PR #100, opened 2022-03-20
Ignore prettier / ember-template-lint in tests

amyrlam:
### :pushpin: Summary

Without this change, when you open an existing test file, there will often be Prettier / ember-template-lint errors.

I don't think that we want Prettier in tests, it seems like overkill and will be very tedious.

TBH I'm not sure why this wasn't happening before. May have come with the monorepo change. I believe the best path forward is to fix it vs investigating the root cause beyond what I timeboxed.

### :camera_flash: Screenshots

WAS:
Existing button/index-test file:
image

Would get Prettier errors in CI, on any new test file: https://github.com/hashicorp/design-system/runs/5595891309?check_suite_focus=true

IS:
No Prettier test errors in /tests/:

image

### :link: External links

https://app.asana.com/0/1201630204387510/1201988425660720/f

***

### 👀 How to review

Reviewer's checklist:

- [N/A] +1 Percy if applicable
- [N/A] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #110, opened 2022-03-22
tell prettier to ignore tests folder

MelSumner:
### :pushpin: Summary

If merged, this PR turns off prettier for the `/tests/` folder (so, tests and dummy app). It's driving me batty, hampering my productivity (rather than helping) and adds extra whitespace in my HTML that makes the documentation appear off for inline code blocks. I endorse using it for our templates in general, but I think it's safe enough to turn off in tests and dummy app.

WDYT?

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/.prettierignore
+++ b/packages/components/.prettierignore
@@ -14,6 +14,7 @@
 /coverage/
 !.*
 .eslintcache
+/tests/
 
 # ember-try
 /.node_modules.ember-try/


amyrlam (commented):
I made this change before and tagged you https://github.com/hashicorp/design-system/pull/100. Can follow up in Slack, but seems like we should close this for now?

MelSumner:
@amyrlam agree; if we decide to do it we should re-open your PR since it came earlier. 👍

PR #87, opened 2022-03-17
CTA component (Spike)

MelSumner:
This is a WIP spike to show how I think the hds:cta component could work.

PR #115, opened 2022-03-23
"CTA" component - Main PR

didoo:
### :pushpin: Summary

This is a follow-up of the [discussion](https://docs.google.com/document/d/1SqsHvv7_Kg5mu9FXdVS1Y7Dw4fbxXYLddNebd4Kd1ic/edit#) around the introduction of a "CTA (Call To Action)" element in the design system, where we came to the decision to introduce this specific (and controversial) component as 

"a _link_ that looks **exactly** as a _primary_ button"

(if it's a button that looks like a button... it's a `Hds::Button`)

_Notice: we have not created a specific CRD (Components Requirement Document) for this component, because is mirroring existing components and would have been a bit of wasted time (we know almost everything about how this CTA component should look and work)._

#### Why it's called "Link(To)::Cta"

Because technically is a link (and we don't want to call it `Button::Cta`, this would lead to even more misunderstanding.
Plus, has to follow the same approach of the `Link(To)::Standalone` component, to be able to differentiate internal vs. external links (this is because of how the internals of Ember work). 

That said, there is a chance that the name in Figma may be slightly different (apart from the _"CTA"_ part). Is something we have initially discussed with @heatherlarsen but we never came to a conclusion. But on the code side this is something that we **strongly** want to keep as is in this PR (unless we find strong reasons for having a different naming).

#### Differences with the `Button (primary)` component

Despite being very similar to the `Button` component, there are some important differences to take into account:
- the `color` argument is pre-selected as primary (it needs to stand out as "call to action")
- it doesn't have a `isIconOnly` option (this would go against being a "call to action")
- it doesn't have a `type` option (it's not a button element)
- it doesn't have the `isDisabled` option (links can't be disabled)
- it needs to responds to the `space` key when selected (being a `<a>` elements that resembles a button, it needs to to mimic the behaviour of a real `<button>` element)

In this PR I have:
- added the `Link::Cta` and `LinkTo::Cta` components as add-ons
  - in doing so I have 
    - re-used parts of the `Link/LinkTo::Standalone` and `Button` implementation in JS
    - completely re-used the CSS styling for the `Button (primary)` using directly the `hds-button` classes within the `Link/LinkTo::CTA` component
    - added a `space` key event listener to the component
- added the documentation page for the two new components
  - in doing so I have:
    - refactored the routing for the `Link/LinkTo::Standalone` components documentation, for consistency
    - added specific documentation for both the `Link::CTA` and `LinkTo::CTA` components in separate pages
      - overview
      - component API
      - usage
      - design guidelines (waiting for docs to be added to Figma)
      - showcase
    - reworked/updated the documentation pages for `Link/LinkTo::Standalone` and `Button` components to be more in sync with the `Link/LinkTo::Cta` ones (now they're all very similar and consistent in content and structure)
- added Percy tests for the new components
- added big alert banners as comments) on top of the files that re-use the styling of the `Button` component to make sure users will notice we’re using code from the “Button” component (I've used [this website](https://patorjk.com/software/taag/#p=display&h=0&c=c%2B%2B&f=ANSI%20Regular&t=NOTICE) to generate the ASCII art)
- added integration tests for the compoents
  - in doing so I have also updated the tests for the `Link/LinkTo::Standalone` for consistency

**Notice:**
- _unlike other components, this has been developed in a single branch, with a single PR (reason being that the component is very similar to other ones, so there was less iteration on it). If you think is a problem, I can break it down in smaller PRs._
- _same for the changes to the `Link::Standalone` and the `Button` documentation pages: if you prefer I can move them to another PR (the reason I didn't is that now they are all very similar, with similar content and structure, so it's easier to compare in the same preview/deploy) _ 
### Alternative solutions
We considered the option to share the `Button` styling using re-usable Sass mixins (and did a spike in code too). 
The `button.scss` file would have looked something like this:
screenshot_1149
but we concluded that it would have been too much abstraction, and the cons would have out-weighted the pros. 

screenshot_1148

### :link: External links

- [[DS-031] A Call To Action (CTA) Component (RFC)](https://docs.google.com/document/d/1SqsHvv7_Kg5mu9FXdVS1Y7Dw4fbxXYLddNebd4Kd1ic/edit#)

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy
- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets)


packages/components/addon/components/hds/link-to/cta.js
```diff
@@ -0,0 +1,179 @@
...
+   * @default null
+   * @description The name of the icon to be used.
+   */
+  get icon() {
+    return this.args.icon ?? null;
+  }
+
+  /**
+   * @param iconPosition
+   * @type {string}
```
amyrlam, line 74: non-blocking question, should be enum (and others)? https://jsdoc.app/tags-enum.html I'm not super familiar with JSDoc, just noticing the diff w/the scrappy page

amyrlam (commented):
It's looking good to me, but will wait to formally approve til the TODOs. Probably good for @MelSumner and @jesdavpet to chime in on eng also!

packages/components/tests/dummy/app/templates/components/button.hbs
```diff
@@ -80,11 +79,7 @@
 <section>
   <h3 class="dummy-h3">Usage</h3>
   <p class="dummy-paragraph">
-    The button component is used to trigger an action or event. They should not be used to route to a URL. This
-    component supports
-    <code class="dummy-code">...attributes</code>
-    .
-  </p>
+    The button component is used to trigger an action or event. They should not be used to route to a URL.</p>
```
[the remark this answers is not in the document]
didoo: 👍

packages/components/addon/components/hds/link-to/cta.js
```diff
@@ -0,0 +1,179 @@
...
+    // we need to register the element to compare it with the one that triggered the "key/space" event
+    this.el = el;
+  }
+
+  // TODO! cleanup this before final review
+  @action
+  onKeySpace(event) {
+    console.log(
+      'This is triggered for every component instance that is present in the page'
+    );
```
MelSumner, line 142:
```suggestion
```

packages/components/tests/dummy/app/templates/components/link-to/cta.hbs
```diff
@@ -0,0 +1,292 @@
...
+    <a
+      href="https://guides.emberjs.com/release/routing/linking-between-routes/#toc_the-linkto--component"
+      target="_blank"
+      rel="noopener noreferrer"
+    >Ember LinkTo</a>
+    component with the same exact visual appearence as the "primary"
+    <LinkTo @route="components.button">Button</LinkTo>
+    component, that points to an internal application link or resource.
+  </p>
+  <p class="dummy-paragraph">
```
MelSumner, line 20: I think we also should say that there should only be one CTA on any given page.

packages/components/tests/dummy/app/templates/components/link-to/standalone.hbs
```diff
@@ -175,74 +121,154 @@
...
+  <p class="dummy-paragraph">Renders to:</p>
+  <Hds::LinkTo::Standalone
+    @icon="film"
+    @iconPosition="trailing"
+    @text="Watch tutorial video"
+    @route="components.link-to.standalone"
+  />
 
-  <h5 class="dummy-h5">Small</h5>
+  <h4 class="dummy-h4">Color</h4>
```
MelSumner, line 178: [question, maybe issue] does this belong here? It's referring to the button component.

packages/components/tests/dummy/app/templates/components/link/standalone.hbs
```diff
@@ -144,65 +129,128 @@
...
     @language="markup"
     @code='
-      <Hds::Link::Standalone @text="Watch tutorial video" @icon="film" href="/" />
+      <Hds::Link::Standalone @icon="film" @text="Watch tutorial video" href="..." />
     '
   />
   {{! prettier-ignore-end }}
   <p class="dummy-paragraph">Renders to:</p>
-  <Hds::Link::Standalone @text="Watch tutorial video" @icon="film" href="/" />
+  <Hds::Link::Standalone @icon="film" @text="Watch tutorial video" href="#" />
```
MelSumner, line 165: This should have thrown an ember-template-lint error; is there a reason that the href value was changed? It should not be "blank" like this.

packages/components/addon/components/hds/link-to/cta.js
amyrlam: Yeah maybe they could go in the future, one source of truth with the scrappy website "Component API" section, as is already drifting). But I'm not sure how they're used by others TBH. No strong opinion

MelSumner (changes requested):
I still think that doing it this way instead of a separate component was both over-engineered (we could have yielded the components in a hash) and locks us out of more flexible CTA designs in the future. But I assume my objections have been noted and I won't block this component moving forward as is. 

I've requested a few changes based on code review.

didoo:
> but will wait to formally approve til the TODOs

@amyrlam fixed all the open TODOs apart from the last two that are not blockers (and likely will be done in a separate PR)

packages/components/addon/components/hds/link-to/cta.js
didoo: removed

packages/components/tests/dummy/app/templates/components/link-to/cta.hbs
didoo: @heatherlarsen can you chime on this specific point? thanks

packages/components/tests/dummy/app/templates/components/link/standalone.hbs
didoo: > This should have thrown an ember-template-lint error
Nope, I didn't see errors.

> is there a reason that the href value was changed? It should not be "blank" like this.
Reason is that, if clicked by mistake while testing, it opens a new window with the index of the dummy website, and it's confusing (using the `#` at least remains on the same page, even if in a new window).

didoo:
> I still think that doing it this way instead of a separate component was both over-engineered (we could have yielded the components in a hash) and locks us out of more flexible CTA designs in the future. 

Can you expand on this? Do you refer to this spike? https://github.com/hashicorp/design-system/pull/87/files 
We discussed this, and I thought it was clear it could not work in that way.

> I've requested a few changes based on code review. 

Commented in all of them, let me know if I addressed your questions.

> I have also provided #116 as a demo that shows a simpler implementation of the spacebar for activation of the CTA (as a link).

Replied in that PR/draft, and ported here the idea of using the `click()` method directly on the active element.

packages/components/tests/dummy/app/templates/components/link-to/standalone.hbs
amyrlam: For some reason I can't add a suggestion below, but I think what Melanie means is to change L180 to "for a LinkTo (Standalone) instead of "for a button"

packages/components/tests/dummy/app/templates/components/link-to/standalone.hbs
```diff
@@ -23,46 +23,7 @@
     <em>Note: Solutions for LinkTo (Inline) are coming soon.</em>
```
amyrlam, line 23: For some reason, can't add suggestion. How about removing this "Note: Solutions for LinkTo (Inline) are coming soon." and/or adding a link to LinkTo (CTA) page before merge? Non-blocking suggestion.

This is an argument to have Percy do the full page, it makes the PR easier to grok, when there's so many changes

amyrlam (approved):
Lint error and a couple docs tweaks to maybe address, but :+1:, I gave the Percy a :+1: as well

MelSumner (approved):

packages/components/yarn.lock
```diff
@@ -0,0 +1,12393 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
amyrlam, line 1: @didoo Just noticed this, there shouldn't be a yarn.lock in anything in /packages/*. Only at the root. Please remove this file before merge 

We could add to the .gitignore, but not sure if solves the root cause of not having the file generated there at all

didoo: I've removed the file.

didoo:
@jesdavpet do you think you will have time to review this and give us thumbs up if it's what you expect for unblocking you in the Cloud UI adoption of HDS components?

didoo:
@heatherlarsen do you think you will have time to review this ticket?

packages/components/tests/dummy/app/templates/components/link-to/cta.hbs
heatherlarsen: I agree that it is too restrictive and hard to enforce, but I do think we can caution against over use in the docs coupled with education on the difference between Buttons, CTAs, and Links. 

@ashleemboyer also had some really good points [in the RFC](https://docs.google.com/document/d/1SqsHvv7_Kg5mu9FXdVS1Y7Dw4fbxXYLddNebd4Kd1ic/edit?disco=AAAAU-E6sYc) about how CTAs get used in our marketing sites which feels important to consider here as well.

heatherlarsen:
@didoo Docs look good. Only think I noticed is that this has the same padding issues as the regular button. Will the updates from #126 land here as well or will the padding have to be adjusted here manually?

didoo:
> @didoo Docs look good. Only think I noticed is that this has the same padding issues as the regular button. Will the updates from #126 land here as well or will the padding have to be adjusted here manually?

Yes, it will automatically be fixed with #126 (is re-using the CSS classes from Button)

heatherlarsen (approved):
lgtm!

packages/components/tests/dummy/app/templates/components/link-to/standalone.hbs
didoo: Ok, I see. No problem. Fixed.

packages/components/tests/dummy/app/templates/components/link-to/cta.hbs
didoo: @heatherlarsen once approved and merged this PR, I'll open a follow-up issue to update the documentation for  Link(To)::CTA, Link(To)::Standalone and Button components, to explain the differences between them, and guide the users in choosing which one to use for different contexts (see first TODO in the PR description) 
That can be a good place where iron out exactly what we want to say to the users/consumers.

didoo:
As discussed with @jesdavpet, we'll wait for @evankerrigan to come back from time off and review this PR too.

evankerrigan:
@didoo & @jesdavpet: 👋 Just catching up from being off and I wanted to double-check if there was specific feedback that you were seeking from me? Not sure if I’m well-suited to offer any useful commentary on this PR’s code changes, but after reading the summary of the changes, it does sound like the proposed changes will offer us a way out of [the challenges outlined here](https://www.figma.com/file/Lu4nv6KKrbv3rMSLLZcfq3/Buttons-%E2%80%98n%E2%80%99-Links?node-id=0%3A1) 🙌

didoo:
@jesdavpet can you chime on this? (see @evankerrigan question above) 
thanks

didoo:
For context: the `ember-keyboard` addon generates **a lot** of deprecation warnings in the console, but I am not sure how to fix this; I have added a comment in the open issue on their repo, let's see if they reply https://github.com/adopted-ember-addons/ember-keyboard/issues/194#issuecomment-1090070179

(not sure what would be a viable alternative, any ideas @MelSumner @hashicorp/design-systems ?)

amyrlam:
I don't think those deprecations are a showstopper, they're somewhat "normal" in Ember as addons needs to make updates as Ember updates. I actually can't find one of the most common ones, `use-destroyables` on https://deprecations.emberjs.com/v3.x/ or https://deprecations.emberjs.com/v4.x/ (the search seems broken so just did a command - F on those pages). Not sure if I'm missing something? I wanted to read about it to see at which version it's not just deprecated (a warning), but no longer works (error).

Reporting it was good, I think it's just repeated cause so many instances of the component on the website.

jesdavpet (approved):
Catching up on the discussion on this PR, looks like @evankerrigan has [approved of these changes](https://github.com/hashicorp/design-system/pull/115#issuecomment-1087796668) from design's perspective.

I _haven't_ reviewed the implementation details for this HDS component, _only_ the documented Ember component API, which looks good to me @didoo. 👍 Nice work!

PR #116, opened 2022-03-23
demo showing spacebar activating link

MelSumner:
### :pushpin: Summary

DO NOT MERGE THIS PR. 

This PR demonstrates the simplest way to add link activation via SPACEBAR.

### DIFF
--- a/packages/components/addon/components/hds/cta.js
+++ b/packages/components/addon/components/hds/cta.js
@@ -1,0 +1,13 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export default class HdsCtaComponent extends Component {
+  @action
+  doRouteTransition(e) {
+    if (document.activeElement) {
+      e.preventDefault();
+      document.activeElement.click();
+    }
+    return;
+  }
+}
--- a/packages/components/addon/components/hds/cta.js
+++ b/packages/components/addon/components/hds/cta.js
@@ -4,7 +4,11 @@ import { action } from '@ember/object';
 export default class HdsCtaComponent extends Component {
   @action
   doRouteTransition(e) {
+    console.log('This is triggered by every instance of the CTA in the page');
     if (document.activeElement) {
+      console.log(
+        'This should be triggered by only the "active" instance of the CTA, but is not (look in the devtools console)'
+      );
       e.preventDefault();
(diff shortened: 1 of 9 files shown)


didoo:
@MelSumner I have pushed a change to your code, to show you what is the issue:
- we're attaching multiple listeners to multiple elements (I have 3 similar CTA in the same page), so when we click the "space" each one of these trigger the event listeners and does something
- in the code as is, we're just checking if there is an `activeElement` not if the active element is the one that corresponds to that instance of the components, so if you look in the console you will see 3 times the same messages, both outside as well as inside the `if (document.activeElement) {` check (makes sense)
- this means we are triggering the `click` on the `activeElement` three times (the number of CTAs we have in the page).
   - some browsers are "smart" (they use heuristics) to understand that if there are three "clicks" on the same link one next after the other, probably is a "bug" and the execute the operation only once (in fact Crome, Safari, Edge they open a single new window)
   - some other browsers may have a different implementation, in which case they will execute the operation three times: if you take Firefox and you select one of the links and click "space" you will see it open three new windows, not just one

So, the solution is to compare the element that triggered the event (that is also the `activeElement`) with the DOM node of the instance of the component, and see if they are the same. if they are, we execute the operation (which is what I did in the CTA code). 

That said, I like the idea of using the `.click()` operation on the element, instead of using `window.open`, it's more correct. I'll adopt it my code.

Let me know if something is not clear, we can pair together if you want.

MelSumner:
@didoo ok, it's just to demo the idea, so it's all good. Since you have what you need I'll close the PR now 👍

packages/components/addon/components/hds/cta.js
```diff
@@ -0,0 +1,17 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+export default class HdsCtaComponent extends Component {
+  @action
+  doRouteTransition(e) {
+    console.log('This is triggered by every instance of the CTA in the page');
+    if (document.activeElement) {
```
MelSumner, line 8: btw @didoo this didn't work correctly because you were missing `hasFocus()`. If you change this line (line 8) to this: 
`if (document.activeElement.hasFocus()) {`
Then this code works. Just thought you would want to know!

PR #94, opened 2022-03-18
Improve readability of the documentation website

didoo:
### :pushpin: Summary

Currently the content in the "scrappy" documentation website is full width, so with large screens it's really hard to read long paragraphs and to scan visually the content. Is better to apply a max-width to it. 

### :hammer_and_wrench: Detailed description

In this PR I have
- applied specific width to the main content
- applied a background to the body (to differentiate the two)

### :camera_flash: Screenshots

**Before:**
before

**After:**
after

***

### 👀 How to review

Reviewer's checklist:

- [ ] +1 Percy (all the screenshots will be udpated)

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #117, opened 2022-03-23
Add z-index isolation to the “Button” component

didoo:
### :pushpin: Summary

In https://github.com/hashicorp/design-system/pull/94 we have changed the layout of the documentation pages. In particular we have applied a background color to the main container, to stand out against the light gray background.

This has surfaced an issue with the focus on the `Button` component, that went undetected until now (it was not noticed in Percy because all the pages were resized, so all tests were failing!).

This is how it looks like now the page:
screenshot_1158
The reason for that is that the pseudo-element used to show the focus has a `z-index: -1` so it goes below the `main` content background (before it was transparent).

In this PR I have:
- added a `z-index` isolation to the “Button” component

This is how it looks after the fix:
screenshot_1159

Notice: I have bumped the package version anyway, for precaution (technically we're changing the behaviour of the button, so better to have a new release for this).

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [x] +1 Percy

### DIFF
--- a/.changeset/gorgeous-singers-punch.md
+++ b/.changeset/gorgeous-singers-punch.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+added a stacking context for the “Button” component so that the focus z-index is isolated in the button
--- a/packages/components/app/styles/components/button.scss
+++ b/packages/components/app/styles/components/button.scss
@@ -21,6 +21,7 @@ $hds-button-focus-border-width: 3px;
   box-sizing: border-box; // TODO https://github.com/hashicorp/design-system-components/issues/46
   display: flex;
   font-family: var(--token-typography-font-stack-text);
+  isolation: isolate;
   justify-content: center;
   outline-style: solid; // used to avoid double outline+focus-ring in Safari (see https://github.com/hashicorp/design-system-components/issues/161#issuecomment-1031548656)
   outline-color: transparent; // We need this to be transparent for a11y


amyrlam (approved):
TIL `isolation` prop 👍 

***

Aside that I just noticed on this PR:

Looks like a pre-existing bug, but the `disabled` state doesn't pass color contrast. I ran Lighthouse in Chrome. (It's tricky to read, because need to ignore the existing issues from ember-prism)

Would suggest that we, separately from this PR, (1) have ember-prism up to a11y and (2) add ember-a11y-testing to the repo if desired, so that can catch color contrast issues in CI

image

didoo:
> Aside that I just noticed on this PR: Looks like a pre-existing bug, but the `disabled` state doesn't pass color contrast. I ran Lighthouse - Accessibility - Desktop in Chrome. (It's tricky to read, because need to ignore the existing issues from ember-prism)

@amyrlam this is the default color of a `<button disabled />` element in HTML, rendered in Chrome:
screenshot_1176
I suspect it doesn't pass the color contrast test either (same for disabled radio/checkbox elements). So I suspect it's not an issue for disabled elements.

Probably @MelSumner @heatherlarsen can comment on this?

MelSumner (approved):
Good catch!

amyrlam:
That totally makes sense, disabled not needing color contrast. I found this: https://sean-elliott.medium.com/a11y-tips-disabled-buttons-and-colour-contrast-f8824d5e9610. Thanks, good to know

PR #126, opened 2022-03-28
Fix internal paddings of the “Button” component

### LINKED ISSUE #120: Bug report: "Button" component sizing is not correct in some cases
## IS: Current behavior
Currently the `medium` (default) variant of the `Hds::Button` component has a computed/rendered height of 38px instead of 36 as in Figma. 

screenshot_1179

Plus, the horizontal padding in the `small` variant is currently `16px` but should be `12px` according to Figma.

The height for the `medium` variant should be **`36px`** as expected in Figma.
The horizontal padding for the `small` variant should be **`12px`** as expected in Figma.

## Possible solution

The reason for this is that we are applying a min-height of `36px` to the button (medium), but also a vertical padding that doesn't take into account the border width (`1px).
There are two options:
- use a `calc()` where we are defining the vertical padding
- slightly reduce the vertical padding (and add a comment explaining why that particular value)

In doing so, better to check also the `large` variant to see it has the same issue.
screenshot_1178

didoo:
### :pushpin: Summary

This will fix #120

In this PR I have:
- fixed the internal paddings of the “Button” component to follow design specs

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy

---
  - https://app.asana.com/0/0/1202033783287268

### DIFF
--- a/.changeset/quick-crabs-check.md
+++ b/.changeset/quick-crabs-check.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+updated the internal padding of the “Button” component to match design specifications
--- a/packages/components/app/styles/components/button.scss
+++ b/packages/components/app/styles/components/button.scss
@@ -99,23 +99,23 @@ $hds-button-focus-border-width: 3px;
 $size-props: (
   "small": (
     "font-size": 0.8125rem, // 13px;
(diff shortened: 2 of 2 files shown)


MelSumner (approved):
👍 :shipit:

PR #127, opened 2022-03-28
Remove the “isDisabled“ prop from the “Button“ component in favour of the native “disabled” attribute

didoo:
### :pushpin: Summary

This PR follows a conversation that happened in the RFC about migration to the `Hds::Button` component in Cloud UI ([see thread here](https://docs.google.com/document/d/1DwnPClfIMdHuGppDIkWJkDCBOJQqgLplHXnh1eI6sVo/edit?disco=AAAATd3VfeI)), in which multiple people raised the question why we were using a custom `isDisabled` prop instead of just relying on the native `disabled` attribute.

I don't remember why we went for the `isDisabled` prop (maybe we wanted to do some stuff with it using a CSS class?) After some thought, I decided it makes perfectly sense to just use the `disabled` attribute (which gets "splatted" via the `...attributes` on the `<button>` element).

In this PR I have:
- removed the “isDisabled” parameter from the “Button” component
- updated accordingly the documentation and the integration tests for the “Button” component

screenshot_1193

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/proud-houses-glow.md
+++ b/.changeset/proud-houses-glow.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+removed the “isDisabled“ prop from the “Button” component in favour of the native “disabled“ attribute
--- a/packages/components/addon/components/hds/button/index.hbs
+++ b/packages/components/addon/components/hds/button/index.hbs
@@ -1,10 +1,4 @@
-<button
-  class={{this.classNames}}
-  ...attributes
-  aria-label={{if this.isIconOnly this.text null}}
-  type={{this.type}}
-  disabled={{if this.isDisabled "disabled" null}}
->
+<button class={{this.classNames}} ...attributes aria-label={{if this.isIconOnly this.text null}} type={{this.type}}>
   {{#if this.isIconOnly}}
     <div class="hds-button__icon">
       <FlightIcon @name={{this.icon}} @size={{this.iconSize}} @stretched={{true}} />
--- a/packages/components/tests/dummy/app/templates/components/button.hbs
+++ b/packages/components/tests/dummy/app/templates/components/button.hbs
@@ -64,12 +64,6 @@
     <dd>
       <p>This indicates that a button should take up the full width of the parent container. The default is false.</p>
     </dd>
-    <dt>isDisabled <code>boolean</code></dt>
-    <dd>
-      <p>Sets the native HTML attribute
-        <em>disabled</em>
-        on the button element. Default is null (doesn't render the attribute).</p>
-    </dd>
     <dt>“splattributes”</dt>
     <dd>
       <p><code class="dummy-code">...attributes</code> spreading is supported on this component.</p>
@@ -227,23 +221,20 @@
   {{! prettier-ignore-end }}
 
   <h4 class="dummy-h4">Disabled buttons</h4>
-  <p class="dummy-paragraph">This is the native button attribute,
-    <code class="dummy-code">disabled</code>. To use this attribute, set
-    <code class="dummy-code">@isDisabled</code>
-    to
-    <code class="dummy-code">true</code>. The default is
-    <code class="dummy-code">null</code>, which means that the attribute will not render at all in the DOM.
+  <p class="dummy-paragraph">To disable a button just use the native
+    <code class="dummy-code">disabled</code>
+    attribute:
   </p>
   {{! prettier-ignore-start }}
   <CodeBlock
     @language="markup"
     @code='
-      <Hds::Button @text="Copy to clipboard" @isDisabled=true />
+      <Hds::Button @text="Copy to clipboard" disabled />
     '
   />
   {{! prettier-ignore-end }}
   <p class="dummy-paragraph">Renders to:</p>
-  <Hds::Button @text="Copy to clipboard" @isDisabled={{true}} />
+  <Hds::Button @text="Copy to clipboard" disabled />
 
   <h4 class="dummy-h4">Full-width buttons</h4>
   <p class="dummy-paragraph">
--- a/packages/components/tests/integration/components/hds/button/index-test.js
+++ b/packages/components/tests/integration/components/hds/button/index-test.js
@@ -108,11 +108,11 @@ module('Integration | Component | hds/button/index', function (hooks) {
     );
     assert.dom('.hds-button__text').doesNotExist();
   });
-  test('it should add the `disabled` attribute to the button if `@isDisabled` is set to true', async function (assert) {
+  test('it should disable to the button if the `disabled` attribute is passed', async function (assert) {
     await render(
-      hbs`<Hds::Button @text="copy to clipboard" @isDisabled={{true}} id="test-button" />`
+      hbs`<Hds::Button @text="copy to clipboard" disabled id="test-button" />`
     );
-    assert.dom('#test-button').hasAttribute('disabled');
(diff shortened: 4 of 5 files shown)


packages/components/addon/components/hds/button/index.hbs
```diff
@@ -1,10 +1,4 @@
-<button
-  class={{this.classNames}}
-  ...attributes
-  aria-label={{if this.isIconOnly this.text null}}
-  type={{this.type}}
-  disabled={{if this.isDisabled "disabled" null}}
->
+<button class={{this.classNames}} ...attributes aria-label={{if this.isIconOnly this.text null}} type={{this.type}}>
```
jesdavpet, line 1: 🤔 💭 @didoo what's your instinct on making this a little bit more flexible in terms of a11y, by allowing consumers of this component to override the `aria-label` explicitly via splattributes? _(BTW I think it makes perfect sense to constrain the supported values for `type`, my question's _only_ about `arial-label`.)_

```suggestion
<button class={{this.classNames}} aria-label={{if this.isIconOnly this.text null}} ...attributes type={{this.type}}>
```

As-is RN, I believe the ordering of this attribute VS splattributes would always override a `arial-label="My special case a11y label ..."`, this might be unexpected now that folks can use `disabled`.

jesdavpet (approved):
Nice, I think this is a great simplification @didoo ! 🙂 👍 

I did have one _(non-blocking)_ question about how `disabled`'s sibling attribute `aria-label` is being treated:
https://github.com/hashicorp/design-system/pull/127/files#r836423040

packages/components/addon/components/hds/button/index.hbs
didoo: @jesdavpet this specific code was created by @MelSumner to force the consumers to add a label when the button is icon-only, but I see your point.

@MelSumner thoughts on this?

jesdavpet: > _"... to force the consumers to add a label when the button is icon-only ..."_

Makes sense, but I don't think they're mutually exclusive. 🙂 The proposed change in ordering would _not_ stop the override from working in the case of an icon-only button ... it would just allow the component consumer to override the `aria-label` if/when necessary.

noelledaley (approved):
thank you for making this fix so quickly!

packages/components/addon/components/hds/button/index.hbs
MelSumner: @jesdavpet I'm not sure I understand your goal, so let's break it down: 

1. right now, the developer has to set the value of `@text` which, if it's an icon-only button, becomes the value of `aria-label`. 

In what cases would you need to override what's already provided? Seems like you're overriding...yourself? Am I missing something?

MelSumner: Also, let's break this out into a different issue so we don't block this one for an unrelated reason.

packages/components/tests/dummy/app/templates/components/button.hbs
```diff
@@ -227,23 +221,20 @@
   {{! prettier-ignore-end }}
 
   <h4 class="dummy-h4">Disabled buttons</h4>
-  <p class="dummy-paragraph">This is the native button attribute,
-    <code class="dummy-code">disabled</code>. To use this attribute, set
-    <code class="dummy-code">@isDisabled</code>
-    to
-    <code class="dummy-code">true</code>. The default is
-    <code class="dummy-code">null</code>, which means that the attribute will not render at all in the DOM.
+  <p class="dummy-paragraph">To disable a button just use the native
```
MelSumner, line 224:
```suggestion
  <p class="dummy-paragraph">To disable a button, manually add the native
```

.changeset/proud-houses-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+removed the “isDisabled“ prop from the “Button” component in favour of the native “disabled“ attribute
```
MelSumner, line 5: [issue] this is misleading, since that's what it already did (provided the native disabled attribute). suggestion: 

```suggestion
removed the “isDisabled“ prop from the “Button” component; added instructions for developers to manually add it themselves if needed.
```

MelSumner (changes requested):
I've requested a few changes, and I think the question raised by @jesdavpet regarding `aria-label` should go in a separate issue (although I'm not inclined to change that design now, as it was done that way for a specific reason).

packages/components/addon/components/hds/button/index.hbs
jesdavpet: @MelSumner the scenario I've raised is that of a component consumer who wants to set a specific value for `aria-label` for a an invocation that _isn't_ "icon-only".

jesdavpet: > _"Also, let's break this out into a different issue..."_

This is the _only_ topic of the issue I started, it's non-blocking.

didoo: @MelSumner what @jesdavpet is suggesting is to allow users to do something like this
```
<Hds::Button @icon="play" @text="this is the visible text" aria-label="this is the additional aria-label that for some reason they may want to add" ... />
```
with the current implementation, this is not possible. I can imagine @jesdavpet position: having to write a codemod that takes care of all possible use cases, and considering the possibility that this may be some strange exception in some parts of the codebase, having it solved before doing the migration would be a plus. if you think it can't be done/decided in the context of this PR, I can open a new one.

didoo:
> I've requested a few changes, and I think the question raised by @jesdavpet regarding `aria-label` should go in a separate issue (although I'm not inclined to change that design now, as it was done that way for a specific reason).

@MelSumner I have made the changes requested, and replied on the topic of Jesse's suggestion. Let me know if you still want this to be discussed/done in a separate PR.

packages/components/addon/components/hds/button/index.hbs
MelSumner: @didoo @jesdavpet ahh yes okay. So we don't allow that because it's a [failure due to the accessible name not containing the visible label text](https://www.w3.org/WAI/WCAG21/Techniques/failures/F96), a violation of [2.5.3- Label in Name](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name).

jesdavpet (approved):

packages/components/tests/integration/components/hds/button/index-test.js
```diff
@@ -108,11 +108,11 @@ module('Integration | Component | hds/button/index', function (hooks) {
...
     assert.dom('.hds-button__text').doesNotExist();
   });
-  test('it should add the `disabled` attribute to the button if `@isDisabled` is set to true', async function (assert) {
+  test('it should disable to the button if the `disabled` attribute is passed', async function (assert) {
     await render(
-      hbs`<Hds::Button @text="copy to clipboard" @isDisabled={{true}} id="test-button" />`
+      hbs`<Hds::Button @text="copy to clipboard" disabled id="test-button" />`
     );
-    assert.dom('#test-button').hasAttribute('disabled');
+    assert.dom('#test-button').isDisabled();
```
MelSumner, line 115: [comment, non-issue] so you could have totally left `.hasAttribute('disabled')` since it's checking for the presence of the HTML attribute. 👍

MelSumner (approved):
👍  :shipit:

PR #91, opened 2022-03-18
Add new Alert component - Part 2

amyrlam:
### :camera_flash: Screenshots

See https://hds-components-git-amy-alert-overall-branch-hashicorp.vercel.app/

### :link: External links

- Placeholder Asana ticket for the rest of the work: https://app.asana.com/0/1201630204387510/1202124690279467
- New base branch. Initial branch, now closed, was https://github.com/hashicorp/design-system/pull/56

***

### 👀 How to review

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable
- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202124690279467

PR #132, opened 2022-03-29
"Alert" component - Dismiss button

didoo:
### :pushpin: Summary

This adds the "dismiss" button to the Alert

In this PR I have:
- added `onClose` hook attached to the "dismiss" button in the “Alert” component
  - this prop is used to determine if the alert is dismissible or not (means showing or not the "dismiss" button)
- fixed vertical alignment of text with icon in case of a single line

#### TODOs in follow up tickets
- [ ] check the accessibility later (events, aria attributes, other relevant things)
    - In Cloud UI `popup` component it has `aria-label="close popup"`

### :link: External links

Other products alerts/notifications:

- Cloud UI: https://unofficial-design-system-search.vercel.app/?path=/story/cloud-ui-storybook_notifications--notifications
- Nomad: https://unofficial-design-system-search.vercel.app/?path=/story/nomad-storybook_components-alerts--dismissal
- Vault: https://unofficial-design-system-search.vercel.app/?path=/story/vault-storybook_alerts-alertpopup--alertpopup
- Boundary: https://unofficial-design-system-search.vercel.app/?path=/story/boundary-storybook_rose-notification--basic-notifications-with-dismissal

***

### 👀 How to review

👉 Review by files changed


amyrlam:
On the spelling, I prefer `hasDismiss` cause it's easier to use. Totally get prefer is...maybe the utility outweighs the confusion. Up to you though!

// 

On the a11y stuff: Keep in mind that the cloud-ui notif (https://unofficial-design-system-search.vercel.app/?path=/story/cloud-ui-storybook_notifications--notifications) is showing an a11y no-no, as I've read and interpreted the specs. Toasts should deliver info to the user, but not actions to the user. So they shouldn't be dismissible.  

(Can see why USWDS and gov.uk don't use any dismiss on their alerts, cause all the alert/notification/banner a11y spec verbiage is somewhat opaque. There is more in the CRD, I only put in the most relevant on the scrappy page for now.)

- For our Alert (Toast), the default isDismissible would be `false`...but we'd allow for it to be true as an escape hatch.
- I don't believe that our Alert (Toast) should have the prescribed hasActions block. If folks want to port over legacy 1:1, with the a11y no-no, they still can in the custom block. AKA "escape hatch but don't recommend" 

If you interpret the a11y specs differently just let me know, just wanted to point this out per the checkboxes above

didoo:
@amyrlam I think this one is ready to go now, can you have a look/review? thanks

packages/components/app/styles/components/alert.scss
```diff
@@ -15,7 +17,15 @@
...
 
+.hds-alert__content {
+  flex: 1 1 auto;
+}
+
 .hds-alert__text {
+  display: flex;
+  flex-direction: column;
+  justify-content: center;
+  min-height: 24px; // same as the icon height, to keep text vertically centered in case is a single line
```
amyrlam, line 28: (non-blocking) Do we want these all in rem? Ditto to 2px below. I am honestly not sure what we want

```suggestion
  min-height: 1.5rem, // 24px, same as the icon height, to keep text vertically centered in case is a single line
```

amyrlam (approved):
looks good, let me take a look at the test failure on some of the ember try's. can go ahead and merge if figure that out so can use this in cloud-ui

amyrlam:
Fixed the test had to be ` @onClose={{"noop"}}` for some of the newer Ember try tests. 

Added your a11y note to https://app.asana.com/0/1201630204387510/1202057405183445/f

Merging in.

didoo:
> Fixed the test had to be ` @onClose={{"noop"}}` for some of the newer Ember try tests.

@amy the correct syntax was the previous one `@onClose={{noop}}` because `noop` is a helper/function, not a string (otherwise the `this.onClose` will always return `false` and the dismiss button will never appear). But apparently there are problems with the syntax, whatever variants one use (see thread here: https://hashicorp.slack.com/archives/C11JCBJTW/p1648751235987409)

I am still trying to fix this in an alternative way, but for now without success. I'll keep you posted.

didoo:
@amyrlam fixed here https://github.com/hashicorp/design-system/pull/91/commits/09b892f50e79526a54cec6a1f3f8b27c7ce64045

amyrlam:
thx

PR #140, opened 2022-03-30
remove disabled state for dropdown toggles

MelSumner:
### :pushpin: Summary

If merged, this PR removes disabled state for dropdown toggles (per design)

### 👀 How to review

👉  Review commit [89a2a75](https://github.com/hashicorp/design-system/commit/89a2a75a3270e90014d0967cde0e505509422066)

Reviewer's checklist:

- [ ] +1 Percy if applicable

---
  - https://app.asana.com/0/0/1202045877853559

### DIFF
--- a/packages/components/addon/components/hds/dropdown/index.hbs
+++ b/packages/components/addon/components/hds/dropdown/index.hbs
@@ -10,10 +10,11 @@
   </:toggle>
   <:content>
     <ul class="hds-dropdown-list hds-dropdown-list--absolute-right">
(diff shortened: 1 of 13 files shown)


didoo (approved):
OK for me. 

My instinct tells me we're gonna need to re-introduce the `disabled` state, expecially if we start to support also the use case of `icon+chevron` as in the image below
screenshot_1215
but we can always re-introduce later, it would not require a huge refactoring.

PR #142, opened 2022-03-30
removed reference to copy-item in docs

MelSumner:
## :pushpin: Summary

If merged, this PR removes the reference to copy-item in the docs. I didn't remove it from the template since I still plan to implement in a separate PR.

## Reviewer's checklist:

- [ ] +1 Percy if applicable

---
  - https://app.asana.com/0/0/1202045371056987


didoo (changes requested):
@MelSumner, it should also be removed from the JS/HBS code (do a general search for `copy-item` in the codebase, you will find more instances)

PR #150, opened 2022-03-31
Cleanup the "box-sizing: border-box" declarations from the codebase

didoo:
### :pushpin: Summary

While developing the Ember components, in many cases we had to add a `box-sizing: border-box` declaration in the styling of a UI element to make sure that the this was the box model used by the browser to render that element. 
We have discussed many times with @MelSumner of how to solve this problem once for all (see for example https://github.com/hashicorp/design-system-components/issues/46) but we never made a decision.

In this PR I make this proposal: we assume that every consumer of the Ember components has a "standard" box-sizing reset/normalize global declaration in their CSS:
```
*, *::before, *::after {
  box-sizing: border-box;
}
```
so that we don't have to add it every time in our components. The reason is that this has become a standard de facto, and to make sure this assumption was correct I asked to the front-end developers if all the applications were using this reset.
Their answer was positive (see [this thread in Slack](https://hashicorp.slack.com/archives/C11JCBJTW/p1648639657008599)) so can safely make this assumption and remove all the box-sizing declarations from our codebase (we will also need to make sure that our websites, the "dummy" app and the Flight website, contain this reset).

In this PR I have:
- added a global `box-sizing: border-box` reset in the CSS of the dummy app + updated the size of main “documentation” page container
   - this was to check that the payout of the pages remained the same, so any diff in Percy would have meant a diff in the components, not in the layout of the containing page
- removed any redundant `box-sizing` declaration from the CSS of dummy app
- removed any redundant `box-sizing` declaration from the CSS of components
- removed any redundant `box-sizing` declaration from the CSS of Flight website
  - notice: the `box-sizing` declaration in this case is already contained in the existing resets (see #151)
- updated the documentation to highlight that we assume the `box-sizing` reset for our components

_Notice: we don't expect any difference in the consuming codebases, once this release is adopted, but for precaution we declare this as a _minor_ version change (which in the current semver strategy is a breaking change)_

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy
    - 🚨 Notice: there a couple of failing test, but only on very small screens (375px) and are due to some strange reflow in the previous version in a specific page (colors) on a specific dummy component used to visualize these colors, they're not about the HDS components (also, the correct ones are the new screenshots)

/cc @hashicorp/design-systems 

---
  - https://app.asana.com/0/0/1202055821223298

### DIFF
--- a/.changeset/curly-emus-begin.md
+++ b/.changeset/curly-emus-begin.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+removed “box-sizing“ declarations from the components (we assume the consumers codebase already have set it to “border-box“ by default
--- a/packages/components/README.md
+++ b/packages/components/README.md
@@ -47,6 +47,8 @@ Finally, add this line to the top of your app's style file (`app.scss` or simila
 @import '@hashicorp/design-system-components';
 ```
 
+⚠️ **Notice**: our component library assumes that a `*, *::before, *::after { box-sizing: border-box; }` reset is applied globally in the CSS of the application. If in your use case this is not true, please speak with the design system team (we can try to find a workaround).
+
 Usage
 ------------------------------------------------------------------------------
 
--- a/packages/components/app/styles/components/badge-count.scss
+++ b/packages/components/app/styles/components/badge-count.scss
@@ -15,7 +15,6 @@ $hds-badge-count-border-width: 1px;
 .hds-badge-count {
   align-items: center;
   border: $hds-badge-count-border-width solid transparent;
-  box-sizing: border-box;
   display: inline-flex;
   font-family: var(--token-typography-font-stack-text);
   max-width: 100%;
--- a/packages/components/app/styles/components/badge.scss
+++ b/packages/components/app/styles/components/badge.scss
@@ -16,7 +16,6 @@
   align-items: center;
   border-radius: 5px;
   border: $hds-badge-border-width solid transparent;
-  box-sizing: border-box;
   display: inline-flex;
   max-width: 100%;
 }
--- a/packages/components/app/styles/components/button.scss
+++ b/packages/components/app/styles/components/button.scss
@@ -18,7 +18,6 @@ $hds-button-focus-border-width: 3px;
   align-items: center;
   border: $hds-button-border-width solid transparent; // We need this to be transparent for a11y
   border-radius: $hds-button-border-radius;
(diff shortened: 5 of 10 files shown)


jgwhite:
I’m supportive of this. Popular CSS libraries like [Tailwind](https://tailwindcss.com/docs/preflight) and [Normalize](https://github.com/sindresorhus/modern-normalize) set `box-sizing: border-box` globally, so we’ll be in good company.

johncowen:
Hey @didoo did I understand right here that you add the reset to your dummy app but don't provide this reset as part of the design system for consumers to use?

didoo:
> Hey @didoo did I understand right here that you add the reset to your dummy app but don't provide this reset as part of the design system for consumers to use?

Yes @johncowen we assume that the applications will already have this reset. It's a big assumption, but at the same time is very realistic (who would work in a different box model in 2022?)

johncowen:
Yeah I think we are on the same page with this, I'm just thinking from my/Consul's perspective, I would rather remove this reset from my app and then consume it from the design system itself, so if this reset was a little file somewhere in the design system package I could just import it from there and remove it from my app (your dummy app could also do the same)

meirish:
I think if you're going to assume it's there, you should provide what the expected values should be - it sounds like you've already found a case where there were slightly differing resets in different code bases so providing would clear that up.

didoo:
This is the only **global** reset we are likely to define (need).

johncowen:
Good opportunity here for me to use someones else's comment as a TLDR; 😸 

> I think if you're going to assume it's there, you should provide what the expected values should be

Also, just to clarify my thoughts

- This 'box-sizing' reset is the only thing I'm discussing right now, not a completely different much more featured reset like a normalize or anything else similar
- The design system depends on this box-sizing rule. If a project doesn't have it then I would guess things won't display properly? i.e. this box-sizing rule is a dependency of the design system. Did I understand that right?
- We all use this box-sizing reset already, some via application level code, some via third party code (e.g. bulma) and I'm assuming the end goal here is to remove things like bulma anyway
- From Consul's standpoint I would rather depend on this from the design system than have it as application level code, but not as a global 'reset' stylesheet, just as a single importable file with this almost single rule in it. Others can optionally import it also if they wanted to or not import it. Either way, as I think you said up the top in the description the design system components will depend on this rule being present. It's a pretty hard dependency, granted a very small one.

All in all I'm fine either way, but I figured if we are saying you _must_ have this rule defined, that's not that far away from the dependency anyway, we may as well let folks include it from the design system if they wanted to.

PR #160, opened 2022-04-01
refactored list-item to support button elements

MelSumner:
## :pushpin: Summary

If merged, this PR refactors the list-item.

## :hammer_and_wrench: Detailed description

- updated `linkTo` and `link` to make sure they include everything they need
- updated the `is-{{state}}` class conditional so it renders correctly
- added support for buttons and links as list-items
- updates list style to support `min-width` of 200px and `max-width` of 400px (per design)

## :camera_flash: Screenshots

image

## 👀 How to review

👉  Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

---
  - https://app.asana.com/0/0/1202058001063648


packages/components/addon/components/hds/dropdown/list-item.js
```diff
@@ -55,6 +55,7 @@ export default class HdsDropdownListItemComponent extends Component {
   get item() {
     let { item = DEFAULT_ITEM } = this.args;
 
+    // do we need this assert anymore?
```
didoo, line 58: it depends :)

packages/components/app/styles/components/dropdown.scss
```diff
@@ -169,7 +170,22 @@ $hds-dropdown-toggle-border-width: 1px;
     isolation: isolate; // used to create a new stacking context (needed to have the pseudo element below text/icon but not the parent container)
     min-height: 36px;
 
-    a {
+    // need to reset a few extra things to make the button visually appear the same as the links
+    // also CR if you want to re-factor, that's fine; just trying to make this work in this first pass
```
didoo, line 174: @MelSumner I can't refactor until we remove the nesting of classes (needs to be done in a separate PR)

packages/components/addon/components/hds/dropdown/list-item.hbs
```diff
@@ -14,28 +14,46 @@
...
             <FlightIcon @name={{@icon}} />
           </div>
         {{/if}}
-        <div class="hds-dropdown-list-item__link_text">
+        <div class="hds-dropdown-list-item__interactive-text">
           {{@text}}
         </div>
       </a>
+    {{else}}
+      <button class="{{if @state (concat 'is-' @state)}}" type="button" ...attributes>
```
didoo, line 47: it may be we will need to define a specific hook/callback handler for the button, but let's see the final API/implementation of index before deciding

packages/components/addon/components/hds/dropdown/list-item.hbs
```diff
@@ -14,28 +14,46 @@
...
+        {{#if @icon}}
+          <div class="hds-dropdown-list-item__interactive-icon">
+            <FlightIcon @name={{@icon}} />
```
didoo, line 50: just noticed now, but all the `<Flight>` icon declarations under the `interactive-icon` list-item element need to have the `@isInlineBlock={{false}}` (otherwise the vertical alignment is not correct).

packages/components/app/styles/components/dropdown.scss
```diff
@@ -169,7 +170,22 @@ $hds-dropdown-toggle-border-width: 1px;
...
+    // TODO this is 2px taller than the link...why?
+    button {
+      box-sizing: border-box;
```
didoo, line 177:
```suggestion
      box-sizing: border-box;
```
all these `box-sizing: border-box` declaration can be removed now (after merging `main`)

packages/components/app/styles/components/dropdown.scss
```diff
@@ -169,7 +170,22 @@ $hds-dropdown-toggle-border-width: 1px;
...
+      border: 1px inset transparent; // this might be the cause of the extra height?
```
didoo, line 178: yes, this is the cause. we have to understand how we can deal with it

packages/components/addon/components/hds/dropdown/list-item.hbs
```diff
@@ -14,28 +14,46 @@
...
+        @current-when={{@current-when}}
+        @models={{hds-link-to-models @model @models}}
+        @query={{hds-link-to-query @query}}
+        @replace={{@replace}}
+        @route={{@route}}
+        ...attributes
+      >
         {{#if @icon}}
-          <div class="hds-dropdown-list-item__link-icon">
+          <div class="hds-dropdown-list-item__interactive-icon">
```
didoo, line 27: not a big fan of the name `interactive` for this element (maybe `action`?)

packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -671,7 +669,40 @@
...
+            @state={{state}}
+          />
+        {{/each}}
+      </ul>
+    </div>
+  {{/each}}
+
+  <h4 class="dummy-h4">Buttons</h4>
+
+  {{#each @model.ITEM_COLORS as |color|}}
```
didoo, line 683: probably we can merge the two blocks in a single one, with another loop outside. if you want I can try in a separate PR

packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -671,7 +669,40 @@
           <Hds::Dropdown::ListItem
             @route="components.dropdown"
             @icon={{if (eq color "critical") "trash" "settings"}}
-            @text="{{state}} with a long multi-line text"
+            @text="{{state}} with a longer text string that may wrap since max-width is defined on the container"
```
didoo, line 672: when I read this, it was confusing for me (I thought the container was an external wrapper; here it's actually the list item itself, so maybe we should say something about the fact that the item has a max width?
```suggestion
            @text="{{state}} with a longer text string that may wrap since the list item has a max-width property applied to it"
```

didoo (approved):
@MelSumner I have added a few comments, but I think the best thing would be to merge this branch, and then do a new pass on the entire code and see what is missing or needs to be cleaned up, refactored, changed, etc (like we did for the https://app.asana.com/0/1202008719910219/1202025538866553 ticket)

packages/components/addon/components/hds/dropdown/list-item.hbs
MelSumner: since "link" is explicitly related to an `<a>` element, and "action" with a `<button>`, we settled on "interactive". If there is another word that is not already tied to one element or the other, but would work generically, we could consider, but let's do that separately.

packages/components/tests/dummy/app/templates/components/dropdown.hbs
MelSumner: that's not technically accurate; it's the container (the `<ul>` that has the sizing: 

```scss
  .hds-dropdown-list {
    ...
    min-width: 200px;
    max-width: 400px;
  }
```

But, we're still working on the proportions and converting these to relative units, so I think we should fine-tune the wording in the last pass of the docs. WDYT?

packages/components/addon/components/hds/dropdown/list-item.hbs
MelSumner: ok I'll do that in a new PR. .

PR #170, opened 2022-04-04
added support for list positioning

MelSumner:
### :pushpin: Summary

If merged, this PR adds support for `@listPosition` so that users can indicate how they want the list positioned. Default position is to the right; if `@listPosition="left"` is added when invoking the component, the list will be positioned on the left. 

I used a string value instead of a boolean because it gives us future flexibility (i.e., adding `top` (left/right) and `bottom` (left/right) variations).

Reviewer's checklist:

- [ ] +1 Percy if applicable

---
  - https://app.asana.com/0/0/1202045371056988


packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -11,6 +11,18 @@
...
   <h3 class="dummy-h3" id="section-component-api">Component API</h3>
+  <dl class="dummy-component-props">
+    <dt>listPosition <code>string</code></dt>
+    <dd>
+      <p>Acceptable values:</p>
+      <ol>
+        <li>left</li>
+        <li class="default">right</li>
+      </ol>
+      <p>If no <span class="dummy-code">listPostion</span> is defined, it will automatically be positioned to the right.</p>
```
didoo, line 22: This will go in the "How to use" section. Here it is already implicit by the fact that `right` is the _default_ value.

```suggestion
```

didoo (approved):

packages/components/addon/components/hds/dropdown/index.hbs
```diff
@@ -14,7 +14,7 @@
     {{/if}}
   </:toggle>
   <:content>
-    <ul class="hds-dropdown-list hds-dropdown-list--absolute-right">
+    <ul class="hds-dropdown-list {{if (eq @listPosition "left") "hds-dropdown-list--position-left" "hds-dropdown-list--position-right"}}">
```
didoo, line 17: [suggestion] this is even more compact (and easy to read, and future-proof).

```suggestion
    <ul class="hds-dropdown-list hds-dropdown-list--position-{{@listPosition}}">
```

MelSumner: This would mean that it always needs to be explicitly defined. If not defined, it would be `hds-dropdown-list--postiion-` 

It would also create the same issue as the `is-{{state}}` classes created (since fixed).

packages/components/tests/dummy/app/templates/components/dropdown.hbs
MelSumner: I'll remove these in a separate PR; right now all of the API definitions have a sentence like this. 👍

packages/components/addon/components/hds/dropdown/index.hbs
MelSumner: Created issue: https://github.com/hashicorp/design-system/issues/173

PR #174, opened 2022-04-04
Update changesets to use updateInternalDependents

amyrlam:
### :pushpin: Summary

Re: https://github.com/changesets/changesets/blob/fcf5e46d2c2b5644db08af0463dfe4c143b21a8d/packages/types/CHANGELOG.md#major-changes-1

With this change, a patch bump of flight-icons should trigger a bump of ember-flight-icons (and then design-system-components)

### :link: External links

https://app.asana.com/0/1201630204387510/1202073042122436/f

***

### 👀 How to review

Reviewer's checklist:

- [N/A] +1 Percy if applicable
- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #176, opened 2022-04-04
Remove Percy from ember try testing 

amyrlam:
### :pushpin: Summary

Follow-on to https://github.com/hashicorp/design-system/pull/174

***

Reviewer's checklist:

- [N/A] +1 Percy if applicable

### DIFF
--- a/packages/components/tests/acceptance/percy-test.js
+++ b/packages/components/tests/acceptance/percy-test.js
@@ -2,9 +2,16 @@ import { module, test } from 'qunit';
 import { visit } from '@ember/test-helpers';
 import { setupApplicationTest } from 'ember-qunit';
 import percySnapshot from '@percy/ember';
+import config from 'dummy/config/environment';
 
 module('Acceptance | Percy test', function (hooks) {
   setupApplicationTest(hooks);
+
+  if (config.emberTryScenario) {
+    // eslint-disable-next-line no-console
+    console.log('Not running percy in ember-try');
+    return;
+  }
 
   test('Take percy snapshots', async function (assert) {
     await visit('/foundations/elevation');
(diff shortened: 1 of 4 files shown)


amyrlam:
I think we shouldn't run Percy in ember-try: https://github.com/percy/percy-ember/issues/4, need to figure it out in GitHub Actions

amyrlam:
strangely this passes locally

amyrlam:
CI error was also on https://github.com/hashicorp/design-system/pull/177

Dhaulagiri:
[question] is percy even running during these tests? I see for the other scenarios we have to actually append `:percy` to the test invocation to have those generate snapshots for percy.

packages/components/tests/acceptance/percy-test.js
```diff
@@ -2,10 +2,17 @@ import { module, test } from 'qunit';
...
+import config from 'dummy/config/environment';
 
 module('Acceptance | Percy test', function (hooks) {
   setupApplicationTest(hooks);
 
+  if (config.emberTryScenario) {
+    // eslint-disable-next-line no-console
+    console.log('Not running percy in ember-try');
+    return;
+  }
```
amyrlam, line 14: I think we don't need this in flight since the ember-try isn't present in `packages/flight-website`

amyrlam:
@Dhaulagiri So it runs the acceptance test that runs Percy, like here on `main` https://github.com/hashicorp/design-system/runs/5820800160?check_suite_focus=true#step:5:33 (not the snapshots though) ... the console.log in this PR makes it more clear?

I guess could try to completely skip the test altogether in ember-try if desired?

Dhaulagiri (approved):

PR #180, opened 2022-04-04
add copy label/item

MelSumner:
### :pushpin: Summary

If merged, this PR adds support for `copy-label` and `copy-item` as `list-item` items.

TODO:
- [x] check proportions with @heatherlarsen
- [x] add state styles for copy-item
- [x] copy-item focus state still needs doing, it's ridiculously complex
- [x] copy-item also wants a success state and a different icon (!)
- [x] add invocation support in `index.hbs`
- [x] add clipboard support ([web API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API))

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

---
  - https://app.asana.com/0/0/1202025538866561
  - https://app.asana.com/0/0/1202049123133276


packages/components/addon/components/hds/dropdown/list-item.hbs
```diff
@@ -25,7 +43,7 @@
       >
         {{#if @icon}}
           <div class="hds-dropdown-list-item__interactive-icon">
-            <FlightIcon @name={{@icon}} />
+            <FlightIcon @name={{@icon}} @isInlineBlock={{false}} />
```
didoo, line 46: [issue] the same change needs to be done at lines `57` and `68`

PS: see why long time ago I told you that I was worried that non abstracting these replicated code would lead to these errors? I suggested using handlebars partials, do you think this would be possible?

packages/components/addon/components/hds/dropdown/list-item.js
```diff
@@ -86,4 +97,12 @@ export default class HdsDropdownListItemComponent extends Component {
 
     return classes.join(' ');
   }
+
+  @action
+  copyCode() {
+    navigator.clipboard.writeText(this.text);
+    if (navigator.clipboard.readText) {
+      this.isSuccess = true;
+    }
```
didoo, line 106: [issue] a couple of issues: 
- this doesn't work for me in both Safari and Chrome: the text is copied but the style is not "success" (remains the same as before being copied).
- maybe we should compare the value of the `readText` with the value of `this.text` to make sure it's really a "success"? we may have a case in which we have already something in clipboard, but is old and for some reasons the `writeText` failed, but the `readText` doesn't fail because there's already the old value available

packages/components/app/styles/components/dropdown.scss
```diff
@@ -25,311 +25,385 @@ $hds-dropdown-toggle-base-height: 36px;
...
+
+  img {
+    border-radius: inherit;
+  }
+}
+
+.hds-button__icon { // TODO this class is not being used.
+  margin-right: -6px; // we apply a negative margin to counter the padding-right of the button and reduce the visual space between the icon and the right border
+  margin-left: 0.5rem; // this overrides the rule `.hds-button__text + .hds-button__icon`
+}
```
didoo, line 90: this can be removed, it was in the `.hds-dropdown-toggle--with-button-component` (see lines `100-103`)

packages/components/app/styles/components/dropdown.scss
```diff
@@ -25,311 +25,385 @@ $hds-dropdown-toggle-base-height: 36px;
...
+      .hds-dropdown-list-item__copy-item-icon {
+        color: var(--token-color-foreground-success);
+      }
+    } 
+
+    .hds-dropdown-list-item__copy-item-text {
+      font-size: var(--token-typography-code-100-font-size);
+      font-weight: var(--token-typography-font-weight-regular);
+      line-height: var(--token-typography-code-100-line-height);
+      max-width: 250px; // TODO we should be able to figure out the proportions here
```
didoo, line 230: you can remove this and it will automatically work without a `max-width` declaration

packages/components/app/styles/components/dropdown.scss
```diff
@@ -25,311 +25,385 @@ $hds-dropdown-toggle-base-height: 36px;
...
+      overflow: hidden;
+      text-align: left;
+      text-overflow: ellipsis;
```
didoo, line 233: [issue/blocker]: here is missing `white-space: nowrap;` to work correctly (otherwise it will be multi-line)

packages/components/app/styles/components/dropdown.scss
```diff
@@ -25,311 +25,385 @@ $hds-dropdown-toggle-base-height: 36px;
...
     }
 
-    .hds-dropdown-toggle-icon__chevron {
-      margin-left: 0.25rem;
+    .hds-dropdown-list-item__copy-item-icon {
+      align-self: flex-end;
```
didoo, line 237: you're setting `justify-content: space-between;` in the parent flex, so you don't need this

packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -119,10 +119,12 @@
...
       </p>
       <ol>
-        <li>title</li>
+        <li class="default">interactive</li>
+        <li>copy-item</li>
+        <li>copy-label</li>
         <li>description</li>
         <li>separator</li>
-        <li class="default">interactive</li>
+        <li>title</li>
```
didoo, line 127: [consistency] can we have the same order everywhere?

didoo (changes requested):
@MelSumner I have added a few comments (some of them are blockers)

I couldn't review properly the CSS because of the huge changes in the un-nesting of the classes (should have been done in a separate PR, remember?)

For other things, we'll do a final review once everything is merged and we can see the full picture.

packages/components/addon/components/hds/dropdown/list-item.js
MelSumner: ok, I'll work on it.

packages/components/app/styles/components/dropdown.scss
MelSumner: I'll double check, I had to add this purposefully before.

PR #182, opened 2022-04-05
Update docs following monorepo changes

amyrlam:
### :pushpin: Summary

- Update docs following monorepo changes
- Add info about ember install

No pressure to merge this, wanted to make these doc changes following the monorepo changes to close the loop on that work. I believe that the table at the top makes it more clear for consumers what's happening here. Happy to make any edits you prefer, also feel free to suggest / commit directly to the branch if desired!

### :link: External links

https://app.asana.com/0/1201630204387510/1201967845133303/f

***

### 👀 How to review

Best way to review is to check out https://github.com/hashicorp/design-system/blob/286780a0d7bb0f12b708be7a6a844ba3d4d254aa/README.md, or

👉 Review commit-by-commit

Reviewer's checklist:

- [N/A] +1 Percy if applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,14 +1,26 @@
 # HashiCorp Design System
 
-## Workspaces
+## Packages
+
+| NPM package | Version | Website | Local package README |
+|---|---|---|---|
+| **@hashicorp/design-system-components** <br />Design System components in Ember.js | [![npm version](https://badge.fury.io/js/%40hashicorp%2Fdesign-system-components.svg)](https://badge.fury.io/js/%40hashicorp%2Fdesign-system-components) | [design-system-components-hashicorp.vercel.app](https://design-system-components-hashicorp.vercel.app) | [components/README](packages/components/README.md) |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,56 +46,39 @@ yarn changeset version
...
-| ------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------------- |
-| **[@hashicorp/design-system-components](./packages/components/)** <br />Design System components    | [![npm version](https://badge.fury.io/js/%40hashicorp%2Fdesign-system-components.svg)](https://badge.fury.io/js/%40hashicorp%2Fdesign-system-components) |
-
-### Contributing
+```bash
+yarn workspace <workspace-package-name> run ember install <package>
+```
 
-See [components/CONTRIBUTING](packages/components/CONTRIBUTING.md) guide for details on how to update the components.
+e.g. `yarn workspace components run ember install ember-a11y-refocus`
```
Dhaulagiri, line 77: Two thoughts here:
* I think moving this into the first section would make sense as that has the other instructions for working with monorepo whereas the sections right above this are about the sub-packages.
* Because yarn 1 is broken this command half works in that it installs the package, but it also generates a `yarn.lock` in the package folder instead of from the root. So if we want to mention this we should also mention that you'll need to remove that file and manually run `yarn` in the root

Dhaulagiri: also, the `workspace` is the name of the package, not the folder. so something like:

`yarn workspace @hashicorp/design-system-components run ember install ember-a11y-refocus`

amyrlam: good catch, made those changes, ready for another look: https://github.com/hashicorp/design-system/blob/3cf62fa36361a79ea2e719929645ad1d637603e2/README.md

Dhaulagiri (approved):

PR #186, opened 2022-04-06
Alert a11y and docs

amyrlam:
### :pushpin: Summary

Take a look, let me know if any feedback
- Seems like this could be merged into `amy/alert-overall-branch` and left an in-scrappy website TODO for Melanie to ok the a11y when back in office.

### :link: External links

https://app.asana.com/0/1201630204387510/1202057405183441/f

***

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ignore percy's, can update https://github.com/hashicorp/design-system/tree/amy/alert-overall-branch with `main`, after this is merged into https://github.com/hashicorp/design-system/tree/amy/alert-overall-branch] +1 Percy if applicable
- [N/A, not for merge into main] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

---
  - https://app.asana.com/0/0/1202064552765978


packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -443,19 +443,22 @@
...
 <section>
-  <h3 class="dummy-h3">🚧 Accessibility: Alert (Page), Alert (Inline), and Alert (Compact)</h3>
-  <p class="dummy-paragraph">
-    The Alert (Page) has
-    <code class="dummy-code">role="alert"</code>
-    on the parent
-    <code class="dummy-code">div</code>, to identify the element as the container where alert content will be added or
-    updated.
-  </p>
+  <h3 class="dummy-h3">🚧 Accessibility</h3>
```
amyrlam, line 446: Wanted to specifically call out this section, from the preview link https://hds-components-git-amy-alert-a11y-hashicorp.vercel.app/components/alert:

image

LMK what you think, needs some more verbiage and logic in code @didoo @cveigt @MelSumner

cveigt: > When the dismiss button is focused, it should close on return / enter
 should it be `return / enter` or `space`?

I've seen that some design systems also use role=status depending on the alert type. I'm going to read a bit more tomorrow. [IBM's Carbon](https://www.carbondesignsystem.com/components/notification/accessibility) is one of them.

Also, this section does not include AlertToast, right?

packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -403,19 +443,22 @@
...
+    <li>color="warning" or color="critical" => role="alert", can have dismiss.</li>
+    <li>color="neutral", color="highlight" or color="success" => role="status", can have dismiss.</li>
+    <li>Dismiss button, type="button"</li>
+    <li>Dismiss has a focus state (e.g. tab to the button)</li>
+    <li>When the dismiss button is focused, it should close on return / enter (Note: Close on ESC, would be for modal or a classic role="alertdialog" that is an alert-like modal)</li>
+    <li>I think it's arguable if the dismiss needs a hover or active state. The hover state seems good though, especially for a static alert (e.g. not a disappearing toast).</li>
+    <li>When the alert has :actions, the role will change to role="alertdialog", as it's requiring a response from the user. role="alertdialog" seems to typically apply to a "modal" alertdialog, but not always. Some DS's refer to all of their alerts with role="alertdialog" instead of role="alert" or role="status".</li>
+  </ul>
   <p class="dummy-paragraph">References:</p>
+  🚧 TODO: Confirm a tag titles for all of these
```
MelSumner, line 457: I don't think we should refer to authoring practices but the success criteria themselves. Authoring practices are examples written by contributors. I left some comments in the alert CRD for the relevant section.

packages/components/tests/dummy/app/templates/components/alert.hbs
amyrlam: @cveigt Let me check on `space`. Where did you see that? I've been doing a lot of keyboard tabbing cause the spec doesn't always list (each one seems pretty scoped). I *think* Enter is for PC.

Am suggesting role="status" for neutral, highlight, and success

For Toast...it's sort of tricky. I think it's arguable that all Toasts could be an alertdialog. Seems like most of the ones we have, have actions => alertdialog. But I think the type is the determinant

Also that reminds me, should check on the Toast dismiss, if it's ESC.

amyrlam: @cveigt Just met with Cristiano and Brian on an eng handoff plan for before April 16. Since we are using the button you are totally right about space. (And we can stick with hover, focus, active to match the button, which is what's on the preview link rn,)

I'm going to flesh out the a11y more, but will do some testing in cloud-ui first. I can still update this by EOD and then will be out Thurs & Fri, here all next week before transferring.

There may be some Toast-specific a11y recommendations (how long before it dismisses, details about the animation), but we won't provide the flash service AFAICT. Specifically will work on a 1:1 swapout of https://cloud-ui-storybook-hashicorp.vercel.app/?path=/story/notifications--notifications for Alert (Toast). Also we can talk at standup

packages/components/tests/dummy/app/templates/components/alert.hbs
amyrlam: Removed

packages/components/tests/dummy/app/templates/components/alert.hbs
amyrlam: updated ^, resolving

amyrlam:
Merging this into the parent branch, since is not main. Most things should be self-explanatory from the scrappy site, but will leave a few notes in a placeholder Asana ticket linked to the parent non-main branch. Thanks for taking this work over

PR #199, opened 2022-04-09
Docs: Expand on the tertiary button icon requirement

DingoEatingFuzz:
### :pushpin: Summary

This describes why tertiary buttons without an icon are deemed inaccessible and links to the WCAG 2.1 techniques for the criterion.

I still have reservations on whether or not this should actually be a restriction on the component.  The applicability section for the [use of color criterion techniques](https://www.w3.org/WAI/WCAG21/Techniques/general/G182.html) states:

> Colored text when the color is used to convey information such as:
> - Words that are links in a paragraph
> - Items in a list where some are different than others and are presented in colored text

Since buttons should never be buried in paragraphs of text like an inline link would be, the criterion shouldn't be applicable--or at the very least so rarely applicable that is shouldn't be a hard restriction encoded in the `Hds::Button` component.

And from the description of the technique on the same page:

> The intent of this technique is to provide a redundant visual cue for users who may not be able to discern a difference in text color.

Interpreting this intent, the position of a button outside the flow of prose is itself a way to discern the difference between text and the button.

Nonetheless, if this restriction is going to be in place on the component, I doubt I'll be the last person to wonder why, so having docs link to the official source should be valuable!

### :link: External links

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/tests/dummy/app/templates/components/button.hbs
+++ b/packages/components/tests/dummy/app/templates/components/button.hbs
@@ -35,9 +35,19 @@
     <dt>icon <code>string</code></dt>
     <dd>
       <p>Use this parameter to show an icon. Acceptable value: any Flight icon name.</p>
-      <p><strong>🚨 IMPORTANT:</strong>
+      <p><strong>🚨 IMPORTANT a11y note:</strong>
         <code class="dummy-code">tertiary</code>
-        buttons are required to have either a leading or trailing icon to be accessible.</p>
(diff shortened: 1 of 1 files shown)


didoo (approved):
@DingoEatingFuzz OK for me (left a comment about replacing the HDS component with a normal link). 

Thanks for the contribution 🙏

amyrlam:
Right now, external folks can't merge their own branches. So merging this in, changes were made!

PR #66, opened 2022-03-15
Dropdown component (feature branch)

MelSumner:
## :pushpin: Summary

If merged, this PR adds the dropdown component.

***

## :white_check_mark: Reviewer's checklist

- [ ] +1 Percy if applicable
- [x] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201758147765505
  - https://app.asana.com/0/0/1201880481980935

PR #202, opened 2022-04-11
"Dropdown" component - CSS cleanup (02)

didoo:
### :pushpin: Summary

Some minor cleanup/refactoring of the CSS for the `Dropdown` component.

- **re-added back Sass variables**
  - removed in 0a974bc3 / see https://github.com/hashicorp/design-system/pull/66#discussion_r846285980)
- **check that all sizes are in `px` and not in `rem`**
  - some leftovers of 06e556a + b02b0e8/ see https://github.com/hashicorp/design-system/pull/66#pullrequestreview-936446035)
- **added `width: max-content` to the list so that it doesn’t compress the text**
  - see https://github.com/hashicorp/design-system/pull/66#discussion_r846317623)
  - _notice: the reason is that the element has position absolute so no intrinsic width, so this impacts on the flex algorithm of the children (in some cases they don't use the full width, probably the layout algorithm has constrasting requirements)_
- **removed extra import (in the dummy app) of the typography CSS helper classes**
  - it was not only duplicating the import of CSS classes, but also coming after the components CSS, so resetting margin/padding applied to text elements
- **replaced typographic tokens with CSS helper classes in “list-item” text elements**
  - see https://github.com/hashicorp/design-system/pull/66#discussion_r846159125)
- **un-indented CSS classes where is not needed**
  - see https://github.com/hashicorp/design-system/pull/66#discussion_r846168092)
- **removed inline size for the image in the `toggle-icon`** 
  - this is to avoid having two distinct places where this size is declared (in CSS **and** in HTML)
- **re-ordered properties alphabetically**
- **some other small refactorings/cleanups**

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ ] +1 Percy if applicable

---
  - https://app.asana.com/0/0/1202108023378525

### DIFF
--- a/packages/components/addon/components/hds/dropdown/toggle-icon.hbs
+++ b/packages/components/addon/components/hds/dropdown/toggle-icon.hbs
@@ -1,7 +1,7 @@
 <button class={{this.classNames}} aria-label={{this.text}} ...attributes {{on "click" this.onClick}} type="button">
   <div class="hds-dropdown-toggle-icon__wrapper">
     {{#if @imageSrc}}
-      <img src={{@imageSrc}} alt="" role="presentation" height="32" width="32" />
+      <img src={{@imageSrc}} alt="" role="presentation" />
     {{else}}
       <FlightIcon @name={{this.icon}} @size="24" />
     {{/if}}
--- a/packages/components/addon/components/hds/dropdown/toggle-icon.hbs
+++ b/packages/components/addon/components/hds/dropdown/toggle-icon.hbs
@@ -1,7 +1,7 @@
 <button class={{this.classNames}} aria-label={{this.text}} ...attributes {{on "click" this.onClick}} type="button">
   <div class="hds-dropdown-toggle-icon__wrapper">
     {{#if @imageSrc}}
-      <img src={{@imageSrc}} alt="" role="presentation" />
+      <img src={{@imageSrc}} alt="" role="presentation" height="32" width="32" />
     {{else}}
       <FlightIcon @name={{this.icon}} @size="24" />
     {{/if}}
--- a/packages/components/app/styles/components/dropdown.scss
+++ b/packages/components/app/styles/components/dropdown.scss
@@ -21,6 +21,9 @@
 
 @use '../mixins/focus-ring' as *;
 
+$hds-dropdown-toggle-base-height: 36px;
+$hds-dropdown-toggle-border-radius: 5px;
+
 
 // TOGGLE/ICON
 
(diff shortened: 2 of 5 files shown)


packages/components/addon/components/hds/dropdown/toggle-icon.hbs
```diff
@@ -1,7 +1,7 @@
 <button class={{this.classNames}} aria-label={{this.text}} ...attributes {{on "click" this.onClick}} type="button">
   <div class="hds-dropdown-toggle-icon__wrapper">
     {{#if @imageSrc}}
-      <img src={{@imageSrc}} alt="" role="presentation" height="32" width="32" />
+      <img src={{@imageSrc}} alt="" role="presentation" />
```
MelSumner, line 4: [issue] height and width should be pre-defined in the html markup for performance reasons. This tells the browser how much space to "save" for the image and makes the page faster. So even if we are declaring it in the CSS, it should still be declared here in the markup. 

```suggestion
      <img src={{@imageSrc}} alt="" role="presentation" height="32" width="32" />
```

packages/components/app/styles/components/dropdown.scss
```diff
@@ -113,33 +118,32 @@
   box-shadow: var(--token-surface-high-box-shadow);
   list-style: none;
   margin: 0;
-  max-width: 25rem;
-  min-width: 12.5rem;
+  max-width: 400px;
```
MelSumner, line 121: [issue] [a11y] these need to be in relative units to support browser zoom. 

The most important reason for using responsive and unitless values in our CSS is for supporting our users that rely on zooming. If you read the [Web Content Accessibility Guidelines](https://www.w3.org/TR/WCAG21/), our users need to be able to zoom the viewport without loss of content or functionality, or restrictions imposed by CSS values or viewport scaling settings. 

In particular, there are a some success criteria to be met:

- [WCAG 1.4.4](https://www.w3.org/TR/WCAG21/#resize-text): Users must be able to resize text without assistive technology up to 200 percent, without loss of content or functionality. (Level AA)
- [WCAG 1.4.10](https://www.w3.org/TR/WCAG21/#reflow): Users must be able to resize text without being forced to scroll both horizontally and vertically to read that content. (Level AA)

MelSumner (changes requested):
two small issues, everything else seems correct, although I'm not sure why we're using helper classes? I thought the helper classes were for our users, not us.

packages/components/addon/components/hds/dropdown/toggle-icon.hbs
didoo: OK.

packages/components/app/styles/components/dropdown.scss
didoo: @MelSumner can you tell me how declaring the size in `px` or `rem` would change the outcome in this case (zooming, not changing base font size, which is something we're not able to support yet)? 

**This is the content at 200% with `px`**

px1
px2

**This is the content at 200% with `rem`**

rem1
rem2

They look exactly the same to me.

PR #205, opened 2022-04-11
"Dropdown" component - Split "list-item" in independent subcomponents (05)

didoo:
### :pushpin: Summary

As discussed with @MelSumner during the implementation of the `Dropdown` in #66 we may want to consider splitting the existing code implementation of the `list-item` in independent sub-components.

Reasons for this change:
- use the same approach as in the `toggle` of the `Dropdown`
  - we may want to consider to move the `toggle` files in a sub-folder too, for consistency with this PR
- avoid overlapping of props that apply to different sub-components
  - this means a simplified API for the sub-components
  - this means also the possibility to have clearer documentation, without having to describe too many exceptions

In this PR I have:
- added the `list-item` sub-elements as independent components
- updated the `index` of the `Dropdown` to use the newly added sub-components
- deleted files for the old `list-item` implementation
- updated the `list-item` declarations/invocations in documentation page
- updated integration tests accordingly, splittin the integration tests for the `list-item` in dedicated tests for the sub-components
  - in doing so, I have also added the missing `...attributes` spread on the single items

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

---
  - https://app.asana.com/0/0/1202108023378540


Dhaulagiri:
I had this exact same thought when reviewing the initial work so I'm excited to see this followup 👏

MelSumner (approved):
still not sure how I feel about this but I won't block on it. 👍

PR #210, opened 2022-04-11
Fix the order of declarations in the Ember components backing classes

### LINKED ISSUE #197: Re-arrange order of declarations in the components
As discussed with @melsumner we want to have a consistent order for the declarations inside the backing class of the components:

tracked properties > everything else > actions, in that order. we can say that in the "Everything else" section, we want `classNames` to come last, but it's still before `actions`.

As a follow-up of this:
- @didoo will document this decision somewhere, so it's easy to refer to in the future
- @Dhaulagiri will investigate if this can be automated via a plugin (https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/order-in-components.md)

_Originally posted by @MelSumner in https://github.com/hashicorp/design-system/pull/66#discussion_r846207904_

didoo:
### :pushpin: Summary

This resolves #197

In this PR I have:
- moved the `@action` declarations to the bottom of the backing classes for the `link(to)/cta` files

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/link-to/cta.js
+++ b/packages/components/addon/components/hds/link-to/cta.js
@@ -127,19 +127,6 @@ export default class HdsLinkToCtaComponent extends Component {
     return route;
   }
 
-  @action
-  didInsert(el) {
-    // we need to register the element to compare it with the one that triggered the "key/space" event
-    this.el = el;
-  }
-
-  @action
-  onKeySpace(event) {
-    if (event.target === this.el) {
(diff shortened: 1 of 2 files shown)


Dhaulagiri (approved):
Failing tests are related to that issue we saw with an upstream dep so no need to re-run it or otherwise block the PR from being merged

PR #212, opened 2022-04-12
"Dropdown" component - Add handling of dynamic "width" for the list (07)

didoo:
### :pushpin: Summary

There is a requirement from the design side to have a way for the designers (and hence the developers) to declare a specific width for the `Dropdown` list.

This means we have to expose a `@width` prop that can be set by the developers at implementation time. But since it's dynamic, it means it needs to be applied via JS at runtime.

In this PR we propose a solution with two possible variants (to be decided/discussed which one suits best our needs and the Ember ecosystem):
- we set an CSS inline style via JS (with safe HTML) and apply it to the list
- we set a CSS custom prop via JS and use it in CSS to control the `width` of the list

In this PR I have:
- updated the `index` in the `Dropdown` to have backing class that exposes a `@width` prop and handles the width accordingly
  - notice: I’m using two different methods, to discuss what is the most suitable for us
- updated one of the examples in the documentation to use as test

### 🤔 Things to discuss

- [ ] we need to understand if it's a `number` (so implicitly always in `px`) or a `string` that can contain any value/unit (in which case we can't control/validate what is used as value)

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

---
  - https://app.asana.com/0/0/1202108023378550


packages/components/addon/components/hds/dropdown/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+    return listPosition;
+  }
+
+  /**
+   * Get a style attribute to apply to the "list" based on the @width argument
+   * @return {string} The style attribute to apply to the list
+   */
+  get listStyle() {
+    return this.args.width ? htmlSafe(`width: ${this.args.width}`) : undefined;
```
jgwhite, line 34: Counter-intuitively, Ember’s [`htmlSafe`](https://api.emberjs.com/ember/4.2/functions/@ember%2Ftemplate/htmlSafe) function does not perform any sanitization, but rather tells Ember “this string is already safe, no need to sanitize it”.

packages/components/addon/components/hds/dropdown/index.js
```diff
@@ -0,0 +1,60 @@
...
+    if (this.args.width) {
+      classes.push('hds-dropdown-list--fixed-width');
+    }
+
+    return classes.join(' ');
+  }
+
+  @action
+  didInsert(el) {
+    el.style.setProperty('--width', htmlSafe(this.args.width));
```
jgwhite, line 58: With this approach we’d need a corresponding `did-update` to update the CSS variable if/when new values for `@width` arrive.

jgwhite:
Thanks for opening this up for discussion. Two thoughts in addition to my inline comments:

First, have you considered using [ember-style-modifier](https://github.com/jelhan/ember-style-modifier)?

Second, regarding the following:

> we need to understand if it's a number (so implicitly always in px) or a string that can contain any value/unit (in which case we can't control/validate what is used as value)

Personally, I think it’s totally fine if the component only accepts CSS-compatible values for `@width`. In other words, while I appreciate the convenience of `@width={{12}}` I think `@width="12px"` is actually clearer.

didoo:
For context: here's the error triggered in Cloud UI when trying to use the code above:
screenshot_1300
Interesting enough, the CSS custom prop is applied and it works.

didoo:
> First, have you considered using [ember-style-modifier](https://github.com/jelhan/ember-style-modifier)?

@jgwhite didn't know of its existence, neat alternative. (I see is imported in Cloud UI but I can't see it's used anywhere?)

> Personally, I think it’s totally fine if the component only accepts CSS-compatible values for `@width`. In other words, while I appreciate the convenience of `@width={{12}}` I think `@width="12px"` is actually clearer.

I agree too (but I want to hear other opinions/points of view before making a decision)

packages/components/addon/components/hds/dropdown/index.js
ChaiWithJai: I think this property doesn't adjust with a changing with. If we want to react to the `width` property changing then I think we should create an element modifier which will autoupdate when our `width` property changes.

johncowen:
Wondered if you considered using:

```
<Hds::Dropdown style="--panel-width: 248px">
```

Custom props cascade so you can pick that value up anywhere you like in the CSS. Document the CSS prop for bonus points. Also pretty easy to add a `--panel-height` then if/when you need to make the panel scrollable.

In fact I don't know the code that well, but wondered if putting a `width 100%` on the `ul` and then controlling width via the root node would work? Then you wouldn't need the custom prop, just use `width`, guessing that as its an inline style it would overwrite whatever the 'default' was you specify as the component author.

Are there min-width/max-width/fit-content concerns here at all from a design perspective?

packages/components/addon/components/hds/dropdown/index.js
randallmorey: I agree with @jgwhite.  `htmlSafe` should be avoided, especially with input data.  I'm curious why this is even necessary though?  Would any part of the concatenated string be sanitized by Ember?

packages/components/addon/components/hds/dropdown/index.js
randallmorey: This approach (used with `did-update`) might avoid linting issues.  `htmlSafe` is probably unnecessary here, for a couple of reasons:
- Since the value isn't passing through a template, marking it as safe would have no effect.  Ember doesn't play a roll in this DOM call.
- The values are probably not subject to sanitization anyway (see my comment on the inline style approach).

packages/components/addon/components/hds/dropdown/index.js
didoo: > I was mostly wondering what kind of input data you were expecting that would even be subject to sanitization.
No user-generated data, just consumers (you) values.

packages/components/addon/components/hds/dropdown/index.hbs
```diff
@@ -8,10 +8,7 @@
     }}
   </:toggle>
   <:content>
-    <ul
-      class="hds-dropdown-list
-        {{if (eq @listPosition 'left') 'hds-dropdown-list--position-left' 'hds-dropdown-list--position-right'}}"
-    >
+    <ul class={{this.listClassNames}} style={{this.listStyle}} {{did-insert this.didInsert}}>
```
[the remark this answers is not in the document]
didoo: > Will inline styling cause linting problems down the line?

That's what I was trying to understand :)

> or clash with also setting the class attribute here?

Not that I know. They will follow the normal CSS specificity rules.

packages/components/addon/components/hds/dropdown/index.js
didoo: Removed this approach in favour of using `ember-style-modifier`.

packages/components/addon/components/hds/dropdown/index.js
didoo: OK, adopted `ember-style-modifier` instead of doing inline styling ourselves.

didoo:
@jgwhite @johncowen @meirish @ChaiWithJai @randallmorey thanks for all the precious feedbacks and suggestions: at the end I went with `ember-style-modifier` instead of inlining the style ourselves, so it's more in line with the current approach in your codebase(s).

didoo:
For context, in this spike/test I've found the `width` used in a few places https://github.com/hashicorp/cloud-ui/pull/2362 (in one case it was even a `max-width` which probably is an error).

didoo:
@MelSumner @Dhaulagiri can I have your review too? thanks

randallmorey:
@didoo have you considered leaving width override to the developer's stylesheet rather than exposing it as a component attribute?  I tend to agree with @MelSumner.  Anyway I think giving developers hooks to override styles where they belong (stylesheets) is the best way to go.

didoo:
@randallmorey sorry but I strongly disagree: in a component-based world, CSS classes are **not** the API of a component. "Consumers" should (ideally) **never** use CSS classnames from the design system for targeting elements in their codebases (for CSS I mean, but ideally also for other usages). A component is a black box, that speaks with the outside world through APIs expressed as `props` (arguments). CSS classes are an internal concern (just think to all the cases in which the component could be styled using things that don't have a CSS class name, or a classname that can't be used as "target" from outside: inline styles, atomic CSS, CSS-in-JS at compile time, dynamic CSS class names generated at runtime/buildtime).

PR #215, opened 2022-04-13
"Dropdown" component - Add support for "LinkToExternal" engine component (08)

didoo:
### :pushpin: Summary

This is a draft PR to discuss (using real code) a possible solution to the problem of supporting `LinkToExternal` links in our components (these are very special kind of links, similar to the `<LinkTo>` but specific for Ember engines.

More context:
- https://github.com/hashicorp/cloud-ui/pull/2362#issuecomment-1097909966
- https://hashicorp.slack.com/archives/CPB8GS9QT/p1649868202616679

👉 Notice: if you want to test it yourself and play with the code:
- clone the repo, then do a `yarn install` in the root of the project, and then call `yarn start` in the folder `packages/components`
- this will build the dev version of the dummy website and run it as localhost (you will see the port in the CLI)
- open the localhost website and navigate to the `Dropdown` page

### :hammer_and_wrench: Detailed description

In this PR I have:
- added `ember-engines` as both _devDependency_ and _peerDependency_
  -  as suggested by Matthew Irish [in this conversation in Slack](https://hashicorp.slack.com/archives/CPB8GS9QT/p1649868722882949?thread_ts=1649868202.616679&cid=CPB8GS9QT)
  - we need both [according to the documentation](https://ember-engines.com/docs/quickstart#:~:text=A%20quick%20note%2C%20Engines%20should%20have%20ember%2Dengines%20listed%20as%20a%20devDependency%20and%20peerDependency)
- updated the `list-item/interactive` to support also a `LinkToExternal` via a `@isRouteExternal` prop
- changed one of the routes in the dropdown doc page to use an external route

🚨 **Notice**: this triggers an error in console, I suspect we need to [configure the app to support engines](https://ember-engines.com/docs/quickstart)

screenshot_1322

I am not sure how to proceed from here, is beyond my Ember knowledge.

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202122650634746
  - https://app.asana.com/0/0/1202122650634755

PR #216, opened 2022-04-14
"Interactive" generic component (01)

didoo:
### :pushpin: Summary

While working on the HDS components, me and @MelSumner we've found ourselves writing over and over the same "logic" to handle the different ways to declare a `link` vs a `LinkTo` vs a `button` in Ember ([here one of the last examples](https://github.com/hashicorp/design-system/pull/160/commits/23836021ef7cedfe55b5b9f8a5b40f3fdcbe7da5)). 

This is the TLDR of the "logic" (this is pseudo-code, the actual implementation in code is slightly different)

```
{{#if @route}}
  {{#if @isRouteExternal}}
    <LinkToExternal>
  {{else}}
    <LinkTo>
  {{/if}}
{{else if @href}}
  {{#if @isHrefExternal}}
    <a target="_blank" rel="noopener noreferrer">
  {{else}}
    <a>
  {{/if}}
{{else}}
  <button/>
{{/if}}
```

The intent of this PR is to create a generic, reusable, "utility" component (almost headless) to centralize this logic in a single place, and abstract it away from the actual UI components.

#### Where it will be used?
At the moment we expect to use it in these components:
- `button`
- `dropdown/list-item/interactive`
- `link-standalone`
- `link-inline` (coming soon)

but is likely that most of all the future interactive components/elements will use it as well.

#### ...but wait, why use `@href` as argument/prop and not just `href` attribute?

We can't support the direct use of the `href` HTML attribute because we need to rely on the `@href` Ember argument to differentiate between different types of generated output (see the "logic" above).

In this PR I have:
- added a generic`Hds:Interactive` component
  - under one single component it handles the generation of `<button>` / `<a>` / `<a target="_blank" rel="noopener noreferrer">` / `<LinkTo>` / `<LinkToExternal>` conditionally, depending on the `@href` / `@route` / `@isHrefExternal` / `@isRouteExternal` arguments passed 
  - no styling applied, is just a "utility" component that abstract the logic above in a single place (we are using it over and over in our components) - the styling will be applied via CSS classes that will be applied to the instance (and its children) in the context where it's used
- updated the `dropdown/list-item/interactive` to use the new `Hds::Interactive` generic component
- added the documentation page ([preview here](https://hds-components-git-spike-interactive-generic-c-6d4c05-hashicorp.vercel.app/utilities/interactive))
- added the integration tests

### 🤔 Things to decide/discuss

- [ ] how do we call this component? `interactive`? `action`? something else?
- [ ] what do we do with the `LinkToExternal`?
   - see #215 and the Slack threads linked in that PR

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/addon/components/hds/interactive/index.hbs
```diff
@@ -0,0 +1,34 @@
+{{#if @route}}
+  {{#if @isRouteExternal}}
+    <LinkToExternal
```
Dhaulagiri, line 3: [question] there was some discussion over whether or not we'd need to add `ember-engines` as a dep to support this, but is my interpretation here correct that this code path will only happen if folks opt into it by passing `@isRouteExternal` meaning only apps that use engines will ever go down this path hence there isn't a big problem to solve?

didoo: Yes. In all my tests, this works. But would be great to have feedback from the Cloud UI/TCE engineers (and more experienced Ember devs) on this, if it's an acceptable way to support external links without having to introduce engines in the design system ember setup/configuration.

Dhaulagiri: That seems logical to me and an "easy" way to solve what I thought might be pretty hard to figure out, but would be good to confirm like you said.

noelledaley:
Just dropping in my note from the RFC ~

nice, i do think it'd be quite useful to combine the link components here (and including the LinkToExternal in that as well). The main thing i'd want to make sure is enforced is that consumers can intuitively pass in a `@route` or `@href` and receive a warning if they try to pass in a @route that should be an @href (i.e. passing in @route='https://foo.com' should not be allowed). i don't necessarily know if anyone would do this, but i think our design system components should make it easy to do the correct thing and not allow developers to mix up params unintentionally.

didoo:
@noelledaley a question about your proposal, in particular re. the `@route/@href` validation:
- I don't see it done in Structure or in Cloud UI so I wonder why it would be necessary: have you or others met this problem in real code?
- I would imagine this would mean do some kind of strange validation via regex:
  - validate the `@href` against any possible valid URL (not so simple)
    - what if for some reasons actually in the product there are special cases in which the `@href` parameter is not a standard one?
    - validate the `@route` parameter against... ? not sure how we could properly validate this: what is a valid _route_ in Ember? afaik is a string, but what regex or validation would you have in mind (and the one you're suggesting is a non-validation: everything that is not a URL then is valid?)
- when I try to use a non valid route in the the `LinkTo` component, Ember throws an error: would it not be enough in this case?

noelledaley:
@didoo nope, we haven't encountered that problem, it was just something that i could see someone accidentally doing. as long as the Ember error coming from and invalid route isn't swallowed i think that should be plenty. thanks for considering the idea!

didoo:
@MelSumner as discussed yesterday, I have added the handling of the `@isHrefExternal` (defaults to `true`) for the `<a>` links. You can see the commits here https://github.com/hashicorp/design-system/pull/216/commits/e12dd4474cd4869998f6e3b5c8bf881a1ae8b8fa and here https://github.com/hashicorp/design-system/pull/216/commits/3544a8ddf8c17a41d2eeaf770b85222b5a810d7e.

packages/components/addon/components/hds/interactive/index.hbs
```diff
@@ -0,0 +1,41 @@
...
+    <a target="_blank" rel="noopener noreferrer" ...attributes href={{@href}}>
+      {{yield}}
+    </a>
+  {{else}}
+    <a ...attributes href={{@href}}>
+      {{yield}}
+    </a>
+  {{/if}}
+{{else}}
+  <button type="button" ...attributes>
```
alex-ju, line 38: I like that we set this by default to `type="button"` to prevent accidental form submissions

alex-ju (commented):
As far as my understanding of `LinkTo` and `LinkToExternal` goes at the moment, this makes sense to me. No suggestions for improvements from my side, I'm afraid.

MelSumner (approved):
Tentative about this one but let's explore how it works!

PR #220, opened 2022-04-15
Extend the "Button" component to support also links as actions (02)

didoo:
### :pushpin: Summary

This is an attempt to solve once for all the issue of "links that look like buttons" in our products.

The previous attempt using a `Link::CTA` component (see https://github.com/hashicorp/design-system/pull/115) was followed by other issues raised by the developers, this time around the need of a `secondary` styling for the link (see [this Slack thread](https://hashicorp.slack.com/archives/C11JCBJTW/p1649412165572899), but also [this Slack thread](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1649267673705649), [this other Slack thread](https://hashicorp.slack.com/archives/C03A0N1QK8S/p1649362956033069), [this Slack thread](https://hashicorp.slack.com/archives/C03A0N1QK8S/p1649850252570679) too, this [comment](https://hashicorp.slack.com/archives/C11JCBJTW/p1649423066529139?thread_ts=1649412165.572899&cid=C11JCBJTW), this [Jira ticket](https://hashicorp.atlassian.net/browse/HVT-1363), [this GitHub PR](https://github.com/hashicorp/cloud-ui/pull/2316)).

Adding a **secondary** style for the `Link::CTA`, which by its own definition is _intended to be a design element that [...] stands out from the rest of the page[and] tells the user “this is the most important action on this page”_ doesn't make sense at this point. The only pragmatic solution is to modify the `Button` component and allow it to handle links too.

This change will facilitate the migration of the Cloud UI codebase to HDS buttons, unblocking some of the tasks that were in standby, waiting for alternatives to the existing UI implementations based on Structure components.

In this PR I have:

- updated the `button` to use the new `interactive` generic component
  - allowed underline of text on links
  - add the `space` key event handler when the element is a `link` (see #291)
  - removed the `@type` argument from the API in favour of the `type` native attribute (this simplified the code)
- improved assertion guards in `button` to work differently between buttons and links
- added special visual treatment to alert the developer if a `href` HTML attribute is used instead of the `@href` Ember argument
- updated documentation page - [preview page](https://hds-components-git-spike-button-with-link-support-hashicorp.vercel.app/components/button)
  - API
  - showcase
- updated integration tests

### 🤔 Things to discuss/decide
- [ ] decide if we want to keep the visual treatment (alert) when the consumer passes a `href` instead of a `@href`
    - screenshot_1338

### :link: Links

- **RFC**: [DS-033 - Button/Link/CTA Rethink](https://docs.google.com/document/d/1Tc99UR3UJo_avncWD1V8uGgDr72W2TjLbxAvPPaNRj8/edit)

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/app/styles/components/button.scss
```diff
@@ -26,6 +26,11 @@ $hds-button-focus-border-width: 3px;
   position: relative;
   width: auto;
 
+  @at-root a#{&} {
+    text-decoration: none;
+    width: fit-content;
+  }
```
alex-ju, line 32: We have [`button.hds-button` defined further down the page](https://github.com/hashicorp/design-system/pull/220/files#diff-bc76736c4fa05350de917ad2426c66e972a17bf42a5d210fe476f48ba7d800afR319), would it be more consistent (and potentially easier to read) if we define these simply as `a.hds-button`?

alex-ju (approved):
If I understand correctly, the space key event handler would be added in a separate PR

packages/components/app/styles/components/button.scss
didoo: I see your point here, but I have mixed feelings for this suggestion. Reason is that the one below is a special override that we apply to the component (intended only for the developers, it's a large visual `console.error`, and may also be removed if we this is too much, or risky) while this one is actual styling. I have added a comment above this rule, so it's more obvious what it does.

didoo:
> If I understand correctly, the space key event handler would be added in a separate PR

Yes: all the accessibility improvements will be done once we agree on what they should be (eg. the `role` was controversial).

packages/components/addon/components/hds/button/index.js
```diff
@@ -126,25 +124,6 @@ export default class HdsButtonIndexComponent extends Component {
     }
   }
 
-  /**
```
didoo, line 129: removed the @type argument from the API in favour of the type native attribute (this simplifies the code)

packages/components/app/styles/components/button.scss
```diff
@@ -304,4 +309,24 @@ $size-props: (
       border-color: transparent;
     }
   }
+}
+
+
+// SPECIAL
+
+// we apply a visual treatment to alert the developer if a "href" HTML attribute is used (instead of the "@href" Ember argument)
```
didoo, line 317: please let me know what you think of this

MelSumner (approved):
Reiterating that we definitely need to iterate on this, it's the worst possible outcome of all of the possible outcomes we discussed. Understand why from a business perspective, but need to have this in writing.

packages/components/app/styles/components/button.scss
alex-ju: seems like a good idea; if we end-up needing this in other places there may be value in converting it into a mixin

alex-ju (approved):
The `Interactive` abstraction makes sense to me; code looks good.

PR #223, opened 2022-04-15
update to yarn 3

Dhaulagiri:
### :pushpin: Summary

This PR moves us from yarn v1 to yarn v3

The main impetus for this that yarn v1 is in maintenance mode and support for workspaces in it is pretty rough. More context is in #341, but in short none of the small bugs we are experiencing in v1 will ever be fixed so this is the best way to get onto a more modern version of yarn.

The one drawback is that this change exposed an underlying issue with our ember-3.24 scenario for the components package. This is a quirk of the monorepo and ember that actually exists locally regardless, this change just exposes it in CI. We've informally agreed that this is not a long term concern though given we expect most teams to be on more modern versions of ember.

### 👀 How to review

👉 Review commit-by-commit


packages/components/config/ember-try.js
```diff
@@ -5,6 +5,10 @@ const { embroiderSafe, embroiderOptimized } = require('@embroider/test-setup');
 
 module.exports = async function () {
   return {
+    useYarn: true,
+    buildManagerOptions() {
```
Dhaulagiri, line 9: add a comment on why this was needed

Dhaulagiri:
Opening this up for review. I'm honestly not 100% convinced this is the best idea long term as my perception is that, in general, yarn as a project is not in the best place (yarn 1 is obsolete but widely used, yarn 3 is more actively worked on but not used much from what I can see, etc). Even if we accept this PR it may be worth exploring how to finalize a migration to use npm which seems somewhat more common 🤷

alex-ju (approved):
I guess it's a bet on increased adoption of `yarn 3` across the web that I'm personally willing to make. Extra 🍬 for updating the docs!

MelSumner (approved):
Thank you for doing this work! 👍

PR #225, opened 2022-04-15
use ember-cli-deprecation-workflow in packages/components

Dhaulagiri:
### :pushpin: Summary

Adds [ember-cli-deprecation-workflow](https://github.com/mixonic/ember-cli-deprecation-workflow) to manage our deprecation warnings

The main benefit of this tool is that it gives more control for us in managing these deprecations. Often times we can't actually fix them so by suppressing them in the console we reduce noise in our consoles during development and testing. Once we are ready to address them we can set the `handler` to `throw` and ensure we don't regress on the deprecation in the future.

As part of this PR I've upgraded `ember-keyboard` to remove deprecated code upstream and set those to `throw` so we won't regress on them in the future.

### 👀 How to review

👉 Review by files changed

### DIFF
--- a/.changeset/mean-sheep-rest.md
+++ b/.changeset/mean-sheep-rest.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+update to ember-keyboard@8.1.0 in packages/components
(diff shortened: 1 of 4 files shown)


MelSumner:
Note: this would be good to add to a regular maintenance check list.

didoo (approved):
Not sure how to review this, but if you need a thumbs up for merging this, here it is :) 

👍

Dhaulagiri:
@didoo I'll give you a demo of this tool when we have a minute so you can see what it does and let me know if it feels useful or not

PR #227, opened 2022-04-19
Unify "Link/LinkTo::Standalone" component (04)

didoo:
### :pushpin: Summary

Now that we have a single `interactive` element that handles both the normal `<a>` links and the `<LinkTo>` links, there's no reason why we should keep separate the "standalone" link in two distinct components, we can use the same method (`@href` vs `@route`) to detect which one is needed, using the underlying `Hds::Interactive` component that does all the work for us.

This will also simplify the adoption of the component in the consumers codebases, and its documentation.

_Notice: even if under the hood it's using the `Hds::Interactive` component, this `Hds::Link::Standalone` can’t be used as a button, because there’s an assertion thrown if nor `@route` nor `@href` are passed as arguments._

In this PR I have:
- removed the `LinkTo::Standalone` component from the codebase
  - component
  - documentation
  - tests (percy + integration)
- updated the `Link::Standalone` to use the new `interactive` generic component
  - in this way it serves both the old `Link::Standalone` and `LinkTo::Standalone` use cases
- unified the documentation page

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


didoo:
@alex-ju @MelSumner @Dhaulagiri can you review also this one? thanks

alex-ju:
Assuming the 'Interactive' component makes its way in, I agree these two shall be merged. Couldn't spot anything off in the code.

MelSumner (approved):
I think we should consider not removing components until we're sure of the unified interactive component approach. WDYT?

didoo:
> I think we should consider not removing components until we're sure of the unified interactive component approach. WDYT?

OK for me. In any case, this PR is branched from the `interactive` branch, and will be merged back into it, so they go hand in hand, so there's no risk of releasing one without the other. So I am merging in the main `feature` branch, and it will stay there until we are OK with the global changes.

PR #231, opened 2022-04-19
Add "Link::Inline" component (05)

didoo:
### :pushpin: Summary

This PR intends to add the missing counter-part of the `Link::Standalone` (_a link that's used in isolation, not inline with adjacent text_) component: the `Link::Inline` (_a link that's used within a body of text_).

#### Why not use just a `<a>` element?

Because in this way:
- we can use the "logic" provided by the `Hds::Interactive` component to handle the different kinds of Ember links we need to support (`<a>`, `<LinkTo>`, `<LinkToExternal>`)
- we can control the colors (by design, only `primary` ("action" color) and `secondary` (dark gray) should be used)
- we can control in detail the visual treatment of the interaction states (`hover/active/focus`)

In this PR I have:
- added `Link::Inline` component
   - to resolve some issues with whitespace (see below) I had to slightly modify the template of the `Interactive` component
- added documentation page ([preview here](https://hds-components-git-spike-link-inline-hashicorp.vercel.app/components/link/inline))
- added tests (integration + percy)

#### 🚨 Known issues

- **Whitespace** - see comment below: https://github.com/hashicorp/design-system/pull/231#issuecomment-1123502499
  - found a fix for the "internal" whitespace, but gave up on trying to solve the "external" whitespace (asked @Dhaulagiri and @MelSumner to investigate more, if they have capacity).
- **Focus ring** - see comment below: https://github.com/hashicorp/design-system/pull/231#issuecomment-1123594656
  - decided to go with the simple override of the whole `outline`: this will make it consistent across all the browsers, apart from the border radius in Safari (something that probably can't be fixed anyway) 

### 🔗 Links

- [Proposal about color naming convention in Figma for Link Inline](https://docs.google.com/document/d/1o2FotMfRdKz508RbW_jg2cu8p_3hI1a-4Trv-I1kD_Q/edit#)

### TODOs
- [x] Decide if it's OK to support also the `icon` in the component, or we want to remove it - see https://github.com/hashicorp/design-system/pull/255
    - let's go with it at this point
- [x] add missing documentation for designers
    - decided we'll add in a follow-up PR

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


didoo:
PLACEHOLDER COMMENT - DON'T DELETE

didoo:
### Issues with focus ring

I'll collect in this comment all the findings around the problems we found with the focus ring in the implementation of this component.

For the "focus ring" we're using the native focus and applying custom styling to it, and we have noticed that the browser implementation of the `focus` on links is pretty inconsistent, depending on how we apply the styles (browsers behave differently if you override just the `outline-***` properties or the whole `outline`, see tables below).

#### Override just the `outline-***` properties

If the overrides are applied this way:
```
.hds-link-inline {
  border-radius: 10px;
  outline-color: #FF0000;
  outline-width: 5px;
  outline-offset: 10px;
}
```

| Browser | outline-color | outline-width | outline-offset | border-radius (outline) |
|---|---|---|---|---|
| Chrome | ✅ | ❌ | ✅ | ✅ | 
| Safari | ❌ | ❌ | ✅ | ✅ |
| Firefox | ❌ | ❌ | ✅ | ✅ |
| Edge | ✅ | ❌ | ✅ | ✅ |

#### Override the whole `outline` property

If the overrides are applied this way:
```
.hds-link-inline {
  border-radius: 10px;

  &:focus,
  &:focus-visible {  
    outline: 10px solid #FF0000;
    outline-offset: 5px;
  }
}
```

| Browser | outline-color | outline-width | outline-offset | border-radius (outline) |
|---|---|---|---|---|
| Chrome | ✅ | ✅ | ✅ | ✅ | 
| Safari | ✅ | ✅ | ✅ | ❌ |
| Firefox | ✅ | ✅ | ✅ | ✅ |
| Edge | ✅ | ✅ | ✅ | ✅ |

cveigt:
@didoo I've collected all the feedback and made the necessary adjustments.
I will apply the changes to `Link primary` and add `Link secondary` after the merge.

Here's the link to the [Figma branch](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/zJFztjPwmavtrdJ3hTtzXY/HDS-Product---Components?node-id=8130%3A20251). They are all using existing styles/tokens and are already connected.

Screen Shot 2022-05-11 at 5 15 12 PM

Let me know if anything is missing or needs to be discussed.
CC @heatherlarsen

cveigt:
I forgot that this won't be a component in Figma, so we'll need to add color styles/design tokens to Foreground for product designers.

For actions, we currently have 
`--token-color-foreground-action`
`--token-color-foreground-action-hover`
`--token-color-foreground-action-active`

what do you think about adding these?
`--token-color-foreground-action-secondary` (neutral-700)
`--token-color-foreground-action-secondary-hover` (neutral-600)
`--token-color-foreground-action-secondary-active` (neutral-500)

packages/components/addon/components/hds/link/inline.hbs
```diff
@@ -0,0 +1,23 @@
+{{! IMPORTANT: we removed the newlines before/after the yield to reduce the issues with unexpected whitespaces (see https://github.com/hashicorp/design-system/pull/231#issuecomment-1123502499) }}
+{{! IMPORTANT: we need to add "squishies" here (~) because otherwise the whitespace added by Ember becomes visible in the link (being an inline element) - See https://handlebarsjs.com/guide/expressions.html#whitespace-control }}
```
alex-ju, line 2: I personally find whitespace control operators rather common in templates that handle non-block elements and whitespace-sensitive contexts; that being said it may be valuable for some people to have these comments in.

packages/components/app/styles/components/link/inline.scss
```diff
@@ -0,0 +1,66 @@
...
+
+.hds-link-inline {
+  border-radius: 2px; // this is used by the outline too
+
+  &:focus,
+  &.is-focus,
+  &:focus-visible {
+    outline: 2px solid var(--token-color-focus-action-internal);
+    outline-offset: 1px;
+    // text-decoration: none; // TBD with CVeight
```
alex-ju, line 18: Leaving a comment only to make sure we don't forget about this line

alex-ju (approved):
Technically this looks good – happy for it to be merged in the feature branch when the design comments are resolved.

packages/components/app/styles/components/link/inline.scss
didoo: fixed.

packages/components/addon/components/hds/link/inline.hbs
didoo: I didn't know of their existence until a couple of weeks ago :)

cveigt (approved):
LGTM 👍 -I like how the focus ring ended up looking. It looks better in the browser than in Figma with the same offset.

PR #240, opened 2022-04-22
Updates code block theme

### LINKED ISSUE #239: Change code block theme
The current code block theme does not provide sufficient color contrast.

MelSumner:
### :pushpin: Summary

If merged, this PR resolves #239 by updating the code block theme in the dummy website so that it passes color contrast criteria.  

Before:
image

After: 
image

### DIFF
--- a/packages/components/ember-cli-build.js
+++ b/packages/components/ember-cli-build.js
@@ -12,10 +12,13 @@ module.exports = function (defaults) {
       ],
     },
     'ember-prism': {
-      components: ['markup'],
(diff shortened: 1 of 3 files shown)


didoo (approved):
OK for me. 

One question though [not a blocker]: why not use the same theme as in Flight?
screenshot_1364
In that case, the copy button is even more visible compared to the one in this PR:
screenshot_1366

MelSumner:
@didoo we could. I just like this one better and didn't know about it at the time when we made Flight.

PR #235, opened 2022-04-21
"Dropdown" component - Finalize the documentation (10)

didoo:
### :pushpin: Summary

TODO

### :hammer_and_wrench: Detailed description

TODO

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #243, opened 2022-04-22
"Dropdown" component - Remove default icon from "toggle/icon" (12)

didoo:
### :pushpin: Summary

While working on the documentation of the dropdown, I noticed that the default icon for the Toggle::Icon is the user one. afaik there’s only one use case in the entire application where this icon in the dropdown is used, which is the user menu, so why have we promoted it to default? for me there should not be a default for this parameter, doesn’t make sense (in which way is helping me, consumer of the component?)

We had a discussion with @MelSumner @heatherlarsen @cveigt, there was no unanimous consensus on this but I decided to  move on with this change for the reasons above.

In this PR I have:
- removed “icon” logic from backing class of `toggle/icon`
  - also added assertion to make sure one of `@icon` or `@imageSrc` is passed to the component 
- updated the documentation 
  - to make sure it compiles, the full update will be in #235
- updated integration tests

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

---
  - https://app.asana.com/0/0/1202166650438500

### DIFF
--- a/packages/components/addon/components/hds/dropdown/toggle/icon.hbs
+++ b/packages/components/addon/components/hds/dropdown/toggle/icon.hbs
@@ -2,8 +2,8 @@
   <div class="hds-dropdown-toggle-icon__wrapper">
     {{#if @imageSrc}}
       <img src={{@imageSrc}} alt="" role="presentation" height="32" width="32" />
-    {{else}}
-      <FlightIcon @name={{this.icon}} @size="24" />
+    {{else if @icon}}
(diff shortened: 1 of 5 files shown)


MelSumner (approved):
I am not in favor of removing defaults but I do not think it's pragmatic to spend any more time on this. In the future, we should endeavor to be more pragmatic about what we are choosing to spend our energy on so we can move the project forward, since there are endless things we could bikeshed about.

PR #244, opened 2022-04-25
Adds an about page

MelSumner:
### :pushpin: Summary

If merged, this PR adds an "about this design system" page.

We are discovering knowledge gaps where we think that providing additional context for the design system will be valuable for our users (and valuable to us to refer users to as we provide support). 

While some of this information is available in RFCs, it is always useful to have a condensed version in the same place as the components themselves for additional context. We believe that users will explore "about this design system" if they are new to the design system in general, and this will help them understand the context better than if they had to intuit it alone. In this way, we are supporting designers and developers at all stages of their career and reducing the risk of loss of time due to assumptions. 

For reviewers: I've added the information I think would be the most useful, but if other information would be useful here, please comment and/or PR to add. 

Screenshot:
CleanShot 2022-04-25 at 10 07 30

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/tests/dummy/app/templates/about.hbs
+++ b/packages/components/tests/dummy/app/templates/about.hbs
@@ -4,26 +4,43 @@
 
 <section aria-labelledby="ds-purpose">
   <h3 class="dummy-h3" id="ds-purpose">Purpose</h3>
-  <p class="dummy-paragraph">The purpose of the design system is to provide a single source of truth for the design vision needed to build accessible, differentiated experiences for HashiCorp's products.</p>
-  <p class="dummy-paragraph">The Design System Team designs, creates and supports the design system, with input from our partners in product design and the engineers that we support.</p>
+  <p class="dummy-paragraph">The purpose of the design system is to provide a single source of truth for the design
+    vision needed to build accessible, differentiated experiences for HashiCorp's products.</p>
+  <p class="dummy-paragraph">The Design System Team designs, creates and supports the design system, with input from our
+    partners in product design and the engineers that we support.</p>
 </section>
 
 <section aria-labelledby="ds-purpose">
   <h3 class="dummy-h3" id="ds-usage">Usage</h3>
-  <p class="dummy-paragraph">While the technical usage instructions are available in the <a href="">components' repository</a>, this section intends to clarify the (other/philosophical/strategic) usage details.</p>
-  <p class="dummy-paragraph">First, it should be make explicitly clear that the design system does not intend to be a 1:1 replacement for existing styles, components, or code that may exist, not only because it seeks to unify design, provide accessibility conformance and code consistency, but also because previous implementations across products have proven to be inconsistent in all of these areas.</p>
+  <p class="dummy-paragraph">While the technical usage instructions are available in the
+    <a href="https://github.com/hashicorp/design-system/blob/main/README.md">design-system repository</a>, this section
+    intends to clarify the (other/philosophical/strategic) usage details.</p>
+  <p class="dummy-paragraph">First, it should be make explicitly clear that the design system does not intend to be a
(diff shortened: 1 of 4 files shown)


heatherlarsen:
@MelSumner I think this is a great start and I like the idea of helping our consumers better understand what we do, what our priorities are, and how to use HDS. Because the Purpose is so vital for us to all be aligned to, maybe we should spend some time as a team wordsmithing it to make sure it's reflective of all of our team's priorities. And maybe we can lean on the [Platform Team Charter](https://docs.google.com/document/d/1bX8JILYfr9pSxpt2hlLYN6pVBSp1Kftp1AUSAZ5RG6E/edit?usp=sharing) to help us as well?

packages/components/tests/dummy/app/templates/about.hbs
```diff
@@ -0,0 +1,46 @@
+{{page-title "About this Design System"}}
+
+<h2 class="dummy-h2">🚧 About This Design System </h2>
```
heatherlarsen, line 3: [nit]: Could this be "About HashiCorp Design System"?

MelSumner:
> @MelSumner I think this is a great start and I like the idea of helping our consumers better understand what we do, what our priorities are, and how to use HDS. Because the Purpose is so vital for us to all be aligned to, maybe we should spend some time as a team wordsmithing it to make sure it's reflective of all of our team's priorities. 

@heatherlarsen having our team brainstorm on the wordsmithing is a great idea. I pulled this from the Design System Team RFC (DS-009) to get us started; I'll add this to the next team agenda so we can work more on it!

packages/components/tests/dummy/app/templates/about.hbs
```diff
@@ -0,0 +1,46 @@
...
+<section aria-labelledby="ds-purpose">
+  <h3 class="dummy-h3" id="ds-usage">Usage</h3>
+  <p class="dummy-paragraph">While the technical usage instructions are available in the
+    <a href="https://github.com/hashicorp/design-system/blob/main/README.md">design-system repository</a>, this section
+    intends to clarify the (other/philosophical/strategic) usage details.</p>
+  <p class="dummy-paragraph">First, it should be make explicitly clear that the design system does not intend to be a
+    1:1 replacement for existing styles, components, or code that may exist, not only because it seeks to unify design,
+    provide accessibility conformance and code consistency, but also because previous implementations across products
+    have proven to be inconsistent in all of these areas.</p>
+</section>
```
heatherlarsen, line 22: I'd love to see this section have a bit more of a positive spin because it's sounding a little blamey (even if it's factual). Maybe it's better positioned as "goals" where we only outline the things we intend to help move forward and drop the extras about "technical usage" and "what we do not intend to be"?

For example: 

- We seek to unify design and provide consistency across products
- We bake accessibility conformance into each component _(this one needs help)_

Better yet, maybe it could repurpose some of the things from the [Platform Experiences Team Charter](https://docs.google.com/document/d/1bX8JILYfr9pSxpt2hlLYN6pVBSp1Kftp1AUSAZ5RG6E/edit?usp=sharing), specifically the sections "We Value" and "The Value We Bring"?

MelSumner: @heatherlarsen I will work on the wording here, thank you for this feedback! 

My goal here was to point out the practical parts of what we are trying to do; while I think that the PE team charter is great, I want to try to avoid phrases that are more on the intellectually ambiguous side and stay more in the explicit, "boots-on-the-ground" side. 

We can definitely point to it as a reference point but we should also keep in mind that this repo (and its documentation) is open source and I am not sure how much of internal team docs we want to reference. So we should check that too!

Finally, apologies that it came across as a blame game, that was definitely not in my mind at all.

PR #104, opened 2022-03-21
"Alert" component - actions implementation

didoo:
### :pushpin: Summary

This is an initial attempt to add support for the `actions` feature in the `Alert` component. 
I have implemented two ways to add actions to an alert:
- using the `@actions` argument (an array of objects) over which to loop and insert the corresponding HDS component
- a generic `<:actions>` named block that can be used to add any content into the "action" area

The reasons for having both is because we want to provide two ways for the users to declare the actions, a "controlled" one (that automatically sets the layout and possibly the `size` of the actions and maybe other properties by default) and an "escape hatch" in case they need to do something different/custom.

👉 **Notice**: at the moment I am not sure the `@actions` prop can do exactly what I have in mind (eg. does the `hash` work also for functions/callbacks? how do we pass extra attributes that then can be _splashed_ on the element? etc). Also, I am not sure the consumers will like this approach (it's a pattern more common/standard in React, where you can pass and spread props more easily than in Ember), they may prefer to just yield the action elements via the named block (but this in this case doesn't allow us to control spacing, layout, etc). See thread below for further discussion.

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1201994286157147

PR #56, opened 2022-03-14
Add new Alert components

amyrlam:
## :pushpin: Summary

WIP, see https://hds-components-git-amy-alerts-new-hashicorp.vercel.app/components/alert

## :link: External links

Tickets: https://app.asana.com/0/1201630204387510/1201772249529641/f in conjunction with https://app.asana.com/0/1201630204387510/1201760392262328/f

***

## :white_check_mark: Reviewer's checklist

- [ ] +1 Percy if applicable
- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #245, opened 2022-04-25
"Alert" + "Toast" components - Finalize implementation in code (feature branch)

didoo:
### :pushpin: Summary

This is the "feature branch" used to finalize the implementation of the `Alert` and `Toast` components. The implementation of these components has not followed the usual process (for [reasons]) so it's a single large PR for both the components. This PR contains the previous #56 and #91, and the follow-up changes were merged directly in this branch (for development convenience).

This PR contains these changes.

#### Alert
- added the `Alert` component (HTML/HBS + JS + CSS)
- added the documentation for the `Alert` component - [page preview here](https://hds-components-git-alert-finalize-implementation-hashicorp.vercel.app/components/alert)
- added testing (Percy + integration) for the `Alert` component

#### Toast
- added the `Toast` component (HBS + CSS) as specialized `Alert(inline)` instance
- added the documentation for the `Toast` component - [page preview here](https://hds-components-git-alert-finalize-implementation-hashicorp.vercel.app/components/toast)
- added testing (Percy + integration) for the `Toast` component

#### Alert & Toast

At the beginning there was a single component, `Alert`, that covered all the possible types of notification (`Page`, `Inline`, `Compact` and `Toast`). Later in the process, it was decided to "extract" the `Toast` in a distinct component, for many reasons (among them, it was much easier to write documentation, because these two kind of notifications have very different usages, logics, etc).

The `Toast` component in Figma is a distinct component, in a distinct page. 

Similarly in code the `Toast` component is a distinct component, with a distinct documentation page. But under the hood is just using the `Hds::Alert(Inline)` component (with just a couple of extra styles in CSS).

#### APIs

There have been a few back & forth around the API for the component, in particular around using named blocks and/or contextual components (see for example this discussion: #104).

When we tested the integration of the components in Cloud UI (see below) we run into a few issues around how Ember was treating whitespace and yielded components inside named blocks, see these threads for more details:
   - https://hashicorp.slack.com/archives/C11JCBJTW/p1651069421716229
   - https://hashicorp.slack.com/archives/C11JCBJTW/p1651583160749839
   - https://hashicorp.slack.com/archives/C11JCBJTW/p1651691482276839

You can see a writeup of what we did and what we learned in the process here: https://github.com/hashicorp/design-system/pull/104#issuecomment-1121170131

At the end we decided to get rid of the named blocks, and use only contextual components, to avoid (part of) these problems (and follow the same code pattern used in the dropdown).

#### Spike for adoption test in Cloud UI

We did a spike to test the integration of the `Hds::Alert` and `Hds::Test` components in Cloud UI, to reduce as much as possible friction in the adoption phase. You can see the results of this test here: https://github.com/hashicorp/cloud-ui/pull/2429

### :link: External links

- [CRD document](https://docs.google.com/document/d/1MC0MTm8-0F26s6Y-gou_RsMj6rtKS6jC-eM7rmruqLY/edit#heading=h.gjdgxs)
- [Figma file](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/Ru0h1oqy6r7SeGxwtQ6085/HDS-Product---Components?node-id=7636%3A35707)

***

### 👀 How to review

👉 Review by files changed (don't even try to go commit by commit :) )

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -0,0 +1,844 @@
+{{page-title "Alert component"}}
+
+<h2 class="dummy-h2">Alerts</h2>
+
+<section>
+  <h3 class="dummy-h3">Overview</h3>
```
MelSumner, line 6: Please remember to add the `id` and section link for all h3's.

packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -0,0 +1,844 @@
...
+
+  <p class="dummy-paragraph">
+    An Alert is an element that displays a brief, important message in a way that attracts the user's attention, without
+    interrupting the user's task.
+  </p>
+
+  <p class="dummy-paragraph">
+    We currently have four types of alerts, slightly different one from another
```
MelSumner, line 14:
```suggestion
    There are four types of alerts, each slightly different one from another. 
```

packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -0,0 +1,844 @@
...
+    <strong>description</strong>
+    (hence they are both required for this type of alert).</p>
+  <p class="dummy-paragraph">The default icon is also slightly different from the other alert types: it's filled instead
+    of outlined.
+  </p>
+
+  <h4 class="dummy-h5">Toast</h4>
+  <p class="dummy-paragraph">
+    This alert is intended for
+    <strong>user-triggered messages</strong>.</p>
```
MelSumner, line 77:
```suggestion
    This alert is intended for
    <strong>messages that are the result of a user's actions</strong>.</p>
```

packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -0,0 +1,844 @@
...
+  </dl>
+</section>
+
+<section>
+  <h3 class="dummy-h3">How to use</h3>
+
+  <h4 class="dummy-h4">Basic use</h4>
+  <p class="dummy-paragraph">
+    The most basic invocation requires the
+    <code class="dummy-code">type</code>,
```
MelSumner, line 216: Question: why don't we have a default `type`?

packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -0,0 +1,844 @@
...
+    <strong>actions</strong>
+    (optional) and a
+    <strong>close</strong>
+    button (optional).</p>
+  <p class="dummy-paragraph">
+    🚨
+    <em><strong>Important</strong>: we provide only the visual styling to this alert, so other features like placement,
+      animations/transitions, and what happens on dismiss, will need to be implemented in your app (eg. with an Ember
+      addon such as
+      <a href="https://github.com/poteto/ember-cli-flash">ember-cli-flash</a>).
```
MelSumner, line 99: We should not be recommending this addon, since it uses component helpers. See: https://github.com/embroider-build/embroider/blob/main/REPLACING-COMPONENT-HELPER.md

packages/components/addon/components/hds/alert/index.hbs
```diff
@@ -0,0 +1,40 @@
...
+            LinkTo::Standalone=(component "hds/link-to/standalone" size="small")
+          )
+          to="actions"
+        ~}}
+      </div>
+    {{~/if~}}
+  </div>
+
+  {{#if this.onClose}}
+    <button class="hds-alert__dismiss" type="button" {{on "click" this.onClose}}>
```
MelSumner, line 36: [issue, blocker] this is missing the `aria-label` attribute.

packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -0,0 +1,844 @@
...
+        <strong>and/or</strong>
+        <code class="dummy-code">title</code>
+        is required.
+      </p>
+    </dd>
+
+    <dt>onClose <code>function</code></dt>
+    <p>
+      The alert can be dismissed by the user. Pass in a function to display this optional parameter.
+    </p>
```
MelSumner, line 188: [issue] This needs to be inside of a `<dd>` element: 
```suggestion
    <dd><p>
      The alert can be dismissed by the user. Pass in a function to display this optional parameter.
    </p></dd>
```
image

packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -0,0 +1,844 @@
...
+      <code class="dummy-code">role="status"</code>, as the content is advisory information for the user, but is not
+      important enough to justify a
+      <code class="dummy-code">role="alert"</code>. Elements with
+      <code class="dummy-code">role="status"</code>
+      have an implicit
+      <code class="dummy-code">aria-live="polite"</code>
+      and
+      <code class="dummy-code">aria-atomic="true"</code>, so we have included them explicitly, for clarity.
+    </li>
+    <ul>
```
MelSumner, line 670: This must also be inside of an `<li>` element (see explanation above)

packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -0,0 +1,844 @@
...
+      <code class="dummy-code">onClose</code>
+      function in your app.
+    </li>
+    <li class="dummy-paragraph">
+      We do not have a
+      <code class="dummy-code">role="alertdialog"</code>
+      yet, as this type of alert interrupts the user's workflow to communicate an important message and acquire a
+      response.
+    </li>
+    <ul>
```
MelSumner, line 698: This must also be in an `<li>` element.

packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -0,0 +1,844 @@
...
+    <code class="dummy-code">Button</code>,
+    <code class="dummy-code">Link::Standalone</code>
+    or
+    <code class="dummy-code">LinkTo::Standalone</code>
+    yielded components.</p>
+  <p class="dummy-paragraph"><code class="dummy-code">Button</code>
+    is a yielded
+    <code class="dummy-code">HDS::Button</code>
+    component, so it exposes exactly
+    <LinkTo @route="components.button">the same API</LinkTo>
```
MelSumner, line 381: [issue] Links with the same name must serve the same purpose. If they go to different URLs they need to have different names. In this case, I would change the text to "the button component's API" (or even exclude this section entirely, as I am not sure it provides any additional clarity or value). 

There are a few instances of this issue in this page. 

image

packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -0,0 +1,844 @@
...
+      @type="inline"
+      @color="critical"
+      @description="With the dismiss button and no title"
+      @onClose={{this.noop}}
+    />
+  </div>
+
+</section>
+
+<section>
```
MelSumner, line 630: Please update this section to be more aligned with this: https://design-system-components-hashicorp.vercel.app/components/dropdown#accessibility

MelSumner:
I wonder if we should still have a working example in the documentation, even if users need to implement some of the alert logic themselves. For example, we could have a button that when pressed, inserted the alert (or reloaded the page to insert the alert, or provided a child page that had alerts rendered on navigation, etc). WDYT?

packages/components/tests/dummy/app/templates/components/alert.hbs
didoo: done.

packages/components/tests/dummy/app/templates/components/alert.hbs
didoo: done.

packages/components/tests/dummy/app/templates/components/alert.hbs
didoo: done.

packages/components/tests/dummy/app/templates/components/alert.hbs
didoo: Because there are no meaningful "defaults" (it depends on the context where it's used).

packages/components/tests/dummy/app/templates/components/alert.hbs
didoo: Ok, removed.

packages/components/tests/dummy/app/templates/components/alert.hbs
didoo: fixed

packages/components/tests/dummy/app/templates/components/alert.hbs
didoo: fixed

packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -0,0 +1,844 @@
...
+      => then
+      <code class="dummy-code">role="alert"</code>, as it's an important message.
+    </li>
+    <ul>
+      <li class="dummy-paragraph">Reference:
+        <a href="https://www.w3.org/TR/wai-aria-practices-1.2/#alert" target="_blank" rel="noopener noreferrer">
+          WAI-ARIA Authoring Practices 1.2 (Alert)
+        </a>
+      </li>
+    </ul>
```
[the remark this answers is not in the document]
didoo: removed content, not relevant anymore

packages/components/tests/dummy/app/templates/components/alert.hbs
didoo: removed content, not relevant anymore

packages/components/tests/dummy/app/templates/components/alert.hbs
didoo: removed content, not relevant anymore

packages/components/tests/dummy/app/templates/components/alert.hbs
didoo: done.

packages/components/addon/components/hds/alert/index.hbs
didoo: done

packages/components/tests/integration/components/hds/alert/index-test.js
```diff
@@ -0,0 +1,214 @@
...
+      hbs`<Hds::Alert @type="inline" @description="yo" id="test-alert" as |A|><A.Link::Standalone @icon="plus" @text="I am a link" href="#" @size="small" @color="secondary" /></Hds::Alert>`
+    );
+    assert
+      .dom(
+        '#test-alert .hds-alert__actions a.hds-link-standalone.hds-link-standalone--size-small.hds-link-standalone--color-secondary'
+      )
+      .exists()
+      .hasText('I am a link');
+  });
+
```
alex-ju, line 140: As I understand a lot of the head scratching on this component was done to make sure it supports conditional statements (or other logic) within the main block. To make this intention more explicit and prevent future refactoring from omitting this, may I suggest we add a test to cover this scenario?

Rough suggestion
```suggestion
  test('it should render an Hds::Button component when conditionally yielded to the "actions" container', async function (assert) {
    await render(hbs`
      <Hds::Alert @type="inline" @title="alert text" id="test-alert" as |A|>
        {{#if true}}
          <A.Button @text="Secondary" @color="secondary" />
        {{/if}}
      </Hds::Alert>
    `);
    assert.dom('#test-alert .hds-alert__actions button.hds-button').exists().hasText('Secondary');
  });
```

packages/components/tests/integration/components/hds/alert/index-test.js
```diff
@@ -0,0 +1,214 @@
...
+  test('it should not render the "dismiss" button by default', async function (assert) {
+    await render(hbs`<Hds::Alert @type="inline" @title="yo" />`);
+    assert.dom('button.hds-alert__dismiss').doesNotExist();
+  });
+  // TODO! how can this be done??
+  // skip('it should render the "dismiss" button if a callback function is passed to the @onDismiss argument', async function (assert) {
+  //   await render(hbs`<Hds::Alert @description="yo" @onDismiss={{NOOP}} />`);
+  //   assert.dom('button.hds-alert__dismiss').exists();
+  // });
```
alex-ju, line 161: As an Ember novice, I don't yet have an answer for this, but can dig into it.

alex-ju (commented):
Looks great to me! 🎉  and provides the flexibility of using logic within the main block, which I understand it eases the adoption. I admit that I played a bit with the branch to see if we can avoid rendering the `hds-alert__actions` element when no buttons or links are provided, but I couldn't find a solution, so I guess we'll have to live with it and rely on CSS.

didoo:
> Looks great to me! 🎉 and provides the flexibility of using logic within the main block, which I understand it eases the adoption. 

Thanks for the review!

> I admit that I played a bit with the branch to see if we can avoid rendering the `hds-alert__actions` element when no buttons or links are provided, but I couldn't find a solution, so I guess we'll have to live with it and rely on CSS.

Thanks for trying :)

packages/components/app/styles/components/toast.scss
```diff
@@ -0,0 +1,16 @@
...
+
+.hds-toast {
+  // we apply an elevation to the "alert/inline" element
+  box-shadow: var(--token-elevation-higher-box-shadow);
+
+  // per design specs
+  min-width: min(360px, 80vw);
+  max-width: min(500px, 80vw);
+  width: fit-content;
+}
```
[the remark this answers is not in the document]
didoo: Yes, this is a commit I added this morning. It's more of a safeguard than an actual "requirement". Our applications (as in "products") are not responsive to mobile viewport, but this was the first component that had a potential size that would outflow a mobile screen width, so I decided to use this `min()` function in CSS to add an extra safety. 
But if you have alternative/better ideas, I'm all ears!

packages/components/app/styles/components/alert.scss
```diff
@@ -0,0 +1,263 @@
...
+  }
+}
+
+// ACTIONS
+
+.hds-alert__actions {
+  align-items: center;
+  display: flex;
+
+  > * {
```
Dhaulagiri, line 97: [question] are these still necessary now that we've moved to the contextual component approach vs named block? I thought no, but very possible I misunderstood.

packages/components/tests/integration/components/hds/alert/index-test.js
```diff
@@ -0,0 +1,214 @@
...
+    assert.dom('#test-alert .hds-alert__content pre').exists().hasText('test');
+  });
+
+  // DISMISS
+
+  test('it should not render the "dismiss" button by default', async function (assert) {
+    await render(hbs`<Hds::Alert @type="inline" @title="yo" />`);
+    assert.dom('button.hds-alert__dismiss').doesNotExist();
+  });
+  // TODO! how can this be done??
```
Dhaulagiri, line 157: what is the question here? how to create a fake callback function to pass into the component or something else?

packages/components/app/styles/components/toast.scss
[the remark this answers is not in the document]
alex-ju: It makes sense – wanted to make sure this was the intent. I guess we can review it when/if we would be addressing the page layout more holistically.

packages/components/app/styles/components/alert.scss
didoo: Yes, we still need this. This is the "solution" :)

packages/components/tests/integration/components/hds/alert/index-test.js
```diff
@@ -0,0 +1,214 @@
...
+  // ACTIONS
+
+  test('it should render an Hds::Button component yielded to the "actions" container', async function (assert) {
+    assert.expect(2);
+    await render(
+      hbs`<Hds::Alert @type="inline" @description="yo" id="test-alert" as |A|><A.Button @text="I am a button" @size="small" @color="secondary" /></Hds::Alert>`
+    );
+    assert
+      .dom(
+        '#test-alert .hds-alert__actions button.hds-button.hds-button--size-small.hds-button--color-secondary'
```
Dhaulagiri, line 123: [nitpick] (non-blocking) my experience has been that chaining selectors together like this makes future refactoring trickier. is there a way we could simplify these to more directly target what we're looking for in the DOM?

packages/components/tests/integration/components/hds/alert/index-test.js
didoo: Yes, because this:
```
hbs`<Hds::Alert @description="yo" @onDismiss={{NOOP}} />`
``` is not a component invocation but a string so I am not sure how to "pass" a function to the handlebars (`hbs`) interpreter (in this case, the `NOOP` function (or any inline function).

If you comment the `NOOP` function at line 6, change `skip` to `test` and run the test, you will see this error:
<img width="715" alt="screenshot_1414" src="https://user-images.githubusercontent.com/686239/167144969-48d3d210-0f13-4d8c-8283-ae1e4c111650.png">

File: packages/components/tests/integration/components/hds/alert/index-test.js (hunk 18 above)
Comment (Dhaulagiri) [reply packages/components/tests/integration/components/hds/alert/index-test.js:123]: I personally always leaned heavily on the use of `data-test-*` attributes while writing tests but I know there are other approaches. Not necessary to completely resolve to one approach for this PR but something I think is worth thinking about as part of our overall strategy.

File: packages/components/tests/integration/components/hds/alert/index-test.js (hunk 13 above)
Comment (didoo) [reply packages/components/tests/integration/components/hds/alert/index-test.js:140]: So, not exactly. What we were testing the other day, was adding a condition(al) inside a named block. The syntax looks like this (pseudo-code):
```
<Hds::Alert @type="inline" @title="alert text" id="test-alert">
    <:actions as |A|>
        {{#if true}}
            <A.Button @text="Secondary" @color="secondary" />
        {{/if}}
    </:actions>
</Hds::Alert>
```
In this case (the one you suggested above) we know it works, it's a normal yield. And even if we wanted to test it, I am not sure we could test what was the problem (the extra visible space above the "actions").

Comment (alex-ju) [reply packages/components/tests/integration/components/hds/alert/index-test.js:142]: I see, it's a problem only with named blocks. Nevermind then 😬

File: packages/components/tests/integration/components/hds/alert/index-test.js (hunk 18 above)
Comment (didoo) [reply packages/components/tests/integration/components/hds/alert/index-test.js]: 100% agree with you. It's super brittle. I have pushed a change that alleviates the problem (and is more semantic for what I was testing): https://github.com/hashicorp/design-system/pull/245/commits/fe7aea918b8436751454d826c35568757881936d Let me know if that works for you.

Thanks for noticing it.

File: packages/components/tests/integration/components/hds/alert/index-test.js (hunk 17 above)
Comment (Dhaulagiri) [reply packages/components/tests/integration/components/hds/alert/index-test.js:159]: I need to resolve this so I can try to make a suggestion in the UI. Wish me luck.

File: packages/components/tests/integration/components/hds/alert/index-test.js (hunk 18 above)
Comment (didoo) [reply packages/components/tests/integration/components/hds/alert/index-test.js]: > I personally always leaned heavily on the use of data-test-*

Yes, but at the moment we've avoided scattering the components' code with `data-test-*` attributes everywhere, and I would take the risk and try to avoid doing it (until it lasts). Maybe we can get away with it :)
If later we'll see that there's no other way, we will systematically add these attributes and update the tests accordingly.

PS: if you want to have a broader discussion on this, maybe open an issue so we have tracking of the decisions made.

File: packages/components/tests/integration/components/hds/alert/index-test.js [hunk 19]
```diff
@@ -0,0 +1,216 @@
...
+
+  test('it should not render the "dismiss" button by default', async function (assert) {
+    await render(hbs`<Hds::Alert @type="inline" @title="yo" />`);
+    assert.dom('button.hds-alert__dismiss').doesNotExist();
+  });
+  // TODO! how can this be done??
+  // skip('it should render the "dismiss" button if a callback function is passed to the @onDismiss argument', async function (assert) {
+  //   await render(hbs`<Hds::Alert @type="inline" @description="yo" @onDismiss={{NOOP}} />`);
+  //   assert.dom('button.hds-alert__dismiss').exists();
+  // });
```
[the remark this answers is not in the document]
didoo: TIL today! Thanks @Dhaulagiri! 🙏
Fixed in https://github.com/hashicorp/design-system/pull/245/commits/b3c9b29a5c500830a9d97dbbb3a598d2d490d125

packages/components/tests/integration/components/hds/alert/index-test.js
didoo: Fixed in https://github.com/hashicorp/design-system/pull/245/commits/b3c9b29a5c500830a9d97dbbb3a598d2d490d125

packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -0,0 +1,730 @@
...
+<section>
+  <h3 class="dummy-h3" id="overview"><a href="#overview" class="dummy-link-section">§</a> Overview</h3>
+
+  <p class="dummy-paragraph">
+    An Alert is an element intended for
+    <strong>system-generated messages</strong>
+    (for messages that are the result of a user's actions see the
+    <LinkTo @route="components.toast">Toast</LinkTo>
+    component.)
+  </p>
```
MelSumner, line 14:
```suggestion
  <p class="dummy-paragraph">
    An Alert is an element intended for
    <strong>system-generated messages</strong>. It is a live region with important, usually time-sensitive information. The use of this alert component will cause immediate notifications for users with assistive technology. Since alerts are not required to receive focus, it should not be required that the user close the alert.<p>
    <p class="dummy-paragraph">For messages that are the result of a user's actions see the
    <LinkTo @route="components.toast">Toast</LinkTo>
    component.)
  </p>
```

packages/components/tests/integration/components/hds/alert/index-test.js
```diff
@@ -0,0 +1,216 @@
...
+      hbs`<Hds::Alert @type="inline" @description="yo" @color="warning" />`
+    );
+    assert.dom('.flight-icon-alert-triangle').exists();
+    await render(
+      hbs`<Hds::Alert @type="inline" @description="yo" @color="critical" />`
+    );
+    assert.dom('.flight-icon-alert-diamond').exists();
+  });
+
+  test('if an icon is declared, the icon should render in the component and override the default one', async function (assert) {
```
MelSumner, line 62: [suggestion] should there be a test here to ensure that the "Default one" renders?

packages/components/tests/integration/components/hds/alert/index-test.js
```diff
@@ -0,0 +1,216 @@
...
+  });
+  test('it should render both the title and the description when both the @title and @description arguments are provided', async function (assert) {
+    assert.expect(2);
+    await render(
+      hbs`<Hds::Alert @type="inline" @title="This is the title" @description="This is the description" />`
+    );
+    assert.dom('.hds-alert__title').hasText('This is the title');
+    assert.dom('.hds-alert__description').hasText('This is the description');
+  });
+  test('it should render rich HTML when the @description argument contains HTML tags', async function (assert) {
```
MelSumner, line 101: [comment] I'm surprised that we are even offering this.

didoo: > [comment] I'm surprised that we are even offering this.
we have multiple of this cases [in the audit](https://www.figma.com/file/h3LtloARmbYVwh7z6XWHVb/?node-id=3%3A4), we have to support them. and we're (more correctly, we'll be) supporting 
basic HTML elements like `a`, `strong`, `em`, `code` ([see CSS and comments](https://github.com/hashicorp/design-system/blob/b3c9b29a5c500830a9d97dbbb3a598d2d490d125/packages/components/app/styles/components/alert.scss#L65-L88))

packages/components/tests/integration/components/hds/alert/index-test.js
didoo: It's the previous test (`'it should render an icon by default depending on the type and color'`)

packages/components/app/styles/components/alert.scss
```diff
@@ -0,0 +1,263 @@
...
+
+  &.hds-alert--type-inline {
+    background-color: var(--token-color-surface-faint);
+    border-color: var(--token-color-border-strong); // notice: in the "neutral" color the "inline" has a slightly darker border color compared to the others to increase contrast (eg. could be used on a light gray background)
+  }
+
+  &.hds-alert--type-compact {
+    // special override by design
+    .hds-alert__icon {
+      color: var(--token-color-foreground-faint);
```
cveigt, line 188: This override applies to `page`, `inline` type alerts and `toast` as well.

packages/components/tests/dummy/app/templates/components/alert.hbs
```diff
@@ -0,0 +1,732 @@
...
+      <A.Button @text="Action" @color="secondary" />
+      <A.Link::Standalone @icon="plus" @text="Action" href="#" @color="secondary" />
+      <A.Generic>
+        <div class="dummy-alert-sample-custom-content-after-actions">This for example could be extra text, specific for
+          a special use case.</div>
+      </A.Generic>
+    </Hds::Alert>
+  </div>
+
+  <h4 class="dummy-h4">Dismiss</h4>
```
cveigt, line 700: @didoo can we change the `@color` for all these alerts for something different? maybe neutral?
We are recommending setting critical alerts as persistent in the usage guidelines.

cveigt (changes requested):

packages/components/app/styles/components/alert.scss
```diff
@@ -0,0 +1,263 @@
...
+  line-height: var(--token-typography-body-200-line-height);
+
+  .hds-alert--type-compact & {
+    font-family: var(--token-typography-body-100-font-family);
+    font-size: var(--token-typography-body-100-font-size);
+    line-height: var(--token-typography-body-100-line-height);
+  }
+}
+
+.hds-alert__title {
```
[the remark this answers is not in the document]
didoo: why do you think so? It makes sense (I have used `hds-alert__text` as place where to define the text style, because both `title` and `description` share the same font-family/font-size/line-height). introducing custom styles for everything goes against the purpose of having base/core styles (unless there are real reasons/differences, which is not the case here).

packages/components/app/styles/components/alert.scss
```diff
@@ -0,0 +1,263 @@
...
+  font-weight: var(--token-typography-font-weight-regular);
+
+  .hds-alert__title + & {
+    margin-top: 4px;
+  }
+
+  // we add very basic styling for elements that may be injected via the "description" string
+
+  strong {
+    font-weight: var(--token-typography-font-weight-medium);
```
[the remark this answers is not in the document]
didoo: @cveigt you mean to change it also in Figma? Ok for me, as soon as they're in sync. Let me know when you've made the change.

packages/components/addon/components/hds/alert/index.hbs
```diff
@@ -0,0 +1,35 @@
+<div class={{this.classNames}} role="alert" ...attributes>
```
MelSumner, line 1:
```suggestion
<div class={{this.classNames}} role="alert" aria-live="polite" ...attributes>
```
After doing a lot of testing on this on Friday and over the weekend, I think we need to add `aria-live="polite"` here.  Right now, our alerts are interrupting _each other_ which is not desirable behavior. 

An element with the role of `alert` implicitly has `aria-atomic="true"` and `aria-live="assertive"`. We want the `aria-atomic` attribute to stay the same so that they user has the context of the entire alert message and associated text/actions; however we want `aria-live` to be polite so it will finish waiting until the thing that is being read to the user with a screen reader is read out all the way. In this way, setting it to `polite` should at least allow the first alert message to be read out before the second alert message is read out, etc.

packages/components/app/styles/components/alert.scss
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  code, pre {
+    font-family: var(--token-typography-font-stack-code);
```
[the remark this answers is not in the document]
didoo: I didn't use the `code-100-font-size/line-height` because they would be `13px/18px` which is smaller than the styles used for the `description` text (`body-200-font-size/line-height` = `18px/20px`) so I just needed the `font-family` in this case.

packages/components/app/styles/components/alert.scss
didoo: fixed in https://github.com/hashicorp/design-system/pull/245/commits/21e86e010aa5f65d0f332ffc2288f9b3071d8511

packages/components/app/styles/components/alert.scss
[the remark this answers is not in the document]
cveigt: I'm not sure what you mean by "change it also in Figma", but we don't have a specific style for `strong` in the Foundations or HDS components file.

If you mean in the Alert itself; the one in the Figma docs already uses semibold.

packages/components/app/styles/components/alert.scss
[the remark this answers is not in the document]
cveigt: Ok, let's leave it as is for now. We should have a separate discussion to discuss our code style in general because it seems like the one we have is intended to be used as a code block and not within body text, so we may need to decide whether we want to add another style and tokens or not.

packages/components/app/styles/components/alert.scss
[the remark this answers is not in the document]
didoo: Ahh now I see what you mean. I'm not sure why I used `medium` (maybe before it was medium weight?)

packages/components/app/styles/components/alert.scss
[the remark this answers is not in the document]
didoo: > We should have a separate discussion to discuss our code style in general because it seems like the one we have is intended to be used as a code block and not within body text, so we may need to decide whether we want to add another style and tokens or not.

@cveigt can you please set up a meeting for next week, me you and @heatherlarsen so we can align on what steps to take for this? Thanks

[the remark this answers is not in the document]
heatherlarsen: Just as an FYI, the need for additional code styles has already come up as DevDot relies on them heavily. There is an [Asana task](https://app.asana.com/0/1202008719910219/1202214539819773) (internal) for this work created in the backlog.

packages/components/app/styles/components/alert.scss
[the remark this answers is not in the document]
cveigt: I'm closing this discussion after our meeting. We'll keep it as is.

MelSumner (approved):
I think this is good enough to ship, let's keep an eye out for how the accessibility shakes out when it's implemented. 👍

cveigt (approved):
LGTM 🚀

PR #200, opened 2022-04-11
"Dropdown" component - Finalize implementation in code (feature branch)

didoo:
### :pushpin: Summary

This is the new "feature branch" used to finalize the implementation of the `Dropdown` component.
The following PRs need to be reviewed and merged in this order:

1. #201 - ✅
2. #202 - ✅ (with new issue opened [here](https://github.com/hashicorp/design-system/issues/228))
3. #203 - ✅
4. #204 - ✅
5. #205 - ✅
6. #211 - ✅
7. #212 - ✅
8. #215 - 🤔
9. #234 - ✅
10. #235 - ✅
11. #236 - ✅
12. #243 - ✅
13. #242 - ✅

Once they're all merged (if approved) in this branch, we'll do a final review of the code to make sure we're not missing or forgetting something, and once this PR is approved it can be merged in `main` and the component officially released.

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable
- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202108023378513

PR #247, opened 2022-04-26
copy-item icon size

MelSumner:
Resolves issue where the `copy-item` icon was too small when text was super long. 

(I wonder if it would make more sense for @didoo to pull the relevant code lines into his most recent PR set? Leaving this in draft state for that reason)

Screenshot showing that the icon no longer shrinks:
CleanShot 2022-04-26 at 10 38 12

---
  - https://app.asana.com/0/0/1202167094155066

### DIFF
--- a/packages/components/app/styles/components/dropdown.scss
+++ b/packages/components/app/styles/components/dropdown.scss
@@ -169,6 +169,7 @@
     border: 1px solid var(--token-color-border-primary);
     color: var(--token-color-foreground-primary);  
     display: flex;
+    flex: 1 0 0;
     font-family: var(--token-typography-font-stack-code);
     justify-content: space-between;
     padding: 12px 8px;
@@ -217,6 +218,7 @@
     .hds-dropdown-list-item__copy-item-icon {
       color: var(--token-color-foreground-action);
       margin-left: 0.5rem;
+      min-width: 16px;
     }
   }
 }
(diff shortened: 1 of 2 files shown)


packages/components/app/styles/components/dropdown.scss
```diff
@@ -169,6 +169,7 @@
     border: 1px solid var(--token-color-border-primary);
     color: var(--token-color-foreground-primary);  
     display: flex;
+    flex: 1 0 0;
```
didoo, line 172: You don't need this.
```suggestion
```

didoo (changes requested):
See my comments.

didoo:
> (I wonder if it would make more sense for @didoo to pull the relevant code lines into his most recent PR set? Leaving this in draft state for that reason)

The best thing would be to PR against the "feature branch" `dropdown-finalize-implementation` (see https://github.com/hashicorp/design-system/pull/200). The problem I suspect is that if you do it now, editing this PR, it will have very complex conflicts. So my suggestion is:
- delete your local `melsumner/copy-item-icon-fix` branch (but don't push!)
- branch from `dropdown-finalize-implementation` using the same name (`melsumner/copy-item-icon-fix`)
- make the changes in this new branch (it's 3 lines to change)
- force push the newly created branch `melsumner/copy-item-icon-fix`
- this will overwrite the commits in this PR
- edit this PR and make it point against `dropdown-finalize-implementation`

MelSumner:
Closed in favor of #253

PR #255, opened 2022-04-28
[WIP] Add icon support to "Link::Inline" component

didoo:
### :pushpin: Summary

This is a spike to understand if it's possible to add support for `icon` to the `Link::Inline` component.

Now, while the results show that it's technically possible, I am a bit wary of implementing it, because is likely not used that much (and can be achieved independently, because the content is `yielded` so the consumer can add generic content (`icon + text` included) and organize its layout how the see fit. 

Another reason for this, is that achieving this in Figma is not possible, because is not a "block" component, so it's more of a style than a component (and you can't apply icons to styles). I can expand on this if it's not obvious what I mean.

Personally I would prefer to document how to do it in the "How to use" section, with a real example, than doing it as done in this PR (likely this will solve only a sub-set of all the possible cases).


didoo:
Decided to merge the branch and ask for a single review instead of two, to speed up the process.

PR #256, opened 2022-04-28
Fix tokens css reference in `package/components`

alex-ju:
### :pushpin: Summary

Fix `design-system-tokens/dist/products/css` reference in `package/components/ember-cli-build.js` and two tweaks to contribution docs.

Currently, running `ember test` in `packages/components` fails with 'Directory not found' because `node_modules` are currently under `packages/components` as a sibling of `ember-cli-build.js` script – so we update the reference accordingly

### Note
Do I need to be added to a team in Vercel? 😬

### DIFF
--- a/packages/components/CONTRIBUTING.md
+++ b/packages/components/CONTRIBUTING.md
@@ -3,7 +3,7 @@
 ## Installation
 
 * `git clone <repository-url>`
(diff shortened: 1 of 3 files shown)


alex-ju:
ah, we have a `workspace` setup at the root level – nice! this may only be a docs update after all

PR #257, opened 2022-04-28
Small fix on component contribution doc

alex-ju:
### :pushpin: Summary

Update contribution docs to fix folder name and promote `yarn install` at the root level instead of package level.

The order of commands in the initial setup made me think `yarn install` is required at the package level.


Dhaulagiri (approved):
🎉

Issue #258, opened 2022-04-28
Dropdown: Visual differences between code and Figma


heatherlarsen:
While reviewing the new Dropdown component, I came across a few visual differences between what was implemented and what is in [Figma](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components?node-id=5633%3A16319) (internal resource). @didoo asked me to put these all in one issue since they're all likely related, but let me know if we need to break these up. Happy to adjust.

## Issue 1
An in-between active and focus style is being applied to the first ListItem by default when the menu is opened via a mouse click. A sub-issue here is that this style is also being retained on the first ListItem when the user moves the mouse over other ListItems.

### Results
| Chrome       | Safari |
| ----------- | ----------- |
| <img width="224" alt="chrome-mouse-style" src="https://user-images.githubusercontent.com/8553306/165665236-ded56b86-42bc-4d05-96af-3d041ef41b41.png">  <img width="230" alt="chrome-mouse-hover" src="https://user-images.githubusercontent.com/8553306/165804031-dbae20cb-7647-45c3-952e-d05fb4726ca2.png">    |  <img width="227" alt="safari-mouse" src="https://user-images.githubusercontent.com/8553306/165665596-66f9720d-faaf-4fbc-82ba-907c8a8f96f8.png"> <img width="233" alt="safari-hover" src="https://user-images.githubusercontent.com/8553306/165804359-edd8b57e-9d4b-408b-a3c3-4b9ba046d3a8.png"> |

### Expected
❓I know focus needs to automatically move to the first interactive ListItem when the menu is opened with the keyboard, but does focus need to be visible on the first interactive ListItem when the menu is opened with a mouse? I found [one article](https://www.sarasoueidan.com/blog/focus-indicators/) that mentions ways of not displaying focus states for mouse users. 

If focus **doesn't need** to be visible when the user clicks the toggle with a mouse, then the first Interactive ListItem should display in the `Default` style.
| Dropdown Opened      | User interacts with menu with a mouse |
| ----------- | ----------- |
| <img width="303" alt="default-list" src="https://user-images.githubusercontent.com/8553306/165664486-b138d9df-3169-4fb3-b7d1-0097e1b08912.png"> | <img width="213" alt="safari-active" src="https://user-images.githubusercontent.com/8553306/165804561-c0c5513b-b093-4a93-bb5b-380eff75ed24.png"> |

[...]

PR #259, opened 2022-04-28
"Dropdown" component - Fix issues with focus - Part 1

didoo:
### :pushpin: Summary

This is a possible fix for part of the problems related to focus states in `Dropdown::ListItem::Interactive` (see #258).

There will be a second round of fix, in which we want to review all the focus management for the components that rely on the `Disclosure` component (with the intent to fix also the bug seen in https://github.com/hashicorp/cloud-ui/pull/2362#issuecomment-1097909966).

In this PR I have:
- fixed the issues with the color of text + combination of `active+focus` states
  - in doing so, did a small refactoring to simplify even more the definition of the colors, and centralize the "logic" of the interactive states in a single place
- ported the changes from https://github.com/hashicorp/design-system/pull/261 into this branch.
- removed integration test

### 🔗 Links
- codepen that can be used/forked to do some tests: https://codepen.io/didoo/pen/OJzeVjG?editors=1100

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

---
  - https://app.asana.com/0/0/1202198964307759

### DIFF
--- a/.changeset/famous-singers-admire.md
+++ b/.changeset/famous-singers-admire.md
@@ -1,0 +1,6 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+- removed autofocus on first item for `Disclosure` component (and as a result also for `Breadcrum` and `Dropdown` components)
+- updated focus state treatment for `Dropdown` component
--- a/.changeset/famous-singers-admire.md
+++ b/.changeset/famous-singers-admire.md
@@ -2,5 +2,5 @@
 "@hashicorp/design-system-components": patch
 ---
 
-- removed autofocus on first item for `Disclosure` component (and as a result also for `Breadcrum` and `Dropdown` components)
+- removed autofocus on first item for `Disclosure` component (and as a result also for `Breadcrumb` and `Dropdown` components)
 - updated focus state treatment for `Dropdown` component
--- a/packages/components/addon/components/hds/disclosure/index.hbs
+++ b/packages/components/addon/components/hds/disclosure/index.hbs
@@ -7,7 +7,7 @@
       class="hds-disclosure__content"
       {{focus-trap
         isActive=this.isActive
-        shouldSelfFocus=false
+        shouldSelfFocus=true
         focusTrapOptions=(hash clickOutsideDeactivates=this.clickOutsideDeactivates onDeactivate=this.onDeactivate)
       }}
     >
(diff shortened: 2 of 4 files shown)


didoo:
@MelSumner can you have a look at this PR?
@heatherlarsen can you check this solves #258?
@alex-ju can you have a look too? 

Thanks

heatherlarsen:
@didoo This appears to resolve all of the issues I noted in #258. 

I did see a new issue popup in Firefox (but not in Chrome) though. I was able to tab to the ToggleButton, open the menu, and then tab to the first list item using the keyboard, but I couldn't tab to any other items in the list. Focus remained on that first list item.

MelSumner:
@heatherlarsen see [this post for how to resolve that issue](https://www.scottohara.me/blog/2014/10/03/link-tabbing-firefox-osx.html)

MelSumner (approved):
This seems to resolve the concerns we were having previously. Tested on macOS with Chrome, Safari and Firefox and was able to confirm that we are seeing what we expect. 

I am not entirely sure about the `shouldSelfFocus` change in addition to the CSS change (I was expecting one or the other but probably not both) but I am not concerned enough to object at this time.

alex-ju:
Leaving a comment to check my understanding. With this PR we're trying to address the 3 issues flagged in #258 as follows:
 - issue 1 is addressed in 13703d7e7c1d57d43b8c62b8871d47ddfa337ada by disabling the self-focus on the disclosed content (the first focusable item, in our scenarios)
 - issue 2 is addressed in a56d81f31fbaa8a63eb749a82880eaf4373cfe6d by updating the `:focus:active` styles while abstracting `color` and `action` definitions
 - issue 3 is addressed in 42da19c3967955734c3aa74d133b3ba8b46a5f38 by explicitly defining the text colour on `:focus` state to be the same with the text colour on `:active` state

didoo:
> I did see a new issue popup in Firefox (but not in Chrome) though. I was able to tab to the ToggleButton, open the menu, and then tab to the first list item using the keyboard, but I couldn't tab to any other items in the list. Focus remained on that first list item.

I checked in my local env and it works for me (in Firefox). Maybe you have to enable the tabbing on all the "control" elements, as per @MelSumner suggested link ?

didoo:
> Do we need a changelog entry for these fixes?

Yes, just pushed one.

.changeset/famous-singers-admire.md
```diff
@@ -0,0 +1,6 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+- removed autofocus on first item for `Disclosure` component (and as a result also for `Breadcrum` and `Dropdown` components)
```
alex-ju, line 5: typo on `Breadcrumb`

heatherlarsen (approved):
@MelSumner's suggested fix worked to resolve what I was seeing in Firefox. I'm seeing things as expected now. Thanks.

PR #261, opened 2022-04-29
resolves issue with focus on dropdown

MelSumner:
Potential resolution for the half-active focus for the dropdown. Changing a setting in the discloser component resolved the issue but it should be checked to see if it's the solution you want, so leaving it in draft mode. Feel free to close if this is not the approach you want to take.

### DIFF
--- a/packages/components/addon/components/hds/disclosure/index.hbs
+++ b/packages/components/addon/components/hds/disclosure/index.hbs
@@ -7,7 +7,7 @@
       class="hds-disclosure__content"
       {{focus-trap
         isActive=this.isActive
-        shouldSelfFocus=false
+        shouldSelfFocus=true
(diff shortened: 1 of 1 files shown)


didoo:
@MelSumner it works for me, if it works for you (I remember it was an accessibility requirement, no?)
A couple of things:
- can you fix also the integration tests (now are failing), I think one test can be removed
- can you point this PR to this branch `dropdown-fix-focus-part1` instead of main, so we use the other PR https://github.com/hashicorp/design-system/pull/259 to solve all the problems in #258 ?
Thanks

didoo:
@MelSumner ported these changes in #259, you can close this one. Thanks

MelSumner:
@didoo I intended this code to demonstrate that the fix could be implemented with one small change, so please feel free to not use it if you decided to take a different approach! No worries!

PR #265, opened 2022-05-02
npx ember-cli-update across all the ember packages

Dhaulagiri:
### :pushpin: Summary

This PR updates all our ember apps to use the same version of Ember and runs the `npx ember-cli-update` script to update blueprints/dependencies as needed.

- [x] `packages/components`
- [x] `packages/ember-flight-icons`
- [x] `packages/flight-website` 

***

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/flight-website/app/components/ds-icon-grid.hbs
```diff
@@ -14,7 +14,6 @@
         id="ds-input-search"
         class="ds-input-search"
         placeholder="e.g. arrow"
-        autofocus={{false}}
```
Dhaulagiri, line 17: this ran afoul of a newly enable rule:

```
$ ember-template-lint .
app/components/ds-icon-grid.hbs
  17:8  error  No autofocus attribute allowed, as it reduces accessibility by moving users to an element without warning and context  no-autofocus-attribute
```

MelSumner:
I don't know if we can move to this version of Ember yet; last I knew, the update that made ember-engines compatible with 4.0+ was still not merged. WDYT?

Dhaulagiri:
@MelSumner that's a good question. Ultimately I _think_ it's up to what version the consuming app is using to determine compatibility, assuming we don't introduce new behavior in our addon via 4.x that is then incompatible with an app using engines _and_ an older version of ember.

Do you have a link to the issue(s) about this on the engines side of things?

packages/flight-website/app/components/ds-icon-grid.hbs
MelSumner: I'll submit an issue for this; we're saying no autofocus which is the same thing

Dhaulagiri: Ahh, that make sense. So deleting this isn't necessarily a problem but it _should_ be able to stay without tripping up the linter.

MelSumner:
@Dhaulagiri ahh looks like it was finally merged - https://github.com/ember-engines/ember-engines/pull/798 👍

MelSumner (approved):
LGTM! 👍

PR #266, opened 2022-05-04
Fix links under Pre-Flight Checklist section

alex-ju:
### :pushpin: Summary

Fix links under Pre-Flight Checklist section

I assumed the CRD template is the internal one and the system of systems is what we now call the design system search. Do let me know if they're not correct or we don't want to expose them.


didoo (approved):
LGTM

Issue #268, opened 2022-05-04
Bug report: CopyItem shows success state for multiple copied items simultaneously


noelledaley:
## IS: Current behavior
When a dropdown has multiple `CopyItem`s, both items can appear copied at once:

https://user-images.githubusercontent.com/903288/166738940-8c215c3a-18de-42df-87d7-c0a6d61591d8.mov

## SHOULD BE: Expected behavior
Only one CopyItem should appear copied at a time.

## Steps to reproduce
1. Navigate to https://design-system-components-hashicorp.vercel.app/components/dropdown
2. Find the section for "ToggleButton + Title, Description, CopyItem, Separator"
3. Click to expand and open the example dropdown
4. Click to copy one item, and then the other
5. Notice you see the green copy icon (aka success state) for both 

## Possible solution
The dropdown should optionally be able to keep track of whether one of its child `CopyItem`s has been copied & clear the copied item when a new one is selected.

MelSumner:
That's a good point; I'm resolving some other issues with this component & can include an update for this as well.

PR #84, opened 2022-03-17
"CopyButton" component - MVP

didoo:
### :pushpin: Summary

This is a "temporary" implementation of the `CopyButton` component (like an MVP) that will be used for the implementation of the `Dropdown` component (so we relieve @MelSumner to implement also this component, and focus on the `Dropdown` items (one of which may be a "copy item" that simply will consume the `CopyButton` as child. 

Preview of the documentation page: https://hds-components-git-copybutton-component-mvp-hashicorp.vercel.app/components/copy-button

🚨 **IMPORTANT**: at the moment we have only implemented the layout and the basic states (likely will be fine-tuned by @heatherlarsen in Figma) so that @MelSumner can use it in her code. We have **not** implemented any behaviour or interaction or extra functionality. In reality, in the products this component does much more than just copying to clipboard a string: it can be without container (just text and icon), it can have a success message or a tooltip, it can have different text between label and actual clipboard text, etc. We will work on these only later, because we will have to go through the standard process to unify all the different implementations (see links below): audit, design, CRD, code, etc.

**TODOs**
- [ ] support wrapping of the content ([example](https://www.figma.com/file/DUjKDINH2ps9ggk6YsAVC6/Dropdown?node-id=4653%3A18803))

### :camera_flash: Screenshots

How it will be used in the `Dropdown`:

screenshot_1114

### :link: External links
- [Cloud UI - Clipboard Copy Button](https://unofficial-design-system-search.vercel.app/?path=/story/cloud-ui-storybook_clipboardcopybutton--clipboard-copy-button)
- [Cloud UI - CopyCard](https://unofficial-design-system-search.vercel.app/?path=/story/cloud-ui-storybook_copycard--copy-card-with-clipboard-text)
- [Nomad - Copy Button](https://unofficial-design-system-search.vercel.app/?path=/story/nomad-storybook_components-copy-button--copy-button)
- [Consul - Copy Button](https://unofficial-design-system-search.vercel.app/?path=/story/consul-copy-button--consul-copy-button)

***

### 👀 How to review

👉 Review by files changed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #270, opened 2022-05-05
[Dropdown] copy-item improvement(s)

MelSumner:
### :pushpin: Summary

If merged, this PR:
1. adds check to copy-item, to ensure that `this.args.text` is equal to `navigator.clipboard.readText` results, BEFORE setting `isSuccess` to `true`.
2. adds `isError` state

I've added `isError` to hook into  because it was convenient to add, has a small footprint, and seemed useful even if we don't have an exact plan for it just yet; totally fine with removing it if there are objections, though.

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

---
  - https://app.asana.com/0/0/1202167094155069


packages/components/addon/components/hds/dropdown/list-item/copy-item.js
```diff
@@ -37,12 +38,21 @@ export default class HdsDropdownListItemCopyItemComponent extends Component {
...
-  copyCode() {
-    navigator.clipboard.writeText(this.args.text);
-    // this if statement resolves to [object Promise] so maybe some improvements
-    // could be made here
-    if (navigator.clipboard.readText()) {
-      this.isSuccess = true;
+  async copyCode() {
+    // If there is text to be copied, then copy it to the clipboard.
+    // If the text in the clipboard is the same as the text to be copied, then set isSuccess to true.
+    // TODO: set any _other_ isSuccess to false, so there is only one indicated successful copy at a time.
```
didoo, line 44: [suggestion] what if instead we just set a timeout that resets the status?

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
```diff
@@ -4,7 +4,8 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 
 export default class HdsDropdownListItemCopyItemComponent extends Component {
-  @tracked isSuccess = this.args.isSuccess ?? false;
+  @tracked isSuccess = false;
+  @tracked isError = false;
```
didoo, line 8: here's my thought: or we do it properly (so we add an "error" visual state too) or we remove the `isError`.
personally I would go with removing it (we will implement the full functionality when we will build the `CopyButton` component, see #84)

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
```diff
@@ -37,12 +38,21 @@ export default class HdsDropdownListItemCopyItemComponent extends Component {
...
-    navigator.clipboard.writeText(this.args.text);
-    // this if statement resolves to [object Promise] so maybe some improvements
-    // could be made here
-    if (navigator.clipboard.readText()) {
-      this.isSuccess = true;
+  async copyCode() {
+    // If there is text to be copied, then copy it to the clipboard.
+    // If the text in the clipboard is the same as the text to be copied, then set isSuccess to true.
+    // TODO: set any _other_ isSuccess to false, so there is only one indicated successful copy at a time.
+    if (this.args.text) {
```
didoo, line 45: I think we can remove this check, because we have an assertion at line 17 that makes sure there is a @text value (otherwise would throw an assertion).

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
```diff
@@ -37,12 +38,21 @@ export default class HdsDropdownListItemCopyItemComponent extends Component {
...
+      navigator.clipboard.writeText(this.args.text);
```
didoo, line 46: - I think you need an `await` in front of it.
- is better to use `this.text` instead of `this.args.text` (in some cases, we alter the initial value, eg. with defaults).

```suggestion
      await navigator.clipboard.writeText(this.text);
```

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
```diff
@@ -37,12 +38,21 @@ export default class HdsDropdownListItemCopyItemComponent extends Component {
...
+      navigator.clipboard.readText().then((result) => {
+        if (result === this.args.text) {
+          this.isSuccess = true;
+          // console.log(`msg: ${result} and ${this.args.text} should be equal`);
+        } else {
+          this.isError = true;
+          // console.log(`copyCode was not successful (for whatever reason). One possible reason could be browser clipboard permissions.`);
+        }
+      });
```
didoo, line 55: - since we're using `async/await`, better to use `await` also here
- the rest of the code can be simplified
```suggestion
    const readText = await navigator.clipboard.readText();
    this.isSuccess = readText === this.text;
```

didoo (commented):
Added a few comments/suggestions

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
MelSumner: Yes, perhaps! Looking into it as part of a different issue though: https://github.com/hashicorp/design-system/issues/268

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
MelSumner: Interesting! This code is really hard for me to read, which would make it confusing to maintain. Is there any reason that the code as it's written now is incorrect? Happy to change it if that's the case, but would prefer to keep it a little more readable (and as a result a little easier to maintain) if it's not incorrect. WDYT?

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
MelSumner: So that's interesting, I didn't know we still planned to build a copy button; I intended for the copy-item to be the final implementation. Do we have a need for a different copy button still? If there are use cases that I missed, I would like to catch them now.

didoo: > I didn't know we still planned to build a copy button; I intended for the copy-item to be the final implementation
If you look at the Cloud UI codebase you can see that the `<ImportToTerraform>` it's using the `CopyCard` component (see https://github.com/hashicorp/cloud-ui/blob/0cb059241c392c32d9ca41d2a1293b0276fdfec0/packages/cloud-ui-core/addon/components/import-to-terraform/index.hbs#L16-L20). Now, this component is used not only in the `<ImportToTerraform>` but in may other places in Cloud UI: https://github.com/hashicorp/cloud-ui/search?q=%3CCopyCard
So we need to have a similar component, to avoid duplication of code. That's the whole reason why I started the `CopyButton` MVP: we will have that component, sooner or later (hopefully sooner).

MelSumner:
Closing this PR, will try a different approach.

PR #274, opened 2022-05-05
dropdown component bugfixes

### LINKED ISSUE #268: Bug report: CopyItem shows success state for multiple copied items simultaneously

MelSumner:
### :pushpin: Summary

If merged, this PR: 

1. updates the `copy-item` to use async/await. While we are using the [native web api for clipboard](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText), it was suggested that `async`/`await` is preferable to `.then` because it is more readable.
2. adds a `setTimeout` for the "success" state to fade back to the default after copying the item (would this sufficiently resolve #268 ?)
3. updates the documentation to include the copy-item API 
4. updates a mismatched/incorrect example with the invocation shown on the page

Remaining question(s):

1. Testing the copy-item action. While we have guards in place (comparing the result of `readText` with `this.args.text`), I've tried a few different approaches to write an accurate test for the action but it doesn't seem like I can get the clipboard action to accurately test - `promise rejected...document not focused`.  
2. The WCAG conformance of using `setTimeout` to change the state from success back to default after being copied.

Added the copy-item API: 
CleanShot 2022-05-05 at 16 39 34

Updated the code sample to match what was implemented:
CleanShot 2022-05-05 at 16 34 25

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

---
  - https://app.asana.com/0/0/1202167094155069

### DIFF
--- a/packages/components/addon/components/hds/dropdown/list-item/copy-item.js
+++ b/packages/components/addon/components/hds/dropdown/list-item/copy-item.js
@@ -4,7 +4,7 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 
 export default class HdsDropdownListItemCopyItemComponent extends Component {
-  @tracked isSuccess = this.args.isSuccess ?? false;
+  @tracked isSuccess = false;
 
   /**
    * @param text
@@ -37,12 +37,19 @@ export default class HdsDropdownListItemCopyItemComponent extends Component {
   }
 
   @action
-  copyCode() {
-    navigator.clipboard.writeText(this.args.text);
-    // this if statement resolves to [object Promise] so maybe some improvements
-    // could be made here
-    if (navigator.clipboard.readText()) {
+  async copyCode() {
+    // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard
+    await navigator.clipboard.writeText(this.args.text);
+    const result = await navigator.clipboard.readText();
+
+    if (result === this.args.text) {
       this.isSuccess = true;
+      // console.log(`result is ${result}`);
+
+      // make it fade back to the default state
+      setTimeout(() => {
+        this.isSuccess = false;
+      }, 1000);
     }
   }
 }
--- a/packages/components/tests/dummy/app/templates/components/dropdown.hbs
+++ b/packages/components/tests/dummy/app/templates/components/dropdown.hbs
@@ -153,9 +153,16 @@
     </dd>
   </dl>
 
-  <h4 class="dummy-h4">ListItem::CopyItem</h4>
-  {{! TODO: add props here if we decide to use the current version }}
-  <p class="dummy-paragraph">[WIP]</p>
+  <h4 class="dummy-h4" id="api-listitem-copyitem">ListItem::CopyItem</h4>
+  <dl class="dummy-component-props" aria-labelledby="api-listitem-copyitem">
+    <dt>copyItemTitle <code>string</code></dt>
+    <dd>Optional. Provides a contextual title for the text to be copied.</dd>
+    <dt>text <code>string</code></dt>
+    <dd>
+      <p>The text to be copied.</p>
+      <p><em>If no text value is defined an error will be thrown.</em></p>
+    </dd>
+  </dl>
 
   <h4 class="dummy-h4">ListItem::Description</h4>
   <p class="dummy-paragraph" id="component-api-listitem-description">Here is the API for the "description" list item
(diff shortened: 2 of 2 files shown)


packages/components/addon/components/hds/dropdown/list-item/copy-item.js
```diff
@@ -4,7 +4,7 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 
 export default class HdsDropdownListItemCopyItemComponent extends Component {
-  @tracked isSuccess = this.args.isSuccess ?? false;
+  @tracked isSuccess = false;
```
didoo, line 7: [nitpick/not blocker] maybe we could leave it `undefined` by default, so we can use the `false` state for when there is an error.

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
```diff
@@ -37,12 +37,19 @@ export default class HdsDropdownListItemCopyItemComponent extends Component {
...
+    await navigator.clipboard.writeText(this.args.text);
+    const result = await navigator.clipboard.readText();
+
+    if (result === this.args.text) {
       this.isSuccess = true;
+      // console.log(`result is ${result}`);
+
+      // make it fade back to the default state
+      setTimeout(() => {
+        this.isSuccess = false;
```
didoo, line 51: [nitpick/not blocker] see comment above about `false` vs `undefined`

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
```diff
@@ -37,12 +37,19 @@ export default class HdsDropdownListItemCopyItemComponent extends Component {
...
+  async copyCode() {
+    // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard
+    await navigator.clipboard.writeText(this.args.text);
+    const result = await navigator.clipboard.readText();
+
+    if (result === this.args.text) {
       this.isSuccess = true;
+      // console.log(`result is ${result}`);
+
+      // make it fade back to the default state
```
didoo, line 49: 👍👍👍

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
```diff
@@ -37,12 +37,19 @@ export default class HdsDropdownListItemCopyItemComponent extends Component {
...
-    navigator.clipboard.writeText(this.args.text);
-    // this if statement resolves to [object Promise] so maybe some improvements
-    // could be made here
-    if (navigator.clipboard.readText()) {
+  async copyCode() {
+    // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard
+    await navigator.clipboard.writeText(this.args.text);
+    const result = await navigator.clipboard.readText();
+
+    if (result === this.args.text) {
```
didoo, line 45: [nitpick/improvement] since we have a "getter" function for the value `this.text`, and potentially thiw function could be manipulating the value in the future, it would be more resilient if here we use `this.text` (and at line 41 too) instead of the argument passed.

didoo (approved):
👍

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
MelSumner: https://github.com/hashicorp/design-system/issues/275

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
MelSumner: there's a reason to use this.args.text but I need to look it up. I'll find the text so we can all have the same knowledge and will update this PR if I am not correct.

Dhaulagiri:
Definitely open to ideas on how to best remember this, but in order for this bugfix to be "released" this PR should have included a changset

didoo:
> Definitely open to ideas on how to best remember this, but in order for this bugfix to be "released" this PR should have included a changeset

My bad, again 🤦‍♂️

PR #279, opened 2022-05-10
Docs: adds a sidebar to component pages

MelSumner:
### :pushpin: Summary

Some consistent feedback is that our docs are good but a lot. In this PR, I've added a sidebar to the component's page, with the sections in the sidebar for hopefully easier navigation.

If this is useful to our ambassadors, we can tidy up the styling a bit and ship it. 

This screenshot shows the focused state, the default state, and the hover state. I've mainly stuck to defaults so that if @heatherlarsen and @cveigt have styling opinions there wouldn't be too much to undo.

CleanShot 2022-05-09 at 19 03 42

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/tests/dummy/app/components/dummy-toc.hbs
+++ b/packages/components/tests/dummy/app/components/dummy-toc.hbs
@@ -1,0 +1,8 @@
+<ul class="dummy-aside-list dummy-list">
+  <li class="dummy-aside-list-item"><a href="{{this.dummyRouteUrl}}">The Component</a></li>
+  <li class="dummy-aside-list-item"><a href="{{this.dummyRouteUrl}}#component-api">Component API</a></li>
+  <li class="dummy-aside-list-item"><a href="{{this.dummyRouteUrl}}#how-to-use">How To Use</a></li>
+  <li class="dummy-aside-list-item"><a href="{{this.dummyRouteUrl}}#design-guidelines">Design Guidelines</a></li>
+  <li class="dummy-aside-list-item"><a href="{{this.dummyRouteUrl}}#accessibility">Accessibility</a></li>
+  <li class="dummy-aside-list-item"><a href="{{this.dummyRouteUrl}}#showcase">Showcase</a></li>
+</ul>
--- a/packages/components/tests/dummy/app/styles/components/dummy-toc.scss
+++ b/packages/components/tests/dummy/app/styles/components/dummy-toc.scss
@@ -1,0 +1,22 @@
+.dummy-aside-list {
+  box-sizing: border-box;
+  list-style: none;
+  padding: 0;
+  margin: auto;
+}
+
+.dummy-aside-list-item {
+  box-sizing: border-box;
+  width: 100%;
+  a {
+    border: 2px solid transparent;
+    box-sizing: border-box;
+    padding: 0.5rem;
+    width: 100%;
+    display: flex;
+    &:hover {
+      background-color: aliceblue;
+      border-left-color: blue;
+    }
+  }
+}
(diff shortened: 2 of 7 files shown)


didoo (changes requested):
A few comments:
- Would it be possible to move the sidebar outside the `main` block so that it doesn't use the space used for the showcase of the components? 
- Do you think it would make sense to apply a `sticky` position to it?
- This is a general sidebar, but in some cases it doesn't reflect the actual content (eg. Badge page, or other pages where there are more section or less sections than the ones in the sidebar nav). Do you think this may be a problem for the users, that may think a link is broken in the sidebar?
- Can you update the CSS selector in the `tests/dummy/app/styles/app.scss` at the bottom, so that the sidebar nav is hidden too in the percy tests?
Thanks

MelSumner:
@didoo these are all great points! 

> * Would it be possible to move the sidebar outside the `main` block so that it doesn't use the space used for the showcase of the components?

Yes, should be possible with a little refactoring

> * Do you think it would make sense to apply a `sticky` position to it?

That would probably be a great idea, makes it even more useful

> * This is a general sidebar, but in some cases it doesn't reflect the actual content (eg. Badge page, or other pages where there are more section or less sections than the ones in the sidebar nav). Do you think this may be a problem for the users, that may think a link is broken in the sidebar?

Great point, I think we might have to hold on the idea until we're sure that all component pages are consistent if we use this particular sidebar approach. It might be the tradeoff of making "the least disruptive to the codebase" kind of sidebar. 

> * Can you update the CSS selector in the `tests/dummy/app/styles/app.scss` at the bottom, so that the sidebar nav is hidden too in the percy tests?

Good catch, can do.

packages/components/tests/dummy/app/styles/components/dummy-toc.scss
```diff
@@ -0,0 +1,22 @@
...
+  a {
+    border: 2px solid transparent;
+    box-sizing: border-box;
+    padding: 0.5rem;
+    width: 100%;
+    display: flex;
+    &:hover {
+      background-color: aliceblue;
+      border-left-color: blue;
+    }
```
heatherlarsen, line 20:
```suggestion
      border-top-right-radius: 5px;
      border-bottom-right-radius: 5px;
    }
```

Added a subtle design polish suggestion. Eventually would love to see us sort out a way to do the border-radius on all sides, but the use of `border-left-color` makes that challenging and I don't want to make this implementation more complicated right now. 

This is all good with me if we get @didoo's suggestions in place.

packages/components/tests/dummy/app/components/dummy-toc.hbs
```diff
@@ -0,0 +1,8 @@
+<ul class="dummy-aside-list dummy-list">
+  <li class="dummy-aside-list-item"><a href="{{this.dummyRouteUrl}}">The Component</a></li>
```
heatherlarsen, line 2: [suggestion] Maybe something like "Overview" instead of "The Component"?

cveigt:
I don't think it needs to be that big, so if it's doable, I'd suggest changing the font-size from `1` rem to `0.9` rem.
I'm not suggesting the change directly, as I noticed that other elements on the page are using the class `.dummy-list` too.

MelSumner:
Update: I'm a little bit blocked on this PR until: 

1. we work out what components are staying, since I need to add the accessibility section to the sidebar but not every page has that section yet
2. we also need to decide what to do re: badge/badge count, because right now both components are being rendered on the same page

If we want consistency in our documentation, this PR's approach will work. 
If we want more flexibility in our documentation (to document each component as we see fit) then we will need a more dynamic approach (if so, we should brainstorm about what that should be).

/cc @Dhaulagiri @didoo

MelSumner:
@didoo "probably a little old school" was exactly my thought too. I guess it's doable, and would unblock this. I'll go that route and will ping if I need help. Thanks for the feedback!

Issue #267, opened 2022-05-04
Bug report: Closing a dropdown causes scrolling


thrashr888:
## IS: Current behavior

When you open a dropdown, scroll away, then open another, the page scrolls back to the first dropdown when closing it.

## SHOULD BE: Expected behavior

The page should not scroll away so I can retain my context and awareness of what I'm doing.

## Steps to reproduce

1. https://design-system-components-hashicorp.vercel.app/components/dropdown
2. click the text toggle dropdown
3. scroll down
4. click the gear dropdown

didoo:
@thrashr888 thanks for submitting this issue. I noticed it myself, but I am not sure if it's a bug or a "feature". The reason why the page scrolls is because the focus is returned to the element that initially "opened" the dropdown (the first one) on click-outside, and Chrome scrolls to bring back the focused element in the viewport (while if you test this in Safari, it works as you say you would expect) which in this case is the `toggle` element of the first dropdown.

@MelSumner any thoughts on this?

thrashr888:
I sort of follow, but should clicking on a new thing focus on that instead? I think that even when clicking on whitespace, I wouldn't expect to scroll back up from that.

MelSumner:
@thrashr888 I'm going to dig through the ember-focus-trap/focus-trap docs to see if this is documented; then I'll try to document potential solutions.

Issue #280, opened 2022-05-10
Review of focus management in Disclosure (+ Dropdown/Breadcrumb) components


didoo:
@alex-ju as your first ticket, I would like you to review holistically the management of the `focus` in the `Disclosure` component (and therefore also in the `Dropdown` and `Breadcrumb` components, that use this generic component under the hood) and in the process fix a few issues/bugs that have been raised. 

Below I'll add as much information I have and context as possible, but please ask me or @MelSumner if there is something not clear. 

The general idea is that you could become the go-to person when comes to focus management in the HDS code library, in the long term :)

_Notice: I'll leave it to you how to organize/submit the PRs (single PR, multiple PRs, etc), no strong preferences._

## Known bugs and other possible issues

These are the bugs we know exist and we want to try to fix, and some other potential issues may be fixed/improved too.

### Fix "focus-trap" bug when the `interactive` opens a modal

There is a bug with the `focus-trap` when a modal is opened as result of one of the dropdown actions. See screen recording in the comment here: https://github.com/hashicorp/cloud-ui/pull/2362#issuecomment-1097778784
Not even sure how to debug/fix it, seems it's caused by the focus being returned to the dropdown list content when the modal is closed, but the dropdown content is not anymore there (but one can see it in the browser, so not sure this is the issue).

- existing Asana ticket: https://app.asana.com/0/1202008719910219/1202122650634732
- issue found while testing the Dropdown component in Cloud UI: https://github.com/hashicorp/cloud-ui/pull/2362#issuecomment-1097778784

_Notice: to test this particular bug in Cloud UI, you have to set up the Cloud UI repo, and then check out the branch I was using for testing. Since Cloud UI takes some time to compile, the development could be extremely slow (and painful). I suggest maybe first try to see if you can reproduce it in another way, locally in the dummy app, and use the Cloud UI only as last resort._

### Scrolling caused by focus returning to the previously active element

You can see the issue reported here: https://github.com/hashicorp/design-system/issues/267 and the comments/thread in it. I am not sure if there is a "solution" to it, if there is a better approach, or of this is how it should work. I'll leave it to you to investigate more on this subject (or maybe you already have an answer based on your previous experience).

### First item focused on dropdown open

This was fixed in https://github.com/hashicorp/design-system/pull/259 but the solution looks counterintuitive to me (the fix was to set `shouldSelfFocus=true` which to me does the contrary of what we wanted, so not sure why this fixed the problem).

[...]

PR #284, opened 2022-05-12
Change focus management in disclosure utility component

alex-ju:
### :pushpin: Summary

This PR is proposing an alternative approach to focus management in our disclosure utility component on the back of #280

The main change is dropping the focus-trap mechanism and letting the focus flow through the component while preserving the rest of the requirements defined in [the associated CRD](https://docs.google.com/document/d/1PSIAdUypvGWCLBAkeOo4HSa3qd6qN-pWvlLrfuwYT4o/edit#). The solution is meant to avoid any change the API of the component.

I propose this change to fix the issues described in #280 (briefly: "fix focus-trap bug when the interactive opens a modal", and the "scrolling caused by focus returning to the previously active element"). It also addresses the WCAG 2.1 success criterion [2.1.2 No Keyboard Trap](https://www.w3.org/TR/WCAG21/#no-keyboard-trap) (Level A) by allowing the users to exit the disclosed content using the default keyboard shortcut (tab).

<table>
<tr><td>

disclosure-focus-before

</td><td>

disclosure-focus-after

</td></tr></table>

Note: the colors went way off on the above compressed gifs 😱 

### :link: External links

[Review of focus management in Disclosure (+ Dropdown/Breadcrumb) components](https://github.com/hashicorp/design-system/issues/280)
Fixes https://github.com/hashicorp/design-system/issues/267 and https://github.com/hashicorp/cloud-ui/pull/2362#issuecomment-1097778784

***

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ ] +1 Percy if applicable

---
  - https://app.asana.com/0/0/1202253786016960

### DIFF
--- a/packages/components/addon/components/hds/disclosure/index.js
+++ b/packages/components/addon/components/hds/disclosure/index.js
@@ -32,6 +32,14 @@ export default class HdsDisclosureComponent extends Component {
   }
 
   @action
+  onKeyUp(event) {
+    if (event.key === 'Escape') {
+      this.deactivate();
+      this.toggleRef.focus();
+    }
+  }
+
+  @action
   deactivate() {
     if (this.isActive) {
       this.isActive = false;
--- a/packages/components/addon/components/hds/disclosure/index.js
+++ b/packages/components/addon/components/hds/disclosure/index.js
@@ -19,6 +19,8 @@ export default class HdsDisclosureComponent extends Component {
       this.toggleRef = event.currentTarget;
     }
     this.isActive = !this.isActive;
+    // we explicitly apply a focus state to the toggle element to overcome a bug in WebKit (see b8abfcf)
+    this.toggleRef.focus();
   }
 
   @action
--- a/packages/components/addon/components/hds/disclosure/index.js
+++ b/packages/components/addon/components/hds/disclosure/index.js
@@ -8,6 +8,11 @@ export default class HdsDisclosureComponent extends Component {
   @tracked isToggleClicked;
 
   @action
+  didInsert(element) {
+    this.element = element;
+  }
+
+  @action
   onClickToggle(event) {
     // we store a reference to the DOM node that has the "onClickToggle" event associated with it
     if (!this.toggleRef) {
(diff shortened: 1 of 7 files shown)


MelSumner:
Thank you for your PR! Can you please add some more details to the PR, explaining why you want to make this change, and what the intent is. Thank you!

didoo (changes requested):
@alex-ju I've found an issue with Safari that I suspect is a blocker: when the dropdown is activated/opened via keyboard (tab to the toggle, open the toggle via spacebar/enter) everything works as expected; but if the dropdown is opened via mouse click, then nor the click outside nor the "esc" key close the dropdown.

other things I notice, that may be helpful in narrowing down the problem: 
- if you add a `console.log` inside the `onFocusOut` function, you will see that is logged in Chrome (when the toggle is clicked to open it) but not in Safari
- in Safari, if you open the toggle via mouse click, but then hit `tab` and focus on the first item, then the click outside and the esc both trigger the close 

I did some tests, but I am not sure what would be the cause root, if something in the logic or a bug in Safari. I'll let you investigate.

This happens in Safari:
screenshot_1448

alex-ju:
> Thank you for your PR! Can you please add some more details to the PR, explaining why you want to make this change, and what the intent is. Thank you!

Sure, I should've been more clear on the why; I guess part of it was left in [the comment on the issue](https://github.com/hashicorp/design-system/issues/280#issuecomment-1124978813). I have added a paragraph to the PR description trying to articulate the motivation behind the proposed change.

alex-ju:
> @alex-ju I've found an issue with Safari that I suspect is a blocker: when the dropdown is activated/opened via keyboard (tab to the toggle, open the toggle via spacebar/enter) everything works as expected; but if the dropdown is opened via mouse click, then nor the click outside nor the "esc" key close the dropdown.
> 
> other things I notice, that may be helpful in narrowing down the problem:
> 
> * if you add a `console.log` inside the `onFocusOut` function, you will see that is logged in Chrome (when the toggle is clicked to open it) but not in Safari
> * in Safari, if you open the toggle via mouse click, but then hit `tab` and focus on the first item, then the click outside and the esc both trigger the close
> 
> I did some tests, but I am not sure what would be the cause root, if something in the logic or a bug in Safari. I'll let you investigate.

It's [a bug in webkit](https://trac.webkit.org/changeset/276698/webkit/), [a fix will be released soon](https://webkit.org/blog/11680/release-notes-for-safari-technology-preview-125/), but I'll add a patch in the meantime.

alex-ju:
@didoo I've added a patch for the inconsistency of WebKit not triggering `focusin` and `focusout` on button elements (among other types of elements) to address the issue you found and added your suggestions to the preview/documentation page.

packages/components/addon/components/hds/disclosure/index.js
```diff
@@ -7,41 +7,47 @@ export default class HdsDisclosureComponent extends Component {
...
+  }
+
   @action
   onClickToggle(event) {
     // we store a reference to the DOM node that has the "onClickToggle" event associated with it
     if (!this.toggleRef) {
       this.toggleRef = event.currentTarget;
     }
     this.isActive = !this.isActive;
+    this.toggleRef.focus();
```
didoo, line 22: [suggestion] generally, I tend to add comments to this kind of "things" (when I make a non-obvious fix for two reasons: a future me that will come in and don't remember why I did something + the fear that a future maintainer will come in and remove the code because they think is unnecessary).

didoo (approved):
All good for me 👍

MelSumner:
Is this a proposal or a foregone conclusion? I think I missed that part of the conversation entirely. Any documents that you can point me to? I don't think we should change the component without appropriate justification.

MelSumner (approved):
Update: After looking through it, I really think we are losing some A11y-specific UX here, but it's also technically a WCAG-conformant solution. As such, I disagree with this change but cannot object to it.

alex-ju:
> Update: After looking through it, I really think we are losing some A11y-specific UX here, but it's also technically a WCAG-conformant solution. As such, I disagree with this change but cannot object to it.

Thank you for the review, Melanie. From an accessibility perspective, my concern with the current implementation is that assistive tech users are not being told that they've entered a region when they moved into the disclosed content and they're not being told that if they want to get out of the region they can use the escape key (some may figure it out, but it's not explicit), which may trap and frustrate some of them. That being said, we'll most probably need this focus-trap-based solution for the modal dialogue component.

PR #285, opened 2022-05-12
[SPIKE]

didoo:
### :pushpin: Summary

🚨 This is a working branch used to explore and play with the APIs of the `form controls` components. **It is not for merging!**

The intention is to have a “spike/playground” branch with all the form controls components showcased in their own documentation pages, to use as “scrapbook” for their APIs (each component page will contain the _Component API_ section
and the _Showcase_ section).

The content will be scrappy, on purpose, and potentially will contain comments, links to external code, etc. Same for the code implementation. The idea is to collect all the notes and relevant information about the form controls coming from the other codebases in this PR, acting as a centralized place for this kind of research and analysis.

Later this code can be used as a base (or directly, via cherry picking or copy&pasting) for the final implementation of the components.

### DIFF
--- a/packages/components/app/styles/components/form/text-input.scss
+++ b/packages/components/app/styles/components/form/text-input.scss
@@ -1,0 +1,6 @@
+//
+// FORM > TEXT-INPUT
+//
+// properties within each class are sorted alphabetically
+//
+
--- a/packages/components/tests/dummy/app/templates/components/form/text-input.hbs
+++ b/packages/components/tests/dummy/app/templates/components/form/text-input.hbs
@@ -1,0 +1,66 @@
+{{page-title "Form::TextInput Component"}}
+
+<h2 class="dummy-h2">Form::TextInput</h2>
+
+{{! ADD YOUR CONTENT BELOW }}
+{{! Please follow as much as possible what it's already done in other components. }}
+{{! You can start by copy&pasting it, if you want, and then adapt it according to your needs. }}
+{{! Once done, please remove these comments (they're created by the generator). }}
+
+<section>
+  <h3 class="dummy-h3" id="overview"><a href="#overview">§</a> Overview</h3>
+
+  {{! ADD YOUR CONTENT HERE }}
+
+</section>
+
+<section>
+  <h3 class="dummy-h3" id="component-api"><a href="#component-api">§</a> Component API</h3>
+  <p class="dummy-paragraph" id="component-api-form-text-input">
+    Here is the API for the component:
+  </p>
+  <dl class="dummy-component-props" aria-labelledby="component-api-form-text-input">
+
+    {{! ADD YOUR CONTENT HERE }}
+
+  </dl>
+</section>
+
+<section>
+  <h3 class="dummy-h3" id="how-to-use"><a href="#how-to-use" class="dummy-link-section">§</a> How to use</h3>
+
+  {{! ADD YOUR CONTENT HERE }}
+
+</section>
+
+<section>
+  <h3 class="dummy-h3" id="design-guidelines"><a href="#design-guidelines" class="dummy-link-section">§</a>Design
+    guidelines</h3>
+  {{! UNCOMMENT THIS BLOCK (once the link and/or the image are available) }}
+  {{!
+  <div class="dummy-design-guidelines">
+    <p class="dummy-paragraph">
+      <a href="[ADD THE LINK TO THE FIGMA FILE/PAGE HERE!]" target="_blank" rel="noopener noreferrer">Figma UI Kit</a>
+    </p>
+    <br />
+    <img class="dummy-figma-docs" src="/assets/images/form-text-input-design-usage.png" alt="" role="none" />
+  </div>
+  }}
+</section>
+
+<section>
+  <h3 class="dummy-h3" id="accessibility"><a href="#accessibility" class="dummy-link-section">§</a> Accessibility</h3>
+
+  {{! ADD YOUR CONTENT HERE }}
+
+</section>
+
+<section data-test-percy>
+  <h3 class="dummy-h3" id="showcase"><a href="#showcase" class="dummy-link-section">§</a> Showcase</h3>
+
+  {{! ADD YOUR CONTENT HERE }}
+
+  {{! This below is just an example of invocation, to get started }}
+  <Hds::Form::TextInput @value="This is the Hds::Form::TextInput component" />
+
+</section>
--- a/packages/components/addon/components/hds/form/checkbox/index.hbs
+++ b/packages/components/addon/components/hds/form/checkbox/index.hbs
@@ -1,0 +1,1 @@
+<input type="checkbox" className={{this.classNames}} ...attributes />
--- a/packages/components/addon/components/hds/form/checkbox/index.js
+++ b/packages/components/addon/components/hds/form/checkbox/index.js
@@ -1,0 +1,23 @@
+import Component from '@glimmer/component';
+
+export default class HdsFormCheckboxIndexComponent extends Component {
+  // UNCOMMENT THIS IF YOU NEED A CONSTRUCTOR
+  // constructor() {
+  //   super(...arguments);
+  //   // ADD YOUR ASSERTIONS HERE
+  // }
+
+  /**
+   * Get the class names to apply to the component.
+   * @method classNames
+   * @return {string} The "class" attribute to apply to the component.
+   */
+  get classNames() {
+    let classes = ['hds-form-checkbox'];
+
+    // add a class based on the @xxx argument
+    // classes.push(`hds-form-checkbox--[variant]-${this.xxx}`);
+
+    return classes.join(' ');
+  }
+}
--- a/packages/components/addon/components/hds/form/radio/index.hbs
+++ b/packages/components/addon/components/hds/form/radio/index.hbs
@@ -1,0 +1,1 @@
+<input type="radio" className={{this.classNames}} ...attributes />
--- a/packages/components/addon/components/hds/form/radio/index.js
+++ b/packages/components/addon/components/hds/form/radio/index.js
@@ -1,0 +1,23 @@
+import Component from '@glimmer/component';
+
+export default class HdsFormRadioIndexComponent extends Component {
+  // UNCOMMENT THIS IF YOU NEED A CONSTRUCTOR
+  // constructor() {
+  //   super(...arguments);
+  //   // ADD YOUR ASSERTIONS HERE
+  // }
+
+  /**
+   * Get the class names to apply to the component.
+   * @method classNames
+   * @return {string} The "class" attribute to apply to the component.
+   */
+  get classNames() {
+    let classes = ['hds-form-radio'];
+
+    // add a class based on the @xxx argument
+    // classes.push(`hds-form-radio--[variant]-${this.xxx}`);
+
+    return classes.join(' ');
+  }
+}
--- a/packages/components/addon/components/hds/form/select/index.hbs
+++ b/packages/components/addon/components/hds/form/select/index.hbs
@@ -1,0 +1,3 @@
+<select className={{this.classNames}} ...attributes>
+  {{yield}}
+</select>
--- a/packages/components/addon/components/hds/form/select/index.js
+++ b/packages/components/addon/components/hds/form/select/index.js
@@ -1,0 +1,23 @@
+import Component from '@glimmer/component';
+
+export default class HdsFormSelectIndexComponent extends Component {
+  // UNCOMMENT THIS IF YOU NEED A CONSTRUCTOR
+  // constructor() {
+  //   super(...arguments);
+  //   // ADD YOUR ASSERTIONS HERE
+  // }
+
+  /**
+   * Get the class names to apply to the component.
+   * @method classNames
+   * @return {string} The "class" attribute to apply to the component.
+   */
+  get classNames() {
+    let classes = ['hds-form-select'];
+
+    // add a class based on the @xxx argument
+    // classes.push(`hds-form-select--[variant]-${this.xxx}`);
+
+    return classes.join(' ');
+  }
+}
--- a/packages/components/addon/components/hds/form/text-input/index.hbs
+++ b/packages/components/addon/components/hds/form/text-input/index.hbs
@@ -1,0 +1,1 @@
+<input type={{this.type}} value={{@value}} className={{this.classNames}} ...attributes />
--- a/packages/components/addon/components/hds/form/text-input/index.js
+++ b/packages/components/addon/components/hds/form/text-input/index.js
@@ -1,0 +1,34 @@
+import Component from '@glimmer/component';
+
+export default class HdsFormTextInputIndexComponent extends Component {
+  // UNCOMMENT THIS IF YOU NEED A CONSTRUCTOR
+  // constructor() {
+  //   super(...arguments);
+  //   // ADD YOUR ASSERTIONS HERE
+  // }
+
+  /**
+   * Sets the type of input
+   *
+   * @param type
+   * @type {string}
+   * @default 'text'
+   */
+  get type() {
+    return this.args.type ?? 'text';
+  }
+
+  /**
+   * Get the class names to apply to the component.
+   * @method classNames
+   * @return {string} The "class" attribute to apply to the component.
+   */
+  get classNames() {
+    let classes = ['hds-form-text-input'];
+
+    // add a class based on the @xxx argument
+    // classes.push(`hds-form-text-input--[variant]-${this.xxx}`);
+
+    return classes.join(' ');
+  }
+}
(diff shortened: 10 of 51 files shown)


packages/components/addon/components/hds/form/utils/describedby.js
```diff
@@ -0,0 +1,12 @@
+import { ID_PREFIX as ERROR_ID_PREFIX } from '../error';
+import { ID_PREFIX as HELPER_TEXT_ID_PREFIX } from '../helper-text';
+
+export function describedBy(element) {
+  let describedBy = [];
+  describedBy.push(`${HELPER_TEXT_ID_PREFIX}${element.id}`);
+  describedBy.push(`${ERROR_ID_PREFIX}${element.id}`);
+  if (element.args.ariaDescribedBy) {
+    describedBy.push(element.args.ariaDescribedBy);
+  }
```
didoo, line 10: @alex-ju I am not sure why you need these lines, we don't have a `@ariaDescribedBy` argument (at the moment, at least)

alex-ju: Indeed, we don't. I saw [your comment on accepting an `@id` argument](https://github.com/hashicorp/design-system/pull/285/files#diff-89af5d947bcadceab7186cf644089e49c44401512d8ede02dac0a488d939e74bR44) and based on the spike on `cloud-ui` I'd say we should accommodate that, and also an `@ariaDescribedBy`. We can discuss it in our sync.

didoo: As discussed, makes totally sense 👍

packages/components/addon/components/hds/form/helper-text/index.hbs
```diff
@@ -0,0 +1,3 @@
+<div class={{this.classNames}} id={{this.id}} ...attributes>
```
MelSumner, line 1: [question] Is there a reason this doesn't have a pre-defined class name?

packages/components/addon/components/hds/form/checkbox/field.js
```diff
@@ -0,0 +1,41 @@
+import Component from '@glimmer/component';
+
+import { guid } from '../utils/guid';
```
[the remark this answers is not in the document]
didoo: there is logic behind it (see `utils/guid`) - the one provided by ember is used inside the method.
we intend to rename it though (to something like `controldId` or `getControlId`) once #353 is done

packages/components/addon/components/hds/form/checkbox/group.hbs
```diff
@@ -0,0 +1,10 @@
+<fieldset class={{this.classNames}}>
+  {{yield (hash Legend=(component "hds/form/legend" _contextualClass="hds-form-group__legend"))}}
+  {{yield (hash HelperText=(component "hds/form/helper-text" _contextualClass="hds-form-group__helper-text"))}}
```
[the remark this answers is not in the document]
didoo: We intend to maintain the "right" order in the DOM.

packages/components/addon/components/hds/form/helper-text/index.hbs
didoo: can you expand on this? I am not sure of the question

MelSumner: Let me try this another way: Is there a specific reason that the helper text has `this.classNames` but the error message has `hds-form-error__message`?

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
```diff
@@ -0,0 +1,355 @@
...
+    There may be cases in which the helper text needs to contain more than just text. In this case it's possible to pass
+    structured content to it (it's just yielded in output):
+  </p>
+  {{! prettier-ignore-start }}
+  <CodeBlock
+    @language="markup"
+    @code='
+      <Hds::Form::HelperText @controlId="control-ID">
+        Some text with a
+        <Hds::Link::Inline @route="components.link.inline">Hds::Link::Inline</Hds::Link::Inline>,
```
MelSumner, line 176: did you get a chance to review https://codepen.io/melsumner/pen/NWyLMOQ? We may need to rethink the docs here.

packages/components/addon/components/hds/form/checkbox/group.hbs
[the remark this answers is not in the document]
didoo: (btw, this "order of elements in the DOM" decision may be revised later)

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
didoo: @MelSumner this refers to the `helper text`, which is a normal element, not a `<label>` (I have updated the documentation removing any reference to link in the label, and added a warning)
screenshot_1544

packages/components/addon/components/hds/form/text-input/field.hbs
```diff
@@ -0,0 +1,18 @@
...
+    )
+  }}
+  <Hds::Form::TextInput::Base
+    @type={{@type}}
+    @value={{@value}}
+    @isInvalid={{@isInvalid}}
+    @width={{@width}}
+    ...attributes
+    id={{this.id}}
+    aria-describedby={{this.ariaDescribedBy}}
```
MelSumner, line 15: [issue] the way this is currently implemented, an error will be thrown because it always include help text and error message text, even if there aren't any (therefore those elements wouldn't exist).

I wonder if you can try something like what we did in the [button element for aria-label](https://github.com/hashicorp/design-system/blob/6a7f752ce59ca284a16e6c48ff0a55f9c425bc48/packages/components/addon/components/hds/button/index.hbs#L12) or something similar? 

`aria-describedby="{{if this.hasHelpText this.helpTextId null}} {{if this.hasErrorText this.errorTextId null}}"`

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
```diff
@@ -0,0 +1,462 @@
...
+<section data-test-percy>
+  <h3 class="dummy-h3" id="showcase"><a href="#showcase" class="dummy-link-section">§</a> Showcase</h3>
+
+  <h4 class="dummy-h4">"Base" control</h4>
+  <h5 class="dummy-h6">Interaction status</h5>
+  <div class="dummy-form-text-input-base-sample">
+    <div>
+      <span class="dummy-text-small">Default</span>
+      <br />
+      <Hds::Form::TextInput::Base />
```
MelSumner, line 275:
```suggestion
      <Hds::Form::TextInput::Base aria-label="default" />
```
We have been consistent about making our own samples conformant. Even if we don't want to show a label here, we can still add an aria-label attribute.

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
```diff
@@ -0,0 +1,462 @@
...
+    </div>
+    <div>
+      <span class="dummy-text-small">With placeholder</span>
+      <br />
+      <Hds::Form::TextInput::Base placeholder="Lorem ipsum dolor" />
```
MelSumner, line 280:
```suggestion
      <Hds::Form::TextInput::Base placeholder="Lorem ipsum dolor" aria-label="with placeholder" />
```

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
```diff
@@ -0,0 +1,462 @@
...
+      <Hds::Form::TextInput::Base @value="Lorem ipsum dolor" />
+    </div>
+  </div>
+  <h5 class="dummy-h5">Types (native)</h5>
+  <div class="dummy-form-text-input-types-grid">
+    {{#each @model.TYPES as |type|}}
+      <div>
+        <span class="dummy-text-small">{{capitalize type}}:</span>
+        <br />
+        <Hds::Form::TextInput::Base @type={{type}} @value={{type}} />
```
MelSumner, line 294:
```suggestion
        <Hds::Form::TextInput::Base @type={{type}} @value={{type}}  aria-label="{{type}}" />
```

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
```diff
@@ -0,0 +1,462 @@
...
+      {{#each variants as |variant|}}
+        {{#each @model.STATES as |state|}}
+          <div>
+            <span class="dummy-text-small">{{capitalize variant}} / {{capitalize state}}:</span>
+            <br />
+            <div class="dummy-form-text-input-sublist" mock-state-value={{state}} mock-state-selector="input">
+              <Hds::Form::TextInput::Base
+                disabled={{if (eq variant "disabled") "disabled"}}
+                readonly={{if (eq variant "readonly") "readonly"}}
+                @isInvalid={{if (eq variant "invalid") true}}
```
MelSumner, line 310:
```suggestion
                @isInvalid={{if (eq variant "invalid") true}}
                aria-label="{{variant}} {{state}}"
```

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
```diff
@@ -0,0 +1,462 @@
...
+              />
+              <Hds::Form::TextInput::Base
+                placeholder="Placeholder"
+                disabled={{if (eq variant "disabled") "disabled"}}
+                readonly={{if (eq variant "readonly") "readonly"}}
+                @isInvalid={{if (eq variant "invalid") true}}
```
MelSumner, line 316:
```suggestion
                @isInvalid={{if (eq variant "invalid") true}}
                aria-label="{{variant}} {{state}}"
```

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
```diff
@@ -0,0 +1,462 @@
...
+              />
+              <Hds::Form::TextInput::Base
+                @value="Lorem ipsum dolor"
+                disabled={{if (eq variant "disabled") "disabled"}}
+                readonly={{if (eq variant "readonly") "readonly"}}
+                @isInvalid={{if (eq variant "invalid") true}}
```
MelSumner, line 322:
```suggestion
                @isInvalid={{if (eq variant "invalid") true}}
                aria-label="{{variant}} {{state}}"
```

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
```diff
@@ -0,0 +1,462 @@
...
+              />
+              <Hds::Form::TextInput::Base
+                @type="date"
+                @value="Lorem ipsum dolor"
+                disabled={{if (eq variant "disabled") "disabled"}}
+                readonly={{if (eq variant "readonly") "readonly"}}
+                @isInvalid={{if (eq variant "invalid") true}}
```
MelSumner, line 329:
```suggestion
                @isInvalid={{if (eq variant "invalid") true}}
                aria-label="{{variant}} {{state}}"
```

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
```diff
@@ -0,0 +1,462 @@
...
+          <br />
+          <div class="dummy-form-text-input-sublist">
+            <div class="dummy-form-text-input-containers__{{display}}">
+              <Hds::Form::TextInput::Base @value="Default width" />
+            </div>
+            <div class="dummy-form-text-input-containers__{{display}}">
+              <Hds::Form::TextInput::Base @value="Custom width" @width="248px" />
+            </div>
+            <div class="dummy-form-text-input-containers__{{display}}">
+              <Hds::Form::TextInput::Base @type="date" />
```
MelSumner, line 371:
```suggestion
              <Hds::Form::TextInput::Base @type="date" aria-label="{{display}}" />
```

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
```diff
@@ -0,0 +1,462 @@
...
+            </div>
+            <div class="dummy-form-text-input-containers__{{display}}">
+              <Hds::Form::TextInput::Base @type="time" />
```
MelSumner, line 374:
```suggestion
              <Hds::Form::TextInput::Base @type="time" aria-label="{{display}}" />
```

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
MelSumner: I suppose this will fall into the category of "technically permissible but doesn't work" - anything that is referred to in `aria-describedby` has only the text read out, so the user won't know that it's actually a link.

Let's brainstorm with the team and see if there's an elegant way to solve this problem.

packages/components/addon/components/hds/form/checkbox/base.hbs
```diff
@@ -0,0 +1 @@
+<input type="checkbox" class={{this.classNames}} ...attributes />
```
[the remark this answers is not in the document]
MelSumner: It's fine for the user to have to provide and define it, but it should still be here. Think about the other components where the user has to define a value for something. We still say which things need to be defined and give helpful errors if they are not there (like the `text` in the `Hds::Button` component).

packages/components/app/styles/components/form/text-input.scss
```diff
@@ -0,0 +1,111 @@
...
+.hds-form-text-input {
+  border: 1px solid var(--token-color-palette-neutral-400);
+  border-radius: 5px;
+  box-shadow: var(--hds-elevation-inset-box-shadow);
+  font-family: var(--token-typography-body-200-font-family);
+  font-size: var(--token-typography-body-200-font-size);
+  line-height: var(--token-typography-body-200-line-height);
+  padding: 7px;
+  width: 100%;
+  max-width: 100%;
```
MelSumner, line 18: Do you think we need a min-width here? Noticed this when I was moving my browser windows around: 
CleanShot 2022-06-10 at 11 11 42

packages/components/addon/components/hds/form/fieldset/index.js
```diff
@@ -0,0 +1,48 @@
...
+
+import { guid } from '../utils/guid';
+import { describedBy } from '../utils/describedby';
+
+export default class HdsFormFieldsetIndexComponent extends Component {
+  /**
+   * Sets the layout of the group
+   *
+   * @param layout
+   * @type {enum}
```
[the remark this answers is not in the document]
didoo: @MelSumner yes, something @alex-ju is going to do in another PR.

packages/components/addon/components/hds/form/text-input/field.hbs
didoo: @alex-ju opened a Asana ticket specifically for this issue: https://app.asana.com/0/1202008719910219/1202443172675747

packages/components/addon/components/hds/form/checkbox/field.js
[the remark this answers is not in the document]
didoo: Refactored in https://github.com/hashicorp/design-system/pull/375

packages/components/addon/components/hds/form/checkbox/base.hbs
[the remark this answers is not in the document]
didoo: Discussed with @MelSumner, for now let's keep it like this and we may review the decision later.

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
didoo: @MelSumner will
- update the documentation for the `HelperTex` in the `base-elements` page, with a note about how to make the content more accessible if there is a link inside the helper text
- create a helper component `ScreenReaderOnly` (name tbd) that can wrap a generic content/text with the existing `.sr-only` class, to make the content available only to screen readers.

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
didoo: As discussed during the code review today, I am not convinced this is the way to go for the "showcase" components. I totally understand the idea to have accessible documentation pages, but the _show_case is intended mainly to _show_ the components, with a particular focus on the developers implementing the components (and Percy) plus as a bonus side-effect, show to the consumers what are the variants (and their combinations) of a component. 

The issue of devs copying&pasting the code from the showcase is minimal, because we don't expect them to copy the HTML, and (unless I'm missing something) is not possible to copy the code from the Ember inspector.

So the last problems remains having pages which will not pass an accessibility test. Here I can propose some alternatives:
- we hide the whole showcase from the accessibility tree (a bit drastic!)
- we hide the parent container/wrapper of the "base" components from the accessibility tree (more specific)
- we accept that not all the "base" elements in the showcase will meet all the accessibility criteria, when considered in isolation
  - this is not something unusual: consider this MDN page: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input - as you can see the top example has a label associated with the `<input>` but the ones in the list of types they don't and they also don't have a `aria-label` associated, and I am not sure I would expect them to have it frankly (their purpose meant to showcase the component, not as real usable/interactive content).

PR #289, opened 2022-05-13
"Alert" and "Toast" components - Convert "Title" and "Description" to be yielded contextual components

didoo:
### :pushpin: Summary

While doing a demo of the `Alert` and `Toast` components to the HDS Engineering Ambassadors, we have found a few use cases where the content of the `description` field was more structured than simple text (it could contain conditional logic, HTML elements, Ember components).

See for example here:
- https://github.com/hashicorp/cloud-ui/blob/c3161a6a5ff5b9cd0007601a4cad68555f4300f2/engines/monitoring/addon/components/audit-logs/archives-unavailable-banner/index.hbs#L6-L12
- https://github.com/hashicorp/atlas/blob/c77e9c1ccd481f199811b80a53f20a8a7f2162c5/frontend/atlas/app/components/subscription-plan-viewer.hbs#L4-L30
- https://github.com/hashicorp/atlas/blob/f35acb239b90211ab1c4905162b814219e6028c7/frontend/atlas/app/components/vcs-repo-select.hbs#L3-L28
- https://github.com/hashicorp/atlas/blob/ccd7d558038ba143c3237f9cfebc6db2e94b0aec/frontend/atlas/app/components/v2/runs-blank-slate.hbs#L5-L31

We have agreed that the best option in this case was to introduce an alternative way to pass the content of the `description` using an ad-hoc contextual component.

While working on the PR, we have realized (see comments below) that at this point it was better/more consistent to have also the `title` as contextual component, and remove them as arguments.

In this PR I have:
- added `Title` and `Description` as yielded children to the `Alert` component
  - removed `@title` and `@description` as arguments
- updated documentation for the `Alert` component
- improved styling of the HTML basic tags inside the `description` field (to be discussed with @cveigt before release)
- updated the `Toast` component to yield also the new `Title` and `Description` contextual components
- updated documentation for the `Toast` component
- updated integration tests for the `Alert` and `Toast` components

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/app/styles/components/alert.scss
+++ b/packages/components/app/styles/components/alert.scss
@@ -74,13 +74,12 @@
     border-radius: 5px;
     display: inline;
     font-family: var(--token-typography-code-100-font-family);
-    // font-size: var(--token-typography-code-100-font-size);
-    font-size: 0.9em; // TODO discuss with CVeight
+    font-size: 0.9em; // as discussed with designers, we reduce the size for optical/visual balance
     line-height: 1em;
     padding: 1px 5px;
   }
 
-  // Notice: in the future this may become a "Link::Inline" component (for now we declare the exact same styles directly here)
+  // Notice: in the future this may become a "Link::Inline" component (for now we declare directly the same styles here)
   a {
     color: var(--token-color-foreground-action);
 
--- a/packages/components/app/styles/components/alert.scss
+++ b/packages/components/app/styles/components/alert.scss
@@ -70,7 +70,7 @@
 
   code, pre {
     background-color: var(--token-color-surface-primary);
-    border: 1px solid var(--token-color-palette-neutral-300);
+    border: 1px solid var(--token-color-palette-neutral-200);
     border-radius: 5px;
     display: inline;
     font-family: var(--token-typography-code-100-font-family);
--- a/packages/components/app/styles/components/alert.scss
+++ b/packages/components/app/styles/components/alert.scss
@@ -69,14 +69,28 @@
   }
 
   code, pre {
-    font-family: var(--token-typography-font-stack-code);
+    background-color: var(--token-color-surface-primary);
+    border: 1px solid var(--token-color-palette-neutral-300);
+    border-radius: 5px;
+    display: inline;
+    font-family: var(--token-typography-code-100-font-family);
+    // font-size: var(--token-typography-code-100-font-size);
+    font-size: 0.9em; // TODO discuss with CVeight
+    line-height: 1em;
+    padding: 1px 5px;
   }
 
-  // Notice: in the future this may become a "Link::Inline" component (for now we declare the styles directly here)
+  // Notice: in the future this may become a "Link::Inline" component (for now we declare the exact same styles directly here)
   a {
     color: var(--token-color-foreground-action);
-    // At the moment the "focus" state is not well defined in design (the one that is in Figma does not work) so we just apply a simple color to the default outline
-    outline-color: var(--token-color-focus-action-external);
+
+    &:focus,
+    &.is-focus,
+    &:focus-visible {
+      outline: 2px solid var(--token-color-focus-action-internal);
+      outline-offset: 1px;
+      text-decoration: none;
+    }
 
     &:hover {
       color: var(--token-color-foreground-action-hover);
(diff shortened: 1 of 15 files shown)


Dhaulagiri:
This may be a separate conversation, but it seems a little awkward/potentially confusing to offer two different ways of setting the same property (`description`) with no way to enforce one choice or the other. If we are offering the ability to yield a value does it make sense to also keep the `@description` property?

didoo:
> This may be a separate conversation, but it seems a little awkward/potentially confusing to offer two different ways of setting the same property (`description`) with no way to enforce one choice or the other. If we are offering the ability to yield a value does it make sense to also keep the `@description` property?

Personally, I totally agree with you. To the point that yesterday we considered also to have a yielded/contextual component for the `Title` too, but then we found no cases where the title was not simple text. Let's see what others say about it (please @MelSumner @alex-ju have your say on this) and then making the change would be fast in case.

MelSumner:
what if we: 
- used `descriptionText` in the API for the string
- used `descriptionYield` for the yield

the more explicit naming should reduce the confusion. WDYT?

Dhaulagiri:
> To the point that yesterday we considered also to have a yielded/contextual component for the Title too, but then we found no cases where the title was not simple text.

_nod_ that makes sense

> the more explicit naming should reduce the confusion

What I was referring to wasn't so much a matter of _naming_ as a matter of offering two ways to do the same thing being redundant.

MelSumner:
> @Dhaulagiri @MelSumner @alex-ju FYI I have updated the API of the `Alert` and `Toast` components to have both `Title` and `Description` as contextual components, and removed them as arguments.

Can you say more about this? Title makes sense as an arg and the feedback we got from the ambassadors was that title was fine to leave that way.

MelSumner:
> > the more explicit naming should reduce the confusion
> 
> What I was referring to wasn't so much a matter of _naming_ as a matter of offering two ways to do the same thing being redundant.

It might be redundant, but I like the option for both, it's nice flexibility. We could deprecate one in the future if we found that it was simply not being used.

didoo:
> Can you say more about this? Title makes sense as an arg and the feedback we got from the ambassadors was that title was fine to leave that way.

> It might be redundant, but I like the option for both, it's nice flexibility. We could deprecate one in the future if we found that it was simply not being used.

When I looked at the implementation in code, it started to be confusing. Where use one or another? One could support only string, one could support strings with encoded HTML, one could support directly structured content/HTML. I followed the "don't make me think" here and made it all the same, reducing complexity (also in the documentation).

MelSumner:
How about an experiment to test an hypothesis?

1. add the ability to yield description
2. let me take a turn at making the documentation and examples really clear
3. review the solution and the docs with our users

My hypothesis is that they will not find it confusing; but if they do, then we change it to be like this and yield both title and description.

WDYT?

didoo:
> How about an experiment to test an hypothesis?
> 
> 1. add the ability to yield description
> 2. let me take a turn at making the documentation and examples really clear
> 3. review the solution and the docs with our users
> 
> My hypothesis is that they will not find it confusing; but if they do, then we change it to be like this and yield both title and description.
> 
> WDYT?

@MelSumner I prefer to use our time in a more efficient way. We have more fish to fry, let's move on.

/cc @Dhaulagiri

MelSumner (approved):
I don't think changing both of these is the right move, but I agree @didoo that we need to keep moving forward and that this component is already overdue as it is. Perhaps in the future we could scope PRs to agreed-upon changes and keep additional ones separate.

packages/components/app/styles/components/alert.scss
```diff
@@ -69,14 +69,27 @@
   }
 
   code, pre {
-    font-family: var(--token-typography-font-stack-code);
+    background-color: var(--token-color-surface-primary);
+    border: 1px solid var(--token-color-palette-neutral-300);
```
cveigt, line 73:
```suggestion
    border: 1px solid var(--token-color-palette-neutral-200);
```

cveigt (approved):
LGTM! 👍

alex-ju (approved):

PR #301, opened 2022-05-19
Convey disclosure state to assistive tech for dropdown and truncated breadcrumbs

alex-ju:
### :pushpin: Summary

Convey the disclosure state to assistive tech for dropdown and truncated breadcrumbs.

**What**: I propose that we use `aria-expanded` on the toggle elements to convey to assistive tech whether the disclosed container is expanded or collapsed.

**Why**: When sighted users interact with the toggle element they see the expanded container and have a visual clue in the chevron icon changing its state. With this change we try to express the same clue to screen reader users so that when they reach a the toggle element, it will be read out as `collapsed` or `expanded`, depending on its state.

Exemplified on the first dropdown example in VoiceOver

Collapsed (default)
Screenshot 2022-05-19 at 12 41 40

Expanded
Screenshot 2022-05-19 at 12 44 14

<table>
<tr><td>

Collapsed (default)/expanded
Screenshot 2022-05-19 at 12 38 27

</td><td>
Collapsed (default)
Screenshot 2022-05-19 at 12 38 37

Expanded
Screenshot 2022-05-19 at 12 38 41

</td></tr>
</table>

***

### 👀 How to review

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/breadcrumb/truncation.hbs
+++ b/packages/components/addon/components/hds/breadcrumb/truncation.hbs
@@ -7,6 +7,7 @@
           {{if @hover 'is-hover'}}
           {{if @active 'is-active'}}
           {{if @focus 'is-focus'}}"
+        aria-expanded={{if t.isActive "true" "false"}}
         {{on "click" t.onClickToggle}}
       >
         <FlightIcon @name="more-horizontal" @size="16" @isInlineBlock={{false}} />
--- a/packages/components/addon/components/hds/dropdown/toggle/button.hbs
+++ b/packages/components/addon/components/hds/dropdown/toggle/button.hbs
@@ -19,5 +19,6 @@
   @iconPosition="trailing"
   @color={{this.color}}
   ...attributes
+  aria-expanded={{this.expanded}}
   {{on "click" this.onClick}}
 />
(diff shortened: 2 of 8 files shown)


didoo (approved):
Code is OK for me. 

For what concerns accessibility, there's more qualified people to comment, so I'll leave it to them :)

MelSumner:
From the spec: 

> If the element with the [aria-expanded](https://www.w3.org/TR/wai-aria/#aria-expanded) attribute controls the expansion of another grouping container that is not 'owned by' the element, the author SHOULD reference the container by using the [aria-controls](https://www.w3.org/TR/wai-aria/#aria-controls) attribute.

So in this case, we can't add an `aria-controls` because the item isn't in the DOM until the click event happens (idk if this is unique to Ember). 

This is the biggest reason why we went with focus trapping in the original implementation.

alex-ju:
> I usually prefer the "write to the spec" approach, but if you don't think it will be confusing for users then I'm fine with us shipping it and adding this to the list of things to check on as time goes forward. 👍

Totally with you on "write to the spec" – I'll have to add this to the not-to-the-spec jar that haunts me so often

PR #304, opened 2022-05-20
automatically update percy deps

Dhaulagiri:
### :pushpin: Summary

*  ~~Move percy deps to the workspace root as they are shared across multiple packages~~
* Adds a weekly dependabot check for these so they stay updated more frequently

***

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ ] +1 Percy if applicable
- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #305, opened 2022-05-20
remove design system deps from dependabot

Dhaulagiri:
### :pushpin: Summary

with the monorepo + changesets setup this is no longer needed (hasn't run at all in quite some time). the package updates happen as part of changeset releases.

Leaving the file here as https://github.com/hashicorp/design-system/pull/304 will add a new check

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -2,18 +2,3 @@
 
 version: 2
 updates:
-  - package-ecosystem: "npm" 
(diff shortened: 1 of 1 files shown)


alex-ju (approved):
makes sense

Dhaulagiri:
merged without waiting for the checks since this wouldn't impact any of our tests anyhow

PR #217, opened 2022-04-14
Button/Link/CTA - Updates for the interactive elements

didoo:
### :pushpin: Summary

This is the "feature branch" used to collect in one single place all the changes proposed to the Button/Link/CTA components. For more details about the reasons and rationale why these changes are proposed, please refer to the single PRs and to the RFC document linked below.

The following PRs need to be reviewed and merged in this order:

1. #216 - ✅
2. #220 - ✅
3. #221 - ✅
4. #227 - ✅
5. #231  - ✅

Once they're all merged (if approved) in this branch, we'll do a final review of the code to make sure we're not missing or forgetting something, and once this PR is approved it can be merged in `main` and the proposed changes officially released.

### 🚧 TODOs

Once all the branches are merged:
- [x] updated screenshots for documentation for designers (from Figma)
  - decided we'll add in a follow-up PR
- [x] update the `Links, Buttons, CTAs...oh my!` block (probably following a similar update in the documentation in Figma)
  - hiding for now, we will re-add it later once it's been updated also in Figma
- [x] decide what kind of release we want to do (minor = `0.12.0` or major = `1.0.0`) - _notice: this will be a major breaking change from the code perspective_
  -  @Dhaulagiri decided to go with `minor` release
  - [x] add the changeset file accordingly

### :link: Links

- **RFC**: [DS-033 - Button/Link/CTA Rethink](https://docs.google.com/document/d/1Tc99UR3UJo_avncWD1V8uGgDr72W2TjLbxAvPPaNRj8/edit)

***

### 👀 How to review

👉 Review by files changed

Updated documentation pages:
- [Interactive](https://hds-components-git-spike-button-link-cta-main-hashicorp.vercel.app/utilities/interactive)
- [Button](https://hds-components-git-spike-button-link-cta-main-hashicorp.vercel.app/components/button)
- [Link::Inline](https://hds-components-git-spike-button-link-cta-main-hashicorp.vercel.app/components/link/inline)
- [Link::Standalone](https://hds-components-git-spike-button-link-cta-main-hashicorp.vercel.app/components/link/standalone)

Reviewer's checklist:

- [ ] +1 Percy if applicable
- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202149050872176

PR #314, opened 2022-05-25
Generalize mocking of states in dummy pages

didoo:
### :pushpin: Summary

This task has been on my todo list for a long time, and while implementing the states for the _form controls_ elements I have realized it was time to bite the bullet and implement it properly (without the proposed changes, would have required a lot of hacks).

What I am proposing here is to replace the way we declare and handle the visual states for the showcase of components, from using _Ember arguments_ (`@state`) passed down in the actual components code (HBS + SCSS), to custom HTML attributes (`mock-state`) applied only in the dummy webpages (and referenced only in the CSS).

**What is the problem I am trying to solve?**

The fact is that with the previous approach we were mixing two concerns, the code for the actual components, and the code used to showcase them in the dummy website, and we were exposing both in production code. 

**Why this solution is better?**

With the proposed change, we get rid of this entanglement on the HBS template side (and we may be able to get rid of the same thing on the CSS side, in the future, using a post-css processor that removed the `mock-[state]` selectors from the generated CSS for production code).

In this PR I have
- added a controller for the “components” pages that takes care of handling `mock-state` attributes
  - the logic is this: we have two possible custom HTML attributes (not prefixed with `data` intentionally, even if it's not valid HTML for our purpose is OK)
    - `mock-state-value` is the value of the _state_ that we want to use, to apply a class `mock-[state]` to an element, so that a custom CSS class can be used to emulate a particular state in a component (typically pseudo states like `:hover` or `:focus`) in a showcase
    - `mock-state-selector` is the DOM selectors we want to use (relative to the element to which is applied) to select the _target_ element on which dynamically apply the mock class at runtime; if no selector is defined, we simply use the element with the `mock-state-value` attribute itself
  - at the moment I don't see this functionality used in other type of pages, so it's OK to have it only here (if other pages/sections will need the same functionality, we will abstract/duplicate it accordingly)
- updated `Dropdown` + `Button` dummy pages (and related components and CSS files) to use the new `mock-status` approach

Notice: since the other pages using the states showcase are `Link::CTA` and `Link::Standalone, we will apply the changes to the remaining dummy pages once this branch is merged to `main` and  `main` merged in #217 (we will apply the changes there).

### 🖤 Credits

Thanks to @jgwhite 🙏 for helping me solve the problem with `extends Component` + `constructor()` + `scheduleOnce('afterRender'...`). Essentially _this is the ultimate gotcha with controllers — they’re singletons. So that constructor will only be run once for the lifetime of the application._. TIL 🤷‍♂️

(the code fix is in https://github.com/hashicorp/design-system/pull/314/commits/ddfc49a004e53c7c216be330fb2170bf4ced158d)

***

### 👀 How to review

👉 Review by files changed

_Notice: I'll add comments inline in the `files changed` section, to explain what I did and why._

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/app/styles/mixins/_focus-ring.scss
+++ b/packages/components/app/styles/mixins/_focus-ring.scss
@@ -8,7 +8,8 @@
 
   // default focus for browsers that still rely on ":focus"
   &:focus,
-  &.is-focus {
+  &.is-focus,
+  &.mock-focus {
     box-shadow: var(--token-focus-ring-#{$color}-box-shadow);
   }
   // undo the previous declaration for browsers that support ":focus-visible" but wouldn't normally show default focus styles
@@ -21,7 +22,8 @@
   }
   // remove the focus ring on "active + focused" state (by design)
   &:focus:active,
-  &.is-focus.is-active {
+  &.is-focus.is-active,
+  &.mock-focus.is-active {
     box-shadow: none;
   }
 }
@@ -45,7 +47,8 @@
 
   // default focus for browsers that still rely on ":focus"
   &:focus,
-  &.is-focus {
+  &.is-focus,
+  &.mock-focus {
     &::before {
       box-shadow: var(--token-focus-ring-#{$color}-box-shadow);
     }
@@ -64,7 +67,8 @@
   }
   // remove the focus ring on "active + focused" state (by design)
   &:focus:active,
-  &.is-focus.is-active {
+  &.is-focus.is-active,
+  &.mock-focus.is-active {
     &::before {
       box-shadow: none;
     }
--- a/packages/components/tests/dummy/app/controllers/components.js
+++ b/packages/components/tests/dummy/app/controllers/components.js
@@ -1,0 +1,24 @@
+import Controller from '@ember/controller';
+import { scheduleOnce } from '@ember/runloop';
+
+function replaceDummyStates() {
+  document.querySelectorAll('[mock-state-value]').forEach(function (element) {
+    let target;
+    if (element.attributes['mock-state-selector']) {
+      target = element.querySelector(
+        element.attributes['mock-state-selector'].value
+      );
+    } else {
+      target = element;
+    }
+    const state = element.attributes['mock-state-value'].value;
+    target.classList.add(`mock-${state}`);
+  });
+}
+
+export default class ComponentsController extends Controller {
+  constructor() {
+    super(...arguments);
+    scheduleOnce('afterRender', this, replaceDummyStates);
+  }
+}
--- a/packages/components/tests/dummy/app/controllers/components.js
+++ b/packages/components/tests/dummy/app/controllers/components.js
@@ -1,4 +1,5 @@
 import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
 import { scheduleOnce } from '@ember/runloop';
 
 function replaceMockStates() {
@@ -19,8 +20,18 @@ function replaceMockStates() {
   });
 }
 export default class ComponentsController extends Controller {
+  @service router;
+
   constructor() {
     super(...arguments);
+    this.router.on('routeDidChange', this, 'routeDidChange');
+  }
+
+  routeDidChange() {
     scheduleOnce('afterRender', this, replaceMockStates);
+  }
+
+  willDestroy() {
+    this.router.off('routeDidChange', this, 'routeDidChange');
   }
 }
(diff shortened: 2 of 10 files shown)


packages/components/app/styles/mixins/_focus-ring.scss
```diff
@@ -8,7 +8,8 @@
 
   // default focus for browsers that still rely on ":focus"
   &:focus,
-  &.is-focus {
+  &.is-focus,
```
didoo, line 11: once the changes done across all the pages, I'll remove the `is-[state]` selectors from here, for now we need both to avoid breaking the percy tests

packages/components/tests/dummy/app/controllers/components.js
```diff
@@ -0,0 +1,24 @@
...
+    if (element.attributes['mock-state-selector']) {
+      target = element.querySelector(
+        element.attributes['mock-state-selector'].value
+      );
+    } else {
+      target = element;
+    }
+    const state = element.attributes['mock-state-value'].value;
+    target.classList.add(`mock-${state}`);
+  });
```
didoo, line 16: The logic here is this: we have two possible custom HTML attributes (not prefixed with `data` intentionally, even if it's not valid HTML for our purpose is OK)
- `mock-state-value` is the value of the _state_ that we want to use, to apply a class `mock-[state]` to an element, so that a custom CSS class can be used to emulate a particular state in a component (typically pseudo states like `:hover` or `:focus`) in a showcase
- `mock-state-selector` is the DOM selectors we want to use (relative to the element to which is applied) to select the _target_ element on which dynamically apply the mock class at runtime; if no selector is defined, we simply use the element with the `mock-state-value` attribute itself

packages/components/tests/dummy/app/templates/components/button.hbs
```diff
@@ -344,7 +344,13 @@
           <div>
             <span class="dummy-text-small">{{capitalize size}}/{{state}}:</span>
             <br />
-            <Hds::Button @icon="plus" @text={{capitalize state}} @size={{size}} @color={{color}} class="is-{{state}}" />
+            <Hds::Button
+              @icon="plus"
+              @text={{capitalize state}}
+              @size={{size}}
+              @color={{color}}
+              mock-state-value={{state}}
```
didoo, line 352: here you can see how one of these mock states is declared on a component, but **only** in the dummy documentation page

packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -1083,12 +1083,17 @@
     {{! template-lint-disable no-inline-styles }}
     <ul class="hds-dropdown-list" style="width: 250px">
       {{#each @model.ITEM_STATES as |state|}}
-        <Hds::Dropdown::ListItem::CopyItem @text="{{state}}: 91ee1e8ef65b337f0e70d793f456c71d" @state={{state}} />
+        <Hds::Dropdown::ListItem::CopyItem
+          @text="{{state}}: 91ee1e8ef65b337f0e70d793f456c71d"
+          mock-state-value={{state}}
+          mock-state-selector="button"
```
didoo, line 1089: here instead you can see a case where the _target_ element is **inside** the component, but using the `mock-state-selector` it's possible to apply the mock state class directly to it, without the need to pass down arguments across components and sub-components

didoo:
@alex-ju @MelSumner @Dhaulagiri now the PR is ready for review.

alex-ju (approved):
I very much prefer using attributes to mock states, especially as I got a bit confused at my first interaction with the repo when looking at those classes. Code looks good to me (as much as I understand the lifecycle of an Ember instance at the moment).

MelSumner (approved):
- I am surprised that Ember is not throwing several errors on this, especially re: having a controller named "components". 

- I am curious why you didn't use a `data-` attribute; seems strange to purposefully craft invalid HTML when valid HTML is available for exactly this sort of thing. 🤔

- The approach itself [seems similar](https://github.com/ember-a11y/ember-a11y-refocus/blob/e254e865424cdad6c62302bd0b5f504cd43da8b0/addon/components/navigation-narrator.js#L36) to what we do with `ember-a11y-refocus`

didoo:
> * I am surprised that Ember is not throwing several errors on this, especially re: having a controller named "components".

Me too :)

> * I am curious why you didn't use a `data-` attribute; seems strange to purposefully craft invalid HTML when valid HTML is available for exactly this sort of thing. 🤔

Shorter, easier to read but more importantly to spot in code or in devtools (there may be other legit `data-` attributes).

> * The approach itself [seems similar](https://github.com/ember-a11y/ember-a11y-refocus/blob/e254e865424cdad6c62302bd0b5f504cd43da8b0/addon/components/navigation-narrator.js#L36) to what we do with `ember-a11y-refocus`
👍

PR #321, opened 2022-05-26
Mock status - Complete migration

didoo:
### :pushpin: Summary

This is a follow-up of #314, where we updated only some of the dummy documentation pages (and components).

In this PR I have:
- updated remaining pages and components to use the new `mock-status` approach for mocking visual states

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/breadcrumb/item.hbs
+++ b/packages/components/addon/components/hds/breadcrumb/item.hbs
@@ -10,7 +10,7 @@
     </div>
   {{else}}
     <LinkTo
-      class="hds-breadcrumb__link {{if @hover 'is-hover'}} {{if @active 'is-active'}} {{if @focus 'is-focus'}}"
+      class="hds-breadcrumb__link"
       @models={{hds-link-to-models @model @models}}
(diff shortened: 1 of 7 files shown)


Dhaulagiri (approved):
Seems to work ok in the review app

Non-blocking ❓ 

I feel like I'm often the changeset 👮‍♂️, but I think you could argue we should include a changeset for these since they do touch public code and _could_ introduce bugs customers might be confused by without a changelog entry 🤷. What do you think?

didoo:
> Seems to work ok in the review app
> 
> Non-blocking ❓
> 
> I feel like I'm often the changeset 👮‍♂️, but I think you could argue we should include a changeset for these since they do touch public code and _could_ introduce bugs customers might be confused by without a changelog entry 🤷. What do you think?

I swear, this time I thought of it (the changeset) and I concluded that it's not necessary (that's why I intentionally removed it from the "Reviewer's checklist:").

Dhaulagiri:
👍  thanks for letting me work through this out loud

Issue #290, opened 2022-05-13
Bug report: "navigator.clipboard.readText is not a function" in Firefox


didoo:
## IS: Current behavior
When clicking on the `CopyItem` in Firefox, a javascript error occurs:
<img width="1792" alt="screenshot_1445" src="https://user-images.githubusercontent.com/686239/168259747-57b8dfaf-3b62-4840-9449-4a8f583967b0.png">

## SHOULD BE: Expected behavior
It should copy the value

## Steps to reproduce
1. Visit https://design-system-components-hashicorp.vercel.app/components/dropdown in Firefox
2. Open the "console" panel in the dev tools
3. Go to the example "Integrate with Terraform Cloud"
4. Click on "Endpoint URL"
5. You will see the error: `navigator.clipboard.readText is not a function`

## Possible solution
A [quick search on StackOverflow](https://stackoverflow.com/questions/67440036/navigator-clipboard-readtext-is-not-working-in-js) seems to indicate that [Firefox doesn't allow the reading of the clipboard](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText#browser_compatibility), so maybe we should add a `guard` that checks if the function exists, otherwise assume the `write` was successful.

MelSumner:
I checked that all browsers supported this before implementing, seems like I missed something somewhere. I'll go back through and figure it out and submit a patch for this issue.

MelSumner:
Update: this fell through the cracks so I'm working on it again, apologies. 

This is from the documentation: 

> However, while [navigator.clipboard.readText()](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/readText) and [navigator.clipboard.writeText()](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) work on all browsers, [navigator.clipboard.read()](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read) and [navigator.clipboard.write()](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write) do not. For example, on Firefox at the time of writing, [navigator.clipboard.read()](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/read) and [navigator.clipboard.write()](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write) are not fully implemented

Now, this is a bit confusing, since I did use writeText and readText here. However, I'll add guards as you've suggested and make the solution more robust.

PR #334, opened 2022-06-01
resolves FF bug related to `copy-item` component

MelSumner:
### :pushpin: Summary

If merged, this PR should resolve issue #290 

I can't quite get the conditional to activate correctly in Firefox, so I ended up taking out the clipboard API and just using Firefox's recommended approach. I tested it in Chrome, FF and Safari and it's working as expected, but I don't like the idea of only using execCommand (which has been deprecated). 

Here's what I wanted to do though, maybe someone can see what I am missing: 

```js
@action
  async copyCode() {
    // check to see if the browser even supports the clipboard-write permission
    if (navigator.permissions.query({ name: 'clipboard-write' })) {
      // query for the clipboard-write navigator permission
      const permResult = await navigator.permissions.query({
        name: 'clipboard-write',
      });

      // if the permission is granted, then we can copy the text
      if (permResult.state === 'granted' || permResult.state === 'prompt') {
        // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard
        await navigator.clipboard.writeText(this.args.text);
        const result = await navigator.clipboard.readText();

        // if the copied text is equal to the text to be copied, then we have been successful
        if (result === this.args.text) {
          this.isSuccess = true;
          console.log(`result is ${result}`);

          // make it fade back to the default state
          setTimeout(() => {
            this.isSuccess = false;
          }, 1000);
        }
      } else {
        // if permission is denied, we want to know about it
        console.log(`Permission denied: ${permResult.state}`);
        alert(
          'you have not granted your browser permission to use the clipboard'
        );
      }
    } else {
      // if the clipboard-write permission doesn't exist for the browser, we want to try something else
      let copyText = await this.args.text;
      // https://codepen.io/chrisdavidmills/full/gzYjag/
      // execCommand is supposedly deprecated but it's also what Firefox says to do in this situation so we'll try it
      document.execCommand('copy', false, copyText || '');

      console.log(`Copied text: ${copyText}`);

      if (copyText === this.args.text) {
        this.isSuccess = true;
        console.log(`copied text is ${copyText}`);

        // make it fade back to the default state
        setTimeout(() => {
          this.isSuccess = false;
        }, 1000);
      }
    }
  }
  ```

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/dropdown/list-item/copy-item.js
+++ b/packages/components/addon/components/hds/dropdown/list-item/copy-item.js
@@ -39,12 +39,15 @@ export default class HdsDropdownListItemCopyItemComponent extends Component {
   @action
   async copyCode() {
     let copyText = await this.args.text;
+    // https://codepen.io/chrisdavidmills/full/gzYjag/
     // execCommand is supposedly deprecated but it's also what Firefox says to do in this situation so we'll try it
-    await document.execCommand('copy', false, copyText || '');
+    // also execCommand is synchronous so we can't await it
+    document.execCommand('copy', false, copyText || '');
 
     if (copyText === this.args.text) {
       this.isSuccess = true;
-      console.log(`success: ${this.isSuccess}; copied text: ${copyText}`);
+      // if you un-comment the next line you'll see what you expect in the browser when you visit the showcase for this component.
+      // console.log(`success: ${this.isSuccess}; copied text: ${copyText}`);
       // make it fade back to the default state
       setTimeout(() => {
         this.isSuccess = false;
--- a/packages/components/addon/components/hds/dropdown/list-item/copy-item.js
+++ b/packages/components/addon/components/hds/dropdown/list-item/copy-item.js
@@ -38,7 +38,7 @@ export default class HdsDropdownListItemCopyItemComponent extends Component {
 
   @action
   async copyCode() {
-    let copyText = await this.args.text;
+    let copyText = this.args.text;
     // https://codepen.io/chrisdavidmills/full/gzYjag/
     // execCommand is supposedly deprecated but it's also what Firefox says to do in this situation so we'll try it
     // also execCommand is synchronous so we can't await it
--- a/packages/components/addon/components/hds/dropdown/list-item/copy-item.js
+++ b/packages/components/addon/components/hds/dropdown/list-item/copy-item.js
@@ -38,14 +38,13 @@ export default class HdsDropdownListItemCopyItemComponent extends Component {
 
   @action
   async copyCode() {
-    // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard
-    await navigator.clipboard.writeText(this.args.text);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/components/addon/components/hds/dropdown/list-item/copy-item.js [hunk 1]
```diff
@@ -38,14 +38,16 @@ export default class HdsDropdownListItemCopyItemComponent extends Component {
 
   @action
   async copyCode() {
-    // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard
-    await navigator.clipboard.writeText(this.args.text);
-    const result = await navigator.clipboard.readText();
+    let copyText = await this.args.text;
```
Dhaulagiri, line 41: i don't think this is asynchronous

```suggestion
    let copyText = this.args.text;
```

Dhaulagiri, line 41: i don't think this is asynchronous

```suggestion
    let copyText = this.args.text;
```

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
```diff
@@ -38,14 +38,16 @@ export default class HdsDropdownListItemCopyItemComponent extends Component {
...
+    // also execCommand is synchronous so we can't await it
+    document.execCommand('copy', false, copyText || '');
 
-    if (result === this.args.text) {
+    if (copyText === this.args.text) {
       this.isSuccess = true;
-      // console.log(`result is ${result}`);
-
+      // if you un-comment the next line you'll see what you expect in the browser when you visit the showcase for this component.
+      // console.log(`success: ${this.isSuccess}; copied text: ${copyText}`);
```
Dhaulagiri, line 50: [suggestion] i generally try to not to leave behind debugging statements, so I think this and the line above could be removed

Dhaulagiri, line 50: [suggestion] i generally try to not to leave behind debugging statements, so I think this and the line above could be removed

Dhaulagiri (commented):
This seems sufficient to me to unblock this. 

Can you add a changeset for this so it gets included in a release?

packages/components/addon/components/hds/dropdown/list-item/copy-item.js
didoo: +1

didoo (changes requested):
@MelSumner a few points:
- I tested the feature in Chrome and Safari and it doesn't work for me
- the issue with Firefox was with the _read_ part (`navigator.clipboard.readText`) not the _write_ part; with the solution proposed in this PR we're not reading at all the clipboard, so at that point we could have just used the old/existing code and remove (or even better conditionally test) the `navigator.clipboard.readText`

Let me know if you prefer to pair on this, ok?

PR #340, opened 2022-06-07
Form controls - Groups of `checkbox`, `radio`, and `toggle`

didoo:
### :pushpin: Summary

Quick PR on top of #285 so @alex-ju can review the changes

Previews: 
- [Checkbox](https://hds-components-git-form-controls-groups-hashicorp.vercel.app/components/form/checkbox#showcase)
- [Radio](https://hds-components-git-form-controls-groups-hashicorp.vercel.app/components/form/radio#showcase)
- [Toggle](https://hds-components-git-form-controls-groups-hashicorp.vercel.app/components/form/toggle#showcase)

---
  - https://app.asana.com/0/0/1202346498517521


packages/components/app/styles/components/form/common.scss
```diff
@@ -13,7 +13,10 @@
 .hds-form-field--vertical-layout {
   display: grid;
   gap: 4px;
-  width: max-content;
+
+  .hds-form-field__label {
+    width: fit-content; // needed to avoid extra invisible space since the <label> is interactive
```
alex-ju, line 18: Good catch! I guess it's worth testing under different types of containers after all 😬

packages/components/addon/components/hds/form/checkbox/group.hbs
```diff
@@ -0,0 +1,9 @@
+<fieldset class={{this.classNames}}>
+  {{yield (hash Legend=(component "hds/form/legend" _contextualClass="hds-form-group__legend"))}}
+  <div class="hds-form-group__control-fields-wrapper">
+    {{yield
+      (hash Checkbox::Field=(component "hds/form/checkbox/field" _contextualClass="hds-form-group__control-field"))
+    }}
+  </div>
+  {{yield (hash Error=(component "hds/form/error" _contextualClass="hds-form-group__error"))}}
+</fieldset>
```
alex-ju, line 9: For discussion: I'm thinking we may be able to abstract this wrapper (fieldset + legend) into a utility component.
```
<Hds::Form::Fieldset>
  {{yield
    (hash Checkbox::Field=(component "hds/form/checkbox/field" _contextualClass="hds-form-group__control-field"))
  }}
</Hds::Form::Fieldset>
```
The reason I'm suggesting this is because we'll likely introduce more logic here that could be shared between group components. A few examples that I can think of:
 - We may want to introduce a helper text at the group level (below the legend)
 - We may need to create relationship between the fieldset and the error message (e.g. `<fieldset aria-describedby="error-message-id">` (and potentially for the helper text).

I appreciate this suggestion can be addressed in a further iteration, but it's something for us to consider.

alex-ju (approved):
The whole refactor and the grouping approach makes sense to me.
We might want to introduce the option of a helper-text at the group level (probably something already on your radar), so I left a comment (around abstracting the fieldset into a shared template) to remind myself to discuss it with you (if it's feasible and brings any value).

packages/components/addon/components/hds/form/checkbox/group.hbs
didoo: https://app.asana.com/0/1202008719910219/1202406318316728

PR #353, opened 2022-06-09
Form controls – abstract fieldset and field

alex-ju:
### :pushpin: Summary

Follow up on #340, [as suggested in the comment](https://github.com/hashicorp/design-system/pull/340#discussion_r891231397) we abstract `Fieldset` (based on `<Hds::Form::[component-name]::Group>`) and `Field` (based on `<Hds::Form::[component-name]::Field>`), mainly to avoid code repetition within the repo and make it easier to maintain.

For `<Hds::Form::Fieldset>` we preserve the same properties we used for `<Hds::Form::[component-name]::Group>`.
In `<Hds::Form::Field>` we introduce, two properties `@fieldType` and `@fieldLayout` to accommodate the input field types we currently support, and the two current layouts (vertical and flag).

### :link: External links

[Asana ticket](https://app.asana.com/0/1202168297424991/1202406318316728)

***

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/addon/components/hds/form/field/index.hbs
```diff
@@ -0,0 +1,57 @@
+<div class={{this.classNames}}>
```
didoo, line 1: Maybe there is a simpler approach:

**hds/form/field/index.hbs**
```
  {{yield (hash Label=(component "hds/form/label" controlId=this.id _contextualClass="hds-form-field__label"))}}
  {{yield
    (hash
      HelperText=(component "hds/form/helper-text" controlId=this.id _contextualClass="hds-form-field__helper-text")
    )
  }}
  {{yield (hash Control=(component "hds/yield"))}}
  {{yield (hash Error=(component "hds/form/error" controlId=this.id _contextualClass="hds-form-field__error"))}}
</div>
```

**hds/form/text-input/field.hbs** (for example)
```
<Hds::Form::Field as |F|>
  {{yield (hash Label=F.Label HelperText=F.HelperText Error=F.Error)}}
  <F.Control>
    <Hds::Form::TextInput::Base
      @type={{@type}}
      @value={{@value}}
      @isInvalid={{@isInvalid}}
      ...attributes
      id={{this.id}}
      aria-describedby={{this.ariaDescribedBy}}
    />
  </F.Control>
</Hds::Form::Field>
```

Let's discuss the pros/cons in a meeting.

packages/components/addon/components/hds/form/field/index.js
```diff
@@ -3,7 +3,18 @@ import Component from '@glimmer/component';
...
-export default class HdsFormRadioFieldComponent extends Component {
+export default class HdsFormFieldIndexComponent extends Component {
+  /**
+   * Sets the layout of the group
+   *
+   * @param layout
+   * @type {enum}
+   * @default 'vertical'
+   */
+  get fieldLayout() {
```
didoo, line 14: since it's implicitly a `field` maybe I would call it just `@layout`?

packages/components/addon/components/hds/form/field/index.js
```diff
@@ -3,7 +3,18 @@ import Component from '@glimmer/component';
...
+    return this.args.fieldLayout ?? 'vertical';
```
didoo, line 15: personal preference, but in this case I would not have a default (it's more explicit in the code what layout we're using, and in any case it's only in 3+3 places where we have to add it)

packages/components/addon/components/hds/form/fieldset/index.hbs
```diff
@@ -0,0 +1,32 @@
+<fieldset class={{this.classNames}} ...attributes aria-describedby={{this.ariaDescribedBy}}>
```
didoo, line 1: maybe we can have the same approach as in the `field` and declare a container that does nothing more that yielding its content?

didoo (commented):
added a few comments. let's discuss them 1:1

packages/components/.eslintrc.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
   ],
   env: {
     browser: true,
+    node: true,
```
didoo, line 21: I have renamed the codemod files (adding a `.txt` to the extension) so you can remove this now

alex-ju: I removed this at the rebase

packages/components/addon/components/hds/form/field/index.js
alex-ju: Declared 'vertical' or 'flag' explicitly – no defaults now.

packages/components/addon/components/hds/form/field/index.js
alex-ju: Update from `fieldLayout` to `layout`

packages/components/addon/components/hds/form/field/index.hbs
didoo: @alex-ju let me know if you want to rubber duck the problem

packages/components/addon/components/hds/form/field/index.js
```diff
@@ -3,7 +3,17 @@ import Component from '@glimmer/component';
...
+  /**
+   * Sets the layout of the field
+   *
+   * @param layout
+   * @type {enum}
+   */
+  get layout() {
+    return this.args.layout;
+  }
+
```
didoo, line 16: [suggestion] maybe we should add the different layouts as ENUMs in the components, and add assertions that check that the user passes the right value (in a different ticket/new PR)

(later we can add also tests for this)

packages/components/addon/components/hds/form/toggle/field.hbs
```diff
@@ -1,18 +1,13 @@
...
-  {{yield (hash Label=(component "hds/form/label" controlId=this.id _contextualClass="hds-form-field__label"))}}
-  {{yield
-    (hash
-      HelperText=(component "hds/form/helper-text" controlId=this.id _contextualClass="hds-form-field__helper-text")
-    )
-  }}
-  {{yield (hash Error=(component "hds/form/error" controlId=this.id _contextualClass="hds-form-field__error"))}}
-</div>
\ No newline at end of file
+<Hds::Form::Field @isInvalid={{@isInvalid}} @layout="flag" as |F|>
```
didoo, line 1: [chore] can you double check that this is still needed?
```suggestion
<Hds::Form::Field @layout="flag" as |F|>
```

didoo (approved):
good to go for me, left only a suggestion about a possible extra argument, and a couple of things we probably want to do (after merge)

packages/components/addon/components/hds/form/toggle/field.hbs
alex-ju: Not needed, it's a leftover. I just added a commit to remove it.

packages/components/addon/components/hds/form/field/index.js
alex-ju: Good suggestion; pushed a commit to raise a warning if `@layout` is not `vertical` or `flag`

PR #356, opened 2022-06-09
Add stylelint to components

Dhaulagiri:
ezgif-3-0c250e884a

### :pushpin: Summary

This PR adds [stylelint](https://stylelint.io/) to our `components` package. If we are happy with this we could duplicate a similar setup in `ember-flight-icons` and `flight-website` or consider how to consolidate them.

This adds stylelint with the following rules disabled either because they didn't seem that valuable or they seemed to complex to fix as part of a first pass:

 * [number-max-precision](https://stylelint.io/user-guide/rules/list/number-max-precision/)
 * [scss/comment-no-empty](https://stylelint.io/user-guide/rules/list/comment-no-empty/)
 * [scss/no-global-function-names](https://github.com/stylelint-scss/stylelint-scss/blob/master/src/rules/no-global-function-names/README.md)
 * [scss/at-import-no-partial-leading-underscore](https://github.com/stylelint-scss/stylelint-scss/blob/master/src/rules/at-import-no-partial-leading-underscore/README.md)
 * [scss/at-mixin-pattern](https://github.com/stylelint-scss/stylelint-scss/blob/master/src/rules/at-mixin-pattern/README.md)
 * [selector-class-pattern](https://stylelint.io/user-guide/rules/list/selector-class-pattern/)

**Note** I had to add this to my settings.json in vscode to get formatting on save to work
```
"[scss]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode",
    "editor.formatOnSave": true
},
```

### :link: External links

* [Stylelint](https://stylelint.io/)

***

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/app/styles/components/alert.scss
```diff
@@ -23,6 +23,8 @@
   .hds-alert--type-compact & {
     height: 14px;
     width: 14px;
+    margin-right: 8px;
```
didoo, line 26: [issue] is this PR adding code?

packages/components/app/styles/components/alert.scss
```diff
@@ -164,14 +173,10 @@
 }
 
 .hds-alert--type-compact {
-  .hds-alert__icon {
```
didoo, line 167: mmm, seems the linter is changing code here. we should avoid this

didoo (commented):
@Dhaulagiri I'm 1000% on adding _stylelint_ to our stack. but I have seen too many changes that I would not want it to do. which leads to my "suggestion": would it be possible to have this PR doing only a sub-set of things:
- add the infra for supporting _stylelint_
- disable all the rules so that technically does nothing to the code (no changes in CSS in the PR)
Once this is in master (and once Kristin has joined the team) we will organize a meeting/workshop/team-paring session and start to enable rules one by one, and decide which one we want, which we don't, etc.

packages/components/app/styles/components/alert.scss
Dhaulagiri: this was because this selector is technically being duplicated. could definitely disable that rule if we don't want it.

https://github.com/hashicorp/design-system/pull/356/commits/81d6fff02b9fa2d060a39c5093618f36d701ea02

Dhaulagiri:
> but I have seen too many changes that I would not want it to do

@didoo I tried to piece out the more tricky changes into specific commits if any of those stand out as a patterns we'd prefer to ignore.

If we want to disable _all_ the rules now then it's probably not worth landing this change right now and I can close the PR. It'd be fairly trivial to cherry pick the initial commit to restart this effort when we're more ready to discuss particular rules we like/don't like.

didoo:
> > but I have seen too many changes that I would not want it to do
> 
> @didoo I tried to piece out the more tricky changes into specific commits if any of those stand out as a patterns we'd prefer to ignore.
> 
> If we want to disable _all_ the rules now then it's probably not worth landing this change right now and I can close the PR. It'd be fairly trivial to cherry pick the initial commit to restart this effort when we're more ready to discuss particular rules we like/don't like.

No need to close it, just move it to draft. When Kristin has joined it will be easy to set up a meeting and go through the rules together (she created a CSS Library, I am sure she has strong preferences too ;) )

Dhaulagiri:
This will be brought back in a more current PR

PR #366, opened 2022-06-13
Update @percy/cli requirement from ^1.2.1 to ^1.3.0 in /packages/flight-website

Dhaulagiri:
### :pushpin: Summary

Manually doing this since the problem in https://github.com/hashicorp/design-system/pull/365 prevented this from happening automatically


Dhaulagiri:
Assumed dependabot would take a minute before opening its version but today is full of surprises

PR #379, opened 2022-06-15
Form controls: Determine `aria-describedby` more accurately

alex-ju:
### :pushpin: Summary

To overcome limitations of not being able to check if a certain contextual component (e.g. helper-text or error) has content, we turn to a client-side approach for an accurate `aria-describedby` value.

The current implementation of `aria-describedby` in our form components lists both the `id` of the helper text and the `id` of the error block regardless of which of these two block are being passed in. While assistive technologies fail gracefully in such scenario (if `aria-describedby` contains the `id` of an element that doesn't exist in the DOM) any automated test will pick up this anomaly and report it, so we'd like to be accurate.

### :link: External links

[Asana task](https://app.asana.com/0/1202168297424991/1202443172675747)

***

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/form/field/index.js
+++ b/packages/components/addon/components/hds/form/field/index.js
@@ -9,10 +9,11 @@ export const LAYOUT_TYPES = ['vertical', 'flag'];
 
 export default class HdsFormFieldIndexComponent extends Component {
   @tracked ariaDescribedBy;
+  @tracked descriptors = [];
 
   @action
-  didInsert(element) {
-    this.HTMLElement = element;
+  appendDescriptors(element) {
+    this.descriptors.push(element.id);
     this.ariaDescribedBy = getAriaDescribedBy(this);
   }
 
--- a/packages/components/addon/components/hds/form/field/index.js
+++ b/packages/components/addon/components/hds/form/field/index.js
@@ -12,7 +12,7 @@ export default class HdsFormFieldIndexComponent extends Component {
   @tracked descriptors = [];
 
   @action
-  appendDescriptors(element) {
+  appendDescriptor(element) {
     this.descriptors.push(element.id);
     this.ariaDescribedBy = getAriaDescribedBy(this);
   }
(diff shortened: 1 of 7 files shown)


alex-ju:
I have added a refactor commit based on feedback from the community, then rebased.

MelSumner (approved):
We did this a little differently in the button component- put the `aria-` attribute in the component template by default, and used a conditional-- render content if there was content to be rendered, or `null` (if `null`, Ember won't render the attribute to the DOM).

I don't immediately see any issues with this approach, if @didoo is fine with the inconsistency then I have no objections. 

Thanks for working on this!

packages/components/addon/components/hds/form/field/index.js
```diff
@@ -1,12 +1,22 @@
...
 import { getAriaDescribedBy } from '../utils/getAriaDescribedBy';
 
 export const LAYOUT_TYPES = ['vertical', 'flag'];
 
 export default class HdsFormFieldIndexComponent extends Component {
+  @tracked ariaDescribedBy;
+  @tracked descriptors = [];
+
+  @action
+  appendDescriptors(element) {
```
didoo, line 15: [nitpick] should this method be called `appendDescriptor` (singular)?

packages/components/addon/components/hds/form/utils/getAriaDescribedBy.js
```diff
@@ -1,13 +1,16 @@
...
-  // append @ariaDescribedBy arg, if provided
-  if (element.args.ariaDescribedBy) {
-    ariaDescribedBy.push(element.args.ariaDescribedBy);
+
+  // append descriptor's IDs, if provided
+  if (element.descriptors) {
+    element.descriptors.forEach((descriptor) =>
+      ariaDescribedBy.push(descriptor)
+    );
+  }
```
didoo, line 9: [nitpick] can be simplified even further, without looping on the entries
```suggestion
  if (element.descriptors) {
    ariaDescribedBy.concat(element.descriptors);
  }
```

didoo (approved):
LGTM, only a couple of nitpicks if you want.

Issue #358, opened 2022-06-09
percy duplicate snapshots error


Dhaulagiri:
After https://github.com/hashicorp/design-system/pull/217 landed we started to see errors like the below [in CI runs for components](https://github.com/hashicorp/design-system/runs/6614452283?check_suite_focus=true#step:6:383). Things seem to be working ok still but it'd be good to fix this if we can
<img width="1378" alt="screenshot_1542" src="https://user-images.githubusercontent.com/1672302/172915935-44fc840d-001b-41ae-85e7-d031d686938c.png">

Dhaulagiri:
The percy tests are running more than once as seen [here](https://github.com/hashicorp/design-system/runs/6919371366?check_suite_focus=true#step:6:26) and [here](https://docs.google.com/document/d/1WPyzphzD0upycOwyvCSCD3cOoGLvfUS1j6YQqX6Z6-I/edit)

PR #381, opened 2022-06-16
Fix odd test bug

Dhaulagiri:
### :pushpin: Summary

Fixes https://github.com/hashicorp/design-system/issues/358

Refactors a single test to avoid an odd bug in our test suite

After some sleuthing 🕵️  I found out how to reproduce the issue (in CI, never locally 🤷 )

* For unclear reasons our acceptance test creating percy snapshots was running _twice_
* It started _after_ https://github.com/hashicorp/design-system/pull/217/files was merged
* A lot changed here but nothing stood out to me until some trial and error highlighted the addition of a test using sinon
* If I commented out this test the problem went away
https://github.com/hashicorp/design-system/blob/cb383903acc5341f2b2619d63879bb8a2901519f/packages/components/tests/integration/components/hds/interactive/index-test.js#L133-L139

I'm a bit rusty, but I made a pass at refactoring this test to not use sinon and a) the test "works" and b) the errors are no longer present in our CI runs if you look at one of the tests for components.

Ideally we could understand _why_ this test was causing this in the event we want to add sinon back in the future but I've hit my timebox limit on this for now. My suspicion is that the sinon usage was causing some hidden error that sent things astray.

***

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/tests/integration/components/hds/interactive/index-test.js
+++ b/packages/components/tests/integration/components/hds/interactive/index-test.js
@@ -2,7 +2,6 @@ import { module, test, skip } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 import { render, triggerKeyEvent } from '@ember/test-helpers';
 import { hbs } from 'ember-cli-htmlbars';
-import sinon from 'sinon';
 
 module('Integration | Component | hds/interactive/index', function (hooks) {
   setupRenderingTest(hooks);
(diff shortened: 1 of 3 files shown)


packages/components/tests/integration/components/hds/interactive/index-test.js
```diff
@@ -131,10 +130,12 @@ module('Integration | Component | hds/interactive/index', function (hooks) {
...
   test('it should dispatch a click event when pressing space key on a link', async function (assert) {
-    await render(hbs`<Hds::Interactive @href="#" id="test-interactive"/>`);
-    var interactiveComponent = document.querySelector('#test-interactive');
-    var keyUpSpy = sinon.spy(interactiveComponent, 'click');
-    await triggerKeyEvent(interactiveComponent, 'keyup', ' ');
-    assert.true(keyUpSpy.calledOnce);
+    let clicked = false;
+    this.set('clickHandler', () => (clicked = true));
+    await render(
+      hbs`<div {{on "click" this.clickHandler}}><Hds::Interactive @href="javascript:;" id="test-interactive"/></div>`
```
alex-ju, line 136: So we rely on the click event bubbling up, then capture it at the parent level using a boolean. Given we only use sinon for this test I'm happy with the change. We may get back to square one if we won't manage to use this pattern for future scenarios, but that's a problem for tomorrow.

alex-ju (approved):
Thanks for investigating this! I've also spent some time yesterday trying to replicate the issue locally but no luck.

packages/components/tests/integration/components/hds/interactive/index-test.js
Dhaulagiri: > We may get back to square one if we won't manage to use this pattern for future scenarios

That was my fear with "fixing" it this way as well.

PR #382, opened 2022-06-17
Form controls - Base elements (01)

didoo:
### :pushpin: Summary

This is the first PR for the `form controls` elements. It's about the low-level base components used to build the "field" and "group" controls:
- `label`
- `helper-text`
- `error`
- `legend`
- `field`
- `fieldset`

For the nomenclature see this illustration:
screenshot_1564

_Notice: the branch has been extracted from `form-controls/spike-playground` in #285 via a diff with the `main` branch and then cherry-picked only the changes needed for this PR._

In this PR we (me and @alex-ju) have:
- added the `base elements` for  the `form controls` components
- added the documentation page for the `base elements`
- added integration tests and percy testing
- done some minor tweaks to other components/files for consistency 

Preview of the documentation page: https://hds-components-git-form-controls-01-base-elements-hashicorp.vercel.app/components/form/base-elements

Notice: the page is not linked in the index file of the dummy/scrappy website on purpose, so it's not visible to the public.

***

### 👀 How to review

Reviewer's checklist:

- [x] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/form/error/index.hbs
+++ b/packages/components/addon/components/hds/form/error/index.hbs
@@ -1,0 +1,6 @@
+<div class={{this.classNames}} id={{this.id}} {{did-insert this.onInsert}} ...attributes>
+  <FlightIcon class="hds-form-error__icon" @name="alert-diamond-fill" />
+  <div class="hds-form-error__content">
+    {{yield (hash Message=(component "hds/form/error/message"))}}
+  </div>
+</div>
--- a/packages/components/addon/components/hds/form/error/message.hbs
+++ b/packages/components/addon/components/hds/form/error/message.hbs
@@ -1,0 +1,3 @@
+<p class="hds-form-error__message" ...attributes>
+  {{yield}}
+</p>
--- a/packages/components/addon/components/hds/form/fieldset/index.hbs
+++ b/packages/components/addon/components/hds/form/fieldset/index.hbs
@@ -1,0 +1,23 @@
+<fieldset class={{this.classNames}} id={{this.id}} ...attributes aria-describedby={{this.ariaDescribedBy}}>
+  {{yield (hash Legend=(component "hds/form/legend" contextualClass="hds-form-group__legend"))}}
+  {{yield
+    (hash
+      HelperText=(component
+        "hds/form/helper-text"
+        controlId=this.id
+        onInsert=this.appendDescriptor
+        contextualClass="hds-form-group__helper-text"
+      )
+    )
+  }}
+  <div class="hds-form-group__control-fields-wrapper">
+    {{yield (hash Control=(component "hds/yield"))}}
+  </div>
+  {{yield
+    (hash
+      Error=(component
+        "hds/form/error" controlId=this.id onInsert=this.appendDescriptor contextualClass="hds-form-group__error"
+      )
+    )
+  }}
+</fieldset>
--- a/packages/components/addon/components/hds/form/utils/getElementId.js
+++ b/packages/components/addon/components/hds/form/utils/getElementId.js
@@ -1,0 +1,14 @@
+import { guidFor } from '@ember/object/internals';
+
+export function getElementId(element) {
+  // use @id arg, if provided
+  if (element.args.id) {
+    return element.args.id;
+  }
+
+  // otherwise, generate and memoize a guid
+  if (!element._id) {
+    element._id = guidFor(element);
+  }
+  return element._id;
+}
--- a/packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
+++ b/packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
@@ -525,7 +525,7 @@
     <code class="dummy-code">controlId</code>
     is provided, no ID is generated (but if needed it can be passed directly as HTML attribute).</p>
   <p class="dummy-paragraph">
-    There may be cases in which the error is made of multiple messages. In this case it's possible to iterate over an
+    There may be cases in which the error is made of multiple messages. In this case it's possible to iterate over a
     collection of error messages:
   </p>
   {{! prettier-ignore-start }}
@@ -710,7 +710,7 @@
     <F.Error>This is the error</F.Error>
   </Hds::Form::Fieldset>
   <p class="dummy-paragraph">Depending on the context/need, one may want to pass just the legend, just the helper text,
-    both or none, while he error message is likely conditional to the validation of the inputs provided by the user.</p>
+    both or none, while the error message is likely conditional to the validation of the inputs provided by the user.</p>
   <p class="dummy-paragraph"><em><strong>Important:</strong>
       in this case the layout/styling of the content inside the "control" container is left to the consumer.</em></p>
 
--- a/packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
+++ b/packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
@@ -613,7 +613,7 @@
           <input
             type="email"
             id=\{{F.id}}
-            value="john.doe@email.com"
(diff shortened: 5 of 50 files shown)


packages/components/addon/components/hds/form/error/message.hbs
```diff
@@ -0,0 +1,3 @@
+<p class="hds-form-error__message" ...attributes>
+  {{yield}}
+</p>
```
MelSumner, line 3: [a11y improvement suggestion] [equal experience] Looking through the docs - since we're supporting multiple error messages, it would be an improvement to put these inside of a `ul`/`li` structure. The reason is that the list count is read out to the user with assistive tech, so they have a good idea of how many errors there are, without manually counting. This allows the user to have a more complete idea of what is going on, and provides them with the same information that users already know if they can visually consume the information on the screen.

The current implementation does not violate any current WCAG Success Criteria.

packages/components/addon/components/hds/form/error/index.hbs
```diff
@@ -0,0 +1,6 @@
+<div class={{this.classNames}} id={{this.id}} {{did-insert this.onInsert}} ...attributes>
+  <FlightIcon class="hds-form-error__icon" @name="alert-diamond-fill" />
```
alex-ju, line 2:
```suggestion
  <FlightIcon class="hds-form-error__icon" @name="alert-diamond-fill"  @title="Error: "/>
```
Suggestion: At the moment, there is not much distinction for a screen reader user between the helper text and the error message, whereas a sighted user can distinguish between the two using either color or the icon. I suggest we prefix the error message with something like 'Error: ' to help screen reader users differentiate them, but I would like to get @MelSumner's thoughts on this beforehand.

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
```diff
@@ -0,0 +1,1240 @@
...
+  <h4 class="dummy-h4">Form::Error</h4>
+  <p class="dummy-paragraph" id="component-api-form-helper-error">
+    Here is the API for the component:
+  </p>
+  <dl class="dummy-component-props" aria-labelledby="component-api-form-helper-error">
+    <dt>controlId <code>string</code></dt>
+    <dd>
+      <p>The ID of the form control associated with the error. This is used to populate the
+        <code class="dummy-code">id</code>
+        HTML attribute (with a
```
alex-ju, line 99:
```suggestion
        HTML attribute (with an
```

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
```diff
@@ -0,0 +1,1240 @@
...
+        value="john.doe@email.com"
+        class="my-custom-class"
+        aria-describedby={{F.ariaDescribedBy}}
+      />
+    </F.Control>
+    <F.Error>This is the error</F.Error>
+  </Hds::Form::Field>
+  <p class="dummy-paragraph">Depending on the context/need, one may want to pass just the label, or the label
+    <em>and</em>
+    the helper text, while he error message is likely conditional to the validation of the input provided by the user.</p>
```
alex-ju, line 640:
```suggestion
    the helper text, while the error message is likely conditional to the validation of the input provided by the user.</p>
```

alex-ju (approved):
Looks good to me, but I'm a biased contributor to the code. Left a suggestion that I'd like to check with Melanie and spotted two typos (likely introduced by me).

packages/components/addon/components/hds/form/error/index.hbs
MelSumner: @alex-ju all Flight icons have aria-hidden on them, so they are not presented to the user with AT at all. This was by design.

packages/components/addon/components/hds/form/utils/getElementId.js
```diff
@@ -0,0 +1,14 @@
+import { guidFor } from '@ember/object/internals';
```
MelSumner, line 1: [question] can you explain why this is in a utils folder inside of the form components folder, instead of a standard helper (`ember generate helper elementId` or something)?

Dhaulagiri: @MelSumner helpers are meant to be used from within `.hbs` templates, not in `.js` files, correct?

packages/components/tests/integration/components/hds/form/field/index-test.js
```diff
@@ -0,0 +1,118 @@
...
+      .hasAttribute(
+        'aria-describedby',
+        `helper-text-${controlId} error-${controlId}`
+      );
+    assert
+      .dom('.hds-form-field__error')
+      .hasAttribute('id', `error-${controlId}`);
+  });
+
+  // ATTRIBUTES
```
Dhaulagiri, line 107: should there be tests for the `@layout` argument?

packages/components/addon/components/hds/form/error/message.hbs
didoo: @alex-ju can chime in on this about his thinking in more detail. What I can say is that we discussed this specifically, but since there's no way in Ember to know which (and so how many) contextual components are passed as children, we can't discern when there's a single error message vs when there are multiple ones. We came to the conclusion that since in most of the cases we have a single message, it would be weird (and also maybe even less accessible) to have a list that contains one single message item

packages/components/addon/components/hds/form/fieldset/index.hbs
```diff
@@ -0,0 +1,23 @@
+<fieldset class={{this.classNames}} id={{this.id}} ...attributes aria-describedby={{this.ariaDescribedBy}}>
+  {{yield (hash Legend=(component "hds/form/legend" contextualClass="hds-form-group__legend"))}}
+  {{yield
+    (hash
+      HelperText=(component
+        "hds/form/helper-text"
+        controlId=this.id
```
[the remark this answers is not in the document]
didoo: Not sure about the issue, can you clarify?

For context:
- if the ID is generated by the component, then it's using Ember's `guidFor` helper method, so it should be unique
- if the ID is passed by the consumer, we don't have a way to check, is their responsibility.

If you mean an integration test, not sure how to do it. How would you test it?

packages/components/addon/components/hds/form/utils/getElementId.js
didoo: yep! what @Dhaulagiri says (and doesn't make sense to move it up in the hierarchy, because is very specific to the form controls needs).

packages/components/tests/integration/components/hds/form/field/index-test.js
didoo: @Dhaulagiri it's at line 22 of this file.

packages/components/addon/components/hds/form/error/index.hbs
MelSumner: That was supposed to be a temporary measure put in place, though. SVG implementation is simply incomplete which is why we went with the simpler route to begin with. It would be sensible to leave the decision as it stands and keep icons aria-hidden.

MelSumner (approved):

alex-ju:
Rebased to main and clean-up the git history a little bit

packages/components/addon/components/hds/form/error/index.hbs
didoo: @alex-ju let's pause on this for the moment, we can re-discuss this idea later when we will do a final check on the form controls before the "official" release

didoo (commented):
@alex-ju I've reviewed everything, and pushed a few minor fixes (see my last commits).
If they are OK with you, feel free to merge, it's good to go for me.

/cc @Dhaulagiri

PR #385, opened 2022-06-17
Form controls - `TextInput` component (02)

didoo:
### :pushpin: Summary

> **Note**
> This is not the final version: we will complete the component (documentation included) when the design specs are finalized.

This PR adds the `TextInput` form control to the system in two variants:
- `base`
- `field`

---

For the nomenclature see this illustration:
screenshot_1564

_Notice: the branch has been extracted from `form-controls/spike-playground` in #285 via a diff with the `main` branch and then cherry-picked only the changes needed for this PR._

In this PR we (me and @alex-ju) have:
- added the "base" and "field" `TextInput` form controls as components
- added the documentation page for the `TextInput`
- added integration tests and percy testing

Preview of the documentation page: https://hds-components-git-form-controls-02-text-input-hashicorp.vercel.app/components/form/text-input

Notice: the page is not linked in the index file of the dummy/scrappy website on purpose, so it's not visible to the public.

***

### 👀 How to review

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/app/styles/components/form/text-input.scss
+++ b/packages/components/app/styles/components/form/text-input.scss
@@ -46,9 +46,18 @@
     border-color: var(--token-color-focus-action-internal);
   }
 
+  // READONLY
+
+  &:read-only,
+  &[readonly] {
+    background-color: var(--token-color-surface-strong);
+    border-color: var(--token-color-palette-neutral-400);
+  }
+
   // DISABLED
 
-  &:disabled {
+  &:disabled,
+  &[disabled] {
     background-color: var(--token-color-surface-interactive-disabled);
     border-color: var(--token-color-border-primary);
   }
--- a/packages/components/app/styles/components/form/text-input.scss
+++ b/packages/components/app/styles/components/form/text-input.scss
@@ -1,0 +1,127 @@
+//
+// FORM > TEXT-INPUT
+//
+// properties within each class are sorted alphabetically
+//
+
+// "BASE" CONTROL
+
+.hds-form-text-input {
+  border: 1px solid var(--token-color-palette-neutral-400);
+  border-radius: 5px;
+  box-shadow: var(--hds-elevation-inset-box-shadow);
+  color: var(--token-color-foreground-primary);
+  font-family: var(--token-typography-body-200-font-family);
+  font-size: var(--token-typography-body-200-font-size);
+  line-height: var(--token-typography-body-200-line-height);
+  padding: 7px;
+  width: 100%;
+  max-width: 100%;
+
+  // PLACEHOLDER
+
+  ::placeholder {
+    color: var(--token-color-foreground-faint);
+  }
+
+  // STATUS
+
+  &:hover,
+  &.mock-hover {
+    border-color: var(--token-color-palette-neutral-500);
+  }
+
+  // focus (same for all the states)
+  // TODO add handling of focus-visible
+  &:focus,
+  &.mock-focus {
+    border-color: var(--token-color-focus-action-internal);
+    // TODO: Safari doesn't apply a rounded border
+    outline: 2px solid var(--token-color-focus-action-external);
+    outline-offset: 0px;
+  }
+
+  &:active,
+  &.mock-active {
+    border-color: var(--token-color-focus-action-internal);
+  }
+
+  // DISABLED
+
+  &:disabled {
+    background-color: var(--token-color-surface-interactive-disabled);
+    border-color: var(--token-color-border-primary);
+  }
+
+  // INVALID
+
+  &.hds-form-text-input--is-invalid {
+    border-color: var(--token-color-foreground-critical);
+  }
+}
+
+
+
+// "TYPE" CUSTOMIZATION
+
+.hds-form-text-input {
+
+  // DATE/TIME
+
+  &[type="date"],
+  &[type="time"] {
+
+    // browsers set a specific width for these controls, we want to keep it
+    width: initial;
+
+    // show the native icon dimmed if disabled (hidden in Chrome)
+    &[disabled]::-webkit-calendar-picker-indicator,
+    &:disabled::-webkit-calendar-picker-indicator {
+        visibility: visible;
+        opacity: 0.5;
+    }
+
+    // show the icon if readonly
+    &[readonly]::-webkit-calendar-picker-indicator {
+        visibility: visible;
+    }
+  }
+
+  // we override the default icon with the Flight corresponding one
+  // notice: the original in Chrome has two assets, one for light and one for dark mode, and uses a special syntax, but apparently it doesn't work if used in a stylesheet
+  &[type="date"] {
+    &::-webkit-calendar-picker-indicator {
+      // notice: the icon color is hardcoded here!
+      background-image: url("data:image/svg+xml,%3csvg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'%3e%3cpath d='M11.5.75a.75.75 0 00-1.5 0V1H6V.75a.75.75 0 00-1.5 0V1H3.25A2.25 2.25 0 001 3.25v9.5A2.25 2.25 0 003.25 15h9.5A2.25 2.25 0 0015 12.75v-9.5A2.25 2.25 0 0012.75 1H11.5V.75zm-7 2.5V2.5H3.25a.75.75 0 00-.75.75V5h11V3.25a.75.75 0 00-.75-.75H11.5v.75a.75.75 0 01-1.5 0V2.5H6v.75a.75.75 0 01-1.5 0zm9 3.25h-11v6.25c0 .414.336.75.75.75h9.5a.75.75 0 00.75-.75V6.5z' fill-rule='evenodd' clip-rule='evenodd' fill='%233B3D45'/%3e%3c/svg%3e");
+      background-position: center center;
+      background-size: 16px;
+    }
+  }
+  &[type="time"] {
+    &::-webkit-calendar-picker-indicator {
+      // notice: the icon color is hardcoded here!
+      background-image: url("data:image/svg+xml,%3csvg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'%3e%3cg fill='%233B3D45'%3e%3cpath d='M8.5 3.75a.75.75 0 00-1.5 0V8c0 .284.16.544.415.67l2.5 1.25a.75.75 0 10.67-1.34L8.5 7.535V3.75z'/%3e%3cpath d='M8 0a8 8 0 100 16A8 8 0 008 0zM1.5 8a6.5 6.5 0 1113 0 6.5 6.5 0 01-13 0z' fill-rule='evenodd' clip-rule='evenodd'/%3e%3c/g%3e%3c/svg%3e");
+      background-position: center center;
+      background-size: 16px;
+    }
+  }
+
+  // SEARCH
+  &[type="search"] {
+    padding-left: 32px;
+    // notice: the icon color is hardcoded here!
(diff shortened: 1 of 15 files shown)


packages/components/app/styles/components/form/text-input.scss
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  // we override the default icon with the Flight corresponding one
+  // notice: the original in Chrome has two assets, one for light and one for dark mode, and uses a special syntax, but apparently it doesn't work if used in a stylesheet
+  &[type="date"] {
+    &::-webkit-calendar-picker-indicator {
+      // notice: the icon color is hardcoded here!
+      background-image: url("data:image/svg+xml,%3csvg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'%3e%3cpath d='M11.5.75a.75.75 0 00-1.5 0V1H6V.75a.75.75 0 00-1.5 0V1H3.25A2.25 2.25 0 001 3.25v9.5A2.25 2.25 0 003.25 15h9.5A2.25 2.25 0 0015 12.75v-9.5A2.25 2.25 0 0012.75 1H11.5V.75zm-7 2.5V2.5H3.25a.75.75 0 00-.75.75V5h11V3.25a.75.75 0 00-.75-.75H11.5v.75a.75.75 0 01-1.5 0V2.5H6v.75a.75.75 0 01-1.5 0zm9 3.25h-11v6.25c0 .414.336.75.75.75h9.5a.75.75 0 00.75-.75V6.5z' fill-rule='evenodd' clip-rule='evenodd' fill='%233B3D45'/%3e%3c/svg%3e");
+      background-position: center center;
+      background-size: 16px;
```
alex-ju, line 97: Thought: I've just realised we have this 'icon size' value in so many places we might want to convert it into a variable, in this file at least.

packages/components/tests/integration/components/hds/form/text-input/base-test.js
```diff
@@ -0,0 +1,90 @@
...
+      hbs`<Hds::Form::TextInput::Base id="test-form-text-input" class="my-class" data-test1 data-test2="test" />`
+    );
+    assert.dom('#test-form-text-input').hasClass('my-class');
+    assert.dom('#test-form-text-input').hasAttribute('data-test1');
+    assert.dom('#test-form-text-input').hasAttribute('data-test2', 'test');
+  });
+
+  // ASSERTIONS
+
+  test('it should throw an assertion if an incorrect value for @type is provided', async function (assert) {
```
Dhaulagiri, line 78: [nitpick] as a reviewer i was expecting this to be next to the other tests for `@type` and started writing up a comment there before finding it down here. I assume there as informal convention on how these are being grouped, but what would you think about moving the tests to be next to each other?

packages/components/tests/integration/components/hds/form/text-input/base-test.js
```diff
@@ -0,0 +1,90 @@
...
+    const errorMessage =
+      '@type for "Hds::Form::TextInput" must be one of the following: text, email, password, url, search, date, time, received: foo';
```
Dhaulagiri, line 80: [suggestion] in the future if we were to change the list of input types we'll have to update this error message which feels a little too tightly coupled. for the purpose of this test we care that an error is captured, not necessarily what the exact message is. what do you think?

packages/components/tests/integration/components/hds/form/text-input/field-test.js
```diff
@@ -0,0 +1,115 @@
...
+    assert.dom('input').hasClass('hds-form-field__control');
+  });
+
+  // TYPE
+
+  test('it should render the "text" type if no type is declared', async function (assert) {
+    await render(hbs`<Hds::Form::TextInput::Field />`);
+    assert.dom('input').hasAttribute('type', 'text');
+  });
+  test('it should render the correct type depending on the @type prop', async function (assert) {
```
Dhaulagiri, line 28: [question] in the base tests we test for invalid `@type` values but don't here, is that intentional?

Dhaulagiri: _sidenote: i don't have a good suggestion otherwise but it strikes me how we are effectively copying the exact same tests from the base file into this wrapper component's tests which makes sense since we want to make sure consumers of the wrapper component have things work but feels also redudnant in a way I can't quite put my finger on_

Dhaulagiri (commented):
I defer to @MelSumner @alex-ju  for a more thorough review but left a few comments thoughts that came up in my review

packages/components/app/styles/components/form/text-input.scss
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  // focus (same for all the states)
+  // TODO add handling of focus-visible
+  &:focus,
+  &.mock-focus {
+    border-color: var(--token-color-focus-action-internal);
+    // TODO: Safari doesn't apply a rounded border
+    outline: 2px solid var(--token-color-focus-action-external);
+    outline-offset: 0px;
```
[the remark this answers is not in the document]
didoo: @alex-ju see comment here: https://github.com/hashicorp/design-system/pull/392#discussion_r902415176. I see the same issue with the `<input type="text">` unfortunately :(

packages/components/app/styles/components/form/text-input.scss
didoo: Yes, the idea is to have a set of component-level design tokens, specific for the `form controls`, once the design specs are finalized, so we can share them with DevDot too.

packages/components/tests/integration/components/hds/form/text-input/base-test.js
didoo: By convention, all the "assertions" are collected at the bottom of the test files. <oving them around would spread them in no specific order, and make them easy to spot.

packages/components/tests/integration/components/hds/form/text-input/base-test.js
didoo: Yes, but that's on purpose. I could import (maybe?) the list from the component JS, but let's imagine I make an error there (eg. deleting one type), then the test would automatically auto-fix itself, and the error go unnoticed. Makes sense my thinking?

packages/components/tests/integration/components/hds/form/text-input/field-test.js
didoo: Yes I agree with you, but I went with redundancy in this case because we want to test that the property is correctly drilled down from the `field` to the `base` control.

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
```diff
@@ -0,0 +1,478 @@
...
+  </div>
+  }}
+</section>
+
+<section>
+  <h3 class="dummy-h3" id="accessibility"><a href="#accessibility" class="dummy-link-section">§</a> Accessibility</h3>
+  <p class="dummy-paragraph">🚧 TODO 🚧</p>
+</section>
+
+<section data-test-percy>
```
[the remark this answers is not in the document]
didoo: Continuing the conversation from this PR: https://github.com/hashicorp/design-system/pull/285/files#r896982292

---

As discussed during the code review [the other day], I am not convinced this is the way to go for the "showcase" components. I totally understand the idea to have accessible documentation pages, but the _show_case is intended mainly to _show_ the components, with a particular focus on the developers implementing the components (and Percy) plus as a bonus side-effect, show to the consumers what are the variants (and their combinations) of a component. 

The issue of devs copying&pasting the code from the showcase is minimal, because we don't expect them to copy the HTML, and (unless I'm missing something) is not possible to copy the code from the Ember inspector.

So the last problems remains having pages which will not pass an accessibility test. Here I can propose some alternatives:
- we hide the whole showcase from the accessibility tree (a bit drastic!)
- we hide the parent container/wrapper of the "base" components from the accessibility tree (more specific)
- we accept that not all the elements in the showcase will meet all the accessibility criteria, when considered in isolation
  - this is not something unusual: consider this MDN page: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input - as you can see the top example has a label associated with the `<input>` but the ones in the list of types they don't and they also don't have a `aria-label` associated, and I am not sure I would expect them to have it frankly (their purpose meant to showcase the component, not as real usable/interactive content).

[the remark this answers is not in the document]
didoo: I did a quick test on other design systems websites that we use as reference, these are the results for similar pages:
screenshot_1572
screenshot_1571
screenshot_1570
screenshot_1569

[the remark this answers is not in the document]
MelSumner: Let's forget that this is something that matters to me (an equal member of the team), and look over the other reasons why this should be done. 

It's still inconsistent from the rest of our documentation pages and our approach so far.

I'm inferring two things from your response: 
1. someone with a screen reader will never need to use our documentation
2. other design systems have non-conformant documentation (so it's fine)

I think the first is perhaps the reason that we have to have accessibility laws in place-  what I have asked for is:
1. a small thing to fix 
2. a thing that will make our page as consistent as our other pages are
3. a thing that demonstrates our accessibility statement: intent to provide at least WCAG conformance

Even if I were to accept the argument that "other design systems have conformance issues in their documentation too", then compare the number of critical issues, because that's what this is. 

What would it mean then that a person who needs a screen reader will never visit this page?
1. that we would never hire someone who needed a screen reader
2. that none of us will never need a screen reader
3. that no auditor will ever be using a screen reader in their audit
4. that no site visitor (maybe works on another design system and has come to ours to see how we do things) will ever be using a screen reader

I don't think any of those things are logically safe to assume. 

However, even if I were to accept that a person who needs a screen reader were to never access this documentation, I have to wonder why you're so resistant to this. All of our other pages are conformant. This clearly is a critical WCAG issue. 

I'll even submit a PR to fix it.

packages/components/tests/integration/components/hds/form/text-input/base-test.js
Dhaulagiri: I think my suggestion is more that asserting that _any_ error is thrown is sufficient and what the actual contents of the error message are is not particularly high value

packages/components/tests/integration/components/hds/form/text-input/field-test.js
Dhaulagiri: makes sense, but should we also test for the invalid types here as well or no?

didoo: The invalid `@type` is tested in the `base` component (file `tests/integration/components/hds/form/text-input/base-test.js`, line 78), the `field` is consuming the `base` as inner component.

packages/components/tests/integration/components/hds/form/text-input/base-test.js
didoo: Ahh. I don't think there was a different way to test assertion (there is almost zero literature on this). I am not sure how to test assertions with just the "throw" part. Do you have ideas/suggestions?

Dhaulagiri: I'd need to do some digging. I think it's safe to table my suggestion for now and revisit if/when this is a problem.

didoo (commented):
@alex-ju I've pushed a couple of small changes. for me is good to go, when you want feel free to merge.

alex-ju (approved):

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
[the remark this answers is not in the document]
Dhaulagiri: I opened a ticket in our backlog to track this discussion

PR #402, opened 2022-06-22
Version Packages

hashibot-hds:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Minor Changes

-   [#399](https://github.com/hashicorp/design-system/pull/399) [`62e3077e`](https://github.com/hashicorp/design-system/commit/62e3077e5cf40e93852e9fa3489d3d1970d45070) Thanks [@alex-ju](https://github.com/alex-ju)! - Enable programmatic close on dropdown component


packages/components/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @hashicorp/design-system-components
 
+## 0.13.0
+
+### Minor Changes
```
didoo, line 5: @alex-ju do you think this should be a `minor` or a `patch`? (considered that our current convention, until v. 1.0.0 is releases, is that `minor` is like a `major`)? I don't think we changed the API of the component, we just expanded it.

@Dhaulagiri what do you think?

alex-ju: If I'm being pedantic, the dropdown component didn't have the `close` functionality exposed before this change, hence we introduce a new feature in this component. However, if we think of this feature as something that was originally intended to be part of the component, but missed in implementation, we can see it as a fix/patch.
Happy to either of the approaches so I'll defer the decision to @Dhaulagiri.

Dhaulagiri: I could go either way 🤷  Technically it's adding new api vs "patching" something that already existed. Ultimately I don't think it matters _too_ much as consumers will need to update either way.

alex-ju: Having a look at previous releases made me stick with a patch, for consistency.

didoo (approved):

PR #406, opened 2022-06-24
Fix badge vertical alignment

alex-ju:
### :pushpin: Summary

Fix badge vertical alignment

When a mix of badges (with and without icon) are presented in a block element they are vertically misaligned.
In this PR we fix the alignment and add visuals to check how the component behaves with different `display` values on the parent element.

[Preview link](https://hds-components-git-alex-ju-fix-badge-vertical-6c42ca-hashicorp.vercel.app/components/badge#showcase)

Thanks @natmegs for [reporting the issue](https://hashicorp.slack.com/archives/CPB8GS9QT/p1656079573971139)! 🦸‍♀️ 

#### Before

badge-before

#### After

badge-after

***

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/eleven-ears-cover.md
+++ b/.changeset/eleven-ears-cover.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Fix badge vertical alignment
--- a/packages/components/tests/dummy/app/templates/components/badge.hbs
+++ b/packages/components/tests/dummy/app/templates/components/badge.hbs
@@ -102,6 +102,23 @@
       <Hds::Badge @icon="activity" @text="Lorem ipsum" /></p>
   </div>
 
+  <h5 class="dummy-h5">Containers</h5>
+  <div class="dummy-form-text-input-containers">
+    {{#let (array "block" "flex" "grid") as |displays|}}
+      {{#each displays as |display|}}
+        <div>
+          <span class="dummy-text-small">Parent with <code class="dummy-code">display: {{display}}</code></span>
+          <br />
+          <div class="dummy-form-text-input-containers__{{display}}">
+            <Hds::Badge @text="Only text" />
+            <Hds::Badge @icon="activity" @text="Text + icon" />
+            <Hds::Badge @icon="activity" @text="Only icon" @isIconOnly={{true}} />
+          </div>
+        </div>
+      {{/each}}
+    {{/let}}
+  </div>
+
   <h4 class="dummy-h4">Size</h4>
   <div class="dummy-badge-base-sample">
     {{#each @model.BADGE_SIZES as |size|}}
--- a/packages/components/tests/dummy/app/templates/components/badge.hbs
+++ b/packages/components/tests/dummy/app/templates/components/badge.hbs
@@ -103,16 +103,18 @@
   </div>
 
   <h5 class="dummy-h5">Containers</h5>
-  <div class="dummy-form-text-input-containers">
+  <div class="dummy-badge-containers">
     {{#let (array "block" "flex" "grid") as |displays|}}
       {{#each displays as |display|}}
         <div>
           <span class="dummy-text-small">Parent with <code class="dummy-code">display: {{display}}</code></span>
           <br />
-          <div class="dummy-form-text-input-containers__{{display}}">
-            <Hds::Badge @text="Only text" />
-            <Hds::Badge @icon="activity" @text="Text + icon" />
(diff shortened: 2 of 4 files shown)


.changeset/eleven-ears-cover.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
```
didoo, line 2: thinking out loud: Is this really just a patch? potentially is a visual change in the places where it's used. maybe we should use a "minor" bump here?

packages/components/tests/dummy/app/templates/components/badge.hbs
```diff
@@ -102,6 +102,23 @@
       <Hds::Badge @icon="activity" @text="Lorem ipsum" /></p>
   </div>
 
+  <h5 class="dummy-h5">Containers</h5>
+  <div class="dummy-form-text-input-containers">
```
didoo, line 106: always better to copy&paste these helper classes from one component page to another (in this case, the class `dummy-form-text-input-containers` is specific to the `text-input` doc page).

_Notice: once we will start to work on the "proper" website, we will also abstract these classes but for now copy&paste between the `db-[component].scss` files is OK_

packages/components/tests/dummy/app/templates/components/badge.hbs
```diff
@@ -102,6 +102,23 @@
...
+    {{#let (array "block" "flex" "grid") as |displays|}}
+      {{#each displays as |display|}}
+        <div>
+          <span class="dummy-text-small">Parent with <code class="dummy-code">display: {{display}}</code></span>
+          <br />
+          <div class="dummy-form-text-input-containers__{{display}}">
```
didoo, line 112: same here.

also, better to add some gap between the child elements, otherwise people may think there is some error in the implementation.

didoo (approved):
I am approving the changes, but I am not 100% convinced. 

Generally we delegate these layout concerns to the consumers (or we introduce a `Badge List` component for this purpose). I understand that adding a line with `vertical-align` solves the issue at hand, but opens the doors to future similar concerns (eg. with `Tags`). Plus, is a single-line away for the consumers too: just add `display: flex` to the parent. (In any case, since we're not handling spacing between badges, this needs to be handled at consumers-level anyway).

I'll leave the final call to you @alex-ju :)

packages/components/tests/dummy/app/templates/components/badge.hbs
alex-ju: I copied the markup without realising it was specific to the text-input component 😬 
Have updated with porting these classes to the `db-badge.scss` – and we'll abstract at a further point, as you said.

alex-ju:
> I am approving the changes, but I am not 100% convinced.
> 
> Generally we delegate these layout concerns to the consumers (or we introduce a `Badge List` component for this purpose). I understand that adding a line with `vertical-align` solves the issue at hand, but opens the doors to future similar concerns (eg. with `Tags`). Plus, is a single-line away for the consumers too: just add `display: flex` to the parent. (In any case, since we're not handling spacing between badges, this needs to be handled at consumers-level anyway).
> 
> I'll leave the final call to you @alex-ju :)

Thanks for taking a look at this @didoo, and for spotting the issues above. Interesting perspective on delegating the layout concerns to the consumers. I was under the (possibly wrong) impression that we're trying to make our components resilient to different layout types (based on the work we recently did on the form controls). I also appreciate this fix is far from perfect, but it's something minor we can do to improve how different types of badges align when in a block/inline-block container.

jesdavpet (approved):
Thanks for make the change for this possible solution to misalignment of the `<Hds::Badge>` component @alex-ju !!! 😄 🙌

I think this would make the component a lot easier to use in Cloud UI, without needing workarounds. ❤️

.changeset/eleven-ears-cover.md
alex-ju: @Dhaulagiri thoughts on this? sorry to drag in most versioning conversations, but I think we'll need a bit of guidance until we (or at least I) get to a common sense on these 😁

Dhaulagiri: > sorry to drag you in most versioning conversations

Not a problem as I feel like I'm often the ones starting them 😅 

Even though my experience is that there is an inherent subjectivity to these choices which makes it hard to create "rules", it does strike me that even a light framework we could reference might be useful. I'll add a task to our backlog.

PR #415, opened 2022-06-29
`Card::Container` component - Add extra `level` args for `hover` and `active` states

didoo:
### :pushpin: Summary

When we implemented the `Card::Container` component (long time ago) we didn't include the specifications for the user interaction with the card (`hover`, `active` and `focused` states) because at that time we considered it only as "container" (hence, the name).

Since then the component has been used in multiple places/products for interactive cards, requiring different visual treatment for the different interaction states. Last in time, [this request](https://hashicorp.slack.com/archives/C7KTUHNUS/p1656122163303119) by @DingoEatingFuzz. As the component is built today, the only way to achieve these states would be via JS (with `onMouseOver/Enter/Leave`), which is not great.

After discussion with the team, we have acknowledged that defining all the possible "correct" states would require a proper audit, and some work on the design side to come up with the correct specs, but at the moment we don't have capacity for this. So we agreed for an intermediate solution: adding two new arguments `@levelHover` and `@levelActive` that would allow the consumers to specify via component API the elevation "level" to use on `:hover` and `:active`, delegating to their product designers which level to use for the different states.

In this PR I have:
- added `@levelHover` and `@levelActive` arguments to `Card::Container` component
- updated the “Component API” and the “Showcase” documentation for the `Card::Container`
  - small changes to the doc file due to linting
- updated the integration tests for `Card::Container`

Preview: https://hds-components-git-card-add-extra-level-args-hashicorp.vercel.app/components/card

#### Things to discuss 

@heatherlarsen while technically this solution works, I _think_ it would be better to expose not just the three levels `base/mid/high` for the default level, but all the possible elevation styles, `base/low/mid/high/higher/[overlay?]` and let the consumers decide which one to use (we provide guidance, not control). This would allow them to have more space for different states combinations, while at the moment using the `Card::Container` as is is very limited. What do you think?

***

### 👀 How to review

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/card/container.js
+++ b/packages/components/addon/components/hds/card/container.js
@@ -5,8 +5,6 @@ export const DEFAULT_LEVEL = 'base';
 export const DEFAULT_BACKGROUND = 'neutral-primary';
 export const DEFAULT_OVERFLOW = 'hidden';
 export const LEVELS = ['base', 'mid', 'high'];
-export const HOVER_LEVELS = ['base', 'mid', 'high'];
-export const ACTIVE_LEVELS = ['base', 'mid', 'high'];
 export const BACKGROUNDS = ['neutral-primary', 'neutral-secondary'];
 export const OVERFLOWS = ['hidden', 'visible'];
 
@@ -44,10 +42,10 @@ export default class HdsCardContainerComponent extends Component {
 
     if (levelHover) {
       assert(
-        `@levelHover for "Hds::Card::Container" must be one of the following: ${HOVER_LEVELS.join(
+        `@levelHover for "Hds::Card::Container" must be one of the following: ${LEVELS.join(
           ', '
         )}, received: ${levelHover}`,
-        HOVER_LEVELS.includes(levelHover)
+        LEVELS.includes(levelHover)
       );
     }
 
@@ -66,10 +64,10 @@ export default class HdsCardContainerComponent extends Component {
 
     if (levelActive) {
       assert(
-        `@levelActive for "Hds::Card::Container" must be one of the following: ${ACTIVE_LEVELS.join(
+        `@levelActive for "Hds::Card::Container" must be one of the following: ${LEVELS.join(
           ', '
         )}, received: ${levelActive}`,
-        ACTIVE_LEVELS.includes(levelActive)
+        LEVELS.includes(levelActive)
       );
     }
 
--- a/packages/components/addon/components/hds/card/container.js
+++ b/packages/components/addon/components/hds/card/container.js
@@ -5,6 +5,8 @@ export const DEFAULT_LEVEL = 'base';
 export const DEFAULT_BACKGROUND = 'neutral-primary';
 export const DEFAULT_OVERFLOW = 'hidden';
 export const LEVELS = ['base', 'mid', 'high'];
+export const HOVER_LEVELS = ['base', 'mid', 'high'];
+export const ACTIVE_LEVELS = ['base', 'mid', 'high'];
 export const BACKGROUNDS = ['neutral-primary', 'neutral-secondary'];
 export const OVERFLOWS = ['hidden', 'visible'];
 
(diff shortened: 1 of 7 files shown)


packages/components/addon/components/hds/card/container.js
```diff
@@ -5,6 +5,8 @@ export const DEFAULT_LEVEL = 'base';
 export const DEFAULT_BACKGROUND = 'neutral-primary';
 export const DEFAULT_OVERFLOW = 'hidden';
 export const LEVELS = ['base', 'mid', 'high'];
+export const HOVER_LEVELS = ['base', 'mid', 'high'];
+export const ACTIVE_LEVELS = ['base', 'mid', 'high'];
```
alex-ju, line 9: What is the rationale for re-defining levels for each state? Is it so we can potentially restrict the levels we allow for each state? Or do we foresee different levels for different states?

packages/components/app/styles/components/card/container.scss
```diff
@@ -6,7 +6,10 @@
 //
 
 
+$hds-card-container-style: ( 'surface', 'elevation' );
 $hds-card-container-levels: ( 'base', 'mid', 'high' );
+$hds-card-container-hover-levels: ( 'base', 'mid', 'high' );
+$hds-card-container-active-levels: ( 'base', 'mid', 'high' );
```
alex-ju, line 12: Same question here about redefining the levels for each state.

packages/components/app/styles/components/card/container.scss
```diff
@@ -16,6 +19,27 @@ $hds-card-container-border-radius: 6px;
...
+      box-shadow: var(--token-#{$style}-#{$level}-box-shadow);
+    }
+  }
+  @each $level in $hds-card-container-active-levels {
+    .hds-card__container--active-level-#{$style}-#{$level}:active,
+    .hds-card__container--active-level-#{$style}-#{$level}.mock-active {
+      box-shadow: var(--token-#{$style}-#{$level}-box-shadow);
+    }
+  }
+}
```
alex-ju, line 42: Neatly packed!

alex-ju (approved):
Technically all good! I'm only curious about the choice to redefine the potential levels for each state.

packages/components/app/styles/components/card/container.scss
didoo: See above.

heatherlarsen:
@didoo I don’t recall us making a decision to limit the options to `base`, `mid`, and `high`. I'm fine leaving this open ended until we have time to do a more thorough audit.

heatherlarsen:
@didoo Just refreshed my memory on the elevation levels...the intention for limiting these to `base`, `mid`, and `high` is:
- to align to the design choice from the visual refresh of "Intentional elevation"
  - "We use shadow with containers to elevate and group key content on the page, and to convey natural page hierarchy and modality."
- to provide a more consistent card experience
- reserve other elevation levels for specific types of components, such as `modal`, `toasts`, form controls, etc 

At this time I don't think we should be adjusting the available elevation levels, but it's a possibility that when we do a more thorough audit, a need may arise to do so. 

Screen Shot 2022-06-29 at 8 14 01 AM

packages/components/tests/dummy/app/templates/components/card.hbs
```diff
@@ -129,4 +160,21 @@
       </div>
     </Hds::Card::Container>
   </div>
+
+  <h4 class="dummy-h4">Interaction:</h4>
```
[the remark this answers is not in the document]
didoo: I've updated the documentation, removing these examples and improving the "How to use" section: https://hds-components-git-card-add-extra-level-args-hashicorp.vercel.app/components/card#how-to-use

packages/components/app/styles/components/card/container.scss
didoo: As per @heatherlarsen feedback, for now we keep the same three levels for all of them, so I have unified the arrays in a single one.

packages/components/addon/components/hds/card/container.js
didoo: As per @heatherlarsen feedback, for now we keep the same three levels for all of them, so I have unified the arrays in a single one.

alex-ju (approved):
Likewise!

PR #421, opened 2022-07-01
Form controls - Bring back `group` variant for `Toggle`

didoo:
### :pushpin: Summary

After the initial implementation, we decided to remove the group variant for the Toggle component from the code implementation as well as the documentation.

Now, looking at the use cases described in the Figma file for this component, we have found that actually is used in the form of a “single item group” to emulate a label and a control with a secondary label.

[See Slack thread here for details](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1656583124340899).

### :hammer_and_wrench: Detailed description

In this PR I have:
- added back the `Toggle::Group` component
- added back the “Component API” and "Showcase" documentation for `Toggle::Group`
- added back the integration tests for `Toggle::Group`
- fixed a CSS declaration to include more use cases (Legend + Helper + Control, so Legend and Control are just siblings, not directly one after the other)

Preview: https://hds-components-git-form-controls-toggle-group-hashicorp.vercel.app/components/form/toggle

***

### 👀 How to review

👉 Review commit-by-commit or by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #425, opened 2022-07-04
Add `isRequired` and `isOptional` to form fields

alex-ju:
### :pushpin: Summary

Add `isRequired` and `isOptional` to form fields.

Following the recommendations, we add these options to the label/legend element and then we expose them via field/fieldset to form components.

This will allow components to be called as follows:

```hbs
<Hds::Form::TextInput::Field @isRequired={{true}} as |F|>
  <F.Label>This is the label text</F.Label>
</Hds::Form::TextInput::Field>
```

An alternative would be to preserve `isRequired` only at the label/legend level, removing the need for most of the code in this PR, but surfacing this argument at the field level brings it in-line with the `isInvalid` argument, so I guess it makes sense to have them aligned.
```hbs
<Hds::Form::TextInput::Field as |F|>
  <F.Label @isRequired={{true}}>This is the label text</F.Label>
</Hds::Form::TextInput::Field>
```

**Note**
I omitted the toggle component as I've not seen a reference in design and don't think it makes sense for the component to have it, but open to discussions. Similarly for checkbox/radio as a standalone field.

### :link: External links

[Jira issue](https://hashicorp.atlassian.net/jira/software/c/projects/HDS/boards/445?modal=detail&selectedIssue=HDS-137)

***

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ ] +1 Percy if applicable


alex-ju:
@heatherlarsen if you'd like to have a look and/or provide feedback at this stage (as you know, the form components are not yet accessible from the index page, but URLs are https://hds-components-git-alex-ju-form-fields-require-8a8dd0-hashicorp.vercel.app/components/form/text-input (change `text-input` with `textarea` or `select`)

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
```diff
@@ -50,6 +50,16 @@
...
+    <dt>isRequired <code>boolean</code></dt>
+    <dd>
+      <p>It appends a badge to the label text to indicate user input is required.</p>
+      <p>Default: <span class="default">false</span></p>
+    </dd>
+    <dt>isOptional <code>boolean</code></dt>
+    <dd>
+      <p>It appends "(Optional)" to the label text to indicate user input is not required.</p>
+      <p>Default: <span class="default">false</span></p>
+    </dd>
```
didoo, line 62: [nitpick] Probably I would generalize the text, so that we don't have to specify there is some kind of difference between`badge` and `("Optional")` (it's kinda an internal thing).  Something like this:
- It appends a [mark|indication] to the label text to indicate that the user input is required.
- It appends a [mark|indication] to the label text to indicate that the user input is optional.

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
```diff
@@ -406,6 +446,40 @@
...
+  {{! template-lint-disable no-unbalanced-curlies }}
+  <CodeBlock
+    @language="markup"
+    @code='
+      <Hds::Form::Label @controlId="control-ID" @isRequired=\{{true}}>My label</Hds::Form::Label>
+    '
+  />
+  {{! template-lint-enable no-unbalanced-curlies }}
+  {{! prettier-ignore-end }}
+  <p class="dummy-paragraph">Renders to:</p>
```
didoo, line 464: [suggestion] probably, now that we have the `Required` badge, I would update the example below this (the one with "Some badge") to have the badge as `highlight` color, so they are different and it doesn't generate confusion (see all the occurrences of "Some badge" in the page).

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
```diff
@@ -905,19 +1013,31 @@
...
+  <Hds::Form::Label @isOptional={{true}}>This is the label</Hds::Form::Label>
+  <br />
   <span class="dummy-text-small">With structured content (eg. a
     <code>flex</code>
     layout and a
     <code>&lt;Badge&gt;</code>)</span>
   <Hds::Form::Label>
     <div class="dummy-form-base-elements-label-with-badge">This is the label
-      <Hds::Badge @size="small" @text="Some badge" /></div>
+      <Hds::Badge @size="small" @text="Some badge" />
```
didoo, line 1028: [question/suggestion] do you think we should showcase an example of "With structured content + Required" and/or "With structured content + Optional" (probably with short text and with long text)? On one side would be nice so we can test it works as expected, but on the other I fear it would confuse the end consumers...

didoo:
> I omitted the toggle component as I've not seen a reference in design and don't think it makes sense for the component to have it

@alex-ju probably for the same reason for which we re-introduced the "group" variant for the toggle, and we showcase it as multiple toggles even if likely is not a real use case, I would be inclined to add it anyway (maybe a user need to turn on a "toggle" for some reasons, like accepting the risk of an operation, or explicitly approving something)

didoo (commented):
Overall code is OK for me. There are a few things I would like to discuss with you (and also with @heatherlarsen / @cveigt about the position of the "required/optional" indicators) before approving the PR.

packages/components/addon/components/hds/form/field/index.js
```diff
@@ -10,6 +10,8 @@ export const LAYOUT_TYPES = ['vertical', 'flag'];
 export default class HdsFormFieldIndexComponent extends Component {
   @tracked ariaDescribedBy = getAriaDescribedBy(this);
   @tracked descriptors = [];
+  @tracked isRequired = this.args.isRequired;
```
MelSumner, line 13: [question] I might be missing something here, but why is this tracked?

MelSumner, line 13: [issue] I don't see where this is also adding the `required` attribute to the input element if true.

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
```diff
@@ -50,6 +50,16 @@
         <code class="dummy-code">&lt;label&gt;</code>
         element.</p>
     </dd>
+    <dt>isRequired <code>boolean</code></dt>
+    <dd>
+      <p>It appends a badge to the label text to indicate user input is required.</p>
+      <p>Default: <span class="default">false</span></p>
```
MelSumner, line 56: [issue] if you are leaving this up to the user to declare, then the default would be `null` instead of `false`, yes?

didoo: Technically correct. But probably, to keep it simple, we can leave it as is? For the consumer what is important is to know is that it can be `true/false` (and in this case the default would be _falsy_, but if we use `null` or `falsy` may be confusing?)

didoo:
@heatherlarsen @cveigt + @alex-ju something I want to discuss with you, as a possible alternative to the current layout implementation of the "required/optional" indicators (currently inline with the text). Let me show you the problem. 

Let's take the default example:
screenshot_1612
this works fine, no problem here. But what if the text is slightly longer? well in some cases (it depends by the length of the text and the single words) this could happen:
screenshot_1613
Essentially, being an inline element, and there not being enough space after the text, the mark goes on a new line. Not only, it has a 4px margin from the left, because... that's what we declared to have, to avoid it touching the text when is next to a word.

Now, not a big problem per se, still works, but I want to suggest an alternative. To have the label text and the mark laid out via a `flex` container, so that this is the alternative behaviour when there is not enough space:
screenshot_1614
Of course, in case _there is_ enough space, the layout would be visually indistinguishable from the current one:
screenshot_1615

Essentially I am suggesting to go from this:
inline
to this:
flex

This, on the Figma side, would also solve the problem you discussed a few days ago, about the layout with "required/optional" not being responsive on resize, while with the proposed change this could be achieved with autolayout.

What do you think?

heatherlarsen:
@didoo While it's harder to accomplish in Figma (and would require a detach of the component), the indicator wrapping exactly to that point is such an edge case that I don't think we should be adjusting our whole approach around it. Personally, the inline approach feels like it provides a more natural reading/scanning experience.

Here is an example of a label from our products that would result in a very unnatural reading experience with the proposed change: 

Frame 3255


[question] Could you move the 4px margin to the text and not the badge so that the gap on the left is avoided?

packages/components/app/styles/components/badge.scss
```diff
@@ -144,3 +144,12 @@ $size-props: (
     }
   }
 }
+
+.hds-form-label,
+.hds-form-legend {
+  .hds-badge {
+    display: inline-block;
+    vertical-align: baseline;
+    margin-left: 4px;
```
heatherlarsen, line 153: [nitpick(ux)] font-weight is explicitly called out for the optional indicator, but for required it's taking on the font-weight from the label. The font-weight of the required badge should be `--token-typography-font-weight-medium`.

cveigt: I agree with Heather here. We may also have more indicators beyond required and optional with longer text, which would make the label even more challenging to read.

cveigt:
I agree with Heather here. We may also have more indicators beyond required and optional with longer text, which would make the label even more challenging to read.

packages/components/addon/components/hds/form/field/index.js
```diff
@@ -10,6 +10,8 @@ export const LAYOUT_TYPES = ['vertical', 'flag'];
...
+  @tracked isOptional = this.args.isOptional;
```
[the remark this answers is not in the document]
alex-ju: We don't need a setter on these values, but we at least need a getter, because of [the way we forward them to the label/legend component](https://github.com/hashicorp/design-system/pull/425/files#diff-61b99252eb7c03439d9402878520f0bfe18b8ecdf6e9d5c71188cdc0035ef168R9).

I can convert them from `@tracked` to only having a getter, e.g.
```
  get isRequired() {
    let { isRequired } = this.args;
    return isRequired;
  }
```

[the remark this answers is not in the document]
didoo: Mmm, ok now I see why (that was the part I missed: the fact that when you hash them, they need to be `this.something` and not `this.args.something`, while I was thinking "in React" where automatically a prop is automatically exposed)

Personally, I have a subtle preference for the getter, probably in an ever simplified way:
```
// I would add a comment here explaining why we need to do this strange thing
get isRequired() {
    return this.args.isRequired;
  }
```

didoo:
> @didoo While it's harder to accomplish in Figma (and would require a detach of the component), the indicator wrapping exactly to that point is such an edge case that I don't think we should be adjusting our whole approach around it. Personally, the inline approach feels like it provides a more natural reading/scanning experience.

FYI this is how it would work in a browser:
screenshot_1617
so the case you mentioned would be only in Figma.

> [question] Could you move the 4px margin to the text and not the badge so that the gap on the left is avoided?

@alex-ju I _think_ this would require to move the mark outside the `<label/legend>` and use a wrapper container, right? in that case do you see any impact from an accessibility standpoint? @MelSumner @alex-ju should the "required/optional" mark live _inside_ the `<label/legend>`, _outside_ of it, or it's irrelevant?

packages/components/app/styles/components/badge.scss
didoo: @heatherlarsen the problem is because we (I) never explicitly set the `font-weight` on the `Badge` component.
I've opened a PR to fix this: https://github.com/hashicorp/design-system/pull/426
/cc @alex-ju

packages/components/addon/components/hds/form/field/index.js
MelSumner: If we are providing a way for the author to say "isRequired" - and that provides a required badge, do you think it would be logical to also add the `required` attribute to the associated `input` element?

didoo: Since we've decided to leave responsibility to the consumers to handle most part of the HTML attributes for the controls, especially validation, I think we should follow the same approach to `required` here: if a field is required and the consumers handle if and how to validate that the field is "filled" by the user, if and how to show an error, then adding a `required` attribute would go against this approach, potentially may be unexpected too (we may have client-side validation interfering with the consumers-validation, at UI level).

packages/components/addon/components/hds/form/label/index.hbs
```diff
@@ -1,3 +1,5 @@
 <label class={{this.classNames}} for={{@controlId}} ...attributes>
   {{yield}}
+  {{#if @isRequired}}<Hds::Badge @size="small" @text="Required" />{{/if}}
```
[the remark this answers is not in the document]
alex-ju: My assumption was that we'd want the 'Required' badge to be a default badge, regardless of what that would look like. Have updated now to explicitly declare it as 'neutral'

packages/components/addon/components/hds/form/legend/index.hbs
```diff
@@ -1,3 +1,5 @@
 <legend class={{this.classNames}} ...attributes>
   {{yield}}
+  {{#if @isRequired}}<Hds::Badge @size="small" @text="Required" />{{/if}}
+  {{#if @isOptional}}<span class="hds-form-legend__optional-indicator">(Optional)</span>{{/if}}
```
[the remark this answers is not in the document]
alex-ju: Have abstracted these into `required-indicator` and `optional-indicator` within the `forms` directory.

packages/components/app/styles/components/badge.scss
```diff
@@ -144,3 +144,12 @@ $size-props: (
     }
   }
 }
+
+.hds-form-label,
+.hds-form-legend {
+  .hds-badge {
```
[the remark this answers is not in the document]
alex-ju: Agreed – have moved these styles into their own files as part of the abstraction suggested above.

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
alex-ju: Seeing Heather backed this suggestion I added a few more examples.

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
alex-ju: Updated with the suggestion – have thought about this, but hesitated as people may take it as a recommendation to use the a colored badge in label.

alex-ju:
> > I omitted the toggle component as I've not seen a reference in design and don't think it makes sense for the component to have it
> 
> @alex-ju probably for the same reason for which we re-introduced the "group" variant for the toggle, and we showcase it as multiple toggles even if likely is not a real use case, I would be inclined to add it anyway (maybe a user need to turn on a "toggle" for some reasons, like accepting the risk of an operation, or explicitly approving something)

@didoo yes, I can rebase to main and add it for the 'group' variant, but I'd like to check with Christian first so we keep in line with design.

@cveigt we're planning to introduce the `isRequired` and `isOptional` indicators at the field level for text-input, textarea and select and at the group level for the checkboxes and radio. what are your thoughts on these indicators for the toggle? shall we have them as an option, and if so at the field or group level? or both?

packages/components/addon/components/hds/form/field/index.js
alex-ju: Sure, I'll get ambassadors' view on this.

alex-ju:
@didoo, @MelSumner I think this is ready for another review. I added a few commits on top of what you reviewed already and rebased to `main` to get [the new additions to the Toggle component](https://github.com/hashicorp/design-system/pull/421).
- Swapped from `@tracked` to getters for `isRequired` and `isOptional`, setting a default of `false`
- Explicitly defined the 'Require' badge color
- Refactored to abstract optional/required indicators, as suggested, removing some styling duplication
- Generalised parameter description then aligned it with test names
- Updated the structured content examples and added more examples for indicators
- Set the `required` attribute on form controls when `isRequired` is `true`
- Add `isRequired` and `isOptional` to `Form::Toggle` to bring it in line with the checkbox component.

packages/components/tests/dummy/app/templates/components/form/checkbox.hbs
```diff
@@ -183,6 +183,18 @@
         <li>horizontal</li>
       </ol>
     </dd>
+    <dt>isRequired <code>boolean</code></dt>
+    <dd>
+      <p>Appends an indicator to the legend text and sets the
+        <code class="dummy-code">required</code>
+        attribute when user input is required.</p>
```
didoo, line 190: I had to read the text multiple times to understand the role of the attribute. Maybe being mode specific makes it more obvious what happens behind the scenes, what receives that attribute.

```suggestion
        attribute on the controls when user input is required.</p>
```

packages/components/tests/dummy/app/templates/components/form/radio.hbs
```diff
@@ -183,6 +183,18 @@
         <li>horizontal</li>
       </ol>
     </dd>
+    <dt>isRequired <code>boolean</code></dt>
+    <dd>
+      <p>Appends an indicator to the legend text and sets the
+        <code class="dummy-code">required</code>
+        attribute when user input is required.</p>
```
didoo, line 190:
```suggestion
        attribute on the controls when user input is required.</p>
```

packages/components/tests/dummy/app/templates/components/form/select.hbs
```diff
@@ -111,6 +111,18 @@
       <p><em>Notice: by default the ID is automatically generated by the component; use this argument if you need to
           pass a custom ID for specific reasons you may have.</em></p>
     </dd>
+    <dt>isRequired <code>boolean</code></dt>
+    <dd>
+      <p>Appends an indicator to the label text and sets the
+        <code class="dummy-code">required</code>
+        attribute when user input is required.</p>
```
didoo, line 118:
```suggestion
        attribute on the control when user input is required.</p>
```

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
```diff
@@ -129,6 +129,18 @@
       <p>It applies an "invalid" appearance to the control (<em>notice: this does _not_ modify its logical validity</em>).</p>
       <p>Default: <span class="default">false</span></p>
     </dd>
+    <dt>isRequired <code>boolean</code></dt>
+    <dd>
+      <p>Appends an indicator to the label text and sets the
+        <code class="dummy-code">required</code>
+        attribute when user input is required.</p>
```
didoo, line 136:
```suggestion
        attribute on the control when user input is required.</p>
```

packages/components/tests/dummy/app/templates/components/form/textarea.hbs
```diff
@@ -100,6 +100,18 @@
       <p><em>Notice: by default the ID is automatically generated by the component; use this argument if you need to
           pass a custom ID for specific reasons you may have.</em></p>
     </dd>
+    <dt>isRequired <code>boolean</code></dt>
+    <dd>
+      <p>Appends an indicator to the label text and sets the
+        <code class="dummy-code">required</code>
+        attribute when user input is required.</p>
```
didoo, line 107:
```suggestion
        attribute on the control when user input is required.</p>
```

packages/components/tests/dummy/app/templates/components/form/toggle.hbs
```diff
@@ -177,6 +177,18 @@
         <li>horizontal</li>
       </ol>
     </dd>
+    <dt>isRequired <code>boolean</code></dt>
+    <dd>
+      <p>Appends an indicator to the legend text and sets the
+        <code class="dummy-code">required</code>
+        attribute when user input is required.</p>
```
didoo, line 184:
```suggestion
        attribute on the controls when user input is required.</p>
```

packages/components/tests/integration/components/hds/form/checkbox/group-test.js
```diff
@@ -66,6 +66,36 @@ module('Integration | Component | hds/form/checkbox/group', function (hooks) {
...
+      hbs`<Hds::Form::Checkbox::Group @isRequired={{true}} as |G|>
+            <G.Legend>This is the legend</G.Legend>
+            <G.Checkbox::Field checked="checked" @value="abc123" as |F|>
+              <F.Label>This is the control label</F.Label>
+            </G.Checkbox::Field>
+          </Hds::Form::Checkbox::Group>`
+    );
+    assert.dom('legend .hds-form-required-indicator').exists();
+    assert.dom('legend .hds-form-required-indicator').hasText('Required');
+    assert.dom('input[required]').exists();
```
didoo, line 83: maybe I would use something more explicit about what we're testing
```suggestion
    assert.dom('input').hasAttribute('required');
```

packages/components/tests/integration/components/hds/form/radio/group-test.js
```diff
@@ -66,6 +66,36 @@ module('Integration | Component | hds/form/radio/group', function (hooks) {
...
+      hbs`<Hds::Form::Radio::Group @isRequired={{true}} as |G|>
+            <G.Legend>This is the legend</G.Legend>
+            <G.Radio::Field checked="checked" @value="abc123" as |F|>
+              <F.Label>This is the control label</F.Label>
+            </G.Radio::Field>
+        </Hds::Form::Radio::Group>`
+    );
+    assert.dom('legend .hds-form-required-indicator').exists();
+    assert.dom('legend .hds-form-required-indicator').hasText('Required');
+    assert.dom('input[required]').exists();
```
didoo, line 83:
```suggestion
    assert.dom('input').hasAttribute('required');
```

packages/components/tests/integration/components/hds/form/select/field-test.js
```diff
@@ -94,6 +94,30 @@ module('Integration | Component | hds/form/select/field', function (hooks) {
...
+  test('it should append an indicator to the label text and set the required attribute when user input is required', async function (assert) {
+    assert.expect(3);
+    await render(
+      hbs`<Hds::Form::Select::Field @isRequired={{true}} as |F|>
+            <F.Label>This is the label</F.Label>
+          </Hds::Form::Select::Field>`
+    );
+    assert.dom('label .hds-form-required-indicator').exists();
+    assert.dom('label .hds-form-required-indicator').hasText('Required');
+    assert.dom('select[required]').exists();
```
didoo, line 108:
```suggestion
    assert.dom('select').hasAttribute('required');
```

packages/components/tests/integration/components/hds/form/text-input/field-test.js
```diff
@@ -101,6 +101,30 @@ module('Integration | Component | hds/form/text-input/field', function (hooks) {
...
+  test('it should append an indicator to the label text and set the required attribute when user input is required', async function (assert) {
+    assert.expect(3);
+    await render(
+      hbs`<Hds::Form::TextInput::Field @isRequired={{true}} as |F|>
+            <F.Label>This is the label</F.Label>
+          </Hds::Form::TextInput::Field>`
+    );
+    assert.dom('label .hds-form-required-indicator').exists();
+    assert.dom('label .hds-form-required-indicator').hasText('Required');
+    assert.dom('input[required]').exists();
```
didoo, line 115:
```suggestion
    assert.dom('input').hasAttribute('required');
```

packages/components/tests/integration/components/hds/form/textarea/field-test.js
```diff
@@ -82,6 +82,31 @@ module('Integration | Component | hds/form/textarea/field', function (hooks) {
...
+  test('it should append an indicator to the label text and set the required attribute when user input is required', async function (assert) {
+    assert.expect(3);
+    await render(
+      hbs`<Hds::Form::Textarea::Field @isRequired={{true}} as |F|>
+            <F.Label>This is the label</F.Label>
+          </Hds::Form::Textarea::Field>`
+    );
+    assert.dom('label .hds-form-required-indicator').exists();
+    assert.dom('label .hds-form-required-indicator').hasText('Required');
+    assert.dom('textarea[required]').exists();
```
didoo, line 97:
```suggestion
    assert.dom('textarea').hasAttribute('required');
```

packages/components/tests/integration/components/hds/form/toggle/group-test.js
```diff
@@ -66,6 +66,36 @@ module('Integration | Component | hds/form/toggle/group', function (hooks) {
...
+      hbs`<Hds::Form::Toggle::Group @isRequired={{true}} as |G|>
+            <G.Legend>This is the legend</G.Legend>
+            <G.Toggle::Field checked="checked" @value="abc123" as |F|>
+              <F.Label>This is the control label</F.Label>
+            </G.Toggle::Field>
+          </Hds::Form::Toggle::Group>`
+    );
+    assert.dom('legend .hds-form-required-indicator').exists();
+    assert.dom('legend .hds-form-required-indicator').hasText('Required');
+    assert.dom('input[required]').exists();
```
didoo, line 83:
```suggestion
    assert.dom('input').hasAttribute('required');
```

packages/components/app/styles/components/form/optional-indicator.scss
```diff
@@ -0,0 +1,10 @@
+//
+// FORM > OPTIONAL INDICATOR
+//
+// properties within each class are sorted alphabetically
+//
+
+.hds-form-optional-indicator {
+  color: var(--token-color-foreground-faint);
+  font-weight: var(--token-typography-font-weight-regular);
+}
```
didoo, line 10: I _think_ we've lost the styling that sets the sizing of the element (probably just the `line-height`? I remember I checked the size, but maybe I misremember).

Reason to have it, is that in some conditions the alignment may appear off:
screenshot_1621

didoo (commented):
Nothing outstanding, I've left a bunch of comments, but I would also discuss a few things in person before approving:
- checkbox/radio/toggle fields: should they have `@isRequired/@isOptional` fields for consistency (and potentially some use cases, especially checkbox and toggle?
- `@isRequired` as extra element also at "control:base" level? (99% no, but want to hear your thoughts on this)
- merging the "required" and "optional" components in a single one: pros vs cons

I'll set up a meeting for later (or tomorrow)

cveigt:
Small detail: There is a slight misalignment between the label and the badge.
This is probably not the solution, but I was able to "fix it" by removing `vertical-align` from the badge.

Screen Shot 2022-07-07 at 1 07 05 PM

Screen Shot 2022-07-07 at 1 11 27 PM

alex-ju:
> This is probably not the solution, but I was able to "fix it" by removing `vertical-align` from the badge.

@cveigt yes, there is a misalignment, indeed, and removing the `vertical-align` property is the solution for this case (a simple badge in a label), but it will [break the alignment for badges with icons](https://github.com/hashicorp/design-system/pull/406) (including the case where a badge with icon is placed in a label) 😭  – I'll give this some more thought.

packages/components/app/styles/components/form/optional-indicator.scss
```diff
@@ -6,5 +6,7 @@
 
 .hds-form-optional-indicator {
   color: var(--token-color-foreground-faint);
+  line-height: 1.23077; // 16px
+  font-size: 0.8125rem; // 13px
```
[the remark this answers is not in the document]
alex-ju: Makes sense, will update – I borrowed the ones from the badge, thinking we want to align the two.

cveigt:
@alex-ju back to the required/optional indicators. We reviewed this today and will have them at the group level, but not the field, which I think is how you have them now.

alex-ju:
@MelSumner I hope I addressed all your comments on this PR as well – would you like to have another look?

alex-ju:
I'll carry on and merge this to avoid potential conflicts in other branches, but let me know if anything and I'll open follow-up PRs

PR #426, opened 2022-07-06
Badge component - Add font weight declaration

didoo:
### :pushpin: Summary

Following this comment by @heatherlarsen https://github.com/hashicorp/design-system/pull/425#discussion_r914230335 I've realized we never explicitly set the `font-weight` property for the text in the `Badge` component. This has two problems: 1) it inherits from the container, as Heather noticed, and 2) it was using the wrong weight, it should be "medium".

In this PR I have:
- added font-weight “medium” to the text in the `Badge` component

***

### 👀 How to review

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/mean-weeks-guess.md
+++ b/.changeset/mean-weeks-guess.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Updated font-weight to "medium" for "Badge" and "BadgeCount" components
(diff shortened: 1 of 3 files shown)


alex-ju (approved):
Makes sense! I'm not sure about the value of [redefining this property for the required badge](https://github.com/hashicorp/design-system/pull/425/files#diff-579dd1c81d1e3941513eba1ef2417d870f32bbc99683c737d8e48e9240c21414R9) in this case.

didoo:
> Makes sense! I'm not sure about the value of [redefining this property for the required badge](https://github.com/hashicorp/design-system/pull/425/files#diff-579dd1c81d1e3941513eba1ef2417d870f32bbc99683c737d8e48e9240c21414R9) in this case.

Yes, we shouldn't in that case.

heatherlarsen (changes requested):
This change also needs to be made for `BadgeCount`

heatherlarsen (approved):
lgtm! thanks @didoo 🙂

PR #427, opened 2022-07-06
Updates to showcase section for accessibility

didoo:
### :pushpin: Summary

In a follow-up meeting with @MelSumner re. this discussion https://github.com/hashicorp/design-system/pull/385#discussion_r902610004 we agreed that a good reasonable pragmatic compromise 🙂 would be the following:
- clearly indicate under every "showcase" that the examples under that section are not to be used as a reference from the accessibility, because they're meant to be visual-only examples
-  add a `aria-hidden="true"` so @MelSumner doesn't get impacted when testing the page for accessibility issues
- to avoid making these changes to for every page, introduce a "section" dummy component that can be used for all the showcase sections, to centralize these choices in a single place (in case we need to change/update things in the future)

In this PR I have:
- added a `DummySection::Accessibility` component, containing the banner message about the “showcase” section potentially not being accessible/conformant
- updated CSS to hide from Percy the newly added “banner” to the “showcase” section, so all the existing tests remain intact
- replaced standard showcase `<sections>` with new `<DummySection::Showcase>` component

***

### 👀 How to review

### DIFF
--- a/packages/components/tests/dummy/app/components/dummy-section/showcase.hbs
+++ b/packages/components/tests/dummy/app/components/dummy-section/showcase.hbs
@@ -1,0 +1,11 @@
+<section data-test-percy aria-hidden="true">
+  <h3 class="dummy-h3" id={{or @id "showcase"}}><a href="#{{or @id 'showcase'}}" class="dummy-link-section">§</a>
+    {{or @title "Showcase"}}</h3>
+  <div class="dummy-banner dummy-banner--info" data-test-section-showcase-not-accessible>
+    <p class="dummy-paragraph"><strong>Notice</strong>: if you inspect the source code for this section, you may see
+      that some of the following examples are not necesssarily full compliant from an accessibility perspective. That's
+      because they're meant to be visual-only examples. If you have any doubt or question about it, please contact the
+      HDS team.</p>
+  </div>
+  {{yield}}
+</section>
--- a/packages/components/tests/dummy/app/components/dummy-section/showcase.hbs
+++ b/packages/components/tests/dummy/app/components/dummy-section/showcase.hbs
@@ -3,7 +3,7 @@
     {{or @title "Showcase"}}</h3>
   <div class="dummy-banner dummy-banner--info" data-test-section-showcase-not-accessible>
     <p class="dummy-paragraph"><strong>Notice</strong>: if you inspect the source code for this section, you may see
-      that some of the following examples are not necesssarily full compliant from an accessibility perspective. That's
+      that some of the following examples are not necessarily fully compliant from an accessibility perspective. That's
       because they're meant to be visual-only examples. If you have any doubt or question about it, please contact the
       HDS team.</p>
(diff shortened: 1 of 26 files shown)


packages/components/tests/dummy/app/components/dummy-section/showcase.hbs
```diff
@@ -0,0 +1,11 @@
+<section data-test-percy aria-hidden="true">
+  <h3 class="dummy-h3" id={{or @id "showcase"}}><a href="#{{or @id 'showcase'}}" class="dummy-link-section">§</a>
+    {{or @title "Showcase"}}</h3>
+  <div class="dummy-banner dummy-banner--info" data-test-section-showcase-not-accessible>
+    <p class="dummy-paragraph"><strong>Notice</strong>: if you inspect the source code for this section, you may see
+      that some of the following examples are not necesssarily full compliant from an accessibility perspective. That's
```
Dhaulagiri, line 6:
```suggestion
      that some of the following examples are not necessarily fully compliant from an accessibility perspective. That's
```

packages/components/tests/dummy/app/components/dummy-section/showcase.hbs
```diff
@@ -0,0 +1,11 @@
+<section data-test-percy aria-hidden="true">
+  <h3 class="dummy-h3" id={{or @id "showcase"}}><a href="#{{or @id 'showcase'}}" class="dummy-link-section">§</a>
```
MelSumner, line 2: [issue] I think we should only hide the content for this section, because it's a [failure to hide an interactive element](https://accessibilityinsights.io/info-examples/web/aria-hidden-focus/).

packages/components/tests/dummy/app/components/dummy-section/showcase.hbs
```diff
@@ -0,0 +1,11 @@
...
+    {{or @title "Showcase"}}</h3>
+  <div class="dummy-banner dummy-banner--info" data-test-section-showcase-not-accessible>
+    <p class="dummy-paragraph"><strong>Notice</strong>: if you inspect the source code for this section, you may see
+      that some of the following examples are not necesssarily full compliant from an accessibility perspective. That's
+      because they're meant to be visual-only examples. If you have any doubt or question about it, please contact the
+      HDS team.</p>
```
MelSumner, line 8:
```suggestion
    <p class="dummy-paragraph"><strong>Notice</strong>: The elements in this section are intended to provide samples for our visual regression testing platform. As such, some showcase elements are not fully conformant with WCAG 2.1 AA Success Criteria.  If you have any questions or concerns, please contact the HDS team.</p>
```

packages/components/tests/dummy/app/components/dummy-section/showcase.hbs
didoo: But what if there are interactive elements inside the content of the section? How would we solve it?

PR #431, opened 2022-07-06
Updated a11y docs for forms/base-elements

MelSumner:
### :pushpin: Summary

If merged, this PR updates the accessibility documentation for the forms/base-elements page.

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
+++ b/packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
@@ -735,7 +735,11 @@
 <section>
   <h3 class="dummy-h3" id="accessibility"><a href="#accessibility" class="dummy-link-section">§</a> Accessibility</h3>
   <p class="dummy-paragraph">Since these are the base elements, they are conditionally conformant; that is, they are not
-    conformant until used with the elements that will make them conformant.</p>
+    conformant until used in conjunction with the other components/elements that will make them conformant.</p>
+  <h4 class="dummy-h4">Known Issues</h4>
+  <p class="dummy-paragraph">If a link is used within a label, helper text, or error text, it will not be presented as a
+    link to the user with a screen reader; only the text content is read out. As such, care should be used when
+    considering this feature.</p>
   <h4 class="dummy-h4">
     Applicable WCAG Success Criteria (Reference)
   </h4>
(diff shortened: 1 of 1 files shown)


packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
```diff
@@ -735,7 +735,11 @@
 <section>
   <h3 class="dummy-h3" id="accessibility"><a href="#accessibility" class="dummy-link-section">§</a> Accessibility</h3>
   <p class="dummy-paragraph">Since these are the base elements, they are conditionally conformant; that is, they are not
-    conformant until used with the elements that will make them conformant.</p>
+    conformant until used in conjunction with the other components/elements that will make them conformant.</p>
+  <h4 class="dummy-h4">Known Issues</h4>
+  <p class="dummy-paragraph">If a link is used within a label, helper text, or error text, it will not be presented as a
+    link to the user with a screen reader; only the text content is read out. As such, care should be used when
+    considering this feature.</p>
```
didoo, line 742: When we spoke about this, you told us the problem was with the `<label>`, not the other elements (`helper text` and `error`). Why in this case a link would not be read out to the user, if it's inside a `<p>` or `<div>` element?

(also, if all of them have the same issue, at this point we should remove the paragraph in "how to use" about not adding links to the `<label>` and let the consumers decide where to put the link, since there's no difference in the outcome).

MelSumner: Links in the help text and the error text are also not identified as links (just the text is read) because it's associated to the input via `aria-describedby`.  

For labels, I still recommend the approach that was explored in https://codepen.io/melsumner/pen/NWyLMOQ. 

For help text and error text, the link _can_ be found through other navigation methods, but it won't be immediately or easily obvious that there's a link in the help and error text, so authors _should_ be aware of this.

didoo: Discussed, ok as it is. If we will have a more general `Form` documentation page, we will try to move the common parts, this included, in that page.

@MelSumner please copy&paste :) the same "known issues" in the other form controls pages, so it's documented in all of them

PR #432, opened 2022-07-07
Set default number of rows on textarea component

alex-ju:
### :pushpin: Summary

Set default number of rows on textarea component

As requested by design, the default value for `rows` on textarea would be 4 – have updated `Form::Textarea::Base` and `Form::Textarea::Field`.

### :link: External links

[Jira issue](https://hashicorp.atlassian.net/browse/HDS-246?focusedCommentId=83720)

***

### DIFF
--- a/packages/components/addon/components/hds/form/textarea/base.hbs
+++ b/packages/components/addon/components/hds/form/textarea/base.hbs
@@ -1,2 +1,2 @@
 {{! Notice: this is not the native HTML <textarea> but the Ember component <Textarea> }}
-<Textarea class={{this.classNames}} {{style width=@width}} ...attributes @value={{@value}} />
+<Textarea class={{this.classNames}} {{style width=@width}} rows="4" ...attributes @value={{@value}} />
--- a/packages/components/tests/integration/components/hds/form/textarea/base-test.js
+++ b/packages/components/tests/integration/components/hds/form/textarea/base-test.js
@@ -24,6 +24,17 @@ module('Integration | Component | hds/form/textarea/base', function (hooks) {
     assert.dom('#test-form-textarea').hasValue('abc123');
   });
 
+  // ROWS
+
+  test('it should render the textarea with the default number of rows', async function (assert) {
+    await render(hbs`<Hds::Form::Textarea::Base />`);
+    assert.dom('textarea').hasAttribute('rows', '4');
+  });
+  test('it should render the textarea with a custom number of rows', async function (assert) {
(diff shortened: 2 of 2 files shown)


packages/components/addon/components/hds/form/textarea/base.hbs
```diff
@@ -1,2 +1,2 @@
 {{! Notice: this is not the native HTML <textarea> but the Ember component <Textarea> }}
-<Textarea class={{this.classNames}} {{style width=@width}} ...attributes @value={{@value}} />
\ No newline at end of file
+<Textarea class={{this.classNames}} {{style width=@width}} rows="{{this.rows}}" ...attributes @value={{@value}} />
```
didoo, line 2: I think we can keep it much simpler! Just do this:
```suggestion
<Textarea class={{this.classNames}} {{style width=@width}} rows="4" ...attributes @value={{@value}} />
```
and you can remove all the code below. If a user pass a `rows` attribute, the `...attributes` spread will override the default.

didoo (changes requested):
See my comment. It can be much much simpler.

packages/components/tests/integration/components/hds/form/textarea/base-test.js
```diff
@@ -24,6 +24,17 @@ module('Integration | Component | hds/form/textarea/base', function (hooks) {
     assert.dom('#test-form-textarea').hasValue('abc123');
   });
 
+  // ROWS
+
+  test('it should render the textarea with the default number of rows', async function (assert) {
+    await render(hbs`<Hds::Form::Textarea::Base />`);
+    assert.dom('textarea[rows="4"]').exists();
```
didoo, line 31: maybe I would use something more explicit about what we're testing
```suggestion
    assert.dom('textarea').hasAttribute('rows', '4');
```

packages/components/tests/integration/components/hds/form/textarea/base-test.js
```diff
@@ -24,6 +24,17 @@ module('Integration | Component | hds/form/textarea/base', function (hooks) {
...
+  });
+  test('it should render the textarea with a custom number of rows', async function (assert) {
+    await render(hbs`<Hds::Form::Textarea::Base rows="2" />`);
+    assert.dom('textarea[rows="2"]').exists();
```
didoo, line 35: maybe I would use something more explicit about what we're testing
```suggestion
    assert.dom('textarea').hasAttribute('rows', '2');
```

heatherlarsen (approved):
lgtm! thanks for getting this in so quickly 🙂

PR #443, opened 2022-07-12
Add Tag component

alex-ju:
### :pushpin: Summary

Add `Hds::Tag` component based on [current design](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/6UTRhciO3hMNBgJyK3aXuY/HDS-Product---Components?node-id=12899%3A33230).

[👉 Preview Tag component](https://hds-components-git-alex-ju-add-tag-component-hashicorp.vercel.app/components/tag)

We're introducing a Tag component (for categorizing or filtering content) comprising of:
 - a dismiss button (displayed when an `onDismiss` function is provided)
 - a text; when `@href` or `@route` is provided we use the `Hds::Interactive` utility component, benefitting from all the logic behind it, otherwise we render a plain text

**Note**
~~We are currently discussing approaches for when the text is long (considering how to handle overflowing text), so we expect to add additional code to handle these scenarios.~~
Tried truncation, but is not conformant, so we'll fall back with text wrapping.

### :link: External links

[Jira issue](https://hashicorp.atlassian.net/browse/HDS-121?atlOrigin=eyJpIjoiNjcwMmU5OGVmMTE3NGEzYWE0NDBmZDgwNDRmYWE3ZjIiLCJwIjoiaiJ9)

***

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/tag/index.hbs
+++ b/packages/components/addon/components/hds/tag/index.hbs
@@ -1,7 +1,7 @@
 <span class="hds-tag" ...attributes>
   {{#if this.onDismiss}}
     <button class="hds-tag__dismiss" type="button" aria-label="Dismiss {{this.text}}" {{on "click" this.onDismiss}}>
-      <FlightIcon @name="x" @size="16" @isInlineBlock={{false}} />
+      <FlightIcon class="hds-tag__dismiss-icon" @name="x" @size="16" @isInlineBlock={{false}} />
     </button>
   {{/if}}
   {{#if (or @href @route)}}
--- a/packages/components/addon/components/hds/tag/index.hbs
+++ b/packages/components/addon/components/hds/tag/index.hbs
@@ -15,13 +15,11 @@
       @isRouteExternal={{@isRouteExternal}}
       @href={{@href}}
       @isHrefExternal={{@isHrefExternal}}
-      title="{{this.text}}"
-      {{style maxWidth=@maxWidth}}
     >
       {{this.text}}
     </Hds::Interactive>
   {{else}}
-    <span class="hds-tag__text" title="{{this.text}}" {{style maxWidth=@maxWidth}}>
+    <span class="hds-tag__text">
       {{this.text}}
     </span>
   {{/if}}
--- a/packages/components/addon/components/hds/tag/index.hbs
+++ b/packages/components/addon/components/hds/tag/index.hbs
@@ -15,11 +15,13 @@
       @isRouteExternal={{@isRouteExternal}}
       @href={{@href}}
       @isHrefExternal={{@isHrefExternal}}
+      title="{{this.text}}"
+      {{style maxWidth=@maxWidth}}
     >
       {{this.text}}
     </Hds::Interactive>
   {{else}}
-    <span class="hds-tag__text">
+    <span class="hds-tag__text" title="{{this.text}}" {{style maxWidth=@maxWidth}}>
       {{this.text}}
     </span>
   {{/if}}
--- a/packages/components/addon/components/hds/tag/index.hbs
+++ b/packages/components/addon/components/hds/tag/index.hbs
@@ -1,4 +1,4 @@
-<span class="hds-tag" ...attributes>
+<span class={{this.classNames}} ...attributes>
   {{#if this.onDismiss}}
     <button class="hds-tag__dismiss" type="button" aria-label="Dismiss {{this.text}}" {{on "click" this.onDismiss}}>
       <FlightIcon class="hds-tag__dismiss-icon" @name="x" @size="16" @isInlineBlock={{false}} />
--- a/packages/components/addon/components/hds/tag/index.hbs
+++ b/packages/components/addon/components/hds/tag/index.hbs
@@ -1,0 +1,26 @@
+<span class="hds-tag" ...attributes>
+  {{#if this.onDismiss}}
+    <button class="hds-tag__dismiss" type="button" aria-label="Dismiss {{this.text}}" {{on "click" this.onDismiss}}>
+      <FlightIcon @name="x" @size="16" @isInlineBlock={{false}} />
+    </button>
+  {{/if}}
+  {{#if (or @href @route)}}
+    <Hds::Interactive
+      class="hds-tag__link"
+      @current-when={{@current-when}}
+      @models={{hds-link-to-models @model @models}}
+      @query={{hds-link-to-query @query}}
+      @replace={{@replace}}
+      @route={{@route}}
+      @isRouteExternal={{@isRouteExternal}}
+      @href={{@href}}
+      @isHrefExternal={{@isHrefExternal}}
+    >
+      {{this.text}}
+    </Hds::Interactive>
+  {{else}}
+    <span class="hds-tag__text">
+      {{this.text}}
+    </span>
+  {{/if}}
+</span>
--- a/packages/components/app/styles/components/tag.scss
+++ b/packages/components/app/styles/components/tag.scss
@@ -5,7 +5,7 @@
 //
 //
 
-$hds-tag-border-radius: 20px;
+$hds-tag-border-radius: 13px;
 
 .hds-tag {
   align-items: stretch;
--- a/packages/components/app/styles/components/tag.scss
+++ b/packages/components/app/styles/components/tag.scss
@@ -6,6 +6,7 @@
 //
 
 $hds-tag-border-radius: 13px;
+$hds-tag-text-max-width: 120px;
 
 .hds-tag {
   align-items: stretch;
@@ -40,6 +41,10 @@ $hds-tag-border-radius: 13px;
   border-radius: inherit;
   flex: 1 0 0;
   padding: 4px 8px;
+  max-width: $hds-tag-text-max-width;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
 }
 
 .hds-tag__link {
@@ -47,6 +52,10 @@ $hds-tag-border-radius: 13px;
   color: var(--token-color-foreground-action);
   flex: 1 0 0;
   padding: 4px 8px;
+  max-width: $hds-tag-text-max-width;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
 
   &:hover,
   &.mock-hover {
--- a/packages/components/app/styles/components/tag.scss
+++ b/packages/components/app/styles/components/tag.scss
@@ -5,6 +5,8 @@
 //
 //
 
+@use "../mixins/focus-ring" as *;
+
 $hds-tag-border-radius: 13px;
 
 .hds-tag {
@@ -28,7 +30,7 @@ $hds-tag-border-radius: 13px;
   border-bottom-right-radius: 0;
   border: none; // reset default button border
   margin: 0; // reset default button margin
-  padding: 6px 4px 6px 6px;
+  padding: 6px 2px 6px 8px;
 }
 
 .hds-tag__dismiss-icon {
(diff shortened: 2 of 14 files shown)


packages/components/app/styles/components/tag.scss
```diff
@@ -0,0 +1,106 @@
...
+}
+
+.hds-tag__dismiss ~ .hds-tag__text,
+.hds-tag__dismiss ~ .hds-tag__link {
+  border-radius: 0 $hds-tag-border-radius $hds-tag-border-radius 0;
+  padding-left: 4px;
+}
+
+.hds-tag__dismiss,
+.hds-tag__link {
```
didoo, line 66: Maybe we can find a way to avoid the duplication with similar selectors, assigning to the icon a color (with higher specificity)?

packages/components/app/styles/components/tag.scss
```diff
@@ -0,0 +1,106 @@
...
+
+  &:focus {
+    box-shadow: 0px 0px 0px 1px var(--token-color-focus-action-internal), 0px 0px 0px 4px var(--token-color-focus-action-external);
+    outline-color: transparent; // prevent default outline as we use box-shadow for custom styling
+    outline-style: solid;
+    z-index: 1; // ensures focus is not obscured by adjacent elements
+  }
+}
+
+// mock classes are applied at the component level
```
didoo, line 86: I _think_ there is a way to apply the mock to both the elements, the "dismiss" and the "link (I have to check in the code implemented to handle the mocks).

didoo (commented):
added a bunch of initial comments, but overall seems very good, and already in very good shape. 🚀

packages/components/app/styles/components/tag.scss
didoo: I tried, but adds a some hidden/implicit knowledge so I am not sure is a good idea. Let's keep it as is for now.

packages/components/app/styles/components/tag.scss
```diff
@@ -0,0 +1,106 @@
...
+}
+
+.hds-tag__dismiss {
+  flex: 0 0 auto;
+  border-radius: $hds-tag-border-radius 0 0 $hds-tag-border-radius;
+  border: none; // reset default button border
+  margin: 0; // reset default button margin
+  padding: 6px 4px 6px 6px;
+
+  .flight-icon {
```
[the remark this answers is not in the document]
alex-ju: Agree – addressed in [57cb8c1](https://github.com/hashicorp/design-system/pull/443/commits/57cb8c16f22c873d0ca7554fd7539eb765b13515)

packages/components/app/styles/components/tag.scss
```diff
@@ -0,0 +1,106 @@
...
+    height: 12px;
+    width: 12px;
+  }
+}
+
+.hds-tag__text {
+  border-radius: $hds-tag-border-radius;
```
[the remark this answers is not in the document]
alex-ju: Bundled the `border-radius` suggestions in [46c5138](https://github.com/hashicorp/design-system/pull/443/commits/46c513862d8c6223b4f16efdc07132b20e2cec44) – the inherited model looks much nicer.

packages/components/app/styles/components/tag.scss
```diff
@@ -0,0 +1,92 @@
...
+  font-weight: var(--token-typography-font-weight-medium);
+  line-height: 1rem; // 16px
+  vertical-align: middle;
+}
+
+.hds-tag__dismiss {
+  flex: 0 0 auto;
+  border-radius: inherit;
+  border-top-right-radius: 0;
+  border-bottom-right-radius: 0;
```
didoo, line 28: I _think_ you can do this, if you want to reduce the "visual noise":
```suggestion
  border-radius: inherit 0 0 inherit;
```

alex-ju: Individual subproperties don't inherit, unfortunately, so we'll have to use the overrides on certain corners

packages/components/app/styles/components/tag.scss
```diff
@@ -0,0 +1,92 @@
...
+  &:active,
+  &.mock-active{
+    color: var(--token-color-foreground-action-active);
+  }
+}
+
+.hds-tag__dismiss ~ .hds-tag__text,
+.hds-tag__dismiss ~ .hds-tag__link {
+  border-radius: inherit;
+  border-top-left-radius: 0;
```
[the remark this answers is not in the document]
alex-ju: We do, otherwise we'll end up with something like this:
Screenshot 2022-07-12 at 14 37 19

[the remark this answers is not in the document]
didoo: Sorry, I removed the wrong line. I meant the line #64 (`border-radius: inherit;`)

packages/components/app/styles/components/tag.scss
didoo: Noooo way!! Didn't know this is a "special case". TIL today 🤷‍♂️

alex-ju:
I pushed a commit which explores visual truncation for tags, as an alternative to wrapping on multiple lines when the content is long (we're using an arbitrary value of 120px – could better be rems)

Before [17312bd](https://github.com/hashicorp/design-system/pull/443/commits/17312bd2ae7fb9be1b4f817b4acbf8f6ec2abeb5)
localhost_3000_components_tag (1)

After [17312bd](https://github.com/hashicorp/design-system/pull/443/commits/17312bd2ae7fb9be1b4f817b4acbf8f6ec2abeb5)
localhost_3000_components_tag

I'm interested in feedback, especially from an accessibility perspective. What do you think, @MelSumner? Is it conformant if a screen-reader has access to the full content or is it still an impairment for sighted users? We also use the `title` attribute, to visually provide the content in full (but it's only available on hover, unfortunately, not on focus).

MelSumner:
@alex-ju some thoughts: 

1. if the `title` attribute is used, the content will be read out twice to users with a screen reader (poor implementation is why it is generally recommended to avoid using the `title` attribute all together)
2. it doesn't seem like this approach would be usable by keyboard-only users (or am I missing something?)

alex-ju:
> @alex-ju some thoughts:
> 
> 1. if the `title` attribute is used, the content will be read out twice to users with a screen reader (poor implementation is why it is generally recommended to avoid using the `title` attribute all together)
> 2. it doesn't seem like this approach would be usable by keyboard-only users (or am I missing something?)

thanks, Melanie! both valid points and I can see how it doesn't help from user experience perspective 😞 I'll check-in with Andrew and see how can we maybe improve the version where it wraps.

didoo (commented):
Added a few comments:
- about how we use/declare/assign colors (but this is personal preference)
- about documentation (could be improved, and some props are missing)

didoo (approved):
All good for me

packages/components/tests/dummy/app/templates/components/tag.hbs
```diff
@@ -0,0 +1,148 @@
...
+    </dd>
+    <dt>...attributes</dt>
+    <dd>
+      <p><code class="dummy-code">...attributes</code> spreading is supported on this component.</p>
+    </dd>
+  </dl>
+</section>
+
+<section>
+  <h3 class="dummy-h3" id="how-to-use"><a href="#how-to-use" class="dummy-link-section">§</a> How to use</h3>
```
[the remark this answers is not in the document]
alex-ju: It's a good question I asked myself as well. I'd like to run it by the team (it's already in the backlog for our sync). I think there's value in providing a list component – either as an alternative (e.g. `Hds::Tag::List`) or as the recommended component, but not sure how easy is to manage a collection of items being passed into the component; being able to run the loop at the application level has its advantages.

heatherlarsen (changes requested):
Took a quick look, found one major design difference that I noted. @agendelHC will go through and note anything else he sees.

packages/components/app/styles/components/tag.scss
```diff
@@ -0,0 +1,101 @@
...
+//
+//
+
+@use "../mixins/focus-ring" as *;
+
+$hds-tag-border-radius: 13px;
+
+.hds-tag {
+  align-items: stretch;
+  background-color: var(--token-color-surface-interactive);
```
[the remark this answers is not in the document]
alex-ju: Thanks, Heather! I applied your correction

alex-ju:
Rebased to `main` to allow Vercel to deploy after the latest changes.

packages/components/app/styles/components/tag.scss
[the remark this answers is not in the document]
alex-ju: However, after a chat with Andrew, it sound like `Surface/Interactive` is the one we'd want here ([as per Figma](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/LdUGD1ub9oPL28gv5w4Tdz/HDS-Product---Components?node-id=13720%3A34360)) so I just changed it back to how it was 😄 

@agendelHC will review them all today and will let me know if there's anything out of place.

packages/components/app/styles/components/tag.scss
```diff
@@ -0,0 +1,101 @@
...
+  border: 1px solid var(--token-color-border-strong);
+  border-radius: $hds-tag-border-radius;
+  color: var(--token-color-foreground-primary);
```
agendelHC, line 17:
```suggestion
  color: var(--token-color-surface-interactive);
```

packages/components/app/styles/components/tag.scss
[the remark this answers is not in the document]
heatherlarsen: Hmm, it's technically the same color (#FFFFFF), but some tags are interactive and some aren't. Curious how we want to handle this...should the interactive ones be `Surface/Interactive` and the static ones be `Surface/Primary`? Seems silly...not sure how strongly we want to be tied to the semantic tokens.

packages/components/app/styles/components/tag.scss
heatherlarsen: @agendelHC `color` refers to the text or icon (foreground) color, not the background color of the component, so we wouldn't use `surface` colors for it. According to the [designs](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/LdUGD1ub9oPL28gv5w4Tdz/HDS-Product---Components?node-id=13720%3A34360), this token is actually correct. Does something look off to you?

agendelHC (approved):

packages/components/app/styles/components/tag.scss
```diff
@@ -0,0 +1,101 @@
+//
+// TAG COMPONENT
+//
+// properties within each class are sorted alphabetically
+//
+//
+
+@use "../mixins/focus-ring" as *;
+
+$hds-tag-border-radius: 13px;
```
heatherlarsen, line 10: @alex-ju When the content wraps, 13px doesn't keep the pill shape...thinking if you bump it up to something ridiculous like 50px that should cover _most_ of the edge cases.

packages/components/addon/components/hds/tag/index.hbs
```diff
@@ -0,0 +1,26 @@
+<span class={{this.classNames}} ...attributes>
+  {{#if this.onDismiss}}
+    <button class="hds-tag__dismiss" type="button" aria-label="Dismiss {{this.text}}" {{on "click" this.onDismiss}}>
+      <FlightIcon class="hds-tag__dismiss-icon" @name="x" @size="16" @isInlineBlock={{false}} />
```
heatherlarsen, line 4:
```suggestion
      <FlightIcon class="hds-tag__dismiss-icon" @name="x" @size="16" @isInlineBlock={{false}} @color="var(--token-color-foreground-primary)" />
```

heatherlarsen (changes requested):
Left a few additional comments.

packages/components/addon/components/hds/tag/index.hbs
alex-ju: Thanks! Moved the declaration to css as we have a `hds-tag__dismiss-icon` that we can leverage.

heatherlarsen (approved):
lgtm :shipit:

alex-ju:
Thanks all for the reviews. To avoid stalling this PR too much (and as we already have the Tag component available in Figma) I will get this it in and follow-up with the 'Design guidelines' when they're ready (likely very soon).

didoo:
> and follow-up with the 'Design guidelines' when they're ready (likely very soon).

if they're ready before this is merged, you can use this PR: #479

PR #449, opened 2022-07-15
Fix the default button type value in doc

alex-ju:
### :pushpin: Summary

Fix the default button type value mentioned in the documentation

The default type for a button is `submit`, not `button`, so I've updated the documentation accordingly. I also find '_To use a different value, declare a different value_' redundant, but happy to keep it if someone feels strongly about it.

### :link: External links

Found this as I checked the guidance we provide before replying to [a query in Slack](https://hashicorp.slack.com/archives/C7KTUHNUS/p1657841679203199).

***


didoo (approved):
👍

PR #456, opened 2022-07-18
Fix computation error on disclosure utility

alex-ju:
### :pushpin: Summary

Fix computation error on disclosure utility component

Ember, possibly on certain versions (3.2x), doesn't support setting and getting a `@tracked` property in the same computation. To avoid this type of errors, we're using an explicit conditional statement to toggle the `isOpen` property, so we get the property in a computation and set it in a different one.

Sample error:
```
Error: You attempted to update `isOpen` on `HdsDisclosureComponent`, but it had already been used previously in the same computation.  Attempting to update a value after using it in a computation can cause logical errors, infinite revalidation bugs, and performance issues, and is not supported.
```

### :link: External links

[Issue raised in Slack](https://hashicorp.slack.com/archives/CPB8GS9QT/p1657913533488579)

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/disclosure/index.js
+++ b/packages/components/addon/components/hds/disclosure/index.js
@@ -18,7 +18,12 @@ export default class HdsDisclosureComponent extends Component {
     if (!this.toggleRef) {
       this.toggleRef = event.currentTarget;
     }
-    this.isOpen = !this.isOpen;
+    // we're using an explicit conditional statement to toggle the `isOpen` property to avoid a backtracking rerender assertion caused by setting and getting a `@tracked` property in the same computation
+    if (this.isOpen) {
+      this.isOpen = false;
+    } else {
+      this.isOpen = true;
+    }
     // we explicitly apply a focus state to the toggle element to overcome a bug in WebKit (see b8abfcf)
     this.toggleRef.focus();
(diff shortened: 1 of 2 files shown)


packages/components/addon/components/hds/disclosure/index.js
```diff
@@ -18,7 +18,11 @@ export default class HdsDisclosureComponent extends Component {
     if (!this.toggleRef) {
       this.toggleRef = event.currentTarget;
     }
-    this.isOpen = !this.isOpen;
+    if (this.isOpen) {
```
didoo, line 21: please add a comment here to explain why it is necessary to do this (otherwise in the future someone - likely me - will refactor this code, thinking that this is an "error") ;)

(and if possible, add a link to this issue so it's easy to find more details about the issue)

didoo (approved):
all good, just please add a comment in the code explaining why it's necessary to have that strange "logic"

packages/components/addon/components/hds/disclosure/index.js
didoo: 👍

PR #441, opened 2022-07-11
Form controls - Styles updates

didoo:
### :pushpin: Summary

Now that the `form controls` are almost completely finalized on the design side, it's time to re-check that their styles in CSS match the latest design specs.

### :hammer_and_wrench: Detailed description

In this PR I have:
- created component-level tokens for form controls (as temporary solution, while we wait to implement them in the proper tokens pipeline)
- applied latest styling specs from design (using component-level design tokens) to:
  - `Legend`
  - `Error`
  - `HelperText`
  - `Label`
  - `TextInput`
  - `Textarea`
  - `Select`
  - `Checkbox`
  - `Radio` 
  - `Toggle` 

Previews:
- [Base elements](https://hds-components-git-form-controls-styles-updates-hashicorp.vercel.app/components/form/base-elements#showcase)
- [TextInput](https://hds-components-git-form-controls-styles-updates-hashicorp.vercel.app/components/form/text-input#showcase)
- [Textarea](https://hds-components-git-form-controls-styles-updates-hashicorp.vercel.app/components/form/textarea#showcase)
- [Select](https://hds-components-git-form-controls-styles-updates-hashicorp.vercel.app/components/form/select#showcase)
- [Checkbox](https://hds-components-git-form-controls-styles-updates-hashicorp.vercel.app/components/form/checkbox#showcase)
- [Radio](https://hds-components-git-form-controls-styles-updates-hashicorp.vercel.app/components/form/radio#showcase)
- [Toggle](https://hds-components-git-form-controls-styles-updates-hashicorp.vercel.app/components/form/Toggle#showcase)

### :link: External links

- Jira issue: https://hashicorp.atlassian.net/browse/HDS-295

### 🤔 Open questions

- The form controls share a lot of styles in terms of CSS declarations: should we introduce a set of common class names that can be applied to the components (via the `classNames` JS method) to avoid code duplication? In that case, how much is the risk to increase the complexity of the CSS code for the form controls? (eg. where does a style come from? what happens if I change a common file?) 

***

### 👀 How to review

👉 Review commit-by-commit or by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #458, opened 2022-07-18
Form controls - Component-level design tokens

didoo:
### :pushpin: Summary

This is a follow-up of #441, in which we add proper design tokens for the `form controls` components.

Re. the naming, there's still an ongoing discussion (see comments in [this Figma file](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/CsGFsY5gotq3YE7USnA0Zr/HDS-Product---Components)) but this is a preliminary attempt to try to converge to a similar (not necessarily identical) naming convention between code and Figma.

In this PR I have:
- added the "source" design tokens for the form base elements and controls
- updated the build script for tokens to transform “time/seconds” values
- generated new tokens in output
- applied the new tokens to the _base elements_ and `textinput` `textarea` `select` `checkbox` `radio` `toggle`
- removed the old (temporary) CSS tokens file

As extra, I have decided that, to be able to distribute the icons used in the form controls also for other consumers (eg. DevDot), it would be great to have these assets defined via specific design tokens. So I have also:
- converted the embedded background images for form controls icons to design tokens (and in the process done some small refactorings)

Notice: while reviewing the naming convention, keep in mind that another consumer of these tokens will be the DevDot codebase, so imagine @ashleemboyer using these tokens, and trying to understand by the naming itself what a token is for).

***

### 👀 How to review

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/tokens/dist/devdot/css/tokens.css
```diff
@@ -176,6 +176,68 @@
...
+  --token-form-text-input-background-image-data-url-search: url("data:image/svg+xml,%3csvg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'%3e%3cg fill='%23656A76'%3e%3cpath d='M7.25 2a5.25 5.25 0 103.144 9.455l2.326 2.325a.75.75 0 101.06-1.06l-2.325-2.326A5.25 5.25 0 007.25 2zM3.5 7.25a3.75 3.75 0 117.5 0 3.75 3.75 0 01-7.5 0z' fill-rule='evenodd' clip-rule='evenodd'/%3e%3c/g%3e%3c/svg%3e"); /* notice: the icon color is hardcoded here! */
+  --token-form-text-input-background-image-data-url-search-cancel: url("data:image/svg+xml,%3csvg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'%3e%3cpath d='M12.78 4.28a.75.75 0 00-1.06-1.06L8 6.94 4.28 3.22a.75.75 0 00-1.06 1.06L6.94 8l-3.72 3.72a.75.75 0 101.06 1.06L8 9.06l3.72 3.72a.75.75 0 101.06-1.06L9.06 8l3.72-3.72z'/%3e%3c/svg%3e"); /* notice: the icon color is hardcoded here! */
+  --token-form-toggle-width: 32px;
+  --token-form-toggle-height: 16px;
+  --token-form-toggle-base-surface-color-default: #f1f2f3; /* the toggle has a different base surface color, compared to the other controls */
```
heatherlarsen, line 231: Is `base` really necessary here? Seems like `toggle` covers the entity well enough because there aren't multiple toggle styles.

heatherlarsen (approved):
lgmt :shipit:

packages/tokens/dist/devdot/css/tokens.css
didoo: I thought the same, too. But in the end I went for consistency with the naming structure of the generic control tokens, which is `--token-form-[control]-base-surface-color-default`.

MelSumner (approved):
Looks correct! :shipit:

PR #474, opened 2022-07-21
move flight-website out of packages

Dhaulagiri:
### :pushpin: Summary

Unlike everything else in that directory, it is not actually a package. This PR moves it out to its own directory in the top level of the monorepo.

The idea for this came to after seeing how a few other design systems organize their monorepos. I'm ok if we don't do this, but I think it ultimately does make more sense than how we had initially organized things. Now `packages` will only contain actual packages we publish to NPM and any other things (flight's website, future websites) can live at the monorepo root.

### Note

This change breaks our current vercel config for the flight website as it's expecting the site to exist in its old location. If we're ok with this change, I'll update vercel _before_ I merge this and do a quick smoke test to ensure everything is 💯 

### 👀 How to review

👉 Review by files changed


KristinLBradley (approved):
Seems fine by me.

didoo (approved):
No strong opinions on this. Works for me. 👍

PR #468, opened 2022-07-20
Generate flight icons as CSS props

didoo:
### :pushpin: Summary

This is a follow up of a conversation I had today with @johncowen about the possibility for us to export the Flight icons embedded in CSS props and for Consul UI to consume them in this way (potentially doing tree-shaking on their side importing single icons when/where needed).

### :hammer_and_wrench: Detailed description

In this PR I have:
- added a script to bundle Flight icons as CSS variables
- generated the icons in CSS-prop format (single file with all icons + distinct files per icon)

TODO:
- [x] if approved, add changeset file

Please @johncowen let me know what you think of this (and if they work of course)

***

### 👀 How to review

👉 Review commit-by-commit or by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable
- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #482, opened 2022-07-26
Update changeset description

didoo:
See https://github.com/hashicorp/design-system/pull/468#discussion_r929897720


Dhaulagiri:
We'll see what happens with the release action. Might have to make the change directly in https://github.com/hashicorp/design-system/pull/483/files.

Dhaulagiri:
Neat! It worked!

PR #485, opened 2022-07-26
install deps for icons sync with --immutable

Dhaulagiri:
### :pushpin: Summary

--frozen-lockfile is deprecated in yarn3. See this recent run [where this was noted](https://github.com/hashicorp/design-system/runs/7509643214?check_suite_focus=true#step:4:7). This now matches [how we do it in other github actions](https://github.com/hashicorp/design-system/blob/ed96740352c8a1bda0cebbcec41eb6eb7de0a67d/.github/workflows/ci-components.yml#L30).


alex-ju (approved):
Nice work chasing this up🕵️

PR #492, opened 2022-07-28
Fix custom id on form fields

alex-ju:
### :pushpin: Summary

Fix custom `id` on form fields

`Form::Field` knows to look up for a custom @id and use it instead of an autogenerated one, but in refactor we forgot to pass the @id value to the Field.

We currently have [a test for custom `@id` on Form::Field](https://github.com/hashicorp/design-system/blob/main/packages/components/tests/integration/components/hds/form/field/index-test.js#L81-L105), but it doesn't capture the case where the `@id` is not provided (e.g. from `Form::Checkbox::Field` to `Form::Field`), so I added an quick test on each form field to make sure we don't break this again in future.

***


didoo (approved):
Nice, clean fix! 

Thank you

PR #496, opened 2022-07-28
Add `multiple`/`size` styles to select component

alex-ju:
### :pushpin: Summary

Add `multiple`/`size` styles to select component.

We don't expect to find this variation in our products, but as we allow these two attributes in the select component we want to make sure they look decently well when they're enabled.

Tried to keep the `multiple`/`size` styles separated from the main select styles, for readability.
Added some states in examples so we can catch any visual regression, but happy to scrap them if we don't want to encourage this.

**Note:**
The selected options are still light blue in Chrome when the `<select>` element is in focus, as there is not way to override them, unfortunately (cc: @heatherlarsen). More on this on Chromium bug tracker [Issue 398417](https://bugs.chromium.org/p/chromium/issues/detail?id=398417) and [Chromium group chat](https://groups.google.com/a/chromium.org/g/blink-reviews/c/2FeNfAR8FAQ).

<table>
<tr><th>Browser</th><th>Before</th><th>After</th></tr>
<tr><td>Chrome</td><td>

select-multiple-chrome-before

</td><td>

select-multiple-chrome-after

</td></tr>
<tr><td>Firefox</td><td>

select-multiple-firefox-before

</td><td>

select-multiple-firefox-after

</td></tr>
</table>

Note: Safari doesn't support custom styling on these elements/states it seems.

### :link: External links

[Design reference](https://www.figma.com/file/vPHO3jpKMaBlvIwVnfbOuW/PowerSelect-styles?node-id=14537%3A35556)

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/app/styles/components/form/select.scss
+++ b/packages/components/app/styles/components/form/select.scss
@@ -67,3 +67,34 @@
     }
   }
 }
+
+// MULTIPLE/SIZE
+
+.hds-form-select {
+  &[multiple],
+  &[size]{
+    background: none;
+
+    option {
+      border-radius: 3px;
+      margin: 2px auto;
+
+      &:hover {
+        color: var(--token-color-foreground-action);
+      }
+      &:disabled {
+        color: var(--token-color-foreground-disabled);
+      }
+      &:checked {
+        color: var(--token-color-foreground-high-contrast);
+        background: var(--token-color-palette-blue-200);
+      }
+    }
+
+    optgroup {
+      color: var(--token-color-foreground-strong);
+      font-weight: var(--token-typography-font-weight-semibold);
+      font-style: normal; // reset Firefox default
+    }
+  }
+}
--- a/packages/components/tests/dummy/app/templates/components/form/select.hbs
+++ b/packages/components/tests/dummy/app/templates/components/form/select.hbs
@@ -637,17 +637,17 @@
   <CodeBlock
     @language="markup"
     @code='
-      <Hds::Form::Select::Field name="infrastructure" multiple size="4" as |F|>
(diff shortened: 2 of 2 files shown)


packages/components/app/styles/components/form/select.scss
```diff
@@ -67,3 +67,35 @@
...
+      margin: 2px auto;
+
+      &:hover {
+        color: var(--token-color-foreground-action);
+      }
+      &:disabled {
+        color: var(--token-color-foreground-disabled);
+      }
+      &:checked {
+        border-radius: 3px;
```
didoo, line 89: [super-giga-nitpick] for me this goes on line 80, under the `option` selector.

didoo (approved):
I like it! 😍

Well done! 🙌

packages/components/app/styles/components/form/select.scss
alex-ju: right – will move it there

MelSumner:
Some reading: https://www.24a11y.com/2019/select-your-poison/#:~:text=some%20usability%20tests.-,Multi%2Dselect,-So%20far%2C%20the

I don't remember this coming up in team discussion, is it a feature requested by a team?

didoo:
> I don't remember reading about this option in the CRD, was it requested later?
> I probably would have suggested some additional reading/conversation about it, so I'm trying to find some clarity. Thank you!

It's a native feature of the `<select>` component, and since users can pass HTML attributes to the form controls, means they may pass `multiple` (and `size`) attributes. This PR does nothing more than styling the native element.

heatherlarsen:
@alex-ju This is awesome! Thanks for tracking down some of those weird bugs and giving this a shot 🙂

PR #470, opened 2022-07-20
Add Stepper Indicator Component

jorytindall:
### :pushpin: Summary

Initial component generation and variant/state styling for the stepper indicator.

### :hammer_and_wrench: Detailed description

Configured the initial stepper indicator component with `status` types, `states`, `isInteractive` boolean, and `isProcessing` state with loading Flight Icon. Additionally I added examples of the different statuses and variants when paired together in the showcase.

### :camera_flash: Screenshots
Screen Shot 2022-07-20 at 9 48 25 AM

### :link: External links

***

### 👀 How to review

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable
- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #497, opened 2022-07-28
Addresses some a11y issues in stepper component

MelSumner:
### :pushpin: Summary

If merged, this PR attempts to address some of the WCAG conformance issues with the stepper component. 

Changes made: 

1. Component: Added template-only component `Hds::Stepper` to provide and element with the appropriate role 
1. Component: moved `aria-hidden` to the SVG itself instead of the wrapper due to the way some browsers try to fix what they perceive as author error
1. Docs: Updated the rendered code with the new component
1. Docs: Added additional prose to the accessibility section
1. Docs: Updated some prose for readability
1. Docs: Fixed cases of multiple non-unique IDs on the same page
1. Docs: Fixed incorrect markup for `<dl>` in API section

If we decide to move forward with these proposed changes, there's a couple other TODOs:
- update sample code
- maybe add a test for a composed version of the component

Other remaining issues to discuss:

1. I do not think that the task indicators are visible enough. If we cannot have this design updated, we need to very clearly indicate that using them would be in violation of WCAG (and list which Criteria they fail in their current state).
2. It seems confusing that `isInteractive` adds a pointer for hover state but there is no visible focus (and if none is added, it would fail the [Focus Visible](https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html) Success Criteria). I wonder if we can shore this up a bit. Since the icon itself isn't interactive, perhaps we should change this API to `hasInteractive`, and put the hover and focus styling on the anticipated nested `<a>` element, or something like that?
3. I do not understand how this component is supposed to add interactivity; more complete instructions and examples are needed. Should it be used as a block component and nest an interactive element in that? If so, which ones are okay? Just a plain link? Are we styling that? We should clarify all of these things.
4. The prose does not use plain language; recommend edits to make it easier to read/understand (and fit the tone of the other components)
5. After testing with `prefers-reduced-motion`, I recommend re-visiting the design. Filed [HDS-441](https://hashicorp.atlassian.net/browse/HDS-441) with video. 
6. After reviewing the full pattern in a call, I understand the eventual goals better and we should improve a couple of parent/child relationships and how they have been defined.

Before this PR: 
image

After this PR:
CleanShot 2022-07-28 at 15 47 34

### :link: External links

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/stepper/index.hbs
+++ b/packages/components/addon/components/hds/stepper/index.hbs
@@ -1,0 +1,3 @@
+<div role="list" class="hds-stepper" ...attributes>
+  {{yield}}
+</div>
--- a/packages/components/addon/components/hds/stepper/indicator/step.hbs
+++ b/packages/components/addon/components/hds/stepper/indicator/step.hbs
@@ -1,6 +1,6 @@
 <div class={{this.classNames}} ...attributes role="listitem" aria-label={{@status}}>
-  <div class="hds-stepper-indicator-step__svg-hexagon" aria-hidden="true">
-    <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+  <div class="hds-stepper-indicator-step__svg-hexagon">
+    <svg aria-hidden="true" width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
       <path
         d="m3.664 6.264 6.99-4.42a2.5 2.5 0 0 1 2.67-.002l7.01 4.422A2.5 2.5 0 0 1 21.5 8.38v7.242a2.5 2.5 0 0 1-1.166 2.115l-7.01 4.422a2.5 2.5 0 0 1-2.67-.002l-6.99-4.42A2.5 2.5 0 0 1 2.5 15.623V8.377a2.5 2.5 0 0 1 1.164-2.113Z"
         stroke-width="1"
--- a/packages/components/tests/dummy/app/templates/components/stepper.hbs
+++ b/packages/components/tests/dummy/app/templates/components/stepper.hbs
@@ -6,12 +6,12 @@
   <p class="dummy-paragraph">A stepper indicator helps the user maintain context and directionality when advancing
     through a multi-step flow or feature, and in certain circumstances acts as a navigational device between steps. It
     is generally assembled as part of a larger stepper pattern.</p>
-  <p class="dummy-paragraph">For practical accessibility purposes, consider a stepper (the whole component) to be a list of items, with each step
-    being a list item.</p>
+  <p class="dummy-paragraph">For practical accessibility purposes, consider a stepper (the whole component) to be a list
+    of items, with each step being a list item.</p>
   <p class="dummy-paragraph">
     As of this writing the HDS team publishes two separate (but related)
     <code class="dummy-code">Stepper::Indicator</code>
-    components that serve different hierarchal purposes.
+    components that serve different hierarchical purposes.
     <ol>
       <li>The
         <code class="dummy-code">Indicator::Step</code>
--- a/packages/components/tests/dummy/app/templates/components/stepper.hbs
+++ b/packages/components/tests/dummy/app/templates/components/stepper.hbs
@@ -6,7 +6,7 @@
   <p class="dummy-paragraph">A stepper indicator helps the user maintain context and directionality when advancing
     through a multi-step flow or feature, and in certain circumstances acts as a navigational device between steps. It
     is generally assembled as part of a larger stepper pattern.</p>
-  <p>For practical accesibility purposes, consider a stepper (the whole component) to be a list of items, with each step
+  <p class="dummy-paragraph">For practical accessibility purposes, consider a stepper (the whole component) to be a list of items, with each step
     being a list item.</p>
   <p class="dummy-paragraph">
     As of this writing the HDS team publishes two separate (but related)
--- a/packages/components/tests/dummy/app/templates/components/stepper.hbs
+++ b/packages/components/tests/dummy/app/templates/components/stepper.hbs
@@ -6,6 +6,8 @@
   <p class="dummy-paragraph">A stepper indictor helps the user maintain context and directionality when advancing
     through a multi-step flow or feature, and in certain circumstances acts as a navigational device between steps. It
     is generally assembled as part of a larger stepper pattern.</p>
+  <p>For practical accesibility purposes, consider a stepper (the whole component) to be a list of items, with each step
+    being a list item.</p>
   <p class="dummy-paragraph">
     As of this writing the HDS team publishes two separate (but related)
     <code class="dummy-code">Stepper::Indicator</code>
@@ -17,7 +19,7 @@
         sequentially.</li>
       <li>The
         <code class="dummy-code">Indicator::Task</code>
-        component: used either on it's own to denote smaller task-oriented flows or in combination with the
+        component: used either on its own to denote smaller task-oriented flows or in combination with the
         <code class="dummy-code">Step</code>
         indicator to list multiple tasks within a step.</li>
     </ol>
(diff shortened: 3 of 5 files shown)


packages/components/tests/dummy/app/templates/components/stepper.hbs
```diff
@@ -106,7 +102,9 @@
     '
   />
   <p class="dummy-paragraph">Renders to:</p>
-  <Hds::Stepper::Indicator::Step @text="1" @status="incomplete" />
+  <Hds::Stepper>
+    <Hds::Stepper::Indicator::Step @text="1" @status="incomplete" />
```
meirish, line 106: [non-blocking]: do we have guidelines on when to use a contextual component and when to compose namespace components? I go back and forth tbh, but I do like the composition here.

packages/components/addon/components/hds/stepper/index.hbs
```diff
@@ -0,0 +1,3 @@
+<div role="list" class="hds-stepper" ...attributes>
```
MelSumner, line 1: So I am not sure if I should leave `...attributes` here. 

Pro: a `data-` attribute could be added for tests?
Cons: Do we need it?

Also, I added the component name as a class even though we are not styling it from within the addon: 
1. it's consistent with all of the other components
2. it's a convenient styling class already attached and ready for use (by consuming teams)

meirish, line 1: 👍

meirish: I'm +1 on leaving both of them in - leaving `...attributes` doesn't hurt - the `data-` passthrough is nice and is certainly used for tests. I think components without `...attributes` sometimes surprising to users, and I agree with your points for the class as well.

packages/components/tests/dummy/app/templates/components/stepper.hbs
```diff
@@ -115,25 +113,34 @@
     '
   />
   <p class="dummy-paragraph">Renders to:</p>
-  <Hds::Stepper::Indicator::Step @text="1" @status="incomplete" @isInteractive={{true}} />
-  <h5 class="dummy-h5">Notating status</h5>
+  <Hds::Stepper>
+    <Hds::Stepper::Indicator::Step @text="1" @status="incomplete" @isInteractive={{true}} />
```
MelSumner, line 117: This is one place where, if we intend for the user to nest an interactive element, we could provide an example: 

```suggestion
    <Hds::Stepper::Indicator::Step @text="1" @status="incomplete" @isInteractive={{true}}><a href="#">Step One</a></Hds::Stepper::Indicator::Step>
```

packages/components/tests/dummy/app/templates/components/stepper.hbs
```diff
@@ -6,6 +6,8 @@
   <p class="dummy-paragraph">A stepper indictor helps the user maintain context and directionality when advancing
     through a multi-step flow or feature, and in certain circumstances acts as a navigational device between steps. It
     is generally assembled as part of a larger stepper pattern.</p>
+  <p>For practical accesibility purposes, consider a stepper (the whole component) to be a list of items, with each step
```
alex-ju, line 9:
```suggestion
  <p class="dummy-paragraph">For practical accessibility purposes, consider a stepper (the whole component) to be a list of items, with each step
```

packages/components/addon/components/hds/stepper/index.hbs
alex-ju: Is there a reason for using a `role="list"` instead of a `<ol>` here?

packages/components/addon/components/hds/stepper/indicator/step.hbs
```diff
@@ -1,6 +1,6 @@
 <div class={{this.classNames}} ...attributes role="listitem" aria-label={{@status}}>
```
alex-ju, line 1: And consequently an `<li>` here?

packages/components/addon/components/hds/stepper/indicator/step.hbs
```diff
@@ -1,6 +1,6 @@
...
-  <div class="hds-stepper-indicator-step__svg-hexagon" aria-hidden="true">
-    <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+  <div class="hds-stepper-indicator-step__svg-hexagon">
+    <svg aria-hidden="true" width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
alex-ju, line 3: Agree with this change and, as mentioned [in a previous comment](https://github.com/hashicorp/design-system/pull/470#pullrequestreview-1051030518), I suggest we move the class to the svg element and remove the div wrapper.

packages/components/tests/dummy/app/templates/components/stepper.hbs
```diff
@@ -6,6 +6,8 @@
   <p class="dummy-paragraph">A stepper indictor helps the user maintain context and directionality when advancing
```
alex-ju, line 6:
```suggestion
  <p class="dummy-paragraph">A stepper indicator helps the user maintain context and directionality when advancing
```

packages/components/tests/dummy/app/templates/components/stepper.hbs
```diff
@@ -6,6 +6,8 @@
...
     through a multi-step flow or feature, and in certain circumstances acts as a navigational device between steps. It
     is generally assembled as part of a larger stepper pattern.</p>
+  <p>For practical accesibility purposes, consider a stepper (the whole component) to be a list of items, with each step
+    being a list item.</p>
   <p class="dummy-paragraph">
     As of this writing the HDS team publishes two separate (but related)
     <code class="dummy-code">Stepper::Indicator</code>
```
alex-ju, line 13: Below this line (on L14) `hierarchal` → `hierarchical`

alex-ju (commented):
Left a few early comments addressing typos and such and a question about potentially using semantic HTML tags (I may be missing something here).

packages/components/addon/components/hds/stepper/indicator/step.hbs
didoo: @MelSumner @alex-ju why are we making this so complex? why don't just remove the `role="listitem"` here? this is not a list item, and if we remove it we can avoid adding a parent wrapper, and all this "problem" goes away.
The indicator should be evalutated in the context of the whole stepper, is a sub-sub-component.
screenshot_1737
/cc @jorytindall

alex-ju: Looking at the diagram, yes `<Hds::Stepper:Step>` is an item, but the indicator isn't. So, I'm happy to remove the roles we defined and postpone that for when we tackle the `<Hds::Stepper:Step>`

alex-ju: I admit, I've not checked the CRD/component anatomy first and got confused by the presence of these `role` attributes.

MelSumner: > why are we making this so complex?

I think it would be good for you to clarify your claim that this PR is adding complexity.  

It is doing the opposite. It's adding a single template-only component that provides a small, flexible wrapper that will help users meet WCAG conformance requirements. 

After looking at this screenshot, I'm even more convinced that there is no complexity being added. It's both flexible and future-proof in implementation.

packages/components/tests/dummy/app/templates/components/stepper.hbs
MelSumner: thanks for catching that, fixed! :thumbsup:

packages/components/addon/components/hds/stepper/indicator/step.hbs
MelSumner: Good catch, thank you! Fixed!

MelSumner:
We've decided to go in a different direction for the time being, so closing this PR.

PR #501, opened 2022-07-29
Addressing a11y issues and updating component structure

jorytindall:
### :pushpin: Summary

If merged, this PR addresses issues with fundamental a11y conformance at the component level and adds clarification and context for why the `Stepper` component is not WCAG-conformant when used in isolation.

Changes made:

* Updated the component structure to `Hds::Stepper::Step/Task::Indicator`, adhering to Ember parent->child architecture.
* Moved `aria-hidden` from the svg wrapper to the svg element itself to account for inconsistencies between browsers.
* Added clarifying information to the Accessibility section to provide context on why the component isn't conformant and how to build a conformant pattern using the indicator.
  * **Note**: Did not remove the relevant WCAG guidelines _for now_ as the content still seems relevant for assembling a larger pattern.
* Fixed some spelling, grammatical, and prose issues.

***

### 👀 How to review

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/app/styles/components/stepper/index.scss
```diff
@@ -1,2 +1,2 @@
-@use './indicator/step.scss';
-@use './indicator/task.scss';
\ No newline at end of file
+@use './step-indicator.scss';
```
Dhaulagiri, line 1: we could do this in a follow-up PR, but it _seems_ like `/.step/indicator.scss` is the likely structure we'll want here eventually

Dhaulagiri (approved):

packages/components/app/styles/components/stepper/index.scss
jorytindall: That's the structure I had initially, but it threw a Sass error for a duplicative namespace 🤔

Dhaulagiri: sounds...sassy 🙃

Dhaulagiri: i see what you mean. something we can dig into later 👍

alex-ju:
Nice one @jorytindall! 🙌

PR #514, opened 2022-08-01
HDS-508 added automated a11y test support

MelSumner:
### :pushpin: Summary

If merged, this PR would add `ember-a11y-testing` as a dependency and sets the test to run at the global level instead of the individual test level. 

Instructions that I followed are here: https://github.com/ember-a11y/ember-a11y-testing 


packages/components/tests/integration/components/hds/alert/index-test.js
```diff
@@ -172,6 +173,19 @@ module('Integration | Component | hds/alert/index', function (hooks) {
     assert.dom('#test-alert').hasAttribute('aria-labelledby', 'custom-id');
   });
 
+  test('it should pass available a11y automation tests', async function (assert) {
```
Dhaulagiri, line 176: @MelSumner I'm curious your thoughts about the approach of creating new tests like this or adding `await a11yAudit()` checks into our existing tests that expose more varied iterations of each component's API? Any pros or cons between the two approaches?

MelSumner: I think that improves the approach, because it's making sure each iteration passes the available accessibility automation checks. Good idea!

MelSumner: follow-up thought: I wonder how much this will slow our tests down if we run this for each iteration? 

As is: 
CleanShot 2022-08-01 at 14 24 53

Let me try adding the a11y audit to just the different iterations and post those results too.

MelSumner: (1992ms vs 5875ms)

KristinLBradley: > @MelSumner I'm curious your thoughts about the approach of creating new tests like this or adding `await a11yAudit()` checks into our existing tests that expose more varied iterations of each component's API? Any pros or cons between the two approaches?

This sounds like a good approach to me but I'm happy with whatever approach you think works best @MelSumner

packages/components/tests/integration/components/hds/alert/index-test.js
```diff
@@ -172,6 +173,19 @@ module('Integration | Component | hds/alert/index', function (hooks) {
...
+    await render(
+      hbs`<Hds::Alert @type="inline" id="test-alert" as |A|><A.Title id="custom-id">This is the title</A.Title><A.Button @text="I am a button" @size="small" /></Hds::Alert>`
+    );
+
+    let axeOptions = {
+      runOnly: ['wcag21a', 'wcag21aa', 'best-practice'],
+    };
```
alex-ju, line 183: I wonder if we could store this at a higher level, so we consistently test against the same criteria.

alex-ju (commented):
Excited to see this! I like the idea of adding these tests where needed. We could have components that would benefit from these in mosts tests and components where they can be applied to only one test as the markup doesn't change.

packages/components/tests/integration/components/hds/alert/index-test.js
MelSumner: @alex-ju perhaps the `test-helper.js` file that's already in the tests folder. WDYT?

alex-ju: yep, that looks like a good place!

KristinLBradley (approved):

alex-ju (approved):
Work for me! Thanks for putting this together!

PR #521, opened 2022-08-03
custom lint rule experiment

MelSumner:
### :pushpin: Summary

This PR tries out adding custom lint rule support for the add-on.

- followed plugin instructions for ember-template-lint: https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/plugins.md
- noted that I might not be able to test in dummy app: based on [a blog post expanding on the instructions in the add-on](https://www.mehulkar.com/blog/2020/03/ember-template-lint-rule-from-addon/#:~:text=Using%20the%20plugin%20in%20an%20app)

I also had to do a couple of other things: 

- change the extension of `.template-lintrc.js` to `.cjs`
- change the extensions of the custom-rules to `.mjs`

TODO: 
- add a test (incl some failing scenarios so I know things aren't just failing silently)
- should the plugin need to be a separate package?
- what instructions do I need to add for consuming apps so they have the linting rules available to them?
- once I know that everything is working, submit a PR to update the docs for ETL to include this information


Dhaulagiri:
> should the plugin need to be a separate package?

it seems like that may be cleaner, especially with respect to the test setup. On the other hand, its rules are _only_ going to be useful in conjunction with the component library.

In either case, what would be useful to experiment with is the experience of actually consuming this from another app to see if however we're setting this up actually works how we want it to.

MelSumner:
I am going to close this PR for now, since we need some analysis first. Specifically, on what things we'd have coverage for with ember-a11y-testing (axe-core) and wouldn't need lint coverage for.

PR #529, opened 2022-08-12
Fix required attribute on form fields

alex-ju:
### :pushpin: Summary

Fix `required` attribute on form fields

We currently set the `required` attribute on a form control based on the `@isRequired` argument (which is used to apply a visual indicator to the label element), but there are cases where the native validation is used, but the indicator is not required. To accommodate for this cases, we allow the required attribute to be set at the component level.

### :link: External links

Merging this will help with [form component migration in Waypoint](https://github.com/hashicorp/cloud-ui/pull/3281).

***

### 👀 How to review

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/fast-bugs-admire.md
+++ b/.changeset/fast-bugs-admire.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Fix `required` attribute on form fields
--- a/packages/components/addon/components/hds/form/checkbox/field.hbs
+++ b/packages/components/addon/components/hds/form/checkbox/field.hbs
@@ -12,10 +12,10 @@
       class="hds-form-field__control"
       @value={{@value}}
       name={{@name}}
+      required={{@isRequired}}
       ...attributes
       id={{F.id}}
       aria-describedby={{F.ariaDescribedBy}}
-      required={{@isRequired}}
     />
   </F.Control>
 </Hds::Form::Field>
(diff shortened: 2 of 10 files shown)


KristinLBradley (approved):

MelSumner:
@alex-ju do you think we should make it optional to show the indicator? I remember when we talked about this previously, we discussed that it would likely be confusing to the developer if they'd set `@isRequired` but then had no styling support. 

Also I think we need a follow-up PR, if you are removing this support, the documentation would need an update too.

alex-ju:
> @alex-ju apologies, just catching up a little from last week. Do you think we should make it optional to show the indicator? I remember when we talked about this previously, we discussed that it would likely be confusing to the developer if they'd set `@isRequired` but then had no styling support.
> 
> Also I think we need a follow-up PR, if you are removing this support, the documentation would need an update too.

To clarify, the way `@isRequired` is applied to a component doesn't change (if set to `true` it will apply the indicator to the label and will set the `required` attribute to the associated form control). What changed with this PR is that we allow for the `required` attribute to be set to `true` without requiring the indicator to be shown. And we do this because we want to accommodate the pattern where the fields are implicitly required and only the optional fields are marked with an indicator.

Do let me know if you think the documentation requires updates, but skimming through it all sounds valid still.

MelSumner:
@alex-ju ahh okay. I think I was a little confused how you could accomplish both things without a separate flag; thank you for explaining!

PR #534, opened 2022-08-16
Fix label indicator for assistive tech

alex-ju:
### :pushpin: Summary

Fix label indicator for assistive tech

There is currently an issue where Chrome (at least) merges the label text with the indicator text, which may cause potential issues to assistive tech, particularly screen readers. After trying a few options to preserve a space from code to browser (including an HTML encoded standard space `&#32;`) the only working solution I could find that generates a new text node (hence prevents the text from being merged without space) is the use of `&nbsp;` right after the label text.

While not proud of this solution, is the only way forward I could find.

<table>
<tr><td>

label indicator AT before

</td><td>

label indicator AT after

</td></tr>
</table>

### :link: External links

[Jira issue](https://hashicorp.atlassian.net/browse/HDS-548)

***

### 👀 How to review

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/orange-shirts-end.md
+++ b/.changeset/orange-shirts-end.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Fix label indicator for assistive technologies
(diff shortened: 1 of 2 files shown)


MelSumner (approved):
Thanks for working on this! FWIW, I don't think that there is probably any _better_ way to do it, it's just one of those things we gotta do.

PR #536, opened 2022-08-17
Added avatar component

MelSumner:
### :pushpin: Summary

 If merged, this PR adds the `<Hds::Avatar />` component.

- intends to replace [pds-avatar](https://unofficial-design-system-search.vercel.app/?path=/docs/structure-design-system_components-avatar--index)
- slightly modified to be more correct; rendered outcome (in the browser) should be the same as the PDS equivalent.

CleanShot 2022-08-17 at 11 39 03@2x

### :link: External links

JIRA ticket: https://hashicorp.atlassian.net/browse/HDS-511 

***

### 👀 How to review

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/app/styles/components/avatar.scss
```diff
@@ -0,0 +1,24 @@
...
+  display: inline-flex;
+  height: 32px;
+  justify-content: center;
+  width: 32px;
+}
+
+.hds-avatar img {
+  border-radius: 2px;
+  height: 32px;
+  width: 32px;
```
KristinLBradley, line 19: suggestion: Would you want to inherit the height and width values on the img? That way you wouldn't need to repeat them here.

Example:
`height: 32px;` => `height: inherit;`

packages/components/app/styles/components/avatar.scss
```diff
@@ -0,0 +1,24 @@
...
+}
+
+.hds-avatar svg {
+  display: inline-block;
+}
```
KristinLBradley, line 24: suggestion: You could make use of Sass's nested syntax to avoid repeating the parent selector.

packages/components/tests/dummy/app/templates/components/avatar.hbs
```diff
@@ -0,0 +1,76 @@
+{{page-title "Avatar Component"}}
+
+<h2 class="dummy-h2">Avatar</h2>
+<section>
+  <h3 class="dummy-h3" id="overview"><a href="#overview" class="dummy-link-section">§</a> Overview</h3>
+
+  <p class="dummy-paragraph">Renders an avatar; a user image if provided, otherwise a user icon by default.</p>
+
+</section>
```
KristinLBradley, line 9: super nit: Since content is indented, this blank line before the closing parent tag could be removed.

packages/components/addon/components/hds/avatar/index.hbs
```diff
@@ -0,0 +1,7 @@
+<div class="hds-avatar" ...attributes data-test-avatar aria-hidden="true">
+  {{#if @src}}
+    <img src={{@src}} alt="" role="none" data-test-avatar-img />
+  {{else}}
+    <FlightIcon @name="user-circle" @size="24" />
+  {{/if}}
+</div>
```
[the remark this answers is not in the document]
MelSumner: I considered it; but considering that I was trying to do as close of a 1:1 port as possible, _and_ we don't have an HTML validator in our build yet, _and_ this markup is similar to other markup that we have in our components, I decided to leave the divs there. Definitely open to having a team discussion and changing this across the board in a future PR, though. 👍

packages/components/addon/components/hds/avatar/index.hbs
```diff
@@ -0,0 +1,7 @@
+<div class="hds-avatar" ...attributes data-test-avatar aria-hidden="true">
```
alex-ju, line 1:
```suggestion
<div class="hds-avatar" ...attributes aria-hidden="true">
```

packages/components/addon/components/hds/avatar/index.hbs
```diff
@@ -0,0 +1,7 @@
...
+  {{#if @src}}
+    <img src={{@src}} alt="" role="none" data-test-avatar-img />
```
alex-ju, line 3:
```suggestion
    <img src={{@src}} alt="" role="none" />
```

alex-ju (approved):
I may be wrong, but we're currently using ids in our tests so we may want to remove the data attributes and leave the consumer define them if needed.

Otherwise looks good – great choice of sample avatar! 🐱

packages/components/addon/components/hds/avatar/index.hbs
MelSumner: @alex-ju for our own tests, yes; I left these in because 1. they were in the PDS component and 2. there could be consumers using these data attributes in their own tests. WDYT?

alex-ju: Currently only `atlas` would benefit from them; I'm a bit skeptical as we've not done this with other components, but also not too opinionated on this.

packages/components/addon/components/hds/avatar/index.hbs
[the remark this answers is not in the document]
KristinLBradley: Thanks for the background information. 

It would be great if we could look into improving markup in the future to try to validate. I understand we have other priorities currently though. :)

PR #538, opened 2022-08-19
DRAFT: Gutters

KristinLBradley:
### :pushpin: Summary
**DRAFT PR:** Spike for proposed “Gutters” layout component to add spacing between child elements. [Refer to CRD](https://docs.google.com/document/d/1ZQ5IbHaqWCsAcJ5KdiQmaiYQySVlrX78OIvwMMLZQvE/edit#) for more background details.

_Please review & add your comments & suggestions. This is still a Work in Progress._

For this PR, I'm exploring both creating a component and exposing the underlying CSS classes as helper classes to give end-users an option as to how they make use of the layout pattern.

This is still a work in progress and is my first fuller-fledged Ember component that has actual arguments and tests. I would appreciate your feedback on how to improve the component and ways to make it more useful (if in fact you do think it would be useful).

Also please let me know if I've violated any Ember best practices or any general code tips for how to improve anything.

* See: https://hds-components-git-hds-550-gutters-component-hashicorp.vercel.app/
  * Added "CSS helper classes" section as part of exploring how to expose helper classes for the Gutters component 
  * I also added a separate section for "Layout" components to separate them from general components
* [Gutters component documentation](https://hds-components-git-hds-550-gutters-component-hashicorp.vercel.app/components/gutters)
* [Gutters helper classes documentation](https://hds-components-git-hds-550-gutters-component-hashicorp.vercel.app/helpers/gutters)

(I experimented with adding a new section for helper classes to make them easier to find. Also added a new section to distinguish Layout components from others.)

Updated to move CSS Helpers under Foundation section:
Screen Shot 2022-08-19 at 3 42 36 PM

### :link: External links
* [Component Requirements Document](https://docs.google.com/document/d/1ZQ5IbHaqWCsAcJ5KdiQmaiYQySVlrX78OIvwMMLZQvE/edit#)

***

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/addon/components/hds/gutters/index.js
```diff
@@ -0,0 +1,95 @@
+import Component from '@glimmer/component';
+import { htmlSafe } from '@ember/template';
+import { assert } from '@ember/debug';
+
+export const DIRECTIONS = ['horizontal', 'vertical'];
+export const SPACING_SIZES = ['xxs', 'xs', 'sm', 'md', 'lg', 'xl', 'xxl'];
+
+const regEx = /^((\d+)|(\d+\.\d+)|(\.\d+))(px|rem|em|%)$/;
```
KristinLBradley, line 8: This is to check for valid css units consisting of a number and some of the most common CSS units. We may want to do something more sophisticated and/or support more types of units.

packages/components/addon/components/hds/gutters/index.js
```diff
@@ -0,0 +1,95 @@
...
+
+export default class HdsGuttersIndexComponent extends Component {
+  /**
+   * Get the direction for the component
+   * Accepted values: 'horizontal', 'vertical'
+   * @param direction
+   * @type {string}
+   * @default ? // TODO: Determine if a default should be set & if so which value
```
KristinLBradley, line 16: I wasn't sure if it would be helpful to set a default value or not for the direction.

packages/components/tests/dummy/app/styles/pages/db-gutters.scss
```diff
@@ -0,0 +1,47 @@
...
+    --hds-spacing-xs: 8px;
+    --hds-spacing-sm: 12px;
+    --hds-spacing-md: 16px;
+    --hds-spacing-lg: 24px;
+    --hds-spacing-xl: 36px;
+    --hds-spacing-xxl: 48px;
+}
+
+// TODO: Determine best default spacing size
+$default-spacing: var(--hds-spacing-md); // set as Sass variable to not expose
```
KristinLBradley, line 16: This seemed like an appropriate default spacing size but could be changed if there's a better default to use.

packages/components/tests/dummy/app/styles/pages/db-gutters.scss
```diff
@@ -0,0 +1,47 @@
...
+
+.hds-gutters {
+    // vertical (make default ?): (Find out most common use cases)
+    // &:not([class*="hds-gutters-horizontal"]), // yes/ no?
```
KristinLBradley, line 20: I wasn't sure if "direction" should be a required argument or if a default should be set.

packages/components/tests/dummy/app/templates/components/gutters.hbs
```diff
@@ -0,0 +1,186 @@
...
+    <dt>direction</dt>
+    <dd>
+      <p>vertical, horizontal (Default, if any, to be determined)</p>
+    </dd>
+
+    <dt>TOOD: tag/inline/block</dt>
+    <dd>
+      <p>Needs more thought. (To prevent invalid markup, users would be able to choose between using a div tag (the
+        default) or a span tag.)</p>
+    </dd>
```
KristinLBradley, line 63: I would like to give users the option to use a span tag instead of the default div tag for the component but not sure what to call the argument.

packages/components/tests/dummy/app/templates/components/gutters.hbs
```diff
@@ -0,0 +1,186 @@
...
+      >Figma UI Kit</a>
+    </p>
+  </div>
+</section>
+
+<section data-test-percy>
+  <h3 class="dummy-h3" id="showcase">
+    <a href="#showcase" class="dummy-link-section">§</a>
+    Showcase
+  </h3>
```
KristinLBradley, line 142: Showcase examples could probably be improved if there are suggestions.

heatherlarsen:
@KristinLBradley I noticed you put these in a new section on the homepage of the scrappy site called "CSS helper classes", but then mention the Foundations section in the description. Is there a reason these can't go directly under the Foundations heading?

KristinLBradley:
> @KristinLBradley I noticed you put these in a new section on the homepage of the scrappy site called "CSS helper classes", but then mention the Foundations section in the description. Is there a reason these can't go directly under the Foundations heading?

It could go under there. I wasn't sure of the best place to put it. The Foundations pages seemed to be more about design tokens though although they also contain related helper classes. I wanted to make it clearer where to find helper classes and to distinguish the Gutters component from the Gutters helper classes.

KristinLBradley:
> > @KristinLBradley I noticed you put these in a new section on the homepage of the scrappy site called "CSS helper classes", but then mention the Foundations section in the description. Is there a reason these can't go directly under the Foundations heading?
> 
> It could go under there. I wasn't sure of the best place to put it. The Foundations pages seemed to be more about design tokens though although they also contain related helper classes. I wanted to make it clearer where to find helper classes and to distinguish the Gutters component from the Gutters helper classes.

I moved "CSS helper classes" to under the Foundations section now.

packages/components/addon/components/hds/gutters/index.hbs
```diff
@@ -0,0 +1,3 @@
+<div class={{this.classNames}} style={{this.customSpacing}} ...attributes>
```
didoo, line 1: In Ember this (using an inline style attribute) is considered an anti-pattern for security reasons (that's what I've been told).

Here you can see an early implementation where we used it:
https://github.com/hashicorp/design-system/blob/c9b9f12a1071b29c607e76688a6eada0d4ec289f/packages/components/addon/components/hds/breadcrumb/item.js#L7-L38
https://github.com/hashicorp/design-system/blob/c9b9f12a1071b29c607e76688a6eada0d4ec289f/packages/components/addon/components/hds/breadcrumb/item.hbs#L1

and here you can see how we ended up doing it in the components after that feedback, using a `style` ember helper:
https://github.com/hashicorp/design-system/blob/c9b9f12a1071b29c607e76688a6eada0d4ec289f/packages/components/addon/components/hds/dropdown/index.hbs#L11

Another alternative approach in this case, since you're setting a CSS custom prop, could be to using the `setProperty ` method in JS to sets the CSS prop value directly on the element (under the hood, it creates an inline style attribute via JS).

packages/components/addon/components/hds/gutters/index.js
```diff
@@ -0,0 +1,99 @@
+import Component from '@glimmer/component';
+import { htmlSafe } from '@ember/template';
+import { assert } from '@ember/debug';
+
+export const DIRECTIONS = ['horizontal', 'vertical'];
+export const SPACING_SIZES = ['xxs', 'xs', 'sm', 'md', 'lg', 'xl', 'xxl'];
+
+const regEx = /^((\d+)|(\d+\.\d+)|(\.\d+))(px|rem|em|%)$/;
```
didoo, line 8: If we use the `style` Ember helper, it automatically validates the CSS unit for us (if the unit is not correct, it simply ignores the declaration is my understanding).

Reason: we avoid making a choice between which CSS units to support and which to leave out.

packages/components/addon/components/hds/gutters/index.js
```diff
@@ -0,0 +1,99 @@
...
+      '@direction for "Hds::Gutters" must have a valid value',
+      DIRECTIONS.includes(direction)
+    );
+
+    return direction;
+  }
+
+  /**
+   * Get the preset spacing size to apply to the item.
+   * Accepted values: 'xxs', 'xs', 'sm', 'md', 'lg', 'xl', 'xxl'
```
didoo, line 31: To be discussed if we want to re-introduce a spacing scale (time ago the decision was not to, and I would like to see strong arguments to justify this change).

packages/components/addon/components/hds/gutters/index.js
```diff
@@ -0,0 +1,99 @@
...
+    if (SPACING_SIZES.includes(spacing)) {
+      return spacing;
+    }
+
+    return undefined;
+  }
+
+  /**
+   * Get the inline custom spacing style to apply to the item.
+   * Accepted values: '1px', '2.4rem', '3em', '2%', etc. (only positive numbers)
```
didoo, line 56: My understanding is that the "only positive numbers" is not a technical constraint, right? In which case what are the benefits of preventing the user to have negative margins (in case they need it, which is unlikely, but still...)

packages/components/addon/components/hds/gutters/index.js
```diff
@@ -0,0 +1,99 @@
...
+   * Get the class names to apply to the component.
+   * @method classNames
+   * @return {string} The "class" attribute to apply to the component.
+   */
+  get classNames() {
+    let classes = ['hds-gutters'];
+
+    // add a class based on the @spacing argument
+    if (this.presetSpacing) {
+      classes.push(`hds-child-spacing-${this.presetSpacing}`);
```
didoo, line 91: if the namespace is `hds-gutters` probably this class should be `hds-gutters-child-spacing-xxx`? (and probably at that point I would omit the `child`, as in `hds-gutters-spacing-xxx` since is applied to the parent container anyway)

packages/components/addon/components/hds/gutters/index.js
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    // add a class based on the @direction argument
+    classes.push(`hds-gutters-${this.direction}`);
```
didoo, line 95: Nitpick: probably I would move it before the "spacing" class (mentally I would thing of it as "gutter > vertical > lg")

packages/components/tests/dummy/app/router.js
```diff
@@ -49,4 +50,7 @@ Router.map(function () {
     this.route('disclosure');
     this.route('interactive');
   });
+  this.route('helpers', function () {
+    this.route('gutters');
+  });
```
didoo, line 55: Let's discuss about the pages organization, there are reasons why we may want to consider an alternative content structure.

packages/components/tests/dummy/app/styles/pages/db-gutters.scss
```diff
@@ -0,0 +1,47 @@
...
+// (See: https://www.figma.com/file/oQsMzMMnynfPWpMEt91OpH/HDS-Product---Foundations?node-id=2008%3A3)
+:root {
+    --hds-spacing-xxs: 4px;
+    --hds-spacing-xs: 8px;
+    --hds-spacing-sm: 12px;
+    --hds-spacing-md: 16px;
+    --hds-spacing-lg: 24px;
+    --hds-spacing-xl: 36px;
+    --hds-spacing-xxl: 48px;
+}
```
didoo, line 13: The intent was **not** to make then generic design tokens. If we define them these way (and if we centralize them) then we revert this decision, that I would prefer not to (unless there's strong evidence about this need).

packages/components/tests/dummy/app/styles/pages/db-gutters.scss
didoo: let's avoid to use `:not` pseudo selectors, they created havoc when used before :)

packages/components/tests/dummy/app/styles/pages/db-gutters.scss
```diff
@@ -0,0 +1,47 @@
...
+
+// TODO: Determine best default spacing size
+$default-spacing: var(--hds-spacing-md); // set as Sass variable to not expose
+
+.hds-gutters {
+    // vertical (make default ?): (Find out most common use cases)
+    // &:not([class*="hds-gutters-horizontal"]), // yes/ no?
+    &-vertical {
```
didoo, line 21: re. our BEM approach, we are not using this technique for concatenation, we just rely on flat declarations

packages/components/tests/dummy/app/styles/pages/db-gutters.scss
```diff
@@ -0,0 +1,47 @@
...
+      > * {
+        &:not(li) {display: block;}
```
didoo, line 23: see comment above about the `:not()` selector

plus I am not sure why we need to intervene on the `display` property of the children (potentially impacting the layout of components used as children, see attached video)

https://user-images.githubusercontent.com/686239/186160262-6c469245-ac19-435a-8835-1f36770a22b1.mov

packages/components/tests/dummy/app/templates/components/gutters.hbs
didoo: ok, let's discuss what options do we have.

packages/components/tests/dummy/app/templates/components/gutters.hbs
didoo: No problem, once the ticket is in a more advanced state I can help with this

packages/components/tests/integration/components/hds/gutters/index-test.js
```diff
@@ -0,0 +1,67 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | hds/gutters/index', function (hooks) {
```
didoo, line 6: nice tests implementation!  🙌

didoo (commented):
@KristinLBradley left a few initial comments, before our meeting today

packages/components/addon/components/hds/gutters/index.hbs
KristinLBradley: Ok thanks.

I used this: `htmlSafe(--hds-spacing: ${spacing})` which I thought would handle security issues but I'll investigate the items mentioned in your feedback.

packages/components/addon/components/hds/gutters/index.js
KristinLBradley: Without a strong use case for negative margins, I wanted to leave them out to prevent possible mis-use. I will update to allow negative values.

packages/components/addon/components/hds/gutters/index.js
KristinLBradley: I wanted to be able to use the same CSS properties to set spacing in the upcoming “Layout” component I have planned. (Also I was hoping to not use overly long names for things.)

I'll leave as is for now but will consider it more.

packages/components/tests/dummy/app/templates/components/gutters.hbs
KristinLBradley: I'm now thinking I will call the argument: "isSpan". The default would be to use a div tag. I'll work on this...

packages/components/tests/integration/components/hds/gutters/index-test.js
KristinLBradley: Thanks!

packages/components/app/styles/components/gutters.scss
```diff
@@ -0,0 +1,51 @@
...
+
+// TODO: Determine best default spacing size
+$default-spacing: var(--hds-spacing-md); // set as Sass variable to not expose
+
+.hds-gutters {
+  // vertical (make default ?): (Find out most common use cases)
+  // &:not([class*="hds-gutters-horizontal"]), // yes/ no?
+  &-vertical {
+    > * {
+      &:not(li) {display: block;}
```
KristinLBradley, line 27: I wanted to avoid using flex box for this simple component (at least for first iteration) but I see that changing the display of the children like this has potential issues so I'm going to try using flexbox after all...

Btw, after I get this “Gutters” component into better shape and have an API others agree with we still don't have to necessarily merge it right away (or at all). I could work on the more complex “Layout” component I have planned next using what we've agreed upon from this one. (I wanted to start with this simpler component first to use as a basis for the more complex one.)

packages/components/tests/dummy/app/templates/components/gutters.hbs
KristinLBradley: Added new isSpan argument

packages/components/addon/components/hds/gutters/index.js
KristinLBradley: Updated to allow negative numbers

packages/components/addon/components/hds/gutters/index.js
KristinLBradley: I think you're right. The `hds-gutters-spacing-xxx` should still make sense to the other planned Layout component so I updated the naming.

packages/components/addon/components/hds/gutters/index.js
KristinLBradley: ### Background behind my original thinking:

Currently within the old Structure library and within Cloud-ui code, they are using spacing scales following very similar naming: "xs", "sm", "md", etc. so this type of naming isn't a new or unfamiliar concept to devs.

The spacing scale we have in Figma uses numbers as names instead of letters but we aren't currently referencing it directly anywhere; it doesn't get exported as tokens. So we could easily add naming such as "xs", "sm" etc. to the Figma file in addition to the numbers or instead of them if we are worried about potential confusion.

### Trying to use px numbers for spacing size names instead...

Currently, for users to use one of the preset spacing sizes with the Gutters component, they just need to add the short abbreviation name.

**Example:**  
```
<Hds::Gutters @direction="horizontal" @spacing="xs">
	<span>item 1</span>
	<span>item 2</span>
	<span>item 3</span>
</Hds::Gutters>
```

If I change the spacing size names to use numbers based on px instead, then users would enter a plain number.

**Example:**  
`<Hds::Gutters @direction="horizontal" @spacing="4" />`

It then becomes confusing what "4" means. Does it mean a pixel value? Or does it mean size #4 in the HDS Figma file which is actually 16px?

To try to prevent this confusion, we could include "px" as part of the spacing names.

**Example:**  
`<Hds::Gutters @direction="horizontal" @spacing="4px" />`

But then there is no difference between using a preset size vs. entering a custom CSS unit size.

So then we have to think of other ways to solve the confusion. Do we require users to enter the full variable names? We could, but then the component becomes harder to use and users need to type more, have more chance of typing the wrong name, or have to constantly look up the names for the spacing scale.

**Example:**  
`<Hds::Gutters @direction="horizontal" @spacing="hds-spacing-4px" />`

To me it seems that using short descriptive abbreviations that devs are already familiar with would make the components simpler and easier to use. Please let me know if you have other ideas I haven't considered though.

packages/components/addon/components/hds/gutters/index.js
KristinLBradley: I'll work on updating my code, thanks

packages/components/tests/dummy/app/styles/pages/db-gutters.scss
KristinLBradley: I can change it but I like to make use of sass nesting to keep things organized and avoid repeating the same selectors for shorter groups of styles.

(Btw, I didn't really follow BEM here since the "parent" class name doesn't really do anything. The other classes can be used on their own as helper classes.)

packages/components/tests/dummy/app/styles/pages/db-gutters.scss
KristinLBradley: I have since updated this to remove the display styles for children, thanks.

packages/components/tests/dummy/app/styles/pages/db-gutters.scss
KristinLBradley: Updated

didoo: > I can change it but I like to make use of sass nesting

Absolutely! It's one of the most powerful thing of using a CSS pre-processor! What I meant using the `&` to construct BEM names/selectors.

packages/components/addon/components/hds/gutters/index.hbs
KristinLBradley: @didoo I'm trying to use setProperty() but I can't figure out how to access the specific element (div or span parent element) to set the style property. Is there a way to do that?

packages/components/addon/components/hds/gutters/index.js
KristinLBradley: @didoo Do you have any comments or suggestions? I tried to figure out how to follow what you wanted but it seems like it introduces problems and confusion of its own (at least in my thinking). Do you have any ideas on what would make sense?

packages/components/addon/components/hds/gutters/index.js
```diff
@@ -0,0 +1,108 @@
...
+  /**
+   * Get the preset spacing size to apply to the item.
+   * Accepted values: 'xxs', 'xs', 'sm', 'md', 'lg', 'xl', 'xxl'
+   * @param spacing
+   * @method presetSpacing
+   * @type {string} The class name to apply to the item.
+   * @default handled by CSS
+   */
+  get presetSpacing() {
+    let { spacing } = this.args;
```
KristinLBradley, line 47: The spacing argument allows 2 types of values:
1. A preset spacing size. (Currently one of: "xxs", "xs", "sm", "md", "lg", "xl", "xxl")
2. A custom CSS unit. (Examples: "2.5rem", "4%", "1em", "-5px", etc.)

Depending on the type of value entered it's handled 2 different ways:
1. Preset size = appends a classname to set the spacing size
2. Custom CSS unit = Adds an inline style modifying the value of a CSS custom property.

Since "spacing" works differently depending on the type of value entered, it adds a few complications. For example: if a preset size is entered, then a class name should be added but a style attribute should not be. Conversely, if a custom CSS unit is entered, then a style attribute should be added but a class name should not be. Also if no "spacing" vale is added, then neither a class name nor a style attribute should get added to the component's underlying HTML element.

The way this is currently implemented, the different cases are taken care of. I tried to refactor to use {{style}} instead but couldn't figure out how to get it to work to handle all the various cases. I've also tried using JS .setProperty() instead to add a style attribute but I don't know how the HTML element should be targeted.

packages/components/addon/components/hds/gutters/index.hbs
```diff
@@ -0,0 +1,9 @@
+{{#if this.isSpan}}
+  <span class={{this.classNames}} style={{this.customSpacing}} ...attributes>
```
KristinLBradley, line 2: See: https://github.com/hashicorp/design-system/pull/538/files#r955427462

packages/components/addon/components/hds/gutters/index.js
didoo: > So we could easily add naming such as "xs", "sm" etc. to the Figma file in addition to the numbers or instead of them if we are worried about potential confusion.

True. But this still would not solve the problem of: _"As a developer, I am inspecting a Figma file with the design specs. I see that the spacing between two elements is 4px. Which size alias should I use? xs? sm? other?"_ The only way is 1) I know it by heart or 2) I have to look out for this info somewhere.

> It then becomes confusing what "4" means. Does it mean a pixel value? Or does it mean size https://github.com/hashicorp/design-system/pull/4 in the HDS Figma file which is actually 16px?

No one is using the size scale in Figma. Not even the designers. We could remove that scale and probably no one would notice. So I don't think there would be confusion, or at least would be much less than with any alternative approach we could use.

> To try to prevent this confusion, we could include "px" as part of the spacing names.
> But then there is no difference between using a preset size vs. entering a custom CSS unit size.

**Exactly** the point I wanted to make during our discussion! Given these "boundary conditions", why do you need a "scale"? Just let the users enter the spacing as they want (including `px`) and you reduce **by a lot** the overall complexity of this component.

All the complexity is rising because of the introduction of an artificial spacing scale that acts as a proxy between the user and the actual value they need. As soon as one removes that scale, and the proxy/mapping, then it's immediate the relation between what they see in Figma (`4px`), the value they need to use in code (`4px`), the value that will be used in the component (`4px`) and the value that will be rendered in the browser (`4px`; until we use a post-css processor that converts this to `rem`s, but this is for the future).

I know this approach is an anti-pattern. It took me quite a while to accept it too, but once you get used to it you see that makes sense: we haven't had any issue since we decided to follow this approach almost one year ago.

packages/components/addon/components/hds/gutters/index.js
didoo: > I tried to refactor to use {{style}} instead but couldn't figure out how to get it to work to handle all the various cases.

Use the [`style` modifier](https://github.com/jelhan/ember-style-modifier) to assign the value of `--hds-spacing` as inline style: 
```
// in .hbs file
<[div|span] class={{this.classNames}} {{style --hds-spacing=this.customSpacing}} ...attributes>
```
```
// in .js file
if (spacing !== undefined) {
  if (spacing.match(regEx)) {
    return spacing;
  }
}
```
Personally, I would also merge `presetSpacing` and `customSpacing` in a single function, that returns the spacing value (without the need for CSS classes). Of course, this also takes away the possibility of using CSS helper classes, but as mentioned elsewhere for other components we're not providing CSS classes that can be used directly as "helpers" so this would not be a bug issue (for me).

> I've also tried using JS .setProperty() instead to add a style attribute but I don't know how the HTML element should be targeted.
I've noticed that internally the [`style` modifier](https://github.com/jelhan/ember-style-modifier) is using the `setProperty()` method, so technically it would be the same as using the modifier; let's use the modifier then, which is more explicit on what it does.
https://github.com/jelhan/ember-style-modifier/blob/master/addon/modifiers/style.js#L57

packages/components/addon/components/hds/gutters/index.js
KristinLBradley: @meirish commented on my CRD that the blocking issue at a past attempt to create a layout component was the lack of a spacing scale.

didoo: He doesn't remember correctly: the blocking issue was lack of time for all of us to continue (we were supposed to continue working on that PR/draft together, but we never did).

KristinLBradley: Ok. It would be nice to start to establish some some of consistent values for spacing within layouts still to my thinking. Also I'd like to at some point be able to switch to using rem units for spacing sizes instead of pixels. I don't know how realistic either of those are though...

packages/components/addon/components/hds/gutters/index.hbs
didoo: > I used this: htmlSafe(--hds-spacing: ${spacing}) which I thought would handle security issues

I did the same (time ago) but then I was told that it doesn't do what one would expect, so the security issue remains using this approach (that's what I've been told).

> I'm trying to use setProperty() but I can't figure out how to access the specific element 
As far as I remember you have to assign an ID to the element, and then use the ID to get the DOM node, and then use target that DOM node. It may be that there is a more straightforward way (I tried long time ago) but I am not sure.

But why not use the `style` modifier as suggested in this comment?
https://github.com/hashicorp/design-system/pull/538#discussion_r958296215
It does all the heavy lifting for you (it's built on top of `setProperty()`).

packages/components/addon/components/hds/gutters/index.js
didoo: > I'd like to at some point be able to switch to using rem units for spacing sizes instead of pixels. 

As mentioned before, this will be done automatically (in the future) using a CSS post-processor. Sizes will still be declared in `px` in the Sass source code.

packages/components/addon/components/hds/gutters/index.hbs
KristinLBradley: Thanks. I tried the style modifier first but I couldn't get it to work how I needed. It doesn't validate the value since I'm trying to set a css custom property value.

packages/components/addon/components/hds/gutters/index.js
KristinLBradley: This is exactly what I tried but coudn't get to work right: `{{style --hds-spacing=this.customSpacing}}`

Anyway I'll come back to this and experiment more when I can, thanks

packages/components/addon/components/hds/gutters/index.js
KristinLBradley: Ok

didoo:
@KristinLBradley since the corresponding JIRA ticket is closed, maybe we could close this PR too (it's a spike) and we can pick up from where we left when we will work again on this. What do you think?

PR #539, opened 2022-08-19
Add custom styles for `ember-power-select` addon

alex-ju:
### :pushpin: Summary

Add custom styles for `ember-power-select` addon

 - Add `ember-power-select` styles (these overrides are used to align the default `ember-power-select` styles with the rest of HDS form components and have also been tested with custom implementations in `cloud-ui`)
 - To be able to manually test these in this repo, I've added `ember-power-select` as a dev dependency and created a showcase page in the dummy app.

Example of basic usage in the [preview link](https://hds-components-git-add-power-select-styles-hashicorp.vercel.app/overrides/power-select)

Example of custom usage in `cloud-ui` with the `hvn-select` component (local development captures)

Default
01 hvn-select default

Focus
02 hvn-select focus

Expanded
03 hvn-select open

### :link: External links

[Jira ticket for PowerSelect](https://hashicorp.atlassian.net/browse/HDS-516)
[Jira ticket for PowerSelectMultiple](https://hashicorp.atlassian.net/browse/HDS-565)

***

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ ] +1 Percy if applicable


MelSumner (approved):
:shipit: no objections from what I can tell. 👍

heatherlarsen (approved):
🎨 🎉 🙌

meirish (changes requested):
Is this meant to style _all_ instances of power-select? Or just the HVN picker in cloud-ui? I think most of our apps are using power-select in some manner, so this change might have a bigger blast radius than you expect.

Just want to call that out before it's merged - apologies if you've already considered that when making this change!

packages/components/ember-cli-build.js
```diff
@@ -9,8 +9,12 @@ module.exports = function (defaults) {
       precision: 4,
       includePaths: [
         '../../node_modules/@hashicorp/design-system-tokens/dist/products/css',
+        '../../node_modules/ember-power-select/vendor',
       ],
     },
+    'ember-power-select': {
+      theme: false,
```
meirish, line 16: Do apps need to do this in their `ember-cli-build.js` too? If so, could this be added to the docs?

meirish:
Oh I clicked through the preview, and it looks like the styles work on the normal select too 🎉 . Do you intend to style the other modes too? with the multiselect, search, etc?

alex-ju:
> Is this meant to style _all_ instances of power-select? Or just the HVN picker in cloud-ui? I think most of our apps are using power-select in some manner, so this change might have a bigger blast radius than you expect.

> Just want to call that out before it's merged - apologies if you've already considered that when making this change!

> Oh I clicked through the preview, and it looks like the styles work on the normal select too 🎉 . Do you intend to style the other modes too? with the multiselect, search, etc?

It is meant to style all generic instances. I have studied the addon for a bit and tested it in 'power-select' and 'basic-dropdown' modes, and with custom markup, namely `triggerComponent`, `selectedItemComponent`, `beforeOptionsComponent` (this is where the search usually sits), `optionsComponent`, `afterOptionsComponent` and `searchMessageComponent`. I've only tested it in `cloud-ui`, but will try and spin up `nomad`, `consul` and `atlas` which contain the other instances we identified. It doesn't cover multisearch (`PowerSelectMultiple`) which is currently only found in `atlas`. I will have to get back to the team on this before committing to an answer.

packages/components/ember-cli-build.js
alex-ju: As far as I can tell, this is [the default configuration](https://github.com/hashicorp/cloud-ui/blob/master/hcp/ember-cli-build.js#L147-L149), but yes it's worth mentioning that these styles assume you use `ember-power-select` with the default theme.

packages/components/tests/dummy/app/controllers/components/power-select.js
```diff
@@ -0,0 +1,9 @@
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+
+export default class PowerSelectController extends Controller {
+  // notice: this is used as "noop" function for the onDismiss callback of the Alert component
```
didoo, line 5: the comment refers to the "Alert" component, but this is the "PowerSelect" one

packages/components/app/styles/components/power-select.scss
```diff
@@ -0,0 +1,183 @@
...
+    background-size: var(--token-form-select-background-image-size) var(--token-form-select-background-image-size);
+    border: var(--token-form-control-border-width) solid var(--token-form-control-base-border-color-default);
+    border-radius: var(--token-form-control-border-radius);
+    box-shadow: var(--token-elevation-low-box-shadow);
+    color: var(--token-form-control-base-foreground-value-color);
+    max-width: 100%;
+    min-height: 34px;
+    padding: calc(var( --token-form-control-padding) + 1px);
+    padding-left: 44px;
+    padding-right: calc(var( --token-form-control-padding) + 24px); // extra space for the icon
```
didoo, line 26: I LOVE how we can use the component-specific design tokens for this! 😍

packages/components/tests/dummy/app/templates/components/power-select/after-options.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="hds-power-select_after-options">
```
didoo, line 1: Is this file needed? 🤔

packages/components/tests/dummy/app/templates/components/power-select.hbs
```diff
@@ -0,0 +1,107 @@
+{{page-title "PowerSelect Component"}}
```
didoo, line 1: There's no link to this page in the "index", is it intentional? Otherwise probably should leave under "Forms" (something like `Form / PowerSelect` maybe?).

packages/components/tests/dummy/app/templates/components/power-select.hbs
```diff
@@ -0,0 +1,107 @@
...
+
+<h2 class="dummy-h2">PowerSelect</h2>
+
+<section>
+  <h3 class="dummy-h3" id="overview"><a href="#overview" class="dummy-link-section">§</a> Overview</h3>
```
didoo, line 6: Probably I would expand this overview, explaining that is not an HDS component but just the styling (we're essentially "re-skinning" it to make is more similar to the other form controls).

packages/components/tests/dummy/app/templates/components/power-select.hbs
```diff
@@ -0,0 +1,107 @@
...
+
+  <p class="dummy-paragraph">
+    A showcase with custom styles applied as overrides to the
+    <a target="_blank" rel="noopener noreferrer" href="https://ember-power-select.com/">ember-power-select addon</a>.
+  </p>
+
+</section>
+
+<section data-test-percy>
+  <h3 class="dummy-h3" id="showcase"><a href="#showcase" class="dummy-link-section">§</a> Showcase</h3>
```
didoo, line 16: Probably before the "Showcase" I would add a "How to use" section in which:
- explain that these styles assume you use ember-power-select with the default theme (see [comment here](https://github.com/hashicorp/design-system/pull/539#discussion_r951247990))
- explain how to use the specific `.hds-power-select` class (initially I thought it should be applied to the component itself, in reality it needs to be applied to a wrapping DOM node).
- explain how to use the `.hds-power-select_after-options`class (move the explanation from the showcase to the howto)

packages/components/tests/dummy/app/templates/components/power-select.hbs
```diff
@@ -0,0 +1,107 @@
...
+
+  <h4 class="dummy-h4">States</h4>
+  <p class="dummy-text-small">Default</p>
+
+  {{! This below is just an example of invocation, to get started }}
```
didoo, line 21: Is this comment still relevant? seems not related to the content below.

didoo (commented):
@alex-ju I did a first pass, and added a few comments. Once you've decided how to "distribute" the styles (via component, as normal styles, as extra Sass to import) let me know and I'll do a final review.

One extra thing: while reading the CSS code, I thought that I'm sure in your "journey" discovering all the internals and quirks of this component you gained a lot of knowledge. Maybe have a read at the CSS file with the styles and add comments where you think it would be helpful for someone in the future to know certain things :) (included your future self)

alex-ju:
> Oh I clicked through the preview, and it looks like the styles work on the normal select too 🎉 . Do you intend to style the other modes too? with the multiselect, search, etc?

@meirish an update on this: we found a few instances of `PowerSelectMultiple` only in atlas and terraform registry and having discussed this with the team we will provide custom styles for these as well.

packages/components/tests/dummy/app/templates/components/power-select.hbs
alex-ju: sure not, will remove it

packages/components/tests/dummy/app/templates/components/power-select/after-options.hbs
alex-ju: We use this to exemplify how a custom component/template block can be passed in. It's called in [one of the examples](https://github.com/hashicorp/design-system/pull/539/files/4c181a89655938a0d004db044bb82273140f0f12#diff-ac16ea0a6f51865dfe0a911f35693d355234436b89274cad9807836ab60e3a89R98).

packages/components/tests/dummy/app/templates/components/power-select.hbs
alex-ju: Was not sure where to place it at the time

packages/components/tests/dummy/app/templates/components/power-select/after-options.hbs
didoo: OK. I can't think of better alternatives, so let's leave it like this.

alex-ju:
This PR is good for another review.

@didoo I addressed your comments in this PR in conjunction with the [decisions agreed during the sync](https://docs.google.com/document/d/1xXxMv70KCOEmAe_l8mfzrnbQ3EGrw3v6Qt107RPyWuA/edit#bookmark=id.twbdrk2s4sxl).

In short in the additional commits I've:
 - moved the PowerSelect styles to a new 'Overrides' section (hopefully we won't have many of these) to reflect it's bespoke nature
 - added the 'Overview' and 'How to use' sections trying to better document these overrides
 - applied the stylelint rules with an exception on class names pattern (as we're not in control of these names)

@heatherlarsen I have updated this PR to style the PowerSelectMultiple based on your designs. There is one limitation: the way the tags are being marked-up by the vendor doesn't allow us to use the hover we have in our Tag component – otherwise I did my best to match the design, but let me know if there's something I missed or you'd want me to adjust.

packages/components/app/styles/@hashicorp/power-select-overrides.scss
```diff
@@ -0,0 +1,225 @@
+//
+// POWER-SELECT
+//
+// selectors is this file are defined to match original PowerSelect styles
+// and are meant to provide overrides through specificity
+//
+
+/* stylelint-disable selector-class-pattern */
```
didoo, line 8: 👍

packages/components/tests/dummy/app/templates/overrides/power-select.hbs
```diff
@@ -0,0 +1,284 @@
...
+      rel="noopener noreferrer"
+      href="https://github.com/hashicorp/design-system/blob/main/packages/components/README.md"
+    >design-system-components</a>
+    in your project the overrides import would look something like this:</p>
+  {{! prettier-ignore-start }}
+  {{! template-lint-disable no-unbalanced-curlies }}
+  <CodeBlock
+    @code='
+      @import "@hashicorp/design-system-components";
+      @import "@hashicorp/power-select-overrides";
```
didoo, line 42: _maybe_ I would be over-specific and explicit, and call the file `@hashicorp/design-system-power-select-overrides`

packages/components/tests/dummy/app/templates/index.hbs
```diff
@@ -175,6 +175,16 @@
...
+      <h2 class="dummy-h2">
+        Overrides
+      </h2>
+      <ol>
+        <li class="dummy-paragraph">
+          <LinkTo @route="overrides.power-select">
+            PowerSelect
+          </LinkTo>
+        </li>
+      </ol>
```
didoo, line 187: I like it!

didoo (approved):
all perfect! 🙌

(left only a couple of minor comments, non blocking)

heatherlarsen:
> the way the tags are being marked-up by the vendor doesn't allow us to use the hover we have in our Tag component

That's ok. These look great. Thanks Alex 👍

packages/components/tests/dummy/app/templates/overrides/power-select.hbs
alex-ju: agree, better over-specific than ambiguous – will update the file name (and references to it)

packages/components/tests/dummy/app/templates/overrides/power-select.hbs
```diff
@@ -0,0 +1,284 @@
...
+  <p class="dummy-paragraph">As these overrides rely on specificity, to apply them you need to wrap each PowerSelect
+    instance in an element with
+    <code class="dummy-code">hds-power-select</code>
+    class applied to it.</p>
+  <p class="dummy-paragraph">Invocation of the component with overrides would look something like this:</p>
+
+  <CodeBlock
+    @language="markup"
+    @code='
+    <div class="hds-power-select">
```
KristinLBradley, line 71: Question: Is it not possible/desirable to add a class name to the "PowerSelect" component itself vs. wrapping it with an extra div?

alex-ju: Great question, @KristinLBradley! I would've loved that to be possible, but applying a `class="hds-power-select"` attribute to a `<PowerSelect>` instance results in something like this:
```
<div class="ember-basic-dropdown">
  <div class="ember-view ember-basic-dropdown-trigger hds-power-select">...</div>
  <div class="ember-basic-dropdown-content">...</div>
</div>
```
So the class is not being applied to the outermost element, but to the dropdown trigger, which doesn't help with styling the dropdown content.

alex-ju: We could use adjacent sibling selector, but we needed some adjustments to the element that holds the `ember-basic-dropdown` hence I decided to use that additional wrapper element, which you can imagine I'm not fond of.

KristinLBradley: I see. Thanks for explaining.

KristinLBradley (approved):

PR #556, opened 2022-09-01
HDS-604 Separate out initial child components.

KristinLBradley:
### :pushpin: Summary
The purpose of this PR is to begin separating out the main child components (not all since some will be dependent upon further steps). Once approved, it will be merged into the [feature branch](https://github.com/hashicorp/design-system/pull/548).

**Completed so far:**
I separated out 2 main child components for each "Tab" & "TabPanel". I will most likely need to separate out other child components but I wanted to start with the absolute basic required ones first and then go from there. I took these as far as I can go without writing JS yet. 

**Question:** Should I start to write the JS as part of this PR or do it as a separate PR? (I don't know how much I should break out the different steps.)

**Next Steps:**
* Would now like to work with others to hammer down component naming and css class naming for the initial child components at this point ("Tab" & "TabPanel").
* Would like to implement JS for the arguments on the initial components but not sure if I should use this same PR?

### :link: External links

***

### 👀 How to review

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/addon/components/hds/tabs/index.hbs
```diff
@@ -2,68 +2,25 @@
 {{! template-lint-disable require-context-role }}
 
 <div class={{this.classNames}} ...attributes>
-  {{!-- {{yield}} --}}
-
   {{! Hds-TabList }}
```
KristinLBradley, line 5: Probably (?) eventually I'll need to separate out the Tab(s)List as another child component but not yet for implementing the initial basic functionality.

packages/components/addon/components/hds/tabs/index.hbs
```diff
@@ -2,68 +2,25 @@
...
 
 Component Structure:
-
-Tabs
-  TabList
-    Tab
+-----------------------
+Hds::Tabs
+  Hds::Tabs.TabList
+    Hds::Tabs.Tab
```
KristinLBradley, line 23: This is still for rough planning but I would like to start hammering down component naming and class names next. Initially I was trying to follow naming in the CRD and Figma but maybe I took that too strictly before?

packages/components/addon/components/hds/tabs/tab-panel.hbs
```diff
@@ -0,0 +1,11 @@
+{{! Hds-TabPanel }}
+<section
+  class="hds-tab-panel"
+  role="tabpanel"
+  {{!-- id="section1" --}}
+  {{!-- aria-labelledby="tab1" --}}
+  {{! hidden TODO: set by arg }}
```
KristinLBradley, line 7: I commented out all attributes with non-static values. Next step is to add these back in with underlying JavaScript.

packages/components/addon/components/hds/tabs/tab-panel.hbs
```diff
@@ -0,0 +1,11 @@
...
+  ...attributes
```
KristinLBradley, line 8: Added this temporarily. Don't know if it will be needed or even desirable for final implementation.

packages/components/app/styles/components/tabs.scss
```diff
@@ -9,14 +9,14 @@ $indicator-height: 3px;
 .hds-tabs {}
 
 // Sub-components:
-// TODO: MOVE!
+// TODO: MOVE?
```
KristinLBradley, line 12: Initially I was thinking child components should live in their own sub-folders with their own styles but maybe not really necessary... Will see.

packages/components/app/styles/components/tabs.scss
```diff
@@ -9,14 +9,14 @@ $indicator-height: 3px;
...
 .hds-tab-list {
   padding-bottom: round(calc($indicator-height / 2));
   // overflow-x: auto;
   // -webkit-overflow-scrolling: touch;
   // white-space: nowrap; // move
 
-  // Un-nest after moving
+  // Un-nest?
```
KristinLBradley, line 19: I can start refactoring class names now as people recommend...

My preference for BEM child class naming is to use sass nested structure like:
```
.parent-name {
  color: red;

  &--child-name {font-weight: bold;}
}
```
but if we don't like to use that I could unnest this to simplify the rendered css selectors.

packages/components/app/styles/components/tabs.scss
```diff
@@ -30,6 +30,7 @@ $indicator-height: 3px;
   position: relative; // for positioning temp indicator
 
   // temporary indicator
+  /* TODO: refactor
```
KristinLBradley, line 33: I commented this part out for now until I'm ready to start implementing the actual selector.

packages/components/tests/dummy/app/templates/components/tabs.hbs
```diff
@@ -194,5 +194,13 @@
...
 
-  <Hds::Tabs />
+  <Hds::Tabs as |T|>
+    <T.Tab id="tab1" href="#section1">One</T.Tab>
+    <T.Tab id="tab2" href="#section2">Two</T.Tab>
+    <T.Tab id="tab3" href="#section3">Three</T.Tab>
+    
+    <T.TabPanel id="section1" aria-labelledby="tab1">Content one</T.TabPanel>
+    <T.TabPanel id="section2" aria-labelledby="tab2">Content two</T.TabPanel>
+    <T.TabPanel id="section3" aria-labelledby="tab3">Content three</T.TabPanel>
```
KristinLBradley, line 204: I know for the final implementation we wouldn't want to require users to have to specify all these individual attributes/arguments. This is just for purposes of testing using child components before I start to write the JS to make a nicer API.

packages/components/addon/components/hds/tabs/tab.hbs
```diff
@@ -0,0 +1,17 @@
+{{! template-lint-disable require-context-role }}
+{{! template-lint-disable no-invalid-role }}
+{{! template-lint-disable link-href-attributes }}
```
KristinLBradley, line 3: This will be deleted after I implement more of the functionality.

packages/components/addon/components/hds/tabs/index.hbs
```diff
@@ -72,10 +28,9 @@ Tabs
         isSelected (boolean, default = false) // Select first tab by default?
         hasIcon (boolean, default = false)
         hasCount (boolean, default = false)
-    Tab (2+ required)
+    Hds::Tabs.Tab (2+ required)
     ...
     ScrollNavigation (displays if more tabs than fit container)
-  TabPanel
-  TabPabel
+  Hds::Tabs.TabPanel (2+ required, same # as Tab components)
```
didoo, line 34: Maybe, since now `tab`s and `panel`s are all yielded a siblings inside the `Tabs`, I would us just `Panel` instead of `TabPanel` (so `Hds::Tabs::Panel`). Unless we want to maintain the exact naming as its "role".

packages/components/addon/components/hds/tabs/tab-panel.hbs
didoo: Very likely we will keep it. We tend to allow consumers/users to pass attributes to the components (eg to add 'data-test' attributes or extra class names).

packages/components/app/styles/components/tabs.scss
didoo: > with their own styles

For context: generally we split the component templates/javascript files, but not the CSS files.

packages/components/app/styles/components/tabs.scss
didoo, line 19: Yes, we tend to make the CSS as flat (lowest specificity) as possible. 

_Given the naming convention we're using, there's zero risk of conflicts or overrides (unless intentional)._

didoo (approved):
Added a couple of minor comments, but for the purpose of the PR everything is OK.

packages/components/addon/components/hds/tabs/index.hbs
KristinLBradley: Updated naming. Refactored class names also to use BEM style

packages/components/app/styles/components/tabs.scss
KristinLBradley: Thanks, I will unnest. 

(Comment/Explanation: I like to keep css relatively flat but I like nesting BEM child element classes and modifiers so it's easier to see the structure and relationships. To still keep the CSS flat and avoid repeating the parent selector, I normally use partial class names for the child elements like `&__child`.)

didoo: Personally, I tend to prefer findability of classes in the code base. Some resources on this, that explain better what I mean:
https://csswizardry.com/2017/02/code-smells-in-css-revisited/#string-concatenation-for-classes
https://bradfrost.com/blog/post/sass-selectors-to-nest-or-not-to-nest/
https://bradfrost.com/blog/post/css-architecture-for-design-systems/

KristinLBradley: I haven't normally needed to search for classes in that way but anyway I unnested those styles.

PR #557, opened 2022-09-02
[DRAFT] `Hds::Text` component (2022)

didoo:
# 🚨 OUTDATED 🚨

### :pushpin: Summary

This is a work-in-progress. Essentially is a porting of the experiment/spike I did in Cloud UI https://github.com/hashicorp/cloud-ui/pull/2574

The idea is to introduce an `<Hds::Text>` component as helper to easily apply pre-defined typographic styles to the text.

### 🤔 FAQ

- Why not use [dynamic tags](https://rfcs.emberjs.com/id/0389-dynamic-tag-names/) instead of the large `switch` declaration?
  - see @gregone comment here: https://github.com/hashicorp/cloud-ui/pull/2574#issuecomment-1183296573 

### :link: External links

For reference see also:
- https://github.com/hashicorp/design-system-components/issues/21
- [RFC on helper components](https://docs.google.com/document/d/1aEDbxGTIIs4udumrWNjnnNPd4u1kcr9T_YSeTlec-3I/edit)
- https://github.com/hashicorp/cloud-ui/pull/2574

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


didoo:
For future reference: yesterday during the frontend guild @jgwhite suggested a possible alternative API for the component, on the line of
```
<HDS::Text @style="body/200/semibold" ... />
```
where the `style` matches exactly the name of the style in Figma (would be case insensitive though).

This has some pros:
- would be explicit for the developers (they could even copy&paste from the inspect panel in Figma)
- would reduce (or even eliminate) the need for alerts when the combination `style + weight` is not recommended by the HDS team

As well as some cons:
- too strictly binded to the naming in Figma: what if designers decide to change things? it already happened (the `/` represents a folder, so may be just re-organization of the styles in folders)

At the moment I don't have strong opinions on this, I think is a clever idea but my "spider sense" tingles a bit, so I want to think more about it, if there would be downsides that I'm not considering. In any case I wanted to record it in the ticket, so it doesn't get lost.

alex-ju (commented):
I like the refactored API, I personally find it easier to understand the heading hierarchy when scanning a page, especially when definitions start with the `@tag` argument as below.
```
<Hds::Text @tag="h1" @variant="display/500/regular">Heading 1</Hds::Text>
[...]
<Hds::Text @tag="h2" @variant="display/400/regular">Heading 2</Hds::Text>
```
I also find it easier to mimic the Figma design by copying/replicating the values rather than setting 3 distinct arguments.

While mapping styles to tags might be beneficial for some of our users, there might also be value in mapping tags to styles. That would lead to simpler declarations, especially for headings, such as: 
```
<Hds::Text @tag="h1">Heading 1</Hds::Text>
```

instead of

```
<Hds::Text @tag="h1" @variant="display/500/regular">Heading 1</Hds::Text>
```

didoo:
> While mapping styles to tags might be beneficial for some of our users, there might also be value in mapping tags to styles (if not bigger value). That would lead to simpler declarations and reduces the overhead of remembering the variants, especially for headings, such as:
> 
> ```
> <Hds::Text @tag="h1">Heading 1</Hds::Text>
> ```
> 
> instead of
> 
> ```
> <Hds::Text @tag="h1" @variant="display/500/regular">Heading 1</Hds::Text>
> ```

@alex-ju I think it's a very **good idea**! Have a look at this change and let me know if this is what you had in mind:
https://github.com/hashicorp/design-system/pull/557/commits/e7fc3c4ff44635d91ec5a78ae6e5b4916be37869
In the meantime I will work on the "How to use" section so we start to see the component APIs in action too.

alex-ju:
> > While mapping styles to tags might be beneficial for some of our users, there might also be value in mapping tags to styles (if not bigger value). That would lead to simpler declarations and reduces the overhead of remembering the variants, especially for headings, such as:
> > ```
> > <Hds::Text @tag="h1">Heading 1</Hds::Text>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > instead of
> > ```
> > <Hds::Text @tag="h1" @variant="display/500/regular">Heading 1</Hds::Text>
> > ```
> 
> @alex-ju I think it's a very **good idea**! Have a look at this change and let me know if this is what you had in mind: [e7fc3c4](https://github.com/hashicorp/design-system/commit/e7fc3c4ff44635d91ec5a78ae6e5b4916be37869) In the meantime I will work on the "How to use" section so we start to see the component APIs in action too.

yes, that looks excellent to me! 🙌

PR #575, opened 2022-09-20
Fix boolean arguments values

### LINKED ISSUE #572: Tweak docs for Card `hasBorder`
Not a bug per se. The Card docs https://design-system-components-hashicorp.vercel.app/components/card say `hasBorder` is a boolean, but it seems like it's intended to be omitted instead of false. 

image

didoo:
### :pushpin: Summary

In our documentation (and in some cases code) we have a few places where we have declared a boolean argument as `@argument="true"` which may be misleading, because if one would do `@argument="false"` it would be evaluated to `true` anyway. Better to avoid this confusion and be explicit in the intent, using the notation `@argument={{true}}`.

This resolves #572.

In this PR I have:
- replaced values for boolean arguments with proper boolean definitions (not strings)

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/badge/index.hbs
+++ b/packages/components/addon/components/hds/badge/index.hbs
@@ -1,7 +1,7 @@
 <div class={{this.classNames}} ...attributes>
   {{#if @icon}}
     <div class="hds-badge__icon">
-      <FlightIcon @name={{this.icon}} @size="16" @stretched="true" />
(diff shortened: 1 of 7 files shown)


alex-ju (approved):
🙌

amyrlam:
thanks!

PR #576, opened 2022-09-20
Add `RadioCard` component

alex-ju:
### :pushpin: Summary

Add `Hds::Form::RadioCard` component based on [current design](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components?node-id=16108%3A37941).

[👉 Preview RadioCard component](https://hds-components-git-alex-ju-add-radio-card-component-hashicorp.vercel.app/components/form/radio-card)

We're introducing a `RadioCard` component comprising of:
 - `Hds::Form::RadioCard` (a bespoke component wrapping a `Form::Radio::Base` instance)
 - `Hds::Form::RadioCard::Group` (very much similar to its cousin `Hds::Form::Radio::Group`)

**Notes**
One interesting challenge with this component is the logic behind selecting a card (and deselecting the other cards), which is currently handled at the application level. After exploring a few options for internalizing this logic as part of the component I ended up leaving it as is, meaning an easier adoption process on the consumer side.

~~I am currently in the process of testing this component in `cloud-ui`, hence the draft status, but feel free to provide feedback in the meantime.~~ Tested this API in HCP Vault and HCP Consul (the main products using this component according to the audit).

I would like to add more examples under the 'How to use' section, showcasing a custom example (using the Generic block) and the `controlPosition` and `align` arguments.

### :link: External links

[Jira issue](https://hashicorp.atlassian.net/browse/HDS-613)

***

### 👀 How to review

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [x] +1 Percy if applicable


packages/components/app/styles/components/form/radio-card.scss
```diff
@@ -0,0 +1,131 @@
+// FORM > RADIO-CARD
+
+// RadioCard Group
+
+.hds-form-group__radio-cards {
+  .hds-form-group__control-fields-wrapper {
+    gap: 16px;
```
didoo, line 7: we can't use `gap` yet because of the browsers versions that we need to support (see thread here: https://hashicorp.slack.com/archives/C11JCBJTW/p1635973292134000)

_(I've asked in Slack again, but I don't expect much difference from one year ago)_

packages/components/app/styles/components/form/radio-card.scss
```diff
@@ -0,0 +1,131 @@
+// FORM > RADIO-CARD
+
+// RadioCard Group
+
+.hds-form-group__radio-cards {
+  .hds-form-group__control-fields-wrapper {
```
didoo, line 6: do you think we could re-use the same approach we have for the "normal" control groups and add a "contextual" class `hds-form-group__control-field` to the radio-card items? or is too much complexity for no real value?

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
```diff
@@ -0,0 +1,625 @@
...
+
+<section>
+  <h3 class="dummy-h3" id="component-api"><a href="#component-api" class="dummy-link-section">§</a> Component API</h3>
+  <h4 class="dummy-h4">Form::RadioCard</h4>
+  <p class="dummy-paragraph" id="component-api-form-radio-card">
+    Here is the API for the
+    <code class="code-dummy">RadioCard</code>
+    component:
+  </p>
+  <dl class="dummy-component-props" aria-labelledby="component-api-form-radio-card">
```
didoo, line 25: I _think_ from this APIs the following arguments/props are missing:
- `@controlPosition`
- `@align`

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
```diff
@@ -0,0 +1,625 @@
...
+    </dd>
+    <dt>&lt;[R].Description&gt; <code>yielded component</code></dt>
+    <dd>
+      <p>It is a container that yields its content inside the card.</p>
+      <p>The content can be a simple string or a more complex/structured one, in which case it inherits the text style.</p>
+    </dd>
+    <dt>&lt;[R].Generic&gt; <code>yielded component</code></dt>
+    <dd>
+      <p>It is a container that yields its content inside the card.</p>
+      <p>The content does not inherit any styles and can be customized as desired.</p>
```
didoo, line 111: Should we mention that, no matter what the order in the declaration in code, the `generic` content appears always at the bottom?

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
```diff
@@ -0,0 +1,625 @@
...
+          (<a
+            href="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attributes"
+            rel="noopener noreferrer"
+          >see whole list here</a>) and some examples of Ember modifiers:
+          <code class="dummy-code">&lcub;&lcub;on "click" [do something]&rcub;&rcub;</code>,
+          <code class="dummy-code">&lcub;&lcub;on "change" [do something]&rcub;&rcub;</code>.
+        </em></p>
+    </dd>
+  </dl>
+  <h5 class="dummy-h5">Contextual components</h5>
```
didoo, line 70: as mentioned above, most of these contextual sub-component accept attributes spreading, so maybe we should make it explicit in the documentation

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
```diff
@@ -0,0 +1,625 @@
...
+    <G.Legend>Group legend</G.Legend>
+    {{#each @model.RADIOCARDS as |item|}}
+      <G.RadioCard @checked={{item.checked}} @value={{item.value}} {{on "change" this.onChange}} as |R|>
+        <R.Icon @name="hexagon" />
+        <R.Label>{{item.label}}</R.Label>
+        <R.Badge @text={{item.badge}} />
+        <R.Description>{{item.description}}</R.Description>
+      </G.RadioCard>
+    {{/each}}
+  </Hds::Form::RadioCard::Group>
```
didoo, line 503: Maybe in this section I would do something that stress-test the component in relation to its content: eg. one card with a longer label that spans multiple lines, one card with a multi-line description and one card without badge or icon. This would also allow a visual-check that all the cards have the same height even when the content is so different.

didoo (commented):
I've left a bunch of comments, but overall is very good and almost ready to go in production.

One thing I want to chat with you about is the _"logic behind selecting a card (and deselecting the other cards)"_, I want to know more about the issues you've found (for my knowledge).

packages/components/app/styles/components/form/radio-card.scss
```diff
@@ -0,0 +1,131 @@
...
+
+    .hds-form-radio-card__control-wrapper {
+      background-color: var(--token-color-surface-faint);
+      border-color: var(--token-color-border-primary);
+    }
+  }
+
+  // ALIGNMENT
+
+  &--align-left {
```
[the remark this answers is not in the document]
alex-ju: Yes, I strongly agree with changing these in favor of findability – not sure why I did this in the first place 😅

packages/components/app/styles/components/form/radio-card.scss
```diff
@@ -0,0 +1,131 @@
...
+
+.hds-form-radio-card__label {
+  margin: 8px 0;
+  color: var(--token-form-label-color);
+  font-weight: bold;
+  font-size: 18px;
+
+  &:first-child {
+    margin-top: 0;
+  }
```
KristinLBradley, line 113: Nit: if a bottom margin were instead added to icons it seems you could remove this conditional style.

packages/components/addon/components/hds/form/radio-card/index.hbs
```diff
@@ -0,0 +1,20 @@
...
+    <Hds::Form::Radio::Base
+      class="hds-form-radio-card__control"
+      @value={{@value}}
+      name={{@name}}
+      checked={{@checked}}
+      disabled={{@disabled}}
+      aria-describedby={{this.ariaDescribedBy}}
+      ...attributes
+    />
+  </div>
```
KristinLBradley, line 19: Nit: Technically the label element should only contain phrasing content (inline-level content) so should not be used to contain div tags.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#technical_summary

KristinLBradley (commented):
See: https://hds-components-git-alex-ju-add-radio-card-component-hashicorp.vercel.app/components/form/radio-card#showcase

Question: Should the disabled state still have a hover effect? To me, when an element changes on hover it indicates that it's somehow interactive so it could possibly be confusing to users.

(_Clarification: I would personally not include any kind of visible hover effect on disabled elements since users may think it's interactive when it isn't._)

@heatherlarsen Opinions?

heatherlarsen:
@KristinLBradley I tend to agree with you, but this is a decision for @cveigt :)

cveigt:
@KristinLBradley I'm not opposed to it. I think it's a good point; will explore the option in Figma.

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
```diff
@@ -0,0 +1,625 @@
...
+    <dd>
+      <p>Appends a
+        <code class="dummy-code">Required</code>
+        indicator next to the legend text and sets the
+        <code class="dummy-code">required</code>
+        attribute on the controls when user input is required.
+      </p>
+      <p>Default: <span class="default">false</span></p>
+    </dd>
+    <dt>isOptional <code>boolean</code></dt>
```
cveigt, line 161: We already discussed it; just a note about removing _isOptional_ from the page.

cveigt (changes requested):
#### Legend
One difference from the radio card group and radio groups in that the spacing between the legend and the card will be 12px and not 8px.
Screen Shot 2022-09-21 at 11 54 49 PM

#### Group width
Question: How can the group's width be managed/controlled? Like fixed vs full width? 👇
Screen Shot 2022-09-22 at 12 07 43 AM

#### Alignment vs Align
Even though the flexbox property is called align, alignment feels clearer to me. I also looked at @jorytindall’s table and the table header and table cell have an alignment property, so I’d like to be consistent with the naming when possible, though I couldn’t find the same property in the code API for the table; I wonder if it’s called the same or not.

#### Disabled attribute
I followed @KristinLBradley's suggestion and removed any elevation from the disabled card, leaving only the border. I opened a new [Figma branch](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/5qLTIwHeJl66yJAq3BAV0u/HDS-Product---Components?node-id=16108%3A37941) for these updates.

#### Motion
I want to explore an animation to make the transitions between states a bit smother.

jorytindall:
@cveigt I'm not sure that the `alignment` property has been implemented in the draft table component yet. I would imagine that they would both be `alignment` in Figma and in the component code though.

didoo:
> @cveigt I'm not sure that the `alignment` property has been implemented in the draft table component yet. I would imagine that they would both be `alignment` in Figma and in the component code though.

from a developer perspective `align` works better for me than `alignment` (reason: the word commonly used in CSS is `align` not `alignment`). in any case I don't see an issue if is `align` in code and `alignment` in Figma, they don't generate confusion, it's clear they're the same

packages/components/app/styles/components/form/radio-card.scss
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// RadioCard
+
+.hds-form-radio-card {
+  display: flex;
+  flex: 1;
+  flex-direction: column;
+  font-family: var(--token-typography-font-stack-text);
```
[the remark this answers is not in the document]
alex-ju: old habits – I was used to avoiding class composition, hence it didn't cross my mind to use the typography classes (with which I need to get more accustomed); removed it from here and used classes on relevant elements.

KristinLBradley:
> 

@cveigt
For the question of managing the width of RadioCards...

We could set a max-width so that the width wouldn't expand past a certain value but would squish down as needed within layouts. For determining the max-width to use, I would look at the most likely use cases within layouts. IE: a single card in a row vs. multiple side by side cards such as 2 or 3 in a row.

packages/components/app/styles/components/form/radio-card.scss
KristinLBradley: Just wanted to comment... From the Slack conversations it looks like John Cowen would be pretty open to updating the browsers they support so we could use flex gap.

https://hashicorp.slack.com/archives/C11JCBJTW/p1636022796147300?thread_ts=1635973292.134000&cid=C11JCBJTW
_"... Consul agreed to use a ~2 years back support, and thats what I'll be PRing today (hopefully). **But that is not set in stone at all.** It sounds like you have a straightforwards workaround, but **if it becomes too much of a workaround please reach out**"_

cveigt:
> from a developer perspective align works better for me than alignment (reason: the word commonly used in CSS is align not alignment). in any case I don't see an issue if is align in code and alignment in Figma, they don't generate confusion, it's clear they're the same

Ok, sounds good. I'll leave the decision of how it'll be called in code up to you (eng). I'll keep it as `alignment` in Figma.

cveigt:
@alex-ju Regarding the transition, I landed on `0.2s`. Here's a [codepen](https://codepen.io/cveigt/pen/BaxwogQ) for reference. Let me know what you think 😬

packages/components/app/styles/components/form/radio-card.scss
alex-ju: yes, that was my initial intent, but `(component "flight-icon" size="24")` doesn't seem to allow for a ` class="hds-form-radio-card__icon"` to be passed in – I can dig a bit to understand why

packages/components/addon/components/hds/form/radio-card/index.hbs
alex-ju: Yep 😬, aware of the spec, and while I'm not a supporter of `label`s wrapping `input`s I think in this very instance following [Melanie's prototype](https://codepen.io/melsumner/pen/oNEGJaw?editors=1100) has many benefits and reduces the complexity of this component (in terms of keeping element states in sync, etc).

packages/components/app/styles/components/form/radio-card.scss
Dhaulagiri: I'm pretty rusty but it's quite possible you can't pass splattributes through component invocations of this type

alex-ju:
Thanks for the thorough review @cveigt! I addressed all your comments in the recent update.

> #### Legend
> One difference from the radio card group and radio groups in that the spacing between the legend and the card will be 

Good spot – have updated through an override for now. I'll check to see if there is scope in passing a modifier class to cope with this difference.

> #### Group width
> Question: How can the group's width be managed/controlled? Like fixed vs full width? 👇

There are a few options here:
 - one is to leave the with constrain to the consumer
 - another is for us to expose a `@maxWidth` attribute (similarly to how we expose `@width` in other components and to what Kristin mentioned above), allowing consumers to constrain the cards without needing to declare a CSS property.
I am leaning more towards the second option, so I will add the attribute and expose an example in the 'Showcase' section.

> #### Alignment vs Align
> Even though the flexbox property is called align, alignment feels clearer to me. I also looked at @jorytindall’s table and the table header and table cell have an alignment property, so I’d like to be consistent with the naming when possible, though I couldn’t find the same property in the code API for the table; I wonder if it’s called the same or not.

As discussed, I don't have strong opinions on this, just acknowledged the different notation in the CSS world. I have updated the component argument to `alignment` to keep it in line with Figma and the table component.

> #### Disabled attribute
> I followed @KristinLBradley's suggestion and removed any elevation from the disabled card, leaving only the border. I opened a new [Figma branch](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/5qLTIwHeJl66yJAq3BAV0u/HDS-Product---Components?node-id=16108%3A37941) for these updates.

Yep, great suggestion here, have updated.

> #### Motion
> I want to explore an animation to make the transitions between states a bit smother.
> Update: Regarding the transition, I landed on 0.2s. Here's a [codepen](https://codepen.io/cveigt/pen/BaxwogQ) for reference. Let me know what you think

Have added the transition – it feels/looks good to me

packages/components/app/styles/components/form/radio-card.scss
alex-ju: yep, you're right Brian, unless we do some argument forwarding (for example adding a `contextualClass` to the FlightIcon then setting `contextualClass="hds-form-radio-card__icon"` on the above call)

didoo: yes, `class` can't be forwarded in contextual components (that's why we had to introduce `contextualClass` in other components)

probably I would not add this complexity and rely on the CSS as implemented now to solve the problem

packages/components/addon/components/hds/form/radio-card/index.hbs
KristinLBradley: Hmm? It's fine for labels to wrap inputs. Just saying preferably we wouldn't use labels to wrap div tags since that's invalid. We could use span tags instead to make the markup validate.

KristinLBradley: Cool, thanks! 

(CSS visual styling doesn't matter in terms of whether HTML markup is valid or not. :) )

packages/components/app/styles/components/form/radio-card.scss
```diff
@@ -0,0 +1,131 @@
...
+    gap: 16px;
+    margin-bottom: auto;
+  }
+}
+
+// RadioCard
+
+.hds-form-radio-card {
+  display: flex;
+  flex: 1;
```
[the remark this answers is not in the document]
KristinLBradley: Personally, I wouldn't set this here as a hard-coded style (`flex: 1;`). I would let the layout be determined by a parent wrapper element of the end-user's choice.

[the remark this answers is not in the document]
alex-ju: @didoo agree that the `<Hds::Form::RadioCard>` shouldn't make assumptions about the parent layout – one option is to only apply this attribute when the `<Hds::Form::RadioCard>` is under a `<Hds::Form::RadioCard::Group>`.

@KristinLBradley the most common scenario is where a set of `<Hds::Form::RadioCard>` components are wrapped with a `<Hds::Form::RadioCard::Group>`. In this case, how would the consumer wrap the `<Hds::Form::RadioCard>`s? Would you be able to exemplify the above suggestion in code? I feel that I'm missing something here.

KristinLBradley:
> 

I would vote for using "align". To me it's clearer (and also shorter which is nice).

packages/components/app/styles/components/form/radio-card.scss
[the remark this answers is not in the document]
KristinLBradley: I think your proposal of applying it when `<Hds::Form::RadioCard>` is within a `<Hds::Form::RadioCard::Group>` makes sense.

packages/components/app/styles/components/form/radio-card.scss
```diff
@@ -0,0 +1,131 @@
...
+  flex-direction: column;
+  font-family: var(--token-typography-font-stack-text);
+  background-color: var(--token-color-surface-primary);
+  border: 1px solid var(--token-color-border-primary);
+  border-radius: 6px;
```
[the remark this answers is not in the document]
alex-ju: Sure, sounds like a good idea and can do in a follow-up PR.

packages/components/app/styles/components/form/radio-card.scss
alex-ju: I forgot about this open discussion. Until we clarify whether or not flex `gap` is something we're confident using across the products I switched to a `margin`-based approach.

packages/components/app/styles/components/form/radio-card.scss
alex-ju: I played a bit with this and the way this class is applied to a field element within a `fieldset` doesn't give us a clean way to pass a `contextualClass` to it as we're using the bare `yield` utility template for these fields. As an alternative, we can potentially do it at the `<Hds::Form::Fieldset>` level, passing it as an argument there, then using it when calling the field (e.g. <Hds::Form::Fieldset contextualFieldClass="hds-form-group__control-fields-wrapper">). I've not yet pushed any commits to address this, as we'll still have a few other overrides within `hds-form-group__radio-cards`, but if the goal is to get rid of these altogether, I can try some refactoring.

packages/components/tests/integration/components/hds/form/radio-card/index-test.js
```diff
@@ -0,0 +1,84 @@
...
+    await render(
+      hbs`<Hds::Form::RadioCard id="my-id" name="my-name" class="my-class" data-test1 data-test2="test" />`
+    );
+    assert.dom('input').hasAttribute('id', 'my-id');
+    assert.dom('input').hasAttribute('name', 'my-name');
+    assert.dom('input').hasClass('my-class');
+    assert.dom('input').hasAttribute('data-test1');
+    assert.dom('input').hasAttribute('data-test2', 'test');
+  });
+});
```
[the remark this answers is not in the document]
didoo: was the thumbs up something todo, or just an ACK?

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
didoo: was the thumbs up something todo, or just an ACK?

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
didoo: was the thumbs up something todo, or just an ACK? on this one I feel more strongly, because it would allow us to test visually (also via visual regression testing) the working of the component

packages/components/tests/integration/components/hds/form/radio-card/index-test.js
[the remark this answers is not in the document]
alex-ju: Thumbs up were meant to be acked and addressed 😅  (nothing scientific though) – I [added tests for thrown assertions here](https://github.com/hashicorp/design-system/blob/alex-ju/add-radio-card-component/packages/components/tests/integration/components/hds/form/radio-card/index-test.js#L31-L57)

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
alex-ju: I tried capturing [the rigid order of arguments (including `Generic` here](https://github.com/hashicorp/design-system/blob/alex-ju/add-radio-card-component/packages/components/tests/dummy/app/templates/components/form/radio-card.hbs#L104-L113)

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
alex-ju: I tried to vary the examples used across sections (including different heights) as follows:
How to use:
 - basic example (using the structured subcomponents Icon, Label, Badge, Description with varied lengths on the Label and Description)
 - custom example (using custom width, generic content, and multiple Badges).

Showcase:
 - contains examples with and without badges; can add one without icon – this may be the only one missing from the suggestions above.

packages/components/tests/integration/components/hds/form/radio-card/index-test.js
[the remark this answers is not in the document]
didoo: @alex-ju super nitpick: assertions are always declared at the bottom of the tests :|

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
didoo: The issue I have with the "How to use" is that it's not covered by Percy. I would feel more "covered" and at peace of mind if the "variable height/content" example is in the showcase too (only one of the "rows", if you want).

Probably something like this :) :
```
  <span class="dummy-text-small">With different content height</span>
  <Hds::Form::RadioCard::Group @name="radio-card-group-custom" as |G|>
    <G.Legend>Group legend</G.Legend>
    <G.RadioCard @checked="true" {{on "change" this.onChange}} as |R|>
      <R.Icon @name="hexagon" />
      <R.Label>Radio card with normal label</R.Label>
      <R.Badge @text="Badge" />
      <R.Description>This is the radio card description text.</R.Description>
    </G.RadioCard>
    <G.RadioCard {{on "change" this.onChange}} as |R|>
      <R.Icon @name="hexagon" />
      <R.Label>Radio card with long label that spans multiple lines</R.Label>
      <R.Badge @text="Badge" />
      <R.Description>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque erat elit, lacinia at magna
        eget, porttitor lobortis nulla.</R.Description>
    </G.RadioCard>
    <G.RadioCard {{on "change" this.onChange}} as |R|>
      <R.Label>Radio card without icon</R.Label>
      <R.Description>This is the radio card description text.</R.Description>
    </G.RadioCard>
  </Hds::Form::RadioCard::Group>
  ```

alex-ju: Valid point about the 'How to use' section – have added the one above, and yes totally agree this example is a good way to prevent layout regression.

packages/components/tests/integration/components/hds/form/radio-card/index-test.js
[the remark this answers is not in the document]
alex-ju: Updated

alex-ju:
> OK, all clear. I _think_ what I would do in this case is to add a "notice" block in the documentation, in the "how to use" section, which explain this peculiarity, so that whoever needs to implement radio cards, they know they have to manage themselves the state logic.

Added a note in the ['How to use' section](http://localhost:4200/components/form/radio-card#how-to-use), let me know if it sounds clear enough.

didoo (approved):
All good for me! 🛳 !

cveigt (approved):
Looks great. Thanks for making that last change! 🚀

alex-ju:
Rebased to `main` – planning to merge this shortly after and follow up to add the 'Design guidelines' and potentially convert some of the CSS values into tokens.

PR #580, opened 2022-09-27
Add feedback to users of `FlightIcon` when icon name is not correct

didoo:
### :pushpin: Summary

As follow-up of [this Slack conversation](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1664215388905739?thread_ts=1664211733.230849&cid=C025N5V4PFZ) I have created a small PR that adds a feedback for the consumers of FlightIcons if a non-existing icon name is provided.

Example of what a developer would see in the console:
screenshot_1817

In this PR I have:
- updated the `build` step in the Flight icons pipeline , to add the generation of a "index" file with the list of all the Flight icon names
- updated the `EmberFlightIcon` component/addon to throw an error if a non-existing icon name is used
  - in doing so, I had to move `ember-auto-import` from `devDependencies` to `dependencies` (otherwise the import of the index file would not work; see https://github.com/ef4/ember-auto-import#usage-from-addons)
- updated the integration tests for `EmberFlightIcon`

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


.changeset/clean-tables-crash.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/flight-icons": minor
```
MelSumner, line 2: Question: do you need two separate change logs?

MelSumner (approved):

.changeset/clean-tables-crash.md
didoo: @MelSumner ysing the `yarn changelog` command, it was not clear (to me) how to apply different semver type to the packages, so I kept it simple and run it twice (and this created two files). I don't think this should make a difference for `changelog` (@Dhaulagiri can you confirm this?)

Dhaulagiri: It should be possible to combine them via `yarn changelog` but in practice I don't think it'll make a difference: changesets will be able to figure it out similar to if these were part of completely separate changes.

PR #583, opened 2022-09-28
Add component-level design tokens for `RadioCard` component

didoo:
### :pushpin: Summary

As a follow-up of #576 this PR intends to add component-level design tokens for the `RadioCard` component

In this PR I have:
- added component-level design tokens for `RadioCard` component
- re-generated design tokens in output
- updated `RadioCard` styles to use component-specific design tokens

***

### 👀 How to review

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/app/styles/components/form/radio-card.scss
+++ b/packages/components/app/styles/components/form/radio-card.scss
@@ -4,7 +4,7 @@
 
 .hds-form-group--radio-cards {
   .hds-form-group__control-fields-wrapper {
-    margin: -8px;
+    margin: calc(-1 * var(--token-form-radiocard-group-gap) / 2);
   }
 
   .hds-form-group__legend {
@@ -13,7 +13,7 @@
 
   .hds-form-radio-card {
     flex: 1;
-    margin: 8px;
+    margin: calc(var(--token-form-radiocard-group-gap) / 2);
   }
 }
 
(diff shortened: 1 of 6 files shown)


packages/components/app/styles/components/form/radio-card.scss
```diff
@@ -4,7 +4,7 @@
 
 .hds-form-group--radio-cards {
   .hds-form-group__control-fields-wrapper {
-    margin: -8px;
+    margin: calc(-1 * var(--token-form-radiocard-group-gap) / 2);
```
alex-ju, line 7: 🌟

alex-ju (approved):
Thank you for taking this on!

So we have a build script that takes the `src`, generates the config for Style Dictionary, and puts the output in `dist`

Dhaulagiri:
This needs a changeset

didoo:
@Dhaulagiri I thought the changeset was provided by #576 (we decided to split this in a new PR just for convenience, instead of doing it in that branch). Makes sense?

Dhaulagiri:
Unless I'm missing something it looks like we also need a new version of the tokens package to be released which is not included in #576

didoo:
@Dhaulagiri ah I see, you mean for the design tokens package! you're right, yes makes sense. I'll open a new PR for that.

PR #586, opened 2022-09-28
make ember-truth-helpers a dependency

Dhaulagiri:
### :pushpin: Summary

More correctly lists `ember-truth-helpers` as a direct dependency of this addon

Randomly came across this while helping investigate something else today. This has worked ok because ember-turth-helpers just happens to be present in all our consuming apps, but it seems good to be more explicity about this just in case.  

For context, way back in https://github.com/hashicorp/design-system-components/pull/25 is when this was added

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


MelSumner (approved):
Oh, interesting, good catch! 👍

PR #588, opened 2022-09-28
HDS-606 `Tab` - Add icon and count arguments

KristinLBradley:
### :pushpin: Summary

This PR is to add arguments to the Tab child component for displaying an optional Flight Icon or BadgeCount. Once approved, it will be merged into the [feature branch](https://github.com/hashicorp/design-system/pull/548).

https://hds-components-git-hds-606-tab-options-hashicorp.vercel.app/

### :link: External links
https://hashicorp.atlassian.net/browse/HDS-606

***

### 👀 How to review

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/addon/components/hds/tabs/tab.hbs
```diff
@@ -1,6 +1,9 @@
 {{! template-lint-disable require-context-role }}
 {{! template-lint-disable no-invalid-role }}
 
+{{! Temporary: fix linter errors with using FlightIcon and BadgeCount inside role="tab" element }}
+{{! template-lint-disable require-presentational-children }}
```
KristinLBradley, line 5: The linter was complaining after I added these. Any suggestions on how to restructure the html DOM? @MelSumner

packages/components/addon/components/hds/tabs/tab.js
```diff
@@ -11,6 +11,28 @@ export default class HdsTabsIndexComponent extends Component {
...
+  /**
+   * @param hasIcon
+   * @type {string}
+   * @default null
+   * @description Determines if the tab displays a FlightIcon (optional)
+   */
+  get hasIcon() {
+    let { hasIcon } = this.args;
+    return hasIcon;
+  }
```
KristinLBradley, line 34: I need to add tests & assertions next. (Will push up to this PR when ready.)

packages/components/app/styles/components/tabs.scss
```diff
@@ -38,26 +38,18 @@ $indicator-height: 3px;
...
-    background: var(--token-color-border-primary);
-    content: "";
-  }
 }
 
 .hds-tabs__tab {
   position: relative; // for positioning indicator
 
+  // prevent overflow gradients from overlaying focused tab
+  &:focus-within { z-index: 2; }
```
didoo, line 51: I _think_ this overlay is still visible (this is in Chrome):
screenshot_1828

didoo (changes requested):
@KristinLBradley I've left a few comments, some of them are blockers. let me know if you want to discuss them in a call (maybe it's easier?).

didoo:
@KristinLBradley one thing I forgot to ask: will integration tests for the `icon` and `count` options be part of this PR, or will they be added in a follow-up?

packages/components/app/styles/components/tabs.scss
```diff
@@ -38,26 +38,18 @@ $indicator-height: 3px;
   overflow-x: auto;
   white-space: nowrap;
   list-style: none;
+  background: linear-gradient(to right, var(--token-color-border-primary), var(--token-color-border-primary)) repeat-x;
+  background-position: bottom floor(calc($indicator-height / 2)) right 0;
+  background-size: 20px 1px;
```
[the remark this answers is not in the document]
didoo: OK, found it: it's in this PR
https://github.com/hashicorp/design-system/pull/581/commits/8d6be094153ff899ad8aacfad97a2c1ca878e059
which is not related to the overflow and was merged without review.

I know it's a big ask, but what do you think of removing this code related to the overflow, and then add it in another PR so it can be properly reviewed?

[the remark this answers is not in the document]
KristinLBradley: ok I'll delete it and see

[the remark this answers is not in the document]
KristinLBradley: I deleted everything related to overflow for now. Will look at it again and push up later as new commit or maybe separate PR

packages/components/app/styles/components/tabs.scss
KristinLBradley: I deleted everything related to overflow for now. Will reexamine later.

packages/components/addon/components/hds/tabs/tab.hbs
```diff
@@ -1,6 +1,9 @@
...
+
 <li class={{this.classNames}} role="presentation" ...attributes>
   <a
```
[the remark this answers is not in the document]
KristinLBradley: I added an explicit height of 36px now which is what I'm seeing in the Figma file. Let me know if this still isn't correct.

https://hds-components-git-hds-606-tab-options-hashicorp.vercel.app/components/tabs#showcase

(@heatherlarsen could use your eyes when you have a chance)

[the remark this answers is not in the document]
KristinLBradley: I pushed up changes to spacing too. Let me know if you want a different implementation. @didoo

KristinLBradley:
> @KristinLBradley one thing I forgot to ask: will integration tests for the `icon` and `count` options be part of this PR, or will they be added in a follow-up?

@didoo I can add them in here. Do you have suggestions on what kind of tests would be useful? I wasn't sure since they each have their own separate integration tests as individual components. I could pass them in and then use the class names to validate they are there as simple tests.

didoo:
> @didoo I can add them in here. Do you have suggestions on what kind of tests would be useful? I wasn't sure since they each have their own separate integration tests as individual components. I could pass them in and then use the class names to validate they are there as simple tests.

@KristinLBradley probably in the context of this PR I would add just the tests for the `icon` and `count` options. That said, my personal preference is to add all the integration tests only at the end, when the API of the component is consolidated and stable (in the end these are UI components, I'm not doing TDD and the testing-while-developing happens in the showcase, not the integration tests; at least for how I work).

packages/components/app/styles/components/tabs.scss
```diff
@@ -77,19 +47,23 @@ $indicator-height: 3px;
...
-  color: var(--token-color-foreground-primary);
-  white-space: nowrap;
-  text-decoration: none;
-  border-radius: var(--token-form-control-border-radius);
+  .hds-tabs__tab-link {
+    @include hds-focus-ring-with-pseudo-element($top: 4px, $right: 6px, $bottom: 4px, $left: 6px);
+    display: flex;
+    align-items: center;
+    height: 36px;
+    padding: 8px 12px;
```
didoo, line 56: in _theory_, since you're setting a fixed vertical height and we don't allow the text to wrap on multiple lines, there's no need for vertical padding; that said, it's not a problem per se (but generally I prefer to omit it, in case in the future I need to change the `height`, I also have to remember to check the `padding`)

packages/components/app/styles/components/tabs.scss
```diff
@@ -77,19 +47,23 @@ $indicator-height: 3px;
...
+    color: var(--token-color-foreground-primary);
+    white-space: nowrap;
+    text-decoration: none;
+    border-radius: var(--token-form-control-border-radius);
+
+    > * + * { margin-left: 6px; }
```
didoo, line 62: generally, we tend not to use these generic selectors, but be more specific/declarative. 
or we use a wrapping element with a "contextual" class (eg. in the button https://github.com/hashicorp/design-system/blob/d24ce3e1a19fe6bf85006148af7005ea5be4b5c6/packages/components/addon/components/hds/button/index.hbs#L15-L17) or apply directly the "contextual" class to the icon (eg. in the https://github.com/hashicorp/design-system/blob/d24ce3e1a19fe6bf85006148af7005ea5be4b5c6/packages/components/addon/components/hds/dropdown/list-item/copy-item.hbs#L11-L14)
then we use the specific "contextual" class to apply styles (typically layout).
so in your case would mean add a class `hds-tabs__tab-icon` to the icon and a class `hds-tabs__tab-badge` to the badge, and then do something like:
```
.hds-tabs__tab-icon { margin-right: 6px; }
.hds-tabs__tab-badge { margin-left: 6px; }
```

packages/components/addon/components/hds/tabs/tab.hbs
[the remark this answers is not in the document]
didoo: there is still something not wright, but I think is the focus ring position; I'll try to attach a screenshot

packages/components/app/styles/components/tabs.scss
KristinLBradley: Alight. I left it since it didn't make a difference but I can change it.

packages/components/app/styles/components/tabs.scss
KristinLBradley: ok

KristinLBradley:
> > @didoo I can add them in here. Do you have suggestions on what kind of tests would be useful? I wasn't sure since they each have their own separate integration tests as individual components. I could pass them in and then use the class names to validate they are there as simple tests.
> 
> @KristinLBradley probably in the context of this PR I would add just the tests for the `icon` and `count` options. That said, my personal preference is to add all the integration tests only at the end, when the API of the component is consolidated and stable (in the end these are UI components, I'm not doing TDD and the testing-while-developing happens in the showcase, not the integration tests; at least for how I work).

Ok I'll do it later at the end then.

packages/components/addon/components/hds/tabs/tab.hbs
KristinLBradley: **Current DOM Structure (simplified):**
```
<li role="presentation">
  <a role="tab" href="#blah">
    <FlightIcon />
    <span>{{yield}}</span>
    <BadgeCount />
  </a>
</li>
```

**I could do something like below instead to satisfy the linter:**
(I could maybe get rid of the inner wrapping span tag.)
```
<li role="presentation">
  <span>
    <FlightIcon />
    <a role="tab" href="#blah">{{yield}}</a>
    <BadgeCount />
  </span>
</li>
```

I could use CSS to expand the target area of the nested anchor tab to cover the full tab area. @MelSumner What do you think or do you have alternate suggestions?

packages/components/addon/components/hds/tabs/tab.hbs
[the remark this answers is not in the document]
heatherlarsen: @KristinLBradley I took a look and left a few minor comments at the Vercel link. Let me know if you need clarification on anything. Great job though! Awesome to see them coming together 🙂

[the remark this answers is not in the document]
didoo: @KristinLBradley OK, I may have found out the difference between Figma and the code.

Here's an overlay between the component in Figma and the component rendered in a browser:
screenshot_1838
as you can see, they overlap almost perfectly (of course the horizontal size of text is different, different rendering engines)

But when I toggle the "focused" state, you can see that the two focus rings are different (the "magenta" one is the Figma one, I've changed the color to see the difference)
screenshot_1839
From what I see the vertical position should be reduced by a couple of pixels (both top and bottom). I _suspect_ this will impact the position of the indicator, but I think we can tune it later when it's implemented properly (with animation too).

I am going to open a small PR to improve the showcase of the Tabs component and subcomponents, so it's easier to debug this things (and plus we can have visual regression testing on them using Percy).

packages/components/addon/components/hds/tabs/tab.hbs
MelSumner: @KristinLBradley I was working on an experiment for the layout, a while back- want to check it out and see what you think? [CodePen: [a11y] tabs: accessible overflow](https://codepen.io/melsumner/pen/PoRgbVv??editors=1000).

We can make time to pair on this today if you want, maybe some 🦆 will help. 👍

KristinLBradley: Thanks. To me it makes sense to use links but I can use buttons instead if that's what you prefer. I tried using a button with the icons inside though and the linter still complained so it doesn't solve that issue. So I'd still have to restructure the DOM to move the icons outside of the button. Unless it's ok to disable the linter for this?

`{{! template-lint-disable require-presentational-children }}`

packages/components/addon/components/hds/tabs/tab.hbs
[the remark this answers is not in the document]
KristinLBradley: I adjusted the focus ring and indicator positions. Please open another comment if you still see issues.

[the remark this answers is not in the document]
didoo: No, all good now. Perfect!

packages/components/addon/components/hds/tabs/tab.hbs
KristinLBradley: I'm going to go ahead and resolve this since I opened it but will continue to work on it in the next iteration. (I'd like to get the other updates I made merged in.)

KristinLBradley: I unresolved again. Sorry for any confusion. So anyway I can go ahead and refactor the DOM so I can delete the lint disable. Would really like to know if you want me to use buttons instead of links though. (As I said my preference is to use links but if others want buttons instead I can change it.)

KristinLBradley:
> @KristinLBradley I've left a few comments, some of them are blockers. let me know if you want to discuss them in a call (maybe it's easier?).

The PR is still showing me "changes requested" and when I click it shows me this comment. Are there any other changes I need to make? Thanks

packages/components/tests/dummy/app/templates/components/tabs.hbs
```diff
@@ -117,6 +129,24 @@
...
+        <T.Tab @icon="download">Two</T.Tab>
+        <T.Tab>Three</T.Tab>
+
+        <T.Panel>Content one <a href="#">Link to test tabbing</a></T.Panel>
+        <T.Panel>Content 2</T.Panel>
+        <T.Panel>Content 3!</T.Panel>
+      </Hds::Tabs>
+    '
+  />
+  {{! prettier-ignore-end }}
```
didoo, line 149: [non-blocker] Not necessarily in this ticket, but can you add after these code examples also a render of the same code as component? see for example how it's done for these components:
- https://design-system-components-hashicorp.vercel.app/components/button#how-to-use
- https://design-system-components-hashicorp.vercel.app/components/form/radio#how-to-use
- https://design-system-components-hashicorp.vercel.app/components/alert#how-to-use
Thanks

didoo (approved):
OK for me to merge.

packages/components/addon/components/hds/tabs/tab.hbs
MelSumner: Can you explain why you'd rather use a link than a button? I'm open to the conversation, I just think a button element is the more appropriate element in this case.

Even if a query param is being appended, we want to still use the button, I think. Using a link will cause the router to view the whole thing as a new URL and there are implications for this in accessible routing in Ember apps.

KristinLBradley: It made sense to me semantically but I'm not trying to push one over the other. I started out using links in the initial basic HTML implementation and didn't receive any feedback to use a button instead. I'll change it to use a button.

KristinLBradley: I refactored the DOM to fix the linter error and to use a button instead of a link.

KristinLBradley: Ok I'll fix it, thanks

PR #591, opened 2022-09-29
Fix spacing in RadioCard

alex-ju:
### :pushpin: Summary

Fix badge-description spacing in RadioCard – now, by error it adds up to `20px`

When a badge exists, the spacing between badge and description should be `12px`
When the badge doesn't exist the spacing between label and description should be `8px`

To match these rules we adjust the CSS properties accordingly and try to rely on `margin-bottom` only, especially when dealing with spacing between optional elements.

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/app/styles/components/form/radio-card.scss
+++ b/packages/components/app/styles/components/form/radio-card.scss
@@ -79,6 +79,11 @@
 
 .hds-form-radio-card--align-center {
   text-align: center;
+
+  // stylelint-disable-next-line selector-class-pattern
+  .flight-icon {
+    margin: auto;
+  }
 }
 
 // CONTROL POSITION
--- a/packages/components/app/styles/components/form/radio-card.scss
+++ b/packages/components/app/styles/components/form/radio-card.scss
@@ -110,7 +110,7 @@
   padding: var(--token-form-radiocard-content-padding);
 
   .hds-badge {
-    margin-bottom: 8px;
+    margin-bottom: 12px;
   }
 }
 
@@ -126,7 +126,6 @@
 
 .hds-form-radio-card__description {
   display: block;
-  margin-top: 12px;
   color: var(--token-form-helper-text-color);
 }
 
--- a/packages/components/app/styles/components/form/radio-card.scss
+++ b/packages/components/app/styles/components/form/radio-card.scss
@@ -131,7 +131,7 @@
 
 .hds-form-radio-card__description {
   display: block;
-  color: var(--token-form-helper-text-color);
+  color: var(--token-color-foreground-primary);
 }
 
 .hds-form-radio-card__control-wrapper {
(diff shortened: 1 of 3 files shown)


didoo (approved):

alex-ju:
Converted this to a draft until Christian reviews the spacing between different elements.

cveigt:
I found a couple of small discrepancies.

* I noticed the text color for the `.hds-form-radio-card__description` class is `var(--token-form-helper-text-color);`. In this case, it should be the foreground primary color.

* I wasn't able to find what's causing this, but there seems to be an extra 4px margin between the icon and the label span.

Screen Shot 2022-09-29 at 11 46 39 PM

Is there a way to remove the top padding from the generic container when the variant is container-only? I noticed that if I leave the container only (see image below), there will be extra padding. or maybe removing the padding and leaving it up to the consumers to add it internally as they see fit? (Thinking out loud 🤔)

Screen Shot 2022-09-29 at 11 39 08 PM

didoo:
> * I noticed the text color for the `.hds-form-radio-card__description` class is `var(--token-form-helper-text-color);`. In this case, it should be the foreground primary color.
> * I wasn't able to find what's causing this, but there seems to be an extra 4px margin between the icon and the label span.
> Screen Shot 2022-09-29 at 11 46 39 PM

Unfortunately the FlightIcon is by default an `inline-block` (/cc @MelSumner) and this means it has an extra space below. @alex-ju I think we need to pass a `isInlineBlock=false` here: https://github.com/hashicorp/design-system/blob/d24ce3e1a19fe6bf85006148af7005ea5be4b5c6/packages/components/addon/components/hds/form/radio-card/index.hbs#L3

> Is there a way to remove the top padding from the generic container when the variant is container-only? I noticed that if I leave the container only (see image below), there will be extra padding. or maybe removing the padding and leaving it up to the consumers to add it internally as they see fit? (Thinking out loud 🤔)
> Screen Shot 2022-09-29 at 11 39 08 PM

@cveigt that is specific to the example @alex-ju created; essentially is working like you expect / described above.

packages/components/app/styles/components/form/radio-card.scss
```diff
@@ -79,6 +79,11 @@
 
 .hds-form-radio-card--align-center {
   text-align: center;
+
+  // stylelint-disable-next-line selector-class-pattern
+  .flight-icon {
```
didoo, line 84: what do you think of this alternative?
```suggestion
  .hds-form-radio-card__icon {
```

alex-ju: yep, that would be the ideal, but the class attribute (and attributes in general) are not spread when the call is made as a contextual component – so `{{yield (hash Icon=(component "flight-icon" size="24" class="hds-form-radio-card__icon" isInlineBlock=false))}}` won't render any class unfortunately.

didoo: Ahhh yes 🤦‍♂️ sorry my bad!

cveigt (approved):
Beautiful 🚀

packages/components/app/styles/components/form/radio-card.scss
alex-ju: Not fond of it though – I'll make a note to maybe consider some sort of contextualClass/additionalClass argument in the flight icon component if we run into this again.

PR #592, opened 2022-09-29
[basic table] Adds table column header component

MelSumner:
### :pushpin: Summary

If merged, this PR will add a table header component to the table component branch.

- removed `table/sort-by` component
- removed sortable examples from docs (for now)
- made row striping optional
- added an example of how to use the table without a value defined for `@model`
- added a `table/th` component
- updated documentation to reflect changes
- updated data model to resemble an API fetch for a more realistic example

simple table code sample

CleanShot 2022-09-30 at 14 24 57

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/tests/dummy/app/templates/components/table.hbs
```diff
@@ -3,17 +3,7 @@
...
-    href="https://unofficial-design-system-search.vercel.app/?path=/docs/cloud-ui-storybook_table--static"
-    rel="noopener noreferrer"
-    target="_blank"
-  >Cloud-UI
-    <code class="dummy-code">&lt;Table&gt;</code>
-    component</a>
-  with a few adjustments for improved WCAG conformance. Users who already have this table component implemented should
-  be able to migrate to this component with little to no difficulty. Additionally, the data-test attributes have been
-  left "as is" to avoid breaking your existing tests.
+  This table component is a simple data table that supports using a data model...or not, to keep the use case flexible.
```
didoo, line 6: [heads-up] also the "Overview" section should be updated to reflect the changes in this PR

packages/components/addon/components/hds/table/index.hbs
```diff
@@ -1,16 +1,11 @@
...
-        to="head"
-      }}
-    </Hds::Table::Row>
+    {{yield to="head"}}
   </thead>
   <tbody class="hds-table__tbody" data-test-table-body>
     {{#if @model}}
-      {{#each @model as |row|}}
-        {{yield row to="body"}}
+      {{#each @model as |modelItemData|}}
```
didoo, line 7: I have noticed that in the Ember ecosystem, a "row" of data in the model is called `record`. Do you think it would make it cleared to call this entity `record` instead of `modelItemData` (which is a bit of a mouthful)?

packages/components/tests/dummy/app/templates/components/table.hbs
```diff
@@ -364,9 +314,11 @@ export default class ComponentsTableRoute extends Route {
...
   <Hds::Table>
     <:head>
-      <th>Cell Header</th>
-      <th>Cell Header</th>
-      <th>Cell Header</th>
+      <tr>
+        <th>Cell Header</th>
+        <th>Cell Header</th>
+        <th>Cell Header</th>
+      </tr>
```
didoo, line 321: Maybe this example should use `<Hds::Table::Row>` and `<Hds::Table::Th>` too?

didoo: I would document it in some "subtle" way, but not suggest it (at lest for now). 

What do you think @alex-ju ?

packages/components/tests/dummy/app/templates/components/table.hbs
```diff
@@ -104,75 +68,46 @@
...
-  <:body as |row|>
-    <tr>
-      <td>\{{row.artist}}</td>
-      <td>\{{row.album}}</td>
-      <td>\{{row.year}}</td>
-    </tr>
-  </:body>
-</Hds::Table>'
-  />
+  <p></p>
```
didoo, line 110: is this empty paragraph intended?

packages/components/tests/dummy/app/templates/components/table.hbs
alex-ju: If I understand correctly, we have two things that we can potentially highlight (either in the 'Showcase' or in the 'How to use' section). One is that they can use the component with or without a `@model` passed to it and the other one is that they can use a subcomponent or simple HTML elements to define the rows and cells. What Melanie did here was to couple the two in a more 'rigid'/structured example and a more 'flexible'/custom one.

If we want to support simple HTML elements (`tr`, `th`) I think it's valuable to 'showcase' them so we can keep an eye on potential visual regression, if we're not committed to supporting them yet then maybe there's not much value. Having them in the showcase doesn't mean we promote them, it's more that we make sure they work as expected.

alex-ju (approved):
It looks good to me! I admit that I liked the `<Hds::Table::Header>` notation as it aligns better with `<Hds::Table::Row>`.

For example
```
<Hds::Table::Row>
  <Hds::Table::Header/><Hds::Table::Header>
</Hds::Table::Row>
```
or
```
<Hds::Table::Tr>
  <Hds::Table::Th/><Hds::Table::Th>
</Hds::Table::Tr>
```

I feel like we're mixing the two notations at this point.

PR #594, opened 2022-09-30
`Tabs` - accessibility refactors

KristinLBradley:
### :pushpin: Summary

This PR is to refactor keyboard controls and remove panel focus. Once approved, it will be merged into this [sub branch](https://github.com/hashicorp/design-system/pull/588).

https://hds-components-git-tabs-accessibility-refactors-hashicorp.vercel.app/components/tabs#showcase

**References**:
- previous PR comments: https://github.com/hashicorp/design-system/pull/566#discussion_r975899588
- thread where this change was discussed: https://hashicorp.slack.com/archives/C025N5V4PFZ/p1664402723593409

* Removed down arrow navigation
* Changed tab keyboard nav behavior so Panel is displayed using "space" or "enter" keys
* Removed focus behavior and styling from Panels

***

### 👀 How to review

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/tabs/index.js
+++ b/packages/components/addon/components/hds/tabs/index.js
@@ -53,24 +53,24 @@ export default class HdsTabsIndexComponent extends Component {
   handleKeyUp(tabIndex, e) {
     const leftArrow = 37;
     const rightArrow = 39;
-    const downArrow = 40;
+    const enterKey =	13;
+    const spaceKey = 32;
 
     if (e.keyCode === rightArrow) {
       const nextTabIndex = (tabIndex + 1) % this.tabIds.length;
-      this.selectTab(nextTabIndex, e);
+      this.focusTab(nextTabIndex, e);
     } else if (e.keyCode === leftArrow) {
       const prevTabIndex =
         (tabIndex + this.tabIds.length - 1) % this.tabIds.length;
-      this.selectTab(prevTabIndex, e);
-    } else if (e.keyCode === downArrow) {
-      this.setSelectedPanelFocus(tabIndex, e);
+      this.focusTab(prevTabIndex, e);
+    } else if (e.keyCode === enterKey || e.keyCode === spaceKey) {
+      this.selectedTabIndex = tabIndex;
     }
   }
 
-  // Select tab for keyboard & mouse navigation:
-  selectTab(tabIndex, e) {
+  // Focus tab for keyboard & mouse navigation:
+  focusTab(tabIndex, e) {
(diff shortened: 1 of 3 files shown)


MelSumner (approved):
👍

didoo (approved):
Code is OK for me. 

Only thing, I have noticed that now when the focus is moved to another tab and the user hits the "space" key, the corresponding panel is activated and the focus remains in the tab. But if the user hits the "enter" key, the corresponding panel is activated **but** the focus disappears (and if you hit arrow left/right nothing happens). I suspect it's because the element triggered is a link (and the URL is changed), while [in the example that Melanie shared](https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html) the elements are buttons. So I suspect that if we want to have the same behaviour we have to convert the elements to buttons, not links.

For context, I tried to add a `e.preventDefault();` to the event handler, but doesn't fix the problem. @MelSumner @alex-ju any other ideas of how we could solve this?

didoo:
Another thing I've noticed: when the URL is changed, the page jumps (see attached video) and personally I don't think this is the behaviour our consumers will desire (/cc @heatherlarsen).

https://user-images.githubusercontent.com/686239/193276368-cf4f689e-a601-4552-a9d6-7667c8422acd.mov

KristinLBradley:
> Another thing I've noticed: when the URL is changed, the page jumps (see attached video) and personally I don't think this is the behaviour our consumers will desire (/cc @heatherlarsen).
> 
>  Screen.Recording.2022-09-30.at.13.52.22.mov

Yes, I overrode the URL change previously which also solved the jumping behavior but I took that out since it seemed we do want the URL to change. Anyway I'm going to merge this.

heatherlarsen:
> Another thing I've noticed: when the URL is changed, the page jumps (see attached video) and personally I don't think this is the behaviour our consumers will desire (/cc @heatherlarsen).
> 

The screen should remain in the same position. It should not jump.

KristinLBradley:
> > Another thing I've noticed: when the URL is changed, the page jumps (see attached video) and personally I don't think this is the behaviour our consumers will desire (/cc @heatherlarsen).
> 
> The screen should remain in the same position. It should not jump.

This is a side effect from removing the code I had overriding the default behavior which I removed based on other feedback. So this is still being refined. My intention is not to have the screen jump. That's just how it is currently after making changes I thought people wanted. This is still a WIP.

PR #590, opened 2022-09-29
`Radiocard` component - Use placeholder for "custom" content

didoo:
### :pushpin: Summary

We have a `DummyPlaceholder` component that we use whenever we want to showcase a "generic" content. We can use it for the generic content in the RadioCard component, so it's also more in line with a similar generic placeholder in Figma:
screenshot_1833

### :hammer_and_wrench: Detailed description

In this PR I have:
- replaced the textual  generic content with the “placeholder” element

### :camera_flash: Screenshots

**Before:**
screenshot_1831

**After:**
screenshot_1832

### 👀 How to review

👉 Review by files changed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #595, opened 2022-09-30
`Tabs` - Showcase improvements

didoo:
### :pushpin: Summary

@KristinLBradley I have opened this PR to improve the showcase of the `Tabs` component, so it was easier for me to debug [possible layout issues](https://github.com/hashicorp/design-system/pull/588/files#r984485993) in the `Tab` element (easier to do when is in isolation); plus as a benefit we can make sure the `Tab` variants are covered by visual regression testing with Percy.

In this PR I have:
- done a small refactoring of the backing classes for `Tab` and `Panel` (see my comments inline)
- improved the showcase in the `Tabs` page

screenshot_1840

### :link: External links

Related JIRA ticket: https://hashicorp.atlassian.net/browse/HDS-606

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/tabs/panel.js
+++ b/packages/components/addon/components/hds/tabs/panel.js
@@ -15,11 +15,15 @@ export default class HdsTabsIndexComponent extends Component {
 
   @cached
   get nodeIndex() {
-    return this.args.panelIds.indexOf(this.panelId);
+    return this.args.panelIds
+      ? this.args.panelIds.indexOf(this.panelId)
+      : undefined;
   }
 
   get tabId() {
-    return this.args.tabIds[this.nodeIndex];
+    return this.nodeIndex !== undefined
+      ? this.args.tabIds[this.nodeIndex]
+      : undefined;
   }
 
   get isSelected() {
@@ -28,9 +32,11 @@ export default class HdsTabsIndexComponent extends Component {
 
   @action
   didInsertNode(element) {
-    this.elementId = element.id;
-    if (typeof this.args.didInsertNode === 'function') {
-      this.args.didInsertNode(this.elementId, ...arguments);
+    let { didInsertNode } = this.args;
+
+    if (typeof didInsertNode === 'function') {
+      this.elementId = element.id;
+      didInsertNode(this.elementId, ...arguments);
     }
   }
 
--- a/packages/components/addon/components/hds/tabs/tab.hbs
+++ b/packages/components/addon/components/hds/tabs/tab.hbs
@@ -8,7 +8,7 @@
   <a
     role="tab"
     class="hds-tabs__tab-link hds-typography-body-200 hds-font-weight-medium"
-    href={{concat "#" this.panelId}}
+    href={{if this.panelId (concat "#" this.panelId)}}
     id={{this.tabId}}
     aria-selected={{this.isSelected}}
     tabindex={{unless this.isSelected "-1"}}
--- a/packages/components/addon/components/hds/tabs/tab.js
+++ b/packages/components/addon/components/hds/tabs/tab.js
@@ -13,11 +13,13 @@ export default class HdsTabsIndexComponent extends Component {
 
   @cached
   get nodeIndex() {
-    return this.args.tabIds.indexOf(this.tabId);
+    return this.args.tabIds ? this.args.tabIds.indexOf(this.tabId) : undefined;
   }
 
   get panelId() {
-    return this.args.panelIds[this.nodeIndex];
+    return this.nodeIndex !== undefined
+      ? this.args.panelIds[this.nodeIndex]
+      : undefined;
   }
 
   /**
@@ -27,7 +29,10 @@ export default class HdsTabsIndexComponent extends Component {
    * @description Determines if the tab is the selected tab
    */
   get isSelected() {
-    return this.nodeIndex === this.args.selectedTabIndex;
+    return (
+      this.nodeIndex !== undefined &&
+      this.nodeIndex === this.args.selectedTabIndex
+    );
   }
 
   get isInitialTab() {
@@ -37,17 +42,33 @@ export default class HdsTabsIndexComponent extends Component {
 
   @action
   didInsertNode() {
-    this.args.didInsertNode(...arguments);
+    let { didInsertNode } = this.args;
+
+    if (typeof didInsertNode === 'function') {
+      didInsertNode(...arguments);
+    }
   }
 
   @action
   onClick() {
-    this.args.onClick(this.nodeIndex, ...arguments);
+    let { onClick } = this.args;
+
+    if (typeof onClick === 'function') {
+      onClick(this.nodeIndex, ...arguments);
+    } else {
+      return false;
+    }
   }
 
   @action
   onKeyUp() {
-    this.args.onKeyUp(this.nodeIndex, ...arguments);
+    let { onKeyUp } = this.args;
+
+    if (typeof onKeyUp === 'function') {
+      onKeyUp(this.nodeIndex, ...arguments);
+    } else {
+      return false;
+    }
   }
 
   /**
--- a/packages/components/app/styles/components/tabs.scss
+++ b/packages/components/app/styles/components/tabs.scss
@@ -20,12 +20,14 @@ $indicator-height: 3px;
   display: flex;
   margin: 0;
   padding: 0;
-  list-style: none;
   border-bottom: 1px solid var(--token-color-border-primary);
 }
 
 .hds-tabs__tab {
   position: relative; // for positioning indicator
+  margin: 0;
+  padding: 0;
+  list-style: none;
 
   &.hds-tabs__tab--is-selected {
     /* indicator TODO: refactor as needed for animation */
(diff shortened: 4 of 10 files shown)


packages/components/addon/components/hds/tabs/panel.js
```diff
@@ -15,11 +15,15 @@ export default class HdsTabsIndexComponent extends Component {
 
   @cached
   get nodeIndex() {
-    return this.args.panelIds.indexOf(this.panelId);
+    return this.args.panelIds
+      ? this.args.panelIds.indexOf(this.panelId)
+      : undefined;
```
didoo, line 20: I have added this logic so the `Hds::Tabs::Panel` can be used as standalone component (without a `Tabs` parent).

packages/components/addon/components/hds/tabs/tab.js
```diff
@@ -27,7 +29,10 @@ export default class HdsTabsIndexComponent extends Component {
    * @description Determines if the tab is the selected tab
    */
   get isSelected() {
-    return this.nodeIndex === this.args.selectedTabIndex;
+    return (
+      this.nodeIndex !== undefined &&
+      this.nodeIndex === this.args.selectedTabIndex
+    );
```
didoo, line 35: needed to add a "guard" here to make sure we have a "selected" state only when is not standalone and so the `nodeIndex` is undefined.

packages/components/addon/components/hds/tabs/tab.js
```diff
@@ -37,17 +42,33 @@ export default class HdsTabsIndexComponent extends Component {
 
   @action
   didInsertNode() {
-    this.args.didInsertNode(...arguments);
+    let { didInsertNode } = this.args;
+
+    if (typeof didInsertNode === 'function') {
+      didInsertNode(...arguments);
+    }
```
didoo, line 49: converted this logic for consistency with other components

packages/components/app/styles/components/tabs.scss
```diff
@@ -20,12 +20,14 @@ $indicator-height: 3px;
...
   padding: 0;
-  list-style: none;
   border-bottom: 1px solid var(--token-color-border-primary);
 }
 
 .hds-tabs__tab {
   position: relative; // for positioning indicator
+  margin: 0;
+  padding: 0;
+  list-style: none;
```
didoo, line 30: because the `Tab` can live as standalone element, I had to move the `list-style` reset here, and reset also the `margin/padding` of the `<li>` element

packages/components/app/styles/components/tabs.scss
```diff
@@ -47,7 +49,12 @@ $indicator-height: 3px;
   }
 
   .hds-tabs__tab-link {
-    @include hds-focus-ring-with-pseudo-element($top: 4px, $right: 6px, $bottom: 4px, $left: 6px);
+    @include hds-focus-ring-with-pseudo-element(
```
didoo, line 52: this is stylelinting doing its thing

packages/components/app/styles/components/tabs.scss
```diff
@@ -57,11 +64,17 @@ $indicator-height: 3px;
...
+    &:hover,
+    &.mock-hover {
       color: var(--token-color-foreground-action-hover);
     }
   }
 }
 
-.hds-tabs__tab-icon { margin-right: 6px; }
-.hds-tabs__tab-badge { margin-left: 6px; }
+.hds-tabs__tab-icon {
```
didoo, line 74: this is stylelinting doing its thing

packages/components/tests/dummy/app/templates/components/tabs.hbs
```diff
@@ -214,36 +214,90 @@
 <section data-test-percy>
   <h3 class="dummy-h3" id="showcase"><a href="#showcase" class="dummy-link-section">§</a> Showcase</h3>
 
-  <p class="dummy-paragraph"><strong>Basic usage:</strong></p>
+  <h4 class="dummy-h4">Tab</h4>
+
+  <h5 class="dummy-h6">Content</h5>
```
didoo, line 219: we tend to use headings for this type of text/subtitle

packages/components/tests/dummy/app/templates/components/tabs.hbs
```diff
@@ -214,36 +214,90 @@
...
+
+  <h5 class="dummy-h6">States</h5>
+  <div class="dummy-tab-base-sample">
+    {{#each @model.TAG_STATES as |state|}}
+      <div class="dummy-tab-base-sample-item">
+        <span class="dummy-text-small">{{capitalize state}}:</span>
+        <Hds::Tabs::Tab mock-state-value={{state}} mock-state-selector="a">Lorem ipsum</Hds::Tabs::Tab>
+        <Hds::Tabs::Tab @icon="hexagon" @count="10" mock-state-value={{state}} mock-state-selector="a">Lorem ipsum</Hds::Tabs::Tab>
+      </div>
+    {{/each}}
```
didoo, line 251: Here instead I've showcased the different kinds of "interactive states" for the component, using the "mock" technique mentioned above. See how it's done in a similar way in (almost) all the other components (look for `Status` or `States` in their showcase).

(I've created two rows, one with just text and one with icon + text + counter, so we can test both, and it's easier to detect possible differences/inconsistencies in the layout)

packages/components/tests/dummy/app/templates/components/tabs.hbs
```diff
@@ -214,36 +214,90 @@
...
+  </div>
+
+  <h4 class="dummy-h4">Panel</h4>
+  <h5 class="dummy-h6">Content</h5>
+  <Hds::Tabs::Panel><DummyPlaceholder
+      @text="Panel with generic content"
+      @height="50"
+      @background="#eee"
+    /></Hds::Tabs::Panel>
```
didoo, line 260: This is mainly for completeness, even though the panel is just a `yield`

packages/components/tests/dummy/app/templates/components/tabs.hbs
```diff
@@ -214,36 +214,90 @@
...
+  <h5 class="dummy-h6">Basic usage</h5>
   <Hds::Tabs as |T|>
     <T.Tab>One</T.Tab>
     <T.Tab>Two</T.Tab>
     <T.Tab>Three</T.Tab>
 
-    <T.Panel>Content one <a href="#">Link to test tabbing</a></T.Panel>
-    <T.Panel>Content 2</T.Panel>
-    <T.Panel>Content 3!</T.Panel>
+    <T.Panel><DummyPlaceholder @height="50" @background="#eee"><span>Content one with
```
didoo, line 269: Here I'm using a "standard" placeholder component, like it's used in other showcases (see for example https://github.com/hashicorp/design-system/pull/590)

packages/components/addon/components/hds/tabs/tab.hbs
```diff
@@ -8,7 +8,7 @@
   <a
     role="tab"
     class="hds-tabs__tab-link hds-typography-body-200 hds-font-weight-medium"
-    href={{concat "#" this.panelId}}
+    href={{if this.panelId (concat "#" this.panelId)}}
```
didoo, line 11: it's a guard, to avoid the standalone `Tab` having an `href="#"` that when clicked would scroll the page to the top

packages/components/app/styles/components/tabs.scss
KristinLBradley: Ok so I guess making the Tab a standalone component is just for the example in the documentation but it does result in invalid HTML. (`li` elements should only be used as children of a `ul`, `ol`, or `menu` parent element.)

KristinLBradley (approved):
This looks fine except for my comment on the invalid HTML structure caused by using the tab as a standalone component in the documentation. I'd prefer not to have invalid HTML structure but don't know if we care about that here.

packages/components/app/styles/components/tabs.scss
didoo: I've pushed a change, to fix this.

didoo:
> This looks fine except for my comment on the invalid HTML structure caused by using the tab as a standalone component in the documentation. I'd prefer not to have invalid HTML structure but don't know if we care about that here.

@KristinLBradley I've pushed a change, to fix the problem of invalid HTML in the documentation. Let me know if it works for you.

packages/components/tests/dummy/app/templates/components/tabs.hbs
```diff
@@ -221,22 +221,30 @@
     <div>
       <span class="dummy-text-small">Text only</span>
       <br />
-      <Hds::Tabs::Tab>Lorem ipsum</Hds::Tabs::Tab>
+      <ul class="dummy-tab-ul-tab-wrapper">
+        <Hds::Tabs::Tab>Lorem ipsum</Hds::Tabs::Tab>
+      </ul>
```
KristinLBradley, line 226: Thanks for adding ul's

packages/components/app/styles/components/tabs.scss
KristinLBradley: Looks good. I will go ahead and merge. Thanks for all your help creating this.

PR #596, opened 2022-09-30
remove unused ember-cli-markdown-it-template

Dhaulagiri:
### :pushpin: Summary

This is leftover from a long time ago when we toyed with using markdown in our docs. Since we'll be moving to a different solution there's no need to leave this behind.


MelSumner (approved):
👍

PR #603, opened 2022-10-03
New docs website - Minimal initial setup

didoo:
### :pushpin: Summary

This is the initial **very minimal** setup to test that things work more or less as expected (I've included some test pages too).

In this PR I have:
- added `@hashicorp/design-system-components` and `@hashicorp/ember-flight-icons` as dependencies
- added `ember-cli-sass` as devDependency so we can use/compile Sass
- included main HDS component Sass files for compilation
- removed `field-guide-default-template` as dependency (so we can build the pages from the ground up)
- copied (and adapted) `field-guide-default-template` files in `website` folder + added `ember-cli-string-helpers`, `ember-math-helpers` and `ember-truth-helpers` as dependencies
- added fake markdown files/pages for testing (later will remove them)

screenshot_1846

### :link: External links

Jira issue: https://hashicorp.atlassian.net/browse/HDS-717

***

### 👀 How to review

👉 Review by files changed


website/package.json
```diff
@@ -72,5 +72,12 @@
   },
   "ember": {
     "edition": "octane"
+  },
+  "dependencies": {
+    "@hashicorp/design-system-components": "workspace:^",
+    "@hashicorp/ember-flight-icons": "workspace:^",
+    "ember-cli-string-helpers": "^6.1.0",
+    "ember-math-helpers": "^2.18.2",
```
Dhaulagiri, line 80: is this actually being used?

website/package.json
```diff
@@ -72,5 +72,12 @@
...
+    "ember-truth-helpers": "^3.1.1"
```
Dhaulagiri, line 81: is this actually being used?

website/package.json
```diff
@@ -72,5 +72,12 @@
   },
   "ember": {
     "edition": "octane"
+  },
+  "dependencies": {
```
Dhaulagiri, line 76: in practice it's all the same difference, but for consistency I think it would make sense to keep these in `devDependencies`

Dhaulagiri (approved):
Left a couple of small comments/questions here:

> copied (and adapted) field-guide-default-template files in website folder

Is the intention to have this exist within this project permanently or would there be value in being able to extract this into some shareable template in the future?

I don't have a strong opinion about a prefix and might argue whether or not one is even necessary, but I do think we should do something other than "dummy". Feels like a word we should avoid if we don't need to use it. I see you also using `doc` in a lot of other places which seems like a reasonable alternative.

website/package.json
didoo: if it's OK I would move/review them in a follow-up PR to avoid conflicts with other PRs I have in progress

website/docs/components/custom-button.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable ember/no-classic-components, ember/no-classic-classes, ember/require-tagless-components, ember/no-actions-hash, prettier/prettier */
```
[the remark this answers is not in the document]
didoo: no, don't worry this is a test file (from `field-guide`) and will be removed soon

didoo:
> > copied (and adapted) field-guide-default-template files in website folder
> Is the intention to have this exist within this project permanently or would there be value in being able to extract this into some shareable template in the future?

@Dhaulagiri I'm not thinking yet in terms of shareable template. It's at the bottom of my priorities, frankly, at the moment.

> I don't have a strong opinion about a prefix and might argue whether or not one is even necessary, but I do think we should do something other than "dummy". Feels like a word we should avoid if we don't need to use it. I see you also using `doc` in a lot of other places which seems like a reasonable alternative.

I've already pushed a commit ([3998dd1](https://github.com/hashicorp/design-system/pull/603/commits/3998dd17dbba92b87bf683f1a928654c50f684b6)) to replace `field-guide` with `doc` and soon I'll do the same with `dummy`.

PR #562, opened 2022-09-06
Add sorting of properties in Stylelint

didoo:
### :pushpin: Summary

This is a follow-up of #543 + #561. 

After discussion with the team, we have agreed to sort the CSS declarations following the so-called "idiomatic" order instead of the "alphabetical" one (see https://github.com/necolas/idiomatic-css and https://codeguide.co/#declaration-order).

_Notice: the actual application of the rules will happen in a follow-up PR to facilitate the code review._

### :hammer_and_wrench: Detailed description

In this PR I have:
- installed `stylelint-config-rational-order`
- added it to the stylelint config file

There were two possible packages I could use:
- https://github.com/ream88/stylelint-config-idiomatic-order
- https://github.com/constverum/stylelint-config-rational-order
both kinda not maintained anymore (eg. https://github.com/constverum/stylelint-config-rational-order/issues) but still working.

I have decided to go with the second one because it's more comprehensive. @Dhaulagiri do you think we should maybe fork it, fix a few major issues just cherry picking them from the PRs already opened on the original repo, and re-publish it under `@hashicorp/hds-stylelint-config-idiomatic-order`?

### :link: External links

https://hashicorp.atlassian.net/browse/HDS-598

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #608, opened 2022-10-05
Remove CSS attribute sorting notes

alex-ju:
### :pushpin: Summary

Remove CSS attribute sorting notes

This is a follow-up on #562 where we moved from alphabetical to idiomatic. Now that we enforce the CSS property order via Stylelint there's probably less value in keeping the note, so we remove it from the blueprint and update the existing files to align with the generator.

### 👀 How to review

👉 Review commit-by-commit


didoo (approved):
Nice!

PR #618, opened 2022-10-07
Update ember to 4.7

alex-ju:
### :pushpin: Summary

Update ember to 4.7

Thanks @Dhaulagiri for showing me the ropes of `ember-cli-update`!


Dhaulagiri:
🥳  nice! This looks good. I think we do need a changeset for the updates to `components` and `ember-flight-icons` though as those do technically impact consumers and require a release.

Dhaulagiri (approved):

didoo:
@Dhaulagiri one consideration:
- we already have a backlog of version updates in Cloud UI that need to be approved (I _think_ at least two PRs)
  - one of them is a potential breaking change (`FlightIcon throwing if the icon name is not correct)
- this I suspect will require another version bump, not sure if it may be potentially breaking for the consumers (or totally safe)
Probably we should coordinate the bumps in Cloud UI to avoid risking to have too many version updates waiting one for another and then be blocked for the release of other stuff. 

just my 2 cents

Dhaulagiri:
> not sure if it may be potentially breaking for the consumers (or totally safe)

There may be exceptions but generally changes like the ones here should not be very high risk for consumers as most of the changes are internal to the packages here in the monorepo or minor updates to dependencies we expose to consumers. If it does pose a problem come adoption time we can always revert these changes. Definitely a good call to be mindful with these types of changes though.

PR #620, opened 2022-10-07
Version Packages

hashibot-hds:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases
### Minor Changes

-   [#548](https://github.com/hashicorp/design-system/pull/548) [`5d1e2cb4`](https://github.com/hashicorp/design-system/commit/5d1e2cb4f953bfbe55ae1fa8a675903712a66a1c) Thanks [@KristinLBradley](https://github.com/KristinLBradley)! - Add new Tabs component

### Patch Changes

-   [#618](https://github.com/hashicorp/design-system/pull/618) [`11d39410`](https://github.com/hashicorp/design-system/commit/11d39410c40725ff34a291e17eac3f8f8321c659) Thanks [@alex-ju](https://github.com/alex-ju)! - Update `ember` to `4.7.0`

-   [#660](https://github.com/hashicorp/design-system/pull/660) [`4d0826a6`](https://github.com/hashicorp/design-system/commit/4d0826a6df220031b36f4918c5c5365012672c27) Thanks [@alex-ju](https://github.com/alex-ju)! - Update the `Hds::Button` style when rendered as a link

-   [#638](https://github.com/hashicorp/design-system/pull/638) [`90182235`](https://github.com/hashicorp/design-system/commit/901822353453e98c914c8d57e523d64b32a23f75) Thanks [@Dhaulagiri](https://github.com/Dhaulagiri)! - explicitly add ember-style-modifiers as a dependency

-   [#636](https://github.com/hashicorp/design-system/pull/636) [`27a283a5`](https://github.com/hashicorp/design-system/commit/27a283a52c2828b32c282401f91df9bd929f9dda) Thanks [@Dhaulagiri](https://github.com/Dhaulagiri)! - Add copyright notice to license file

[...]

PR #622, opened 2022-10-10
move website deps to devDeps

Dhaulagiri:
### :pushpin: Summary

moves website deps to devDeps

I had noted that this was probably good hygiene, but it seems like practically speaking this also impacts how changesets treats package updates. https://github.com/hashicorp/design-system/pull/620 shows how this is causing changeset to actually version the website package because one of its `dependencies` has had a version bump.

After moving these to `devDependencies` and running `yarn changeset version` locally to test I see that we're no longer versioning the website package which seems more like what we want.

Screen Shot 2022-10-10 at 08 37 02

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -33,6 +33,8 @@
     "@ember/test-helpers": "^2.8.1",
     "@glimmer/component": "^1.1.2",
     "@glimmer/tracking": "^1.1.2",
+    "@hashicorp/design-system-components": "workspace:^",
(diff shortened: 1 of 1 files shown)


didoo (approved):
OK for me. In any case we can move them back to the normal dependencies, if/when needed :)

PR #624, opened 2022-10-11
Remove unnecessary inline width/height attributes for image in Dropdown Icon toggle

didoo:
### :pushpin: Summary

Following [this thread](https://hashicorp.slack.com/archives/C03UR6RR56C/p1665433984551619), I have realized that we have left inline `width/height` attributes in the `img` element of the `Dropdown::Toggle::Icon`. Technically the size of the image is set via CSS so these attributes are not necessary (also wrong, in value) and confusing. Better to remove them.

In this PR I have:
- removed the inline `width/height` attributes for the `` element in the Dropdown Icon toggle

***

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/pretty-clocks-explode.md
+++ b/.changeset/pretty-clocks-explode.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Remove unnecessary inline width/height attributes for image in `Dropdown::Toggle::Icon` component
--- a/packages/components/addon/components/hds/dropdown/toggle/icon.hbs
+++ b/packages/components/addon/components/hds/dropdown/toggle/icon.hbs
@@ -8,7 +8,7 @@
 >
   <div class="hds-dropdown-toggle-icon__wrapper">
     {{#if @imageSrc}}
-      <img src={{@imageSrc}} alt="" role="presentation" height="32" width="32" />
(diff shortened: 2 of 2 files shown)


MelSumner (changes requested):
This is known to cause problems. I'd rather that they make their own component if they don't want to use it the way it was designed to be used.

Edit to clarify: 

1. the image size declared in the HTML is how the browser knows what size the image will be and "save" room for it, avoiding layout shift once the image loads. 

2. the underlying issue is that teams are not using the component correctly. We should not adjust the component we carefully and purposefully designed for accessibility reason. This is where the rubber meets to road, I think.

didoo:
> This is known to cause problems. 
Can you expand on this? The only problem I am aware of if the flash of unstyled content, but considered our use base (professionals with fast connections and quick devices) I think is a negligible side effect.

> I'd rather that they make their own component if they don't want to use it the way it was designed to be used.

This PR is not to solve their problem, is only that is better for the code (if it's not needed, better to remove it).

On a side note: the new side navigation is a strategic element for our products, a lot of time and effort have gone into its design and development, so as HDS team I would prefer to support them as much as possible and see if there is a sensible common ground and a pragmatic solution to the issue they've raised, not just tell them to make their own component for this (unless absolutely and undoubtedly necessary).

PR #626, opened 2022-10-11
New docs website - HTML to Markdown conversion (spike)

didoo:
### :pushpin: Summary

This is an attempt to create a set of scripts and codemods that convert the current "scrappy" documentation in HTML to Markdown(-like[^1]) format for the new website. It works, but not yet as I would like (there are some things that _almost_ work) but for now it's good to go (we will revisit the scripts later, with more specific requirements on the generated output).

[^1]: Notice the format actually generated is actually a superset of Markdown, it contains also Handlebars code (like a MDX file, only for Ember and not React).

In this PR I have done a lot of things :) here is the TLDR:
- I have added a set of node `scripts` in the `package.json` file under the `website-html-to-markdown` folder; they are called in sequence via the `yarn generate` script.
  - have a look at the scripts and check that they make sense to you and it's clear what they're doing
- I have added a few codemod scripts that process the "source" files and update some of the code
- I have added a few scripts that also process the source files, splitting them in "sections", and converting them to markdown
  - in one specific case (`how-to-use`) the output is not exactly how I want it, because of how Turndown works (I'll ask @alex-ju to brainstorm together in the next days, to see what we can do); for now I simply copy in output the HTML/Handlebars code

⚠️ **Important reminder**: keep in mind that this code is temporary, just needs to work for one or two weeks more, and then will be completely deleted from the repo.

### :link: External links

JIRA issue: https://hashicorp.atlassian.net/browse/HDS-723

***

### 👀 How to review

👉 Review by files changed (you can also try by commits, if you want, I've tried to keep them as clean as possible squashing all the intermediate steps)


Dhaulagiri:
@didoo it looks like there is some issue getting `yarn install` to work correctly in CI. Let me know if I can help with that.

didoo:
@Dhaulagiri I have re-run `yarn install` in my local env and noticed that the `yarn.lock` file got updated. Pushed it in a new commit, let's see if this fixes the problem.

didoo:
@Dhaulagiri all good now 👍

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
```diff
@@ -603,7 +603,7 @@
         <span class="dummy-text-small">{{capitalize state}}</span>
         <br />
         <div mock-state-value={{state}} mock-state-selector="label">
-          <Hds::Form::RadioCard {{on "change" this.onChange}} @disabled={{(eq state "disabled")}} as |R|>
+          <Hds::Form::RadioCard {{on "change" this.onChange}} @disabled={{eq state "disabled"}} as |R|>
```
Dhaulagiri, line 606: @didoo do you know if this actually works? I know I've always included `()` around ember truth helpers, so I wouldn't want us to document something that doesn't work.

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
```diff
@@ -620,8 +620,8 @@
         <div mock-state-value={{state}} mock-state-selector="label">
           <Hds::Form::RadioCard
             {{on "change" this.onChange}}
-            @checked="true"
-            @disabled={{(eq state "disabled")}}
+            @checked={{true}}
+            @disabled={{eq state "disabled"}}
```
Dhaulagiri, line 624: same question here

Dhaulagiri (approved):
added one question about the changes to the existing docs but otherwise I'm assuming that rest of the temporary changes are what you need them to be

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
didoo: yes, it's OK: https://cs.github.com/?scope=org%3Ahashicorp&scopeName=hashicorp&q=%3D%7B%7Beq

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
didoo: see above

packages/components/tests/dummy/app/templates/components/form/radio-card.hbs
didoo: you can see that it works here too: https://hds-components-git-new-docs-website-spike-html-b2436e-hashicorp.vercel.app/components/form/radio-card#showcase (see the disabled state)

screenshot_1990

Dhaulagiri: nice! thanks for the context

PR #627, opened 2022-10-12
Remove `assert.expect(n)` declarations in the integration tests

didoo:
### :pushpin: Summary

While looking at how @KristinLBradley was writing integration tests, I noticed she didn't use `assert.expect(n)` in her tests, and this made me realize that probably we don't actually need them (they introduce a cost of maintenance when we add/remove one of the assertions in the tests).

Checking [the documentation](https://api.qunitjs.com/assert/expect/) seems it's mainly used with a zero value (_"indicates that a test may pass without making any assertions"_), so I have decided to open a PR to remove these declarations from the integration tests (where they were useless).

In this PR I have:
- removed all the `assert.expect(n)` declarations in the integration tests (with the exception of the assertions tests, where they're actually needed)

### 👀 How to review

👉 Review by files changed

### DIFF
--- a/packages/components/tests/integration/components/hds/alert/index-test.js
+++ b/packages/components/tests/integration/components/hds/alert/index-test.js
@@ -30,7 +30,6 @@ module('Integration | Component | hds/alert/index', function (hooks) {
 
   test('it should render an icon by default depending on the type and color', async function (assert) {
     // here we don't test all the possible combinations, only some of them as precaution
-    assert.expect(6);
     await render(hbs`<Hds::Alert @type="inline" />`);
     assert.dom('.flight-icon-info').exists();
     await render(hbs`<Hds::Alert @type="compact" />`);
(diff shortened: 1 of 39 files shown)


MelSumner (commented):
I don't think I object to this but I would want to double check on why this is here in the first place before approving this change. If others approve and you are confident to move forward without my explicit approval, please do so.

Dhaulagiri:
I have actually never used qunit before so I don't have a strong perspective on this. I assume this is a convention for a reason, but I agree that at least on the surface this seems redundant. 

FWIW a quick search through cloud-ui reveals sporadic adherence to this convention so it doesn't look like something they've found a need to strictly require.

KristinLBradley:
When I was writing tests for Tabs I saw `assert.expect(n)` being used in other component integration tests I looked at but it didn't really make sense to me or seem useful so I left it out. I figured if it was actually important someone would just tell me to add it when they reviewed my PRs...

KristinLBradley (approved):

alex-ju (approved):
I suspect the intent behind this was to make sure that no assertions fail silently by being skipped. Considering our tests don't do much async or involve conditions/loops (which is great) I think we're safe to remove them.

Issue #114, opened 2022-03-23
Enhancement: z-index calculation


MelSumner:
## Summary

Automate (as much as we can) z-index so we don't try to juggle/ remember the numbers. 

## Potential Solution

In sass, make a variable with a list of named z-index layers, and a function that accepts a name and returns what number it is in the list, that way you’re not trying to remember a number; when new things are added to the list, everything will adjust automatically:

```scss
$z: base, header, tooltip, dialog, alert;
.alert { z-index: z(alert); }
```

didoo:
@MelSumner more "correctly" we could introduce
- specific design tokens for `z-index`
- CSS helpers (if it makes sense, I am not sure)
- a sass mixin to apply the index (TBD)

didoo:
I am closing this one for the time being. We will re-open one if/when we will see this a an issue (maybe with the introduction of `Modal` component?).

PR #631, opened 2022-10-13
Add `Modal` component

alex-ju:
### :pushpin: Summary

Add ~~modal `Dialog`~~ `Modal` component

~~Raising this early for feedback on the direction of travel~~
In this modal dialog implementation, we try to leverage the `<dialog>` element as much as possible and kept the codebase to a minimum.

👉 [Preview `Hds::Modal` component](https://hds-components-3l02g127q-hashicorp.vercel.app/components/modal)

### Functionality
- [x] provide a name for the modal dialog (generate `id` and set `aria-labelledby` to the title)
- [x] inform users that it is a modal dialog (`role="dialog"` `aria-modal="true"`) [dialog]
- [x] on `ESC` key the modal dialog is closed [dialog]
- [x] on 'dismiss' button the modal dialog is closed
- [x] on click outside the modal dialog is closed (review, consider `ember-focus-trap` or `ember-click-outside`)
- [x] prevent tabbing to outside the modal dialogue while open (review, consider `ember-focus-trap`)
- [x] on close: allow callback function
- [x] on close: return focus to the element that invoked the modal dialog [dialog]
- [x] on open: set initial focus on the first focusable element [dialog]
- [x] manage overflowing content (scroll only the body container)

### Variations
- [x] `size` variation (small/medium/large)
- [x] ~~type~~ `color` variation (neutral/warning/critical)
- [x] header (title, tagline, icon)
- [x] footer (primary, secondary, tertiary action)

### Documentation

- [x] Overview
- [x] API
- [x] How to use
- [x] Design guidelines
- [x] Accessibility
- [x] Showcase

The current implementation of `Hds::Modal` component in Safari 14 (tested via BrowserStack) which does not have native support for `<dialog>`. In this instances, where there is no native support, [we use Google Chrome's polyfill](https://googlechrome.github.io/dialog-polyfill/).
dialog-polyfill on Safari14

### :link: External links

[Jira issue](https://hashicorp.atlassian.net/browse/HDS-740)

***

### 👀 How to review

👉 Review commit-by-commit (for the 'evolution' of the component and a more digestible review)
👉 Review by files changed (for an overall review)

Reviewer's checklist:

- [x] +1 Percy if applicable


packages/components/addon/components/hds/dialog/index.hbs
```diff
@@ -0,0 +1,12 @@
+{{! template-lint-disable no-invalid-interactive }}
+<dialog
```
didoo, line 2: I _think_ we should break down the component in three parts:

- the `<dialog> element itself (and children)
- the backdrop element
- the union of the two in the full modal component

(class names to be discussed/decided)

More than happy to brainstorm on this.

packages/components/addon/components/hds/dialog/index.js
```diff
@@ -0,0 +1,101 @@
...
+export default class HdsDialogIndexComponent extends Component {
+  @action
+  setupDialog(element) {
+    this.element = element;
+    element.showModal();
+  }
+
+  @action
+  onClick(event) {
+    // This may seem counterintuitive, but the <dialog> element is the target only when the backdrop is clicked
```
didoo, line 20: probably another point for having the backdrop as a standalone element, instead of using a pseudo-element?

packages/components/addon/components/hds/dialog/index.js
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * Sets the type of the dialog
+   * Accepted values: neutral, warning, critical
+   *
+   * @param type
+   * @type {string}
+   * @default 'neutral'
+   */
+  get type() {
```
didoo, line 65: not sold on the `@type` name, for multiple reasons:
- we have tried to avoid it as much as possible, to avoid problems/conflicts with the native `type` HTML attribute
- in the `Alert/Toast` component, this is called `@color` (see [here](https://design-system-components-hashicorp.vercel.app/components/alert#component-api)). I know @jorytindall doesn't like this, and I see his point, but I don't think we should have two different names for the same property on similar components

So I propose that we choose one of these options:

- we call this `@color` as in the `Alert/Toast` component, for consistency
- we rename the argument in the `Alert/Toast` component (refactoring the code) to whatever name we choose for this one ([`appearance`](https://atlassian.design/components/modal-dialog/examples)? [`intent`](https://evergreen.segment.com/components/dialog#default_with_a_danger_intent)? other?) and then refactor the code in the consumers (likely just using a simple codemod) to use the new argument name

I'm totally aware the second option has an intrinsic cost, but this would become technical debt in our codebase, and in my experience tech debt in a design system should be kept as low as possible, to avoid paying much higher "interests" later.

packages/components/addon/components/hds/dialog/index.js
```diff
@@ -0,0 +1,101 @@
...
+    // This may seem counterintuitive, but the <dialog> element is the target only when the backdrop is clicked
+    // otherwise the target would be a child of the <dialog>
+    if (event.target === this.element) {
+      this.closeDialog();
+    }
+  }
+
+  @action
+  closeDialog() {
+    // we call the "onClose" callback if it exists (and is a function)
```
didoo, line 29: Should we consider more granular hooks? eg. `beforeOpen`, `afterOpen`, `beforeClose`, `afterClose`? or maybe just two `onOpen` and `onClose`? thinking out loud here, but eager to hear your opinion on this

(I know this would be the first component for which we would go so granular, but it's also the first component like this)

packages/components/app/styles/components/dialog.scss
```diff
@@ -0,0 +1,123 @@
...
+
+@use "../mixins/focus-ring" as *;
+
+.hds-dialog {
+  flex-direction: column;
+  padding: 0;
+  background: var(--token-color-surface-primary);
+  border: 1px solid rgba(59, 61, 69, 40%); // --token-color-border-strong;
+  border-radius: 10px;
+  box-shadow: var(--token-elevation-higher-box-shadow);
```
didoo, line 13: this needs to be discussed with the designers, because we have an `overlay` style already predefined (in Figma as well in code) but if this is the one we are using for something which is an "overlay" them we need to understand if that needs to remain (in which case would it be used then?) or we can remove it

/cc @heatherlarsen @jorytindall

packages/components/app/styles/components/dialog.scss
```diff
@@ -0,0 +1,123 @@
...
+
+  &[open] {
```
didoo, line 15: I have looked at the code, but I don't understand how this works: it's not `display: none` by default, nothing is setting an `open` attribute on the `<dialog>` element, so... I'm a bit lost 😅

packages/components/app/styles/components/dialog.scss
```diff
@@ -0,0 +1,123 @@
...
+  flex: none;
+  margin-right: 16px;
+}
+
+.hds-dialog__title {
+  flex-grow: 1;
+  font-weight: var(--token-typography-font-weight-semibold);
+  font-size: var(--token-typography-display-200-font-size);
+  font-family: var(--token-typography-display-200-font-family);
+  line-height: var(--token-typography-display-200-line-height);
```
didoo, line 43: better to use typography helper class names in the HTML/HBS element directly

didoo (commented):
@alex-ju I did a first pass on the code and added a few comments. more than happy to discuss them in a call/chat

packages/components/addon/components/hds/dialog/header.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="hds-dialog__header" ...attributes>
+  {{#if @icon}}
+    <FlightIcon class="hds-dialog__icon" @name={{@icon}} @size="24" @isInlineBlock={{false}} />
+  {{/if}}
+  <span class="hds-dialog__title" id={{@id}}>
+    {{yield}}
+  </span>
+  <button class="hds-dialog__dismiss" type="button" aria-label="Dismiss" {{on "click" @closeDialog}}>
+    <FlightIcon @name="x" @size="16" @isInlineBlock={{false}} />
+  </button>
```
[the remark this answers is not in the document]
jorytindall: We did decide to decouple the `Dismissbutton` in Figma to it's own utility component. I'll work with @cveigt to get that handled.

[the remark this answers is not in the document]
didoo: @jorytindall **probably** – and probably you already know it :) – the best and safest way to do it is to move the component already existing in the `Alert` and `Toast` component to its own page, so that possible designs already using the `Alert/Toast` components don't get the instance detached/reset.

packages/components/app/styles/components/dialog.scss
jorytindall: Yes, we have discussed this, there are couple of options that I presented:

1. Use the `overlay` token as it currently exists
2. Use the `overlay` token but update it a bit. It was created before really being used in context so might not really fit our needs.
3. Use the `higher` token. There are issues with this, the usage and semantic isn't really correct.
4. Create a whole new token for the dialog/modal. This seems unreasonable at this time.

I believe I had raised the question in review but there wasn't a clear answer. I'm an advocate for **option 2** and will raise it in review.

packages/components/addon/components/hds/dialog/index.js
jorytindall: @didoo I've considering raising this as an issue in the backlog to discuss as a team, maybe this is the time. I'm an advocate of **option 2** as well. We're getting our lines crossed a little bit with `@color` but not actually labelling the variant with a color (`warning`, `neutral`, etc).

packages/components/addon/components/hds/dialog/header.hbs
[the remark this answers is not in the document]
jorytindall: Following up on this, I've moved the `_dismissButton` out of the Alert and Toast components in [this branch](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/pNGhFl305jijdRTqrxBdpX/HDS-Product---Components?node-id=4452%3A7661). It's now on the **Utilities** page, I'll go ahead and merge the Figma branch for now!

packages/components/addon/components/hds/dialog/header.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="hds-dialog__header" ...attributes>
```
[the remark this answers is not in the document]
alex-ju: I can see how the suggested names may help to locate them in the markup (when you're in CSS, for example), however `hds-dialog__header-title` may also read as 'the title of the header' whereas it is actually the title of the modal (placed in the header); same for the `hds-dialog__header-dismiss`, it is not dismissing the header, but the dialog. So I'm a bit on the fence here.

packages/components/addon/components/hds/dialog/header.hbs
```diff
@@ -0,0 +1,11 @@
...
+  {{#if @icon}}
+    <FlightIcon class="hds-dialog__icon" @name={{@icon}} @size="24" @isInlineBlock={{false}} />
+  {{/if}}
+  <span class="hds-dialog__title" id={{@id}}>
```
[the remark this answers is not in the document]
alex-ju: I've been pondering about this. I guess semantically it makes sense for this to be heading and as it should only be a single H1 on the page, the best candidate would be an H2, but depending on the page hierarchy it could also be an H3.

My worry with using a heading tag (regardless of the level) is that users of this component may place any content in the modal (let's say an H4 followed by a paragraph), increasing the chances of skipping headings which leads to further confusion for assistive tech users. Not setting it as part of the component, means (in my mind) that users could pass an H2 as part of the `<D.Header>` block.

I'm on the fence with this one too. @MelSumner how do you feel about this?

packages/components/addon/components/hds/dialog/header.hbs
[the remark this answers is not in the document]
alex-ju: Based on the thread above, shall I mimic this in the code as well? To have a `DismissButton` 'utility' component shared between `Alert` and `Dialog`?

packages/components/addon/components/hds/dialog/index.js
alex-ju: as discussed during team sync, I've update this to `@color`, for consistency and we can review these instances at a future point, across components.

packages/components/addon/components/hds/dialog/index.js
alex-ju: from what I've tested in `cloud-ui` `onOpen` (as in: `afterOpen`) and `onClose` (as in: `beforeClose`) do cover all the current needs. `beforeOpen` and `afterClose` could suggest that the events happen before the component is inserted and after the component is destroyed, respectively, which would be misleading. I would start with these 2 and only expand if needed.

packages/components/app/styles/components/dialog.scss
alex-ju: Rebased after #646 and updated to `--token-surface-overlay-box-shadow` – @jorytindall let me know if I got it wrong though.

packages/components/app/styles/components/dialog.scss
alex-ju: In short, a `<dialog>` is `display:none` by default (I think what the UAs do is actually an exclusion `dialog:not([open]) { display: none; }`. The native `showModal` method (that we invoke in our code) adds the `open` attribute and the native `close` removes it. That's the complete `<dialog>` API in a gist.

packages/components/app/styles/components/dialog.scss
alex-ju: sure, will swap to class names

packages/components/addon/components/hds/dialog/header.hbs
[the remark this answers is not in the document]
jorytindall: That makes sense to me if they can easily share. It was literally just a copy/paste in Figma, so the visual and states are all exactly the same.

packages/components/app/styles/components/dialog.scss
jorytindall: Nailed it 👍

packages/components/tests/dummy/app/templates/components/dialog.hbs
```diff
@@ -0,0 +1,172 @@
...
+          </form>
+        </D.Footer>
+      </Hds::Dialog>
+    </div>
+  {{/each}}
+
+  {{! This below is just an example of invocation, to get started }}
+  <h4 class="dummy-h4">Invocation</h4>
+  <Hds::ButtonSet>
+    <Hds::Button
```
jorytindall, line 108: This is probably a common problem when disabling page content, but there's a small layout jump when the dialog is invoked due to the scroll bar being removed. Not sure if it's an issue or if it's something that can even be solved, more just a note.

packages/components/addon/components/hds/dialog/header.hbs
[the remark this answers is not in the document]
MelSumner: I don't think we can assign a heading tag **_unless_** we have a way to know where that heading will fall into the hierarchy of the page as a whole and provide the correct heading level. 

If we're going to allow the user to nest a heading element themselves when they invoke a component, the wrapper should be a div and not a span, since a span should only contain [phrasing content](https://html.spec.whatwg.org/multipage/dom.html#phrasing-content-2) and a heading is not in that category.

For this specific issue, the outcomes we're after from an a11y perspective are:
- DOM order matches reading order
- Heading levels are in order (H1-H6) 

Documentation about headings and page structure: https://www.w3.org/WAI/tutorials/page-structure/headings/ 

For general information purposes, here's a video that shows how users with screen readers use headings to navigate content: https://youtu.be/QP1rzxMRap4

packages/components/addon/components/hds/dialog/header.hbs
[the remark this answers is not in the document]
didoo: > Based on the thread above, shall I mimic this in the code as well? To have a `DismissButton` 'utility' component shared between `Alert` and `Dialog`?

Yes, for me. My suggestion is to open a new PR (or I can do it), extract it from the `Alert/Toast` component and then when it's merged "consume" it in the `Dialog`.

packages/components/addon/components/hds/dialog/index.js
didoo: > from what I've tested in `cloud-ui` `onOpen` (as in: `afterOpen`) and `onClose` (as in: `beforeClose`) do cover all the current needs. `beforeOpen` and `afterClose` could suggest that the events happen before the component is inserted and after the component is destroyed, respectively, which would be misleading. I would start with these 2 and only expand if needed.

Works for me 👍

packages/components/addon/components/hds/dialog/header.hbs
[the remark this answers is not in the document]
didoo: OK.

packages/components/tests/dummy/app/templates/components/dialog.hbs
alex-ju: Ah, I guess this happens if you set your scrollbars to be always shown in macOS. Good spot – I'll try and think about a workaround. This is caused by setting overflow to hidden on the main body element, not part of the component, but part of our showcase styling.

packages/components/addon/components/hds/dialog/header.hbs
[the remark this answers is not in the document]
alex-ju: Agree with the usage of a `div` element as we'll allow users to nest a heading element. Have updated accordingly!

packages/components/addon/components/hds/dialog/index.js
alex-ju: I think the suggestion makes sense from a pragmatic point of view – I have updates this to use an element (that we can call overlay, for consistency with what we have in Figma) instead of the pseudo-element. This way we can remove this logic and leverage the code we require for the `focus-trap` anyway.

packages/components/app/styles/components/dialog.scss
alex-ju: Replaced these with class names

packages/components/addon/components/hds/dialog/index.hbs
alex-ju: I know we touched on this offline but leaving a comment for posterity.

Indeed Safari 14 and below doesn't offer support for the `<dialog>` element, so I added a polyfill (developed by a few folks at Google a few years back), for now, so we can make the most of this relatively new and admirable alignment in standard across current versions of the major browsers, with the hope to remove the polyfill in 1 year time or so.

However, as you will see in the code, I decided not to use the standard `::backdrop` pseudoelement, but a classic 'overlay' element. I did this so I can remove the additional logic around modal dismissal when 'clicking outside' – as you suggested in the other comment – and leveraged the `focus-trap` library to handle the 'clicking outside' events.

packages/components/app/styles/components/modal.scss
```diff
@@ -0,0 +1,156 @@
...
+
+.hds-modal__tagline {
+  margin-bottom: 4px;
+}
+
+.hds-modal__dismiss {
+  flex: none;
+  align-self: center;
+  margin-left: 16px;
+  padding: 4px;
```
didoo, line 62: could it be that this padding makes it visually different than the similar dismiss in the alert? plus, is not in the same position (16px from the right side for the alert, 24px for the dismiss)
maybe something @jorytindall can confirm?

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -0,0 +1,669 @@
+{{page-title "Modal Component"}}
+
+<h2 class="dummy-h2">Modal</h2>
+
+{{! ADD YOUR CONTENT BELOW }}
+{{! Please follow as much as possible what it's already done in other components. }}
+{{! You can start by copy&pasting it, if you want, and then adapt it according to your needs. }}
+{{! Once done, please remove these comments (they're created by the generator). }}
```
didoo, line 8:
```suggestion
```

didoo (commented):
@alex-ju I have left a bunch of comments. overall looks very good, it's just a few remaining details to iron out and then it should be ready to go (demoed to the ambassadors).

packages/components/addon/components/hds/modal/index.js
```diff
@@ -0,0 +1,124 @@
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { assert } from '@ember/debug';
+import { getElementId } from '../form/utils/getElementId';
```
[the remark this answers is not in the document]
MelSumner: I've created an issue: https://hashicorp.atlassian.net/browse/HDS-942

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -0,0 +1,669 @@
...
+
+<section>
+  <h3 class="dummy-h3" id="overview"><a href="#overview" class="dummy-link-section">§</a> Overview</h3>
+
+  <p class="dummy-paragraph">Modals are used as an interstitial to request information or feedback from the user,
```
MelSumner, line 13: [Suggestion, non-blocker] I don't think I've seen the word `interstitial` used outside of a medical context. Perhaps a link to an acceptable definition on the web would be useful, especially for users who are unfamiliar with the word or have English has a non-primary language.

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -0,0 +1,669 @@
...
+        <dt>icon <code>string</code></dt>
+        <dd>
+          <p>Acceptable values: any
+            <a href="https://flight-hashicorp.vercel.app/" target="_blank" rel="noopener noreferrer">
+              Flight</a>
+            icon name.</p>
+        </dd>
+        <dt>tagline <code>string</code></dt>
+        <dd>
+          <p>A string that helps the user maintain context when a modal dialog is open.</p>
```
MelSumner, line 94: [Suggestion, non-blocker]:

```suggestion
          <p>A string that helps the user maintain context when a modal dialog is open. (Note: this is NOT the header text, but a small piece of text above the header text.)</p>
```

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -0,0 +1,669 @@
...
+          <p><code class="dummy-code">...attributes</code> spreading is supported on this component.</p>
+        </dd>
+      </dl>
+    </dd>
+    <dt>&lt;[M].Body&gt; <code>yielded component</code></dt>
+    <dd>
+      <p>It is an unstyled, generic container that yields as the main content of the modal dialog.</p>
+      <p>It accepts
+        <code class="dummy-code">...attributes</code>
+        spreading.</p>
```
MelSumner, line 107: [Suggestion, non-blocker]: 

```suggestion
      <p>It supports
        <code class="dummy-code">...attributes</code> and has a scrollbar by default.</p>
```

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -0,0 +1,669 @@
...
+            @color="secondary"
+            {{on "click" (fn this.deactivateModal "formModalActive")}}
+          />
+        </Hds::ButtonSet>
+      </M.Footer>
+    </Hds::Modal>
+  {{/if}}
+  {{! template-lint-enable no-autofocus-attribute }}
+  <p class="dummy-banner dummy-banner--info dummy-paragraph"><FlightIcon @name="info" />
+    Note: If a modal dialog contains interactive elements, such as a form, the initial focus should be set on the first
```
MelSumner, line 244: [question] where is this guidance coming from?

[suggestion] I would probably try to re-word this to be more explicit about the context.

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -0,0 +1,669 @@
...
+</section>
+
+<section>
+  <h3 class="dummy-h3" id="how-to-use"><a href="#how-to-use" class="dummy-link-section">§</a> How to use</h3>
+
+  <h4 class="dummy-h4">Basic use</h4>
+  <p class="dummy-paragraph">The simplest way to invoke a modal is by doing something like this:</p>
+  {{! prettier-ignore-start }}
+  {{! template-lint-disable no-unbalanced-curlies }}
+  <CodeBlock
```
MelSumner, line 129: [a11y issue] Since we're running into "scrollable regions must have keyboard access" issues here, recommend trying to resolve it (e.g., control it via CSS or putting in some line breaks in the code sample so that no scrolling is required (at least at default size))

Screenshot: 
CleanShot 2022-10-31 at 17 16 34

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -0,0 +1,669 @@
...
+    </dd>
+    <dt>&lt;[M].Footer&gt; <code>yielded component</code></dt>
+    <dd>
+      <p>It is a container that yields its content as the footer of the modal dialog.</p>
+      <p>We recommend using it exclusively for actions using the
+        <LinkTo @route="components.button-set"><code class="dummy-code">ButtonSet</code></LinkTo>
+        component. If a tertiary action is presented, it will always be aligned at the end of the row.</p>
+      <p>It accepts
+        <code class="dummy-code">...attributes</code>
+        spreading.</p>
```
MelSumner, line 117: [non-blocker] this is _really_ strange phrasing to me, but I'm going to go ahead and guess that we've already used it elsewhere so it's "standard" now. Definitely recommend that we change this as soon as is expedient, though.

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -0,0 +1,669 @@
...
+    @code='
+      <Hds::Button @text="Open form modal" @color="secondary" \{{on "click" (fn this.activateModal "formModalActive")}} />
+
+      \{{#if this.formModalActive}}
+        <Hds::Modal id="form-modal" @onClose=\{{fn this.deactivateModal "formModalActive"}} as |M|>
+          <M.Header>
+            Why do you want to leave the beta?
+          </M.Header>
+          <M.Body>
+            <Hds::Form::Select::Field autofocus @width="100%" as |F|>
```
MelSumner, line 187: [question] how did we get around the template-linter's no-autofocus rule?

MelSumner (changes requested):
Most of the comments are suggestions but everything is indicated properly. Please let me know if you have any questions or if I can help clarify. Great job!

packages/components/addon/components/hds/dialog/header.hbs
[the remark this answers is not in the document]
alex-ju: I added [a commit](https://github.com/hashicorp/design-system/pull/631/commits/879ba447841d1a9df0f76d1896d8e79cd4eabbfd) to abstract the `DismissButton` as a utility component shared by Alert/Toast and Dialog. Hopefully, it's small enough to be reviewed as part of this work, but if not let me know and I can open a separate PR.

packages/components/addon/components/hds/modal/header.hbs
```diff
@@ -0,0 +1,16 @@
...
+  {{/if}}
+  <div class="hds-modal__title hds-typography-display-300 hds-font-weight-semibold" id={{@id}}>
+    {{#if @tagline}}
+      <div class="hds-modal__tagline hds-typography-body-100 hds-font-weight-regular">
+        {{@tagline}}
+      </div>
+    {{/if}}
+    {{yield}}
+  </div>
+  <button class="hds-modal__dismiss" type="button" aria-label="Dismiss" {{on "click" @close}}>
```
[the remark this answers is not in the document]
alex-ju: yeah, I've been pondering over this for a while. I guess we're in a position where within this repo we have both [dismiss/onDismiss](https://github.com/hashicorp/design-system/blob/main/packages/components/tests/dummy/app/templates/components/alert.hbs#L143-L148) and [close/onClose](https://github.com/hashicorp/design-system/blob/main/packages/components/tests/dummy/app/templates/components/dropdown.hbs#L104-L129) then, on top of this, `focus-trap` uses `onDeactivate` to make this alignment even more difficult. Happy to move to `dismiss`/`onDismiss` as I don't have strong opinions on it, but we'd still have inconsistency. The reason I went with `close` was that this was the native web API naming, but I can see how it is very similar in functionality with the Alert – especially now when they share the same subcomponent.

packages/components/addon/components/hds/modal/index.hbs
```diff
@@ -0,0 +1,15 @@
+{{! template-lint-disable no-invalid-interactive }}
+<dialog
+  class={{this.classNames}}
+  ...attributes
+  aria-labelledby={{this.id}}
+  {{did-insert this.setup}}
+  {{focus-trap isActive=this.isActive focusTrapOptions=(hash onDeactivate=this.close clickOutsideDeactivates=true)}}
+>
+  {{yield (hash Header=(component "hds/modal/header" id=this.id close=this.close))}}
```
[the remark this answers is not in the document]
alex-ju: yep, `close` in this instance is not the native one, but the one we defined in Ember, so it can be renamed

packages/components/addon/components/hds/modal/index.js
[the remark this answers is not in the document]
alex-ju: yep, it makes total sense – I can also try to converge the way we generate ids (a thing we discussed before) while moving this to a more generic place

packages/components/addon/components/hds/modal/index.js
```diff
@@ -0,0 +1,124 @@
...
+    // Register "onClose" callback function to be called when a native 'close' event is dispatched
+    this.element.addEventListener('close', () => {
+      this.isActive = false;
+
+      if (this.args.onClose && typeof this.args.onClose === 'function') {
+        this.args.onClose();
+      }
+    });
+
+    // If the modal dialog is set to be open and it is not already open
```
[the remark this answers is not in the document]
alex-ju: You're right, the comment you suggested is the accurate one. I've been experimenting with an `isOpen` attribute, [then removed it](https://github.com/hashicorp/design-system/pull/631/commits/0ac2d17532f44715f4ee50757bfeca0ba1cbf1a0) as it was redundant and forgot to update the comment.

packages/components/tests/dummy/app/templates/utilities/dismiss-button.hbs
```diff
@@ -0,0 +1,52 @@
...
+  <h3 class="dummy-h3" id="how-to-use"><a href="#how-to-use" class="dummy-link-section">§</a> How to use</h3>
+
+  <p class="dummy-paragraph">Invocation of the component would look something like this:</p>
+  {{! prettier-ignore-start }}
+  {{! template-lint-disable no-unbalanced-curlies }}
+  <CodeBlock @language="markup" @code="<Hds::DismissButton @onClick=\{{this.noop}} />" />
+  {{! template-lint-enable no-unbalanced-curlies }}
+  {{! prettier-ignore-end }}
+  <p class="dummy-paragraph">Renders to:</p>
+  <Hds::DismissButton @onClick={{this.noop}} />
```
didoo, line 37: as discussed, this probably should be something like: `<Hds::DismissButton {{on "click" this.onClickDismissButton}} />`

packages/components/addon/components/hds/dialog/header.hbs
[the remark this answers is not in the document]
didoo: Works for me. Already reviewed and add a comment (what we discussed in the call)

packages/components/app/styles/components/modal.scss
alex-ju: yes, the extra padding made it bigger, but now that I abstracted the `DismissButton` they look the same across components. in terms of 'positioning', the header padding is indeed `24px` at the moment (similarly with the other modal containers: body and footer) – would be good to get confirmation from Jory that this is how was intended

packages/components/package.json
```diff
@@ -71,6 +72,7 @@
     "ember-cli-terser": "^4.0.2",
     "ember-concurrency": "^2.2.0",
     "ember-disable-prototype-extensions": "^1.1.3",
+    "ember-focus-trap": "^1.0.1",
```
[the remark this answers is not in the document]
alex-ju: I changed it right after you raised it 😁 – it should hopefully be in the right section now, but let me know if it's not and I'll figure things out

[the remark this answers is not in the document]
Dhaulagiri: i think you had moved the dialog shim but not this one. AFAICT they both need to be in `depedencies`

packages/components/tests/dummy/app/templates/components/modal.hbs
alex-ju: Yeah, I can see how this may sound like that (many Latin words often have these scientific connotations); the intention, I guess, was to express the fact that modals are usually placed between two states – as interacting with the modal changes the state of what invoked the modal. Luckily for me, Jory provided new wording for this section which does not include this one.

packages/components/tests/dummy/app/templates/components/modal.hbs
alex-ju: I went with line breaks for now. We'll have to address this issue more widely though.

packages/components/tests/dummy/app/templates/components/modal.hbs
alex-ju: Replaced with 'It supports `...attributes`' – if you have better terminology at hand I'm happy to try a cross-repo update.

packages/components/tests/dummy/app/templates/utilities/dismiss-button.hbs
alex-ju: Updated as such

packages/components/tests/dummy/app/templates/components/modal.hbs
alex-ju: From memory, based on [the W3C example used as a reference](https://www.w3.org/WAI/ARIA/apg/example-index/dialog-modal/dialog.html) 😬 
Happy to try a re-word, but if you have some time please adjust as needed and I'll commit it.

packages/components/addon/components/hds/modal/index.js
```diff
@@ -0,0 +1,124 @@
...
+    if (!this.element.open) {
+      this.open();
+    }
+  }
+
+  @action
+  open() {
+    // Make modal dialog visible using the native `showModal` method
+    this.element.showModal();
+    this.isActive = true;
```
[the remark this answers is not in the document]
alex-ju: no good reason – `isOpen` sounds more in-line with the rest of the code so I will update

packages/components/tests/dummy/app/templates/components/modal.hbs
didoo: Not sure what is wrong with `spreading`. It refers to [the `spread` operator in JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax).

packages/components/package.json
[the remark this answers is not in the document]
alex-ju: 🤦  I somehow managed to only move the shim, indeed – updated now, and thank you for spotting this!

packages/components/ember-cli-build.js
```diff
@@ -23,6 +23,8 @@ module.exports = function (defaults) {
   });
 
   app.import('app/styles/prism-dracula.css');
+  app.import('node_modules/dialog-polyfill/dist/dialog-polyfill.css');
+  app.import('node_modules/dialog-polyfill/dist/dialog-polyfill.js');
```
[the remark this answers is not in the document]
Dhaulagiri: we found a way to do this automagically with ember-auto-import 🎉  so it'll only load the css/js in the cases where it's actually needed without consumers needing to do anything

packages/components/addon/components/hds/modal/header.hbs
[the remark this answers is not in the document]
alex-ju: Updated this to use the [onDismiss notation](https://github.com/hashicorp/design-system/pull/631/commits/89f04624e5499830190390f25f2ffd0529de171c)

packages/components/addon/components/hds/modal/index.js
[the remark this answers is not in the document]
alex-ju: I suggest I address this refactor separately as part of the issue created by Melanie – I assigned myself to it.

packages/components/addon/components/hds/modal/index.js
```diff
@@ -0,0 +1,124 @@
...
+import { tracked } from '@glimmer/tracking';
+
+export const DEFAULT_SIZE = 'medium';
+export const DEFAULT_COLOR = 'neutral';
+export const SIZES = ['small', 'medium', 'large'];
+export const COLORS = ['neutral', 'warning', 'critical'];
+
+export default class HdsModalIndexComponent extends Component {
+  @tracked isActive = false;
+  @action
```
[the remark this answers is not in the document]
alex-ju: [Moved them accordingly](https://github.com/hashicorp/design-system/pull/631/commits/97a3cf346c0c66ddf086de4c0aeec18074f9b599)

packages/components/addon/components/hds/modal/index.js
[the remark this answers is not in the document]
alex-ju: Switched to [isOpen](https://github.com/hashicorp/design-system/pull/631/commits/811bab2ef585d48079359f59e7796ab953353feb), it aligns better with other components.

packages/components/tests/dummy/app/controllers/components/modal.js
```diff
@@ -0,0 +1,25 @@
...
+  @action
+  deactivateModal(modal) {
+    this[modal] = false;
+    document.body.style.overflow = 'auto';
+  }
+
+  @action
+  noop() {
+    console.log('noop');
+  }
```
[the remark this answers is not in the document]
alex-ju: This was actually a leftover from manual testing, [removed it](https://github.com/hashicorp/design-system/pull/631/commits/743a45edac90be47cf34f40054b4f3153c656162) now as it's not being used anywhere at the moment

packages/components/tests/integration/components/hds/modal/index-test.js
```diff
@@ -0,0 +1,192 @@
...
+    assert.dom('button.hds-modal__dismiss').exists();
+  });
+  test('it should close the modal when the "dismiss" button is pressed', async function (assert) {
+    await render(
+      hbs`<Hds::Modal id="test-modal" as |M|><M.Header>Title</M.Header></Hds::Modal>`
+    );
+    assert.dom('#test-modal').isVisible();
+    await click('button.hds-modal__dismiss');
+    assert.dom('#test-modal').isNotVisible();
+  });
```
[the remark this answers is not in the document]
alex-ju: I tried to [add tests to these callback functions](https://github.com/hashicorp/design-system/pull/631/commits/d4f8ece4944e9f4572c5ff625324ab67d6b597b7) without requiring a testing library that contains spies (e.g. `sinon.js` in most Ember circles), but they turned out to be flakey so I'm tempted to remove them before merge 😬

packages/components/tests/integration/components/hds/modal/index-test.js
```diff
@@ -0,0 +1,192 @@
...
+  test('it uses the title as name for the dialog', async function (assert) {
+    await render(
+      hbs`<Hds::Modal id="test-modal" as |M|>
+            <M.Header>Title</M.Header>
+          </Hds::Modal>`
+    );
+    assert.dom('.hds-modal__title').exists();
+    assert.dom('.hds-modal__title').hasText('Title');
+    assert.dom('.hds-modal__icon').doesNotExist();
+    assert.dom('.hds-modal__tagline').doesNotExist();
```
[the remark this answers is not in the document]
alex-ju: Indeed, I [removed them](https://github.com/hashicorp/design-system/pull/631/commits/46d7f7a71a16773ecbf659f1d374d7a14a45ab44)

packages/components/tests/integration/components/hds/modal/index-test.js
```diff
@@ -0,0 +1,192 @@
...
+
+  test('it should render the component with CSS classes that reflect the arguments provided', async function (assert) {
+    await render(
+      hbs`<Hds::Modal @size="small" @color="warning" id="test-modal" as |M|><M.Header>Title</M.Header></Hds::Modal>`
+    );
+    assert.dom('#test-modal').hasClass('hds-modal--size-small');
+    assert.dom('#test-modal').hasClass('hds-modal--color-warning');
+  });
+
+  // CONTEXTUAL COMPONENTS
```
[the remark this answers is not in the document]
alex-ju: Right, [added a test](https://github.com/hashicorp/design-system/pull/631/commits/5dfc17af446d46d7538ba924c571dc500255cbf9) for this element too

packages/components/app/styles/components/modal.scss
```diff
@@ -0,0 +1,156 @@
...
+  border: none;
+  border-radius: 10px;
+  box-shadow: var(--token-surface-overlay-box-shadow);
+
+  &[open] {
+    position: fixed;
+    display: flex;
+  }
+
+  &::backdrop {
```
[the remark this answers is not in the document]
alex-ju: Added in [8bb706](https://github.com/hashicorp/design-system/pull/631/commits/8bb70e2e59911f3eb53da73832234c8dd8856b6e#diff-1a817da79a052ea40b90a3ffb1242de27341c17069f7a819fd860df02b87ec76R18-R19)

packages/components/tests/dummy/app/templates/components/modal.hbs
alex-ju: Removed

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -0,0 +1,669 @@
+{{page-title "Modal Component"}}
```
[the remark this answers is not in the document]
alex-ju: Added two new sections under 'How to use' to expand on these. TBD on the second one, based on the feedback we receive.

packages/components/tests/dummy/app/templates/components/modal.hbs
alex-ju: Sorry for the delay on this – we do disable the linter for this specific line.

packages/components/app/styles/components/modal.scss
```diff
@@ -0,0 +1,156 @@
...
+
+.hds-modal__overlay {
+  position: fixed;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
+  background: var(--token-color-palette-neutral-700);
+  opacity: 0.5;
+}
```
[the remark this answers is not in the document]
alex-ju: Good question. Because the scrappy website didn't have any elements with `z-index` set on them, the `auto` did the trick for these examples. However, this doesn't work in any project, so I did a bit of research the other day to see what's out there.

#### `design-system`
 - [`breadcrumbs` is `300`](https://github.com/hashicorp/design-system/blob/main/packages/components/app/styles/components/breadcrumb.scss#L182) (for the truncated content)
 - [`dropdown` is `2`](https://github.com/hashicorp/design-system/blob/main/packages/components/app/styles/components/dropdown.scss#L134) (with [a note](https://github.com/hashicorp/design-system/issues/114) to review these values)
 - [`tabs` is `10`](https://github.com/hashicorp/design-system/blob/main/packages/components/app/styles/components/tabs.scss#L100)

#### `cloud-ui`
 - [the navigation container is `10`](https://github.com/hashicorp/cloud-ui/blob/master/addons/hc-nav/addon/styles/components/hc-app-frame.scss#L37)
 - [the close navigation button is `15`](https://github.com/hashicorp/cloud-ui/blob/f8a3e3a9387a967654742fde2512cae816a1b79d/addons/hc-nav/addon/styles/components/container/index.scss#L101)
 - [the modal container/slot is `100`](https://github.com/hashicorp/cloud-ui/blob/master/addons/hc-nav/addon/styles/components/hc-app-frame.scss#L74)
 - [the modal component is `10`](https://github.com/hashicorp/cloud-ui/blob/b7e088b2034ba35e6ac5643964944d58e09b64a7/packages/cloud-ui-core/addon/styles/components/modal-dialog/index.scss)
 - [the footer is `5`](https://github.com/hashicorp/cloud-ui/blob/master/addons/hc-nav/addon/styles/components/hc-app-frame.scss#L97)
 - [tooltip overlays](https://github.com/hashicorp/cloud-ui/blob/9d461657715b362f42902d252ce0a0e28aad9c17/engines/v-insights/addon/styles/components/tooltip-overlay.scss) and [notifications](https://github.com/hashicorp/cloud-ui/blob/0b239795535d62acdb40886d7ad5351af3241abf/packages/cloud-ui-core/addon/styles/components/notifications/index.scss) are `100`
 - [the application loading is `1000`](https://github.com/hashicorp/cloud-ui/blob/b7e088b2034ba35e6ac5643964944d58e09b64a7/hcp/app/styles/components/application-loading/index.scss)
 - [the app frame](https://github.com/hashicorp/cloud-ui/blob/a88d9eac87c55639ec68f7a9e42f32ee0744b080/packages/cloud-ui-core/addon/styles/components/hcp-app-frame/_global-header-overrides.scss#L7) and [the code editor](https://github.com/hashicorp/cloud-ui/blob/5f8209b7ca93c62ec4da48ecf289f8b025e61b71/packages/cloud-ui-core/addon/styles/components/code-editor/index.scss#L39) are `5`

#### `structure`
- [`dropdown` is `300`](https://github.com/hashicorp/structure/blob/main/packages/pds-ember/app/styles/pds/components/dropdown/__config.scss#L16)
 - [`modal` is `10`](https://github.com/hashicorp/structure/blob/main/packages/pds-ember/app/styles/pds/components/app/modal-area/_config.scss#L5)
 - [`tooltip` is `1`](https://github.com/hashicorp/structure/blob/main/packages/pds-ember/app/styles/pds/components/tooltip/__config.scss#L43)

I can dig into other apps, but I can't really say I see a pattern.

packages/components/tests/integration/components/hds/dismiss-button/index-test.js
```diff
@@ -0,0 +1,23 @@
...
+  test('it should render with a CSS class that matches the component name', async function (assert) {
+    await render(hbs`<Hds::DismissButton id="test-dismiss-button" />`);
+    assert.dom('#test-dismiss-button').hasClass('hds-dismiss-button');
+  });
+  test('it should render with custom classes', async function (assert) {
+    await render(
+      hbs`<Hds::DismissButton class="hds-alert__dismiss" id="test-dismiss-button" />`
+    );
+    assert.dom('#test-dismiss-button').hasClass('hds-alert__dismiss');
+  });
```
didoo, line 22: I don't think you need this test, because almost all the Ember components support this behaviour. You may want to test that the `...attributes` are spread, if you want.

```suggestion
```

packages/components/tests/integration/components/hds/modal/index-test.js
```diff
@@ -0,0 +1,221 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { click, render, resetOnerror, setupOnerror } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | hds/modal/index', function (hooks) {
```
didoo, line 6: general comment on these tests: now that the `<M.Header>` is not required anymore, you may remove it from these tests `render()` too (where it's not needed, of course).

packages/components/tests/integration/components/hds/modal/index-test.js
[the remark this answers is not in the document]
didoo: OK for me to not include them, if they're flakey.

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -0,0 +1,729 @@
...
+    <code class="dummy-code">Body</code>,
+    <code class="dummy-code">Footer</code>
+    keys.</p>
+  <dl class="dummy-component-props" aria-labelledby="contextual-components-modal">
+    <dt>&lt;[M].Header&gt; <code>yielded component</code></dt>
+    <dd>
+      <p>It is a container that yields its content as the title of the modal dialog.</p>
+      <p>It supports
+        <code class="dummy-code">...attributes</code>
+      </p>
```
didoo, line 78: You don't need this, you're already making it explicit at lines 92-95.
```suggestion
```

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -0,0 +1,729 @@
...
+    make it clear to the users that the underlying elements are not interactive and to avoid confusion, we recommend
+    disabling the page scroll
+    <code class="dummy-code">onOpen</code>
+    and enabling it back
+    <code class="dummy-code">onClose</code>
+    (for example, by setting
+    <code class="dummy-code">overflow: hidden;</code>
+    and
+    <code class="dummy-code">overflow: auto;</code>
+    respectively).
```
didoo, line 154: maybe would be better to specify that the `overflow` property should be applied to the `<body>` element of the page? (may be not obvious).

didoo (approved):
Left a few minor comments, but for me is good to go. 🚀

packages/components/addon/components/hds/dismiss-button/index.hbs
```diff
@@ -0,0 +1,3 @@
+<button class="hds-dismiss-button" type="button" aria-label="Dismiss" ...attributes>
```
MelSumner, line 1: [question] Is there a way to allow the aria-label attribute to be internationalized? 
It might also be something that we want to address all at once (as an initiative, for example) across all of our components, though. I guess if we don't do it now, we will definitely need to support it later.

MelSumner (approved):
Looks pretty great. 👍 

I noticed that the page jumps a little when I open/close the modal- any idea what might be causing that? It wasn't immediately obvious to me, and I'm not sure how much that might be because of our scrappy website and not the component itself.

alex-ju:
> Looks pretty great. 👍
> 
> I noticed that the page jumps a little when I open/close the modal- any idea what might be causing that? It wasn't immediately obvious to me, and I'm not sure how much that might be because of our scrappy website and not the component itself.

Yes, [Jory spotted this too](https://github.com/hashicorp/design-system/pull/631#pullrequestreview-1151244810) – it's because of the overflow changes on the `body` element – not part of the component at the moment, but part of the recommendation. I wasn't able to find an untwisted solution for it yet, but still looking for one.

I did some experimentation with making the whole modal dialog focusable so that it gets the focus on open but I'm having a hard time achieving this with the native `<dialog>` element, as in the native implementation it only looks for focusable descendants. Sure, it can be achieved if I prevent the native behavior and manage the focus in our script, but I'm on the fence at this point – more tempted to go with the native one.

packages/components/addon/components/hds/dismiss-button/index.hbs
alex-ju: Yes, I think it's a good idea to do it across the repo. We captured this in a [Jira issue](https://hashicorp.atlassian.net/browse/HDS-567).

jorytindall:
Looks great! I have a couple of small comments:

[non-blocker] In the showcase all of the examples are dismissible and disappear from the showcase when dismissed. This   is technically correct, but referencing the showcase for `alert`, the variants that are dismissible, the dismiss button doesn't _actually_ dismiss I assume for the sake of the showcase.

[question] Setting `overflow-y` to `scroll` results in the scroll bar always being shown even when the content doesn't overflow/scroll. This does shift the layout a bit with an "empty" scroll bar, is there a reason why this is set to an explicit value as opposed to `auto`? I did some hacky browser testing and the layout was respected with `overflow-y: scroll`, curious if there are other thoughts? _attached an image for reference_

Screen Shot 2022-11-08 at 1 27 22 PM

Otherwise, everything looks good to me!

alex-ju:
> Looks great! I have a couple of small comments:
> 
> [non-blocker] In the showcase all of the examples are dismissible and disappear from the showcase when dismissed. This is technically correct, but referencing the showcase for `alert`, the variants that are dismissible, the dismiss button doesn't _actually_ dismiss I assume for the sake of the showcase.

Fair concern, I can add some 'glue code' to prevent them from closing in the showcase section.

> [question] Setting `overflow-y` to `scroll` results in the scroll bar always being shown even when the content doesn't overflow/scroll. This does shift the layout a bit with an "empty" scroll bar, is there a reason why this is set to an explicit value as opposed to `auto`? I did some hacky browser testing and the layout was respected with `overflow-y: scroll`, curious if there are other thoughts? _attached an image for reference_
> 
> Screen Shot 2022-11-08 at 1 27 22 PM

Excellent point here – `auto` works best in this case, will update.

> Otherwise, everything looks good to me!

packages/components/tests/integration/components/hds/modal/index-test.js
alex-ju: I remember what happened here. While the header is not required, `ember-focus-trap` needs an interactive element to be present, so I used the header (which always renders the close button) consistently across the examples to avoid an error from `ember-focus-trap`. We don't recommend using the modal without a dismiss action at the least (be that in the header or in the actions), so I don't think we'll ever see this issue in production but do let me know if you'd like me to handle this differently.

packages/components/tests/dummy/app/templates/components/modal.hbs
alex-ju: Indeed, it's redundant – I removed it

packages/components/tests/dummy/app/templates/components/modal.hbs
alex-ju: [Updated](https://github.com/hashicorp/design-system/pull/631/commits/59d7b136cfb84fbe62d3d21d711fdcfd25b3f424) to make this more explicit.

packages/components/tests/integration/components/hds/modal/index-test.js
didoo: Makes sense. Maybe in this case add a comment somewhere, explaining this special case?

alex-ju: yep, will do now

alex-ju: [Tried to cover this](https://github.com/hashicorp/design-system/pull/631/commits/15945680222593a844b4877bca703ca96d24cb31) by explaining how we use this Ember modifier, what are the requirements, and how we achieve them

alex-ju:
Based on feedback from engineers during office hours I [added an extra commit](https://github.com/hashicorp/design-system/pull/631/commits/c6b3ed348e47eabf8e7c5b46159f1ba887d09d9a) to allow the callback function to cancel the close action (by returning `false`) and updated the form example to illustrate how it can be used in practice to prevent form data from being lost accidentally.

While doing more testing in various codebases I realized we can simplify things for the consumer by [exposing a `close` function in the footer](https://github.com/hashicorp/design-system/pull/631/commits/793b523b3cc0f6494b799403e0334dd253ffb08f) to ease the implementation of a 'Cancel' action (which behaves similarly to the dismiss button, pressing ESC key or clicking outside the modal).

packages/components/tests/integration/components/hds/modal/index-test.js
```diff
@@ -111,6 +111,18 @@ module('Integration | Component | hds/modal/index', function (hooks) {
     await click('button.hds-modal__dismiss');
     assert.dom('#test-modal').isNotVisible();
   });
+  test('it should close the modal when the "close" function is called', async function (assert) {
+    await render(
+      hbs`<Hds::Modal id="test-modal" as |M|>
+            <M.Footer as |F|>
+              <Hds::Button id="cancel-button" type="submit" @text="Cancel" @color="secondary" {{on "click" F.close}} />
```
MelSumner, line 118:
```suggestion
              <Hds::Button id="cancel-button" type="button" @text="Cancel" @color="secondary" {{on "click" F.close}} />
```

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -270,30 +273,27 @@
...
+      <M.Footer as |F|>
         <Hds::ButtonSet>
           <Hds::Button type="submit" @text="Leave Beta" />
-          <Hds::Button
-            type="submit"
-            @text="Cancel"
-            @color="secondary"
-            {{on "click" (fn this.deactivateModal "formModalActive")}}
-          />
+          <Hds::Button type="submit" @text="Cancel" @color="secondary" {{on "click" F.close}} />
```
MelSumner, line 307:
```suggestion
          <Hds::Button type="button" @text="Cancel" @color="secondary" {{on "click" F.close}} />
```
You'd really only want one submit button in a form- you could use `reset` as the value but in general I don't see that used anymore.

alex-ju: ah, yes, good spot! – copy/paste habits, will update to `button`

alex-ju: they were all over the examples 🙀

didoo:
> Based on feedback from engineers during office hours I [added an extra commit](https://github.com/hashicorp/design-system/pull/631/commits/c6b3ed348e47eabf8e7c5b46159f1ba887d09d9a) to allow the callback function to cancel the close action (by returning `false`) and updated the form example to illustrate how it can be used in practice to prevent form data from being lost accidentally.

I have a couple of things I want to discuss with you, I'll set up a meeting for Monday.

> While doing more testing in various codebases I realized we can simplify things for the consumer by [exposing a `close` function in the footer](https://github.com/hashicorp/design-system/pull/631/commits/793b523b3cc0f6494b799403e0334dd253ffb08f) to ease the implementation of a 'Cancel' action (which behaves similarly to the dismiss button, pressing ESC key or clicking outside the modal).

Makes sense. Great addition! 🙌

alex-ju:
> > Based on feedback from engineers during office hours I [added an extra commit](https://github.com/hashicorp/design-system/pull/631/commits/c6b3ed348e47eabf8e7c5b46159f1ba887d09d9a) to allow the callback function to cancel the close action (by returning `false`) and updated the form example to illustrate how it can be used in practice to prevent form data from being lost accidentally.
> 
> I have a couple of things I want to discuss with you, I'll set up a meeting for Monday.
> 

Thanks for catching up on this, @didoo! I pushed an additional commit switching from checking the returned value of the `onClose` callback function to checking the value of `@isDismissDisabled` – conceptually decoupling the `onClose` callback from the ability to dismiss the modal, although they will often be used in conjunction.

While writing the test for this new argument I discovered the `settled` helper which helped me fix the flakey test issue I was running into when testing the `onClose` callback.

packages/components/addon/components/hds/modal/index.js
```diff
@@ -0,0 +1,142 @@
...
+      if (this.args.onClose && typeof this.args.onClose === 'function') {
+        this.args.onClose();
+      }
+
+      // If the dismissal of the modal is disabled, we keep the modal open/visible otherwise we mark it as closed
+      if (this.isDismissDisabled) {
+        // If, in a chain of events, the element is not attached to the DOM, the `showModal` would fail
+        // so we add this safeguard condition that checks for the `<dialog>` to have a parent
+        if (this.element.parentElement) {
+          this.element.showModal();
```
didoo, line 112: In a previous commit, you had this comment that made clear why we're "opening" the modal inside a "close" event. Is this still the case for this behaviour? If yes, probably I would leave the comment (I've found it very clear when I read it the first time)

```suggestion
          // If the callback function returns `false`, we keep the modal open/visible
          // As there is no way to `preventDefault` on `close` events, we call the `showModal` function
          // preserving the state of the modal dialog
          this.element.showModal();
```

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -0,0 +1,760 @@
...
+    <dt>
+      isDismissDisabled
+      <code>boolean</code>
+    </dt>
+    <dd>
+      <p>Default: <span class="default">false</span></p>
+      <p>Set this boolean to
+        <code class="dummy-code">true</code>
+        if you want to prevent the modal from being closed (for instance, to avoid accidental data loss in an
+        unsubmitted form). Make sure you communicate to users the reason why the modal is still open.
```
didoo, line 68: Maybe add something like this? (it will need to be formatted or the linter will complain)

```suggestion
        unsubmitted form). Make sure you communicate to users the reason why the modal is still open, and what they need to do to resolve the problem that is preventing the modal to be closed.
```

didoo (approved):
Added a couple of super minor comments. 

It's good to go! 🚢

packages/components/tests/dummy/app/templates/components/modal.hbs
didoo: (feel free to reword it :) )

.changeset/tough-emus-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+Add `Modal` component
```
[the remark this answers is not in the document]
alex-ju: Good call – mentioned it now as a utility component and by which component it is being used

packages/components/addon/components/hds/modal/index.js
alex-ju: Added back

packages/components/tests/dummy/app/templates/components/modal.hbs
alex-ju: Thank you!

PR #643, opened 2022-10-20
[Table] adds sample table with some sortable columns

MelSumner:
### :pushpin: Summary

If merged, this PR adds an example of a table with some sortable columns to the table showcase. API documentation will be updated in a separate PR.

CleanShot 2022-10-19 at 21 31 16@2x

### 👀 How to review
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/table/index.js
+++ b/packages/components/addon/components/hds/table/index.js
@@ -12,7 +12,6 @@ export default class HdsTableIndexComponent extends Component {
 
   get isSortableTable() {
     const { columns } = this.args;
-    console.log('isSortableTable', columns !== undefined);
     return columns !== undefined;
   }
 
--- a/packages/components/tests/dummy/app/templates/components/table.hbs
+++ b/packages/components/tests/dummy/app/templates/components/table.hbs
@@ -161,11 +161,11 @@ export default class ComponentsTableRoute extends Route {
   @sortOrder=\{{this.sortOrder}}
   @columns=\{{this.model.columns}}
 >
-  <:body as |folkData|>
+  <:body as |data|>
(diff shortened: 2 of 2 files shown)


didoo (approved):
👍

alex-ju (approved):

didoo:
@MelSumner one thing I noticed (can be fixed in a follow-up PR) is that if there is a pre-sorted table (via the `@sortBy/@sortOrder` arguments, than the `a11y message` is not filled at the first run, only after the user clicks a column to sort it.

I don't know if this is intentional (maybe the intent is to announce it only when the user takes an action) but if not I suggest to move this code
```
this.sortedMessageText = `Sorted by ${this.sortBy} ${this.sortOrder}ending`;
```
to its own getter so that is populated also at first render. And possibly, while you're at it, avoid doing the concatenation `${this.sortOrder}ending` but use normal text, so it's easier to do localization later (see https://hashicorp.atlassian.net/browse/HDS-567).

Notice: if you use the getter, then the check in the `table/index.hbs` file it show or not the `a11y message` needs to be updated too, because it checks only if it's sortable, not if it's also _actually_ sorted.

MelSumner:
I'll look into it BUT if the table header already reads out the sorting then the aria-live region doesn't need the announcement. It's just for changes after the page has been loaded/rendered.

didoo:
> I'll look into it BUT if the table header already reads out the sorting then the aria-live region doesn't need the announcement. It's just for changes after the page has been loaded/rendered.

OK, in that case let's leave it as is, and let's check this functionality at the end when everything is consolidated. Thanks for clarifying this for me.

PR #645, opened 2022-10-20
[Table] add example of badge in th cell

MelSumner:
### :pushpin: Summary

If merged, this PR adds an example of a badge in a `th` element. 

CleanShot 2022-10-20 at 11 16 33

@jorytindall We'll need to look at the design since it doesn't pass color contrast: 
CleanShot 2022-10-20 at 11 14 48

### 👀 How to review
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/tests/dummy/app/templates/components/table.hbs
+++ b/packages/components/tests/dummy/app/templates/components/table.hbs
@@ -329,7 +329,7 @@ export default class ComponentsTableRoute extends Route {
   <Hds::Table @model={{this.model.data}}>
     <:head>
       <Hds::Table::Tr>
-        <Hds::Table::Th>Artist</Hds::Table::Th>
+        <Hds::Table::Th>Artist <Hds::Badge @text="New" @color="success" @type="outlined" /></Hds::Table::Th>
         <Hds::Table::Th>Album</Hds::Table::Th>
         <Hds::Table::Th>Release Year</Hds::Table::Th>
       </Hds::Table::Tr>
--- a/packages/components/tests/dummy/app/templates/components/table.hbs
+++ b/packages/components/tests/dummy/app/templates/components/table.hbs
@@ -329,7 +329,8 @@ export default class ComponentsTableRoute extends Route {
   <Hds::Table @model={{this.model.data}}>
     <:head>
       <Hds::Table::Tr>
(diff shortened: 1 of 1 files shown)


packages/components/tests/dummy/app/templates/components/table.hbs
```diff
@@ -329,7 +329,7 @@ export default class ComponentsTableRoute extends Route {
   <Hds::Table @model={{this.model.data}}>
     <:head>
       <Hds::Table::Tr>
-        <Hds::Table::Th>Artist</Hds::Table::Th>
+        <Hds::Table::Th>Artist <Hds::Badge @text="New" @color="success" @type="outlined" /></Hds::Table::Th>
```
didoo, line 332: probably I would use a `@size="small"` Badge here

didoo (approved):
👍

(left only a minor comment)

alex-ju (approved):

jorytindall:
There weren't any examples of this in the audit (not saying that there won't be a scenario in the future), I would recommend against consumers nesting components within the header. Not only does this add visual clutter but competes with the column label hierarchy. In the case that a column is sortable, that problem is compounded.

I'm inclined to either restrict or strongly recommend against nesting components within the `th`. If this is a scenario that needs to be accounted for we'll need to rework the design language to introduce more hierarchy within the text styles, as well as rethink how the Figma component should be structure to account for custom content. In this scenario the outlined/filled badge variants only pass contrast criteria on a `#ffffff` background, and that's only accounting for a single component instance and type.

cc @didoo and @MelSumner

MelSumner:
@jorytindall thanks for the feedback- and that makes total sense. There is more design exploration that would need to be done to support this feature. 

@didoo @alex-ju @KristinLBradley Unless there are objections, I'm going to close this PR based on the design feedback. 👍

didoo:
> @didoo @alex-ju @KristinLBradley Unless there are objections, I'm going to close this PR based on the design feedback. 

@MelSumner I am OK with not showing this use case in the documentation page of the component, to avoid someone thinking is a possible design, **but** I strongly suggest to use the `{{yield}}` instead of the `{{@text}}` (which was the original intent) in the `table/th-sort.hbs` template (similar to what we do in `table/th.hbs`) so that if in the future we ever need to support extra content in the table head, for any reasons, the component API is resilient and doesn't require any refactoring not only on our side, but expecially on the consumers side.

PR #647, opened 2022-10-21
[Table] Incremental improvements

MelSumner:
### :pushpin: Summary

If merged, this PR primarily adds API comments for the table component and makes other small improvements.

- added API info to the table component (via jsdoc style comments)
- removed `guidFor` since it was no longer needed
- refactored the position of the a11y-live message to be in a table caption 
- updated the demo public api files to be more accurate 
- removes example of badge in table header (it's not intended design and it doesn't pass a11y standards)
- ~~changes `getSortCriteria` to `_getSortCriteria` to indicate that it's "private" (or that developers don't need to think about/use this)~~
- updates the `th-sort` component to use `{{yield}}` instead of `{{@text}}`
- updates the `table` component to use a block form with `{{column.label}}` instead of using inline `@text={{column.label}}`
- added a data attribute for testing to the `th-sort` component (since it doesn't support splattributes)

### 👀 How to review
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/addon/components/hds/table/th-sort.hbs
```diff
@@ -1,6 +1,6 @@
-<th scope="col" class="hds-table__th-sort" aria-sort={{this.ariaSort}}>
+<th scope="col" class="hds-table__th-sort" aria-sort={{this.ariaSort}} data-test-table-th-sort>
```
didoo, line 1: question related to this element: in the `Th` template I see that the `scope` is declared before the `...attributes` spreading. Is it OK if the consumers override this value? if not, then should be moved after it.

didoo (commented):
Added a couple of comments. I haven't commented on the tests because there's another PR for this.

packages/components/addon/components/hds/table/index.hbs
```diff
@@ -1,4 +1,7 @@
 <table class={{this.classNames}} data-test-table ...attributes>
+  {{#if this.isSortableTable}}
+    <caption class="sr-only" aria-live="polite">{{this.sortedMessageText}}</caption>
```
alex-ju, line 3: I like how we moved this message to be a caption 🙌

packages/components/addon/components/hds/table/th-sort.hbs
alex-ju: Thought: we currently don't have any other component outputting `data-test-` attributes in the consumer's codebase. I see how it's being used in the test, but I guess we can get away with checking that the `hds-table__th-sort` class exists on a `th` element produced by this component – we do these sort of checks rather often in our codebase.

packages/components/tests/integration/components/hds/table/th-sort-test.js
```diff
@@ -36,4 +31,14 @@ module('Integration | Component | hds/table/th-sort', function (hooks) {
...
     assert.dom('[data-test-table-th-sort]').hasClass('hds-table__th-sort');
   });
+  test('if @sortOrder is not defined, the swap-vertical icon should be displayed', async function (assert) {
+    await render(hbs`<Hds::Table::ThSort
+  @sortBy='artist'
+  @setSortBy={{this.setSortBy}}
+>artist</Hds::Table::ThSort>`);
+
+    assert
+      .dom(this.element.querySelector('.flight-icon.flight-icon-swap-vertical'))
```
alex-ju, line 41: I think we can simplify this to to `assert.dom('.flight-icon-swap-vertical').exists();` – there shouldn't be anything else in the DOM as it gets cleared after each test

alex-ju (commented):
Looks good overall! Not sold on the `data-test-table-th-sort` attribute, I'm thinking we can avoid it – otherwise it should be good to get in.

packages/components/tests/integration/components/hds/table/th-sort-test.js
MelSumner: @alex-ju I re-worked the table tests in the PR after this, I would like to review the tests in that PR if that's okay. WDYT?

packages/components/addon/components/hds/table/index.hbs
MelSumner: Thanks @alex-ju! After you pointed out that the generated ID was not needed, it occurred to me that I needed to better associate the message with the table just in case there were multiple sortable tables on the same page, and using the `caption` seemed to be an acceptable way to do that.

packages/components/addon/components/hds/table/th-sort.hbs
MelSumner: @didoo this component doesn't support `...attributes` - @alex-ju that's why there's a data-test attribute.

packages/components/addon/components/hds/table/index.hbs
```diff
@@ -20,7 +24,7 @@
   </thead>
   <tbody class="hds-table__tbody" data-test-table-body>
     {{#if this.isSortableTable}}
-      {{#each (sort-by this.getSortCriteria @model) as |data|}}
+      {{#each (sort-by this._getSortCriteria @model) as |data|}}
```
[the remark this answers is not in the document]
MelSumner: hmmm. maybe let's discuss, since this is a common way to indicate private API and a signal that developers should not use this and cannot rely on it (in Ember, private API can change without an RFC).

packages/components/addon/components/hds/table/th-sort.hbs
MelSumner: Also, I have focused more on the tests in the PR that follows this one: https://github.com/hashicorp/design-system/pull/650

didoo: > @didoo this component doesn't support 

@MelSumner yes, I was referring to the `Th` (the non sortable one).

packages/components/addon/components/hds/table/index.hbs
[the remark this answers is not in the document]
didoo: There's plenty of similar methods/getters in the other components, that are not directly exposed, and this didn't come up before, so I would avoid to introduce a new pattern, unless we change it across all the other components, which is not feasible for me.
Plus, when you say _signal that developers should not use this and cannot rely on it_ what do you mean? Us or the consumers? If it's us, I don't see the problem; if it's the consumers, I don't understand how they could invoke that function from outside, unless they do some convoluted thing like this:
```
<Hds::Table as |T|>
  <some component that calls T._getSortCriteria />
```

packages/components/tests/integration/components/hds/table/th-sort-test.js
alex-ju: Yes, that makes sense, I've seen the other PR after dropping this comment 😁

packages/components/addon/components/hds/table/index.hbs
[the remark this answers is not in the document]
MelSumner: I will switch it back for now, but I strongly disagree with the premise that just because we hadn't discussed it until now, it should not be allowed. There are a lot of things we haven't discussed yet because we haven't needed to. We should be able to iterate and improve over time.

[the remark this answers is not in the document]
didoo: For clarity: I am not saying it's not allowed. I am saying that **in the context of this PR** doesn't make sense because:
- it introduces a new pattern in the codebase, only for this component, without tackling all the other components
- it's not clear enough (at least for me, from the description above) what are the reasons why this would be beneficial (and for who)

Maybe open a new issue in GitHub, in which you explain why we should change the way in which we declare the getters/setters in the codebase, differentiating between public and private methods (and what that means in the context of a UI component) and what are the benefits of changing the current approach (what are the downsides of the current? what the upside of the proposed one?)

alex-ju (approved):

PR #650, opened 2022-10-24
[Table] Add & Update Table Component Tests

MelSumner:
### :pushpin: Summary

If merged, this PR adds new table component tests and updates existing table component tests for consistency and tidiness.

- note: the `hds/table/th-sort` component does not support `...attributes` because of the way it is automatically integrated into the table component itself, so it uses a `data-test` attribute. This is an exception to our general process of using the `id` attributes for testing. 

(Tip: If you want to see these tests running locally, pull down the branch and run `ember s -p 0` and then visit `http://localhost:7020/tests?nocontainer&filter=table`)

CleanShot 2022-11-02 at 12 55 33

### 👀 How to review
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/tests/integration/components/hds/table/index-test.js
```diff
@@ -10,8 +10,14 @@ module('Integration | Component | hds/table/index', function (hooks) {
     await render(hbs`<Hds::Table />`);
     assert.dom(this.element).exists();
   });
+
```
didoo, line 13: [blocker] I would probably add more tests for the main `Table` component:
- that renders correctly the children for the different use cases
- that the table sorting functionality works as expected
- that renders the live-region caption, with the expected text, and that gets updated when sorting change

packages/components/addon/components/hds/table/index.hbs
```diff
@@ -4,20 +4,22 @@
...
-            @sortOrder={{this.sortOrder}}
-            @setSortBy={{this.setSortBy}}
-          >
-            {{column.label}}
-          </Hds::Table::ThSort>
-        {{else}}
-          <Hds::Table::Th>{{column.label}}</Hds::Table::Th>
-        {{/if}}
-      {{/each}}
+      <Hds::Table::Tr>
```
didoo, line 7: Ops! I missed this "detail" in the previous PRs 🤦‍♂️

packages/components/tests/integration/components/hds/table/th-sort-test.js
```diff
@@ -14,31 +14,44 @@ module('Integration | Component | hds/table/th-sort', function (hooks) {
...
-  @sortBy='artist'
-  @sortOrder='asc'
   @setSortBy={{this.setSortBy}}
 >artist</Hds::Table::ThSort>`);
 
     assert.dom(this.element).hasText('artist');
   });
+  test('it has the scope attribute, and it is set to column', async function (assert) {
+    await render(hbs`<Hds::Table::ThSort
+  @setSortBy={{this.setSortBy}}
```
didoo, line 24: [blocker] if I understand correctly, this is not actually saving you from repetition (and also makes the code more obscure/less explicit)

packages/components/tests/integration/components/hds/table/th-sort-test.js
```diff
@@ -14,31 +14,44 @@ module('Integration | Component | hds/table/th-sort', function (hooks) {
...
+    assert.dom('[data-test-icon="swap-vertical"]').exists();
+  });
+
+  test('if `@sortOrder` is set to asc, the arrow-up icon should be displayed', async function (assert) {
+    await render(hbs`<Hds::Table::ThSort
+  @sortBy='artist'
+  @sortKey='artist'
+  @sortOrder='asc'
+  @setSortBy={{this.setSortBy}}
+>artist</Hds::Table::ThSort>`);
```
didoo, line 53: [nitpick] I am not sure why this strange formatting. have you tried to use prettier on the page?

packages/components/tests/integration/components/hds/table/th-sort-test.js
```diff
@@ -14,31 +14,44 @@ module('Integration | Component | hds/table/th-sort', function (hooks) {
...
   test('it should render the component with a CSS class that matches the component name', async function (assert) {
     await render(hbs`<Hds::Table::ThSort
-  @sortBy='artist'
-  @sortOrder='asc'
   @setSortBy={{this.setSortBy}}
 >artist</Hds::Table::ThSort>`);
 
     assert.dom('[data-test-table-th-sort]').hasClass('hds-table__th-sort');
   });
+
```
didoo, line 37: I would add these tests
- tests the `onClick` functionality of the button, to check that invokes the callback provided with the argument passed to it
- test that the `aria-sort` attribute is correcly set

packages/components/tests/integration/components/hds/table/th-sort-test.js
```diff
@@ -14,31 +14,44 @@ module('Integration | Component | hds/table/th-sort', function (hooks) {
...
 >artist</Hds::Table::ThSort>`);
 
     assert.dom(this.element).hasText('artist');
   });
+  test('it has the scope attribute, and it is set to column', async function (assert) {
+    await render(hbs`<Hds::Table::ThSort
+  @setSortBy={{this.setSortBy}}
+>artist</Hds::Table::ThSort>`);
+
+    assert.dom('[data-test-table-th-sort]').hasAttribute('scope', 'col');
```
didoo, line 27: [nitpick] for the tests of `Td` we target the element using the ID, for this one the data attribute. We should stick to one (probably the ID, for consistency with all the other components' tests)

didoo (changes requested):
Left a few comments (some are blockers, sorry).

alex-ju (commented):
Looks good overall! Not sold on the `data-test-table-th-sort` attribute, I'm thinking we can avoid it – otherwise it should be good to get in.

packages/components/tests/integration/components/hds/table/th-sort-test.js
MelSumner: This test is included because it is a clue that something is missing if the attribute `scope="col"` is ever missing. It can by all means be removed but the test is an indicator that it is put there purposefully and should not be removed.

packages/components/tests/integration/components/hds/table/index-test.js
MelSumner: I added tests for the second two items, can you say more about the first one?

packages/components/tests/integration/components/hds/table/th-sort-test.js
MelSumner: in some instances there will be an id and some will use a data-test attribute- it depends on whether or not the component supports splattributes, I think.

packages/components/tests/integration/components/hds/table/th-sort-test.js
MelSumner: yes, prettier is the one that formatted it this way. 🤷

packages/components/tests/integration/components/hds/table/th-sort-test.js
MelSumner: Also, can you clarify what you consider a blocker here?

packages/components/tests/integration/components/hds/table/th-sort-test.js
MelSumner: added

packages/components/tests/integration/components/hds/table/th-sort-test.js
didoo: I think the code has been refactored since my comment (or github assigned the comment to the wrong block of code). My comment was about this code: 
```
  // since we'd have to repeat this for each test, we'll set it up once to be used in all tests
  hooks.beforeEach(function () {
    this.set('setSortBy', function (sortBy) {
      return sortBy || 'artist';
    });
  });
````

File: packages/components/tests/integration/components/hds/table/th-sort-test.js (hunk 6 above)
Comment (didoo) [reply packages/components/tests/integration/components/hds/table/th-sort-test.js:27]: I mean in the tests themselves. why use different approaches? what is the benefit in that case?

Comment (didoo) [reply packages/components/tests/integration/components/hds/table/th-sort-test.js:27]: OK, I see; the `Th-Sort` doesn't have splattributes. But at this point, why not add the `...attributes` also to the `Th-Sort` (even if it's not used directly when it's in a loop)? Would allow us to avoid exceptions, make it more similar to the `Th` and (bonus) make it potentially ready to be used for custom-built sortable table if needed.

Comment (MelSumner) [reply packages/components/tests/integration/components/hds/table/th-sort-test.js:27]: I don't think there's a way for them to invoke `...attributes` here and `data-test-*` attributes are stripped out for production. It's a legitimate use case and it's reasonable to leave it be.

File: packages/components/tests/integration/components/hds/table/index-test.js (hunk 1 above)
Comment (MelSumner) [reply packages/components/tests/integration/components/hds/table/index-test.js:13]: I'm going to add more tests after I finish a couple of tiny features. So I think I will consider this resolved for now.

File: packages/components/tests/integration/components/hds/table/th-sort-test.js (hunk 3 above)
Comment (MelSumner) [reply packages/components/tests/integration/components/hds/table/th-sort-test.js:24]: The tests fail if I take it out: 
CleanShot 2022-10-25 at 19 45 21@2x

Comment (didoo) [reply packages/components/tests/integration/components/hds/table/th-sort-test.js:24]: What I mean is this. You can have much a simpler set of tests like this:

```
module('Integration | Component | hds/table/th-sort', function (hooks) {
  setupRenderingTest(hooks);

  test('it renders', async function (assert) {
    await render(hbs`<Hds::Table::ThSort>Artist</Hds::Table::ThSort>`);
    assert.dom(this.element).hasText('Artist');
  });
  test('it has the scope attribute, and it is set to column', async function (assert) {
    await render(hbs`<Hds::Table::ThSort>Artist</Hds::Table::ThSort>`);
    assert.dom('[data-test-table-th-sort]').hasAttribute('scope', 'col');
  });

    await render(hbs`<Hds::Table::ThSort>Artist</Hds::Table::ThSort>`);
    assert.dom('[data-test-table-th-sort]').hasClass('hds-table__th-sort');
  });

  test('if @sortOrder is not defined, the swap-vertical icon should be displayed', async function (assert) {
    await render(
      hbs`<Hds::Table::ThSort @sortBy='artist'>artist</Hds::Table::ThSort>`
    );
    assert.dom('[data-test-icon="swap-vertical"]').exists();
  });

  test('if `@sortOrder` is set to asc, the arrow-up icon should be displayed', async function (assert) {
    await render(
      hbs`<Hds::Table::ThSort @sortBy='artist' @sortKey='artist' @sortOrder='asc'>Artist</Hds::Table::ThSort>`
    );
    assert.dom('[data-test-icon="arrow-up"]').exists();
  });
});
```
where all the code related to `setSortBy` is completely removed, and to do that we need to change how the `setSortBy` is invoked. 

Currently is invoked like this:
```
<button type="button" {{on "click" (fn @setSortBy @sortKey)}}>
```
but this triggers an error if `@setSortBy` is not set. Instead we generally use a "guard" to avoid this error (you can see it applied in many components: `Alert`, `Breadcrumb`, `Disclosure`, `Toggle Button`, `Form Error + Helper Text`, `Tag`). Something like this (in `th-sort.js`):
```
const NOOP = () => {};

export default class HdsTableThSortComponent extends Component {
   ...

  /**
   * @param onClick
   * @type {function}
   * @default () => {}
   */
  get onClick() {
    let { setSortBy, sortKey } = this.args;

    // notice: this is a guard used in case the component is used as standalone element
    if (typeof setSortBy === 'function') {
      return () => {
        setSortBy(sortKey);
      };
    } else {
      return NOOP;
    }
  }
}
```

Now the invocation in the template is like this:
```
<button type="button" {{on "click" this.onClick}}>
```
and the `@setSortBy` argument is not a required argument anymore for the `ThSort` (sub) component (when used inside the `Table` is controlled and provided by the table itself) and this makes `ThSort` simpler and avoids the need to define it in the tests as argument for the component.

PS: if you are not convinced yet, do this experiment: go in the test file `table/th-sort-test.js` and replace this code:
```
  // since we'd have to repeat this for each test, we'll set it up once to be used in all tests
  hooks.beforeEach(function () {
    this.set('setSortBy', function (sortBy) {
      return sortBy || 'artist';
    });
  });
```
with this one (essentially **`setSortBy` becomes a completely empty function**):
```
  // since we'd have to repeat this for each test, we'll set it up once to be used in all tests
  hooks.beforeEach(function () {
    this.set('setSortBy', function () {});
  });
```
and run the tests: you will see that **they pass anyway**. This means the `setSortBy` defined in the tests is not needed by the component per se, it's just there to avoid to trigger an error when the compilation occurs if it's not defined (a check in the `fn` helper, I imagine).

Comment (MelSumner) [reply packages/components/tests/integration/components/hds/table/th-sort-test.js:24]: So this is more about adding `NOOP` to the component in order to simplify the tests. Okay, got it. I'll look into it, ty for the feedback.

Comment (MelSumner) [reply packages/components/tests/integration/components/hds/table/th-sort-test.js]: Ok after looking into it some more (I updated the `beforeEach` hook to use optional chaining and I see what you mean), I sort of agree but also am not very convinced that the th-sort component should be made generic, since it's entire explicit purpose is to create a sortable component. 

I will think about it some more tonight and come back to it tomorrow morning.

Comment (didoo) [reply packages/components/tests/integration/components/hds/table/th-sort-test.js]: The fact that the `ThSort` component can live without the `setSortBy` for me is telling us that this component should know anything about sorting. If you think about it the component does not "sort", but on "click" invokes a function (provided from outside) without even knowing what that function does. The sorting logic happens inside the `Table`, not inside the `ThSort` (it's the name that leads to think that this "sorts").

File: packages/components/tests/integration/components/hds/table/th-sort-test.js (hunk 6 above)
Comment (didoo) [reply packages/components/tests/integration/components/hds/table/th-sort-test.js:27]: @MelSumner we're already had this conversation multiple times: is common practice in out components to **not** add `data-test` attributes in the components. If you do a full search in the codebase (`/packages/components/addon/components/hds`) you will see that the only place where you find `data-test` attributes in the components it's in the new Table components.

File: packages/components/tests/integration/components/hds/table/th-sort-test.js [hunk 7]
```diff
@@ -8,37 +8,46 @@ module('Integration | Component | hds/table/th-sort', function (hooks) {
   // since we'd have to repeat this for each test, we'll set it up once to be used in all tests
   hooks.beforeEach(function () {
     this.set('setSortBy', function (sortBy) {
-      return sortBy || 'artist';
+      return sortBy?.noop();
```
didoo, line 11: This was not the suggested change. 

Let's discuss in person tomorrow, I'll set up a meeting.

packages/components/tests/integration/components/hds/table/th-sort-test.js
didoo: Sorry, but I don't see them? 🤔

packages/components/tests/integration/components/hds/table/index-test.js
didoo: sorry, I don't see the tests for the second two items, is it me? 🤔

also, regarding the first item, what I mean is to test the different possible contents:
- table with no model
- table with model
- table with model + sortable

and for each one of them, check that the expected cells (tr/th and tr/td) are rendered correctly

didoo (changes requested):

packages/components/tests/integration/components/hds/table/th-sort-test.js
MelSumner: I've updated all of these except the th-sort, that will need to stay the exception for now. If you disagree, let's set up a pairing time and you can show me how you would invoke `...attributes`. Otherwise consider this resolved, but will bring it up for future team discussion in general.

packages/components/tests/integration/components/hds/table/index-test.js
MelSumner: These have been moved to the integration test. Please see the PR description for the command to invoke the tests for this branch and a screenshot of the list of tests itself.

packages/components/tests/integration/components/hds/table/th-sort-test.js
MelSumner: I'll file an issue to make the change, but since that change is about the component and this PR is about the tests, let's keep it separate. Resolving this for now.

packages/components/tests/integration/components/hds/table/index-test.js
```diff
@@ -1,17 +1,224 @@
...
+  ]);
+  context.set('columns', [
+    { key: 'artist', label: 'Artist' },
+    { key: 'album', label: 'Album' },
+    { key: 'year', label: 'Year' },
+  ]);
+  context.set('sortBy', 'artist');
+  context.set('sortOrder', 'asc');
+};
+const renderSortableTable = async () => {
```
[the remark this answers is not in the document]
didoo: since the model is assigned directly, and not fetched, you don't need an async function. for this reason you can simplify the code even further, changing the `renderSortableTable` async function to a `const` string:
```
const hbsSortableTable = hbs`
  <Hds::Table ...>
    ...
  </Hds::Table>
`;
```
and then call the render function passing directly the string itself
```
await render(hbsSortableTable);
```

packages/components/tests/integration/components/hds/table/index-test.js
```diff
@@ -1,17 +1,224 @@
...
 module('Integration | Component | hds/table/index', function (hooks) {
   setupRenderingTest(hooks);
 
   test('it renders the component', async function (assert) {
     await render(hbs`<Hds::Table />`);
     assert.dom(this.element).exists();
   });
+
   test('it should render with a CSS class that matches the component name', async function (assert) {
+    await render(hbs`<Hds::Table id="data-test-table"/>`);
```
didoo, line 68: to avoid confusion, probably better to remove the `data-` prefix from the IDs

packages/components/tests/integration/components/hds/table/index-test.js
```diff
@@ -1,17 +1,224 @@
...
+
+  test('it should render with a CSS class appropriate for the @density value', async function (assert) {
+    await render(hbs`<Hds::Table @density="short" />`);
+    assert.dom('[data-test-table]').hasClass('hds-table--density-short');
+  });
+
+  test('it should render with a CSS class appropriate if no @density value is set', async function (assert) {
     await render(hbs`<Hds::Table />`);
-    assert.dom('[data-test-table]').hasClass('hds-table');
+    assert.dom('[data-test-table]').hasClass('hds-table--density-medium');
```
didoo, line 79: since the `Table` has `...attributes` we can remove also this `data-test` selector and use an ID

packages/components/tests/integration/components/hds/table/index-test.js
```diff
@@ -1,17 +1,224 @@
...
+          <Hds::Table::Tr>
+            <td>Cell Content 2 1</td>
+            <td>Cell Content 2 2</td>
+            <td>Cell Content 2 3</td>
+          </Hds::Table::Tr>
+          <Hds::Table::Tr>
+            <td>Cell Content 3 1</td>
+            <td>Cell Content 3 2</td>
+            <td>Cell Content 3 3</td>
+          </Hds::Table::Tr>
```
didoo, line 116: since we're testing only the first row, we can remove the other two and reduce the overall code

didoo (commented):
added a bunch of comments

packages/components/tests/integration/components/hds/table/index-test.js
MelSumner: sorry, where's the confusion here? it's just a fake id on a test?

packages/components/tests/integration/components/hds/table/index-test.js
MelSumner: I'm confused by this comment. Does this have an effect on how the code functions?

packages/components/tests/integration/components/hds/table/th-sort-test.js
MelSumner: ok I've refactored the th-sort component in a separate PR. #666

didoo (approved):

PR #655, opened 2022-10-26
[Table] Adds support for table density

MelSumner:
### :pushpin: Summary

If merged, this PR adds support for table density. Tracked internally with [HDS-816](https://hashicorp.atlassian.net/browse/HDS-814).

- added support for density. Current acceptable values are "packed", "default", and "spacious".
- added CSS classes for density
- added assertion if incorrect value is defined
- added test to make sure that correct class is applied if density is defined

➡️ ➡️ ➡️ **Note: there are some other small CSS fixes but the primary CSS update will come in a separate PR.** 

CleanShot 2022-10-27 at 11 44 37

### 👀 How to review
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/table/index.js
+++ b/packages/components/addon/components/hds/table/index.js
@@ -87,7 +87,7 @@ export default class HdsTableIndexComponent extends Component {
 
     // add a class based on the @density argument
     if (this.density) {
-      classes.push(`hds-table--${this.density}`);
+      classes.push(`hds-table--density-${this.density}`);
     }
 
     return classes.join(' ');
--- a/packages/components/addon/components/hds/table/index.js
+++ b/packages/components/addon/components/hds/table/index.js
@@ -52,12 +52,11 @@ export default class HdsTableIndexComponent extends Component {
     return this.args.isStriped ?? true;
   }
 
-  // values are still being decided by design. For now, we'll use packed, default, and spacious.
   /**
    * @param density
    * @type {string}
-   * @default 'default'
-   * @description Determines the density of the table cells; defaults to 'default'.
+   * @default 'medium'
+   * @description Determines the density of the table cells; options are "short", "medium" and "tall". If no density is defined, "medium" is used.
    */
   get density() {
     let { density = DEFAULT_DENSITY } = this.args;
--- a/packages/components/addon/components/hds/table/index.js
+++ b/packages/components/addon/components/hds/table/index.js
@@ -48,6 +48,17 @@ export default class HdsTableIndexComponent extends Component {
     return this.args.isStriped ?? true;
   }
 
+  // values are still being decided by design. For now, we'll use packed, medium, and spacious. We've suggested "size" like we use for some of the other components like the button.
+  /**
+   * @param density
+   * @type {string}
+   * @default 'default'
+   * @description Determines the density of the table cells; defaults to 'medium'.
+   */
+  get density() {
+    return this.args.density ?? 'default';
+  }
+
   /**
    * Get the class names to apply to the component.
    * @method classNames
@@ -59,6 +70,11 @@ export default class HdsTableIndexComponent extends Component {
     // add a class based on the @isStriped argument
     if (this.isStriped) {
       classes.push('hds-table--striped');
+    }
+
+    // add a class based on the @density argument
+    if (this.density) {
+      classes.push(`hds-table--${this.density}`);
     }
 
     return classes.join(' ');
--- a/packages/components/addon/components/hds/table/index.js
+++ b/packages/components/addon/components/hds/table/index.js
@@ -3,8 +3,8 @@ import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { assert } from '@ember/debug';
 
-const DENSITIES = ['packed', 'default', 'spacious'];
-const DEFAULT_DENSITY = 'default';
+const DENSITIES = ['short', 'medium', 'tall'];
+const DEFAULT_DENSITY = 'medium';
 
 export default class HdsTableIndexComponent extends Component {
   @tracked sortBy = this.args.sortBy;
--- a/packages/components/addon/components/hds/table/index.js
+++ b/packages/components/addon/components/hds/table/index.js
@@ -1,6 +1,10 @@
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
+import { assert } from '@ember/debug';
+
+const DENSITIES = ['packed', 'default', 'spacious'];
+const DEFAULT_DENSITY = 'default';
 
 export default class HdsTableIndexComponent extends Component {
   @tracked sortBy = this.args.sortBy;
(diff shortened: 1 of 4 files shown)


jorytindall (approved):
Visually everything looks good here 🚀, barring the conversation about the `density` prop naming. To capture here, could be:

- Size
  - Small
  - Medium (default)
  - Large

_or_

- Padding
  - Small
  - Medium
  - Large
  
_or_

- Density
  - Small
  - Medium
  - Large
 
My 2 cents: density makes the most sense to me since it's referring to the spacing around the content as it relates to the container. Though, understanding that the terms associated with density might be a little esoteric.

MelSumner:
@jorytindall density as a prop name is fine, it's the values that feel a little off to me. Not in a strong way but in a small nagging "you forgot something on your shopping list" way. 🤷

packages/components/tests/integration/components/hds/table/index-test.js
```diff
@@ -16,6 +16,11 @@ module('Integration | Component | hds/table/index', function (hooks) {
     assert.dom('[data-test-table]').hasClass('hds-table');
   });
 
+  test('it should render with a CSS class appropriate for the @density value', async function (assert) {
```
didoo, line 19: we should test also the default is set, when no `@density` is provided

packages/components/tests/integration/components/hds/table/index-test.js
```diff
@@ -16,6 +16,11 @@ module('Integration | Component | hds/table/index', function (hooks) {
...
+    await render(hbs`<Hds::Table @density="packed" />`);
+    assert.dom('[data-test-table]').hasClass('hds-table--packed');
```
didoo, line 21: plus, better to use `id` to target the element instead of `data-test`, when `...attributes` is available on the element (also for the other tests in this file).

didoo (changes requested):
added a few comments, one of them is a blocker (the height of the table header is not the same as in Figma)

packages/components/addon/components/hds/table/index.js
```diff
@@ -61,6 +85,11 @@ export default class HdsTableIndexComponent extends Component {
       classes.push('hds-table--striped');
     }
 
+    // add a class based on the @density argument
+    if (this.density) {
+      classes.push(`hds-table--${this.density}`);
```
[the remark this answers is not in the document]
MelSumner: ahh I copied this from another component, I should have caught that. Thanks!

packages/components/app/styles/components/table.scss
```diff
@@ -16,32 +17,53 @@
...
 .hds-table--striped {
   tbody {
-    tr:nth-child(odd) {
+    tr:nth-child(even) {
       background-color: var(--token-color-surface-faint);
     }
   }
 }
 
+.hds-table--density-short td {
```
didoo, line 26: since in the other PR we're using `.hds-table__tbody td` as selector, to avoid risks of collision in specificity I would suggest to use `.hds-table--density-short .hds-table__tbody td` here (it works also without, because we're changing different props, but is safer)

packages/components/app/styles/components/table.scss
```diff
@@ -16,32 +17,53 @@
...
   border-bottom: 1px solid var(--token-color-border-primary);
 
-  th {
+  .hds-table__th,
+  .hds-table__th-sort {
+    height: 48px;
     padding: 8px 16px;
     color: var(--token-color-foreground-strong);
     font-weight: var(--token-typography-font-weight-semibold);
+    font-size: 14px;
```
didoo, line 48: I think I have already mentioned elsewhere: we should use the CSS helper classes in the templates (or design tokens in the CSS) instead of these values (in Figma I see they've used standard text styles)

packages/components/app/styles/components/table.scss
```diff
@@ -16,32 +17,53 @@
...
 
-    button { // will refactor this to a button. Use Hds::Button?
+    button {
       display: flex;
       align-items: center;
       width: 100%;
+      height: 100%;
       padding: 8px 16px;
       color: var(--token-color-foreground-strong);
+      font-weight: var(--token-typography-font-weight-semibold);
```
didoo, line 64: I think I have already mentioned elsewhere, but here you can use `inherit` for the font stylings

didoo (commented):
left a few comments

packages/components/app/styles/components/table.scss
MelSumner: is it any riskier? we don't have td elements in the thead element.

packages/components/app/styles/components/table.scss
MelSumner: Let's work on this in the PR related to styles (https://github.com/hashicorp/design-system/pull/658)

packages/components/app/styles/components/table.scss
didoo: `.hds-table--density-short td` has the same specificity as `.hds-table__tbody td`, so there are risks of conflict. if the "baseline" of the `TD` is declared as `.hds-table__tbody td` then `.hds-table--density-short` is a specialized version of the `TD` which means its selector should be `.hds-table--density-short hds-table__tbody td`

MelSumner: this has been changed in PR #658

packages/components/addon/components/hds/table/index.js
```diff
@@ -48,6 +52,25 @@ export default class HdsTableIndexComponent extends Component {
     return this.args.isStriped ?? true;
   }
 
+  /**
+   * @param density
+   * @type {string}
+   * @default 'medium'
+   * @description Determines the density of the table cells; options are "short", "medium" and "tall". If no density is defined, "medium" is used.
```
KristinLBradley, line 59: Comment: "short", "medium" and "tall" sound more like descriptors for `height` vs. `density`. Have we looked at what other design systems use for similarly named props? @jorytindall @MelSumner @heatherlarsen 

Looking at Gmail, they have a `density` setting with values for "default", "comfortable", and "compact".

Other possible descriptors could be:
* high, medium, low
* compact, medium/default, loose?

or other descriptive words more along those lines...

MelSumner: @KristinLBradley there's a long Slack thread, I'll ping you in our channel 👍

KristinLBradley: Something along the lines of what @heatherlarsen mentioned in the Slack thread makes sense to me. 

IE: "dense", "default/medium", and "spacious"

packages/components/tests/dummy/app/templates/components/table.hbs
```diff
@@ -343,9 +343,9 @@ export default class ComponentsTableRoute extends Route {
     </:body>
   </Hds::Table>
   <h4 class="dummy-h4">
-    Static table styled as "packed"
+    Static table with density set to "short"
   </h4>
-  <Hds::Table @model={{this.model.data}} @density="packed">
+  <Hds::Table @model={{this.model.data}} @density="short">
```
alex-ju, line 348: thought: we seem to be in the habit of taking the enum values from the component and [looping through them in the showcase section](https://github.com/hashicorp/design-system/blob/main/packages/components/tests/dummy/app/templates/components/dropdown.hbs#L1041) – I guess it has the advantage of keeping the two in sync with more ease.

alex-ju (approved):
Code looks good to me – I think it would be relatively easy to change the allowed argument values before release if we decide to call them something else.

packages/components/app/styles/components/table.scss
didoo: OK, I'll review/check it there.Thanks

didoo (approved):
for changes related to table density, it's OK for me to merge as is now. 
for other things, I'll comment in the other two PRs (#650 #658)

packages/components/addon/components/hds/table/index.js
MelSumner: I wonder if you'd want to file a JIRA ticket and then we can ask Jory to re-visit when he's back from vacation? Otherwise, I'm tempted to leave it as is for now so we have the alignment with the Figma API. WDYT?

packages/components/tests/dummy/app/templates/components/table.hbs
MelSumner: Ahh yes, I suppose that does happen sometimes. Good catch! 👍

PR #657, opened 2022-10-26
[WIP] New docs website -  Exploration of page templates and routes in Field-Guide

didoo:
### :pushpin: Summary

TODO

TODO

### :link: External links

JIRA ticket: https://hashicorp.atlassian.net/browse/HDS-826

***

### 👀 How to review

👉 Review by files changed


website/app/components/doc-landing-callout-blocks.hbs
```diff
@@ -0,0 +1,7 @@
+<ul class="doc-landing-callout-blocks doc-landing-callout-blocks--layout-{{@cols}}cols">
+  {{#each @items as |item|}}
+  <li class="doc-landing-callout-blocks__item">
+    <a class="doc-landing-callout-blocks__link" href={{item.url}}>{{item.text}}</a>
```
Dhaulagiri, line 4: I don't know that it needs to be addressed in this PR, but manually crafting urls like this feels a bit un-Ember-y to me and there is likely a more idiomatic way to use a normal `{{link-to}}` to accomplish this same thing.

website/app/components/doc-landing-callout-blocks.hbs
```diff
@@ -0,0 +1,7 @@
+<ul class="doc-landing-callout-blocks doc-landing-callout-blocks--layout-{{@cols}}cols">
+  {{#each @items as |item|}}
+  <li class="doc-landing-callout-blocks__item">
```
Dhaulagiri, line 3: [very minor nit that I'm not sure why the linter doesn't complain] this block should be indented inside the `{{#each}}`

website/app/components/doc/page/sidebar.js
```diff
@@ -0,0 +1,41 @@
+import Component from '@glimmer/component';
+
+export default class DocPageSidebarComponent extends Component {
+  get navigationTree() {
```
Dhaulagiri, line 4: it might be useful to add some comments as to what this function is doing so we can better understand it

website/app/controllers/foundations.js
```diff
@@ -0,0 +1,30 @@
+import Controller from '@ember/controller';
+
+export default class FoundationsController extends Controller {
+  items = [
+    {
+      text: 'Tokens',
+      url: 'tokens/03--how-to-use/',
```
Dhaulagiri, line 7: i think eventually we'll want to find a way to uncouple these so tightly from the specific paths like this as the names of these files seem like they could change on us?

Dhaulagiri (approved):
couple of minor comments/thoughts but nothing blocking from my perspective

website/app/components/doc-landing-callout-blocks.hbs
didoo: these are super-temporary placeholders, they will definitely be replaced when we will build the actual landing pages. but I'll keep in mind it for then. thanks

website/app/controllers/foundations.js
didoo: yes, once we finalize the information architecture and we decide which tool to use, we will also have the information about what are called the "index" pages of the components. for now this is the only way to link them

website/app/components/doc/page/sidebar.js
didoo: OK

didoo:
@Dhaulagiri I've pushed a couple of small updates per your comments.

website/app/controllers/foundations.js
Dhaulagiri: 👍  makes sense!

PR #658, opened 2022-10-27
[Table] Adds border-radius to table and refines table style

MelSumner:
### :pushpin: Summary

If merged, this PR adds a border radius to the table and cleans up the table styles.

CleanShot 2022-10-27 at 12 30 54

### 👀 How to review

🚨 **THIS IS THE PR WHERE ITS FINE TO REVIEW ALL STYLE FOR THE TABLE** 🚨 

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/app/styles/components/table.scss
```diff
@@ -2,16 +2,21 @@
 // TABLE
 //
 //
+
+$hds-table-border-radius: 6px; // table border radius
+$hds-table-border-width: 1px;
+$hds-table-inner-border-radius: $hds-table-border-radius - $hds-table-border-width;
+$hds-table-border-color: var(--token-color-border-primary);
```
didoo, line 9: [note to self] later we may convert these to component-level design tokens

packages/components/app/styles/components/table.scss
```diff
@@ -23,11 +28,11 @@
   }
 }
 
+// DENSITY (ROW HEIGHT)
```
didoo, line 31:
```suggestion
// DENSITY (ROW HEIGHT)

```

packages/components/app/styles/components/table.scss
```diff
@@ -36,8 +41,27 @@
   padding: 20px 16px;
 }
 
+
 .hds-table__thead {
-  border-bottom: 1px solid var(--token-color-border-primary);
+  .hds-table__tr {
+    color: #0c0c0e;
+    background-color: #f1f2f3;
```
didoo, line 48: Looking at the design in Figma, these are standard colors. Better use design tokens (`--token-color-blah-blah`).

packages/components/app/styles/components/table.scss
```diff
@@ -36,8 +41,27 @@
...
+
+    &:first-of-type {
+      border-top-left-radius: $hds-table-inner-border-radius;
+      border-top-right-radius: $hds-table-inner-border-radius;
```
didoo, line 52: I _think_ that if you move the color/background declarations there are above (for the row) inside the `hds-table__th, .hds-table__th-sort {}` selector, probably you can remove these two (there's no background color, no need to set a border radius)

packages/components/app/styles/components/table.scss
```diff
@@ -36,8 +41,27 @@
...
+
+      th:first-child {
+        border-left: none;
```
didoo, line 55: I _think_ you don't need this (see line 79)

packages/components/app/styles/components/table.scss
didoo: plus, the background-color is already set in the `TH` elements, so it's useless here (try to set it red, and you will not see it). my suggestion is to delete the background color from here, move the `color` declaration in the `TH` elements, so you can also remove the border radius applied to the `TR` (see comment below)

packages/components/app/styles/components/table.scss
```diff
@@ -103,6 +131,36 @@
   background-color: var(--token-color-surface-primary);
 
   .hds-table__tr {
-    border-bottom: 1px solid var(--token-color-border-primary);
+    td {
+      border-top: none;
+      border-right: $hds-table-border-width solid $hds-table-border-color;
+      border-bottom: $hds-table-border-width solid $hds-table-border-color;
+      border-left: none;
```
didoo, line 138: I see here you're using a different approach: you're applying the color to the `hds-table__tbody` and not to the `TD` (compared to the table head). maybe use the same approach you've used above, also here? (apply the background color and the text color to the `TD` element)?

packages/components/app/styles/components/table.scss
```diff
@@ -103,6 +131,36 @@
...
+
+      &:last-child {
+        border-right: none;
+      }
+    }
+
+    &:nth-of-type(2n) {
+      background-color: #f8f8f8;
```
didoo, line 146: you can use a design token (as CSS custom prop) here

packages/components/app/styles/components/table.scss
```diff
@@ -103,6 +131,36 @@
...
+    }
+
+    &:last-of-type {
+      border-bottom-right-radius: $hds-table-inner-border-radius;
+      border-bottom-left-radius: $hds-table-inner-border-radius;
```
didoo, line 151: I _think_ you can remove these two, they're not needed

packages/components/app/styles/components/table.scss
```diff
@@ -103,6 +131,36 @@
...
+      td {
+        border-bottom: none;
+      }
+
+      td:first-child {
+        border-bottom-left-radius: $hds-table-inner-border-radius;
+      }
+
+      td:last-child {
+        border-bottom-right-radius: $hds-table-inner-border-radius;
```
didoo, line 162: I _think_ you have to remove also the border right here

didoo (changes requested):
added a few comments

jorytindall:
These are all potentially nit-picky points, but I'll capture them here:

The non-sortable `th` and the sortable `th` seem to have a very slight difference in font-weight, but I can't quite figure out what is causing the difference as the token values are the same. Image attached for reference.
table-header-label-comparison

The focus ring on the sortable `th` isn't inheriting the rounded corner of the table. Attaching two examples, the second with the value explicitly set on the focus state for visual reference.

Screen Shot 2022-10-27 at 1 45 57 PM

Screen Shot 2022-10-27 at 1 45 46 PM

There were some concerns about changing the color indeterminate sort icon to match the ascending/descending sort, I'll circle back on that after discussing with the team.

packages/components/app/styles/components/table.scss
MelSumner: Good catch, updated.

MelSumner:
@jorytindall 
1. I noticed the sortable header text thing too but since the code was the same I thought it was my imagination. I'll dig deeper. 
2. I see what you mean about the border on the focus, I'll fix.

packages/components/app/styles/components/table.scss
```diff
@@ -51,11 +51,27 @@ $hds-table-border-color: var(--token-color-border-primary);
...
+            border-top-left-radius: $hds-table-inner-border-radius;
+          }
+        }
       }
 
       th:last-child {
         border-right: none;
         border-top-right-radius: $hds-table-inner-border-radius;
+
+        button { // I don't like this, it feels weird to have declared in multiple places- open to ideas on how to organize this CSS so we don't have to repeat selectors like this.
```
MelSumner, line 68: cc @didoo I'm open to some suggestions on this - I am not _quite_ sure how to organize this.

packages/components/app/styles/components/table.scss
```diff
@@ -107,6 +124,7 @@ $hds-table-border-color: var(--token-color-border-primary);
       &:focus-visible {
         color: var(--token-color-foreground-strong);
         border-color: var(--token-color-focus-action-external);
+        outline: 0; // I'm not sure what to do here because we'll need our outline style but you can't change the outline radius
```
MelSumner, line 127: cc @didoo I am not sure how to approach this, since we can't control the radius of the outline.

MelSumner:
Update: Using the focus mixin that we've used on other components seems problematic, trying to figure out exactly what's going wrong so it can be fixed

Screenshot of the issue: 
CleanShot 2022-10-28 at 14 26 10@2x

didoo:
> Update: Using the focus mixin that we've used on other components seems problematic

I've done a quick test and I've been able to have it working by:
- removing the border from the button (there are different places where to remove it)
- removing the outline from the button (`outline: none;`)
- setting `border-radius: inherit;` on the button (so it has the exact same border as the table cell)
- creating a new stacking context for the button via `isolation: isolate;` (we may consider applying it only on focus)

https://user-images.githubusercontent.com/686239/198724879-fea33d49-a8ea-4864-ae19-cada55628b4d.mov

if you want we can see it together on Monday

MelSumner:
Ok I've re-factored this and have the focus styles working. There are a lot of comments in this file but I felt like they were necessary because the answers were not super-obvious to me so they might not be to the rest of the team either. 

I tried using the focus mixin (in a few different locations) but it does not give the effect we want, so I copy/pasted the box-shadow portion directly and it seems to work. 

Here's a video of moving the focus around:
https://user-images.githubusercontent.com/4587451/198897648-6c3b2fd7-fb61-4dbe-a961-b4a9105df1ac.mp4

packages/components/app/styles/components/table.scss
```diff
@@ -2,107 +2,173 @@
...
+      color: var(--token-color-foreground-strong);
+      font-weight: var(--token-typography-font-weight-semibold);
+      font-size: 14px;
+      line-height: 1.4;
+      text-align: left;
+      background-color: var(--token-color-surface-strong);
+      border-top: none;
+      border-right: $hds-table-border-width solid $hds-table-border-color;
+      border-bottom: $hds-table-border-width solid $hds-table-border-color;
+      border-left: none;
```
KristinLBradley, line 40: I experimented with simplified styles fro borders and border radius. The main drawback is that it requires an extra wrapper div around the table: https://codepen.io/KristinB/pen/rNKxxPz/ec9d9ad82552b4794d7cea417b71936e
@MelSumner

MelSumner: Thanks for sharing! I considered that approach, but overall I tend to lean toward less markup, since there is a limit to how many DOM nodes can exist before a screen reader will crash (and this is pretty easy to do in technical applications).

packages/components/app/styles/components/table.scss
```diff
@@ -2,107 +2,173 @@
...
+    }
+
+    .hds-table__th {
+      padding: 12px 16px;
```
alex-ju, line 44: thought: I see this padding quite a bit across the code (`.hds-table__th`, `.hds-table__th-sort button` and `.hds-table--density-medium .hds-table__tbody td`) and was wondering if we'd want to store these padding values (and maybe the other densities) in SCSS variables.

alex-ju (approved):
Cristiano and Kristin already did the heavy lifting with the review and experimentation on this one. I checked the final code and had a play in the browser and it looks good!

packages/components/app/styles/components/table.scss
MelSumner: That's probably a good idea, thanks for mentioning it!

MelSumner:
@didoo I figured this out offline so I didn't see your comment until after I'd already shipped a solution, thanks for being willing to pair though! At any rate, can you give this PR another look and see if this implementation is good to go? Thanks!

packages/components/app/styles/components/table.scss
MelSumner: ok I've extracted the padding values to variables. Looking at other components it seems like we _also_ do some more complex loops, but since the padding is the only thing that changes (vs the button where nearly all the props change based on the size), didn't go that route.

KristinLBradley (approved):

didoo (approved):

PR #661, opened 2022-10-31
`Pagination` - Main feature branch (HDS-943)

KristinLBradley:
### :pushpin: Summary
If merged, this PR will add the PaginationBar component to the design system.

**Previews:** 
- helios documentation: https://hds-website-git-hds-943-pagination-component-hashicorp.vercel.app/components/pagination
- showcase: https://hds-components-git-hds-943-pagination-component-hashicorp.vercel.app/components/pagination

- Implemented the following components and sub-components:
   - `Hds::Pagination::Numbered`
   - `Hds::Pagination::Compact`
   - `Hds::Pagination::Info`
   - `Hds::Pagination::SizeSelector`
   - `Hds::Pagination::Nav::Arrow`
   - `Hds::Pagination::Nav::Number`
   - `Hds::Pagination::Nav::Ellipsis`
- Added documentation ("helios") + showcase ("scrappy") for the pagination components/sub-components
- Added testing (integration + acceptance) for the pagination components/sub-components

### :link: External links
- https://hashicorp.atlassian.net/browse/HDS-943
- https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components?node-id=18709%3A42011&t=m6fSsOk5Y67zfZRB-0

***

### 👀 How to review

Would like feedback on implementations, documentation content, and any details that still need to be addressed before release.

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/app/styles/components/pagination.scss
```diff
@@ -0,0 +1,177 @@
...
+  display: flex;
+  align-items: center;
+  padding: 0 var(--token-pagination-control-padding-horizontal);
+}
+
+.hds-pagination-nav__control {
+  height: var(--token-pagination-control-height);
+  text-decoration: none;
+  background: transparent;
+  border: none;
```
didoo, line 64: if this is a button, this should be 1px solid transparent (so the height needs to be calculated: `height: calc(var((--token-pagination-control-height) - 2px)`

packages/components/app/styles/components/pagination.scss
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  &:disabled,
+  &.mock-disabled {
+    color: var(--token-color-foreground-disabled);
+    cursor: not-allowed;
+  }
+}
+
+// Make buttons with "Previous" & "Next" labels the same min-width
```
didoo, line 90: maybe explay the "why" of this, in the comment

packages/components/app/styles/components/pagination.scss
```diff
@@ -0,0 +1,177 @@
...
+    }
+  }
+
+  // indicator underline
+  &::before {
+    position: absolute;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    width: var(--token-pagination-indicator-width);
```
didoo, line 143: this fixed width is not resilient. what if the text is 4 chars? (eg. page 1234)? in that case the width would be be correct (unless by design is fixed).
 if instead is proportional to the actual width of the container, then the `left/right` properties are more apt for this purpose

packages/components/app/styles/components/pagination.scss
```diff
@@ -0,0 +1,177 @@
...
+  justify-self: flex-end;
+
+  > label {
+    white-space: nowrap;
+  }
+
+  > select {
+    height: 28px;
+    margin-left: 12px;
+    padding-top: 0;
```
didoo, line 167: probably we can use a single `padding` declaration here

didoo (commented):
@KristinLBradley I had a look at the CSS code and found nothing out of order :) 
only added very minor comments (let me know if you want I can open directly the PR for these small fixes)

packages/components/app/styles/components/pagination.scss
didoo: Above browser, below Figma
screenshot_2209

packages/components/app/styles/components/pagination.scss
KristinLBradley: Why do they need a transparent border? Should they display a border on hover or something? I don't see something like that in the Figma mocks.

I tried making the above changes and the buttons were too short.

didoo: > Why do they need a transparent border? Should they display a border on hover or something? I don't see something like that in the Figma mocks.

It's because of accessibility (something I learned from @MelSumner 🙏): it's for users that may use high-contrast mode/settings in their browser.

> I tried making the above changes and the buttons were too short.

Yes, now looking at the actual CSS code is slightly more complex (but not too complex)
- you can add a 1px border without changing the actual `height` value (since there is no padding, and the button has `box-sizing: border-box`
- but you have to change the position of the focus ring, reducing its "inset" value by 1px (the position absolute is calculated from the inside of the border): `--token-pagination-control-focus-inset: 4px;`

KristinLBradley: I pushed up a number of changes to add the transparent border without shifting the rest of the layout.

didoo (approved):
🚢 !

Issue #662, opened 2022-11-01
Bug report: design-system-components 1.3.0 incompatible w/ Ember 3.24


nfagerlund:
## IS: Current behavior

After upgrading TFC to design-system-components 1.3.0, I got whomped by a build error (shown below). 

The problem seems to be in the Hds::Tabs::Panel template, which includes an explicit helper invocation directly within a value expression (like `{{(not thing)}}` w/ inner parens, as opposed to `{{not thing}}`).

Explicit helper invocations directly inside a value expression [became valid syntax in Ember 3.27](https://rfcs.emberjs.com/id/0432-contextual-helpers/), to support the contextual helpers as values feature. 

TFC is still on Ember 3.24.

```
Expecting 'ID', 'STRING', 'NUMBER', 'BOOLEAN', 'UNDEFINED', 'NULL', 'DATA', got 'OPEN_SEXPR'
        in /tmp/broccoli-115axhrR6gnNZSn/out-268-template_compiler
        at broccoli-persistent-filter:Babel > [Babel: @*********/design-system-components] (Babel: @*********/design-system-components)
  - name: Error
  - nodeAnnotation: Babel: @*********/design-system-components
  - nodeName: broccoli-persistent-filter:Babel > [Babel: @*********/design-system-components]
  - originalErrorMessage: /home/circleci/project/frontend/atlas/@*********/design-system-components/components/hds/tabs/panel.js: Parse error on line 7:
...anelId}}  hidden={{(not this.isSelected
```

## SHOULD BE: Expected behavior

Should be compatible with Ember 3.24, since design-system-components declares 3.24 as supported. This compatibility break is definitely accidental!

## Steps to reproduce

1. Upgrade to design-system-components 1.3.0 in an Ember 3.24 app, and attempt to build the app.

## Possible solution

There's only one instance of the problematic syntax in the whole addon. I'll get a PR up.

PR #670, opened 2022-11-02
clarify ember support

Dhaulagiri:
### :pushpin: Summary

https://github.com/hashicorp/design-system/issues/662 noted that we specify support for Ember 3.24 when in reality our explicit support is for 3.28 and newer, with 3.24 more of a best effort. This PR makes a slight adjustment to the readmes to make that more clear.

### DIFF
--- a/packages/components/README.md
+++ b/packages/components/README.md
@@ -9,8 +9,9 @@ A package containing the components for the HashiCorp Design System.
 Compatibility
 ------------------------------------------------------------------------------
 
-* Ember.js v3.24 or above
-* Ember CLI v3.24 or above
+* Ember.js v3.28 or above
+  * Note: The library _should_ work with earlier versions of Ember, but we only test with Ember 3.28 and newer
+* Ember CLI v3.28 or above
 * Node.js v12 or above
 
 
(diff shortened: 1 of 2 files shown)


packages/components/README.md
```diff
@@ -9,7 +9,8 @@ A package containing the components for the HashiCorp Design System.
 Compatibility
 ------------------------------------------------------------------------------
 
-* Ember.js v3.24 or above
+* Ember.js v3.28 or above
+  * Note: The library _should_ work with earlier versions of Ember, but we only test with Ember 3.28 and newer
 * Ember CLI v3.24 or above
```
MelSumner, line 14: Should this version align as well?

Dhaulagiri: That's a good question. In practice my experience is that ember versions are almost entirely decoupled from ember-cli and we don't have any way of testing what the oldest version of ember-cli that would work with our components is either. It almost makes me feel like taking out the ember-cli line makes the most sense. What do you think?

MelSumner: Since we have custom blueprints, we should probably leave it in, even though that relates more to creating components than using them.

Dhaulagiri: I updated the ember-cli lines to also state 3.28

alex-ju (approved):

PR #674, opened 2022-11-03
New docs website / New design specs - Typography

didoo:
### :pushpin: Summary

Ported the marketing typographic styles to our new website (using this file: https://github.com/hashicorp/mktg-global-styles/blob/main/components/g-type.css).

Preview: https://hds-website-dxqytbhbc-hashicorp.vercel.app/testing/00-testing-new-designs/typography/

**Things to discuss/decide:**

- should the heading class names called `doc-text-h1` or `doc-text-display1` ?
- ~~I have implemented the styles exactly as they are in the Marketing codebase, but they may be different from the Figma files. In that case, what counts as source of truth for us?~~
  - Discussed with @heatherlarsen: our Figma files will be the source of truth
- I have implemented the typography so that is "responsive": was this intended also for us?
  - the markdown headings instead are not responsive (as visualized in the Figma file): is this OK?

_Notice: the `inline code` treatment is not implemented yet, only the `code` typographic style._

### :link: External links

JIRA tasks: 
- https://hashicorp.atlassian.net/browse/HDS-838 (design)
- https://hashicorp.atlassian.net/browse/HDS-967 (development)

Figma file:
- https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=181%3A23422

***

### 👀 How to review

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


didoo:
@heatherlarsen I have ported the typographic styles in the new website codebase, you can preview them here: https://hds-website-dxqytbhbc-hashicorp.vercel.app/testing/00-testing-new-designs/typography/ - have a look also at the things to discuss/decide in the PR description

@KristinLBradley I know you don't have much context in what we're doing, but could you please review the CSS part of this ticket and see if there are things that don't convince you or you would do differently? Thanks

website-html-to-markdown/moveover/testing/00-testing-new-designs/typography.md
```diff
@@ -0,0 +1,64 @@
+<section class="doc-page-temp-testing-typography">
+    <kbd>&lt;h1 class="doc-text-h1"&gt;</kbd>
```
KristinLBradley, line 2: Super Nit: The `code` element seems more appropriate here. (`kbd` is meant to represent keyboard input.)

(I know this is just sample code to show the visual designs though.)

website/app/styles/markdown/typography.scss
```diff
@@ -1,29 +1,29 @@
-@use "../typography/mixins.scss";
+@use "../typography/mixins";
 
 .doc-markdown-content {
 
   // HEADINGS
 
   > h1 {
```
KristinLBradley, line 7: Would we be able to ensure that these tags are always direct children of the `.doc-markdown-content` parent element when the markdown is rendered to HTML?

website/app/styles/typography/index.scss
```diff
@@ -1,2 +1,56 @@
 @import "./font-faces";
 @import "./mixins";
+
+// TYPOGRAPHIC STYLES
+
+// HEADERS
+// TODO! discuss with Heather if we use `docs-text-h1` or `docs-text-display1`
```
KristinLBradley, line 7: For me, I would prefer to reference the tag names because I think that makes the class names clearer and easier to use and it also matches what's in the Figma doc: https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=181%3A23422 

(I don't know how formal that Figma is though or if it's just a rough representation of curent visual styles.)

website/app/styles/typography/mixins.scss
```diff
@@ -1,63 +1,235 @@
...
+}
+
+
+// ---------
+// CODE
+// ---------
+
 @mixin doc-font-style-code {
   @include doc-font-family("mono");
   font-size: 16px;
```
didoo, line 235: @didoo check the value here, is not the right one

website-html-to-markdown/moveover/testing/00-testing-new-designs/typography.md
didoo: yes, the reason why I used `kbd` instead of `code` is because `<code>` is styled in markdown and I didn't want to mix & match tags, so I picked one HTML tag that was "easy" ;) 

in any case it will be deleted later, is only for testing

heatherlarsen:
Notes from our chat (for tracking):

1. re. Markdown headings: This was my mistake; I'd missed doing this for the markdown styles. I've updated them in the [Figma file](https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=182%3A18483) to match the regular type styles. 
2. I took the `inline code` definition out of the markdown styles; we'll address this with a component later.
3. I've updated the `blockquote` definition to 17/30. 
4. Let's move forward with the `code` style (13.5/26) from Marketing, with the understanding that we will likely change it as we work on the code components.

website/app/styles/markdown/typography.scss
didoo: This is the challenge :) 

The reason for this way of writing CSS selectors is that we want to avoid leaking of the styles to the HDS components, in case they're embedded within the markdown.

In _theory_, we could imagine to _always_ make sure that the rendered markdown is wrapped in a container with class `doc-markdown-content` (potentially with `display: contents`).

In _practice_ what we will try to do is to leverage [this feature of Showdown](https://showdownjs.com/docs/tutorials/add-default-class-to-html/) that allows to assign custom classes to the HTML elements generated via markdown, in which case we can completely avoid the need of direct descendants and `doc-markdown-content` wrappers (there's a ticket for that: 
[HDS-823](https://hashicorp.atlassian.net/browse/HDS-823))

website/app/styles/typography/mixins.scss
```diff
@@ -1,63 +1,235 @@
...
+
+// H1
+
+@mixin doc-font-style-h1($responsive: true) {
   @include doc-font-family("gilmer");
   font-weight: 700;
-  font-size: 3.052rem;
-  line-height: 1.3;
+
+  @if $responsive {
```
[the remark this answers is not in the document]
didoo: Yes, good idea. I am not sure we will have a "styleguide page" for the website, we'll see (in which case we'll definitely follow your suggestion).

website/app/styles/markdown/typography.scss
KristinLBradley: Thanks for the background info. I understand you did it that way to avoid style leakage.

Would it be possible to render HDS components inside iframes as another possible solution to style leakage? Just curious as that's the way Storybook handles it.

didoo: > Would it be possible to render HDS components inside iframes as another possible solution to style leakage? Just curious as that's the way Storybook handles it.

That was my first though! Hence my toying with the idea of a React application for the documentation, that would render "rich" markdown (via MDX) + Ember components (via iframe). 

The "problems" to overcome when using iframes are multiple though:
- each iframe should load its own JS files and CSS files; essentially, it would be a full blown Ember app - which means we should have a second web-app that renders the content of the iframe, but in that case how would we pass the "code" to the iframe? it would be a URL (not great)? or inline code that is rendered at runtime (but would it work in Ember? I did something similar with React, but I am not sure is possible with Ember)
- each frame should auto-resize to fit its content (and in that case, what if the content can overflow the bounding box, like a dropdown for example?)
- in a typical documentation page, we could have a hundred instances of a component, which means hundreds of iframes loading in parallel (probably not an issue from the bandwidth perspective, because of the cache, but JS should still need to be interpreted for each one of that instances) and hundreds of iframes being resized in parallel

If you want we can rubberduck a bit more this idea, and see if we can come up with solutions to the problems above.

> that's the way Storybook handles it

Yes, but in Storybook it's only one single large iframe, no?

PR #676, opened 2022-11-03
`PaginationBar` styling (HDS-944)

KristinLBradley:
### :link: External links

***

### 👀 How to review

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/pagination-bar/index.hbs
+++ b/packages/components/addon/components/hds/pagination-bar/index.hbs
@@ -1,4 +1,65 @@
 {{! ADD YOUR TEMPLATE CONTENT HERE }}
 <div class={{this.classNames}} ...attributes>
-  {{yield}}
+  <div class="hds-pagination-bar__total-count">1–10 of 100</div>
+
+  <nav class="hds-pagination-bar__pagination">
+    <ul class="hds-pagination-bar__nav-list">
+      <li>
+        <a class="hds-pagination__nav-item hds-pagination__nav-item--previous" href="/page-?">
+          <FlightIcon @name="chevron-left" class="hds-pagination__nav-icon" /><span class="hds-pagination__nav-item-label">Previous</span>
+        </a>
+      </li>
+      <li class="hds-pagination__current-page">
+        <a class="hds-pagination__nav-item" href="/page-1">1</a>
+      </li>
+      <li>
+        <a class="hds-pagination__nav-item" href="/page-2">2</a>
+      </li>
+      <li>
+        <a class="hds-pagination__nav-item" href="/page-3">3</a>
+      </li>
+      <li>
+        <a class="hds-pagination__nav-item" href="/page-4">4</a>
+      </li>
+      <li>
+        <a class="hds-pagination__nav-item" href="/page-5">5</a>
+      </li>
+      <li>
+        <span class="hds-pagination__nav-item hds-pagination__nav-item-ellipsis">...</span>
+      </li>
+      <li>
+        <a class="hds-pagination__nav-item" href="/page-10">10</a>
+      </li>
+      <li>
+        <a class="hds-pagination__nav-item hds-pagination__nav-item--next" href="/page-?" aria-label="Next page">
+          <span class="hds-pagination__nav-item-label">Next</span><FlightIcon @name="chevron-right" class="hds-pagination__nav-icon" />
+        </a>
+      </li>
+    </ul>
+  </nav>
+
+  <div class="hds-pagination-bar__page-size">
+    <Hds::Form::Select::Field class="hds-pagination-bar__page-size-select" as |F|>
+      <F.Label>Items per page</F.Label>
+      <F.Options>
+        <option value="Kubernetes">10</option>
+        <option value="Other">30</option>
+        <option value="Other">50</option>
+      </F.Options>
+    </Hds::Form::Select::Field>
+  </div>
 </div>
+
+{{!
+  TEMP ROUGH NOTES
+  ------------------
+  WIP Structure:
+    * pagination-bar
+      * total-count
+      * pagination (separate out below children or keep grouped together?)
+        * page-prev
+        * pages
+          * page-item
+        * page-next
+      * page-size
+}}
--- a/packages/components/addon/components/hds/pagination-bar/index.js
+++ b/packages/components/addon/components/hds/pagination-bar/index.js
@@ -13,7 +13,7 @@ export default class HdsPaginationIndexComponent extends Component {
    * @return {string} The "class" attribute to apply to the component.
    */
   get classNames() {
-    let classes = ['hds-pagination-bar'];
+    let classes = ['hds-pagination-bar', 'hds-typography-body-200', 'hds-font-weight-medium'];
 
     // add a class based on the @xxx argument
     // classes.push(`hds-pagination--[variant]-${this.xxx}`);
--- a/packages/components/app/styles/components/pagination-bar.scss
+++ b/packages/components/app/styles/components/pagination-bar.scss
@@ -2,4 +2,111 @@
 // PAGINATIONBAR
 //
 
-// .hds-pagination-bar {}
+@use "../mixins/focus-ring" as *;
+
+.hds-pagination-bar {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  height: 36px;
+  color: var(--token-color-foreground-primary);
+  font-size: 13px;
+}
+
+// TODO: refactor!!! (Should not reference nested component classes)
+.hds-pagination-bar__page-size {
+  > * {
+    display: flex;
+    gap: 12px; // TODO: refactor
+    align-items: center;
+    // outline: 1px dotted blue;
+  }
+
+  // attributes are spread on mested select element, not on the component's parent wrapper
+  .hds-form-label {
+    font-weight: inherit;
+    font-size: inherit;
+  }
+
+  .hds-form-select {
+    height: 28px;
+    margin-top: 0 !important;
+    padding-top: 0;
+    padding-right: calc(0px + 24px);
+    padding-bottom: 0;
+    padding-left: 8px;
+    font-size: inherit;
+    background-position: center right 5px;
+  }
+}
+
+.hds-pagination-bar__pagination {
+  display: flex;
+  align-self: stretch;
+}
+
+.hds-pagination-bar__nav-list {
+  display: flex;
+  margin: 0;
+  padding: 0;
+  list-style-type: none;
+
+  > li {
+    display: flex;
+  }
+}
+
+.hds-pagination__nav-item {
+  @include hds-focus-ring-with-pseudo-element(
+    $top: 5px,
+    $right: 5px,
+    $bottom: 5px,
+    $left: 5px,
+  );
+  display: flex;
+  align-items: center;
+  padding: 0 12px;
+  color: inherit;
+  text-decoration: none;
+}
+
+// NOTE: Trying to account for white space around icons inside the links
+.hds-pagination__nav-item--previous { padding-left: 8px; }
+.hds-pagination__nav-item--next { padding-right: 8px; }
+
+.hds-pagination__nav-icon + .hds-pagination__nav-item-label,
+.hds-pagination__nav-item-label + .hds-pagination__nav-icon {
+  margin-left: 6px;
+}
+
+.hds-pagination__nav-item-ellipsis {
+  color: var(--token-color-foreground-faint);
+}
+
+.hds-pagination__current-page {
+  position: relative;
+  color: var(--token-color-foreground-action);
+  // outline: 1px dotted;
+
+  &:hover {
+    color: var(--token-color-foreground-action-hover);
+  }
+
+  &:active {
+    color: var(--token-color-foreground-action-active);
+  }
+
+  // indicator underline
+  &::before {
+    position: absolute;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    width: 19px;
+    height: 2px;
+    margin: 0 auto;
+    background-color: currentColor;
+    border-radius: 2px;
+    content: "";
+  }
+}
--- a/packages/components/tests/dummy/app/templates/components/pagination-bar.hbs
+++ b/packages/components/tests/dummy/app/templates/components/pagination-bar.hbs
@@ -35,8 +35,16 @@
 <section>
   <h3 class="dummy-h3" id="how-to-use"><a href="#how-to-use" class="dummy-link-section">§</a> How to use</h3>
 
-  {{! ADD YOUR CONTENT HERE }}
-
+  {{! prettier-ignore-start }}
+  {{! template-lint-disable no-unbalanced-curlies }}
+  <CodeBlock
+    @language="markup"
+    @code="
+      <Hds::PaginationBar />
+    "
+  />
+  {{! template-lint-enable no-unbalanced-curlies }}
+  {{! prettier-ignore-end }}
 </section>
 
 <section>
@@ -68,5 +76,4 @@
 
   {{! This below is just an example of invocation, to get started }}
   <Hds::PaginationBar />
-
 </section>


KristinLBradley:
@cveigt This is still a rough work in progress but would appreciate if you could start taking a look at the visual styles I implemented so far: https://hds-components-git-hds-944-pagination-styling-hashicorp.vercel.app/components/pagination-bar#

NOTE: I didn't "gray out" inactive previous and next links yet since no functionality has yet been implemented. I did add focus styles though which you can see if you tab through the pagination links or use browser dev tools to add focus to the links.

packages/components/addon/components/hds/pagination-bar/index.hbs
```diff
@@ -1,4 +1,73 @@
...
+    * pagination-bar
+      * total-count
+      * pagination (separate out below children or keep grouped together in one list?)
+        * page-prev
+        * pages
+          * page
+            * page-link (some content aren't links though...)
+        * page-next
+      * page-size
+}}
```
KristinLBradley, line 73: This is still rough static DOM. Would appreciate feedback on overall structure and class-naming scheme. (Will break out into child components in subsequent steps.)

packages/components/addon/components/hds/pagination-bar/index.hbs
```diff
@@ -1,4 +1,73 @@
...
+
+  <div class="hds-pagination-bar__page-size">
+    <Hds::Form::Select::Field class="hds-pagination-bar__page-size-select" as |F|>
+      <F.Label>Items per page</F.Label>
+      <F.Options>
+        <option value="Kubernetes">10</option>
+        <option value="Other">30</option>
+        <option value="Other">50</option>
+      </F.Options>
+    </Hds::Form::Select::Field>
```
KristinLBradley, line 54: Should I be using this component here? Right now I have rough hacky CSS styles for it to match Figma mocks which I plan to refactor...

packages/components/addon/components/hds/pagination-bar/index.hbs
```diff
@@ -1,4 +1,73 @@
...
+  WIP Structure (some of these will be separated out as sub-components, others just represent HTML/CSS structure):
+    * pagination-bar
+      * total-count
+      * pagination (separate out below children or keep grouped together in one list?)
+        * page-prev
+        * pages
+          * page
+            * page-link (some content aren't links though...)
+        * page-next
+      * page-size
```
KristinLBradley, line 72: I'm not sure about some of the naming and what makes the most sense. Also not sure if I should keep the Previous and Next links in the same list with the number links. I may move them outside it later on...

packages/components/app/styles/components/pagination-bar.scss
```diff
@@ -2,4 +2,127 @@
...
+.hds-pagination-bar {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  height: 36px;
+  color: var(--token-color-foreground-primary);
+  font-size: 13px;
+}
+
+// TODO: refactor!!! (Should not reference nested component classes)
```
KristinLBradley, line 16: These styles are still very rough and are mainly for getting the rough visual styling in place. (Wasn't sure of best way of styling the `Hds::Form::Select::Field` I used for example so I added kind of hacky styles for now.)

packages/components/app/styles/components/pagination-bar.scss
```diff
@@ -2,4 +2,127 @@
...
+}
+
+// Hide "Previous" & "Next" labels by default (TODO: Remove from DOM using JS?)
+.hds-pagination__page-label { display: none; }
+
+// NOTE: Trying to account for white space around icons inside the links (will adjust as requested)
+.hds-pagination__page--previous .hds-pagination__page-link { padding-left: 8px; }
+.hds-pagination__page-next .hds-pagination__page-link { padding-right: 8px; }
+
+// "Compact" Pagination Very Rough WIP - TODO: Refactor
```
KristinLBradley, line 79: Once I implement more of the actual functionality, I should be able to delete some of theses style for the "default" vs. "compact" variants of the component.

packages/components/tests/dummy/app/templates/components/pagination-bar.hbs
```diff
@@ -64,9 +72,9 @@
...
 
-  {{! ADD YOUR CONTENT HERE }}
-
-  {{! This below is just an example of invocation, to get started }}
+  <h4 class="dummy-h4">Default (numbered)</h4>
   <Hds::PaginationBar />
 
+  <h4 class="dummy-h4">Compact</h4>
+  <Hds::PaginationBar class="hds-pagination-bar--is-compact" />
 </section>
```
KristinLBradley, line 80: The docs are all just preliminary. Will clean up, modify, and expand upon them as I continue...

packages/components/addon/components/hds/pagination-bar/index.hbs
```diff
@@ -1,4 +1,73 @@
...
+    <ul class="hds-pagination-bar__page-list">
+      <li class="hds-pagination__page hds-pagination__page--previous">
+        <a class="hds-pagination__page-link" aria-label="Previous page" href="/page-?">
+          <FlightIcon @name="chevron-left" class="hds-pagination__nav-icon" /><span
+            class="hds-pagination__page-label"
+          >Previous</span>
+        </a>
+      </li>
+      <li class="hds-pagination__page hds-pagination__page--current">
+        <a class="hds-pagination__page-link" href="/page-1">1</a>
```
MelSumner, line 15: If this is the current link for the current page, you'd want to add `aria-current="page"` here.

packages/components/addon/components/hds/pagination-bar/index.hbs
```diff
@@ -1,4 +1,73 @@
 {{! ADD YOUR TEMPLATE CONTENT HERE }}
 <div class={{this.classNames}} ...attributes>
-  {{yield}}
-</div>
\ No newline at end of file
+  <div class="hds-pagination-bar__total-count">1–10 of 100</div>
+
+  <nav class="hds-pagination-bar__pagination">
```
MelSumner, line 5: [issue] this will need support for `aria-label` since it won't be the only `nav` element on the page.

packages/components/addon/components/hds/pagination-bar/index.hbs
MelSumner: I think you probably only have to name things if you're going to make them separate components, yeah? 

Thinking it through, you probably wouldn't need to name the `<a>` element- it'll be within an `<li>` -> `<ul>` -> `<nav>` and if you've named all of those, that link will be really tightly scoped.

packages/components/addon/components/hds/pagination-bar/index.hbs
MelSumner: Seems like we should be able to use the component, if you scope the styles to this new parent component it should work out.

packages/components/addon/components/hds/pagination-bar/index.hbs
```diff
@@ -1,4 +1,73 @@
...
+          <FlightIcon @name="chevron-left" class="hds-pagination__nav-icon" /><span
+            class="hds-pagination__page-label"
+          >Previous</span>
+        </a>
+      </li>
+      <li class="hds-pagination__page hds-pagination__page--current">
+        <a class="hds-pagination__page-link" href="/page-1">1</a>
+      </li>
+      <li class="hds-pagination__page">
+        <a class="hds-pagination__page-link" href="/page-2">2</a>
```
MelSumner, line 18: Recommend adding a visually hidden "page" text to the page numbers (or support for custom text). This way users with AT will have context for the link (especially in the ROTA).

```suggestion
        <a class="hds-pagination__page-link" href="/page-2">
          <span class="sr-only">page </span>2</a>
```

packages/components/addon/components/hds/pagination-bar/index.hbs
KristinLBradley: Thanks, my preference is to not add class names to everything for styling like I did here but it seemed like that was the team preference? So thats why I did. Normally I would probably target nested elements by tag such as `<a>` and `<li>` for something like this.

I may break this out into sub-components but probably only 3 at the most for the main three modules within the PaginationBar. (Still considering what would make sense and will work on that in a new PR for the next step.)

packages/components/addon/components/hds/pagination-bar/index.hbs
KristinLBradley: Should I use an anchor link for the current page like I have here now or else remove/replace it with something non-interactive?

packages/components/addon/components/hds/pagination-bar/index.hbs
KristinLBradley: I followed this example: https://www.a11ymatters.com/pattern/pagination/

packages/components/addon/components/hds/pagination-bar/index.hbs
MelSumner: Let's bring it up in the eng sync, because I didn't do classes all the way down for the table component. I think we should use them when it makes sense (and mostly so far it's made sense, but there are a few exceptions).

packages/components/addon/components/hds/pagination-bar/index.hbs
```diff
@@ -1,4 +1,85 @@
...
 <div class={{this.classNames}} ...attributes>
-  {{yield}}
-</div>
\ No newline at end of file
+  <div class="hds-pagination-bar__total-count">1–10 of 100</div>
+
+  <nav class="hds-pagination-bar__pagination" aria-label="Pagination navigation">
+    <ul class="hds-pagination-bar__page-list">
+      <li class="hds-pagination__page hds-pagination__page--previous">
+        <a class="hds-pagination__page-link" aria-label="Previous page" href="/page-?">
```
MelSumner, line 8: Do you know if they use query params for this?

packages/components/addon/components/hds/pagination-bar/index.hbs
KristinLBradley: @alex-ju @didoo do either of you have any input? Thanks!

packages/components/addon/components/hds/pagination-bar/index.hbs
KristinLBradley: The href values are just placeholders at this point.

didoo (commented):
I have added a few comments. 

FYI, as suggested, I haven't looked at the CSS at all. Only at the overall HBS/HTML structure and class namings.

packages/components/addon/components/hds/pagination-bar/index.hbs
```diff
@@ -1,4 +1,85 @@
...
+          <FlightIcon @name="chevron-left" class="hds-pagination__nav-icon" /><span
+            class="hds-pagination__page-label"
+          >Previous</span>
```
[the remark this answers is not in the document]
KristinLBradley: I wasn't sure either if the word "page" was the best choice but wasn't sure what else to use.

packages/components/addon/components/hds/pagination-bar/index.hbs
```diff
@@ -1,4 +1,85 @@
...
+          <span class="sr-only">page </span>4
+        </a>
+      </li>
+      <li class="hds-pagination__page">
+        <a class="hds-pagination__page-link" href="/page-5">
+          <span class="sr-only">page </span>5
+        </a>
+      </li>
+      <li class="hds-pagination__page">
+        <span class="hds-pagination__page-link hds-pagination__page-link--ellipsis">...</span>
```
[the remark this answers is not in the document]
KristinLBradley: Yeah, you're right that it's weird. That was there to match the other link elements. I would like to probably remove some of these extra class names and target some of the styles by element but that didn't seem to be the team's preferred approach. Will think about it more...

packages/components/addon/components/hds/pagination-bar/index.hbs
KristinLBradley: I can see... I was also wondering if there would be a possibility to add a variation to this component for a "compact" version or something but maybe that wouldn't make sense since so far this is just a one-off use case.

packages/components/addon/components/hds/pagination-bar/index.hbs
KristinLBradley: Personally, putting classes on everything makes things seem cluttered, difficult to read, and seems like overkill (just my opinion, no offense). I'm happy to follow whatever the preferred approach is though.

packages/components/addon/components/hds/pagination-bar/index.js
```diff
@@ -13,7 +13,11 @@ export default class HdsPaginationIndexComponent extends Component {
    * @return {string} The "class" attribute to apply to the component.
    */
   get classNames() {
-    let classes = ['hds-pagination-bar'];
+    let classes = [
+      'hds-pagination-bar',
+      'hds-typography-body-200',
+      'hds-font-weight-medium',
```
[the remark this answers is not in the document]
KristinLBradley: Thanks, I planned to probably move them later as I break out the child components and refine the styling.

packages/components/addon/components/hds/pagination-bar/index.hbs
cveigt: The select only has different (decreased) padding and smaller font size. In Figma, it is a separate component to avoid unexpected behaviour from the main component and ensure we the component doesn't have weird resets

packages/components/app/styles/components/pagination-bar.scss
```diff
@@ -2,4 +2,127 @@
...
-// .hds-pagination-bar {}
+@use "../mixins/focus-ring" as *;
+
+.hds-pagination-bar {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  height: 36px;
+  color: var(--token-color-foreground-primary);
+  font-size: 13px;
```
cveigt, line 13:
```suggestion
  font-size: var(--token-typography-body-300-font-size);
```

packages/components/app/styles/components/pagination-bar.scss
```diff
@@ -2,4 +2,127 @@
...
+.hds-pagination-bar__page-list {
+  display: flex;
+  margin: 0;
+  padding: 0;
+  list-style-type: none;
+}
+
+.hds-pagination__page {
+  display: flex;
+}
```
cveigt, line 56: I can see all states for `.hds-pagination__page--current`, but in `.hds-pagination__page` they seem to be missed.

cveigt (commented):
It's looking pretty good overall. I left a few comments.

I noticed that there seems to be something weird with the padding in the page link or the focus ring, but I haven't been able to figure it out.
Screen Shot 2022-11-07 at 4 44 11 PM

packages/components/addon/components/hds/pagination-bar/index.hbs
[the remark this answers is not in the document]
KristinLBradley: I'm going to try removing some of the extra class names I added to see if that helps simplify things. Then if you or others don't like it I can refactor again.

packages/components/addon/components/hds/pagination-bar/index.hbs
KristinLBradley: I went ahead and moved the previous and next links outside the list of page numbers btw. (They aren't quite the same as the numbered links so I don't think they need to be part of the list semantically and also it makes it easier to hide and show the number list for the compact variant of Pagination)

packages/components/addon/components/hds/pagination-bar/index.hbs
KristinLBradley: Is it ok if I look into using Hds::Interactive as the next step? With this PR I'm focusing on the overall structure and preliminary visual styling. 

Once I start breaking things out into sub-components I can focus more on that part.

packages/components/app/styles/components/pagination-bar.scss
KristinLBradley: @cveigt I made some updates to refactor some things. Could you please check again and let me know your feedback? https://hds-components-git-hds-944-pagination-styling-hashicorp.vercel.app/components/pagination-bar#showcase

packages/components/addon/components/hds/pagination-bar/index.hbs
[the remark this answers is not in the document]
KristinLBradley: I refactored it to remove the confusing class name

packages/components/addon/components/hds/pagination-bar/index.hbs
KristinLBradley: If I use `Hds::Form::Select::Base` then it seems I have to create my own html `label` tag for the "Items per page" text to label it and then I have to figure out how best to add an id to associate the label element with the `select` or else I could nest the `Hds::Form::Select::Base` inside the `label` element but don't know if we like doing that... 

Anyway not sure if it's the best solution yet to use that component over `Hds::Form::Select::Field` which lets you specify the label element as part of it and then automatically associates them.

packages/components/app/styles/components/pagination-bar.scss
```diff
@@ -2,4 +2,142 @@
...
+// * Pagination Parent (wrapper) component
+// ----------------------------------------
+.hds-pagination-bar {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  height: 36px;
+  color: var(--token-color-foreground-primary);
+  // font-size: var(--token-typography-body-300-font-size); // = 1rem, not same as 13px, TODO: investigate
+  font-size: 13px;
```
KristinLBradley, line 16: I put back the hard-coded font-size for now because the token wasn't rendering as the correct size. I will clean up and refine the styling as I continue working on this in upcoming PRs @cveigt (This initial PR is meant to focus on overall structure and getting the basic visual look into place vs. the final cleaned up code.)

cveigt: Makes sense. Thanks for clarifying

packages/components/app/styles/components/pagination-bar.scss
```diff
@@ -2,4 +2,142 @@
...
+  position: relative;
+  color: var(--token-color-foreground-action);
+
+  &:hover {
+    color: var(--token-color-foreground-action-hover);
+  }
+
+  &:active {
+    color: var(--token-color-foreground-action-active);
+  }
```
cveigt, line 100: I'm okay if this is sorted out later on, but for some reason, the color doesn't seem to change when the button is clicked/active

packages/components/app/styles/components/pagination-bar.scss
```diff
@@ -2,4 +2,142 @@
...
+    $right: 5px,
+    $bottom: 5px,
+    $left: 5px,
+  );
+  color: inherit;
+  text-decoration: none;
+
+  &:hover {
+    color: var(--token-color-foreground-action-hover);
+  }
```
cveigt, line 56: I think the active state is missing here
`&:active {
    color: var(--token-color-foreground-action-active);
  }`

packages/components/app/styles/components/pagination-bar.scss
KristinLBradley: Sorry about that, I think I may have messed it up when I refactored how the current page class was applied earlier. It should be fixed now

cveigt (approved):
It's coming along nicely 🚀

packages/components/addon/components/hds/pagination-bar/index.hbs
didoo: @KristinLBradley let's figure out this is a follow-up PR (or in the main one). for now let's not consider it a blocker (I personally think we should not mix the two, even if this means duplicating some code, but I am interested to hear also @alex-ju opinion on this)

didoo (approved):
@KristinLBradley I am approving this for now, with the assumption that we will do a final review of the whole code before merging the main feature branch

packages/components/addon/components/hds/pagination-bar/index.hbs
alex-ju: In theory, I'm incline to favor a variant for the select component ('compact' or so, which would mean adding a modifier class). Now, if we don't want to make this variant public (and I can see reasons), I'd fall back on using `Form::Select::Base`

KristinLBradley: Thanks @alex-ju, I would personally favor adding a "Compact" variant for the select component but if the use case is too one off and we don't want it used widely elsewhere then that would be a reason not to.

Using `Hds::Form::Select::Base` vs. `Hds::Form::Select::Field` though means I have to recreate the label element and if we need to add an id to the select to reference it in the label's for attribute then that's a further complication...

PR #679, opened 2022-11-07
[DRAFT] New docs website / Basic test for `Doc::DoDont` component (with yielded markdown)

didoo:
### :pushpin: Summary

Scope of this simple test was to see what happens, with the current implementation, to an Ember component that yields some content in markdown format: how it's interpreted, how it's rendered, what we can do on our side, etc.

I have created a very simple `Doc::DoDont` component that exposes "items" as contextual components that simply yield their content (with some `<div>`s acting as containers/wrappers.

I considered four code samples.

**Normal formatting:**
```
<Doc::DoDont as |dd|>
  <dd.Do>
    Do this:
    - this
    - and this
  </dd.Do>
  <dd.Dont>
    But don't do this:
    - not this
    - and expecially not this
    > this should be a block quote
  </dd.Dont>
</Doc::DoDont>
```
This was rendered in this way:
screenshot_2025
The content is interpreted as inline text. 
Plus, the component is wrapped in a `<p>` element:
screenshot_2026

**Formatted with no whitespace before the content:**
```
<Doc::DoDont as |dd|>
<dd.Do>
Do this:
- this
- and this
</dd.Do>
<dd.Dont>
But don't do this:
- not this
- and expecially not this
> this should be a block quote
</dd.Dont>
</Doc::DoDont>
```
This crashed the app, because the content was (partially) intended as markdown, but the wrapping of the HTML elements was completely off:
screenshot_2023
modified

<doc-do-dont>
  Do this:
  - this
  - and this
</doc-do-dont>

```
<doc-do-dont>
  Do this:
  - this
  - and this
</doc-do-dont>
```
This also crashed the app (for the same reason above).
screenshot_2027

**A simple div:**
```
<div>
Do this:
- this
- and this
</div>
```
This generates a simple `<div>` with no `<p>` wrapper, and the content is interpreted as inline text, not markdown.
screenshot_2029
screenshot_2028

### 🤔 What next?

I think we have to consider our options at this point: I am not so sure we will be able to implement custom ember components that can render their content (as children) in markdown format, unless we pass the markdown code as a string (as argument), which leads to a lot of other problems (can we nest custom components? will we have linting or autosuggestion?).
In theory we could imagine that for some simple components, like this one, we interpret the yielded content and convert it from markdown to HTML, but AFAIK the yielded content is not accessible in the backing class of a component, so... not sure how we can solve this.
Plus, I am not sure how the Ember compilation works in relation with the Markdown to HTML conversion: in which order does it happen? 

I think we should have a brainstorming session, and see what we can do (and maybe have someone else have a go to the problem as well and see what alternatives they can come up with).

Also, I suspect we may have to involve Bryce Kalow for a "consultation" and see from his point of view (and experience) what he thinks of what we're trying to achieve (maybe we're asking too much, looking at the MDX files for DevDot documentation, it seems to me that there are not many custom components used inside the markdown files).

### DIFF
--- a/website-html-to-markdown/moveover/testing/06-test-do-dont/index.js
+++ b/website-html-to-markdown/moveover/testing/06-test-do-dont/index.js
@@ -1,0 +1,8 @@
+import Component from '@glimmer/component';
+
+export default class FileWithFrontmatter extends Component {
+  get test() {
+    console.log(this, this.attributes);
+    return 'this is a test';
+  }
+}
--- a/website-html-to-markdown/moveover/testing/06-test-do-dont/index.md
+++ b/website-html-to-markdown/moveover/testing/06-test-do-dont/index.md
@@ -1,0 +1,51 @@
+# Test "do/dont" component
+
+This is a test using the `Doc::DoDont` component
+
+<!-- THIS DOES NOT WORK AS INTENTED -->
+<Doc::DoDont as |dd|>
+  <dd.Do>
+    Do this:
+    - this
+    - and this
+  </dd.Do>
+  <dd.Dont>
+    But don't do this:
+    - not this
+    - and expecially not this
+    > this should be a block quote
+  </dd.Dont>
+</Doc::DoDont>
+
+<!-- THIS DOES NOT WORK AT ALL (IT CRASHES THE APP) -->
+<!--
+<Doc::DoDont as |dd|>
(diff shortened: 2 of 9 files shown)


didoo:
Closing this one because it was a spike to experiment around some ideas. The actual implementation will happen in a separate PR.

PR #680, opened 2022-11-07
[DRAFT] New docs website / Testing of the `show` template/controller

didoo:
### :pushpin: Summary

This is a test I made to understand if (and how) we could leverage the fact that all the routes generated via markdown files stored in the `docs` folder respond to the same template/controller called `show`.

What I did was to create a template and a controller inside the `website` application (exactly the same as the ones provided by `field-guide`) and see what happened.

### 🕵️‍♀️ Findings

- if the files are added while the web server is running, the pages are rendered correctly and the content in markdown is visible (using the custom template)
  -  notice: adding also a `show` router I was able to generate in output also the frontmatter values
- but if I stop the server an restart it (with `yarn build`) I receive an error related to how (I think) Ember compiles the files
  - let's see what happens in the CI 
screenshot_2030

This for me is a strong indication that we should simply copy the files from `field-guide` inside the `website` app (essentially forking it) and then adapt the application to what we need to do. Otherwise we are investing a large amount of time trying to figure out things, without being able to build something out of our findings.

### :link: External links

JIRA ticket: https://hashicorp.atlassian.net/browse/HDS-875

### DIFF
--- a/website/app/controllers/show.js
+++ b/website/app/controllers/show.js
@@ -1,9 +1,12 @@
 import Controller from '@ember/controller';
-import showdown from 'showdown';
+// import showdown from 'showdown';
 
 export default class ShowController extends Controller {
   get renderedContent() {
-    const converter = new showdown.Converter();
-    return converter.makeHtml(this.model.content);
+    console.log(this.model);
+    // const converter = new showdown.Converter();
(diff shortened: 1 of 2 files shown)


didoo:
Same error in the CI build:
screenshot_2031

PR #685, opened 2022-11-09
Figma design guidelines migration - Part 1

jorytindall:
### :pushpin: Summary

If merged this PR will migrate Figma Design Guidelines for several components to markdown.

- Form::Select
- Toast
- Alert
- Badge
- Breadcrumb
- Button
- Dropdown

These are representative of the different layout and documentation patterns present in our Figma Design Guidelines, though are migrated without any custom elements. Thus, there are some caveats and points that I've noted in the markdown implementation:

- ~~Component examples and visuals of different variants are showcased using only images in lieu of a custom doc component to render Ember components within markdown~~.
- Many (not all) component examples use HDS components within an HTML wrapper. This is a temporary solution and should be simple to update in the future.
- Other documentation focused components (state matrices, banners, etc) are annotated using basic markdown elements
- No _major_ changes to the actual documentation, more or less a copy/paste situation. Therefore, there are some grammar and prose areas that could use improvement.

Since images are used to showcase the component API and features, I've added a `/components` directory within the `public/assets` directory to store images organized in the same fashion as the docs.

Some additional points worth mentioning:

- Content migration (at this point) was relatively trivial since no custom components were being used within markdown. I would estimate 20-25 minutes per component given simple "vanilla" markdown migration.
- I've done my best to call out specific areas to be replaced by documentation components and HDS components in the future given the necessary feature support.
- The typographic hierarchy in the design guidelines doesn't conform 1:1 with markdown. The Figma design guidelines don't map very well to a clear type stack given the lack of layout customization. Hence, the markdown structure is very flat and should be "easy" 😉 to customize going forward.

***

### 👀 How to review

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


jorytindall:
Additional point to note (I'm sure it's on the radar already): there seems to be an issue with the routing of the sidebar in which components that are nested within a category (ie Form controls) are not rendering routes preceded by `/components/`. The routes actually exist, I'm able to manually view `/components/form/checkbox`, but the actual links in the sidebar seem to be broken.

jorytindall:
After deploying, I figured it would be easier going forward to replace the component variants/examples with HDS components in a semantic HTML wrapper, so I replaced _most_ of the images in the `Alert` design guidelines with rendered components for now.

jorytindall:
My image paths aren't working here ☹️ , @didoo could you shed some insight on what the path needs to look like in markdown?

didoo:
> Additional point to note (I'm sure it's on the radar already): there seems to be an issue with the routing of the sidebar in which components that are nested within a category (ie Form controls) are not rendering routes preceded by /components/. The routes actually exist, I'm able to manually view /components/form/checkbox, but the actual links in the sidebar seem to be broken.

Yes, something we have tracked here: https://hashicorp.atlassian.net/browse/HDS-774

> After deploying, I figured it would be easier going forward to replace the component variants/examples with HDS components in a semantic HTML wrapper, so I replaced most of the images in the Alert design guidelines with rendered components for now.
> It's not the cleanest situation in the end, but for representational purpose this seems like the best solution for now.

Something that will need to be discussed (later) in detail: there are significative pros and significative cons in both approaches (using images vs using the "real" components)

> My image paths aren't working here ☹️ , @didoo could you shed some insight on what the path needs to look like in markdown?

Strange! In my local environment they work, but not in the deployed one 🤔
@Dhaulagiri can you have a look? Something strange we're doing?
(even more strange, the SVG of the logo is rendered correctly, and is in the same `public` folder; but maybe I'm missing something)

Dhaulagiri:
I made a commit to temporarily fix the image issue but a longer term fix will require further investigation (i'll create a jira ticket). The short version is that Ember adds a fingerprint to images in production and then rewrites the content of the page to point to those fingerprinted assets. It's unclear but perhaps also unsurprising that it might have a problem doing this within `.md` files.

didoo (approved):
I haven't reviewed all the files (🙂) but in general I think is OK as much as the `yarn generate` doesn't overwrite any file in the `docs` folder. We can always change things in follow-up PRs (surely we'll do).

PR #690, opened 2022-11-10
[WIP] New docs website / Test porting `field-guide`

didoo:
### :pushpin: Summary

In this PR I have started to test if it's possible to port (and fork) `field-guide` as part of the monorepo, so we can customize it to our needs (TBD if and how this work will then be converted in upstream work/contribution to the original project).

In this PR I have done two simple things:
- removed `field-guide` as dependency
- added `@hashicorp/field-guide` as custom "addon" in our repo
- made a quick fix to the imports of files to avoid the application crash (not finding the files) - see https://github.com/hashicorp/design-system/commit/d95163e08d169df7e7fed422e2132f22a3b98e0f
  - notice: later should be converted to relative path; plus maybe these files can be removed (they reference other files, but being already an addon maybe is not necessary)

Result: **it works!!** 🎉

### :link: External links

JIRA ticket:


Dhaulagiri:
> made a quick fix to the imports of files to avoid the application crash (not finding the files)

I think this is the main downside to "forking" the project in this way. What you did is correct and I don't think it would make sense to try to solve it in a different way.

Dhaulagiri:
@didoo what are we doing about tests for field-guide? I don't see the existing tests present, but I would think we'd want to be able to continue to test the addon's behavior as we iterate or add new features.

didoo:
Same approach: since we're going to refactor **a lot** of stuff, I'd prefer to add the test at the end of the process (our current test is: are pages working?)

didoo:
@Dhaulagiri https://hashicorp.atlassian.net/browse/HDS-1022

Dhaulagiri (approved):

PR #692, opened 2022-11-11
New docs website / Add support for tables in markdown in Showdown

didoo:
### :pushpin: Summary

This PR enables the support of markdown tables in our porting of `field-guide` by providing custom settings to `showdown`.

### :link: External links

JIRA ticket: https://hashicorp.atlassian.net/browse/HDS-819

Issue in `field-guide`: https://github.com/empress/field-guide/issues/73

***

### 👀 How to review

### DIFF
--- a/addons/field-guide/addon/controllers/show.js
+++ b/addons/field-guide/addon/controllers/show.js
@@ -6,7 +6,11 @@ export default class ShowController extends Controller {
   fieldGuideConfig = config['field-guide'];
 
   get renderedContent() {
-    const converter = new showdown.Converter();
+    // SET SHOWDOWN SETTINGS HERE:
+    const showdownConfig = {
+      tables: true,
+    };
+    const converter = new showdown.Converter(showdownConfig);
     return converter.makeHtml(this.model.content);
   }
 }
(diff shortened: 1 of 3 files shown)


addons/field-guide/addon/controllers/show.js
```diff
@@ -6,7 +6,11 @@ export default class ShowController extends Controller {
   fieldGuideConfig = config['field-guide'];
 
   get renderedContent() {
-    const converter = new showdown.Converter();
+    // SET SHOWDOWN SETTINGS HERE:
+    const showdownConfig = {
```
Dhaulagiri, line 10: I'm not necessarily opposed to this as it's more direct, but I would suggest that it'd be more reusable and follow Ember conventions to have the config in `config/environment.js` and then access it here similar to how I outlined it in the issue I opened. [ember-cli-showdown docs for reference](https://github.com/empress/ember-cli-showdown#global-showdown-options)

didoo: We're going to have quite a number of configurations for all the tools we're using, and I am not convinced using `config/environment.js` is the right place where to store them, vs co-locating them where they're actually used.
In any case, we can decide later, for now works :)

Dhaulagiri: > I am not convinced

I can see that, but it's worth keeping in mind that this is how configuration is done across all Ember apps so deviating from that convention is going to be surprising to most other contributors to this project.

didoo: @Dhaulagiri https://hashicorp.atlassian.net/browse/HDS-1021

Dhaulagiri (approved):

PR #702, opened 2022-11-14
Table / Use standard mixin for focus ring of `ThSort` button element

didoo:
### :pushpin: Summary

A small PR to remove the custom code and use the standard `focus-ring` mixing to handle focus for the `ThSort` button element

***

### 👀 How to review

### DIFF
--- a/packages/components/app/styles/components/table.scss
+++ b/packages/components/app/styles/components/table.scss
@@ -2,6 +2,9 @@
 // TABLE
 //
 //
+
+@use "../mixins/focus-ring" as *;
+
 $hds-table-border-radius: 6px;
 $hds-table-border-width: 1px;
 $hds-table-inner-border-radius: $hds-table-border-radius - $hds-table-border-width;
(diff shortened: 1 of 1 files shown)


MelSumner:
This seems okay with the high-contrast mode extension: 
CleanShot 2022-11-14 at 15 42 45@2x

didoo:
> This seems okay with the high-contrast mode extension: CleanShot 2022-11-14 at 15 42 45@2x

Yes, this one doesn't require to remove the border to the button (this mixin uses an absolutely positioned pseudo-element, instead of relying on inner/outer box shadows)

MelSumner (approved):
Based on previewing this with the high-contrast extension in Chrome, it seems to work fine- let's take an action item to more thoroughly test our focus states with a Windows machine and proper high-contrast mode when we're able, though.

didoo:
@MelSumner as a bonus, while checking percy that was failing I realized this also fixed an issue we didn't notice (the focus ring was 1px too big).

**before:**
screenshot_2056

**after:**
screenshot_2057

PR #703, opened 2022-11-14
[SPIKE] Table / `Td` component

didoo:
### :pushpin: Summary

This PR is to show what the refactoring of the API of the `Table` component would require, in order to be able to use `Td` as contextual component, similar to what we already do for `Tr`, `Th` and `ThSort`.

In this PR I have:
- added `Td` as sub-component for `Table`
  - added template (no backing class for now, but likely we'll need it)
  - removed typography styling for the `td` element from the `table.scss` file (we're using helper classes directly in the template
- updated the `Table` template so that `Tr` `Td` and `data` are hashed and provided to the `body` named block
- updated the `Table` documentation page to use the new API for `Table`, `:body` and `Td`
- updated integration tests for the `Table` component + added tests for `Td` sub-component

**What these changes entail**:

- we can now use `Td` as contextual component
  - this will open the doors for managing vertical and horizontal alignment in the table cells too, using Ember `@arguments` instead of client-side classes (not something we want to do)
- the way in which the children of the `Table` are declared have been standardized: 
  screenshot_2058
- the "data" of the model is not accessed directly as before, but via a `data` key:
  - before: `<td>{{data.artist}}</td>` → after: `<B.Td>{{B.data.artist}}</B.Td>`
- all tests (integration + visual) passed 🎉

### 👀 How to review

### DIFF
--- a/packages/components/addon/components/hds/table/index.hbs
+++ b/packages/components/addon/components/hds/table/index.hbs
@@ -28,14 +28,14 @@
   <tbody class="hds-table__tbody">
     {{#if this.isSortableTable}}
       {{#each (sort-by this.getSortCriteria @model) as |data|}}
-        {{yield data to="body"}}
+        {{yield (hash Tr=(component "hds/table/tr") Td=(component "hds/table/td") data=data) to="body"}}
       {{/each}}
     {{else if @model}}
-      {{#each @model as |record|}}
-        {{yield record to="body"}}
+      {{#each @model as |data|}}
+        {{yield (hash Tr=(component "hds/table/tr") Td=(component "hds/table/td") data=data) to="body"}}
       {{/each}}
     {{else}}
-      {{yield to="body"}}
+      {{yield (hash Tr=(component "hds/table/tr") Td=(component "hds/table/td")) to="body"}}
     {{/if}}
   </tbody>
 </table>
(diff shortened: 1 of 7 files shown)


didoo:
@MelSumner I did a quick spike to see what changes would be required to have a `Td` sub-component, similar to the other ones we are already using (`Tr`, `Th`, `ThSort`) and the results are not bad at all! Have a look and let me know what you think. Importantly: this also opens the door to (more) easily control the horizontal/vertical alignment of the table cells, using ad-hoc arguments instead of client-side class names.

@alex-ju have a look too, so we can have another point of view (maybe you have other/better ideas)

MelSumner (approved):
Seems like it works; I have no objection if this is what you think is best.

didoo:
> Seems like it works; I have no objection if this is what you think is best.

@MelSumner maybe a way to decide if this is the right approach could be do a quick spike/experiment and try to introduce the alignment for the table cells (`Th` / `ThSort` / `Td`). You could do it using a "sortable table" as litmus test, and compare the results of using the main feature branch (`melsumner/table`) vs this branch (`table-spike-td-component`) and see the difference: what makes it easier (or doesn't make the code too complex/convoluted). I think this would be a good indicator of the direction to take.

(if you want, we can do another pairing session together and we can test together)

alex-ju (approved):
Looks good – makes the component API/notation look more consistent too

MelSumner:
So I guess if I'm thinking about it some more: 

1. Pro: I was concerned about terseness and you kept it terse.
2. Pro: it looks like everything else (well I guess this could be a pro or con, depending on one's point of view)
3. Con: I am not convinced we need it. Especially in table components, where it is likely that lots of rows will need to be rendered, keeping things as lean as possible ends up being rather important. By making the table cell a component we've just given the table component another thing that needs to compile. I would ask if the justification is worth the tradeoff.

PR #716, opened 2022-11-18
[Table] Updates and refactors to remove blockers

MelSumner:
### :pushpin: Summary

If merged, this PR addresses the blockers for shipping the table component ([HDS-980](https://hashicorp.atlassian.net/browse/HDS-980))

I think this is a balanced approach for what we want, so if folks like this PR then I'll add tests so we can ship it. 

- refactors the way sortable columns were implemented to remove deprecation warning ([HDS-1087](https://hashicorp.atlassian.net/browse/HDS-1087))
- added `th-sort` to the `:head` block yield, so users can do a custom implementation if they wish. For awareness: this is not aligned with the design intent. ([HDS-1094](https://hashicorp.atlassian.net/browse/HDS-1094))
- added support for text-alignment and width in `th-sort` ([HDS-1095](https://hashicorp.atlassian.net/browse/HDS-1095))
- added support for text-alignment and width in `th` and `td` ([HDS-1096](https://hashicorp.atlassian.net/browse/HDS-1096))
- added documentation for setting column with and text alignment  (in both API and Usage sections) ([HDS-1088](https://hashicorp.atlassian.net/browse/HDS-1088))
- added extra showcase examples ([HDS-1089](https://hashicorp.atlassian.net/browse/HDS-1089))

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/addon/components/hds/table/td.hbs
```diff
@@ -1,3 +1,6 @@
-<td class="hds-table__td hds-typography-body-200 hds-font-weight-regular" ...attributes>
+<td
+  class="hds-table__td hds-typography-body-200 hds-font-weight-regular {{if @alignRight 'hds-table__td--text-right'}}"
```
MelSumner, line 2: I know some won't like a conditional here; I tried to pass this in the hash (feel like I should have been able to) but it didn't work. I'll try again on Monday if it really bothers folks too much.

packages/components/app/styles/components/table.scss
```diff
@@ -22,6 +22,50 @@ $hds-table-cell-padding-tall: 20px 16px;
   border-spacing: 0;
 }
 
+// TABLE COLGROUP/COL
+.hds-table__col--width-auto {
```
MelSumner, line 26: @jorytindall do you have design guidance on what column width values we should accept? For now, I went with percentages in multiples of ten (mostly because the mental math is quick to do). 

One thing to note is that since Figma doesn't support relative units, things tend to come out in fixed pixel width instead. We'd expect folks to convert this to relative units in code, though (for a11y reasons). Maybe we could work out some guidance for designers about this, what do you think?

Dhaulagiri: If I recall, the approach we had discussed for this was to allow consumers to pass in css of their choosing similar to how we do this for several other components using the `{{style}}` helper behind the scenes. Here's the docs from one of the form components:

Screen Shot 2022-11-21 at 09 16 06

MelSumner:
we're going to try something else, so I'm going to close this PR for now.

PR #714, opened 2022-11-18
New docs website / Implement "Foundations" and "Components" landing pages (with automatic list generation)

didoo:
### :pushpin: Summary

This PR is to start to have the "Foundations" and "Components" landing pages look more like the current design (still not final)

Previews:
- [Foundations](https://hds-website-git-new-docs-website-components-la-e008ee-hashicorp.vercel.app/foundations/)
- [Components](https://hds-website-git-new-docs-website-components-la-e008ee-hashicorp.vercel.app/components/)
  - Notice: the links to the `form` components don't work because of a bug in `field-guide` (see [HDS-774](https://hashicorp.atlassian.net/browse/HDS-774))

### :hammer_and_wrench: Detailed description

See commits.

### :camera_flash: Screenshots

**Current designs** (⚠️ STILL IN PROGRESS ⚠️):

screenshot_2084

### :link: External links

JIRA ticket: https://hashicorp.atlassian.net/browse/HDS-1102

***

### 👀 How to review

👉 Review commit-by-commit or by files changed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #718, opened 2022-11-18
New docs website / Implement "Home" landing page

didoo:
### :pushpin: Summary

Simple follow-up #714, this time for the homepage.

Preview: https://hds-website-git-new-docs-website-homepage-cards-hashicorp.vercel.app/

See commits.

### :link: External links

JIRA ticket: https://hashicorp.atlassian.net/browse/HDS-1110

***

### 👀 How to review

### DIFF
--- a/website/app/components/doc/page/hero.hbs
+++ b/website/app/components/doc/page/hero.hbs
@@ -1,7 +1,10 @@
 <div class="doc-page-hero" ...attributes>
-  {{#if (has-block)}}
-    <h1 class="doc-text-h1">{{yield}}</h1>
-  {{else}}
-    <h1 class="doc-text-h1">This is an hero block</h1>
-  {{/if}}
+  <div class="doc-page-hero__inner-wrapper">
+    {{#if @title}}
+      <h1 class="doc-page-hero__title doc-text-h1">{{@title}}</h1>
+    {{/if}}
+    {{#if @description}}
+      <p class="doc-page-hero__description doc-text-body-large">{{@description}}</p>
+    {{/if}}
+  </div>
 </div>
--- a/website/app/components/doc/page/hero.hbs
+++ b/website/app/components/doc/page/hero.hbs
@@ -4,7 +4,7 @@
       <h1 class="doc-page-hero__title doc-text-h1">{{@title}}</h1>
     {{/if}}
     {{#if @description}}
-      <p class="doc-page-hero__description doc-text-body-large">{{@description}}</p>
+      <p class="doc-page-hero__description doc-text-body">{{@description}}</p>
     {{/if}}
   </div>
 </div>
--- a/website/app/styles/pages/application/hero.scss
+++ b/website/app/styles/pages/application/hero.scss
@@ -3,8 +3,23 @@
 .doc-page-hero {
   display: flex;
   align-items: center;
-  justify-content: center;
-  padding: 120px;
+  justify-content: flex-start;
+  padding: 210px 120px;
+}
+
+.doc-page-hero__inner-wrapper {
+  display: flex;
+  flex-direction: column;
   color: #fff;
-  background-color: #727374;
+}
+
+.doc-page-hero__title {
+  margin: 0;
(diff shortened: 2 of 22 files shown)


website/app/styles/pages/application/hero.scss
```diff
@@ -3,8 +3,23 @@
 .doc-page-hero {
   display: flex;
   align-items: center;
-  justify-content: center;
-  padding: 120px;
+  justify-content: flex-start;
+  padding: 210px 120px;
```
heatherlarsen, line 7:
```suggestion
  padding: 160px 120px;
```

Now that I've seen it in the browser, I think it'd be a good idea to try to surface the cards a bit higher on the page, so looking to reduce the padding in the hero.

website/app/styles/pages/home.scss
```diff
@@ -4,7 +4,14 @@
   grid-area: sidebar / sidebar / content / content;
 }
 
+.doc-page-home__hero {
+  padding-bottom: 80px;
```
heatherlarsen, line 8: I'm not following how this math worked out (maybe we can pair on this next week?), but something is a bit off with the height of the cut-out in the hero...it's a little too tall and should only overlap the bottom of the hero by 52px.

website/app/components/doc/page/hero.hbs
```diff
@@ -1,7 +1,10 @@
...
-    <h1 class="doc-text-h1">{{yield}}</h1>
-  {{else}}
-    <h1 class="doc-text-h1">This is an hero block</h1>
-  {{/if}}
+  <div class="doc-page-hero__inner-wrapper">
+    {{#if @title}}
+      <h1 class="doc-page-hero__title doc-text-h1">{{@title}}</h1>
+    {{/if}}
+    {{#if @description}}
+      <p class="doc-page-hero__description doc-text-body-large">{{@description}}</p>
```
heatherlarsen, line 7:
```suggestion
      <p class="doc-page-hero__description doc-text-body">{{@description}}</p>
```

I must've pulled in the wrong style, but I checked some of the marketing files again and it looks like we want the standard body size, not the large body size. Apologies for the oversight; I've updated the figma wires already.

heatherlarsen (changes requested):
This is a great start 🙂 Left a few comments in the code and had another more general question:

Can we add a max-width to the content area in the hero to 1312px (`1216px + (2 * var(--doc-page-stage-gutter-medium))`) and center-align that block like the rest of the content? Essentially mimicking [hashicorp.com](https://www.hashicorp.com/).

Is it possible for the cards to fill the space evenly if, say, there are only 2 or 3 cards in the row? It looks like they don't adapt in width when cards are removed.

Nothing is blocking at this stage. If you want to merge and adjust these later, I'm cool with that too.

website/app/styles/pages/home.scss
didoo: it's a super-preliminary layout. I am pretty sure will change anyway, it was mainly to mimic (more or less) what you had in Figma, so definitely not pixel perfect (not by a mile)

didoo:
> Can we add a max-width to the content area in the hero to 1312px (`1216px + (2 * var(--doc-page-stage-gutter-medium))`) and center-align that block like the rest of the content? Essentially mimicking [hashicorp.com](https://www.hashicorp.com/).

The math here is slightly more complex (I think) because I need to add also the internal padding of the cards container, so... let's keep it like this and fine-tune when we have the final designs, ok?

> Is it possible for the cards to fill the space evenly if, say, there are only 2 or 3 cards in the row? It looks like they don't adapt in width when cards are removed.

To be discussed together (next week?). If there are 3 cards, and the viewport is small, what happens to them? they squeeze indefinitely? Probably not? So we need to define different layouts for different conditions (2 cards, 3 cards, 4 cards, etc).
Maybe in a later phase?

heatherlarsen:
Sounds good, we can flush out at a later date 🙂

heatherlarsen (approved):
🚀

PR #724, opened 2022-11-23
New docs website: wcag component

MelSumner:
### :pushpin: Summary

If merged, this PR adds a wcag component to the docs website so all relevant success criteria and related information doesn't have to be manually included each time. 

In this initial approach, I tried to mostly emulate what was already there. If we don't like this approach, there are some other things I can try.

As an example of a different approach: I use frontmatter in a11y-automation.dev (which is similar to this website in that the content is all in markdown that is rendered to templates). Here's an example: https://github.com/MelSumner/a11y-automation/blob/041f8f9d9c0ba20b413bcc7f8fa649d623616208/violations/alt-color-diffs.md?plain=1#L3 

Then the tag contents can be rendered in the template: https://github.com/MelSumner/a11y-automation/blob/041f8f9d9c0ba20b413bcc7f8fa649d623616208/app/templates/violations/violation.hbs#L35

There are probably other ways to do it too, so let me know what direction is preferred for this project. 

### 👀 How to review

👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


website/docs/components/alert/05--accessibility.md
```diff
@@ -11,9 +11,11 @@ This component is conditionally conformant. That is, it is conformant when there
 
 This section is for reference only. This component intends to conform to the following WCAG success criteria:
 
-<dummy-wcag-success-criteria-list data-list="1.3.1|1.3.2|1.4.1|1.4.3|1.4.10|1.4.11|1.4.12|2.1.1|2.1.2|2.2.1|2.5.3|4.1.1|4.1.2|4.1.3">Placeholder for the WCAG Success Criteria List component - Don't delete!</dummy-wcag-success-criteria-list>
+<Doc::WcagList @criteriaList={{array 'wcag-1-3-1' 'wcag-1-3-2' 'wcag-1-4-1' 'wcag-1-4-3' 'wcag-1-4-10' 'wcag-1-4-11' 'wcag-1-4-12' 'wcag-2-1-1' 'wcag-2-1-2' 'wcag-2-2-1' 'wcag-2-5-3' 'wcag-4-1-1' 'wcag-4-1-2' 'wcag-4-1-3' }} />
```
didoo, line 14: [suggestion] to make things simpler, why not just pass the values as `1-3-1` and prepend the `wcag-` prefix inside the component logic/code? this would make it shorter, more readable.

website/app/components/doc/wcag-list.hbs
```diff
@@ -0,0 +1,14 @@
+<div class="doc-wcag-list">
+  <ul class="doc-wcag-list__ul">
+    {{#each this.criteria as |criterion|}}
+      {{#if (and @criteriaList (includes criterion.id @criteriaList))}}
```
didoo, line 4: suggestion: since you already have the backing class, why not move all the logic inside of it (so the filtering is done there, and here you loop only on the actual criteria that need to be rendered)

didoo (approved):

website/docs/components/alert/05--accessibility.md
MelSumner: I agree; since we're not using any of the failures or techniques in our documentation, I could simplify the data. 👍

website/docs/components/alert/05--accessibility.md
```diff
@@ -5,15 +5,21 @@ component: alert
...
+It is fully conformant when there are no interactive elements present inside of the alert. 
+
+There is future work planned to improve this component's WCAG conformance by adding support for the correct ARIA roles when interactive elements are contained within the alert.
 
 #### Applicable WCAG Success Criteria (Reference)
 
-This section is for reference only. This component intends to conform to the following WCAG success criteria:
+This section is for education and reference. This component intends to conform to the following WCAG success criteria:
+
+<Doc::WcagList @criteriaList={{array '1.3.1' '1.3.2' '1.4.1' '1.4.3' '1.4.10' '1.4.11' '1.4.12' '2.1.1' '2.1.2' '2.2.1' '2.5.3' '4.1.1' '4.1.2' '4.1.3' }} />
```
MelSumner, line 18: @alex-ju @KristinLBradley what do you think about this? Would this be relatively easy for you to use? Is there something else that would make it easier for you?

website/app/components/doc/wcag-list.hbs
MelSumner: okay. I did it this way because it was faster for me, & I wanted to get a prototype out there for our eng team to discuss. If the overall approach is acceptable, then I will work on it some more.

website/docs/components/alert/05--accessibility.md
KristinLBradley: This seems pretty straight forward to use for me.

KristinLBradley (approved):

MelSumner:
@didoo I updated this component so that it is like the other components from a structure perspective; I'm going to merge it for now and make more updates in a follow-up PR so we can start using it.

PR #725, opened 2022-11-23
New docs website / Add possibility to control rendering of "cover" and "sidecar" elements via "frontmatter"

didoo:
### :pushpin: Summary

In this PR I have used the recently added support for "frontmatter" attributes in markdown, to control the hiding of the "cover" and "sidecar" elements.

In this PR I have:
- added possibility to remove “cover” and “sidecar” rendering for “docs” pages via “frontmatter”
- added a couple of “testing” examples

Preview:
- [no sidecar](https://hds-website-git-page-layout-frontmatter-hashicorp.vercel.app/testing/05-test-frontmatter/no-sidecar/)
- [no cover + no sidecar](https://hds-website-git-page-layout-frontmatter-hashicorp.vercel.app/testing/05-test-frontmatter/no-cover-no-sidecar/)

@heatherlarsen @jorytindall let's use this sparingly plz :)

***

### 👀 How to review

### DIFF
--- a/website-html-to-markdown/moveover/testing/05-test-frontmatter/no-cover-no-sidecar.md
+++ b/website-html-to-markdown/moveover/testing/05-test-frontmatter/no-cover-no-sidecar.md
@@ -1,0 +1,18 @@
+---
+order: 1
+title: Page with no sidecar and no cover
+category: testing
+layout:
+  cover: false
+  sidecar: false
+---
+
+# Page with no sidecar and no cover
+
+Using "frontmatter" properties it's possible to remove the generation of the "cover" and "sidecar" elements in a "doc" page:
+
+```
+layout:
+  cover: false
+  sidecar: false
+```
--- a/addons/field-guide/addon/routes/show.js
+++ b/addons/field-guide/addon/routes/show.js
@@ -36,6 +36,7 @@ export default class ShowRoute extends Route {
           'group',
           'component',
           'section',
+          'layout',
           'title',
           'description',
           'caption',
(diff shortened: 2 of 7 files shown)


website-html-to-markdown/moveover/testing/05-test-frontmatter/no-cover-no-sidecar.md
```diff
@@ -0,0 +1,18 @@
+---
+order: 1
+title: Page with no sidecar and no cover
+category: testing
+layout:
+  cover: false
```
jorytindall, line 6: Only question I have here is whether the property should be have the same name in the frontmatter. Probably doesn't matter, but thinking of bulk find/replace/search

didoo: @jorytindall can you expand on this? are you suggesting to change this to `hasCover` or change `hasCover` to `cover`?

(the thinking behind the different naming is in the context of code, it's more common to use `has` or `is` as prefix for "logical/binary" variables, while for the frontmatter I favoured simplicity for the editor (easy to remember, and more or less all the frontmatter attributes for now have very simple (and memorable) names.

jorytindall: Yes, that's what I was referring to, making the frontmatter property consistent with the code property (`hasCover` and `hasSidecar`). For context we mirror that property naming in Figma. Though it also makes sense for simplicity in the frontmatter.

Not necessarily advocating for one over the other, more just raising out of curiousity.

didoo: Keep in mind that there will be some (rare) conditions in which, even if the `sidecar` is not explicitly set to `false` (so it's should be `true` by default, in reality the sidecar will be hidden anyway (eg. no headings of a certain level in the page); the `hasSidecar` encompasses more than just the frontmatter value.

alex-ju (approved):

PR #726, opened 2022-11-23
New docs website / Component API refactor

alex-ju:
### :pushpin: Summary

Component API refactor to extract the properties from the current `<dl>` structure and translate it into a custom component syntax.

This PR explores a refactoring of the Component API/properties section as follows:
 - uses a codemod to extract the current property-related information and convert `<dl>` to a custom, intermediary `<api>`/`<output>` structure
 - ~~uses `turndownService` in conjunction with `TurndownPluginGfm` to convert `<table>`s to markdown~~
 - uses `turndownService` to the intermediary `<api>`/`<output>` structure to custom a custom `<Doc::ComponentApi>` component and markdown
 - adds the `refactor-properties` script to the `yarn generate` command (to test the experiment e2e)

Notes:
 - ~~it currently uses only markdown (no custom components) although a mix of these or even only custom Ember components are potential options~~
 - ~~doesn't contain styling (sorry for the bad contrast on some rows)~~ (thanks @didoo for getting this sorted!)

Learnings:
 - although not explicit, `turndown` doesn't convert any XML-style tags to markdown, only well-known HTML tags, meaning a strict list of tags and attributes are passed through the markdown convertor engine; this is the reason we went with the intermediary step (consisting of `<api>` and `<output>`) instead of converting straight to Handlebars syntax
 - even within custom tags (such as `<Doc::ComponentApi>` or `<Doc::ComponentApi::Property>`) `showdown` still adds `<p>` elements based on new lines, resulting in interposed tags (e.g. `<p><Doc::ComponentApi></p></Doc::ComponentApi>`); this is the reason we're currently outputting the Handlebars tags inline, instead of nesting them for better visibility

Follow-up work to address the following issues:
 - deep levels of nesting in the DOM tree causes `showdown` to not render some of the nodes – resulting in a `¨C#C` notation
 - extend `showdown` renderer to manage paragraph rendering more tightly and allow multiline Handlebars syntax

#### Current

Screenshot 2022-11-23 at 16 40 57

#### New

Screenshot 2022-11-24 at 11 26 02

### :link: External links

[Jira issue](https://hashicorp.atlassian.net/browse/HDS-1056)

***

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ ] +1 Percy if applicable


MelSumner:
@alex-ju is this just about the moving of the documentation vs the style? The table presents as a "giant-wall-o-text" and seems like it's more difficult to read. Will that be addressed in a separate PR?

alex-ju:
> @alex-ju is this just about the moving of the documentation vs the presentation? The table presents as a "giant-wall-o-text" and seems like it's more difficult to read. Will that be addressed in a separate PR?

The intent is to extract the current API and abstract it away (in our case via arguments to an ember component), then use that component to render it to something else. The [early design explorations](https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=179%3A18810&t=icyNLCC5zQ8FeQUF-1) are presenting these properties in a table – which is why I'm using a table at the moment; but chatting with design, they are not convinced it's going to look like this in the end. My current thinking is that having it abstracted to an ember component should allow us to render it however we'd like in the end.

MelSumner:
@alex-ju that makes sense. We could also _probably_ eventually generate these from the comments in the code, yeah? Anyway, thank you for clarifying!

website/app/components/doc/component-api/index.hbs
```diff
@@ -0,0 +1,13 @@
+<table class="doc-markdown-table doc-component-api" ...attributes>
```
didoo, line 1: since we have not yet defined what the actual output is (a table? a list? other?) it's OK to use `doc-markdown-***` classes here, but later we have to avoid using them for styling custom components, so we can separate what is actual markdown (generated from the markdown to HTML converter) from what is a custom component (for which we have total control).

website/app/styles/markdown/tables.scss
```diff
@@ -8,7 +8,6 @@ $doc-markdown-table-border-color: rgba(101, 106, 118, 20%);
   width: 100%;
   max-width: 100%;
   overflow: auto;
-  table-layout: fixed;
```
didoo, line 11: 👍

website/app/styles/markdown/tables.scss
```diff
@@ -33,4 +32,21 @@ $doc-markdown-table-border-color: rgba(101, 106, 118, 20%);
   .doc-markdown-tr:last-child & {
     border-bottom: none;
   }
+
+  code {
```
didoo, line 36: [issue] these styles work for our specific use case, but may impact other cases when these HTML elements are used inside a generic markdown table. probably better to create a specific CSS file for the `doc-component-api` component, and use the class `.doc-component-api` to target these (even if they are temporary).

website/docs/components/alert/02--component-api.md
```diff
@@ -7,12 +7,12 @@ section: component-api
...
+<Doc::ComponentApi as |C|><C.Property @name="type" @required="true" @type="enum" @value="page inline compact">Sets the type of alert.</C.Property><C.Property @name="color" @type="enum" @value="neutral highlight success warning critical" @default="neutral">Sets the color scheme for `background`, `border`, `title`, and `description`, which **cannot** be overridden. `color` results in a default `icon`, which **can** be overridden.</C.Property><C.Property @name="icon" @type="string | false" @value="–">Override the default `icon` name, which is determined by the `color` argument. [Flight](https://flight-hashicorp.vercel.app/) icon name or pass `false` for no icon.</C.Property><C.Property @name="onDismiss" @type="function" @value="–">The alert can be dismissed by the user. When a function is passed, the "dismiss" button is displayed.</C.Property><C.Property @name="...attributes" @type="–" @value="–">`...attributes` spreading is supported on this component.</C.Property></Doc::ComponentApi>
```
didoo, line 10: My understanding is that for how `turndown` works they need to be inline, not multiline/nested right?
Not in this PR but in the follow-up, if you can't find a way around this, maybe you could consider a post-processing script that using a simple regex takes this text block, and then pass it to a prettier function* and let it do the job for you. just an idea

\* _prettier can be imported as module; I'm sure you already know it, but if you don't... see for example how it's done [here](https://github.com/hashicorp/design-system/blob/main/packages/flight-icons/scripts/build-parts/getCssForIconAnimation.ts) or [here](https://github.com/hashicorp/design-system/blob/main/packages/flight-icons/scripts/build-parts/generateBundleSVGReact.ts)_

didoo (commented):
Added a couple of comments (one of them is an issue, probably easy to fox directly in this PR)

website/app/styles/markdown/tables.scss
alex-ju: leftover from when playing with markdown table – will remove

didoo (approved):
👍

PR #740, opened 2022-12-01
Add `indeterminate` style to `Checkbox` component

alex-ju:
### :pushpin: Summary

Add `indeterminate` style to `Checkbox` component

After wrangling with the logic to reflect the state of an indeterminate checkbox input I realized that we already had support for such a feature, without our knowledge, so we only needed to add the styles and document it.

What Ember does is that it automagically uses the `indeterminate` attribute as a property (I suspect this is determined by Glimmer). So even though the HTML standard doesn't support the `indeterminate` state to be set via an `attribute`, it is possible in Ember. To avoid introducing unnecessary complexity in our codebase and our consumer's codebase, I propose we rely on this feature. I added a test to make sure this behaviour doesn't regress in future Ember releases.

### :link: External links

[Jira issue](https://hashicorp.atlassian.net/browse/HDS-1145)

***

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ ] +1 Percy if applicable


.changeset/itchy-games-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
```
MelSumner, line 2: Question: should there be two change sets?

alex-ju: Not sure of the etiquette here – I added [one for the tokens](https://github.com/hashicorp/design-system/pull/740/files#diff-e2e962f0d1fde94e1cd08f5c87a7229e92dab2895b894ca96fd3c6729014c602) and [one for the component](https://github.com/hashicorp/design-system/pull/740/files#diff-1a80ffc289fe128548c9d95fc57c51422a9dc2038dc4421d7ea231030ec91d04) because I wanted distinct (and hopefully more relevant) messages, but happy to merge them if that makes more sense.

MelSumner: I am not entirely sure either. @Dhaulagiri can you clarify?

Dhaulagiri: I think the way @alex-ju has it here makes sense. In practice I think either approach works more or less the same, this just gives a little more control over the changelog message like Alex mentioned.

MelSumner: Thank you both for answering my question! 👍

MelSumner (approved):
:shipit:

cveigt (approved):
It looks great, Alex. Thank you!

I realized that we have a few tokens named `--token-form-control-checked-...`. We should've probably chosen a less specific name than `checked` back when we build the form components 😣

didoo:
> I realized that we have a few tokens named `--token-form-control-checked-...`. We should've probably chosen a less specific name than `checked` back when we build the form components 😣

@cveigt can you expand on why you think this is an issue?

alex-ju:
> Super neat! 👌
> 
> Maybe one small (nit) thing I would do, is to have an indeterminate state in one of the examples where we use the "field" version. In this way we make sure that the `indeterminate` works also in that variant (I am sure it works, but it's nice to have it covered as use case too)

Good call – added `indeterminate` to our `Checkbox::Group` examples and rebased to main.

> > I realized that we have a few tokens named --token-form-control-checked-.... We should've probably chosen a less specific name than checked back when we build the form components 😣
>
> @cveigt can you expand on why you think this is an issue?

I think it's the fact that we share some `checked` tokens between the `checked` and `indeterminate` states (for the blue background and border) – ideally we would've abstracted those to something more generic; or I can duplicate them with 'indeterminate' instead of 'checked'

didoo:
> I think it's the fact that we share some `checked` tokens between the `checked` and `indeterminate` states (for the blue background and border) – ideally we would've abstracted those to something more generic; or I can duplicate them with 'indeterminate' instead of 'checked'

Ahhh! I get it. Yes,you're right, they should have a different (more generic) name. But I see they're already used in other codebases (https://cs.github.com/?scope=org%3Ahashicorp&scopeName=hashicorp&q=%22--token-form-control-checked%22) so renaming them would be a breaking change. Probably a not super urgent problem to solve, so I would say let's close our eyes and merge it :)

PR #746, opened 2022-12-02
New docs website / Added custom colors for markdown list "markers"

didoo:
### :pushpin: Summary

Follow-up of this thread: https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1?node-id=964:51241#319421092

In this PR I have:
- added custom colors for markdown list markers

_Notice: due to inconsistent browser support for the `::marker` pseudo element, is not possible to control the spacing between the marker and the actual content that follows it (we can hack things to match the designs, but I would prefer not go down that path)._

Preview: 

### :link: External links

***

### 👀 How to review


didoo:
Adding a comment here, then I'll delete it. @heatherlarsen see how the inline `code` element is rendered within different content blocks:

# This is an h1 and this is a `code block`
## This is an h2 and this is a `code block`
### This is an h3 and this is a `code block`
#### This is an h4 and this is a `code block`
##### This is an h5 and this is a `code block`
###### This is an h6 and this is a `code block`

This is a paragraph and this is a `code block`

- This is a list item and this is a `code block`

heatherlarsen:
@didoo I adjusted the marker color in [Figma](https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=1305%3A85529&t=YiLYEvZcSepjMN0v-4) because they were too light. I've updated them to `Gray 3` - `#727374`. All good on the spacing front...I've updated the Figma file to match as close as I could. 

Is it possible to update the font of the ordered list markers to use `Body`? It looks like they're inheriting the font from the content, which is feeling a little large in comparison to the unordered markers. 

The spacing of the nested items is a little off, but I don't have spacing for them yet. Is this something we can update with a follow-up ticket?

didoo:
> @didoo I adjusted the marker color in [Figma](https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=1305%3A85529&t=YiLYEvZcSepjMN0v-4) because they were too light. I've updated them to `Gray 3` - `#727374`. All good on the spacing front...I've updated the Figma file to match as close as I could.

OK, updated also the code, using the newly introduced color tokens.

> Is it possible to update the font of the ordered list markers to use `Body`? It looks like they're inheriting the font from the content, which is feeling a little large in comparison to the unordered markers.

It may seem, but is actually using the same font.

modified

> The spacing of the nested items is a little off, but I don't have spacing for them yet. Is this something we can update with a follow-up ticket?

Yes.

heatherlarsen (approved):
🚀

PR #751, opened 2022-12-02
[docs][wcag-component] moves filtering logic from template to backing class

MelSumner:
### :link: External links

***

### 👀 How to review

👉 Review commit-by-commit
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


website/app/components/doc/wcag-list/index.hbs
```diff
@@ -1,7 +1,7 @@
 <div class={{this.classNames}}>
   <ul class="doc-wcag-list__ul">
-    {{#each this.criteria as |criterion|}}
-      {{#if (and @criteriaList (includes criterion.number @criteriaList))}}
+    {{#each this.filteredCriteria as |criterion|}}
+      {{#if (@criteriaList)}}
```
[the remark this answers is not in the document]
MelSumner: @alex-ju it's really just a catch- I need to remove all of it, really.

MelSumner:
FWIW, the error I'm seeing is `maximum nesting of 10 spans reached!!!` so I wonder if it's coming from something else in the website, since this component doesn't use span elements. 🤔

website/app/components/doc/wcag-list/index.js
```diff
@@ -1,810 +1,817 @@
...
+        number: '5.3.2',
+        url: 'https://www.w3.org/TR/WCAG21/#conformance-optional',
+        description: '',
+      },
+    ];
+    let criteriaList = this.args.criteriaList;
+    let filteredCriteria = criteria.filter((criterion) =>
+      criteriaList.includes(criterion.number)
+    );
+    return filteredCriteria;
```
didoo, line 813: possible improvements:
- you don't need to have a `criteriaList` variable, if you use it only once in the code, you can use directly the `this.args.criteriaList` array (which is already in memory)
- (if you want) you don't need to create a temporary `filteredCriteria`, you can directly return the result of the `filter` method on the original array (it already creates a new array) 
```suggestion
    return criteria.filter((criterion) =>
      this.args.criteriaList.includes(criterion.number)
    );
```

didoo (approved):
left a couple of suggestions, if you want

website/app/components/doc/wcag-list/index.js
MelSumner: @didoo I was actually kind of confused about the conditions under which it's okay to filter the original array, I'll DM you.

alex-ju (approved):

PR #752, opened 2022-12-05
New docs website / Revamp navigation / Page `tabs` and `sidecar` MVP

didoo:
### :pushpin: Summary

This is a **very MVP** implementation of the `tabs` and `sidecar` functionalities for the docs pages. 
The performance of the page rendering is clearly not great, but at least it can be used in the meantime by the designers to have an idea of how the final pages will look like and work, and later we can revisit this approach with a clearer perspective of what is actually needed (at the moment there are a lot of gray areas, around the actual content).

In this PR I have:
- introduced a function that, on `insert` of the content in the `show` page controller/template, parses the DOM tree, looking for specific "sections" blocks, and then  using the ID of the page as "name" for  the section, builds a list of objects ([special Ember arrays](https://api.emberjs.com/ember/release/classes/MutableArray)) passed down in the `show` template to the `tabs` and `sidecar` components, so they can use it to dynamically build themselves at runtime.
- applied some basic styling to the `tabs` and `sidecar` elements (they will be updated later, according to the final design specs)

Preview: https://hds-website-git-new-docs-website-revamp-naviga-10b0cc-hashicorp.vercel.app/components/alert/

_Notice: I have (on purpose) avoided to squash the commits, just in case in the future we need to get back to the service-based approach for any reasons, and we can use some of the work done here as starting point._

### :link: External links

Jira tickets:
- https://hashicorp.atlassian.net/browse/HDS-941
- https://hashicorp.atlassian.net/browse/HDS-940

***

### 👀 How to review

👉 Review by files changed


Dhaulagiri (approved):

Dhaulagiri:
I didn't do a close review of the CSS or tab code itself, but the overall approach here is 👍 for where we are at

PR #754, opened 2022-12-05
Table: small tweaks based on engineer feedback

MelSumner:
### :pushpin: Summary

If merged, this PR makes a few small changes to the table component.

- removed `table-layout: fixed` from the table styles by default. Added an `isFixedLayout` option and updated docs. Further decisions on this will be tracked by HDS-1183.
- reduced options for `valign` to `top` and `middle` and updated docs
- The `isStriped` option now set to `false` by default, updated docs and showcase example
-  Updated the documentation to make it more clear that click events on table elements should not be used

### 👀 How to review
👉 Review by files changed

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/addon/components/hds/table/index.js
```diff
@@ -5,7 +5,7 @@ import { assert } from '@ember/debug';
 
 const DENSITIES = ['short', 'medium', 'tall'];
 const DEFAULT_DENSITY = 'medium';
-const VALIGNMENTS = ['top', 'middle', 'bottom', 'baseline', 'sub', 'text-top'];
+const VALIGNMENTS = ['top', 'middle'];
```
KristinLBradley, line 8: Question: Would it be helpful to include "bottom" as well? ('baseline', 'sub', and 'text-top' seem like they would be much less commonly used but "bottom" seems to have more potential use cases although I don't know if that's true within HashiCorp across various products.)

packages/components/tests/dummy/app/templates/components/table.hbs
```diff
@@ -482,12 +488,13 @@ export default class ComponentsTableRoute extends Route {
...
-      <li>The table row element (<code class="dummy-code">tr</code>) is not eligible to receive interactions. If an
-        interactive element is desired, place it within a table cell element (<code class="dummy-code">td</code>).</li>
+      <li>The table row element (<code class="dummy-code">tr</code>) is not eligible to receive interactions (e.g.,
+        cannot have an onClick event handler attached directly to it). If an interactive element is desired, place it
+        within a table cell element (<code class="dummy-code">td</code>).</li>
       <li>When providing additional or alternative styles to the table element, do not change the
         <code class="dummy-code">display</code>
         property in the CSS. This alters how the table is presented to the user with assistive technology, and they will
-        no longer be presented with a table.
+        no longer be presented with a table, as the table element will have lost its implicit role.
```
KristinLBradley, line 497: Question: Would it be helpful to include a reference link here? To me, it seems non-intuitive that altering a visual display property would affect semantics so it seems like a reference could help HDS users to better understand.

I also wondered if we could possibly add aria roles to the various table elements in case the display property were changed? (Or would it be more harmful to include redundant aria roles for tables using the default "table" display?) I found this article which recommends adding aria roles if you change the display for example: https://www.accessibility-developer-guide.com/examples/tables/layout-changes/

packages/components/addon/components/hds/table/index.js
MelSumner: @KristinLBradley great question! "Top" will only really only apply to situations where there is more than one line of content, otherwise it's vertically aligned to the middle (as a browser default).  We added this option because the `hds-button` component is _just_ tall enough to make the cells look a little "off" if the rest of the row only has a single line of content in each cell. 

Then I thought, "oh I wonder if I should add all the acceptable values" but then it was confusing for devs trying to use the component. So I figured I would take out all but the two I absolutely am sure they need, and figured we could add more back in if we had to. WDYT?

packages/components/tests/dummy/app/templates/components/table.hbs
MelSumner: Hmm, maybe- I have brought it up before but the idea was discarded. Perhaps it's time to try again? 

I could also revert this specific change- maybe I made it more complicated by trying to clarify. I don't _want_ developers reaching for aria-roles if they don't have to.

WDYT?

packages/components/addon/components/hds/table/index.js
KristinLBradley: That makes sense. Thanks for explaining.

packages/components/addon/components/hds/table/index.js
```diff
@@ -20,11 +20,21 @@ export default class HdsTableIndexComponent extends Component {
...
+   * @default false
+   * @description Determines whether the table rows should have alternating background colors; defaults to false.
    */
   get isStriped() {
-    return this.args.isStriped ?? true;
+    return this.args.isStriped ?? false;
+  }
+
+  /**
+   * @param isFixed
```
didoo, line 31: For consideration: should we make it more explicit what it refers to and call it `isFixedLayout`?

packages/components/addon/components/hds/table/index.js
```diff
@@ -50,7 +60,7 @@ export default class HdsTableIndexComponent extends Component {
    * @param valign
    * @type {string}
    * @default 'top'
-   * @description Determines the vertical alignment of the table cells; options are all of the values accepted by the CSS "vertical-align" property: "top", "middle", "bottom", "baseline", "sub", and "text-top". If no valign is defined, "top" is used.
+   * @description Determines the vertical alignment of the table cells; options are: "top", "middle". If no valign is defined, "top" is used.
```
didoo, line 63: One consideration. If we define the vertical alignment at table level, consumer may not be able to achieve some complex layouts that _may_ require different alignments for different cells in a row.
screenshot_2125
Of course it's an edge case, but still. By moving instead the `valign` property to `Td` level (with the default we want) we allow consumers to achieve a more granular alignment. It may be more verbose for the consumers, but for me it's a more straightforward API.

packages/components/addon/components/hds/table/index.js
```diff
@@ -78,6 +88,11 @@ export default class HdsTableIndexComponent extends Component {
       classes.push('hds-table--striped');
     }
 
+    // add a class based on the @isFixed argument
+    if (this.isFixed) {
+      classes.push('hds-table--fixed');
```
didoo, line 93: suggestion: maybe make it more explicit: `hds-table--layout-fixed`

packages/components/tests/dummy/app/templates/components/table.hbs
```diff
@@ -121,8 +123,15 @@
...
-      <p>Default: <span class="default">true</span></p>
-      <p>If set to <code class="dummy-code">false</code>, zebra striping on the table will not be applied.</p>
+      <p>Default: <span class="default">false</span></p>
+      <p>If set to <code class="dummy-code">true</code>, zebra striping on the table will be applied.</p>
+    </dd>
+    <dt>isFixed <code>boolean</code></dt>
+    <dd>
+      <p>Default: <span class="default">false</span></p>
+      <p>If set to
+        <code class="dummy-code">true</code>, table-display will be set to fixed, which will automatically distribute
```
didoo, line 133:
```suggestion
        <code class="dummy-code">true</code>, the table layout will be set to <code class="dummy-code">fixed</code>, which will automatically distribute
```

packages/components/tests/dummy/app/templates/components/table.hbs
```diff
@@ -136,16 +145,13 @@
     </dd>
     <dt>valign <code>enum</code></dt>
     <dd>
-      <p>If set, determines the vertical alignment of table's cell (td) content. While the acceptable values contain all
-        of the values that the CSS property accepts, the default (top) and middle are the values most likely to be used.</p>
+      <p>If set, determines the vertical alignment of table's cell (td) content. While it is set to top by default,
+        content will be middle aligned vertically unless there are two or more lines of contentin any cell in a given
```
didoo, line 149:
```suggestion
        content will be middle aligned vertically unless there are two or more lines of content in any cell in a given
```

didoo: I am not sure this statement is clear, for a reader. 
- if by default the alignment is top, how can be _middle aligned vertically_?
- this may be true for text-only content, but we have to consider every possible content (eg. an image/logo or a custom content/component)

I suggest to remove entirely the explanation (`While it is ... in a give row`), to avoid confusion.

packages/components/tests/dummy/app/templates/components/table.hbs
```diff
@@ -482,12 +488,13 @@ export default class ComponentsTableRoute extends Route {
   </p>
   <p class="dummy-paragraph">There are a few critical items for developers to note:
     <ul>
-      <li>The table row element (<code class="dummy-code">tr</code>) is not eligible to receive interactions. If an
-        interactive element is desired, place it within a table cell element (<code class="dummy-code">td</code>).</li>
+      <li>The table row element (<code class="dummy-code">tr</code>) is not eligible to receive interactions (e.g.,
+        cannot have an onClick event handler attached directly to it). If an interactive element is desired, place it
+        within a table cell element (<code class="dummy-code">td</code>).</li>
```
didoo, line 493:
```suggestion
        within the table cell element (<code class="dummy-code">td</code>).</li>
```

I fear someone would interpret this as the need to do something like this:
```
<td>
  <td>
    <button onclick>
  </td>
<td>
```

packages/components/tests/dummy/app/templates/components/table.hbs
```diff
@@ -715,9 +722,9 @@ export default class ComponentsTableRoute extends Route {
     </:body>
   </Hds::Table>
   <h4 class="dummy-h4">
-    Static table with no striping
+    Static table with striping
   </h4>
-  <Hds::Table @model={{this.model.data}} @isStriped={{false}} @caption="Static table with no striping">
+  <Hds::Table @model={{this.model.data}} @isStriped={{true}} @caption="Static table with no striping">
```
didoo, line 727:
```suggestion
  <Hds::Table @model={{this.model.data}} @isStriped={{true}} @caption="Static table with striping">
```

didoo (approved):
Added a few comments, nothing blocking per se. But I would strongly consider moving the `valign` argument at `Td` level (instead of `Table` level).

packages/components/addon/components/hds/table/index.js
MelSumner: So for this one, Jory and I have already figured out that this shouldn't be a table (no table should have this kind of complexity to it). We are going to offer to pair with that team to help them translate this into a different component/element.

However, if teams do want to still use the table components and achieve complex layouts, they do that now by putting `div` elements inside of the table cell, and using that as their styling hook.

packages/components/tests/dummy/app/templates/components/table.hbs
MelSumner: Ok, I've edited the text for now, hopefully reduces potential confusion.

PR #760, opened 2022-12-06
[Table] column refactor

MelSumner:
### :pushpin: Summary

If merged, this PR would refactor how table columns are invoked. 

This is a result of pairing with HDS Engineer Ambassadors and resulting feedback (added as reviewers as FYI)

- table columns are now invoked with an array hash
- `isSortable`, `width` and `align` are all invocable on the column hash;  `width` and `align` can also be added to a TH and TD component (if not using `@columns`)
- documentation code samples updated to reflect change
- tests updated
- added `th` with `scope="row"` to the `:body` named block (to support accessibility)

New invocation examples: 
CleanShot 2022-12-12 at 10 03 03

Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/addon/components/hds/table/td.hbs
```diff
@@ -1,3 +1,6 @@
-<td class="hds-table__td hds-typography-body-200 hds-font-weight-regular" ...attributes>
+<td
+  class="hds-table__td hds-typography-body-200 hds-font-weight-regular{{if @align (concat ' hds-text-' @align)}}"
```
didoo, line 2: [issue] let's try not to do string concatenation and logic in the templates, we tend to do it these operations in a backing class

packages/components/app/styles/@hashicorp/design-system-components.scss
```diff
@@ -43,3 +43,17 @@
   clip: rect(1px, 1px, 1px, 1px) !important; /* 1 */
   clip-path: inset(50%) !important;  /* 2 */
 }
+
+// text alignment helpers - only use these if you absolutely must.
+
+.hds-text-left {
```
didoo, line 49: [issue] these global classes are not ok. we should have "local" classes, specific only for the table

packages/components/addon/components/hds/table/td.hbs
didoo: also, we should move the `valign` argument here (in the Td` element, instead of the `Table`) which is more specific to where it's used

Reason: this would also allow consumers to have a more granular control of the vertical alignment of cells (what if for any reason two td cells need to have different vertical alignment? with the `valign` applied to the `table` it would not be possible, and would require a breaking change to our component API to achieve it, with the suggested change it's already possible without any extra change or refactoring)

didoo (commented):
Left a couple of comments so you can maybe address them before our meeting today.

packages/components/addon/components/hds/table/index.hbs
```diff
@@ -4,30 +4,34 @@
   {{else if @caption}}
     <caption class="sr-only">{{@caption}}</caption>
   {{/if}}
+  <colgroup>
+    {{#if @columns}}
+      {{#each @columns as |column|}}
+        {{#if column.width}}
+          {{!template-lint-disable no-shadowed-elements}}
+          <col width={{column.width}} />
```
didoo, line 12: [issue/blocker?] one thing I forgot to mention: with this approach, is not possible to apply a width in case of a table with no model defined (any table for which we don't provide a `columns` object).

didoo (commented):
Forgot an important issue

packages/components/addon/components/hds/table/index.hbs
MelSumner: we changed it so you always have to provide a `columns` array hash. This brings other challenges but I scheduled a pairing session for us to discuss.

packages/components/addon/components/hds/table/td.hbs
MelSumner: This would be supporting bad design by default. Tables should not be vertically aligned by individual cell but rather at the table level. This has been the table best practice for years.

didoo: "Guidance over control".

packages/components/app/styles/@hashicorp/design-system-components.scss
MelSumner: I removed these.

didoo (approved):
@MelSumner OK for me to merge, once the documentation has been updated to reflect the latest changes and maybe some tests added to make sure the `@width` and `@scope` arguments are correctly handled.

PR #767, opened 2022-12-07
Move markdown generation to in-repo addon

Dhaulagiri:
### :pushpin: Summary

This PR migrates the logic remaining in the field-guide addon into an "in-repo" addon within the website itself.

The main benefits to this is that it more directly houses the logic for the website in one place and allows us to remove a large number of confusing cruft from the field-guide transfer into this repo.

In my view, the main drawback is that it creates a separation between the logic (in the in-repo addon) and the dependencies which are now mixed with all the other dependencies for the project. I think it's a minor one, but it's worth calling out.

### 👀 How to review

👉 Review commit-by-commit


website/lib/markdown/table-of-contents.js
```diff
@@ -21,17 +21,6 @@ const _ = require('lodash');
 //   'testing',
 // ];
 
-function getStructuredPageTree(pageList) {
```
Dhaulagiri, line 24: AFIACT this wasn't being used anywhere

addons/field-guide/index.js
```diff
@@ -1,118 +0,0 @@
...
-      processedDocsJsonFilesTree,
-      'docs'
-    );
-
-    return new MergeTrees([processedDocsJsonFilesTree, processedTocFiles]);
-  },
-
-  // TODO! not sure how to test if this logic below works
-  // https://github.com/ef4/prember#using-prember-from-an-addon
-  // urlsForPrember(distDir) {
```
Dhaulagiri, line 111: I did just say earlier today that I was leaving this behind but now I'm a bit 🤷‍♂️  about it. I think we'll have to rewrite this to get it to work if/when we get back to prember.

website/lib/markdown/index.js
```diff
@@ -0,0 +1,63 @@
...
+const resolve = require('resolve');
+
+const MarkdownProcessIncludeDirectives = require('./markdown-process-includes');
+const MarkdownToJson = require('./markdown-to-jsonapi');
+const TableOfContents = require('./table-of-contents');
+const MergeTrees = require('broccoli-merge-trees');
+const Funnel = require('broccoli-funnel');
+
+module.exports = {
+  name: 'markdown',
```
Dhaulagiri, line 15: I didn't spend much time thinking of what to call this so if there is a better suggestion I'm happy to change it

didoo: in any case we can change it later, so I would not worry too much about it

website/lib/markdown/table-of-contents.js
didoo: Yes, a leftover from some experiments I did. Since you're here, do you mind also deleting the commented code right above, too?

didoo (approved):
you-are-my-hero-you-my-hero

Thank you ♥️

addons/field-guide/index.js
didoo: My 2 cents: this one leave it ;)

didoo:
@Dhaulagiri I am going to merge, since it's approved, to avoid large conflicts with other PRs I am working on.

PR #768, opened 2022-12-07
Fixed missing escape for curly brackets in backticked code block

didoo:
### :pushpin: Summary

Per conversation with @Dhaulagiri, this small PR fixes a bug in the `Form/Toggle` documentation page.

essentially, expressions like this in markdown
```
`{{on "change" [do something]}}`
```
need to have the first curly brace escaped, or otherwise it's interpreted by Ember (rightly so)
```
`\{{on "change" [do something]}}`
```


Dhaulagiri (approved):
good catch!

PR #773, opened 2022-12-08
New docs website / Finalize `Page:Cover` element (component)

didoo:
### :pushpin: Summary

This PR builds on top of another PR (will open soon) that refactors how the "frontmatter" attributes (and which ones) are provided to the `show` template/controller. This has allowed me to finalize the `Page::Cover` component (page _"cover"_ element) in terms visual design (according to the [latest design specs](https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=1345%3A79255&t=YNJcoWCVScYCh1p5-1)).

In this PR I have:
- added global styling for generic links (added both a mixin and a CSS class, will be used also in other places)
- finalized the `cover` (with responsiveness and correct spacing) including `tabs` (colors, scrolling)

👉 👉 👉 **Preview**: https://hds-website-git-new-docs-website-page-cover-extra-hashicorp.vercel.app/components/alert/ (Notice: only the `Alert` doc page has the proper "frontmatter" attributes).

@heatherlarsen can you check and see if it adheres to the current specifications?
@KristinLBradley @alex-ju can you check and see if you see something off in terms of code?

screenshot_2150

### :link: External links

JIRA ticket: https://hashicorp.atlassian.net/browse/HDS-1202

***

### 👀 How to review

### DIFF
--- a/website/app/styles/pages/application/cover.scss
+++ b/website/app/styles/pages/application/cover.scss
@@ -1,22 +1,30 @@
 // COVER
 
 @use "../../breakpoints" as breakpoint;
+@use "../../typography/mixins" as *;
 
 .doc-page-cover {
   position: relative;
   grid-area: cover;
-  // TODO check with heather what she wants to do with the vertical padding (responsive)
-  padding: 16px var(--doc-page-stage-gutter-small);
+  min-width: 0; // This is needed to allow the scrolling of the tabs
+  padding: 16px var(--doc-page-stage-gutter-small) 0;
   background-color: var(--doc-color-white);
   border-bottom: 1px solid var(--doc-color-gray-500);
 
   @include breakpoint.medium () {
-    padding: 32px var(--doc-page-stage-gutter-medium);
+    padding: 64px var(--doc-page-stage-gutter-medium) 68px;
   }
 
   @include breakpoint.large () {
-    padding: 64px var(--doc-page-stage-gutter-large);
+    padding: 64px var(--doc-page-stage-gutter-large) 68px;
   }
+}
+
+.doc-page-cover__title-extra-wrapper {
+  display: flex;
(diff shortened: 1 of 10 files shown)


website/app/styles/pages/application/cover.scss
```diff
@@ -1,31 +1,69 @@
...
+  display: flex;
+  flex-wrap: nowrap;
+  height: min-content;
+  margin-top: 24px;
+  overflow-x: auto;
+  overflow-y: visible; // needed or it will show a small vertical scrollbar
+  -webkit-overflow-scrolling: touch;
+
+  @include breakpoint.medium-and-above () {
+    position: absolute;
```
KristinLBradley, line 64: I usually try to avoid using absolute positioning for something like this. It seems like maybe you could remove the bottom padding of .doc-page-cover instead (plus maybe adjusting the margin)?

alex-ju (approved):

website/app/styles/pages/application/cover.scss
didoo: @KristinLBradley I though about it, and even started to make the changes, but then realized that since some pages have tabs and others don't, then I have to do different calculations and pass around props to account the different cases, so I decided to roll back the changes. we can revisit this later, when we will have more stable content and we can see what are the emerging patterns.

heatherlarsen:
@didoo This looks great! I saw 2 small issues when testing responsiveness. 

1. When the viewport width is smaller than 768px, the cover's `margin-top` should be `32px`; it's currently `16px`, which feels a bit too tight. 
2. A vertical scrollbar appears when the tabs overflow and need to scroll horizontally. The vertical scrollbar shouldn't be visible.
Screen Shot 2022-12-12 at 1 09 09 PM

PR #777, opened 2022-12-08
Re-enable fastboot + prember

Dhaulagiri:
### :pushpin: Summary

This PR brings back fastboot & prember in a way that works

Long story short is that this _does_ actually work because this combo generates the static files as part of the build process and serves them via vercel (or locally).

To test it locally, one can do `yarn start:prember` and you'll interact with the prember "server" as documented [here](https://github.com/ef4/prember#using-prember-in-development).

One can verify this in production by disabling javascript.

### :link: External links

* https://github.com/ef4/prember

### 👀 How to review

👉 Review commit-by-commit

Reviewer's checklist:

- [ ] +1 Percy if applicable


website/lib/markdown/index.js
```diff
@@ -60,4 +62,12 @@ module.exports = {
 
     return new MergeTrees([processedDocsJsonFilesTree, processedTocFiles]);
   },
+  urlsForPrember(distDir) {
+    const flatPageListJson = fs.readJsonSync(`${distDir}/toc.json`);
+    // TODO is there a way to have this list generated automatically (or exported) from the routes (`website/app/router.js`)?
```
Dhaulagiri, line 67: I think this is doable but I did not try it as the router also contains routes we wouldn't want to crawl in this way so the fixed list seemed like an easier bet for now

didoo (approved):
fantastic

PR #780, opened 2022-12-08
implement query params

Dhaulagiri:
### :pushpin: Summary

This PR creates query params to handle "routing" for:
* tabs on "show" pages
* Icon search and size

* [ ] Why do we lose `#` when reloading a query param'ed route (but we scroll to the right content)

### DIFF
--- a/website/app/controllers/show.js
+++ b/website/app/controllers/show.js
@@ -45,7 +45,7 @@ export default class ShowController extends Controller {
 
     let tab = this.tabs.find((el) => {
       // trim trailing slashes from query param
-      return el.label === this.selectedTab.replace(/\/$/, '');
+      return el.label.toLowerCase() === this.selectedTab.replace(/\/$/, '');
     });
     return tab ? tab.index : 0;
   }
@@ -151,7 +151,7 @@ export default class ShowController extends Controller {
       set(toc, 'isCurrent', toc.index === current);
     });
 
-    set(this, 'selectedTab', this.tabs[current].label);
+    set(this, 'selectedTab', this.tabs[current].label.toLowerCase());
 
     // leave for debugging
     // console.log('show setCurrent', this.sections, this.tabs, this.tocs);
--- a/website/app/controllers/show.js
+++ b/website/app/controllers/show.js
@@ -24,11 +24,29 @@ const getTOCs = (container) => {
 };
 
 export default class ShowController extends Controller {
+  queryParams = [
+    {
+      selectedTab: 'tab',
+    },
+  ];
+
   @service fastboot;
 
   @tracked sections = A([]);
   @tracked tabs = A([]);
   @tracked tocs = A([]);
+
+  get selectedTabIndex() {
+    if (!this.selectedTab) {
+      return 0;
+    }
+
+    let tab = this.tabs.find((el) => {
+      // trim trailing slashes from query param
+      return el.label === this.selectedTab.replace(/\/$/, '');
+    });
+    return tab ? tab.index : 0;
+  }
 
   get title() {
     return this.model.frontmatter?.title ?? '';
@@ -101,8 +119,7 @@ export default class ShowController extends Controller {
     this.sections.setObjects(sections);
     this.tabs.setObjects(tabs);
     this.tocs.setObjects(tocs);
-    // TODO handle when the page loads which one is the current, based on the URL query params
-    this.setCurrent(0);
+    this.setCurrent(this.selectedTabIndex);
     // leave for debugging
     // console.log('show didInsert', this.sections, this.tabs, this.tocs);
   };
@@ -131,6 +148,9 @@ export default class ShowController extends Controller {
     this.tocs.forEach((toc) => {
       set(toc, 'isCurrent', toc.index === current);
     });
+
+    set(this, 'selectedTab', this.tabs[current].label);
+
     // leave for debugging
     // console.log('show setCurrent', this.sections, this.tabs, this.tocs);
   }
--- a/website/app/controllers/show.js
+++ b/website/app/controllers/show.js
@@ -28,6 +28,8 @@ export default class ShowController extends Controller {
     {
       selectedTab: 'tab',
     },
+    'searchQuery',
+    'selectedIconSize',
   ];
 
   @service fastboot;
--- a/website/app/controllers/show.js
+++ b/website/app/controllers/show.js
@@ -44,6 +44,7 @@ export default class ShowController extends Controller {
     }
 
     let tab = this.tabs.find((el) => {
+      // for consistency we always compare the query param tab to a lowercase version of the tab label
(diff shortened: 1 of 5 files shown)


website/app/controllers/show.js
```diff
@@ -7,10 +7,27 @@ import { A } from '@ember/array';
 import { showdownConfig } from '../shared/showdown-config';
 
 export default class ShowController extends Controller {
+  queryParams = [
+    {
+      selectedTab: 'tab',
+    },
+  ];
+
+  selectedTab = 'guidelines';
```
Dhaulagiri, line 16: TODO: this isn't the right default and speaks to maybe turning this into an integer index value

website/app/controllers/show.js
```diff
@@ -7,10 +7,27 @@ import { A } from '@ember/array';
...
+
+  // selectedTab = 'code';
```
Dhaulagiri, line 18: ✂️

website/app/controllers/show.js
Dhaulagiri: Because we don't know the tabs until after the page is rendered we can't change this based on which page we're on. for now I'll delete this which means every tab will have a query param, even the default.

Dhaulagiri:
I think this is ready for some initial review to get some 👀 on what we have here so far

website/app/controllers/show.js
```diff
@@ -9,12 +9,32 @@ import { inject as service } from '@ember/service';
...
   @tracked tocs = A([]);
 
+  get selectedTabIndex() {
+    if (!this.selectedTab) {
+      return 0;
+    }
+
+    let tab = this.tabs.find((el) => {
+      // trim trailing slashes from query param
+      return el.label.toLowerCase() === this.selectedTab.replace(/\/$/, '');
```
Dhaulagiri, line 33: this is so it's `tab=library` vs `tab=Library` (because that's how i had all the tests written). happy to change this and update the tests if keeping things lowercase in the URL seems of low value

didoo: the button inside the tab has this HTML in the DOM:
```
<button id="tab-guidelines" class="doc-text-navigation" role="tab" aria-controls="guidelines" aria-selected="" type="button">Guidelines</button>
```

maybe we can use the `id` of the tab, so we can use it then to identify the element more easily? (I know there would be a duplication, `tab=tab-guidelines` but maybe is not that of a problem?)

website/app/controllers/show.js
```diff
@@ -116,6 +135,9 @@ export default class ShowController extends Controller {
     this.tocs.forEach((toc) => {
       set(toc, 'isCurrent', toc.index === current);
     });
+
+    set(this, 'selectedTab', this.tabs[current].label.toLowerCase());
```
didoo, line 139: all these string manipulations scattered around the code, to be used as identifiers, make me nervous. better to use an ID or an ad-hoc `data-` attribute

website/docs/foundations/icons/index.js
```diff
@@ -25,14 +21,18 @@ export default class Index extends Component {
   });
 
   get filteredIcons() {
-    if (this.searchQuery) {
+    // this.router.currentRoute.queryParams['searchQuery'];
+    const searchQuery = this.router.currentRoute.queryParams['searchQuery'];
```
didoo, line 25: I would add a comment explaining that these queryParams are to be found also in the `show.js` file, why they're are there, and recommend to keep the names in sync

website/app/controllers/show.js
```diff
@@ -9,12 +9,32 @@ import { inject as service } from '@ember/service';
...
   @tracked tabs = A([]);
   @tracked tocs = A([]);
 
+  get selectedTabIndex() {
+    if (!this.selectedTab) {
+      return 0;
+    }
+
+    let tab = this.tabs.find((el) => {
+      // trim trailing slashes from query param
```
didoo, line 32: I see in the icons controller you're using this
```
this.router.currentRoute.queryParams['searchQuery']
```
which returns directly the query param without the trailing slash (I imagine), could something similar also be used in this context?

didoo (approved):
I'm approving it (so you can merge if you want) because it's definitely an improvement but there are still a couple of things to figure out
- the use of the "label" for identification, with string manipulation, makes me nervous, I would prefer to have a more solid approach
- the "back" in the browser doesn't work: I see the URL change, but the selected tab remains the same 😬

website/app/controllers/show.js
Dhaulagiri: that's what I initially used but having `tab-` as part of the query param felt off to me. in practice behind the scenes it's not any easier or harder, so sticking with the label seems slightly cleaner to me.

website/app/controllers/show.js
```diff
@@ -9,12 +9,32 @@ import { inject as service } from '@ember/service';
 import { showdownConfig } from '../shared/showdown-config';
 
 export default class ShowController extends Controller {
+  queryParams = [
```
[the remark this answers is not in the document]
Dhaulagiri: Good question and it highlights a big challenge with only have one route effectively for so many different pages. Thinking out loud: _probably_ we could re-use the same name for search behavior on other pages and it would be ok. In the event we did have to "break" things the impact is low in my view as the page will still load, someone might just need to re-do their search.

Dhaulagiri:
> the "back" in the browser doesn't work:

😬  I see that too. I'll spend a short amount of time seeing if I can figure out how to get that to re-trigger the tab logic as it does not currently refresh the route in any way which is how it works otherwise.

website/app/controllers/show.js
Dhaulagiri: it _is_ limited to the two instances in this file and has tests around it all, and it would help us return the correct tab regardless of how a consumer happens to capitalize the query param (unlikely, but possible). I do see your point though, so I'll see if there is a way to  address this more consistently and/or at least make comments explicitly noting this.

website/app/controllers/show.js
Dhaulagiri: it turns out the trailing slash was from previous work I had done and it is no longer present, so I just removed that part

website/docs/foundations/icons/index.js
Dhaulagiri: added a comment

website/app/controllers/show.js
Dhaulagiri: added a comment for the two instances of this

Dhaulagiri:
@didoo I addressed your comments. I'm going to open an issue for the back button bug. It's a small issue but I could see it coming up for folks.

didoo:
> @didoo I addressed your comments. I'm going to open an issue for the back button bug. It's a small issue but I could see it coming up for folks.

@Dhaulagiri OK feel free to merge then

PR #790, opened 2022-12-09
New docs website / Finalize "inline code" styles

didoo:
### :pushpin: Summary

In this PR I have updated the styling for the inline `code` elements, to follow the design specs.

In this PR I have:
- finalized styling for `inline code` element
- updated the code snippets markdown testing page to have more inline cases

_Notice: it may be that we will need to tweak the CSS selector for when this style need to be applied. For now works in most of the case, so we'll collect cases in which it needs fixes and then we'll update it accordingly (if needed)._ 

**In code**:
screenshot_2162

**In design** ([Figma file](https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=1364%3A123139&t=SG6x8piu4ObfVyAR-1)):
screenshot_2163

@heatherlarsen it was not possible to have a pixel-perfect representation of your designs, this is the best that could be achieved while leveraging the browser defaults.

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-1215
Figma file: https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=1364%3A123139&t=SG6x8piu4ObfVyAR-1


website/app/styles/markdown/other-elements.scss
```diff
@@ -12,19 +14,25 @@
...
-// TODO do we need this or in reality we will not use them directly?
-
-.doc-markdown-code { // for the typographic styles see "typography"
-  margin: 0;
-  padding: 4px 8px;
-  color: white; // TODO
-  background-color: black; // TODO
-  border-radius: 3px;
+
+// TBD if we need a more specific selector to target only the "inline code" elements and not overlap with the other types of code snippets
```
alex-ju, line 18: Hopefully, we won't need it 🤞

alex-ju (approved):
The code looks good – and will come in handy for the Component API work (will rebase after this gets in) as the description of a component property uses plain markdown and often contains code blocks.

didoo:
@heatherlarsen since today you're off, I am merging this PR. if you see we need to do any tweaking we can do it in a follow-up PR

heatherlarsen:
I think there's too much padding on the H1 example, but it's not worth splitting hairs over since it's an unlikely use case. If we find we're using it later, we can revisit then.

PR #797, opened 2022-12-09
Revert "Re-enable fastboot + prember"

Dhaulagiri:
Reverts hashicorp/design-system#777

Strike 2 😄 . This doesn't work correctly in production because our tabs logic cannot be pre-rendered. Seems better to back this out and bring back later after more investigation.


Dhaulagiri:
Interestingly enough I see the same problems in this branch with this all turned off 🤔

Dhaulagiri:
But I don't see the same behavior locally...

Dhaulagiri:
This was because of https://github.com/hashicorp/design-system/pull/779

PR #798, opened 2022-12-09
refactor tab parsing logic to run every time model changes

Dhaulagiri:
### :pushpin: Summary

refactor tab parsing logic to run every time model changes so that navigation routing shows correct component pages

A side effect of https://github.com/hashicorp/design-system/pull/779 is that because we are routing via Ember, when you're on a component page and use the navigation to route to another a full transition does not occur and the `did-insert` helper we rely on to parse the tabs doesn't actually run past the initial page load.

This PR refactors the logic to schedule this for after the `renderedContent` function does it's thing which is triggered on each transition because the underlying model is changing (which is what we want).

~~The main drawback is that this is slightly slower on initial render leading to the content jumping a bit once the tabs logic happens, but this does happen with the previous method as well.~~ I think it's the same actually, but please test me on this.

### DIFF
--- a/website/app/controllers/show.js
+++ b/website/app/controllers/show.js
@@ -3,10 +3,14 @@ import showdown from 'showdown';
 import { set, action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { A } from '@ember/array';
+import { schedule } from '@ember/runloop';
+import { inject as service } from '@ember/service';
 
 import { showdownConfig } from '../shared/showdown-config';
 
 export default class ShowController extends Controller {
+  @service fastboot;
+
   @tracked sections = A([]);
   @tracked tabs = A([]);
   @tracked tocs = A([]);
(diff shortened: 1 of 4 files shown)


website/app/controllers/show.js
```diff
@@ -20,16 +24,21 @@ export default class ShowController extends Controller {
...
 
-  @action
-  didInsertContent(docPageContentElement) {
+  didInsertContent = () => {
     let sections = [];
     let tabs = [];
     let tocs = [];
-    docPageContentElement
+    document
       .querySelectorAll(`section[id^=section-]`)
```
didoo, line 42: since we changed the parent that we use to execute the `querySelectorAll` DOM query, I would prefer to be more specific then in the selection

```suggestion
      .querySelectorAll(`.doc-page-content section[id^=section-]`)
```

didoo (approved):
Another amazing PR! Please don't stop! 😀

@Dhaulagiri now that you have explained it, makes perfectly sense that it didn't work anymore.
In any case I think we will do an overall rethink of the whole _"load + tabs + sidecar + query params"_ thing sooner or later.
So perfectly fine to continue to iterate on these things, until we find exactly **what we want**, and then reverse engineer **what is best**.

didoo:
@Dhaulagiri now that you have explained it, makes perfectly sense that it didn't work anymore.
In any case I think we will do an overall rethink of the whole _"load + tabs + sidecar + query params"_ thing sooner or later.
So perfectly fine to continue to iterate on these things, until we find exactly **what we want**, and then reverse engineer **what is best**.

Dhaulagiri:
> In any case I think we will do an overall rethink of the whole "load + tabs + sidecar + query params" thing sooner or later.

Agreed on that.

PR #803, opened 2022-12-12
New docs website / Renamed “About/Overview” section

didoo:
### :pushpin: Summary

This small PR fixes the wrong content folder organization for the "About" section, as discussed with @heatherlarsen


heatherlarsen (approved):
Ahh, I see that "About" and "Overview" are separate pages. We'll only have content for one of these pages, so we should drop "Overview" from the sidenav. And I'll need to think more about how a user gets back to that page since it's not a link in the sidenav (it's a problem for all overview/main pages, actually), and clicking the global navigation link isn't obvious. But we can tackle that in a follow-up, as necessary.

didoo:
> Ahh, I see that "About" and "Overview" are separate pages. We'll only have content for one of these pages, so we should drop "Overview" from the sidenav. And I'll need to think more about how a user gets back to that page since it's not a link in the sidenav (it's a problem for all overview/main pages, actually), and clicking the global navigation link isn't obvious. But we can tackle that in a follow-up, as necessary.

For me, technically, the `About` (as is now in this PR) is the `index` of the whole `About` section, not the `About HDS` page. In the same way that the `Components` index does show the cards for all the subsections, not only the `Components`. That said, for now I am removing the 'Overview' page and we'll see when we have real content what to do with it.

heatherlarsen:
Thanks for jumping on a quick call @didoo. Makes sense now to keep "Overview" and not bother with any sort of "back" functionality. We'll use `/about` as the index and `/about/overview` for the "About HDS" content.

PR #809, opened 2022-12-13
New docs website / "Typography" page styling

didoo:
### :pushpin: Summary

Small PR to add some styling to the list of CSS helper classes in the "Typography" page.

### :hammer_and_wrench: Detailed description

In this PR I have:
- added `Doc::FontHelpersList` component
- updated the documentation to use the newly created `Doc::FontHelpersList` component

Notice: I haven't changed the showcase, we have to discuss it in person, it's not as easy as it seams.

👉 👉 👉 **Preview:** 

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-1240
Figma file: https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=1131%3A67683

***

### 👀 Reviewer's checklist:

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #791, opened 2022-12-09
New docs website / Finalize "banner" `Doc::Banner` component

didoo:
### :pushpin: Summary

This PR build on top of #761 to introduce the `Doc::Banner` element (with the corresponding `!!! Banner` custom syntax for markdow).

### :hammer_and_wrench: Detailed description

In this PR I have:
- implemented the `Doc::Banner` component
- added a testing page for `Banner` content blocks

👉 👉 👉 **Preview**: https://hds-website-git-new-docs-website-content-block-654209-hashicorp.vercel.app/testing/components/banner/

**Important:** 

In order to avoid possible issues with the order of the headings, I had to do a "trick" based on convention and consider the first isolated "strong" element in the markdown (which is then wrapped in a paragraph) as the "title" of the banner (there were no other ways I could think of).

This means this markdown code:

```
!!! Info

**This is the title**

This is a paragraph
!!!
```

gets converted to this HTML markdown:

```
<div class="doc-banner doc-banner--type-information">
    <svg class="flight-icon">
        <use href="#flight-info-24"></use>
    </svg>
    <div class="doc-banner__content">  
        <p class="doc-markdown-p"><strong>This is the title</strong></p>
        <p class="doc-markdown-p">This is a paragraph</p>
    </div>
</div>
```
and rendered as shown below.

@cveigt @jorytindall @heatherlarsen can you check if the rendered version is what you expect?
@MelSumner @alex-ju can you review the code?
@MelSumner can you have a look also from the accessibility perspective?

### :camera_flash: Screenshots

**In code**:
screenshot_2166

**In design** ([Figma file](https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=1467%3A66784&t=SG6x8piu4ObfVyAR-1))
screenshot_2167

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-939
Figma file: https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=1467%3A66784&t=SG6x8piu4ObfVyAR-1

PR #810, opened 2022-12-13
New docs website / Fix markdown processing

didoo:
### :pushpin: Summary

While working on #809 I realized that the `Banner` syntax added in #791 worked in the test page/example, but actually didn't work **at all** when used in an imported file (via custom `@import` markdown syntax).

After a few experimentations trying to fix the problem, I started to fear that the root cause what that `data-markdown="1"` attribute applied to the main `<section>` containers (the banner blocks worked perfectly well in an imported file, without a section with that attribute). So I decided to bite the bullet and try to fix the problem at the root: use the same technique used in #791 and apply it for the `<section>` main containers elements (that generate the tabs).

At the end the effort paid out, and seems we solved two problems with one change: now I don't see anymore duplicated CSS class names assigned to the elements (🎉).

In this PR I have:
- updated the “Banner” test examples to use includes, in different ways, to cover all the use cases
- fixed the crashing of the app introducing a new `page-sections` extension for `Showdown`
  - it was working but elements still had too many identical classes applied to them
- fixed the multiple classes problem by removing the `<div data-markdown="1">` wrapper for the content of both `page-sections` and `content-blocks
- updated the "typography" doc page, using a `Banner`, to make sure everything was working as expected
- replaced all the `<section>` declarations in the `docs` using the new syntax
- updated the "tabs" logic in the `show` controller to target new sections format

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-1191

***

### 👀 Reviewer's checklist:

### DIFF
--- a/website/app/controllers/show.js
+++ b/website/app/controllers/show.js
@@ -44,19 +44,20 @@ export default class ShowController extends Controller {
     let tabs = [];
     let tocs = [];
     document
-      .querySelectorAll(`.doc-page-content section[id^=section-]`)
+      .querySelectorAll(`.doc-page-content section[data-tab]`)
       .forEach((section, index) => {
         // SECTIONS
-        const name = section.id.replace(/^section-/, '');
+        const id = section.id;
+        const name = section.getAttribute('data-tab');
         section.setAttribute('role', 'tabpanel');
         section.setAttribute('tabindex', '0');
-        section.setAttribute('aria-labelledby', `tab-${name}`);
+        section.setAttribute('aria-labelledby', `tab-${id}`);
         section.setAttribute('hidden', true);
         sections.push(section);
         // TABS
         tabs.push({
           index,
-          id: `tab-${name}`,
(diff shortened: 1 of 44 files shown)


alex-ju (approved):
Looks like a way forward! :shipit: Those php/asp escape hatches are 🤯

didoo:
> This also seems to fix the `maximum nesting of 10 spans reached!!!` error as well!

@alex-ju do you want to do a spike and see if we can remove your custom function, in this case? maybe things are fixed and we don't need the extra code

Dhaulagiri (approved):

alex-ju:
> > This also seems to fix the `maximum nesting of 10 spans reached!!!` error as well!
> 
> @alex-ju do you want to do a spike and see if we can remove your custom function, in this case? maybe things are fixed and we don't need the extra code 

Ah, amazing – I'm confident we can. Will follow up on this

PR #735, opened 2022-11-29
New docs website / Icons page

didoo:
### :pushpin: Summary

This PR implements the `Icons` page of the `Foundation` section, per designs.

### :hammer_and_wrench: Detailed description

In this PR I have:
- added a markdown "doc" page for `Icons` 
- added a `Doc::IconList` component as content to the "Library" section
  - the component has been ported from the FlightIcons website and then refactored and re-skinned to follow the latest design specs
- ported the existing documentation ([Getting started for designers](https://flight-hashicorp.vercel.app/design), [Getting started for engineers](https://flight-hashicorp.vercel.app/engineering)) and associated images, from the FlightIcons website to the "Code", "Guidelines" and "Contributing" sections

👉 👉 👉 **Preview**: Preview: https://hds-website-git-new-docs-website-icons-page-hashicorp.vercel.app/foundations/icons/

### 🚧 TODOs
- [ ] ~~understand why the query parameters are not added to the URL~~
    - opened a specific ticket ([HDS-1216](https://hashicorp.atlassian.net/browse/HDS-1216)) to tackle this issue to avoid this being blocked

### :camera_flash: Screenshots

**In code**
screenshot_2156

**In design** ([Figma file](https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=924%3A50478))
screenshot_2155

### :link: External links

JIRA ticket: https://hashicorp.atlassian.net/browse/HDS-931
Figma file: https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=924%3A50478

***

### 👀 How to review

👉 Review commit-by-commit or by files changed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #812, opened 2022-12-13
Updating CopyButton interactive state and fixing width bug.

jorytindall:
### :pushpin: Summary

This PR solves a small bug that was occurring in the `ColorSwatch` component in which the `CopyButton` was filling the rest of the grid container and looked strange on hover interaction.

Also updated the hover interaction to cursor on hover.

Before:
_GitHub is preventing me from uploading images for some reason_

### :link: External links

[HDS-1058](https://hashicorp.atlassian.net/browse/HDS-1058?atlOrigin=eyJpIjoiNTNiYTcyODUyZmYzNDhiMWI4MWY0YzAzNDljNTA3ZGEiLCJwIjoiaiJ9)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/website/app/styles/components/copy-button.scss
+++ b/website/app/styles/components/copy-button.scss
@@ -5,9 +5,13 @@
 .doc-copy-button {
   display: inline-flex;
   align-items: center;
-  max-width: 100%;
+  max-width: fit-content;
   border: 1px solid transparent;
   border-radius: 4px;
+
+  &:hover {
+    cursor: pointer;
+  }
 }
 
 .doc-copy-button--type-solid {


MelSumner:
I think we should have an option to have a full width copy button since it's used in multiple places. WDYT?

website/app/styles/components/copy-button.scss
```diff
@@ -5,9 +5,13 @@
 .doc-copy-button {
   display: inline-flex;
   align-items: center;
-  max-width: 100%;
+  max-width: fit-content;
```
didoo, line 8: @jorytindall I tend to use `fit-content` very rarely, because despite what the name says, it tends to do much more than one would expect (and behave unexpectedly).
I've just merged #735 in `main` and if you pull you will see that with the proposed change, this happens:

**with current code (`max-width: 100%`)**
screenshot_2181

**with proposed code (`max-width: fit-content`)**
screenshot_2182

(try to change the value in the browser here https://hds-website-hashicorp.vercel.app/foundations/icons/ and see what happens)

didoo (changes requested):

website/app/styles/components/copy-button.scss
didoo: @MelSumner if you wrap the `Doc::CopyButton` in a container (eg. a `<div>`) instead of using it as direct child of the grid, it will fix the issue. Or don't use a grid at all but a flexbox (see if how it's done here suits your needs: https://hds-website-hashicorp.vercel.app/foundations/icons/)

jorytindall: @didoo ok got it, makes sense, thanks for the note

didoo (approved):
👍

(please update the PR title to reflect the actual changes)

PR #817, opened 2022-12-14
New docs website / Fix `Page::Sidecar`

didoo:
### :pushpin: Summary

This PR fixes a couple of things in the `Page::Sidecar` component.

In this PR I have:
- changed logic in `show` so that the sidecar gets populated also when there are no `<section data-tab="">` containers
- updated styling for sidecar according to latest designs

The selection of the "active" item according to the URL is something still to do (low priority)

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-1218
Figma file: https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=1315%3A84058&t=lSi4bFHEq0ZWgvSk-1]

### DIFF
--- a/website/app/components/doc/page/sidecar.hbs
+++ b/website/app/components/doc/page/sidecar.hbs
@@ -1,7 +1,7 @@
 <div class='doc-page-sidecar'>
-  <p class='doc-page-sidecar__on-this-page doc-text-label'>On this page:</p>
+  <p class='doc-page-sidecar__on-this-page doc-text-label' id="nav-sidecar">On this page:</p>
   {{#each @tocs as |toc|}}
-    <ul class='doc-page-sidecar__list' id={{toc.id}} hidden={{not toc.isCurrent}}>
+    <ul class='doc-page-sidecar__list' id={{toc.id}} hidden={{not toc.isCurrent}} aria-labelledby="nav-sidecar">
       {{#each toc.list as |item|}}
         <li
           class='doc-page-sidecar__item doc-page-sidecar__item--depth-{{item.depth}}'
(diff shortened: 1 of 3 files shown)


website/app/components/doc/page/sidecar.hbs
```diff
@@ -1,12 +1,12 @@
 <div class='doc-page-sidecar'>
-  <p class='doc-page-sidecar__on-this-page doc-text-h6'>On this page:</p>
+  <p class='doc-page-sidecar__on-this-page doc-text-label'>On this page:</p>
```
MelSumner, line 2:
```suggestion
  <p class='doc-page-sidecar__on-this-page doc-text-label' id="nav-sidecar">On this page:</p>
```

didoo:
> Added a couple of a11y improvement suggestions but they could also be implemented in a separate PR.

Applied your suggestions. Thanks @MelSumner 🙏

website/app/controllers/show.js
```diff
@@ -8,6 +8,21 @@ import { inject as service } from '@ember/service';
 
 import { showdownConfig } from '../shared/showdown-config';
 
+const getTOCs = (container) => {
+  let headings = [];
+  container.querySelectorAll(`:scope > h2, :scope > h3`).forEach((element) => {
+    // we need to add a class to avoid the element being hidden behind the fixed top header
+    element.classList.add('doc-page-sidecar-scroll-margin-top');
```
Dhaulagiri, line 15: it's not clear in the preview app what this is doing as there is already `margin-top` being applied to these same elements. something I'm missing?

Dhaulagiri (approved):
one question about something that I see now already existed, but otherwise 👍

website/app/controllers/show.js
didoo: they're not the same thing. one is the actual margin, the other one is how much space to leave before the element when it's scrolled into the viewport

didoo:
merging to reduce conflicts. @heatherlarsen if you see something not correct, you know what to do :)

PR #804, opened 2022-12-12
initial contribution docs

Dhaulagiri:
### :pushpin: Summary

initial contribution docs

PR #818, opened 2022-12-14
New docs website / Fix bug `elements-to-classnames` Showdown modifier

didoo:
### :pushpin: Summary

While reviewing #804 I noticed that the content was not formatted correctly (the `doc-markdown-***` class names were all over the place). After some testing, I realized that the issue was in the excessive "greadyness" of the regex for tags with attributes in the `elements-to-classnames` Showdown modifier.

This small PR fixes the issue.

**Before:**
screenshot_2191

**After:**
screenshot_2190

### DIFF
--- a/website/app/shared/showdown-extensions/elements-to-classnames.js
+++ b/website/app/shared/showdown-extensions/elements-to-classnames.js
@@ -31,7 +31,6 @@ export const elementsToClassNames = Object.keys(mapElementsToClassNames).map(
     // IMPORTANT: we NEED to set the "g" global option here!
     regex: new RegExp(`<${element}>|<${element} ([^>]*)>`, 'g'),
     replace: function (text) {
-      console.log('text', text);
       // IMPORTANT: we DO NOT NEED to set the "g" global option here!
       const regexBasic = new RegExp(`<${element}>`);
       const matchBasic = text.match(regexBasic);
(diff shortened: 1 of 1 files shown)


website/app/shared/showdown-extensions/elements-to-classnames.js
```diff
@@ -29,12 +29,13 @@ export const elementsToClassNames = Object.keys(mapElementsToClassNames).map(
     // this is a custom regex, modified from the one found in the original tutolrial, to make it more solid and encompass more use cases
     // for testing see: https://regex101.com/r/jLk7wN/2 + https://regex101.com/r/jLk7wN/5
     // IMPORTANT: we NEED to set the "g" global option here!
-    regex: new RegExp(`<${element}>|<${element} (.*)>`, 'g'),
+    regex: new RegExp(`<${element}>|<${element} ([^>]*)>`, 'g'),
     replace: function (text) {
+      console.log('text', text);
```
alex-ju, line 34: Remnant from debugging

alex-ju (approved):
I had an attempt at refactoring this while debugging the double class addition, but I dropped it when I realized the cause wasn't from here. Good spot!

website/app/shared/showdown-extensions/elements-to-classnames.js
didoo:
```suggestion
```

PR #821, opened 2022-12-14
New Docs Website / Adds accessibility statement

MelSumner:
### :pushpin: Summary

If merged, this PR adds a page with an accessibility statement.

### DIFF
--- a/website/docs/about/03--accessibility-statement.md
+++ b/website/docs/about/03--accessibility-statement.md
@@ -1,0 +1,60 @@
+---
+title: Accessibility Statement
+description: This is an accessibility statement from the HashiCorp Design Systems team.
+---
+
+Accessibility is a core requirement, not an optional feature; this aligns with our [company principles](https://www.hashicorp.com/our-principles), especially our principle of integrity.
+
+> Integrity requires a consistency of our thoughts, words, and actions and a dedication to the truth.
+
+To this end, we have built the HashiCorp Design System with accessibility as a core requirement. In this way, we are giving HashiCorp product designers and developers a greater ability to deliver accessible product experiences for all HashiCorp users.
+
+## Ensuring accessible outcomes
+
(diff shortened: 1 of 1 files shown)


Dhaulagiri (approved):

website/docs/about/03--accessibility-statement.md
```diff
@@ -0,0 +1,60 @@
+---
+title: Accessibility Statement
+description: This is an accessibility statement from the HashiCorp Design Systems team.
```
heatherlarsen, line 3: This description doesn't feel like it's adding extra value, so we should consider removing it.

website/docs/about/03--accessibility-statement.md
```diff
@@ -0,0 +1,60 @@
...
+
+We intend to be compatible with the following assistive technologies:
+
+- Safari with VoiceOver on macOS
+- Chrome with Jaws on Windows
+- Firefox with JAWS and NVDA on Windows
+
+## Conformance status
+
+The Web Content Accessibility Guidelines (WCAG) defines requirements for designers and developers to improve accessibility for people with disabilities. It defines three levels of conformance: Level A, Level AA, and Level AAA. The HashiCorp Design System intends to be fully conformant with WCAG 2.1 Level AA. Fully conformant means that the content fully conforms to the accessibility standard.
```
heatherlarsen, line 33: Could we make "for designers and developers" a bit more general? It should be clear that Accessibility is really everyone's responsibility. For example, if we had a content writer on our team, there are WCAG guidelines that would also be relevant to their work.

heatherlarsen (approved):

website/docs/about/03--accessibility-statement.md
MelSumner: ok, taking a note to work on this, will do in a separate PR so we can ship the first iteration.

PR #823, opened 2022-12-14
Fix tabs error when no tabs are present

Dhaulagiri:
### :pushpin: Summary

Fixes issues with tab logic on pages where there aren't any tabs

Visiting a page like `about/02--hds-principles/` in prod you'll see a console error because we're trying to set the query param for a tab that doesn't exist. You'd also see that query params from pages with tabs persist to pages without tabs. This should help with that.


MelSumner (approved):
💯 thank you for fixing this bug!

PR #826, opened 2022-12-14
fix fingerprint rewrite bug

Dhaulagiri:
### :pushpin: Summary

Fixes a bug where markdown images with dimension attributes are not rendered properly in production on the icons page

After much investigation I learned the following
* It was clear that the dimension attribute was causing these files to not be fingerprinted whereas regular images were fingerprinted fine
* Ember's asset pipeline code is broccoli-based, old and hard (for me) to debug
* I ultimately tracked this down to the way that `broccoli-asset-rewrite` looks in files and attempts to replace the non-fingerprinted string with the fingerprinted string

This leaves us a few options, the first of which I've taken here but the others are good to enumerate should this approach not work:
* Monkeypatch `broccoli-asset-rewrite` to adjust its regex so that it can finger print these files
* Don't use dimensions on images in markdown
* Turn off fingerprinting of images

I've chosen the first option, however it comes with some amount of risk:
* I don't really understand regexes. What I did works, but I don't know what side effects this may cause and the only way to really test this is to smoke test the site in production
* I've never actually used [`yarn patch`](https://yarnpkg.com/cli/patch) before. It seems like this is an ideal case for it, but it also seems like a way to hurt yourself if you use it too much. Given that `broccoli-asset-rewrite` is effectively unmaintained and unlikely to ever be updated, the risk here seems low.

### :link: External links

Jira ticket: [HDS-1245](https://hashicorp.atlassian.net/browse/HDS-1245)


.yarn/patches/broccoli-asset-rewrite-npm-2.0.0-c4ce42084a.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/index.js b/index.js
+index 7d3e96c4fafc3ef01e7f6a5a884d0500fa031b31..c78d43eff3a69b24fc23605d8ee0e8c3a15e64e3 100644
+--- a/index.js
++++ b/index.js
+@@ -106,7 +106,7 @@ AssetRewrite.prototype.rewriteAssetPath = function (string, assetPath, replaceme
+    * ["\')>\s] - Match one of "'(\n> exactly one time
+    */
+ 
+-  var re = new RegExp('["\'(=]\\s*([^"\'()=]*' + escapeRegExp(assetPath) + '[^"\n\'()\\>=]*)(\\?[^"\')> ]*)?\\s*\\\\*\\s*["\')>\s]', 'g');
++  var re = new RegExp('["\'(=]\\s*([^"\'()=]*' + escapeRegExp(assetPath) + '[^"\n\'()\\>=]*)(\\?[^"\')> ]*)?\\s*\\\\*\\s*["=\')>\s]', 'g');
```
Dhaulagiri, line 10: ensures we also match `(/assets/getting-started/contribution/icon-dimensions-sq-24.png =112x112)` which does not have a `)` directly after the filename

Dhaulagiri: _but_ i don't know for sure what other ramifications this might have

Dhaulagiri: You can play with this regex [here](https://regex101.com/r/yY3YBG/1)

Dhaulagiri:
@didoo  great question. I'm going to update the image testing page to use an image from our assets folder and that will be a perfect place to test this.

Dhaulagiri:
@didoo have a look at https://hds-website-4ojhj3yki-hashicorp.vercel.app/testing/markdown/more-complex-images/. I updated it to use a local file and I think it captures most all of the examples on the showdown docs

MelSumner (approved):
I think it'll probably be obvious if this does something else that we're not happy about,  and so far this looks like it's working as expected.

didoo:
@Dhaulagiri give me another day to think about this, I am not super enthusiast of the idea of using a yarn patch.

Dhaulagiri:
👍  no rush if we want to consider our options

didoo (approved):

didoo:
@Dhaulagiri let's merge it as is for now (otherwise will hinder the cleanup/review process of the documentation).

I've captured a follow-up ticket here: https://hashicorp.atlassian.net/browse/HDS-1304

PR #827, opened 2022-12-14
Add changeset for finalized Hds::Table component

MelSumner:
### :pushpin: Summary

Adds a changeset for the finalized `Hds::Table` component to prep for release.

### DIFF
--- a/.changeset/long-walls-vanish.md
+++ b/.changeset/long-walls-vanish.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": minor
 ---
 
-Finalized Hds::Table component
+[PR 760](https://github.com/hashicorp/design-system/pull/760) Finalized Hds::Table component; resolved text alignment and column width invocation options. Added support for `th` elements with the scope of `row`. 
--- a/.changeset/long-walls-vanish.md
+++ b/.changeset/long-walls-vanish.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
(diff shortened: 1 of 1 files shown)


.changeset/long-walls-vanish.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+Finalized Hds::Table component
```
Dhaulagiri, line 5: Can you add a more detailed entry here so consumers know what we did and also link back to the original PR(s) where the changeset would have been

Dhaulagiri (changes requested):

.changeset/long-walls-vanish.md
MelSumner: I think that would go in the version PR, yes? Don't these changelogs get deleted? 

(I'm asking because I wasn't sure how to do it, so I went back and looked at previous files)

Dhaulagiri: They do, but the content in the changelog is taken from here as part of that process. 

It _is_ possible to manually adjust the changelog in the version PR, but that should be an exception.

MelSumner, line 5:
```suggestion
[PR 760](https://github.com/hashicorp/design-system/pull/760) Finalized Hds::Table component; resolved text alignment and column width invocation options. Added support for `th` elements with the scope of `row`. 
```

Dhaulagiri (approved):

PR #829, opened 2022-12-14
New docs website / "Icons" page tweaks

didoo:
### :pushpin: Summary

Follow up of conversation with @cveigt, here's some updates requested for the "Icons" page:
- change "Filter records" to "Filter"
- reduce space below the filters' labels
- have `24px` as default selected size for the filter instead of `16px`

👉 👉 👉 **Preview**: https://hds-website-git-new-docs-website-icons-tweaks-hashicorp.vercel.app/foundations/icons?tab=library/

@Dhaulagiri can you check that the changes I made to the way the query params are consumed in the page works for you?

### DIFF
--- a/website/docs/foundations/icons/index.js
+++ b/website/docs/foundations/icons/index.js
@@ -1,5 +1,6 @@
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
 import { restartableTask, timeout } from 'ember-concurrency';
 import { inject as service } from '@ember/service';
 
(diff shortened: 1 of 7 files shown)


cveigt (approved):

website/docs/foundations/icons/index.js
```diff
@@ -10,6 +11,9 @@ const DEBOUNCE_MS = 250;
 export default class Index extends Component {
   @service router;
 
+  @tracked selectedIconSize =
```
Dhaulagiri, line 14: curious if you found it necessary to move this to be tracked here? I didn't do the same for `searchQuery` which may or may not be a problem actually.

didoo: yes, otherwise it didn't work in updating the "selected" value in the `<select>` element (but if you want I can try to remove it and see what happens

PR #832, opened 2022-12-14
Breadcrumb docs: Adds `description` and `caption`

MelSumner:
### :pushpin: Summary

If merged, this PR adds `description` and `caption` to breadcrumb component documentation. 

- description: A breadcrumb is a type of secondary navigation that reveals the user's location in a website.
- caption: A secondary navigation that shows the user's current location.

***


website/docs/components/breadcrumb/index.md
```diff
@@ -1,5 +1,7 @@
 ---
 title: Breadcrumb
+description: A breadcrumb is a type of secondary navigation that reveals the user's location in a website.
```
heatherlarsen, line 3: We use "application" more than we use "website", so probably good to keep them consistent.

```suggestion
description: A breadcrumb is a type of secondary navigation that reveals the user's location in an application.
```

heatherlarsen (changes requested):

website/docs/components/breadcrumb/index.md
alex-ju: I sense a bit of inconsistency with this description, in a sense that we state it as a definition [a Horse is a mammal that...], whereas in other places we don't mention the component name [a mammal that...]

alex-ju (approved):

PR #842, opened 2022-12-15
New docs website / Update code blocks

alex-ju:
### :pushpin: Summary

Update code blocks' style and enable copy functionality.

To enable copy functionality on code blocks, we pass the encoded code block to the `CopyButton` component and decode it before passing it to the clipboard. We do this because multiline code and ember syntax are causing issue to the renderer.

The updated code block style is based on the 'dracula' theme with some color replacements to match the design reference.

<table>
<tr><td>

localhost_5000_components_alert_tab=code_ (1)

</td><td>

localhost_5000_components_alert_tab=code_

</td></tr>
</table>

### :link: External links

Jira ticket: [HDS-938](https://hashicorp.atlassian.net/browse/HDS-938)
[Figma file](https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=1368%3A86135&t=gjMnCvtJWgauQzEF-1)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/website/app/initializers/showdown-extensions.js
+++ b/website/app/initializers/showdown-extensions.js
@@ -58,6 +58,9 @@ export function initialize(/* application */) {
         var end = (options.omitExtraWLInCodeBlocks) ? '' : '\n';
 
         let codeblock = inputCodeblock;
+        
+        // We encode the codeblock so we can safely pass multi-line code and ember syntax to the `Doc::CopyButton` component
+        let codeblockEncoded =  encodeURI(inputCodeblock);
 
         // First parse the github code block
         // codeblock = showdown.subParser('encodeCode')(codeblock, options, globals);
@@ -97,6 +100,7 @@ export function initialize(/* application */) {
         selfExecutingBlock += '  <div class="doc-code-block__code-rendered">';
         selfExecutingBlock += `    ${inputCodeblock}`;
         selfExecutingBlock += '  </div>';
+        selfExecutingBlock += `  <Doc::CopyButton @type="solid" @textToCopy='${codeblockEncoded}' @encoded={{true}} />`;
         selfExecutingBlock += `  ${preBlock}`;
         selfExecutingBlock += '</div>';
 
--- a/website/app/styles/app.scss
+++ b/website/app/styles/app.scss
@@ -14,6 +14,9 @@
 @import "pages/application"; // main application frame
 @import "pages/home"; // homepage (index)
 
+// Third-party declarations
+@import "prism-dracula";
+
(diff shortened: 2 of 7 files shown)


website/app/initializers/showdown-extensions.js
```diff
@@ -97,6 +98,7 @@ export function initialize(/* application */) {
         selfExecutingBlock += '  <div class="doc-code-block__code-rendered">';
         selfExecutingBlock += `    ${inputCodeblock}`;
         selfExecutingBlock += '  </div>';
+        selfExecutingBlock += `  <Doc::CopyButton @textToCopy='${codeblockcopy}' @type="solid" @encoded={{true}} />`;
```
didoo, line 101:
```suggestion
        selfExecutingBlock += `  <Doc::CopyButton @type="solid" @textToCopy='${codeblockcopy}' @encoded={{true}} />`;
```

website/app/initializers/showdown-extensions.js
```diff
@@ -58,6 +58,7 @@ export function initialize(/* application */) {
         var end = (options.omitExtraWLInCodeBlocks) ? '' : '\n';
 
         let codeblock = inputCodeblock;
+        let codeblockcopy =  encodeURI(inputCodeblock);
```
didoo, line 61: nitpicking
- maybe add a comment explaining why we need to encode the content (same where we use it in the encoded form)
- maybe use the word `encoded` in the variable name (eg. `codeblockEncoded`)

website/app/styles/app.scss
```diff
@@ -5,6 +5,7 @@
 
 // Tokens (CSS props)
 @import "tokens";
+@import "prism-dracula";
```
didoo, line 8: not sure which method is better, no strong opinions, but just in case see how it's done here: https://github.com/hashicorp/design-system/blob/main/packages/components/ember-cli-build.js#L25

didoo (approved):
mostly nitpick comments. all good, great job!

website/app/styles/app.scss
alex-ju: The thinking behind placing/vendoring it in an SCSS file was that this way we'll be in full control of the import order – which could matter when dealing with overrides. This was the `code-block.scss` will always be loaded after the theme.

didoo: good thinking! didn't realize it. 👍

jorytindall (approved):
Baring the changes that we discussed, looks good! 🚀

PR #844, opened 2022-12-15
New docs website / Responsive page "sidebar" (and "burger" menu)

didoo:
### :pushpin: Summary

This is the second PR for the finalization of the page "sidebar" (the first one is #840).

This involves adding a responsive behaviour for the "sidebar" (controlled by a "burger" menu) and decouple the scrolling of the sidebar from the scrolling of the content.

_Notice: I am going to have a session with @heatherlarsen in which I'll go through the behaviour with her and see if we need to do some changes to the current implementation._

### :hammer_and_wrench: Detailed description

In this PR I have:
- added event handling logic between “burger” menu in the header, the main “application” controller, and the “sidebar”
- refactored breakpoints mixins to simplify their definitions and make them more easy to use
- added `z-index` scale as CSS tokens
- added links to “top level” routes in the sidebar navigation (only for small viewports)
- updated page “header” to use the updated breakpoint mixins
- added responsive behaviour to the “sidebar” (differentiated between homepage an other pages)
  - show/hide via "burger" menu on small viewports
  - fixed on medium/large viewports
  - for the homepage, always show/hide via "burger" menu
  - allowed the sidebar to scroll independently from the content

**Notice**: the actual styling of the sidebar content is done in #845.

👉 👉 👉 **Previews**:
- homepage: https://hds-website-git-new-docs-website-responsive-sidebar-hashicorp.vercel.app/
- details page: https://hds-website-git-new-docs-website-responsive-sidebar-hashicorp.vercel.app/components/alert/

### 🚧 TODOs

- ~~add a reset for the `showSidebarOnSmallViewport` on route change~~
  - tracked here: https://hashicorp.atlassian.net/browse/HDS-1258

### :link: External links

Jira tickets:
- https://hashicorp.atlassian.net/browse/HDS-934 (main ticket)
- https://hashicorp.atlassian.net/browse/HDS-1253 (behaviour)

***

### 👀 Reviewer's checklist:

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #840, opened 2022-12-15
New docs website / Sidebar TOC sorting

didoo:
### :pushpin: Summary

This PR introduces sorting for the TOC of the application (used by the sidebar and the cards lists)

### :hammer_and_wrench: Detailed description

In this PR I have:
- added the missing `weight` to the list of attributes parsed by showdown
- added a custom sorting function to the TOC generation
- removed the numbered prefix from the “docs” markdown files and set correct ordering via “weight” attribute
- renamed `weight` “frontmatter” attribute to `order` per team decision (see screenshot below)

**Preview**: https://hds-website-git-new-docs-website-toc-sorting-hashicorp.vercel.app/about/

Notice: the sorting "algorithm" is covering our most common needs; if special needs will arise, we will consider possible improvements (but they're not straightforward as one may think: how do you rank/order a folder that doesn't contain just an index but multiple pages? do we need to add a specific file - eg. `meta.md/yaml` or `frontmatter.md/yaml` - just to specify its order?)

### :camera_flash: Screenshots

screenshot_2198
([poll](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1671039421969139))

### :link: External links

Jira tickets:
- https://hashicorp.atlassian.net/browse/HDS-934 (main ticket)
- https://hashicorp.atlassian.net/browse/HDS-1252 (sorting)

***

### 👀 Reviewer's checklist:

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #807, opened 2022-12-13
New docs website / "Home" page + `Page::Header` + `Page::Footer`

didoo:
### :pushpin: Summary

This is a large PR that encompass multiple things in one:
- the new design for the homepage
- the revision of the page `header`
- the revision of the page `footer`

To achieve this result, I had to do other changes and refactoring in adjacent code/components.

### :hammer_and_wrench: Detailed description

⚠️ Refer to the list of commits for a more granular view of the changes I made.

In this PR I have:
- implemented page `header` (with responsiveness)
  - added barebones “burger menu” to the header (actual sidebar behaviour will be implemented in a follow-up PR)
- implemented **homepage “hero”** to follow design specs (with responsiveness)
  - added abstract decoration/illustration (with responsiveness)
- implemented **homepage “cards”** to follow design specs (with responsiveness)
- added a `Doc::LinkWithIcon` component (similar to what we call "standalone" in the HDS components)
- added [`ember-body-class`](https://github.com/mansona/ember-body-class) as dependency (to have page-specific class names in the `<body>` of the page, to differentiate the styles depending on the page
- added `hashicorp` logo + `hds` logo (temporary)
- implemented page `footer`(with responsiveness)

👉 👉 👉 **Preview**:  https://hds-website-git-new-docs-website-home-page-hashicorp.vercel.app/

_Notice: when reviewing, look also at the other pages: the header and footer are shared with the homepage_

### 🚧 Still to do
- sidebar show/hide behaviour → will be done in [HDS-934](https://hashicorp.atlassian.net/browse/HDS-934)
- ~~blend the header in the homepage (transparent background, but becomes black on scroll) → nice to have, will be done if we have time before the release~~
  - tracked here: https://hashicorp.atlassian.net/browse/HDS-1257

### :camera_flash: Screenshots

**Large viewport:**
homepage-large

**Medium viewport:**
homepage-medium

**Small** viewport:**
homepage-small

### :link: External links

Jira tickets: 
- https://hashicorp.atlassian.net/browse/HDS-1164
- https://hashicorp.atlassian.net/browse/HDS-933
- https://hashicorp.atlassian.net/browse/HDS-1219

Figma file: 
- https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=771%3A58440
- https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=691%3A55731

***

### 👀 Reviewer's checklist:

[...]

PR #845, opened 2022-12-15
New docs website / Page "sidebar" styling (and other polishing)

didoo:
### :pushpin: Summary

This PR styles the "sidebar" elements according to the design specs. It's built on top of  #807, #840 and #844 (it's been closed in favour of a single PR, this one)

In #844 I have:
- added event handling logic between “burger” menu in the header, the main “application” controller, and the “sidebar”
- refactored breakpoints mixins to simplify their definitions and make them more easy to use
- added `z-index` scale as CSS tokens
- added links to “top level” routes in the sidebar navigation (only for small viewports)
- updated page “header” to use the updated breakpoint mixins
- added responsive behaviour to the “sidebar” (differentiated between homepage an other pages)
  - show/hide via "burger" menu on small viewports
  - fixed on medium/large viewports
  - for the homepage, always show/hide via "burger" menu
  - allowed the sidebar to scroll independently from the content

In this PR I have: 
- done a first pass at styling the "sidebar" elements following the current design specs
  - to be discussed/reviewed with @heatherlarsen because I am not sure what is the "final" design
- did a quick change to make the page "side**c**ar" sticky when the page is scrolled

After a review session with @heatherlarsen I have then:
- updated breakpoint for when we show/hide the sidebar in “mobile” mode
  - changed from `1000px` to `880px`, value determined testing in a browser
- made the variable `--doc-page-sidebar-width` responsive
  - at small/medium viewport the sidebar is slightly smaller to optimize the available space
- changed `max-height` value for the homepage illustration on large viewports
- removed GitHub icon from top-route link in sidebar
- refactored sidebar code to split home vs other pages + implemented home-specific animated version with gradient
- implemented collapsible “folder” in sidebar using `<details>`

👉 👉 👉 **Previews**:
- homepage: https://hds-website-git-new-docs-website-sidebar-styling-hashicorp.vercel.app/
- details page: https://hds-website-git-new-docs-website-sidebar-styling-hashicorp.vercel.app/components/alert/

### 🚧 TODOs
- ~~add a reset for the `showSidebarOnSmallViewport` on route change~~
  - tracked here: https://hashicorp.atlassian.net/browse/HDS-1258
- [ ] understand if we really need to handle overflowing of the content, or is an edge case and we can live without it for now

### :link: External links

Jira tickets::
- https://hashicorp.atlassian.net/browse/HDS-934 (main ticket)
- https://hashicorp.atlassian.net/browse/HDS-1253 (behaviour)
- https://hashicorp.atlassian.net/browse/HDS-1220 (styling)
Figma file: https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=188%3A18330

***

### 👀 Reviewer's checklist:

### DIFF
--- a/website/app/controllers/application.js
+++ b/website/app/controllers/application.js
@@ -1,0 +1,16 @@
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class ApplicationController extends Controller {
+  @tracked showSidebarOnSmallViewport = false;
+
+  // TODO add a reset for the `showSidebarOnSmallViewport` on route change
+
+  @action
+  onToggleBurgerMenu(isOpen) {
+    console.log('onToggleBurgerMenu', isOpen);
+    this.showSidebarOnSmallViewport = isOpen;
+    document.body.style.overflow = isOpen ? 'hidden' : 'auto';
+  }
+}
--- a/website/app/controllers/application.js
+++ b/website/app/controllers/application.js
@@ -9,7 +9,6 @@ export default class ApplicationController extends Controller {
 
   @action
   onToggleBurgerMenu(isOpen) {
-    console.log('onToggleBurgerMenu', isOpen);
     this.showSidebarOnSmallViewport = isOpen;
     document.body.style.overflow = isOpen ? 'hidden' : 'auto';
   }
--- a/website/app/components/doc/page/header/index.hbs
+++ b/website/app/components/doc/page/header/index.hbs
@@ -16,5 +16,11 @@
         ><FlightIcon @name="github" @size="24" @isInlineBlock={{false}} /><span class="sr-only">GitHub</span></a></li>
     </ul>
   </nav>
-  <button type="button" class="doc-page-header__burger-menu" {{on "click" this.toggleBurgerMenu}}><FlightIcon @name={{if this.burgerMenuOpen "x" "menu"}} /> <span class="doc-text-navigation">Menu</span></button>
+  <button type="button" class="doc-page-header__burger-menu" {{on "click" this.onToggleBurgerMenu}}>
+    {{#if this.burgerMenuOpen}}
+      <FlightIcon @name="x" /> <span class="doc-text-navigation">Close</span>
+    {{else}}
+      <FlightIcon @name="menu" /> <span class="doc-text-navigation">Menu</span>
+    {{/if}}
+  </button>
 </header>
--- a/website/app/components/doc/page/header/index.js
+++ b/website/app/components/doc/page/header/index.js
@@ -12,7 +12,12 @@ export default class DocPageHeaderComponent extends Component {
   }
 
   @action
-  toggleBurgerMenu() {
+  onToggleBurgerMenu() {
     this.burgerMenuOpen = !this.burgerMenuOpen;
+    let { onToggleBurgerMenu } = this.args;
+
+    if (typeof onToggleBurgerMenu === 'function') {
+      onToggleBurgerMenu(this.burgerMenuOpen);
+    }
   }
 }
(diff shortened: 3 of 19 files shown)


website/app/controllers/application.js
```diff
@@ -0,0 +1,16 @@
...
+import { tracked } from '@glimmer/tracking';
+
+export default class ApplicationController extends Controller {
+  @tracked showSidebarOnSmallViewport = false;
+
+  // TODO add a reset for the `showSidebarOnSmallViewport` on route change
+
+  @action
+  onToggleBurgerMenu(isOpen) {
+    console.log('onToggleBurgerMenu', isOpen);
```
Dhaulagiri, line 12:
```suggestion

```

Dhaulagiri:
question that i don't see answered in the PR description. on a page like [patterns](https://hds-website-hhgnwqsuk-hashicorp.vercel.app/patterns/) where there is no content in the sidebar, it's now visible but in current production it is not. Is that intentional?

Dhaulagiri:
On [components](https://hds-website-hhgnwqsuk-hashicorp.vercel.app/components/) I got into a state where I couldn't scroll the content of the main body but the sidebar was still scrollable. I could only reproduce it with the following steps:

1. Visit [components](https://hds-website-hhgnwqsuk-hashicorp.vercel.app/components/) 
2. Collapse window so sidebar is hidden
3. Open sidebar via hamburger menu
4. Resize page back to full view
5. Sidebar scrolls, main body doesn't

Re-narrowing the screen and closing the sidebar fixes the issue, so maybe there's some class or logic that's happening when the sidebar is open to prevent main body scrolling that isn't getting cleaned up this case?

didoo:
> On [components](https://hds-website-hhgnwqsuk-hashicorp.vercel.app/components/) I got into a state where I couldn't scroll the content of the main body but the sidebar was still scrollable. I could only reproduce it with the following steps:
> 
> 1. Visit [components](https://hds-website-hhgnwqsuk-hashicorp.vercel.app/components/)
> 2. Collapse window so sidebar is hidden
> 3. Open sidebar via hamburger menu
> 4. Resize page back to full view
> 5. Sidebar scrolls, main body doesn't
> 
> Re-narrowing the screen and closing the sidebar fixes the issue, so maybe there's some class or logic that's happening when the sidebar is open to prevent main body scrolling that isn't getting cleaned up this case?

Yes, we noticed this bug today with @alex-ju. Do you mind creating a bug and assign it to him (we discussed a possible way to fix it). Thanks

didoo:
> question that i don't see answered in the PR description. on a page like [patterns](https://hds-website-hhgnwqsuk-hashicorp.vercel.app/patterns/) where there is no content in the sidebar, it's now visible but in current production it is not. Is that intentional?

Something I have to think about. It's not simple to solve. Also, in theory there should be content there. So... it's temporary. I'll think about it more, maybe I can come up with something.

Dhaulagiri:
Opened https://hashicorp.atlassian.net/browse/HDS-1301 for the scrolling issue. I think it's mostly a nice to have given it can be fixed by reloading the page.

> Also, in theory there should be content there. So... it's temporary.

Makes sense. I think for launch we can either fix by updating the content or just live with it for now.

MelSumner (approved):
Let's ship it if we can, we can iterate.  👍

didoo:
> 1. When the menu is open, I can't scroll through the page's main content. This is happening on all pages.
Known issue, tracked in https://hashicorp.atlassian.net/browse/HDS-1301

> 2. On the components overview page, in the sidebar, the "Form" and "Links" categories should be closed by default.

OK, changed.

> 3. On the detail pages, around 879px, the sidebar collapses, but the global navigation doesn't change until 767px. This results in a doubling up of the "main category" links. Main category links shouldn't appear in the sidebar until 767px.

Yes, my bad, wrong media query. Fixed.

> 4. [nice to have] I didn't see the transition on the homepage when the menu opens. It just appears.

Yes, I have disabled it, but I'll re-enable and track things in a follow up ticket: https://hashicorp.atlassian.net/browse/HDS-1306

> 5. Not sure if this is the correct place to ask this, but is there a way to hide "On this page" in the sidecar if there are no links available?

We'll review how to handle this case when we have more realistic content (so, in January)

PR #867, opened 2022-12-15
Bugfix: Resolve issue where extra spacing before `required` badges in form labels was not being applied consistently

MelSumner:
### :pushpin: Summary

 If merged, this PR provides a bugfix that removes the extra space between the label and the required badge that wasn't present in the legend, causing an inconsistency.  Instead, the non-breaking space has been added to the `required` indicator, so it's consistently applied to both.

Adds patch changeset.

Reported issue: 
CleanShot 2022-12-15 at 13 56 41@2x

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/honest-lemons-dream.md
+++ b/.changeset/honest-lemons-dream.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Bugfix: removed extra space in label that was causing label and legend to be inconsistent
--- a/.changeset/honest-lemons-dream.md
+++ b/.changeset/honest-lemons-dream.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-Bugfix: removed extra space in label that was causing label and legend to be inconsistent
+Bugfix: removed extra space in label that was causing label and legend to be inconsistent. Moved the `&nbsp;` to the `required` indicator instead for consistency.
--- a/packages/components/addon/components/hds/form/label/index.hbs
+++ b/packages/components/addon/components/hds/form/label/index.hbs
@@ -1,4 +1,4 @@
 <label class={{this.classNames}} for={{@controlId}} ...attributes>
-  {{yield}}{{#if (or @isRequired @isOptional)}}&nbsp;{{/if}}
+  {{yield}}{{#if (or @isRequired @isOptional)}}{{/if}}
   <Hds::Form::Indicator @isRequired={{@isRequired}} @isOptional={{@isOptional}} />
 </label>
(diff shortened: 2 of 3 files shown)


Dhaulagiri:
it looks like @alex-ju [added this intentionally](https://github.com/hashicorp/design-system/pull/534) so I'll let him chime in on if this change impacts the intent from #534

MelSumner:
@Dhaulagiri ahh yes it could need to be flipped (both have the space instead of neither having the space). It seems like maybe that should be in the CSS though? Fine with waiting for @alex-ju on this one.

alex-ju:
> @Dhaulagiri ahh yes it could need to be flipped (both have the space instead of neither having the space). It seems like maybe that should be in the CSS though? Fine with waiting for @alex-ju on this one.

Yes, I'd say it's best to flip this PR (add it to the legend as well) – I remember trying lots of different options (including adding the space via CSS content) but the only thing that worked at the time was a non-breaking space, so unless we want to spend more time investigating potential alternatives, I'd say we can use the same approach for the legend.

@MelSumner if you don't have the bandwidth for this let me know and I can pick it up

MelSumner:
@alex-ju ok I am going to try putting the non-breaking space in front of the required badge instead of in the label or legend. That way it _should_ apply to both. WDYT?

.changeset/honest-lemons-dream.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Bugfix: removed extra space in label that was causing label and legend to be inconsistent
```
MelSumner, line 5:
```suggestion
Bugfix: removed extra space in label that was causing label and legend to be inconsistent. Moved the `&nbsp;` to the `required` indicator instead for consistency.
```

packages/components/addon/components/hds/form/label/index.hbs
```diff
@@ -1,4 +1,4 @@
 <label class={{this.classNames}} for={{@controlId}} ...attributes>
-  {{yield}}{{#if (or @isRequired @isOptional)}}&nbsp;{{/if}}
+  {{yield}}{{#if (or @isRequired @isOptional)}}{{/if}}
```
alex-ju, line 2: The conditional statement here shall go away with the `&nbsp;`
```suggestion
  {{yield}}
```

alex-ju:
> @alex-ju ok I am going to try putting the non-breaking space in front of the required badge instead of in the label or legend. That way it _should_ apply to both. WDYT?

yep, adding it to the indicator makes sense to me, this way we make sure that whatever solution we have applies consistently across form elements

Dhaulagiri:
Gentle nudge to remember to check for percy updates on PRs like this. It doesn't look like [the diff here](https://percy.io/HashiCorp/ds-components/builds/23724440/changed/1327118832?browser=firefox&browser_ids=30%2C31&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=1280&widths=375%2C1280) got approved which is causing diffs against `main` from other PRs to now fail as well.

Dhaulagiri:
ok, i was at least half wrong since the failure i was seeing was due to my branch being out of sync. I think the original percy diff _wasn't_ approved and anything that makes it to main is just assumed correct, but percy is a bit nebulous to me.

PR #868, opened 2022-12-15
More tab query param fixes

Dhaulagiri:
### :pushpin: Summary

This fixes a couple of small issues:
* Navigating back from a component page required hitting the back button multiple times
* Once you navigated to the non-default tab for a component, the link in the sidebar would then include that tab in the link when we would want that to take someone to the default tab
* If you're on the default tab, we don't need to have that in the query params

Does not fix
* Hitting the back button from the non-default tab doesn't update which tab you're on 😢 

### :link: External links

Jira ticket: [HDS-1251](https://hashicorp.atlassian.net/browse/HDS-1251)

### DIFF
--- a/website/app/components/doc/table-of-contents/index.hbs
+++ b/website/app/components/doc/table-of-contents/index.hbs
@@ -4,7 +4,7 @@
   {{#each-in @structuredPageTree as |key item|}}
     <li class="doc-table-of-contents__item doc-table-of-contents__item--depth-{{@depth}}">
       {{#if item.pageURL}}
-        <LinkTo class="doc-table-of-contents__link doc-text-navigation" @route="show" @model={{item.pageURL}}>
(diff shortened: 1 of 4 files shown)


didoo:
@Dhaulagiri I'm doing a lot of work around the sidebar and the links in it. probably better to move it to draft for now, and wait that they're merged and then rebase (or bring the changes in a fresh branch)

Dhaulagiri:
@didoo no particular rush for me, but the sidebar change could be easily dropped and done later if that would make it cleaner

didoo (approved):
@Dhaulagiri I have rebased and fixed the conflicts. I'm approving this PR, in any case we will review all the "navigation + query-params + tabs + sidecar" soon, so we'll have a better picture of what we want to change there.

PR #872, opened 2022-12-16
`Pagination` truncation (HDS-1174)

KristinLBradley:
### :pushpin: Summary

Adds a "truncated" type variant to display ellipses for some numbers.

**Preview:** https://hds-components-git-hds-1174-pagination-truncation-hashicorp.vercel.app/components/pagination

* This branch will be merged into the https://github.com/hashicorp/design-system/pull/661 once approved. (It will not be merged directly into the main branch.)

Screen Shot 2022-12-19 at 8 36 00 AM

### :link: External links
* Jira ticket: [HDS-1174](https://hashicorp.atlassian.net/browse/HDS-1174)
Figma file: [[if it applies]](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components?node-id=18709%3A42011&t=iKPD0dWMQTqOm4I1-0)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/addon/components/hds/pagination/nav/ellipses.hbs
```diff
@@ -0,0 +1,3 @@
+<li class="hds-pagination-nav__page-item" ...attributes>
+  <div class="hds-pagination-nav__ellipses">...</div>
```
KristinLBradley, line 2: I used 3 dots vs. an ellipses character because visually it more closely matched the Figma mocks. (I'm fine with using either one.)

didoo: Super uber nitpick: should be `ellipsis` and not `ellipses`, which is plural? 
https://www.merriam-webster.com/dictionary/ellipsis

packages/components/addon/components/hds/pagination/nav/index.hbs
```diff
@@ -18,6 +18,22 @@
     </ul>
   {{/if}}
 
+  {{#if (eq this.type "truncated")}}
+    <ul class="hds-pagination-nav__page-list">
+      {{#each this.truncatedPages as |page|}}
```
didoo, line 23: [issue] you don't need a double loop, and a double getter (`this.pages` vs `this.truncatedPages`).

Instead you can do something like this:
```
  get pages() {
    let pages = [];

    for (let i = 1; i <= this.totalPages; i++) {
      pages.push(i);
    }

    if (this.type === 'truncated') {
      return this.elliptize({ pages, current: this.currentPage });
    } else {
      return pages;
    }
  }
```

packages/components/addon/components/hds/pagination/nav/index.hbs
```diff
@@ -18,6 +18,22 @@
...
+        {{#if (eq page "…")}}
+          <Hds::Pagination::Nav::Ellipses />
+        {{else}}
+          <Hds::Pagination::Nav::ButtonNumber
```
didoo, line 27: general comment: I see the text and the buttons jump/move a bit when clicking next/prev buttons (because each character has a different width). A couple of options I can think of:
- apply a small min-width (removing the horizontal padding) - _probably would work only for 1 digits pages_
- apply a `font-variant-numeric: tabular-nums` style to the pages - _this slightly increases the kerning between numeric character, not sure if it's OK (maybe check with @cveigt)_

packages/components/addon/components/hds/pagination/nav/index.js
```diff
@@ -67,6 +67,14 @@ export default class HdsPaginationNavIndexComponent extends Component {
     return pages;
   }
 
+  get truncatedPages() {
+    console.log(
```
didoo, line 71: left a console log.

(also, this function/getter is not need, see comment above)

packages/components/addon/components/hds/pagination/nav/index.js
```diff
@@ -67,6 +67,14 @@ export default class HdsPaginationNavIndexComponent extends Component {
...
+      this.elliptize({ pages: this.pages, current: this.currentPage })
+    );
+    console.log(this.currentPage);
```
didoo, line 74: console log.

packages/components/addon/components/hds/pagination/nav/index.js
```diff
@@ -75,6 +83,46 @@ export default class HdsPaginationNavIndexComponent extends Component {
     return this.currentPage === this.totalPages;
   }
 
+  elliptize({ pages, current }) {
+    const array = Array.from(pages);
```
didoo, line 87: you don't need to create a copy of the array, you can use directly `pages` I think (as far as I remember, arrays are not passed as reference)

packages/components/tests/integration/components/hds/pagination/nav/index-test.js
```diff
@@ -22,13 +22,23 @@ module('Integration | Component | hds/pagination/nav/index', function (hooks) {
 
   // Test API:
 
-  test('it has the same numbr of pages displayed as are passed in', async function (assert) {
+  test('it has the same number of pages displayed as are passed in for the "numbereed" type', async function (assert) {
```
didoo, line 25:
```suggestion
  test('it has the same number of pages displayed as are passed in for the "numbered" type', async function (assert) {
```

packages/components/tests/integration/components/hds/pagination/nav/index-test.js
```diff
@@ -22,13 +22,23 @@ module('Integration | Component | hds/pagination/nav/index', function (hooks) {
...
+  test('it displays a truncated list of page numbers for the "truncated" type', async function (assert) {
+    await render(hbs`
+      <Hds::Pagination::Nav @totalPages={{1000}} @currentPage={{1}} @type="truncated" />
+    `);
+    // displays 7 items (the max number displayed)
+    assert.dom('.hds-pagination-nav__page-item').exists({ count: 7 });
+    // It displays an ellipses for the 5th item
+    assert.dom('.hds-pagination-nav__page-item:nth-child(5)').hasText('...');
+  });
+
```
didoo, line 41: since this is quite tricky algorithm, I would also test that
- when next/prev are clicked, the ellipsis moves
- the ellipsis is in the right position (and the numbers too) for the different combinations
  - `1* 2 3 4 ... 99 100`
  - `1 2* 3 4 ... 99 100`
  - `1 2 3* 4 ... 99 100`
  - `1 ... 3 4* 5 ... 100`
  - `1 ... 96 97* 98 ... 100`
  - `1 ... 97 98* 99 100`
  - `1 ... 97 98 99* 100`
  - `1 ... 97 98 99 100*`

packages/components/addon/components/hds/pagination/nav/index.js
```diff
@@ -75,6 +83,46 @@ export default class HdsPaginationNavIndexComponent extends Component {
...
+    const limit = 7; // limt # of page numbers shown at a time
```
didoo, line 88: sorry, there is a bug in the algorithm (the bug is already in the code I sent you as demo, so my bad)
if you change the limit to 9 you will see that it doesn't work as expected
the reason is in this line:
```
result = [].concat(sliceStart.shift(), ellipsis, sliceCurr, ellipsis, sliceEnd.pop());
```
which assumes that the first block and the last block are of 1 element which is not true if the limit is different (eg. `9`). 

It's not a blocker per se, and it's a "bug" only if one changes the limit from `7` (which btw is something we should consider as an option, seems more easy to use/grasp, maybe check with @cveigt what he thinks)

didoo (changes requested):
Left some comments, some are blockers unfortunately (but I can help you if you want)

packages/components/addon/components/hds/pagination/nav/ellipses.hbs
KristinLBradley: I'll change it, thanks

packages/components/addon/components/hds/pagination/nav/index.hbs
KristinLBradley: I'll take a closer look, thanks

packages/components/addon/components/hds/pagination/nav/index.hbs
```diff
@@ -18,6 +18,22 @@
     </ul>
   {{/if}}
 
+  {{#if (eq this.type "truncated")}}
```
KristinLBradley, line 21: I tried using `#unless` to avoid repeating these chunks of code but the linter complained it was too complex. (I searched but there doesn't seem to be a built in way to do a logical `or`)

packages/components/addon/components/hds/pagination/nav/index.js
```diff
@@ -75,6 +79,57 @@ export default class HdsPaginationNavIndexComponent extends Component {
     return this.currentPage === this.totalPages;
   }
 
+  elliptize({ pages, current }) {
+    const limit = 7; // limit # of page numbers shown at a time (should always be an even number!)
```
KristinLBradley, line 83: @didoo Are you sure it shouldn't say it should be an "odd" number? (vs. "even") or should the default limit be changed?

KristinLBradley: Also right now the limit isn't exposed as an argument, should it be?

didoo: > @didoo Are you sure it shouldn't say it should be an "odd" number? (vs. "even") or should the default limit be changed?

Yes, my bad.

> Also right now the limit isn't exposed as an argument, should it be?

I though the same, yes, but wasn't sure :) As you prefer/think is better (in any case if we had to add it later as option, it would not be a breaking change).

packages/components/addon/components/hds/pagination/nav/index.hbs
didoo: Let's sync on Monday, OK?

packages/components/addon/components/hds/pagination/nav/index.js
KristinLBradley: Ok, not sure if this feature (truncation) would be used enough to require it so I think we could add it later if requested.

packages/components/tests/integration/components/hds/pagination/nav/index-test.js
KristinLBradley: I started adding tests for these combinations but ran into trouble. (I commented out the test above.)

It seemed like it wasn't waiting for the DOM to update after the click fired or at least that's my guess. When I try testing it manually within the Scrappy documentation page it seems to work as expected.

Dhaulagiri: > It seemed like it wasn't waiting for the DOM to update after the click fired

Can take a look with you next week, but try using `await settled();` and see if that helps things wait

KristinLBradley: I can try it again but it didn't work when I tried before (see comment in code). 

I'll keep experimenting though.

packages/components/addon/components/hds/pagination/nav/index.hbs
KristinLBradley: I asked @cveigt  to weigh in on this. 

Personally, I think it would be worse to replace the in-between padding with min-width because then the space between numbers would then be inconsistent. I think adding `font-variant-numeric: tabular-nums` is an ok option to reduce jumping but overall the slight jumping doesn't bother me too much as to me it seems natural for the button sizes to change slightly to accommodate a different length of content.

(@cveigt you can see the different between "normal" and the "tabular-nums" variant on the page: https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-numeric)

cveigt: The shift is an acceptable trade-off. There will be a small shift, but depending on what the current page is, this should happen only once in a while. When it happens, it'll likely be a few pixels ~6px and might not even be noticeable since the pages usually re-render when going from one page to another.

I don't think setting min-width or adding kerning would be the best solution. We likely won't find an ideal value that works for all numbers, and it'll be noticeable in most cases, unlike the shift.

KristinLBradley: Thanks @cveigt I'm going to leave this as is then.

packages/components/tests/integration/components/hds/pagination/nav/index-test.js
KristinLBradley: After discussing with @didoo earlier today I created some new simplified tests for truncation. I can look into adding a more robust test a little later on if needed (also as discussed).

KristinLBradley:
If there are no objections or further feedback, I'd like to just merge this into my feature branch then continue iterating.

alex-ju (approved):
It seems like Cristiano's comments were addressed, so I'm happy it a tick

PR #877, opened 2022-12-16
Enable proper page titles

Dhaulagiri:
### :pushpin: Summary

Fixes a minor issue preventing page titles from updating as expected to the following format

`Page Title | DS Name`

I can't find the thread we had on this from this week, so if there is another desired format we can see how to make that happen.  I think something like `Page Title | Category | DS Name` was the other option.

### :link: External links

Jira ticket: [HDS-1237](https://hashicorp.atlassian.net/browse/HDS-1237)

***

### DIFF
--- a/website/app/templates/show.hbs
+++ b/website/app/templates/show.hbs
@@ -1,4 +1,7 @@
 <HeadLayout />
+
+{{page-title this.title}}
+
 <Doc::Page::Stage>
   {{#if this.model.hasCover}}
     <Doc::Page::Cover @title={{this.title}} @description={{this.description}} @tabs={{this.tabs}} @extra={{this.extra}} />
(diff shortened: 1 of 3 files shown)


MelSumner (approved):
This is a great start! I wonder if we could get to something like `Alert | Components | HDS - Design System`.

website/app/templates/show.hbs
```diff
@@ -1,4 +1,7 @@
 <HeadLayout />
+
+{{page-title this.title}}
```
didoo, line 3: 😱 🤯 😍

Dhaulagiri:
In theory it may be possible but in practice is made more complex by the fact that the detail pages are not actually children of their respective parent routes. 

There's also a bit of a conflict between ember-page-title and ember-meta that I want to sort out at some point.

website/app/templates/show.hbs
Dhaulagiri: i think this is where we could do what @MelSumner is suggesting and somehow buildup `title` to contain the category _and_ the title.

PR #880, opened 2022-12-16
hide content that won't exist at launch

Dhaulagiri:
### :pushpin: Summary

This hides some placeholders for content that we don't expect to exist.

### DIFF
--- a/website/docs/getting-started/adoption.md
+++ b/website/docs/getting-started/adoption.md
@@ -2,6 +2,7 @@
 title: Adoption
 description: This is the (missing) long description of the component, that will come from the frontmatter attributes
 order: 103
+hidden: true
 ---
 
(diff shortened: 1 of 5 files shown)


website/docs/getting-started/adoption.md
```diff
@@ -1,9 +0,0 @@
----
-title: Adoption
-description: This is the (missing) long description of the component, that will come from the frontmatter attributes
-order: 103
```
MelSumner, line 4: Shouldn't we just add `hidden: true` to these pages instead of removing them?

Dhaulagiri: Admittedly I did forget that was an option. I'm inclined to remove them, but for now I've marked them as hidden.

MelSumner (approved):

PR #885, opened 2022-12-16
New docs website / Page "Tokens"

didoo:
### :pushpin: Summary

This PR aligns (as much as possible) the "Tokens" page with the latest designs.

In this PR I have:
- fixed query string for `querySelectorAll` for sidecar
- extracted `select` and `search` controls from the “icons” page and shared as `form` components
  - now we have `Doc::Form::Label`, `Doc::Form::Select`, `Doc::Form::Search`
- added `Doc::TokensList` component
- Added `Doc::TokensList` to “Tokens” page as “library” and implemented filtering logic
- merged all the CSS for the `Doc::IconsList` to a single file
- extracted `Doc::MetaRow` as standalone component
- grouped component-related CSS files under an index file
- implemented “Tokens” list in line as much as possible with the designs, but also took the initiative to implement the "token card" as I would imagine to use it as developer/consumer.

👉 👉 👉 **Preview**: https://hds-website-git-new-docs-website-tokens-page-styling-hashicorp.vercel.app/foundations/tokens?tab=library/

@cveigt this is the best I could achieve, for different technical reasons. 

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-1261
Figma file: https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?node-id=1490%3A62802

***

### 👀 Reviewer's checklist:


alex-ju (approved):
I think this looks great for a first pass. A few things I discovered in review that could potentially be addressed in follow-ups: 
 - the sidecar navigation doesn't work on this page (not sure if we want to fix it or remove it)
 - the disclosure functionality needs some accessibility enhancements (to communicate the expanded/collapsed state of a token list item element

MelSumner (approved):

cveigt:
@didoo here’s my feedback. You can disregard what can’t be achieved due to the technical reasons you mention.

The border radius on the bottom of the `doc-tokens-list__content` are inverted.

Screen Shot 2022-12-19 at 8 58 57 AM

The content inside the list item jumps when the toggle is clicked. It shouldn’t move.

tokens-page

This page should be broken down into Token list, Guidelines, and Code. I can understand if this will be part of the content review.

didoo:
> * the sidecar navigation doesn't work on this page (not sure if we want to fix it or remove it)

@alex-ju we will entirely review how the sidecar works once we have more realistic content across the pages, let's leave this for then (all the logic may change entirely).

> * the disclosure functionality needs some accessibility enhancements (to communicate the expanded/collapsed state of a token list item element

Do you mind taking care of this, and also the same in the `sidebar`, which is also using a native `<details>` element (I would have expected the browser to do the work for us).

didoo:
> The border radius on the bottom of the `doc-tokens-list__content` are inverted.

You're right, fixed.

> The content inside the list item jumps when the toggle is clicked. It shouldn’t move.

I know it would be better, but it's not possible: since we don't know the size of the content inside the "preview" area, there's no way to visually center it when it's closed **and** keep it exactly at the same distance from the top when it's opened.

> This page should be broken down into Token list, Guidelines, and Code. I can understand if this will be part of the content review.

Yes, part of the content review.

cveigt:
If it's not possible at the moment, it'd be better to remove the toggle functionality and just display all the values. Most of them are just showing an additional field, some not even one, and the only ones significantly long are the background-image URLs, but they are currently just two.

PR #886, opened 2022-12-19
New docs website / Remove showdown parsing at build time

didoo:
### :pushpin: Summary

Small PR to remove Showdown parsing of markdown at build time (it’s not used anymore, the page-level TOC for the sidecar is done at runtime now)


Dhaulagiri (approved):
I'm still not convinced we shouldn't do this logic here to avoid some of the client-side challenges, but at least for now 👍  on removing it

PR #892, opened 2022-12-19
Fix `Hds::Table` showcase section

alex-ju:
### :pushpin: Summary

Follow up on #796 to fix the `Hds::Table` showcase section.

At the time, we couldn't figure out what caused the table examples with dynamic data to fail from rendering.
At debugging it turns out we were using an old set of examples for the showcase, as the table component's API evolved since then, so we're updating the examples to match the ones in the dummy website.

### :link: External links

Jira ticket: [HDS-1246](https://hashicorp.atlassian.net/browse/HDS-1246)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


MelSumner (approved):
🎉

PR #884, opened 2022-12-16
[docs] Add About/Overview page content

jorytindall:
### :pushpin: Summary

If merged this PR will add an overview to the about section.

### :hammer_and_wrench: Detailed description

Opening this to get some early feedback and suggestions on the content. I've broken it down as such to give a bit of a intro to what a design system is, what value it brings, and how that relates to HDS.

- What is a design system?
  - Design system architecture/structure
  - Why use the HashiCorp Design System?
  - Our consumers
- Resources
  - Inspiration
  - Additional reading

### :link: External links

Jira ticket: [HDS-1066](https://hashicorp.atlassian.net/browse/HDS-1066?atlOrigin=eyJpIjoiN2Q4YmE0NTU3Y2QyNDA4OTk0NjA2ZjcyYmJhMDY3NTIiLCJwIjoiaiJ9)
Figma file: [if it applies]

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #903, opened 2022-12-20
About / Overview updates

jorytindall:
### :pushpin: Summary

If merged, this PR will update some parts of the prose and grammar in the `About / Overview` content.

This is a follow-on to PR #884 with based on team feedback.

### DIFF
--- a/website/docs/about/overview.md
+++ b/website/docs/about/overview.md
@@ -17,6 +17,7 @@ While the scale of a design system largely depends on the needs of those consumi
 
 - **Visual Language**: colors, typography, icons, imagery, content strategy
 - **Components**: component design, component development, tokens, development standards
+- **Patterns**: repetitive combinations of components and foundational styles that yield more complex functions and features
 - **Documentation**: education, issue tracking, contribution, content management, processes
 
 At HaschiCorp we structure our components, icons, and design tokens in a [monorepo](https://github.com/hashicorp/design-system) using yarn workspaces and publish a collection of libraries in [Figma](https://www.figma.com/files/team/1030156573400567478) for designers.
--- a/website/docs/about/overview.md
+++ b/website/docs/about/overview.md
@@ -22,7 +22,7 @@ While the scale of a design system largely depends on the needs of those consumi
 
 At HaschiCorp we structure our components, icons, and design tokens in a [monorepo](https://github.com/hashicorp/design-system) using yarn workspaces and publish a collection of libraries in [Figma](https://www.figma.com/files/team/1030156573400567478) for designers.
 
-### Why use the HashiCorp Design System?
+## Why use the HashiCorp Design System?
 
 While increasing visual and functional consistency accross products is the most outwardly notable result of design system adoption, the instrinsic value a design system brings can often result in a groundswell of benefits:
 
--- a/website/docs/about/overview.md
+++ b/website/docs/about/overview.md
@@ -11,7 +11,7 @@ A design system is a set of guidelines, standards, assets, and processes to help
 
 Design systems are often equated with a UI kit or component library; in reality an effective design system spans the entire lifecycle of a product through ideation to iteration and everything in between.
 
-### Architecture
+## Architecture
 
 While the scale of a design system largely depends on the needs of those consuming it, the fundamental structure and architecture remains largely the same across systems.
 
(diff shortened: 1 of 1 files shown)


website/docs/about/overview.md
```diff
@@ -11,28 +11,38 @@ A design system is a set of guidelines, standards, assets, and processes to help
...
 - Increased velocity in the design and development of new products and features
 - Dismantling of silos between departments that adopt and use the system
 - More time spent on solving complex problems, less time spent on duplicative work rebuilding small atoms and molecules
 - Components, foundations, and usage guidelines published by the the HashiCorp Design System meet or exceed WCAG 2.1 AA conformance out of the box in accordance with our [accessibility statement](/about/accessibility-statement).
 
-### Our consumers
+## Our consumers
 
-The HashiCorp Design System has been adopted at varying levels by all products within the R&D organization. We work closely with these teams at the integration and implementation level while collecting feedback to better understand consumer needs and inform our [roadmap](https://go.hashi.co/hds-rollout).
+HashiCorp develops and maintains a number of different products with varying sub-brands and approaches to the user experience. In an effort to bring these different product lines and teams towards a common visual language and single source of truth, the HashiCorp Design System was born.
```
jorytindall, line 36: @heatherlarsen here's the largest change in this PR, have a look at let me know what you think!

website/docs/about/overview.md
```diff
@@ -11,28 +11,38 @@ A design system is a set of guidelines, standards, assets, and processes to help
...
+
+Adoption of the HashiCorp Design System across teams is driven by a number of high-value user personas to identify convergence across products:
```
Dhaulagiri, line 38: While this section is interesting, my initial reaction is that I would not include it. It's a lot of detail and exposes some internal ideas that I'm not sure we necessarily want or need to.

website/docs/about/overview.md
```diff
@@ -11,28 +11,31 @@ A design system is a set of guidelines, standards, assets, and processes to help
...
 
 - **Visual Language**: colors, typography, icons, imagery, content strategy
 - **Components**: component design, component development, tokens, development standards
+- **Patterns**: repetitive combinations of components and foundational styles that yield more complex functions and features
 - **Documentation**: education, issue tracking, contribution, content management, processes
 
 At HaschiCorp we structure our components, icons, and design tokens in a [monorepo](https://github.com/hashicorp/design-system) using yarn workspaces and publish a collection of libraries in [Figma](https://www.figma.com/files/team/1030156573400567478) for designers.
 
-### Why use the HashiCorp Design System?
+## Why use the Helios Design System?
```
heatherlarsen, line 25: Rethinking this heading a bit. What if it's "Benefits of using Helios" instead?

website/docs/about/overview.md
```diff
@@ -11,28 +11,31 @@ A design system is a set of guidelines, standards, assets, and processes to help
...
-- Components, foundations, and usage guidelines published by the the HashiCorp Design System meet or exceed WCAG 2.1 AA conformance out of the box in accordance with our [accessibility statement](/about/accessibility-statement).
+- Components, foundations, and usage guidelines published by the the Helios Design System meet or exceed WCAG 2.1 AA conformance out of the box in accordance with our [accessibility statement](/about/accessibility-statement).
 
-### Our consumers
+## Our consumers
 
-The HashiCorp Design System has been adopted at varying levels by all products within the R&D organization. We work closely with these teams at the integration and implementation level while collecting feedback to better understand consumer needs and inform our [roadmap](https://go.hashi.co/hds-rollout).
+HashiCorp develops and maintains a number of different products with varying sub-brands and approaches to the user experience. In an effort to bring these different product lines and teams towards a common visual language and single source of truth, the Helios Design System was born.
```
heatherlarsen, line 36: @Dhaulagiri Was your earlier comment just about the personas or this section as well? 

Personally, I worry this isn't providing value to consumers about who can use the system and I'm not sure it will be received in a positive way, so it might be worth removing the whole paragraph.

heatherlarsen (commented):
One suggestion and one paragraph that I think we need to finesse before merging up.

website/docs/about/overview.md
Dhaulagiri: it was mostly about the personas. i don't think i mind this part of this section.

Dhaulagiri (approved):

PR #906, opened 2022-12-20
Adds missing `isFullWidth` to `copybutton` component

MelSumner:
### :pushpin: Summary

If merged, this PR adds the ability to indicate whether or not the `copybutton` component should take up the full width of the parent container.

I think this should have no visual effect on things as they are now (it sorts out an issue I'm having using the component elsewhere), but let's pay attention to visual diff for this one. 

- similar in the button component: https://github.com/hashicorp/design-system/blob/704e51e17dc3db57b43621840b95cfef64131874/packages/components/app/styles/components/button.scss#L68


KristinLBradley (approved):

didoo:
@MelSumner where is this used? seems I can't find it in the current code

PR #907, opened 2022-12-20
Resolves accessibility issues with toggle button in token list

MelSumner:
### :pushpin: Summary

If merged, this PR:

- adds a unique accessible name via `aria-label` to the toggle buttons in the token list.
- removes the `sr-only` span and adds `aria-expanded` to the button element itself

Note: FWIW, I don't think we need to include the toggle or the `type`, not really sure why those are there. Also, we could work some more on how the list content is present to the AOM. Right now it just says "generic." 😬 

Screenshot of Issue: 
CleanShot 2022-12-20 at 09 45 50

After fix: 

`aria-label`:
CleanShot 2022-12-20 at 09 48 47

`aria-expanded`: 
CleanShot 2022-12-20 at 10 00 34
CleanShot 2022-12-20 at 10 02 15

### DIFF
--- a/website/app/components/doc/tokens-list/item.hbs
+++ b/website/app/components/doc/tokens-list/item.hbs
@@ -18,7 +18,7 @@
     {{! template-lint-enable no-inline-styles }}
   </div>
   <div class="doc-tokens-list__content">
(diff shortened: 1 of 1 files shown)


alex-ju (approved):
Thanks for raising this so quickly, Melanie! 🙌  I made a note yesterday to [follow up on the token](https://github.com/hashicorp/design-system/pull/885#pullrequestreview-1222825926) work done so far, so I can cross that off the list now.

PR #911, opened 2022-12-20
Query param clean up

Dhaulagiri:
### :pushpin: Summary

Fixes a number of query-param related issues you can test on both `foundations/icons` and `foundations/tokens`

* Removes the trailing `/` from our routing so we don't need to manually remove it from query params
* Clearing search inputs also clears out query params and resets search results
* Adds tests for token search
* Reloading the page uses query param

[HDS-1255](https://hashicorp.atlassian.net/browse/HDS-1255)


MelSumner (approved):
Nice!! 👍

PR #913, opened 2022-12-20
Add component & foundation illustrations

MelSumner:
### :pushpin: Summary

* Implements title-inference naming of illustration images
* Disables fingerprinting (at least for now) for illustrations since this approach doesn't work with fingerprinting
* Adds all the images for components & foundations

**Note** There are in flight PRs that will resolve form illustrations that don't work correctly (they are being renamed)

https://hds-website-git-melsumner-update-component-art-hashicorp.vercel.app/components
https://hds-website-git-melsumner-update-component-art-hashicorp.vercel.app/foundations

Screen Shot 2022-12-23 at 09 29 22
Screen Shot 2022-12-23 at 09 29 24


website/app/components/doc/cards/card.hbs
```diff
@@ -1,7 +1,14 @@
 <li class={{this.classNames}} ...attributes>
   {{#if @route}}
     <LinkTo class="doc-cards-card__link" @route={{@route}} @model={{@model}}>
-      <img class="doc-cards-card__image" src={{@image}} alt="" role="presentation" />
+      <img
+        class="doc-cards-card__image"
+        src={{@image}}
+        alt={{(concat @title " - " @caption)}}
```
Dhaulagiri, line 7: @MelSumner this triggered a linting error so it seems like I should undo this. Can you confirm?

website/ember-cli-build.js
```diff
@@ -21,6 +21,8 @@ module.exports = function (defaults) {
     fingerprint: {
       // override defaults to also include json files which our markdown is compiled to. without this images don't render properly.
       replaceExtensions: ['html', 'css', 'js', 'json'],
+      exclude: ['illustrations/foundation', 'illustrations/component'],
+      enabled: true,
```
Dhaulagiri, line 25:
```suggestion

```

jorytindall (approved):
🚀

KristinLBradley (approved):

website/app/components/doc/cards/card.hbs
MelSumner: @Dhaulagiri what was the linting error? Might be a false positive, so I'd just add an inline ignore if so. But the way that we use most of the card images, they are superfluous anyway so I'd probably do an empty alt and a role of none.

Dhaulagiri: i recall that we discussed this at the time and i removed the alt tag entirely

PR #920, opened 2022-12-21
locationType: use history instead of deprecated auto

Dhaulagiri:
### :pushpin: Summary

https://deprecations.emberjs.com/v4.x/#toc_deprecate-auto-location


MelSumner (approved):
Oh good catch, I spaced on that! 👍

PR #922, opened 2022-12-21
Adds missing apostrophe in `dont` associated `label`

MelSumner:
### :pushpin: Summary

Updates the label for the `dont` case from `Dont` to `Don’t`.

CleanShot 2022-12-21 at 10 20 01@2x


Dhaulagiri (approved):
Thanks!

PR #926, opened 2022-12-21
update interactive docs

Dhaulagiri:
### :pushpin: Summary

This was tricky because it's not really meant for consumer usage, and it uses a lot of alerts which I did my best with those. Open to feedback on it but I'm also inclined to keep this one simple.

### DIFF
--- a/website/docs/utilities/interactive/partials/code/how-to-use.md
+++ b/website/docs/utilities/interactive/partials/code/how-to-use.md
@@ -1,4 +1,5 @@
 ## How to use this component
+
 ### Basic use (`<button>`)
 
 Invocation of the component would look like this:
@@ -20,7 +21,7 @@ In this case, since no `@href` or `@route` argument is provided it will generate
 
 !!! Critical
 
-We can't support the direct use of the `href` HTML attribute because we need to rely on the `@href` Ember argument to differentiate between different types of generated output.
+We can’t support the direct use of the `href` HTML attribute because we need to rely on the `@href` Ember argument to differentiate between different types of generated output.
 !!!
 
 If an `@href` argument is provided:
--- a/website/docs/utilities/interactive/partials/code/how-to-use.md
+++ b/website/docs/utilities/interactive/partials/code/how-to-use.md
@@ -1,6 +1,7 @@
-#### Basic use (`<button>`)
+## How to use this component
+### Basic use (`<button>`)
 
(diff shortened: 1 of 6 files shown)


website/docs/utilities/interactive/partials/code/how-to-use.md
```diff
@@ -1,6 +1,7 @@
-#### Basic use (`<button>`)
+## How to use this component
```
Dhaulagiri, line 1:
```suggestion
## How to use this component

```

website/docs/utilities/interactive/partials/guidelines/guidelines.md
```diff
@@ -0,0 +1,3 @@
+## Usage
```
Dhaulagiri, line 1: i wanted to have _some_ header here so the sidecar wouldn't be empty 🤷

Dhaulagiri: created a ticket to follow up on these. it'll be easy to adjust

website/docs/utilities/interactive/partials/code/how-to-use.md
```diff
@@ -10,11 +11,17 @@ Invocation of the component would look something like this:
...
-#### With "@href" parameter (`<a>`)
+`type="button"` HTML attribute is applied by default to the element, but this can be overwritten using the "splattributes".
+!!!
 
-**🚨 ATTENTION**: we can't support the direct use of the `href` HTML attribute because we need to rely on the `@href` Ember argument to differentiate between different types of generated output.
+### With `@href` parameter (`<a>`)
+
+!!! Critical
+
+We can't support the direct use of the `href` HTML attribute because we need to rely on the `@href` Ember argument to differentiate between different types of generated output.
```
MelSumner, line 23:
```suggestion
We can’t support the direct use of the `href` HTML attribute because we need to rely on the `@href` Ember argument to differentiate between different types of generated output.
```

MelSumner (approved):
It needs a few tweaks but after that it can ship. 👍

website/docs/utilities/interactive/partials/code/how-to-use.md
Dhaulagiri: 🤔  what is the difference between these characters? The one I have is the `'` key on my keyboard.

PR #931, opened 2022-12-21
Demo: markup for footer nav

MelSumner:
@Dhaulagiri this is what the footer markup should look like. I didn't change the CSS.


didoo:
@MelSumner implemented the changes here: https://github.com/hashicorp/design-system/pull/991

PR #934, opened 2022-12-21
link standalone content

Dhaulagiri:
### :pushpin: Summary

link standalone content

**Preview:** https://hds-website-git-br-link-stand-hashicorp.vercel.app/components/link/standalone

See https://github.com/hashicorp/design-system/pull/933 for similar updates to Link::Inline

### :link: External links

Jira ticket: [HDS-1281](https://hashicorp.atlassian.net/browse/HDS-1281)

PR #948, opened 2022-12-22
update title to link::standalone

MelSumner:
Update link::standalone to have human readable title

### DIFF
--- a/website/docs/components/link/standalone/index.md
+++ b/website/docs/components/link/standalone/index.md
@@ -1,5 +1,5 @@
 ---
-title: Link::Standalone
+title: Link (Standalone)
 description: A link used in isolation and not as a part of surrounding body text.
 caption: A link used in isolation and not as a part of surrounding body text.
 status: released


jorytindall (approved):

website/docs/components/link/standalone/index.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Link::Standalone
+title: Link (Standalone)
 description: A link used in isolation and not as a part of surrounding body text.
```
KristinLBradley, line 3: For this, I think the more human readable option would be "Standalone Link" with "Inline Link" for the other link type. Just my opinion though.

MelSumner: @KristinLBradley I thought about that- but then they wouldn't be next to each other in the list, which seems more user-friendly. WDYT?

Dhaulagiri: I would make the changes in the same PR. I think they'll still be grouped next to each other based on how the form controls are showing up.

MelSumner: @Dhaulagiri do you want to incorporate these changes into your existing PRs and I'll close these?

jorytindall: My two cents: I read this is <name> <type>, so Link (standalone) makes cents (😉) to me. Categorically it's a link with a variant

KristinLBradley: They would still be grouped under the "Link" category I would assume.

Dhaulagiri:
Will do in https://github.com/hashicorp/design-system/pull/934

PR #952, opened 2022-12-22
Actually run our template linting

Dhaulagiri:
### :pushpin: Summary

We have rules defined but they aren't actually running. This enables them and applies the necessary fixes.


website/.template-lintrc.js
```diff
@@ -1,5 +1,10 @@
 'use strict';
 
 module.exports = {
-  extends: 'recommended',
+  plugins: ['ember-template-lint-plugin-prettier'],
+  extends: ['recommended', 'ember-template-lint-plugin-prettier:recommended'],
+  rules: {
```
Dhaulagiri, line 6: i copied these from `packages/components` assuming we'd want them here as well

MelSumner (approved):
Thank you for doing this! 👍

PR #961, opened 2022-12-22
one weird trick

MelSumner:
### :pushpin: Summary

If merged, this PR _really_ fixes the copybutton that we thought we already fixed. Turns out we did need the quotes around the text value (which I don't actually understand why)

Before the change: 
CleanShot 2022-12-22 at 17 35 55@2x

After the change:
CleanShot 2022-12-22 at 17 37 00@2x


KristinLBradley (approved):

didoo:
@MelSumner reverted this changes in https://github.com/hashicorp/design-system/pull/982

PR #962, opened 2022-12-23
Review Form::Select Docs

jorytindall:
### :pushpin: Summary

Reviewing select docs!


website/docs/components/form/select/partials/code/how-to-use.md
```diff
@@ -175,9 +190,9 @@ If you want to connect one or more extra elements describing the field to the co
 </Hds::Form::Select::Field>
 ```
 
-##### HTML native attributes
+#### HTML native attributes
```
MelSumner, line 193:
```suggestion
#### Native HTML attributes
```

website/docs/components/form/select/partials/code/how-to-use.md
```diff
@@ -202,7 +217,7 @@ Similarly, you can pass HTML attributes to the `<option/optgroup>` elements.
 
 This can be useful in case you want to add specific native behaviors to the field, that are not exposed directly by the component (eg. providing a `name` for the control, or adding `multiple` and `size` attributes to it)
 
-##### Events handling
+#### Events handling
```
MelSumner, line 220:
```suggestion
#### Event handling
```

website/docs/components/form/select/partials/code/how-to-use.md
```diff
@@ -143,7 +155,7 @@ To show the user that their input is not valid, you have to do two things: decla
 </Hds::Form::Select::Field>
 ```
 
-##### Custom control ID
+#### Custom control ID
 
 In case it's necessary to have custom ID for the control, instead of the one automatically generated by the component (eg. because it needs to be referenced in the code for other reasons), you just need to pass a `@id` argument to the "field":
```
MelSumner, line 160:
```suggestion
In cases where it’s necessary to have custom id attribute for the control, instead of the one automatically generated by the component (e.g., because it needs to be referenced somewhere else in the code), you just need to pass a `@id` argument to the "field":
```

website/docs/components/form/select/partials/guidelines/guidelines.md
```diff
@@ -0,0 +1,108 @@
...
+
+### When to use
+
+- To allow user to make a selection from a pre-defined set of options in a form.
+- When the length of the form is a concern or screen real estate is vital.
+
+### When not to use
+
+- When a user can select more than on option, consider using [Checkbox](/components/form/checkbox/overview).
+- When the set of options is fewer than 5, consider using [Radio](/components/form/radio/overview).
```
MelSumner, line 11:
```suggestion
- When there are fewer than 5 options, consider using [Radio](/components/form/radio/overview) instead.
```

website/docs/components/form/select/partials/specifications/states.md
```diff
@@ -0,0 +1,49 @@
...
+![Select invalid with hover state](/assets/components/select/select-state-invalid-hover.png)
+
+Focus
+
+![Select invalid with focus state](/assets/components/select/select-state-invalid-focus.png)
+
+### Disabled
+
+- Disabled fields are not editable by the user.
+- If the user doesn't need to review the data, consider using a hidden field instead.
```
MelSumner, line 48: I would consider moving this section up to the previous disabled state section. 

I would also suggest adding the difference between Disabled and Readonly.

I also don't recommend hiding data that the user doens't need to review. How do we know it? If it is a form that the user is interacting with, they should at least see the disabled and readonly fields, for transparency.

jorytindall: - I've reorganized the disabled content
- I don't see a readonly variant for the select, only for the other textinput fields unless I'm missing something? Readonly is listed as a potential attribute that can be passed in `...attributes` though
- I agree with you on this, but I think this goes deeper than just removing it here. Let's log it away and revisit holistically for all of the form elements.

alex-ju (approved):

KristinLBradley (approved):

PR #971, opened 2022-12-23
Fix repetitive required field indicator for screen readers

alex-ju:
### :pushpin: Summary

Fix repetitive required field indicator for screen readers.

`@isRequired` applies both a visual indicator (a badge) and the `required` attribute on the associated form field, causing a repetitive 'required' when a screen reader user reaches the form element. We set `aria-hidden="true"` to remove the indicator from the a11y tree and rely on the native HTML attribute to communicate this to users.

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/orange-suits-guess.md
+++ b/.changeset/orange-suits-guess.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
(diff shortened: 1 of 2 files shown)


MelSumner (approved):
👍

PR #982, opened 2023-01-03
Improve fix for text in "CopyButton"

didoo:
### :pushpin: Summary

This reverts #961 and applies a fix at template level.

The fix done in #961 worked because was converting this condition:
https://github.com/hashicorp/design-system/blob/main/website/app/components/doc/color-swatch/index.js#L14
to a string (`"false"`) so the `CopyButton` component would not complain, but the user would copy `"false"` as CSS helper (exists only for the 'foreground', 'page', 'surface', 'border' groups, see: https://github.com/hashicorp/design-system/blob/85055c0be427826cae79f231d9758296f20439d4/website/docs/foundations/colors/index.js#L37-L40)

In this PR I have:
- reverted the #961 fix
- added a conditional check around the "css helper" block of code

***

### 👀 Reviewer's checklist:

### DIFF
--- a/website/app/components/doc/color-swatch/index.hbs
+++ b/website/app/components/doc/color-swatch/index.hbs
@@ -8,15 +8,15 @@
         <Doc::CopyButton @textToCopy={{this.cssVariable}} @type="ghost" />
 
       </li>
-      <li class="doc-color-swatch__listitem">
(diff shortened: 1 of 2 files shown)


Dhaulagiri (approved):
Nice! Thanks for circling back on the root cause here.

PR #995, opened 2023-01-03
`Pagination` - design tokens (HDS-971)

KristinLBradley:
### :pushpin: Summary
If merged, this PR adds component-level design tokens for Pagination

- This branch will be merged into the [PaginationBar feature branch](https://github.com/hashicorp/design-system/pull/661) once approved. (It will not be merged directly into the main branch.)

### :link: External links

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/app/styles/components/pagination.scss
+++ b/packages/components/app/styles/components/pagination.scss
@@ -4,17 +4,6 @@
 
 @use "../mixins/focus-ring" as *;
 
-// TODO: Move (Create component-level design tokens)
-:root {
-  // control - button-number, button-arrow:
-  --token-pagination-control-height: 36px;
-  --token-pagination-control-padding-horizontal: 11px;
-  --token-pagination-control-focus-inset: 4px;
-  --token-pagination-ellipsis-padding-horizontal: 12px;
-  // indicator for current page:
-  --token-pagination-indicator-height: 2px;
-}
-
 // * Pagination Parent (wrapper) component
 // ----------------------------------------
 .hds-pagination {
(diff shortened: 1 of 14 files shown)


didoo (commented):
some considerations (happy to discuss them in person):
- since the `control` is under the `nav`, maybe the tokens should be called `token-pagination-nav-control-***`?
- the only difference in horizontal padding between the `control` and the `ellipsis` is due to the `1px` transparent border: what do you think of unifying them under the `control` token (with the `12px` value) and in the CSS code do `calc(var(--token-pagination-control-padding-horizontal) - 1px)` and add a comment of why we remove 1px from the padding?
- since it's `0px`, do you think we need a token for the vertical padding of the `control/ellipsis`? what technically controls the size is the `height` value
- there is a `6px` margin between text and icon in the prev/next buttons, maybe it should be tokenized too?
- there is a `6px` left/right inset for the indicator, maybe tokenize it too?

KristinLBradley:
> some considerations (happy to discuss them in person):
> 
> * since the `control` is under the `nav`, maybe the tokens should be called `token-pagination-nav-control-***`?
> * the only difference in horizontal padding between the `control` and the `ellipsis` is due to the `1px` transparent border: what do you think of unifying them under the `control` token (with the `12px` value) and in the CSS code do `calc(var(--token-pagination-control-padding-horizontal) - 1px)` and add a comment of why we remove 1px from the padding?
> * since it's `0px`, do you think we need a token for the vertical padding of the `control/ellipsis`? (technically, what controls the size is the `height` value)
> * there is a `6px` margin between text and icon in the prev/next buttons, maybe it should be tokenized too?
> * there is a `6px` left/right inset for the indicator, maybe tokenize it too?

Thanks, I went ahead and made these updates.

didoo (approved):
👌

PR #1001, opened 2023-01-04
Setup meta tags

Dhaulagiri:
### :pushpin: Summary

Configures some basic meta tags for use in SEO and social media

Sets up meta tags for the followin

* Description: Caption from frontmatter where it exists otherwise generic blurbe
* Images
  * Caveat: we don't currently have an image for pages without an illustration, but once we have our logos sorted out we can add one
 * Title
   * This is _complicated_. We use ember-page-title for the `<title>` title as it does a better job building up the titles as we traverse route.  This PR we use the `title` defined by ember-meta for the title used in the `og` attributes but **not** the `<title>`. It's definitely possible there's a better way to do this, but it's a solution I was able to make work.

One can use a site like https://www.opengraph.xyz/ to test this out

#### Note

I copied in the `head.hbs` template from the addon and commented out the aspects of I didn't think applied for our use case. If you think otherwise leave a comment and we can sort it out.

Screen Shot 2023-01-04 at 14 25 19

### :link: External links

Jira ticket: [HDS-1114](https://hashicorp.atlassian.net/browse/HDS-1114)

[HDS-1114]: https://hashicorp.atlassian.net/browse/HDS-1114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/services/head-data.js
+++ b/website/app/services/head-data.js
@@ -1,24 +1,19 @@
 import HeadDataService from 'ember-meta/services/head-data';
+import config from 'ember-get-config';
 
 export default class CustomHeadDataService extends HeadDataService {
-  // this is needed or `head-data.js` will crush
-  get config() {
-    return {};
-  }
-
   get title() {
     return (
       this.currentRouteMeta?.title ??
       this.currentRouteMeta?.frontmatter?.title ??
-      'Helios Design System'
+      config['ember-meta'].title
     );
   }
 
-  get siteName() {
-    return 'Helios Design System';
-  }
-
   get description() {
-    return this.currentRouteMeta?.frontmatter?.description ?? null;
+    return (
+      this.currentRouteMeta?.frontmatter?.description ??
+      config['ember-meta'].description
+    );
   }
 }
--- a/website/app/services/head-data.js
+++ b/website/app/services/head-data.js
@@ -16,4 +16,13 @@ export default class CustomHeadDataService extends HeadDataService {
       config['ember-meta'].description
     );
   }
+
+  get imgSrc() {
+    return this.currentRouteMeta?.id
+      ? `/assets/illustrations/${this.currentRouteMeta?.id.replace(
+          /\/index$/,
+          ''
+        )}.jpg`
+      : config['ember-meta'].imgSrc;
+  }
 }
--- a/website/app/templates/head.hbs
+++ b/website/app/templates/head.hbs
@@ -6,12 +6,7 @@ We've made some small customizations to simplify this compared to the default te
 
 }}
 
-{{! Begin General }}
-
-{{! TODO: Determine how to avoid this conflict between ember-meta and ember-page-title where they both set a <title> which causes the latter addon to throw an exception }}
-{{!-- {{#if this.model.title}}
-  <title>{{this.model.title}}</title>
-{{/if}} --}}
+{{! BEGIN GENERAL }}
 
 {{#if this.model.description}}
   <meta name="description" content={{this.model.description}} />
@@ -19,43 +14,26 @@ We've made some small customizations to simplify this compared to the default te
 
 <meta name="referrer" content="unsafe-url" />
 
-{{#if this.model.canonical}}
-  <link rel="canonical" href={{this.model.canonical}} />
-{{else if this.model.url}}
+{{#if this.model.url}}
   <link rel="canonical" href={{this.model.url}} />
-{{/if}}
-{{! End General }}
-
-{{! Begin Article }}
-{{!-- {{#if this.model.date}}
-  <meta property="article:published_time" content={{this.model.date}} />
+  <meta property="og:url" content={{this.model.url}} />
 {{/if}}
 
-{{#each this.model.tags as |tag|}}
-  <meta property="article:tag" content={{tag}} />
-{{/each}} --}}
-{{! End Article }}
+{{! END GENERAL }}
 
-{{! Begin opengraph (Facebook, Slack, etc) }}
+{{! BEGIN OPENGRAPH }}
+
 {{#if this.model.siteName}}
   <meta property="og:site_name" content={{this.model.siteName}} />
 {{/if}}
 
-{{#if (or this.model.articleTitle this.model.title)}}
-  <meta property="og:title" content={{if this.model.articleTitle this.model.articleTitle this.model.title}} />
-{{/if}}
-
-{{#if this.model.url}}
-  <meta property="og:url" content={{this.model.url}} />
+{{#if this.model.title}}
+  <meta property="og:title" content={{this.model.title}} />
 {{/if}}
 
 {{#if this.model.description}}
   <meta property="og:description" content={{this.model.description}} />
 {{/if}}
-
-{{!-- {{#if this.model.type}}
-  <meta property="og:type" content={{this.model.type}} />
-{{/if}} --}}
 
 {{#if this.model.imgSrc}}
   <meta property="og:image" content={{this.model.imgSrc}} />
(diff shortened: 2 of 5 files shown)


website/app/services/head-data.js
```diff
@@ -1,24 +1,28 @@
...
+      this.currentRouteMeta?.frontmatter?.description ??
+      config['ember-meta'].description
+    );
   }
 
-  get description() {
-    return this.currentRouteMeta?.frontmatter?.description ?? null;
+  get imgSrc() {
+    return this.currentRouteMeta?.id
+      ? `/assets/illustrations/${this.currentRouteMeta?.id.replace(
```
Dhaulagiri, line 22: not super jazzed about replicating this logic here, but until we have a better solution for these I'm not sure there is another option?

didoo: If you have capacity, what do you think of opening another PRs that adds support for a `preview-image` property to the frontmatter attributes (as you suggested) and populate it there, so we avoid all of this logic in the code?

website/app/services/head-data.js
```diff
@@ -1,24 +1,28 @@
...
-  get config() {
-    return {};
-  }
-
   get title() {
     return (
       this.currentRouteMeta?.title ??
       this.currentRouteMeta?.frontmatter?.title ??
-      'Helios Design System'
+      config['ember-meta'].title
```
didoo, line 9: still not a big fan of using the ember config this way (not sure what is the added value), but I can live with that :)

website/app/templates/head.hbs
```diff
@@ -0,0 +1,90 @@
...
+{{! Begin Article }}
+{{!-- {{#if this.model.date}}
+  <meta property="article:published_time" content={{this.model.date}} />
+{{/if}}
+
+{{#each this.model.tags as |tag|}}
+  <meta property="article:tag" content={{tag}} />
+{{/each}} --}}
+{{! End Article }}
+
```
didoo, line 38: Why leave them here?

```suggestion
```

website/app/templates/head.hbs
```diff
@@ -0,0 +1,90 @@
...
+{{! Begin opengraph (Facebook, Slack, etc) }}
+{{#if this.model.siteName}}
+  <meta property="og:site_name" content={{this.model.siteName}} />
+{{/if}}
+
+{{#if (or this.model.articleTitle this.model.title)}}
+  <meta property="og:title" content={{if this.model.articleTitle this.model.articleTitle this.model.title}} />
+{{/if}}
```
didoo, line 46: For consistency with lines 22-26 (and better readability)

```suggestion
{{#if this.model.articleTitle}}
  <meta property="og:title" content={{this.model.articleTitle}} />
{{else if this.model.title}}
  <meta property="og:title" content={{this.model.title}} />
{{/if}}
```

website/app/templates/head.hbs
```diff
@@ -0,0 +1,90 @@
...
+We've made some small customizations to simplify this compared to the default template
+
+}}
+
+{{! Begin General }}
+
+{{! TODO: Determine how to avoid this conflict between ember-meta and ember-page-title where they both set a <title> which causes the latter addon to throw an exception }}
+{{!-- {{#if this.model.title}}
+  <title>{{this.model.title}}</title>
+{{/if}} --}}
```
didoo, line 14: This (lines 12-14) is causing issues in my local linter, are you sure is correct?

website/app/templates/head.hbs
```diff
@@ -0,0 +1,90 @@
+{{!
+
+Per https://github.com/shipshapecode/ember-meta#defining-your-own-headhbs this is a place where we can customize what information is injected by the ember-meta addon
+
+We've made some small customizations to simplify this compared to the default template
+
+}}
+
+{{! Begin General }}
```
didoo, line 9: I find it hard to see when things begin and end, with this simple comments. can you maybe make them move evident (eg. ALL UPPERCASE) and separate them more from the code?

website/app/templates/head.hbs
```diff
@@ -0,0 +1,90 @@
...
+{{!-- {{#if this.model.twitterUsername}}
+  <meta name="twitter:site" content={{this.model.twitterUsername}} />
+  <meta name="twitter:creator" content={{this.model.twitterUsername}} />
+{{/if}} --}}
+
+{{#if this.model.imgSrc}}
+  <meta name="twitter:image" content={{this.model.imgSrc}} />
+{{/if}}
+
+{{#if (or this.model.articleTitle this.model.title)}}
```
didoo, line 80: see suggestion above about simpler code

website/app/templates/head.hbs
```diff
@@ -0,0 +1,90 @@
...
+  <title>{{this.model.title}}</title>
+{{/if}} --}}
+
+{{#if this.model.description}}
+  <meta name="description" content={{this.model.description}} />
+{{/if}}
+
+<meta name="referrer" content="unsafe-url" />
+
+{{#if this.model.canonical}}
```
didoo, line 22: where is this `canonical` value coming from? I can't see it in the code (apart from here) so my understanding is that it's never a real case

website/app/templates/head.hbs
didoo: Also, where is this `articleTitle ` value coming from? I can't see it in the code (apart from here) so my understanding is that it's never a real case

website/app/templates/head.hbs
```diff
@@ -0,0 +1,90 @@
...
+{{/if}}
+
+{{#if this.model.description}}
+  <meta property="og:description" content={{this.model.description}} />
+{{/if}}
+
+{{!-- {{#if this.model.type}}
+  <meta property="og:type" content={{this.model.type}} />
+{{/if}} --}}
+
```
didoo, line 59:
```suggestion
```

didoo (commented):
Added a bunch of comments. Happy to discuss them in person, if you want.

website/app/templates/head.hbs
Dhaulagiri: in our case this logic relies on `this.model.title` existing, so we could update it to remove the `articleTitle` references completely.

Dhaulagiri:
@didoo in reading your feedback it seems like it might be more useful to simply remove all the parts of the template we are not using to reduce confusion, knowing we can always go back and reference the default template in the addons code via the link in the comment?

website/app/templates/head.hbs
Dhaulagiri: I'm not. For some reason the linting on the file is behaving quite strangely. Perhaps once I clean it up this will resolve itself.

website/app/templates/head.hbs
didoo: Yes, that's what I meant

didoo:
> @didoo in reading your feedback it seems like it might be more useful to simply remove all the parts of the template we are not using to reduce confusion, knowing we can always go back and reference the default template in the addon's code via the link in the comment?

Yes, agree

website/app/templates/head.hbs
```diff
@@ -0,0 +1,67 @@
...
+{{#if this.model.description}}
+  <meta property="og:description" content={{this.model.description}} />
+{{/if}}
+
+{{#if this.model.imgSrc}}
+  <meta property="og:image" content={{this.model.imgSrc}} />
+  <meta property="og:image:type" content="image/png" />
+  <meta property="og:image:width" content="256" />
+  <meta property="og:image:height" content="256" />
```
didoo, line 42: are these required? because the images we're using in the cards (and that end up used here) don't have this size, and expecially this 1:1 ratio)

didoo (approved):
left only one minor comment

website/app/templates/head.hbs
Dhaulagiri: even though this is the wrong aspect ratio this seems to be required to get these images to appear on facebook for example. they do appear in a larger size at a correct aspect ratio though 🤷

PR #1009, opened 2023-01-05
New docs website / Refactored logic for `showSidebarOnSmallViewport`

didoo:
### :pushpin: Summary

This PR refactors the logic of the `showSidebarOnSmallViewport` that controls when the sidebar is visible or not (from the JS side, the CSS part will be handled in another set of PRs).

The reason for this, is that currently in mobile view, when a user opens the sidebar via the burger menu and clicks one of the links inside it, the sidebar remains open while the preferable behaviour would be to automatically close (similar to what Developer does, for reference)

In this PR I have
- unified the logic `showSidebarOnSmallViewport` and `burgerMenuOpen` in a single place (`application.js`)
  - in doing so, I've removed the `index.js` file (not needed anymore; maybe this will solve also the problem with the error page receiving an `index` class in production environment?)
- added reset of the status of the burger menu and sidebar visibily on route change

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-1258
Related PR: [New docs website / Fix <body> overflow + Sidebar animation/transition/position/display - #1010](https://github.com/hashicorp/design-system/pull/1010)

***

### 👀 Reviewer's checklist:


website/app/controllers/application.js
```diff
@@ -1,15 +1,44 @@
...
+
+  routeDidChange() {
+    scheduleOnce('afterRender', this, this.onToggleBurgerMenu);
+  }
+
+  willDestroy() {
+    this.router.off('routeDidChange', this, 'routeDidChange');
+  }
+
+  get currentTopRoute() {
```
Dhaulagiri, line 27: it doesn't look like this behavior is working any longer. if you navigate to `/components` the class `is-current-route` isn't present where it is in production.

It may be possible to rework this logic to leverage the `current-when` behavior i was investigating yesterday but i'm not sure

Dhaulagiri:
I think there's a small bug:
* Narrow your viewport
* Go to a page with multiple tabs
* Sidebar is closed
* Click on a tab other than the currently active tab
* Sidebar opens when it shouldn't

website/app/controllers/application.js
```diff
@@ -1,15 +1,44 @@
...
-  // TODO add a reset for the `showSidebarOnSmallViewport` on route change
+  // notice: don't set it to false, otherwise the sidebar menu will appear open on page load
+  @tracked isSidebarVisibleOnSmallViewport;
+
+  constructor() {
+    super(...arguments);
+    this.router.on('routeDidChange', this, 'routeDidChange');
+  }
+
+  routeDidChange() {
```
Dhaulagiri, line 19: I think this is also getting triggered on query param "transition", hence toggling the sidebar back and forth per my comment on the main thread

website/app/controllers/application.js
didoo: fixed in https://github.com/hashicorp/design-system/pull/1009/commits/3bf4b96c9fb11a8cca40a82e00be735cb4db198b

Dhaulagiri (approved):

didoo:
@Dhaulagiri I am going to merge this one.

PR #1010, opened 2023-01-05
New docs website / Fix `<body>` overflow + Sidebar animation/transition/position/display

didoo:
### :pushpin: Summary

This PR refactors the way in which the scrolling overflow is applied to the page `<body>`, so that there are no more cases when the user can't scroll (if they open the sidebar in mobile mode, and then resize the window going to a larger viewport).

In the process, I have also completely refactored how the sidebar is positioned /displayed and animated/transitioned, at different viewports and on different pages (homepage vs other pages). 

This refactor fixes multiple issues raised (see Jira tickets below).

In this PR I have:
- refactored how the scrolling overflow is applied to the page `<body>`, using a specific class names instead on an inline style
- refactored of how the sidebar is positioned /displayed and animater/transitioned, at different viewports and on different pages (homepage vs other pages)
- changed header and sidebar to fixed position to avoid overscoll effects

👉 👉 👉 **Preview:** https://hds-website-git-new-docs-website-fix-body-overflow-hashicorp.vercel.app/

### :link: External links

Jira tickets:
- https://hashicorp.atlassian.net/browse/HDS-1301
- https://hashicorp.atlassian.net/browse/HDS-1306
- https://hashicorp.atlassian.net/browse/HDS-1307
- https://hashicorp.atlassian.net/browse/HDS-1310
- https://hashicorp.atlassian.net/browse/HDS-1311

***

### 👀 Reviewer's checklist:


KristinLBradley:
I wonder if we could instead detect when the sidebar becomes "stuck" as in this article:
https://css-tricks.com/how-to-detect-when-a-sticky-element-gets-pinned/

didoo:
> I wonder if we could instead detect when the sidebar becomes "stuck" as in this article: https://css-tricks.com/how-to-detect-when-a-sticky-element-gets-pinned/

@KristinLBradley mmm... interesting! How would you do it?

Dhaulagiri:
> so that there are no more cases when the user can't scroll (if they open the sidebar in mobile mode, and then resize the window going to a larger viewport).

In the preview app I'm still unable to scroll the main body if the sidebar is open which is I think slightly different than the original bug, but as a user I would expect to be able to scroll that content underneath the sidebar. What do you think?

website/app/controllers/application.js
```diff
@@ -28,17 +29,52 @@ export default class ApplicationController extends Controller {
     return this.router.currentURL.split('/')[1];
   }
 
+  resetSidebar() {
+    this.isSidebarVisibleOnSmallViewport = false;
+    if (!this.fastboot.isFastBoot) {
+      document.body.classList.remove('isSidebarVisibleOnSmallViewport');
+      document.body.classList.remove('isSidebarInRenderTree');
+      if (this.runNext) {
```
Dhaulagiri, line 37: maybe we could pair on this but there should be a more straightforward way to interact with the run loop without needing to do this sort of thing manually

Dhaulagiri:
Related to my previous comment, you also can't scroll the contents of the sidebar when it's open

KristinLBradley:
> > I wonder if we could instead detect when the sidebar becomes "stuck" as in this article: https://css-tricks.com/how-to-detect-when-a-sticky-element-gets-pinned/
> 
> @KristinLBradley mmm... interesting! How would you do it?

The article seemed relevant to the behavior you were trying to fix although I don't have an exact idea of how or whether we should implement something like that.

website/app/controllers/application.js
```diff
@@ -28,17 +29,52 @@ export default class ApplicationController extends Controller {
...
+  }
+
+  showSidebarOnSmallViewport() {
+    if (!this.fastboot.isFastBoot) {
+      document.body.classList.add('isSidebarInRenderTree');
+      if (this.runLater) {
+        cancel(this.runLater);
+      }
+      this.runNext = next(() => {
+        document.body.classList.add('isSidebarVisibleOnSmallViewport');
```
KristinLBradley, line 53: Super nit: To me, the class name seems a little hard to understand. Would it make sense to use the wording from the function name instead?

IE, use: `document.body.classList.add('showSidebarOnSmallViewport');`

website/app/styles/doc-components/table-of-contents.scss
```diff
@@ -23,6 +23,10 @@ $doc-toc-item-inset: 8px;
       background-color: var(--doc-color-gray-500);
       content: "";
     }
+    // last one doesn't have the separator
+    &:last-child::after {
+      content: none;
+    }
```
KristinLBradley, line 29: This is fine but I usually do something like below to avoid having to override the style:

```
.doc-table-of-contents__item--depth-1 + .doc-table-of-contents__item--depth-1 {
   /* in-between border styles here (using ::before vs. ::after) */
}
```

website/app/styles/pages/application/sidebar.scss
```diff
@@ -115,3 +93,64 @@ body.application:not(.index) {
...
+    // when the `isSidebarInRenderTree` class is added to the `<body>` (via JS) we add the sidebar to the render tree (so we can then animate it)
+    &.isSidebarInRenderTree .doc-page-sidebar {
+      display: flex;
+      transform: translate3d(-100%, 0, 0);
+      transition: transform 0.25s ease-in;
+    }
+
+    &.isSidebarVisibleOnSmallViewport .doc-page-sidebar {
+      transform: translate3d(0, 0, 0);
+      transition: transform 0.25s ease-out;
```
KristinLBradley, line 148: Nit: I would try to avoid repeating these styles maybe by using a sass variable or a mixin or something.

KristinLBradley (approved):
Added a few minor comments

website/app/controllers/application.js
Dhaulagiri: > so we need to use two "ticks" in the loop

My point with the comment is that my impression/recollection is that there is (or should be) a cleaner way to accomplish this in Ember than manually tracking runloop events like this.

Dhaulagiri (approved):
Just to be clear, my comment about the runloop was not a blocker either.

My personal preference would be for the content underneath the sidebar to be scrollable while it's open but I'm open to that being a personal preference.

website/app/styles/doc-components/table-of-contents.scss
didoo: yeah, I found I simpler way to do it using your suggestion, thanks

didoo:
> Just to be clear, my comment about the runloop was not a blocker either.

Yes. Also, I don't think there's another way :)
 
> My personal preference would be for the content underneath the sidebar to be scrollable while it's open 

It was a specific design requirement (and I agree with it).

PR #1011, opened 2023-01-05
add fathom analytics

Dhaulagiri:
### :pushpin: Summary

Sets up basic analytics tracking. For now it's going to track events across our "production" site and preview apps which is not ideal but will require further investigation. It should not track local events.

I'll share a more detailed demo in person

Screen Shot 2023-01-05 at 09 37 32

Jira ticket: [HDS-932](https://hashicorp.atlassian.net/browse/HDS-932)

[HDS-932]: https://hashicorp.atlassian.net/browse/HDS-932?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/index.html
+++ b/website/app/index.html
@@ -13,10 +13,6 @@
     <link rel="shortcut icon" href="/assets/logos/favicon.png" />
 
     {{content-for "head-footer"}}
-
-    <!-- Fathom - beautiful, simple website analytics -->
-    <script src="https://cdn.usefathom.com/script.js" data-spa="auto" data-site="PPQLJJKK" data-excluded-domains="localhost" defer></script>
-    <!-- / Fathom -->
   </head>
   <body>
     {{content-for "body"}}
--- a/website/app/index.html
+++ b/website/app/index.html
@@ -13,6 +13,10 @@
     <link rel="shortcut icon" href="/assets/logos/favicon.png" />
 
     {{content-for "head-footer"}}
+
+    <!-- Fathom - beautiful, simple website analytics -->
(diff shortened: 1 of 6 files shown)


website/app/index.html
```diff
@@ -13,6 +13,10 @@
     <link rel="shortcut icon" href="/assets/logos/favicon.png" />
 
     {{content-for "head-footer"}}
+
+    <!-- Fathom - beautiful, simple website analytics -->
```
didoo, line 17: @Dhaulagiri do you think we can use an approach like this:
https://github.com/hashicorp/design-system/blob/main/packages/ember-flight-icons/index.js
in which we have a custom `content-for` and conditionally we check, where we define it, if to add the code snippet or not, depending on the ENV variable (that should know if we are in production or not)

Dhaulagiri: good idea. let me take a look at that. there's supposed to be a built in way for fathom to do this but a) our domain doesn't exist yet and b) even when it does it feels harder than it should be

Dhaulagiri: I did refactor to only inject this if a `FATHOM_ENABLED` environment variable is set, which we could set _only_ on the main production build. Let me know what you think.

didoo (approved):
👍

website/app/index.html
alex-ju: Do we want to drop a line in the readme or somewhere to document this variable? `ember-cli-dotenv` seem to be throwing a warning if not set.

Dhaulagiri: Yes, I have an open followup to add some documentation.

I'd also like to see if this warning can not happen since it is not useful

alex-ju: 🙌

PR #1015, opened 2023-01-05
refactor how illustrations are referenced

Dhaulagiri:
### :pushpin: Summary

This PR refactors how we reference illustrations, changing from an file name inference to an explicit reference in the frontmatter.

The main reasons for this change are twofold
* Makes these reference more resilient by not having them depend on the name of a component which _could_ change
* By explicitly referencing the files like this we can reenable their fingerprinting

Test this works by checking  these links for correct illustrations
* [Components](https://hds-website-git-br-preview-hashicorp.vercel.app/components)
* [Foundations](https://hds-website-git-br-preview-hashicorp.vercel.app/foundations)
* Inspect the source of an [individual component](https://hds-website-git-br-preview-hashicorp.vercel.app/components/alert) and confirm you see a value like `<meta property="og:image" content="assets/illustrations/components/alert-dc12b37ebbc20cfbc2a1ff618c20e1c4.jpg">`

### :link: External links

Jira ticket: [HDS-1367](https://hashicorp.atlassian.net/browse/HDS-1367)

[HDS-1367]: https://hashicorp.atlassian.net/browse/HDS-1367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/services/head-data.js
+++ b/website/app/services/head-data.js
@@ -18,11 +18,8 @@ export default class CustomHeadDataService extends HeadDataService {
   }
 
   get imgSrc() {
-    return this.currentRouteMeta?.id
-      ? `/assets/illustrations/${this.currentRouteMeta?.id.replace(
-          /\/index$/,
-          ''
-        )}.jpg`
+    return this.currentRouteMeta?.frontmatter?.previewImage
+      ? this.currentRouteMeta.frontmatter.previewImage
       : config['ember-meta'].imgSrc;
   }
 }
--- a/website/docs/components/alert/index.md
+++ b/website/docs/components/alert/index.md
@@ -6,7 +6,7 @@ status: released
 links:
   figma: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components?node-id=1377%3A11987
   github: https://github.com/hashicorp/design-system/tree/main/packages/components/addon/components/hds/alert
-previewImage: illustrations/components/alert.jpg
+previewImage: assets/illustrations/components/alert.jpg
 ---
 
 <section data-tab="Guidelines">
--- a/website/docs/components/alert/index.md
+++ b/website/docs/components/alert/index.md
@@ -6,6 +6,7 @@ status: released
 links:
   figma: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components?node-id=1377%3A11987
   github: https://github.com/hashicorp/design-system/tree/main/packages/components/addon/components/hds/alert
+previewImage: illustrations/components/alert.jpg
 ---
 
 <section data-tab="Guidelines">
--- a/website/docs/foundations/border/index.md
+++ b/website/docs/foundations/border/index.md
@@ -2,6 +2,7 @@
 title: Border
 caption: Borders define and separate content visually. They can be used alone or in combination with elevation.
 description: Borders define and separate content visually. They can be used alone or in combination with elevation.
+previewImage: assets/illustrations/foundations/border.jpg
 ---
 
 <section data-tab="Guidelines">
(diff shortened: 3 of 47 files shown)


website/docs/components/alert/index.md
```diff
@@ -6,6 +6,7 @@ status: released
 links:
   figma: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components?node-id=1377%3A11987
   github: https://github.com/hashicorp/design-system/tree/main/packages/components/addon/components/hds/alert
+image: /assets/illustrations/components/alert.jpg
```
Dhaulagiri, line 9: @didoo before I get too far along does this match what you were thinking?

Next steps would be to update the frontmatter for all components and then do all these same adjustments for foundations as well.

Dhaulagiri: > maybe we could call the field preview-image

I initially tried that but got confused how to plumb something with a hyphen it through to `pageAttributes`

 > can you check what parts of the path can be omitted
 
unfortunately the full path seems to be required, but i can do a little bit more digging on that

didoo: > I initially tried that but got confused how to plumb something with a hyphen it through to `pageAttributes`

if you mean here `image: page.pageAttributes.image`, this should work _I think_: `image: page.pageAttributes['preview-image']`

> > can you check what parts of the path can be omitted

for the time being, let's leave it as is (`/assets/illustrations/***`) we will investigate later if we find issues

Dhaulagiri: Given (most?) everything else in the Ember parts of the codebase is camelCase it seems worth being consistent the yaml definitions as well and doing something like `previewImage`. Does that work for you?

website/app/services/head-data.js
```diff
@@ -18,11 +18,8 @@ export default class CustomHeadDataService extends HeadDataService {
   }
 
   get imgSrc() {
-    return this.currentRouteMeta?.id
-      ? `/assets/illustrations/${this.currentRouteMeta?.id.replace(
-          /\/index$/,
-          ''
-        )}.jpg`
+    return this.currentRouteMeta?.frontmatter
+      ? this.currentRouteMeta.frontmatter.previewImage
```
didoo, line 22: what if the `frontmatter` key exists, but the `previewImage` value is not defined? in that case the image is undefined. maybe we should do something like this?

```suggestion
    return this.currentRouteMeta?.frontmatter?.previewImage
      ? this.currentRouteMeta.frontmatter.previewImage
```

website/docs/foundations/border/index.md
```diff
@@ -2,6 +2,7 @@
 title: Border
 caption: Borders define and separate content visually. They can be used alone or in combination with elevation.
 description: Borders define and separate content visually. They can be used alone or in combination with elevation.
+previewImage: assets/illustrations/foundations/elevation.jpg
```
didoo, line 5:
```suggestion
previewImage: assets/illustrations/foundations/border.jpg
```

wiki/Website-Doc-folder.md
```diff
@@ -137,6 +138,8 @@ The "frontmatter" attributes that we support are the following:
     Used to control the order of the pages in navigational lists (lower value moves up the page, higher value moves it down) - default is `100`
 *   `hidden`
     Used to hide the page from the sidebar navigation and the lists on the landing pages - default is `false`
+*   `previewImage`
+  An optional full path to an image used when listing the page as "card" (eg. in landing pages).
```
didoo, line 142:
```suggestion
  An optional full path to an image used when listing the page as "card" (eg. in landing pages). The path refers to the `dist` folder generated at build time, so is relative to the content of the `/website/public` folder.
```

didoo (approved):
One of the paths was wrong. Apart from that, all good (left a few suggestions)

website/docs/foundations/border/index.md
Dhaulagiri: good catch!

PR #1020, opened 2023-01-06
`Pagination` - API refactoring

didoo:
### :pushpin: Summary

This PR collects all the work around the refactoring of the `Pagination` component API to make it more simple to adopt in the consumers codebase (see linked tickets).

For more details about why this change is needed, see the "More context" section below.

See commits for details.

Preview: https://hds-components-git-hds-972-pagination-cloud-ui-c491d7-hashicorp.vercel.app/components/pagination

### 🚧 Things to do:
- [x] how do we pass `@route/@model/@query` params to the numbered pages? 
   - after meeting and discussion with ambassadors, we've decided to try to explore this idea
   - this will be done in a follow-up PR (is not trivial)
- [x] ~~fix all the tests that are broken~~
   - ~~this will be done in a follow-up PR (to avoid too much noise)~~ 
   - https://github.com/hashicorp/design-system/pull/1050

### 🖇 Related tickets/PRs
- https://github.com/hashicorp/design-system/pull/661
- https://github.com/hashicorp/design-system/pull/1038
- https://github.com/hashicorp/design-system/pull/1050

Relevant thread
- https://hashicorp.slack.com/archives/C01KCU4HDPY/p1673471011040269

### 📒 More context

The problem that we're trying to solve with this PR arose once we tried to adopt the component (after #661) in Cloud UI, and we realized that our API (based on the idea of buttons triggering events) was not compatible with how the pagination was consumed in Cloud UI (based on `@route/@query` params provided to the controls).

So we had to find a way to "expose" the internals of the navigation controls, to the outside consuming codebase. We had different options, but we ended up deciding to use the contextual components within the `Nav` element.

screenshot_2201

This lead to this syntax:

```
  <Hds::Pagination as |P|>
    <P.Nav as |N|>
      <N.ButtonPrev @route="..." @query={{hash prev="previousPage" next=undefined}} />
      <N.ButtonNext @route="..." @query={{hash prev=undefined next="nextPage"}} />
    </P.Nav>
  </Hds::Pagination>
```

With this first attempt to allow passing dow the link parameters to the prev/next controls, we partially solved the problem, but not completely. How could we pass the routing parameters to the controls for the "page numbers" too? Plus, the resulting code (and components API) was quite convoluted at times (a lot of indirection and arguments passed around).
Plus, for the numbered pagination the parameters were passed to both the container as well as the contextual child components:

```
<Hds::Pagination @totalItems={{100}} @itemsPerPage={{30}} @currentPage={{3}} as |P|>
  <P.Info @showTotalItems={{false}} />
  <P.Nav @hideLabels={{false}} @isTruncated={{true}} />
  <P.SizeSelector @sizes={{array 10 30 50}} />
</Hds::Pagination>
```

The reason for this is that the previous implementation was trying to do too much in having a single pagination component handling two very different things (pagination with a known number of items, and with an unknown number of items) under the single component/code implementation.

So we decided to take another approach and opened a new PR (#1038) proposing to split the code in two different components:
- `Hds::Pagination::Numbered` - for use cases when the number of items is known, and this handles all the complexity of the pagination information shown to the user ("info", "page numbers", "current page", "page size") 
- `Hds::Pagination::Compact` - for use cases when the number of items is unknown, so it's impossible to have the "info", the "page numbers" and the "select-size" elements, and there is no concept of "current page"

This allowed us to bring the content of the `Pagination::Nav` directly in the `Compact` and `Numbered` component, removing one level of indirection, and make the `Pagination::Nav` not needed anymore. Same for the `Pagination` hybrid component, now replaced by two distinct/specialized components.

After discussing this approach in a dedicated meeting with the HDS ambassadors, we all came to the conclusion that the new proposed API (and overall approach) made sense, and was the right solution to the problem.

During the discussion the need to have "links" also for the "Numbered" pagination came up too (but this will be tackled in a follow-up PR).

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/app/styles/components/pagination.scss
```diff
@@ -49,6 +49,7 @@
 .hds-pagination-nav__control {
   // border width is subtracted from padding:
   padding: 0 calc(var(--token-pagination-nav-control-padding-horizontal) - 1px);
+  color: var(--token-color-foreground-primary);
```
KristinLBradley, line 52: I committed directly to this PR to add this line. I think I may have had the color as a CSS class previously and then deleted it accidentally in a later refactor? (Not sure.)

KristinLBradley:
For things to discuss #1:

To make the ButtonArrow components disabled in my test code, I used conditional logic so that an href value would not be provided. This makes the component render as an html button vs. an anchor tag so when the "disabled" attribute is applied it's valid.

```
<Hds::Pagination::Nav::ButtonArrow
    @direction="prev"
    @type={{this.type}}
    @disabled={{this.isDisabledPrev}}
    {{! Default - Uses button with onClick: }}
    @onClick={{unless @model (this.onPageChange)}}
    {{! Optional - Uses model with url query & route: }}
    @query={{if (and @model @model.previousPageToken) (hash next=undefined prev=@model.previousPageToken)}}
    @route={{if (and @model @model.previousPageToken) this.currentRouteName}}
/>
```

The syntax is a bit complex though.

KristinLBradley (commented):
(I'm going through individual commits, this feedback is one the first one)

The changes to "numbered", "compact", & "truncated" types seems simpler and more elegant and it allows some of the logic such as in the hbs to be simplified, so I'm happy with this update.

packages/components/addon/components/hds/pagination/nav/index.hbs
```diff
@@ -1,7 +1,7 @@
 <nav class="hds-pagination-nav" aria-label="Pagination navigation" ...attributes>
   <Hds::Pagination::Nav::ButtonArrow
     @direction="prev"
-    @type={{this.type}}
+    @hideLabel={{eq this.type "numbered"}}
```
KristinLBradley: Thanks, this is a good update. The `type` I originally used here was actually tripping me up too since HTML buttons have their own `type` attribute.

packages/components/app/styles/components/pagination.scss
```diff
@@ -70,10 +70,18 @@
...
+  // but from the product perspective there may be use cases where they need to have a the control as a link that looks disabled anyway
+
   &:disabled,
+  &[disabled],
   &.mock-disabled {
     color: var(--token-color-foreground-disabled);
     cursor: not-allowed;
+    // TODO discuss with Kristin if we want to do it (removes the cursor too!)
+    pointer-events: none;
   }
```
KristinLBradley, line 86: I added a comment to the PR with my test code as an example. I would prefer a solution which doesn't add an invalid attribute (the `disabled` attribute) to anchor link tags.

(Cloud-ui's Pagination also renders a button tag with the `disabled` attribute when the Prev & Next controls are disabled.)

KristinLBradley (commented):
Looks good! Overall I agree with all your changes, the only thing I would change is how the disabled buttons are handled but that's relatively minor.

KristinLBradley (approved):

didoo:
> Looks good! Overall I agree with all your changes, the only thing I would change is how the disabled buttons are handled but that's relatively minor.

@KristinLBradley yes, you're right. I've changed the `ButtonArrow` control implementation. It's a bit move verbose, but it doesn't generate invalid code. See this commit:
https://github.com/hashicorp/design-system/pull/1020/commits/41d7c55c979a12757d589f3148145d1d9fa4986f

I've also fixed the tests too (I forgot 🤦‍♂️).

packages/components/addon/components/hds/pagination/nav/index.hbs
```diff
@@ -1,10 +1,16 @@
+{{! template-lint-disable simple-unless }}
 <nav class="hds-pagination-nav" aria-label="Pagination navigation" ...attributes>
-  <Hds::Pagination::Nav::ButtonArrow
-    @direction="prev"
-    @hideLabel={{eq this.type "numbered"}}
-    @onClick={{this.onPageChange}}
-    @disabled={{this.isDisabledPrev}}
-  />
+  {{! TODO use `has-block-params`? https://api.emberjs.com/ember/4.4/classes/Ember.Templates.helpers/methods/action?anchor=has-block-params }}
+  {{#unless (has-block)}}
```
alex-ju, line 4: I think `has-block` is what we want here, as this would cover the cases where we have contextual components (`<N.ButtonPrev` and `N.ButtonNext`).

`<Hds::Pagination::Nav as |N|>` would return true for `has-block-params` but so will`<Hds::Pagination::Nav @totalPages={{10}} @currentPage={{3}} />`, so it's not probably what we want.

packages/components/app/styles/components/pagination.scss
alex-ju: Not sure if it helps with the matter, but in my experience `<a role="link" aria-disabled="true">Link</a>` is one of the few solutions that convey a disabled link to screen reader users.

packages/components/app/styles/components/pagination.scss
```diff
@@ -70,23 +70,19 @@
...
+}
 
-  // IMPORTANT: we need to use also the [disabled] selector because if the "disabled" attribute is applied to the control that,
-  // since internally it's using the `Hds::Interactive` component, it can be a link or a button, depending on the parameters provided, but
-  // the ":disabled" pseudo-selector is not applied to the element in the browser (rightly) because a link can't be disabled
-  // but from the product perspective there may be use cases where they need to have a the control as a link that looks disabled anyway
-
+// only the "Previous" & "Next" (`ButtonArrow`) controls can be disabled
+.hds-pagination-nav__button-arrow {
+  // notice: when `@disabled={{true}}` is applied, the code always renders a `<button>` element
```
alex-ju, line 77: nice one! 👏

alex-ju (approved):
No suggestions for improvement from me. Enjoyed the granularity of the commits and the descriptions really helped with the review.

packages/components/app/styles/components/pagination.scss
KristinLBradley: Good to know. Would you need to add the role though since it's already a link? Or is that needed so that the screen reader both announces that it's a link AND that it's disabled.

It would still require extra work though as the link isn't truly disabled. (If a user clicks on it, a valid link would still work.)

packages/components/app/styles/components/pagination.scss
```diff
@@ -69,15 +70,19 @@
...
     color: var(--token-color-foreground-disabled);
     cursor: not-allowed;
   }
 }
 
-// Make buttons with "Previous" & "Next" labels the same min-width
+// Make buttons with "Previous" & "Next" buttons the same min-width
 // to make them more visually centered in relationship to other sub-components
 .hds-pagination-nav__button-arrow--labeled {
   min-width: 103px;
```
KristinLBradley, line 88: Reminder: this is one we need to discuss again with @cveigt before final release.

packages/components/tests/dummy/app/controllers/components/pagination.js
```diff
@@ -9,4 +9,14 @@ export default class PaginationController extends Controller {
...
   }
+
+  @action
+  logClickedControl(control) {
+    console.log(`Control "${control}" clicked!`);
+  }
+
+  @action
+  handlePageChange(page) {
+    console.log(`Page changed to "${page}"!`);
```
KristinLBradley, line 20: Do you want to leave these console logs in for now?

KristinLBradley (approved):

packages/components/app/styles/components/pagination.scss
alex-ju: If the anchor doesn't have an href it is often treated as an inline element (similar to a span) which is why we're adding the role. The whole example is not needed anymore though, as we seem to be always rendering buttons for disabled states.

didoo: This (the `pointer-events: none` and the disabled link) is outdated now, we are showing a disabled button. Resolving the conversation.

packages/components/tests/dummy/app/controllers/components/pagination.js
didoo: no, it's on purpose: it's to have something to show to the consumers, that they have clicked a button and triggered an event/callback, when they play with one of the examples in the "how to use" or "showcase"

packages/components/app/styles/components/pagination.scss
didoo: I am going to remove it for simplicity, we can always add it later if we see it's needed.

didoo:
@KristinLBradley @alex-ju I have asked you to re-review this PR, because I've made **a lot** of changes and refactorings to the component. I can'd describe in detail all the changes, please have a look at the commit messages (I've tried to be clear in what I did) and at the overall changes.

Please review (and test) thoroughly, because I expect to have missed things. As mentioned above, the integration tests will be fixed in a follow-up PR (and then merged here once both the PRs are approved)

----

@MelSumner do you mind to do another pass on the component, to make sure it's at the level of accessibility we want? thanks

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -23,94 +18,316 @@
 <section>
   <h3 class="dummy-h3" id="component-api"><a href="#component-api" class="dummy-link-section">§</a> Component API</h3>
   <p class="dummy-paragraph">
-    The
-    <code class="dummy-code">Pagination</code>
-    component is a wrapper component for the Pagination-related sub-components, each with their own APIs:
-  </p>
+    The term «<code class="dummy-code">Pagination</code>
+    component» is an umbrella term used to indicate a set of Pagination-related components and sub-components, each with
+    their own APIs.</p>
```
KristinLBradley, line 23: I would use regular English quotes: 
“”
vs.
«»
which I think of as being "French quotes" (may be used in other languages too but I took French in high school & college).

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -23,94 +18,316 @@
 <section>
   <h3 class="dummy-h3" id="component-api"><a href="#component-api" class="dummy-link-section">§</a> Component API</h3>
   <p class="dummy-paragraph">
-    The
-    <code class="dummy-code">Pagination</code>
-    component is a wrapper component for the Pagination-related sub-components, each with their own APIs:
-  </p>
+    The term «<code class="dummy-code">Pagination</code>
```
KristinLBradley, line 21:
```suggestion
    The term “<code class="dummy-code">Pagination</code>
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -23,94 +18,316 @@
...
+    component» is an umbrella term used to indicate a set of Pagination-related components and sub-components, each with
```
KristinLBradley, line 22:
```suggestion
    component” is an umbrella term used to indicate a set of Pagination-related components and sub-components, each with
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -23,94 +18,316 @@
...
     </li>
   </ul>
 
-  <h4 class="dummy-h4">Pagination</h4>
-  <p class="dummy-paragraph" id="component-api-pagination">Here is the API for the main (“container”) component:</p>
-  <dl class="dummy-component-props" aria-labelledby="component-api-pagination">
-    <dt>totalItems <code>number</code> <strong class="required">required</strong></dt>
-    <dd>Pass the total number of items to be paginated. If no value is defined an error will be thrown.</dd>
+  <p class="dummy-paragraph">
+    These pagination sub-elements may be used directly if you need to cover a very specific use need that is not covered
```
KristinLBradley, line 74:
```suggestion
    These pagination sub-elements may be used directly if you need to cover a very specific use case that is not covered
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -23,94 +18,316 @@
...
+      <p>If no value is defined an error will be thrown.</p>
+    </dd>
+    <dt>pageSizes <code>array of integers</code> <strong class="required">required</strong></dt>
+    <dd>
+      <p>Set the page sizes users can select from. If no value is defined an error will be thrown.</p>
+      <p>
+        Example:
+        {{! prettier-ignore-start }}
+        {{! template-lint-disable no-unbalanced-curlies }}
+        <code>@pageSizes=\{{array 10 30 50}}</code>
```
KristinLBradley, line 103: Btw, I wondered if it would be useful to provide these values as default page sizes which could be overridden if desired.

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -23,94 +18,316 @@
...
+    <dt>showLabels <code>boolean</code></dt>
+    <dd>
+      <p>Used to control the visibility of the "prev/next" text labels.</p>
+      <p>Default: <span class="default">false</span></p>
+    </dd>
+    <dt>onPageChange <code>function</code></dt>
+    <dd>
+      <p>Callback function invoked (if provided) when one of the navigation controls is clicked, and a "page" change is
+        triggered.</p>
+      <p>The function receives the value of the "go to" page as argument (integer number).</p>
```
KristinLBradley, line 127:
```suggestion
      <p>The function receives the value of the "go to" page as an argument (integer number).</p>
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -23,94 +18,316 @@
...
+    Here is the API for the
+    <code class="dummy-code">Pagination::Nav</code>
+    component:
+  </p>
+  <dl class="dummy-component-props" aria-labelledby="component-api-pagination-compact">
+    <dt>route[Prev|Next] <code>string</code></dt>
+    <dd>
+      <p>The
+        <code>@route</code>
+        parameter passed down as arguments to internal render a
```
KristinLBradley, line 150:
```suggestion
        parameter passed down as arguments to internally render a
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -23,94 +18,316 @@
...
+    <dt>...attributes</dt>
     <dd>
-      <p>Used in displaying the page numbers for "numbered" type and for determining the next page to navigate to.</p>
-      <p>totalPages is calculated by HDS::Pagination wrapper component when used. However, it is required if
-        HDS::Pagination::Nav is used as a stand alone component.</p>
+      <p><code class="dummy-code">...attributes</code>
+        spreading is supported on this component.</p>
+      <p><em>Notice: it will not be applied to the
+          <code class="dummy-code">&lt;li&gt;</code>
+          wrapping element but to the
```
KristinLBradley, line 324:
```suggestion
          wrapping element but to the nested
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -124,20 +341,20 @@
...
+      <p>Used to indicate which of the provided
         <code class="dummy-code">sizes</code>
-        options.</p>
-      <p>Normally passed in as an argument on the Pagination wrapper component but can be added to
+        options is pre-selected.</p>
+      <p>Normally this value is passed automatically by the Pagination wrapper component but can be provided as argument
+        to the
         <code class="dummy-code">Pagination::SizeSelector</code>
-        itself when used as a stand alone component</p>
+        component itself when used as a stand alone component</p>
```
KristinLBradley, line 357:
```suggestion
        component itself when used as a stand alone component.</p>
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+  <p class="dummy-paragraph">This differentiation is necessary to cover both use cases of a pagination for a list with a
+    known number of elements (<strong>"numbered"</strong>) and one in which this information is not available or is
+    <a
+      href="https://jsonapi.org/profiles/ethanresnick/cursor-pagination/"
+      target="_blank"
+      rel="noopener noreferrer"
+    >cursor-based</a>
+    (<strong>"compact"</strong>).</p>
+  <p class="dummy-paragraph">In the first one, the user is presented with a list of navigation controls ("prev/next" and
+    "page numbers" to go directly to a specific page) and other optional UI elements; in the second, much simpler, the
```
KristinLBradley, line 382:
```suggestion
    "page numbers" to go directly to a specific page) and other optional UI elements; in the second, much simpler one, the
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+    user is presented with only the "prev/next" controls.</p>
+
+  <p class="dummy-paragraph">When the pagination is invoked using directly one of these two components, it will
```
KristinLBradley, line 385:
```suggestion
  <p class="dummy-paragraph">When pagination is invoked directly using one of these two components, it will
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+    <strong>automatically</strong>:</p>
+  <ul>
+    <li>
+      <p class="dummy-paragraph">provide the correct responsive layout for the entire pagination and its sub-parts</p>
+    </li>
+    <li>
+      <p class="dummy-paragraph">manage the "current page" status across the different sub-components it's made of
+        (based on the arguments provided to it and its children)</p>
```
KristinLBradley, line 393:
```suggestion
        (based on the arguments provided to it and its children).</p>
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+    specific use case, you can use the pagination sub-elements directly (<code
+      class="dummy-code"
+    >Pagination::Info/Nav(*)/SizeSelector</code>).</p>
+  <p class="dummy-paragraph">In this case, you will have to take care of different things <strong>yourself</strong></p>
+  <ul>
+    <li>
+      <p class="dummy-paragraph">the organization/layout of the elemens in the page</p>
+    </li>
+    <li>
+      <p class="dummy-paragraph">the logic to handle the "current page" status</p>
```
KristinLBradley, line 417:
```suggestion
      <p class="dummy-paragraph">the logic to handle the "current page" status.</p>
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+    components expose an
+    <code class="dummy-code">onPageChange</code>
+    callback function, invoked whenever a page change occurs. All the "navigation controls" in this cases are
+    <code class="dummy-code">&lt;button&gt;</code>
+    elements that fire an
+    <code class="dummy-code">onClick</code>
+    event that calls the
+    <code class="dummy-code">onPageChange</code>
+    function.</p>
+  <p class="dummy-paragraph">This means that if you need to update the URL when the user changes "page" in the
```
KristinLBradley, line 438:
```suggestion
  <p class="dummy-paragraph">This means that if you need to update the URL when the user changes the "page" in the
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+    Pagination (eg. to add/remove/update some query parameters) you have to do it inside the
```
KristinLBradley, line 439:
```suggestion
    Pagination (eg. to add/remove/update some query parameters), you have to do it inside the
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+    <code class="dummy-code">onPageChange</code>
+    callback you provide to the component.</p>
+  <p class="dummy-paragraph">If instead you need to update direcly the URL when the user clicks on one of the
```
KristinLBradley, line 442:
```suggestion
  <p class="dummy-paragraph">If instead you need to update the URL directly when the user clicks on one of the
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+
+  {{! ================================= }}
+  {{! ===== PAGINATION::NUMBERED ===== }}
+  {{! ================================= }}
+
+  <h4 class="dummy-h4">Pagination::Numbered</h4>
+
+  <p class="dummy-paragraph">The basic invocation of the "numbered" pagination requires the
+    <code class="dummy-code">@totalItems</code>
+    <code class="dummy-code">@itemsPerPage</code>
```
KristinLBradley, line 455:
```suggestion
    <code class="dummy-code">@itemsPerPage</code>,
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
-
-  <h4 class="dummy-h4">“Numbered” Pagination with Nav</h4>
+  <Hds::Pagination::Numbered @totalItems={{40}} @itemsPerPage={{10}} @pageSizes={{array 10 30 50}} />
+  <p class="dummy-paragraph">By default the "Info" and "SizeSelector" are displayed, and the component takes care of
+    updating the values and the states of the different elements, according to the user interactions with the component.</p>
+
+  <h5 class="dummy-h5">Extra arguments</h5>
+  <p class="dummy-paragraph">It's possible to customize the "Info", "Controls", and "SizeSelector" providing additional
+    arguments to them. For more details about these parameters, refer to the "Components API" section.</p>
+  <p class="dummy-paragraph">Below an example of some of these extra arguments:</p>
```
KristinLBradley, line 477:
```suggestion
  <p class="dummy-paragraph">Below is an example of some of these extra arguments:</p>
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
   <p class="dummy-paragraph">Renders to:</p>
-  <Hds::Pagination @totalItems={{103}} @itemsPerPage={{10}} as |P|>
-    <P.Info />
-    <P.Nav />
-    <P.SizeSelector @sizes={{array 10 30 50}} />
-  </Hds::Pagination>
-
-  <h4 class="dummy-h4">“Numbered” pagination with all optional child components</h4>
+  <Hds::Pagination::Numbered @totalItems={{150}} @itemsPerPage={{10}} @pageSizes={{array 10 30 50}} />
+  <p class="dummy-paragraph">If it's necessary it's possible to disable this functionality using a specific
```
KristinLBradley, line 523:
```suggestion
  <p class="dummy-paragraph">If necessary, it's possible to disable this functionality using a specific
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+  <p class="dummy-paragraph">Renders to:</p>
+  <Hds::Pagination::Numbered
+    @totalItems={{40}}
+    @itemsPerPage={{10}}
+    @pageSizes={{array 10 30 50}}
+    @onPageChange={{this.handlePageChange}}
+    @onPageSizeChange={{this.handlePageSizeChange}}
+  />
+  <p class="dummy-paragraph">The first function
+    <code class="dummy-code">onPageChange</code>
```
KristinLBradley, line 575: I would change this to:

"The first function, `onPageChange`, is ..."

Or perhaps to:

"The first `onPageChange` function is ..."

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+    is invoked when a user interacts with a navigation control ("prev/next" or "page number") and so can be used to
+    respond to a "page" change (eg. updating the list of items in the page and/or updating the routing/URL).</p>
+  <p class="dummy-paragraph">The second function
+    <code class="dummy-code">onPageSizeChange</code>
```
KristinLBradley, line 579: Similar to above, I would change this to:

"The second function, `onPageSizeChange`, is ..."

Or perhaps to:

"The second `onPageSizeChange` function is ..."

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+    is invoked when a user interacts with the "size selector" and so can be used to respond to a "page size" change (eg.
+    updating the number of items listed in the page, updating the routing/URL, updating other elements in the page).</p>
```
KristinLBradley, line 581:
```suggestion
    updating the number of items listed in the page, updating the routing/URL, and/or updating other elements in the page).</p>
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+
+  <h5 class="dummy-h5">Routing/URL updates</h5>
+  <p class="dummy-paragraph">If you want the pagination to change directly the URL of the page (eg. adding/removing
```
KristinLBradley, line 584:
```suggestion
  <p class="dummy-paragraph">If you want the pagination to change the URL of the page directly (eg. adding/removing
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+  <pre>🚨 TODO </pre>
+  <p class="dummy-paragraph">Renders to:</p>
+  <pre>🚨 TODO 🚨</pre>
+
+  {{! =============================== }}
+  {{! ===== PAGINATION::COMPACT ===== }}
+  {{! =============================== }}
+
+  <h4 class="dummy-h4">Pagination::Compact</h4>
+  <p class="dummy-paragraph">By default, the basic use of the pagination provides .</p>
```
KristinLBradley, line 595:
```suggestion
  <p class="dummy-paragraph">By default, the basic use of the pagination provides:</p>
```

KristinLBradley: Not sure if something was left out of the above.

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+
+  <p class="dummy-paragraph">The basic invocation of the "compact" pagination doesn't require any argument (apart from
```
KristinLBradley, line 597:
```suggestion
  <p class="dummy-paragraph">The basic invocation of the "compact" pagination doesn't require any arguments (apart from
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
   {{! template-lint-enable no-unbalanced-curlies }}
   {{! prettier-ignore-end }}
   <p class="dummy-paragraph">Renders to:</p>
-  <Hds::Pagination::Nav @totalPages={{10}} @type="numbered" />
+  <Hds::Pagination::Compact />
+  <p class="dummy-paragraph">In this variant only the "prev" and "next" navigation controls are displayed.</p>
 
-  <h4 class="dummy-h4">“Truncated” PaginationNav</h4>
+  <h5 class="dummy-h5">Extra arguments</h5>
+  <p class="dummy-paragraph">If necessry, it's possible to hide the labels in the controls:</p>
```
KristinLBradley, line 614:
```suggestion
  <p class="dummy-paragraph">If necessary, it's possible to hide the labels in the controls:</p>
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
-    <p class="dummy-paragraph">
-      <a href="[ADD THE LINK TO THE FIGMA FILE/PAGE HERE!]" target="_blank" rel="noopener noreferrer">Figma UI Kit</a>
-    </p>
-    <br />
-    <img class="dummy-figma-docs" src="/assets/images/pagination-design-usage.png" alt="" role="none" />
-  </div>
-  }}
-</section>
+  <h5 class="dummy-h5">Events handling</h5>
+  <p class="dummy-paragraph">The component exposes a callback function that can be used to respond to :</p>
```
KristinLBradley, line 629: Missing some language? Perhaps change to below:

"respond to :"
=>
"respond to page changes:"

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+    @code="
+      <Hds::Pagination::Compact
+        @onPageChange=\{{this.handlePageChange}}
+      />
+    "
+  />
+  <p class="dummy-paragraph">Renders to:</p>
+  <Hds::Pagination::Compact @onPageChange={{this.handlePageChange}} />
+  <p class="dummy-paragraph">The function
+    <code class="dummy-code">onPageChange</code>
```
KristinLBradley, line 643: I would change the order to:

The `onPageChange` function is ...

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+    is invoked when a user interacts with a "prev" or "next" navigation control and so can be used to respond to a
+    "page" change (eg. updating the list of items in the page and/or updating the routing/URL).</p>
+
+  <h5 class="dummy-h5">Routing/URL updates</h5>
+  <p class="dummy-paragraph">If you want the pagination to change directly the URL of the page (eg. adding/removing
```
KristinLBradley, line 648:
```suggestion
  <p class="dummy-paragraph">If you want the pagination to change the URL of the page directly (eg. adding/removing
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -148,157 +365,295 @@
...
+    query parameters) you need to pass the routing parameters to the component:</p>
```
KristinLBradley, line 649:
```suggestion
    query parameters), you need to pass the routing parameters to the component:</p>
```

KristinLBradley (approved):
Looks great overall. I just added minor grammar and spelling fix suggestions.

packages/components/tests/dummy/app/templates/components/pagination.hbs
didoo: Yeah, I thought the same. Good point. Can you investigate this with @cveigt + the ambassadors, and see if it's possible to come up with two reasonable defaults for:
- `@pagesizes` (available options in the `<select>/Hds::Pagination::SizeSelect`)
- `@selectedSize` (selected option)
and in that case open a PR against this one?

didoo: [Note: this is not a blocker for the release, because this change would not be a breaking change, so we may introduce it later as a _minor_ update]

packages/components/tests/dummy/app/templates/components/pagination.hbs
didoo: Fixed

packages/components/tests/dummy/app/templates/components/pagination.hbs
didoo: Fixed

packages/components/tests/dummy/app/templates/components/pagination.hbs
didoo: Fixed

packages/components/tests/dummy/app/templates/components/pagination.hbs
didoo: Fixed

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -23,94 +18,316 @@
...
+    <dt>href + route/models/model/query/current-when/replace</dt>
+    <dd>
+      <p>These are the parameters that are passed down as arguments to the
+        <code class="dummy-code">Hds::Interactive</code>
+        component (used internally). For more details about how this low-level component works please refer to
+        <LinkTo @route="utilities.interactive">its documentation page</LinkTo>.</p>
+    </dd>
+    <dt>disabled <code>boolean</code></dt>
+    <dd>
+      <p>If the control is disabled</p>
```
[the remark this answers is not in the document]
didoo: Fixed

packages/components/addon/components/hds/pagination/nav/index.hbs
didoo: Outdated.

packages/components/tests/dummy/app/templates/components/pagination.hbs
didoo: Outdated.

packages/components/tests/dummy/app/templates/components/pagination.hbs
didoo: Outdated.

packages/components/tests/dummy/app/templates/components/pagination.hbs
didoo: Fixed

didoo:
> Looks great overall. I just added minor grammar and spelling fix suggestions.

Thanks for the thorough spell/grammar checking! 🙏

I think I have _fixed_ all the suggestions you made, but if I missed something let me know.

packages/components/addon/components/hds/pagination/nav/ellipsis.hbs
```diff
@@ -1,3 +1 @@
-<li class="hds-pagination-nav__page-item" ...attributes>
-  <div class="hds-pagination-nav__ellipsis">...</div>
-</li>
\ No newline at end of file
+<div class="hds-pagination-nav__ellipsis" ...attributes>...</div>
```
MelSumner, line 1: Can you remind me how this would be handled for assistive tech?

didoo: @MelSumner looking at other design systems, this is exactly how it's implemented in all of them:
- https://design-system.service.gov.uk/components/pagination/
- https://design-system.w3.org/components/pagination.html
- https://atlassian.design/components/pagination/examples
- https://ons-design-system.netlify.app/components/pagination/
- https://primer.style/css/components/pagination

What would you suggest to do?

alex-ju (approved):
Looks great! Sorry for the slow(er) review on this.

thiskevinwang:
Just offering up some observations (since I'm doing related work on the React side of things).
It appears that none of the `Pagination::Nav::Number` buttons are tab-focusable on Safari.

didoo:
> ~Just offering up some observations (since I'm doing related work on the React side of things). It appears that none of the `Pagination::Nav::Number` buttons are tab-focusable on Safari (`Version 16.0 (17614.1.25.9.10, 17614)`).~ 
> Edit 1: Disregard! I had to **Enable Full Keyboard Access** 

Exactly! I was going to share this link 😀
https://jansensan.net/technical/how-fix-focus-rings-not-showing-links-macos

didoo:
since the PR #1060 for the API refactoring (to support routing-based pagination) has been much more complex than was expected, with a lot of commits, doesn't make sense to merge it into this PR; so I've decided to merge this one on the `feature branch` and then open a separate PR (squashing and cleaning up the commits) for the API refactoring for routing

PR #1021, opened 2023-01-06
Error page fixes part 1/n

Dhaulagiri:
### :pushpin: Summary

* [x] pre-render the error route (missed this before)
* [x] add some dimensions to the image
* [x] adjust styles to account for sidebar
* [ ] Probably in a followup PR - figure out why vercel is going index -> error page instead of directly to it

### :link: External links

Jira ticket: [HDS-1309](https://hashicorp.atlassian.net/browse/HDS-1309)

[HDS-1309]: https://hashicorp.atlassian.net/browse/HDS-1309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/styles/pages/application/error.scss
+++ b/website/app/styles/pages/application/error.scss
@@ -3,6 +3,16 @@
 @use "../../breakpoints" as breakpoint;
 @use "../../typography/mixins";
 
+// temporary fix - see:
+// - https://github.com/hashicorp/design-system/pull/1021
+// - https://hashicorp.atlassian.net/browse/HDS-1309?focusedCommentId=178673
+
+// `ember-body-class` sets specific classes on the `<body>` element
+body.application.error {
+  background-color: var(--doc-color-white) !important;
+}
+
+
 .doc-page-error {
   display: flex;
   grid-area: sidebar / sidebar / stage / stage;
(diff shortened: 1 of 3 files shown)


Dhaulagiri:
> pre-render the error route

This slightly improves the situation in that once you get to the error page, reloading it will apply the correct styles

didoo (approved):

MelSumner:
> figure out why vercel is going index -> error page instead of directly to it

Two potential fixes:
1. We don't have [wildcard/globbing](https://guides.emberjs.com/v3.24.0/routing/defining-your-routes/#toc_wildcard--globbing-routes) set up in our `router.js` file.
2. Vercel might need a config file like Netlify does ([example](https://github.com/MelSumner/a11y-automation/blob/main/.netlifyredirects))

Dhaulagiri:
1) I think we do but it's to render the `show` routes?

2) is something I was wondering as well as the prember docs [seem to indicate as much](https://github.com/ef4/prember#configuring-your-webserver), and it would conceptually make sense why we're seeing the behavior we're seeing if vercel doesn't know how to route to an error page correctly.

I did try this in a commit in this branch that I've since dropped and it didn't help, but it may be that vercel.json changes don't take effect in a preview app or at least when a vercel.json didn't exist previously. What we could do it try merging a PR with that change just to test in production and we can always revert it if it doesn't work.

Dhaulagiri:
Part of the underlying issue or challenge is that the request doesn't actually trigger a `404` which may be something amiss with our error handling

Dhaulagiri:
Will be fixed in #1074

PR #1024, opened 2023-01-06
New docs website/share card

Dhaulagiri:
### :pushpin: Summary

Adds share card to the meta tags for pages that don't have an illustration. Unfortunately the card for twitter still doesn't work but I'm not quite sure why (this PR doesn't change that)

Screen Shot 2023-01-06 at 15 40 50
Screen Shot 2023-01-06 at 15 41 11


website/app/templates/head.hbs
```diff
@@ -37,9 +37,7 @@ We've made some small customizations to simplify this compared to the default te
 
 {{#if this.model.imgSrc}}
   <meta property="og:image" content={{this.model.imgSrc}} />
-  <meta property="og:image:type" content="image/png" />
-  <meta property="og:image:width" content="256" />
```
Dhaulagiri, line 41: it turns out these didn't seem to matter and would be problematic anyhow since the dimensions of this image depend on which type we're sharing, and AFACT there is no way to change these dynamically

didoo: 👍

didoo (approved):

cveigt (approved):

MelSumner:
@Dhaulagiri Here's [a sample of a working social meta](https://github.com/MelSumner/please-fund-a11y/blob/aac2759bbcc9cb7e7161a76e3f755166620247a4/index.html#L8), it took a lot of testing to get it to work properly. See lines 8-18. A lot of the tiny details mattered.

PR #1027, opened 2023-01-09
Fix power select styles

alex-ju:
### :pushpin: Summary

This PR brings the `<PowerSelect>` style overrides to work on the [new website](https://hds-website-hashicorp.vercel.app/overrides/power-select) as they did on the [old one](https://design-system-components-hashicorp.vercel.app/overrides/power-select).

For an unknown reason, on the old website, we were benefitting from a series of styles that are no longer applied to the new website. I have checked the configuration and the only difference between the two is the following (extract from `ember-cli-build`)

```
    'ember-power-select': {
      theme: false,
    },
```

Applying the same configuration to the new website doesn't change anything visually, hence the only option I could see was to port the missing CSS declarations into the override to make the overrides more resilient.

I'm still scratching my head around this, but not sure spending more time on would be of help.

There are many subtle changes, but the most obvious one is in the multiselect version.

<table>
<tr><td>

hds-website-hashicorp vercel app_overrides_power-select

</td><td>

localhost_4200_overrides_power-select

</td></tr>
</table>

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/website/app/styles/app.scss
+++ b/website/app/styles/app.scss
@@ -26,7 +26,7 @@
 
 // Third-party declarations
 @import "prism-dracula";
-
+@import "ember-power-select";
 
 // GLOBALS
 
--- a/website/docs/overrides/power-select/partials/code/how-to-use.md
+++ b/website/docs/overrides/power-select/partials/code/how-to-use.md
@@ -78,7 +78,7 @@ To consistently style the `@afterOptionsComponent` use the `hds-power-select__af
 
 Where `power-select/after-options.hbs` would look like this:
 
-```handlebars
+```handlebars{data-execute=false}
 <div class="hds-power-select__after-options">
   5 results
 </div>
--- a/website/docs/overrides/power-select/partials/code/showcase.md
+++ b/website/docs/overrides/power-select/partials/code/showcase.md
@@ -10,7 +10,7 @@
     <div class="hds-power-select">
       <PowerSelect
         @options={{this.OPTIONS}}
-        @selected={{@model.SELECTED}}
+        @selected={{this.SELECTED}}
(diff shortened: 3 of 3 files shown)


Dhaulagiri:
> For an unknown reason, on the old website, we were benefitting from a series of styles

@alex-ju is [this line](https://github.com/hashicorp/design-system/blob/e93e8a603139c12cd92bd29536007e2c177118de/packages/components/ember-cli-build.js#L12) what is making the magic happen on the old site?

alex-ju:
> > For an unknown reason, on the old website, we were benefitting from a series of styles
> 
> @alex-ju is [this line](https://github.com/hashicorp/design-system/blob/e93e8a603139c12cd92bd29536007e2c177118de/packages/components/ember-cli-build.js#L12) what is making the magic happen on the old site?

I thought about that too, but adding it (with the updated path) doesn't change things 🤔

alex-ju:
> > > For an unknown reason, on the old website, we were benefitting from a series of styles
> > 
> > 
> > @alex-ju is [this line](https://github.com/hashicorp/design-system/blob/e93e8a603139c12cd92bd29536007e2c177118de/packages/components/ember-cli-build.js#L12) what is making the magic happen on the old site?
> 
> I thought about that too, but adding it (with the updated path) doesn't change things 🤔

It turns out I wasn't following my own instructions – which is follow the [Ember Power Select installation guide](https://ember-power-select.com/docs/installation) so I forgot to import the styles 🤦 

I dropped the first commit with CSS adjustments and added a new one with the import.

didoo (approved):

Dhaulagiri:
Nice!

PR #1031, opened 2023-01-09
Updates "explore" links on home page to add description for AT

MelSumner:
### :pushpin: Summary

If merged, this PR adds `aria-label` to the "explore" links on the home page so they have **unique** accessible names for users with AT. Typically we would want the visible link name to be the same as the accessible link name, but in this case, they both start with the same word (clutch), and they provide an equivalent experience.

Before:
CleanShot 2023-01-09 at 12 30 48

After: 
image

***

### DIFF
--- a/website/app/templates/index.hbs
+++ b/website/app/templates/index.hbs
@@ -23,21 +23,19 @@
     </div>
     <ul class="doc-page-home__cards" role="list">
       {{#each this.cards as |card|}}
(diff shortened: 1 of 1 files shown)


MelSumner:
Of course, the second I submitted this PR I wondered if I should add the card title instead of the description, so it reads "explore components" etc. instead. I'll update the PR.

Dhaulagiri (approved):

PR #1034, opened 2023-01-09
Updates to error page

MelSumner:
@Dhaulagiri tidied some things up, there's still some oddness with the way the page layout was done but I think this could maybe improve things. Feel free to use or discard, no worries either way!

### DIFF
--- a/website/app/routes/show.js
+++ b/website/app/routes/show.js
@@ -113,6 +113,10 @@ export default class ShowRoute extends Route {
 
   @action
   error() {
-    this.router.replaceWith('error');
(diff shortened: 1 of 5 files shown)


website/app/routes/show.js
```diff
@@ -113,6 +113,10 @@ export default class ShowRoute extends Route {
 
   @action
   error() {
-    this.router.replaceWith('error');
+    if (this.error.status === '404') {
```
Dhaulagiri, line 116: this may be part of the problem, but i don't think our error on line 32 is actually triggering a `404`

PR #1041, opened 2023-01-10
Button docs: added more a11y info

MelSumner:
### :pushpin: Summary

If merged, this PR adds more a11y information for the button component. 

***

### DIFF
--- a/website/docs/components/button/partials/accessibility/accessibility.md
+++ b/website/docs/components/button/partials/accessibility/accessibility.md
@@ -10,7 +10,7 @@ Provide annotations alongside each design of the non-visual experience. This cou
 
 ![Accessibility annotation example](/assets/components/button/button-annotation_example.png)
 
-## Accessibility in engineering
+## Accessibility considerations
 
 Years of misuse have muddled the purpose of a button element vs. a link element, so here is a good rule of thumb to follow:
 
--- a/website/docs/components/button/partials/accessibility/accessibility.md
+++ b/website/docs/components/button/partials/accessibility/accessibility.md
@@ -14,7 +14,7 @@ Provide annotations alongside each design of the non-visual experience. This cou
 
 Years of misuse have muddled the purpose of a button element vs. a link element, so here is a good rule of thumb to follow:
 
-If it _goes to URL_, use a link. If it does an _action_ (toggle a state, submit a form, etc.), then use a button.
+If it _goes to a URL_, use a link. If it does an _action_ (toggles a state, submits a form, etc.), then use a button.
 
 Now, you may be thinking, “What about instances where I submit a form and redirect a user to a different URL?” This is still a button use case.
 
--- a/website/docs/components/button/partials/accessibility/accessibility.md
+++ b/website/docs/components/button/partials/accessibility/accessibility.md
@@ -12,7 +12,7 @@ Provide annotations alongside each design of the non-visual experience. This cou
 
 ## Accessibility in engineering
 
-Years of misuse have muddled the purpose of a button element and a link element, so here is a good rule of thumb to follow:
+Years of misuse have muddled the purpose of a button element vs. a link element, so here is a good rule of thumb to follow:
 
 If it _goes to URL_, use a link. If it does an _action_ (toggle a state, submit a form, etc.), then use a button.
 
--- a/website/docs/components/button/partials/accessibility/accessibility.md
+++ b/website/docs/components/button/partials/accessibility/accessibility.md
@@ -12,7 +12,7 @@ Provide annotations alongside each design of the non-visual experience. This cou
 
 ## Accessibility in engineering
 
-While years of misuse have muddled the purpose of a button element and a link element, so here is a good rule of thumb to follow:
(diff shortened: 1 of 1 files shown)


website/docs/components/button/partials/accessibility/accessibility.md
```diff
@@ -10,6 +10,18 @@ Provide annotations alongside each design of the non-visual experience. This cou
 
 ![Accessibility annotation example](/assets/components/button/button-annotation_example.png)
 
+## Accessibility in engineering
+
+While years of misuse have muddled the purpose of a button element and a link element, so here is a good rule of thumb to follow:
```
MelSumner, line 15:
```suggestion
Years of misuse have muddled the purpose of a button element and a link element, so here is a good rule of thumb to follow:
```

website/docs/components/button/partials/accessibility/accessibility.md
```diff
@@ -10,6 +10,18 @@ Provide annotations alongside each design of the non-visual experience. This cou
 
 ![Accessibility annotation example](/assets/components/button/button-annotation_example.png)
 
+## Accessibility in engineering
```
heatherlarsen, line 13: This is an important section for designers and engineers, so this heading gives me pause. Could it be something more specific to the content in the section...maybe something like "Button vs link" or "When to use a button or a link"

MelSumner: I definitely can change the heading; @didoo asked me to go through each component and make sure there was an "accessibility for engineers" section with pertinent information, hence the heading.

website/docs/components/button/partials/accessibility/accessibility.md
```diff
@@ -10,6 +10,18 @@ Provide annotations alongside each design of the non-visual experience. This cou
...
+
+Years of misuse have muddled the purpose of a button element and a link element, so here is a good rule of thumb to follow:
```
KristinLBradley, line 15:
```suggestion
Years of misuse have muddled the purpose of a button element vs. a link element, so here is a good rule of thumb to follow:
```

website/docs/components/button/partials/accessibility/accessibility.md
```diff
@@ -10,6 +10,18 @@ Provide annotations alongside each design of the non-visual experience. This cou
...
+
+If it _goes to URL_, use a link. If it does an _action_ (toggle a state, submit a form, etc.), then use a button.
```
KristinLBradley, line 17:
```suggestion
If it _goes to a URL_, use a link. If it does an _action_ (toggles a state, submits a form, etc.), then use a button.
```

website/docs/components/button/partials/accessibility/accessibility.md
KristinLBradley: To me, it seems helpful to group content more relevant to designers under "Accessibility in design" vs. content more relevant to engineers under "Accessibility in engineering" as is done here. It doesn't necessarily mean that designers and engineers wouldn't find value in the content under the different headings, just that the content is generally more relevant to one group vs. the other.

The question of using a button vs. a link is related to the underlying HTML code, which seems to be more of an engineering concern.

Maybe there is other information related to Figma button and link components that it would be helpful to include though? If there is, then maybe that could be added to the "Accessibility in design" section? @heatherlarsen

KristinLBradley (approved):

website/docs/components/button/partials/accessibility/accessibility.md
heatherlarsen: Typically I would agree with you on that, @KristinLBradley. But this specific content has been an ongoing challenge for us and our consumers to get "right". We've consistently seen that there is a misunderstanding of how buttons and links work, mainly among our designers (because of the common misuse of them in the industry). As such, they tend to use buttons and links interchangeably within the designs, which is something we'd like to see change/improve.

I fear that if this information is under the heading "Accessibility for engineers," a Designer may overlook or skip over this content thinking that it isn't relevant to them.

didoo (commented):
left a small suggestion

website/docs/components/button/partials/accessibility/accessibility.md
KristinLBradley: Thanks for the context @heatherlarsen, that makes sense. I have also noticed similar confusion on the difference between buttons vs. links in the past on the design side.

In that case, I think some other header text which calls out the information more would be useful. Something like "Important accessibility considerations" or something else more general that emphasizes the importance or that it's a common confusion.

Dhaulagiri: > @didoo asked me to go through each component and make sure there was an "accessibility for engineers" section with pertinent information, hence the heading.

We did discuss amongst engineers this morning to clarify that the intent is not to carve out specific _engineer_ guidance as part of the a11y updates, but more along the lines of what @heatherlarsen has been saying here by keeping things more inclusive of all consumers.

MelSumner, line 13:
```suggestion
## Accessibility considerations
```

heatherlarsen (approved):

PR #1044, opened 2023-01-10
fix scrolling issue

MelSumner:
Fixes scroll issue

### DIFF
--- a/website/app/routes/show.js
+++ b/website/app/routes/show.js
@@ -112,6 +112,11 @@ export default class ShowRoute extends Route {
   }
 
   @action
+  didTransition() {
+    window.scrollTo(0, 0);
+  }
+
+  @action
   error() {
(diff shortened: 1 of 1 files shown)


Dhaulagiri:
@MelSumner this does not fix the issue if you are transitioning between routes that are not `show` like from `components` to `foundations`

MelSumner:
@Dhaulagiri I see what you mean. 

@didoo let's schedule a pairing session, I have some questions like why `application.js` is using classic, where are the prember settings configured, etc.

Dhaulagiri:
@MelSumner it looks like we can convert the application route to octane and move this logic there and things will work. I suspect it's classic because ported it over from field-guide.

MelSumner:
@Dhaulagiri I think we have this working how we wanted after our group pairing session, yes?

PR #1055, opened 2023-01-12
Simplify analytics enablement

Dhaulagiri:
### :pushpin: Summary

This PR simplifies our analytics configuration to always be enabled if the vercel environment is production. 

The upshot is that we can remove ember-cli-dotenv and not have the warning in our console all the time.

No strong attachment to this, but more putting it forth as an option.

### Testing this

I made a commit with `VERCEL_ENV === 'preview'` and confirmed in that review app it was working as expected. Will need to confirm once this goes to prod that is true there as well.


MelSumner (approved):
:shipit:

PR #1063, opened 2023-01-13
Iterations for elevation

heatherlarsen:
### :pushpin: Summary

If merged, this PR introduces adjustments to the Elevation documentation.

- Updated tab order (Guidelines -> Code)
- Making the code section more actionable
- Adding image content into a table and deleting the image
- Adding basic usage details

### :link: External links

Jira ticket: [HDS-1350](https://hashicorp.atlassian.net/browse/HDS-1350)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-1350]: https://hashicorp.atlassian.net/browse/HDS-1350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


website/docs/foundations/elevation/partials/guidelines/guidelines.md
```diff
@@ -1,4 +1,20 @@
...
+
+- to surface interactivity
+- to elevate and group key content on the page
+- to convey natural page hierarchy and modality
+
+## Levels of elevation
+
+| Level  |  Elevation  |   |  Surface  |    |  Object type  |
+|--------|-------------|---|-----------|----|---------------|
+| **Overlay**  |  <div class="hds-elevation-overlay" style="width: 40px; height: 40px; border-radius: 3px;"><Doc::Placeholder @width="40" @height="40" @background="transparent" /></div>  |   |  <div class="hds-surface-overlay" style="width: 40px; height: 40px; border-radius: 3px;"><Doc::Placeholder @width="40" @height="40" @background="transparent" /></div>  |  Modals  |    |
```
didoo, line 13: this is totally unmaintainable. let's replace it with an image.

heatherlarsen: The text in an image cannot be searched, and I don't imagine we'll be changing our elevation values often, so I personally don't see much risk in keeping this.

website/docs/foundations/elevation/partials/code/how-to-use.md
```diff
@@ -1,47 +1,52 @@
...
-Which one to use will depend on the context: since both refer to the same design token values, from the design system perspective there is no difference between the two methods (it’s more of a preference for the "consumer" codebase).
+!!! Info
+
+Note that `border-radius` is not included with this token and needs to be set according to the specs of the UI element.
+!!!
 
 ### Design tokens
 
-You can use the "elevation" and "surface" [design tokens](./tokens) directly in your CSS definitions:
+Use the `elevation` and `surface` [design tokens](./tokens) directly in your CSS definitions. Note that it can only be used with the `box-shadow` property.
```
jorytindall, line 12: Should this be 'Note that **they** should only be use...' since you reference design tokens plural?

website/docs/foundations/elevation/partials/code/how-to-use.md
```diff
@@ -1,47 +1,52 @@
 ## How to use these styles
 
-There are two different ways to apply an "elevation" or "surface" effect to a UI element, via **design tokens** or via **CSS helper classes**.
+You can apply an `elevation` or `surface` effect to an element via **design tokens** or **CSS helper classes**.
 
-Which one to use will depend on the context: since both refer to the same design token values, from the design system perspective there is no difference between the two methods (it’s more of a preference for the "consumer" codebase).
+!!! Info
+
+Note that `border-radius` is not included with this token and needs to be set according to the specs of the UI element.
```
jorytindall, line 7: In the other update docs (I think Focus ring?) there was a point about recommending `inherit` for the border radius value. Is that important/relevant to also note here?

jorytindall (approved):
I have no strong opinion on the nested markup in the table (understanding it's hard to maintain, but it probably won't be changed much), but a compromise could be using images instead for the actual visual but still use a table so that the values are searchable. Not sure how much more work that would be.

website/docs/foundations/elevation/partials/code/how-to-use.md
heatherlarsen: It wasn't included in the initial documentation, so I figured it may not be the same recommendation here. @didoo, is using `inherit` recommended here as well?

website/docs/foundations/elevation/partials/guidelines/guidelines.md
KristinLBradley: I would at least probably do the table all in HTML vs. a combo of markup with nested HTML and also create a doc component for the repeated examples vs. all the inline styles. That would help readability of the code a lot I think.

KristinLBradley: Jory's idea of using images in a table is a good idea too I think, then you wouldn't need to create a special component like I suggested.

Btw, to get the HTML markup for the table, you could use dev tools to view the rendered code of your page, then copy and paste it. Then edit it to add images for the examples instead of divs with inline styles.

didoo: > The text in an image cannot be searched, and I don't imagine we'll be changing our elevation values often, so I personally don't see much risk in keeping this.
> Jory's idea of using images in a table is a good idea too I think, then you wouldn't need to create a special component like I suggested.
> Btw, to get the HTML markup for the table, you could use dev tools to view the rendered code of your page, then copy and paste it.

@heatherlarsen @KristinLBradley @jorytindall + @Dhaulagiri 

When we started the new website project, we discussed the pros and cons of the different options that were available, and one of the clear limitations of using markdown was that it could become complex to achieve specific layouts. 
This is one of these cases. The fact that technically it can be done, doesn't mean it should be done.
I know an image is not ideal, but the code above, even if we use single images, is beyond an acceptable level of maintenance (I know you think we will not change it often that code, but we will change other code, eg. the markdown table parser, or the placeholder, or something else, and then things can become messy: just one character off, and the entire rendering blows up, and debugging and finding the error becomes a pain).
Please, use an image in this specific case.

website/docs/foundations/elevation/partials/code/how-to-use.md
didoo: In this case, unlike the focus ring, there may be no "parent" elements that have a border radius. This means the consumer may need to apply the border radius directly to the same element that has the elevation applied to it. So in some cases they can use `inherit`, in others no. Better to keep it generic.

website/docs/foundations/elevation/partials/guidelines/guidelines.md
heatherlarsen: After speaking with @Dhaulagiri, we decided to move forward with an image for launch. I've updated the image to use styles from the website to fit more seamlessly within the documentation.

KristinLBradley (approved):

PR #1068, opened 2023-01-17
New docs website / Add a "filter" to the sidebar navigation

didoo:
### :pushpin: Summary

One of the suggestions received by the ambassadors was to have a "filter" functionality for the sidebar.
This PR implements such functionality, using as a model the "filter" on https://developer.hashicorp.com/

In this PR I have:
- added support for `keywords` attribute in frontmatter
- added filtering logic to the sidebar
- updated the `wiki` documentation adding `keywords` to the "frontmatter" explanation

👉 👉 👉 **Preview**: https://hds-website-git-new-docs-website-sidebar-filter-hashicorp.vercel.app/components

### 🚧 TODOs

- [ ] Once approved and merged, create Jira ticket to add keywords to all the pages (that we think should have them)

screenshot_2280

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-1417

***

### 👀 Reviewer's checklist:

### DIFF
--- a/website/app/components/doc/form/filter.hbs
+++ b/website/app/components/doc/form/filter.hbs
@@ -1,0 +1,10 @@
+<div class="doc-form-filter">
+  <input
+    type="search"
+    class="doc-form-filter__control"
+    placeholder={{@placeholder}}
+    value={{@filterQuery}}
+    {{on "input" (perform @onInput value="target.value")}}
+    ...attributes
+  />
+</div>
--- a/website/app/components/doc/form/filter.hbs
+++ b/website/app/components/doc/form/filter.hbs
@@ -3,6 +3,7 @@
     type="search"
     class="doc-form-filter__control"
     placeholder={{@placeholder}}
+    aria-label={{@placeholder}}
     value={{@filterQuery}}
     {{on "input" (perform @onInput value="target.value")}}
     ...attributes
--- a/website/app/components/doc/page/sidebar.hbs
+++ b/website/app/components/doc/page/sidebar.hbs
@@ -30,7 +30,7 @@
         </li>
       </ul>
     </nav>
-    {{#if this.structuredPageTree}}
+    {{#if (or this.structuredPageTree this.isFiltered)}}
       <div class="doc-page-sidebar__filter">
         <Doc::Form::Filter
           @placeholder="Filter sidebar"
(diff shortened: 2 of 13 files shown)


website/app/components/doc/page/sidebar.hbs
```diff
@@ -31,8 +31,22 @@
...
+      <div class="doc-page-sidebar__filter">
+        <Doc::Form::Filter
+          @placeholder="Filter sidebar"
+          autocomplete="off"
+          autocorrect="off"
+          autocapitalize="none"
+          spellcheck="false"
+          @onInput={{this.filterPageTree}}
+        />
+      </div>
```
alex-ju, line 43: I believe the conditional statement sould start after the filter input, otherwise, when we don't have results it gets removed too and we may end up with blank sidebar.
```suggestion
      <div class="doc-page-sidebar__filter">
        <Doc::Form::Filter
          @placeholder="Filter sidebar"
          autocomplete="off"
          autocorrect="off"
          autocapitalize="none"
          spellcheck="false"
          @onInput={{this.filterPageTree}}
        />
      </div>
    {{#if this.structuredPageTree}}
```

Then we can also have an else for this if, where we tell users that we couldn't find any relevant items.

website/app/components/doc/form/filter.hbs
```diff
@@ -0,0 +1,10 @@
+<div class="doc-form-filter">
+  <input
+    type="search"
+    class="doc-form-filter__control"
+    placeholder={{@placeholder}}
```
alex-ju, line 5:
```suggestion
    placeholder={{@placeholder}}
    aria-label={{@placeholder}}
```

alex-ju (commented):
I really like the idea of keywords, and it's a good way to define alias names for components 👏  

Found the following issues: 
 - we lose the filter input when no matches are found, otherwise happy to approve (left an inline comment about this)
 - the about and patterns pages are crashing with: 'TypeError: Cannot convert undefined or null to object' (have not investigate further, but I'm not sure we want the filter on all pages).

Dhaulagiri:
It would be good if we could add some basic acceptance tests around this behavior as well

jorytindall (commented):
This might be unrelated to the changes in this branch, but I noticed a weird bug in the routing in this branch that wasn't present in Heather's branch that I just reviewed for content changes: https://www.loom.com/share/563fe5b164aa45d1acdcf9c770e948e4

If it's unrelated or being fixed elsewhere, happy to approve and move on!

website/app/components/doc/page/sidebar.hbs
didoo: Fixed (in a slightly different way).

website/app/components/doc/form/filter.hbs
didoo: Fixed

didoo:
> It would be good if we could add some basic acceptance tests around this behavior as well

@Dhaulagiri done.

> Found the following issues:

@alex-ju fixed.

> This is probably the same thing as @alex-ju 's point above:

Yes @jorytindall it was. Fixed now.

jorytindall (approved):

alex-ju (approved):
Excellent!

PR #1071, opened 2023-01-17
add url to meta

Dhaulagiri:
### :pushpin: Summary

Adds our final domain name to ember-meta's config which will inject it a few helpful places in our meta tags

This will also be useful for the twitter meta tag work we're doing, but short term it will also fix a bug preventing our site from being crawled. See [this link](https://support.google.com/webmasters/answer/7440203#duplicate_page_without_canonical_tag) for details.


alex-ju (approved):
Makes sense!

Dhaulagiri:
FYI @KristinLBradley as this may impact your twitter work. I have some ideas on how we could make this more dynamic if need be.

PR #1072, opened 2023-01-17
Website: Add ember-router-scroll

MelSumner:
### :pushpin: Summary

If merged, this PR adds `ember-router-scroll` as a dependency, to handle "scroll to top" behavior.

Note: if I open a new tab and visit a hash URL, it's being removed. I don't think it's this addon, since we were seeing it before, I think it's something in one of the controllers. I haven't been able to pinpoint exactly what is causing this behavior, but I'll keep working on it. 

In the meantime, this addon is a step in the right direction for our app.

***

### DIFF
--- a/website/config/environment.js
+++ b/website/config/environment.js
@@ -6,6 +6,8 @@ module.exports = function (environment) {
     environment,
     rootURL: '/',
     locationType: 'history',
+    targetElement: '#main',
+    historySupportMiddleware: true,
 
     EmberENV: {
       FEATURES: {
@@ -18,8 +20,6 @@ module.exports = function (environment) {
       // Here you can pass flags/options to your application instance
       // when it is created
     },
-
-    historySupportMiddleware: true,
 
     fastboot: {
       hostWhitelist: [/^localhost:\d+$/],
--- a/website/config/environment.js
+++ b/website/config/environment.js
@@ -8,7 +8,7 @@ module.exports = function (environment) {
     locationType: 'history',
     historySupportMiddleware: true,
     routerScroll: {
-      targetElement: '#main',
+      targetElement: '#topofpage',
       scrollWhenAfterRender: true,
     },
 
--- a/website/config/environment.js
+++ b/website/config/environment.js
@@ -6,8 +6,11 @@ module.exports = function (environment) {
     environment,
     rootURL: '/',
     locationType: 'history',
-    targetElement: '#main',
     historySupportMiddleware: true,
+    routerScroll: {
+      targetElement: '#main',
+      scrollWhenAfterRender: true,
+    },
 
     EmberENV: {
       FEATURES: {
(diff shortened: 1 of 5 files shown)


Dhaulagiri:
> if I open a new tab and visit a hash URL, it's being removed

I see that too 😢 . Not sure where or when that broke 😬 

This PR fixes _most_ of the issues outlined in the original bug, but still doesn't work in the instance of the navigating to the the accessibility link in the footer. Any idea why that might be?

Dhaulagiri:
> Not sure where or when that broke 😬

Looking back and checking out old code it seems possible this never actually worked 😫

MelSumner:
> > Not sure where or when that broke 😬
> 
> Looking back and checking out old code it seems possible this never actually worked 😫

Ok try it now, I think I had the config incorrect. I *think* it's working locally.

website/config/environment.js
```diff
@@ -6,6 +6,11 @@ module.exports = function (environment) {
     environment,
     rootURL: '/',
     locationType: 'history',
+    historySupportMiddleware: true,
+    routerScroll: {
+      targetElement: '#main',
```
MelSumner, line 11: TODO try to target higher in the page

MelSumner: TODO Make sure to try not targeting anything to see if it works better.

website/config/environment.js
```diff
@@ -38,8 +41,8 @@ module.exports = function (environment) {
   if (environment === 'development') {
     // ENV.APP.LOG_RESOLVER = true;
     // ENV.APP.LOG_ACTIVE_GENERATION = true;
-    // ENV.APP.LOG_TRANSITIONS = true;
-    // ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
+    ENV.APP.LOG_TRANSITIONS = true;
+    ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
```
didoo, line 45: do we need these uncommented? there is  already a lot of noise in the console when building the app, and we may risk to lose important messages there

Dhaulagiri: @MelSumner can we get these commented out again so we can get this PR finished?

website/config/environment.js
```diff
@@ -38,8 +41,8 @@ module.exports = function (environment) {
   if (environment === 'development') {
     // ENV.APP.LOG_RESOLVER = true;
     // ENV.APP.LOG_ACTIVE_GENERATION = true;
-    // ENV.APP.LOG_TRANSITIONS = true;
-    // ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
+    ENV.APP.LOG_TRANSITIONS = true;
```
MelSumner, line 44:
```suggestion
    // ENV.APP.LOG_TRANSITIONS = true;
```

website/config/environment.js
```diff
@@ -39,7 +42,7 @@ module.exports = function (environment) {
     // ENV.APP.LOG_RESOLVER = true;
     // ENV.APP.LOG_ACTIVE_GENERATION = true;
     // ENV.APP.LOG_TRANSITIONS = true;
-    // ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
+    ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
```
MelSumner, line 45:
```suggestion
    // ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
```

Dhaulagiri (approved):

PR #1076, opened 2023-01-18
Responsive cleanup / List Items

jorytindall:
### :pushpin: Summary

if merged this PR adds a breakpoint to assist in the list items in smaller viewports.

The `tokens-list`, `font-helpers-list`, and `icons-list` all break at around less that 500px. I added a new `x-small` breakpoint to account for the smaller viewport (though now the mixins are starting to become a little non-semantic).

All of these pages use relatively the same structure so I mirrored the stacking in all three.

Tokens before
tokens-before

Tokens after
tokens-after

Font helpers before
typography-before

Font helpers after
typography-after

Icons before
icons-before

Icons after
icons-after

### :link: External links

No jira ticket

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/website/app/styles/doc-components/font-helpers-list.scss
+++ b/website/app/styles/doc-components/font-helpers-list.scss
@@ -45,7 +45,8 @@
 
 .doc-font-helpers-list__content {
   padding: 16px;
-  border: solid var(--doc-color-gray-500);
+  border-style: solid;
+  border-color: var(--doc-color-gray-500);
   border-width: 0 1px 1px 1px;
   border-radius: 0 0 3px 3px;
 
--- a/website/app/styles/doc-components/font-helpers-list.scss
+++ b/website/app/styles/doc-components/font-helpers-list.scss
@@ -1,36 +1,58 @@
 // DOC FONT HELPERS LIST
 
 @use "../typography/mixins" as *;
+@use "../breakpoints" as breakpoint;
 
 .doc-font-helpers-list {
-  display: grid;
-  grid-template-columns: minmax(120px, max-content) auto;
+  display: flex;
+  flex-direction: column;
   gap: 16px 0;
   margin: 1.5rem 0;
   padding: 0;
   list-style: none;
+
+  @include breakpoint.medium-and-above () {
+    display: grid;
+    grid-template-columns: minmax(120px, max-content) auto;
+  }
 }
 
 .doc-font-helpers-list__item {
-  display: contents;
+  display: flex;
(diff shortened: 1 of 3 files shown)


heatherlarsen (approved):
I'll let @didoo review the code, but this is a great change visually. Thanks, @jorytindall!

MelSumner:
@jorytindall this feels a little messy to me; if we do ship this now, would you create a backlog ticket to revisit and potentially tidy up things later? Thanks for considering!

didoo (commented):
@jorytindall apart from the comment about avoiding a new breaking point value (and mixin) I am not too worried about the code, for me is ok. one thing I would suggest though is to try to use a mobile-first approach, which is what we have tended to do in other parts of the codebase. Essentially, have the basic behaviour (flex with vertical stacking) by default, and create a media-query that acts above a certain size for the grid-based layout. 

(if you need help in testing this alternative approach, just ping me)

jorytindall:
@didoo I updated the breakpoints in each of these components to be mobile first, I think it works. The one thing that tripped me up is the `display: contents` property which if I understand inherits the grid specs of the larger container? Might be good to merge now, but I'll brush up on the spec.

website/app/styles/doc-components/font-helpers-list.scss
```diff
@@ -1,36 +1,58 @@
...
+    border-radius: 3px 0 0 3px;
+  }
 }
 
 .doc-font-helpers-list__content {
   padding: 16px;
-  border: 1px solid var(--doc-color-gray-500);
-  border-left: none;
-  border-radius: 0 3px 0 3px;
+  border: solid var(--doc-color-gray-500);
```
didoo, line 48: Nitpick: this declaration in interpreted by the browser as
```
border: initial solid var(--doc-color-gray-500);
```
(where the `initial` value in the browser is usually `1px`)

What you intend to do here, I think, is:
```
border-style: solid;
border-color: var(--doc-color-gray-500);
```

didoo (approved):
Left only a minor comment. All good to go.

didoo:
> The one thing that tripped me up is the `display: contents` property which if I understand inherits the grid specs of the larger container? Might be good to merge now, but I'll brush up on the spec.

@jorytindall to make it simple: the `display: contents` property makes the container a "ghost"  in term of layout, it's like a "pass-through" in terms of CSS layout property. Reference: https://developer.mozilla.org/en-US/docs/Web/CSS/display#box

PR #1082, opened 2023-01-18
Website: Don’t render sidebar on pages that don't need it.

MelSumner:
### :pushpin: Summary

If merged, this PR provides one approach to make the sidebar not render if we don't need it for that page.

CleanShot 2023-01-18 at 10 33 02

***

### DIFF
--- a/website/app/styles/pages/application/footer.scss
+++ b/website/app/styles/pages/application/footer.scss
@@ -7,7 +7,7 @@
   // z-index: var(--doc-z-index-footer);
   grid-area: footer;
 
-  body.application:not(.index, .error) & {
+  body.application:not(.index, .error, .patterns) & {
     @include breakpoint.with-fixed-sidebar() {
       padding-left: var(--doc-page-sidebar-width); // we need to compensate for the fixed position of the sidebar
     }
--- a/website/app/styles/pages/application/footer.scss
+++ b/website/app/styles/pages/application/footer.scss
@@ -7,7 +7,8 @@
   // z-index: var(--doc-z-index-footer);
   grid-area: footer;
 
-  body.application:not(.index) & {
+  body.application:not(.index),
+  body.application:not(.error) & {
     @include breakpoint.with-fixed-sidebar() {
       padding-left: var(--doc-page-sidebar-width); // we need to compensate for the fixed position of the sidebar
     }
--- a/website/app/styles/pages/application/footer.scss
+++ b/website/app/styles/pages/application/footer.scss
@@ -7,8 +7,7 @@
   // z-index: var(--doc-z-index-footer);
   grid-area: footer;
 
-  body.application:not(.index),
-  body.application:not(.error) & {
+  body.application:not(.index, .error) & {
     @include breakpoint.with-fixed-sidebar() {
       padding-left: var(--doc-page-sidebar-width); // we need to compensate for the fixed position of the sidebar
     }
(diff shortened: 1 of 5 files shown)


didoo (changes requested):
Have you considered an alternative approach that just uses the `.error` class in the body? 

Something like this, added to the `sidebar.scss` file, seem to work as well:
```
body.application.error {
  .doc-page-sidebar {
    display: none;
  }
}
```

MelSumner:
@didoo good idea, I can slim this down.

didoo:
> @didoo I will try to slim this down, but I think we should try to build less when possible, yeah? It feels a little bit tenuous to use display none like this, if we can prevent it from being built in the first place.
> 
> Why I think it's tenuous: There's this thing where sometimes browsers will say "we know you said display none, but we think some of the content inside is important to assistive tech" and they have their own custom overrides. :(

If you show me a real use case/example, I'm happy to go with this approach. Otherwise better to keep things simple.

website/app/styles/pages/application/sidebar.scss
```diff
@@ -52,17 +52,18 @@ body.application.index {
   }
 }
 
-// ERROR PAGE
+// PAGES THAT DO NOT WANT SIDEBARS
```
MelSumner, line 55: @hashicorp/hds-engineering I think this is the suggested approach but if you look through this commit, you'll see that there are more than a few places where we have to add the new class for any page where we don't want a sidebar. 

The other issue is that the sidebar code is still in the DOM. 

I don't mind it if we feel like we should ship it via hiding in CSS instead of conditional statements in our JS, but it's not the recommended approach.

website/app/styles/pages/application/footer.scss
```diff
@@ -7,7 +7,7 @@
   // z-index: var(--doc-z-index-footer);
   grid-area: footer;
 
-  body.application:not(.index) & {
+  body.application:not(.index, .error, .patterns) & {
```
KristinLBradley, line 10: Question: I wonder if it would make sense to instead add a "show-sidebar" or else "hide-sidebar" type class to certain pages?

KristinLBradley: Not sure if that would simplify anything though...

website/app/styles/pages/application/sidebar.scss
didoo: @MelSumner @KristinLBradley 
- adding more and more classes to the `not()` selection was is not the solution. When the logic for the sidebar was built, was never intended to be used this way. Please, let's not start adding more classes there. 
- If the sidebar should be visible or not, should be controlled via JS and conditionals _IF_ it's necessary. At the moment we have only two use cases:
  - the sidebar in the "patterns" page has already been discussed: we know is a temporary case, and soon (probably even before the launch) will have at least one item, so we can leave it this way. it's not perfect, but it's OK
  - the sidebar in the "error" page is needed, we can't just hide it from the DOM (so my suggestion yesterday is also not entirely correct, needs to be tweaked) because on small screens the sidebar contains the main navigation

As asked yesterday, please remove all the extra code and keep the solution simple. We're spending too much time and energy on a problem that can be solved with a few simple lines of CSS:
```
body.application.error {
  @include breakpoint.with-fixed-sidebar () {
    .doc-page-sidebar {
      display: none;
    }
  }
}
```

Dhaulagiri: @MelSumner can we get this suggestion applied so we can get this change wrapped up?

MelSumner:
Closing this PR, will replace with a new PR shortly.

PR #1085, opened 2023-01-18
Retain input focus when filtering

alex-ju:
### :pushpin: Summary

Retain input focus when performing a filter/search by disabling `ember-a11y-refocus`'s transition validation on two specific paths.

I tried to come up with a more generic solution, but couldn't account for edge cases (for example looking for `searchQuery` to exist wouldn't work in a scenario where the user clears the input to try another search).

### :link: External links

Jira ticket: [HDS-1356](https://hashicorp.atlassian.net/browse/HDS-1356)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-1356]: https://hashicorp.atlassian.net/browse/HDS-1356?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/components/doc/page/header/index.hbs
+++ b/website/app/components/doc/page/header/index.hbs
@@ -1,5 +1,5 @@
 <header class="doc-page-header" ...attributes>
-  <NavigationNarrator />
+  <NavigationNarrator @routeChangeValidator={{@routeChangeValidator}} />
   <LinkTo @route="index" class="doc-page-header__logo" aria-label="home page">
     <Doc::Logo::DesignSystem />
   </LinkTo>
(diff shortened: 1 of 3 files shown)


website/app/components/doc/page/header/index.hbs
```diff
@@ -1,5 +1,5 @@
 <header class="doc-page-header" ...attributes>
-  <NavigationNarrator />
+  <NavigationNarrator @routeChangeValidator={{@routeChangeValidator}} />
```
didoo, line 2: I _think_ that, since the `transitionValidator` function is declared in the `application` controller, then it's available here as `this.transitionValidator` (in which case you don't need to propagate/pass it down via an argument). Can you check?

didoo (approved):
Just a small comment, may be that it can be simplified even more.

website/app/components/doc/page/header/index.hbs
alex-ju: I tried that the first time and it didn't work, unfortunately. I'll give it another go, maybe a missed something.

PR #1089, opened 2023-01-18
Website: Adds a11y audit in acceptance tests

MelSumner:
### :pushpin: Summary

If merged, this PR will add acceptance tests for each page so we can use ember-a11y-testing on our website.

- adds an acceptance test for each page in the website
- adds the a11y audit
- fixes the issues in failing tests
- cleans up some typos and linting failures that I came across

Related PRs:
- https://github.com/hashicorp/design-system/pull/1127
- https://github.com/hashicorp/design-system/pull/1130

***


website/app/components/doc/page/footer.hbs
```diff
@@ -1,6 +1,6 @@
 <footer class="doc-page-footer" ...attributes>
-  <nav class="doc-page-footer__nav-menu" aria-label="footer navigation">
-    <ul class="doc-page-footer__nav-list" role="list">
+  <nav class="doc-page-footer__nav-menu" aria-label="footer">
+    <ul class="doc-page-footer__nav-list" aria-label="footer navigation">
```
MelSumner, line 3: I tested the page with Safari and VoiceOver, and it seems as though a few things needed tweaking: 
1. the bug that doesn't identify a list as a list in Safari no longer seems to exist
2. added aria-label to the list because otherwise it just is called "list" in the rotor and there's no way to differentiate. 
3. removed the redundant navigation from the nav aria-label because it was reading "footer navigation navigation" (oops!)

Sharing context in case it's useful for reviewers.

website/tests/test-helper.js
```diff
@@ -4,9 +4,18 @@ import * as QUnit from 'qunit';
 import { setApplication } from '@ember/test-helpers';
 import { setup } from 'qunit-dom';
 import { start } from 'ember-qunit';
+import { setRunOptions } from 'ember-a11y-testing/test-support';
 
 setApplication(Application.create(config.APP));
 
+// TODO why isn't this working?
```
MelSumner, line 11: This _should_ work, but it doesn't.  Maybe because it's a non-standard prember app? (either of those things would make it challenging)

Dhaulagiri: Is there an error your seeing? Even though we are using prember I don't think there is anything non-standard about how the tests are being run.

MelSumner: There's not an error, there's just...nothing. I tried running it with `PREMBER=true ember s -p 0" and then visiting /tests in my browser.

Dhaulagiri: 🤔  That seems correct since tests shouldn't be part of a production build. Any reason not to `ember s` and then visit `/tests`?

MelSumner: oh! I should have said, I tried that _first_. Then I thought, maybe I need to `PREMBER=true` first. So neither worked for me. 🤔

Dhaulagiri:
If they are simple fixes it would be good to fix the failing tests otherwise I think they should be disabled so we have a clean test suite run

MelSumner:
@Dhaulagiri ok, I'll fix the items.

alex-ju (approved):
This looks good overall! It's a shame we have to have separate files with a similar configuration, but once they're set up I don't see us tweaking them often, so not a burdain from a maintenance perspective.

didoo (changes requested):
@MelSumner added a comment, that I think is a blocker (we need to take a decision as a team if we want to follow this approach for the documentation).

website/docs/components/breadcrumb/partials/code/how-to-use.md
```diff
@@ -24,7 +23,7 @@ A few parameters were omitted for clarity.
 Add the correct `@route/@models/@model/@query` parameter to each Breadcrumb Item.
 
 ```handlebars
-<Hds::Breadcrumb>
+<Hds::Breadcrumb aria-label="breadcrumbs with routing params">
```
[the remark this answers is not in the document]
MelSumner: I would ask you to consider this: We expect our examples to work, we expect our CSS and JavaScript to be correct, why would Accessibility be any different? 

1. If I am a user with a screen reader, and I visit this page, I am presented with a list of breadcrumbs, with no way to differentiate among them. So it becomes a non-equitable discoverability issue: while a user who can see the screen can quickly differentiate among examples and copy the example that is appropriate for them, the user with AT has to go through each example and remember which examples are there, then try to find the one that they want to copy. 
2. I do not think our consumers will be confused by the presence of different aria-labels. The presence of unique labels will reinforce the Success Criteria and provide supplementary learning _through the examples_. 

I hope you will re-consider your point of view.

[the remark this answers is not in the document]
heatherlarsen: When consumers copy and paste the code snippet into their project, do we expect them to keep the `aria-label`, or is this something they would likely remove? 

If it's something they'll remove, is there a way to attach the `aria-label` to the code snippet block (container) instead of directly in the component itself?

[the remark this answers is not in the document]
MelSumner: @heatherlarsen @KristinLBradley there's already an `aria-label` on the component itself (default is `aria-label="breadcrumb"`) so that would take precedence.

MelSumner:
@didoo @Dhaulagiri I've created https://hashicorp.atlassian.net/browse/HDS-1454 to track the additional exploration for reducing the amount of code the developer needs to change from a code sample. I believe this should be sufficient to unblock this PR so we can get it merged before launch, and as such I've requested a re-review.

MelSumner:
After further discussion with @Dhaulagiri, I've removed the aria-label from the code samples and skipped the a11y tests for the breadcrumb page for the time being.

website/docs/components/modal/partials/guidelines/guidelines.md
```diff
@@ -16,8 +16,8 @@
 
 Small
 
-<Hds::Modal @size="small" id="size-small-modal" open as |M|>
-  <M.Header>
+<Hds::Modal @size="small" id="size-small-modal-first" open as |M|>
+  <M.Header as |H|>
```
didoo, line 20:
```suggestion
  <M.Header>
```

website/tests/acceptance/about-test.js
```diff
@@ -0,0 +1,22 @@
+import { module, test } from 'qunit';
+import { visit, currentURL } from '@ember/test-helpers';
+import { setupApplicationTest } from 'website/tests/helpers';
+import { a11yAudit } from 'ember-a11y-testing/test-support';
+
+module('Acceptance | about', function (hooks) {
+  setupApplicationTest(hooks);
+
+  test('visiting /about', async function (assert) {
```
didoo, line 9: _in theory_ we could imagine to get the list of pages at build time (see https://github.com/hashicorp/design-system/blob/main/website/lib/markdown/index.js#L65-L78) and loop over the list, in a single test file, instead of duplicating all the files, one for each page

/cc @Dhaulagiri @alex-ju

didoo (approved):
added a minor comment and a possible idea.

all good! 👍

website/tests/acceptance/about-test.js
MelSumner: Created https://hashicorp.atlassian.net/browse/HDS-1458 to track this 👍

Dhaulagiri: We did discuss this as an idea but decided the manual approach was more pragmatic for the time being

PR #1090, opened 2023-01-18
`Pagination` responsive layout (HDS 971)

KristinLBradley:
### :pushpin: Summary
If merged, this PR will add a responsive layout so that Pagination components will fit better within smaller screens.

### :link: External links

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/app/styles/components/pagination.scss
```diff
@@ -10,19 +10,43 @@
 
 .hds-pagination {
   display: grid;
-  grid-template-areas: "left center right";
+  grid-template-areas: "info nav selector";
```
KristinLBradley, line 13: I renamed these for clarity and find-ability.

packages/components/app/styles/components/pagination.scss
```diff
@@ -10,19 +10,43 @@
...
   grid-template-rows: auto;
   grid-template-columns: 1fr auto 1fr;
   align-items: center;
   margin: 0 auto;
 }
 
+// Responsive layout
```
KristinLBradley, line 20: This is just a simple stacked layout. We can change it from this if we want some other layout instead.

packages/components/app/styles/components/pagination.scss
```diff
@@ -10,19 +10,43 @@
...
+      margin-right: auto;
+      margin-left: auto;
+    }
+  }
+
+  .hds-pagination-nav { order: -1; }
+
+  .hds-pagination-info,
+  .hds-pagination-size-selector {
+    margin-top: 8px;
```
KristinLBradley, line 40: This value can be adjusted if desired and can be exposed as a component-level token (CSS custom property) before release.

packages/components/app/styles/components/pagination.scss
```diff
@@ -10,19 +10,41 @@
...
+// max-width is based on where component layout potentially breaks down
+@media screen and (max-width: 1000px) {
+  .hds-pagination {
+    grid-template-areas:
+      "nav"
+      "info"
+      "selector";
+    grid-template-rows: auto auto auto;
+    grid-template-columns: 1fr;
+    row-gap: 8px;
```
KristinLBradley, line 30: This value can be adjusted if desired and can be exposed as a component-level token (CSS custom property) before release.

packages/components/app/styles/components/pagination.scss
didoo: @KristinLBradley @cveigt what do you think of this approach/layout? It would optimize even more the vertical space (precious in small screens context) and it's also very simple to achieve:

screenshot_2282

packages/components/app/styles/components/pagination.scss
didoo: @KristinLBradley @cveigt what do you think of this approach/layout? It would optimize even more the vertical space (precious in small screens context) and it's also very simple to achieve:

screenshot_2282

KristinLBradley: I will try it, thanks

packages/components/app/styles/components/pagination.scss
```diff
@@ -10,19 +10,44 @@
...
+    order: -1;
+    width: 100%;
+    margin: 0 auto;
+  }
+
+  .hds-pagination-info,
+  .hds-pagination-size-selector {
+    margin-top: 8px;
+    margin-right: 20px;
+    margin-left: 20px;
```
KristinLBradley, line 41: These spacing values can be adjusted and tokenized before release.

packages/components/app/styles/components/pagination.scss
KristinLBradley: I pushed revised layout styles. Screen shots of both layout approaches are included in this PR's description. (Original, 3 row centered & stacked vs. updated, 2 row centered & stacked layouts)

packages/components/app/styles/components/pagination.scss
```diff
@@ -10,19 +10,44 @@
...
+  grid-template-areas: "info nav selector";
   grid-template-rows: auto;
   grid-template-columns: 1fr auto 1fr;
   align-items: center;
   margin: 0 auto;
 }
 
+// Responsive layout
+// max-width is based on where component layout potentially breaks down
+@media screen and (max-width: 1000px) {
```
didoo, line 22: @KristinLBradley have you tried inverting the approach, using `flexbox` by default and switching to a grid above `1000px`? I _think_ the overall code will simplify quite a bit, without the need to override so many properties.

KristinLBradley: Ok, I'll look into that...

KristinLBradley: I cleaned up some of the code a bit but I want to wait until we decide which layout we are going with before I do further optimizations.

packages/components/app/styles/components/pagination.scss
```diff
@@ -4,25 +4,56 @@
...
+
+  .hds-pagination-info,
+  .hds-pagination-size-selector {
+    margin:
+      var(--token-pagination-child-spacing-vertical)
+      var(--token-pagination-child-spacing-horizontal)
+      0
+      var(--token-pagination-child-spacing-horizontal);
+  }
+}
```
KristinLBradley, line 49: I'm grouping all the responsive styles together into one media query for now until we decide which layout we want for narrower screens, then I will optimize the code more.

cveigt (approved):
@KristinLBradley, as discussed I think the first approach (2 row centered & stacked layout) makes a better use of the available space and works even in the smallest devices. Thanks for exploring these options. Looks great 🙌

packages/components/app/styles/components/pagination.scss
KristinLBradley: I broke up the responsive styles and nested them with styles for the elements they affect. To me, it didn't look like it would simplify the code to use flexbox by default (I cleaned up and simplified the styles since I first opened this PR though.)

KristinLBradley:
I merged in the code for the default pageSizes into this PR since it was approved and I had branched off of here originally for that code.

packages/components/app/styles/components/pagination.scss
```diff
@@ -4,27 +4,48 @@
...
 
 .hds-pagination-info {
-  grid-area: left;
+  grid-area: info;
   justify-self: flex-start;
   white-space: nowrap;
+
+  @media screen and (max-width: 1000px) {
+    margin:
+      var(--token-pagination-child-spacing-vertical)
```
didoo, line 44: not a blocker, but probably at this point I would do something like this (also for the selector, only on the other side):

```
.hds-pagination-info {
  ...
  margin-right: var(--token-pagination-child-spacing-vertical);

  @media screen and (max-width: 1000px) {
    margin-top: var(--token-pagination-child-spacing-vertical);
  }
}

because in any case, a right margin for the info is useful, also for the non-mobile version, to avoid the "prev" control and the "info" block to touch. what do you think?
```

packages/components/app/styles/components/pagination.scss
```diff
@@ -33,7 +54,13 @@
 
 .hds-pagination-nav {
   display: flex;
-  grid-area: center;
+  grid-area: nav;
+
+  @media screen and (max-width: 1000px) {
+    justify-content: center;
+    order: -1;
+    width: 100%;
```
didoo, line 62: maybe add a comment here, to explain why this `width: 100%` is needed? (I know it, you know it, but may help the next person that works on this code)

didoo (approved):
left only a couple of minor, non-blocking comments

PR #1102, opened 2023-01-20
Hds 971 pagination default sizes

KristinLBradley:
### :pushpin: Summary
If merged, this PR will add an array of default page sizes so end users will no longer be required to specify an array value unless they wish to override the default.

### :link: External links

Slack thread: https://hashicorp.slack.com/archives/C03A0N1QK8S/p1673546329082759

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/addon/components/hds/pagination/size-selector/index.js
```diff
@@ -17,12 +18,7 @@ export default class HdsPaginationSizeSelectorComponent extends Component {
    * @description Set the page sizes users can select from.
    */
   get sizes() {
-    let { sizes } = this.args;
-
-    assert(
-      '@sizes for "Pagination::SizeSelector" must be defined',
-      sizes !== undefined
-    );
+    let { sizes = DEFAULT_PAGE_SIZES } = this.args;
```
didoo, line 21: I would suggest to declare the default sizes not here, but in the `Pagination::Numbered` component. The advantage in this case is that we can make also the `itemsPerPage` optional, inferring it as first item of the `pageSizes` argument. This will simplify a lot the API of the component (and the documentation itself, now that we don't need to add both values for every instance).

(And the fact that we don't have a default when used as standalone element, is not a problem for me: if the consumer need to build their own pagination using the base elements, defining an array with default sizes is the last of their problems).

KristinLBradley: Pushed updates

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -439,21 +441,19 @@
...
     @code="
-      <Hds::Pagination::Numbered @totalItems=\{{40}} @itemsPerPage=\{{10}} @pageSizes=\{{array 10 30 50}} />
+      <Hds::Pagination::Numbered @totalItems=\{{40}} />
     "
   />
   {{! template-lint-enable no-unbalanced-curlies }}
   {{! prettier-ignore-end }}
   <p class="dummy-paragraph">Renders to:</p>
-  <Hds::Pagination::Numbered @totalItems={{40}} @itemsPerPage={{10}} @pageSizes={{array 10 30 50}} />
+  <Hds::Pagination::Numbered @totalItems={{40}} @itemsPerPage={{10}} />
```
didoo, line 456:
```suggestion
  <Hds::Pagination::Numbered @totalItems={{40}} />
```

packages/components/tests/integration/components/hds/pagination/numbered-test.js
```diff
@@ -24,24 +24,55 @@ module('Integration | Component | hds/pagination/numbered', function (hooks) {
...
+    await render(hbs`
+      <Hds::Pagination::Numbered @totalItems={{100}} @itemsPerPage={{40}} @sizes={{array 20 40 60}} />
+    `);
+    assert.dom('.hds-pagination .hds-pagination-info').hasText('1–40 of 100');
+  });
+
   test('it renders the "nav" content', async function (assert) {
     await render(hbs`
-      <Hds::Pagination::Numbered @totalItems={{100}} @itemsPerPage={{10}} @pageSizes={{array 10 30 50}} />
+      <Hds::Pagination::Numbered @totalItems={{100}} @pageSizes={{array 10 30 50}} />
```
didoo, line 75:
```suggestion
      <Hds::Pagination::Numbered @totalItems={{100}} />
```

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -484,7 +484,6 @@
     @showLabels={{true}}
     @showTotalItems={{false}}
     @showSizeSelector={{false}}
```
didoo, line 486: I think there is a difference between the code snippet and the "renders to:" code. can you check?

packages/components/tests/dummy/app/templates/components/pagination.hbs
```diff
@@ -653,59 +639,33 @@
...
-    @pageSizes={{array 10 30 50}}
-    @showPageNumbers={{false}}
-    @showLabels={{true}}
-  />
+  <Hds::Pagination::Numbered @totalItems={{40}} @showPageNumbers={{false}} @showLabels={{true}} />
 
   <h5 class="dummy-h6">Truncation</h5>
   <p class="dummy-text-small">With large number of pages (truncated by default)</p>
-  <Hds::Pagination::Numbered @totalItems={{100}} @itemsPerPage={{10}} @pageSizes={{array 10 30 50}} />
+  <Hds::Pagination::Numbered @totalItems={{100}} @pageSizes={{array 10 30 50}} />
```
didoo, line 664:
```suggestion
  <Hds::Pagination::Numbered @totalItems={{100}} />
```

didoo (approved):
Just left minor comments and suggestions. All good 👍

packages/components/tests/dummy/app/templates/components/pagination.hbs
KristinLBradley: Fixed it, thanks

KristinLBradley:
I was calling the same argument "pageSizes" when it was used in Pagintion::Numbered and "sizes" when it was used directly in the Pagination::SizeSelector component. I will update these to both be named "pageSizes" for clarity and consistency.

PR #1117, opened 2023-01-23
Website: hide sidebar on error and patterns pages

MelSumner:
### :pushpin: Summary

If merged, this PR hides the sidebar (with CSS)

While we will eventually want to conditionally render layouts based on viewport, for now it's been decided that the sidebar will be hidden with CSS. 

Using the CSS approach that was suggested, here's what the error page looks like: 
CleanShot 2023-01-23 at 13 25 11

***

### DIFF
--- a/website/app/styles/pages/application/sidebar.scss
+++ b/website/app/styles/pages/application/sidebar.scss
@@ -54,6 +54,7 @@ body.application.index {
 
 // ERROR PAGE
 // now with extra specificity
+// also if you want to change this, change footer.scss too
 body.application.error {
   .doc-page-wrapper {
     @include breakpoint.with-fixed-sidebar () {
--- a/website/app/styles/pages/application/sidebar.scss
+++ b/website/app/styles/pages/application/sidebar.scss
@@ -52,19 +52,6 @@ body.application.index {
   }
 }
 
-// ERROR PAGE
-// now with extra specificity
-// also if you want to change this, change footer.scss too
-body.application.error {
-  .doc-page-wrapper {
-    @include breakpoint.with-fixed-sidebar () {
-      aside.doc-page-sidebar {
-        display: none;
-      }
-    }
-  }
-}
-
 // ALL THE OTHER PAGES
 
 body.application:not(.index) {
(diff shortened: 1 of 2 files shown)


MelSumner:
Here's the patterns page after making the footer full-width:

CleanShot 2023-01-23 at 13 44 03

The specificity of the  `body.application:not(.index)` appears to be winning out here: 
CleanShot 2023-01-23 at 13 44 55

didoo:
> The specificity of the `body.application:not(.index)` appears to be winning out here: CleanShot 2023-01-23 at 13 44 55

Increase the specificity of the selector (eg. `body.application.error.error ...` or `body.application.error .doc-page-wrapper .doc-page-sidebar ...`) or use an `!important`. As I said above, the `Patterns` section will have a page very soon, and the only one remaining one will be the `Error` page, so handling it as an exception is OK.

website/app/styles/pages/application/sidebar.scss
```diff
@@ -52,6 +52,18 @@ body.application.index {
   }
 }
 
+// ERROR PAGE
+// now with extra specificity
+// also if you want to change this, change footer.scss too
+body.application.error {
```
didoo, line 58: @MelSumner OK, I was sure it worked when I tested the suggested code:
```
body.application.error {
  @include breakpoint.with-fixed-sidebar () {
    .doc-page-sidebar {
      display: none;
    }
  }
}
```
and now I see where the issue is. 

It's not a problem of specificity:
screenshot_2288
but of order: the declaration for the `Error` page needs to come *after* the more "generic" declaration for the `not:(.index)` pages.

If you change the order, it will work without need to change specificity.

MelSumner: So I guess that has a tradeoff, yeah? Like, do we want the exceptions at the top (maybe easier to find?) then followed by the other things, or the exceptions at the bottom and we all just learn that the exceptions go at the bottom? 

I don't think I have an opinion (it felt like the exceptions were already first so that's where I put them), as long as we are consistent with our own selves. It's one of those things that just gets out of hand really quickly otherwise.

didoo: I don't think there's a pattern or convention to follow, apart from the way CSS works.
In this case, since we want to override a rule, adding if after makes more sense (and doesn't require specificity battles).

didoo (approved):
👍

PR #1119, opened 2023-01-23
Adding `cspell` spell check

jorytindall:
### :pushpin: Summary

**Listed as a draft to compare pros/cons with `markdown-spellcheck`**

If merged this will add `cspell` as a dependency to run semi-automated spell checks on the website content. Several points to note:

### Pros

- `cspell` covers all file types, not just markdown.
- an array of dictionaries can be passed via a config file (`cspell.json`), including a custom dictionary with which you can add words that are specific to your project to ignore.
- not configured in this example, but you can also add forbidden words which will _always_ trigger the checker
- Generally easy to configure and update

### Cons

- does not automatically correct errors, but rather shows you where they are. I think with the complexity/specificity of content in the website, this is better than automatically correcting everything.

In this simple config I've created a custom dictionary (`/project-words.txt`) containing a set of words that we commonly use that might not be part of a traditional dictionary. We can assess whether this list is correct pretty easily.

I've also created a custom dictionary for lorem ipsum (`/dictionary-lorem.txt`) which ignores common placeholder text, which we use quite a bit and was triggering the checker too much.

Run the checker from the website directory with `yarn lint:spell:docs` which will run `cspell` in the docs folder and return errors along with suggestions. This can be customized to whatever directory we want, but I felt it most useful in the docs directory for now. I haven't fixed any of the errors that were thrown intentionally so that the tool could be tested.

### :link: External links

If you are interested, the docs are comprehensive and viewable [here](https://cspell.org/)

Jira ticket: [HDS-1337](https://hashicorp.atlassian.net/browse/HDS-1337)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-1337]: https://hashicorp.atlassian.net/browse/HDS-1337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MelSumner:
@jorytindall after reading both PRs, I would prefer cspell. 
thank you for looking into this! 👍

MelSumner (approved):
💯

alex-ju:
Excellent work, Jory! 🙌

PR #1127, opened 2023-01-24
Adds aria-label to breadcrumb truncation toggle

MelSumner:
### :pushpin: Summary

If merged, this PR adds the missing `aria-label` attribute to the breadcrumb component's truncation button. 

Because the truncation toggle button has only an icon inside of it (hidden with `aria-hidden=true`), the button is required to have an `aria-label` to supply the button element with an accessible name.

This changes nothing that is visually rendered in the browser. 

I've also added a test to make sure the aria-label is always there, just in case the code is refactored in the future.

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/yellow-dolls-arrive.md
+++ b/.changeset/yellow-dolls-arrive.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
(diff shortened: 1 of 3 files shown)


didoo (approved):
👍

PR #1130, opened 2023-01-24
Tabs component: moves the icon and count inside of the button for WCAG conformance

MelSumner:
### :pushpin: Summary

If merged, this PR resolves the issue where the icon was not being rendered inside of the button (with the role of tab) element, causing a WCAG failure.

Failure that was fixed:
CleanShot 2023-01-24 at 10 10 04
 
Screenshot after styles were updated, showing the active/focused tab since that should demonstrate the layout:

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/silly-dots-smile.md
+++ b/.changeset/silly-dots-smile.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-Moves icon inside of tab for WCAG conformance
+Moves icon and count inside of tab (button element) for WCAG conformance
--- a/.changeset/silly-dots-smile.md
+++ b/.changeset/silly-dots-smile.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Moves icon inside of tab for WCAG conformance
--- a/packages/components/app/styles/components/tabs.scss
+++ b/packages/components/app/styles/components/tabs.scss
@@ -66,8 +66,9 @@
     $bottom: var(--token-tabs-tab-focus-inset),
     $left: var(--token-tabs-tab-focus-inset),
   );
-
   position: static;
+  display: flex;
+  align-items: center;
   padding: 0;
   color: inherit;
   background-color: transparent;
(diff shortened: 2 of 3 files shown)


.changeset/silly-dots-smile.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Moves icon inside of tab for WCAG conformance
```
MelSumner, line 5:
```suggestion
Moves icon and count inside of tab (button element) for WCAG conformance
```

KristinLBradley:
This causes a layout shift: https://percy.io/HashiCorp/ds-components/builds/24517146/changed/1370069610?browser=chrome&browser_ids=33%2C34&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

The layout should also be accounted for with this change.

MelSumner:
@KristinLBradley good catch; I'll fix.

KristinLBradley:
Btw, I don't remember exactly, but I think I purposely left the icons outside the button since they didn't seem to make sense as labels for the panels.

MelSumner:
I commented out the styles that expand the click area since I don't _think_ that we need them anymore now that the interactive element contains all of the items, but if someone could confirm that would be helpful!

packages/components/app/styles/components/tabs.scss
```diff
@@ -76,11 +77,12 @@
...
-    position: absolute;
-    content: "";
-    inset: 0;
-  }
+  // I don't think we need this anymore?
+  // &::after {
+  //   position: absolute;
+  //   content: "";
+  //   inset: 0;
+  // }
```
KristinLBradley, line 85: It makes the button click target cover the full "tab" li including the padding and other space around the button in the li wrapper so I think it's still helpful.

MelSumner: ok, I'll put it back in 👍

MelSumner:
> Btw, I don't remember exactly, but I think I purposely left the icons outside the button since they didn't seem to make sense as labels for the panels.

so the icons have aria-hidden on them, so they shouldn't affect the accessible name of the tab at all.

KristinLBradley (approved):
Looking good now! Thanks for fixing this issue

MelSumner:
Yeah, it's the modal component which I definitely didn't touch; maybe just a flaky test, I'll run it again and see.

PR #1147, opened 2023-01-26
style updates to wcag list component

MelSumner:
### :pushpin: Summary

If merged, this PR updates the styles to the wcag list component to make it visually present the same as other lists. 
This is a followup to a previous pr that did some changes but I noticed that it wasn't enough of a change for consistency (compared to https://helios.hashicorp.design/components/form/primitives#formerror) 

Before:
CleanShot 2023-01-26 at 10 35 08

After:
CleanShot 2023-01-26 at 10 35 40

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


didoo (commented):
@MelSumner two quick comments:
- is there any way in which we can leverage the styles provided by the `doc-list-generic` class name (eg adding this class name to the `Doc::WcagList` component) to avoid duplication of certain declarations, and leave only the custom/specific ones?
- for the font styling, we should use one of the existing styles, by using the typography helper class `doc-text-body`

MelSumner:
@didoo ok I removed everything specific to the component and used the generic classes. I _think_ it matches your suggestion now.

website/app/components/doc/wcag-list/index.hbs
```diff
@@ -1,8 +1,8 @@
 <div class={{this.classNames}}>
-  <ul class="doc-wcag-list__list">
+  <ul class="doc-list-generic">
```
didoo, line 2: we need to add the typographic styling (currently works because the element is wrapped in a `<p class="doc-markdown-p">`, but soon - once this https://hashicorp.atlassian.net/browse/HDS-1366 is done - will not work anymore)

```suggestion
  <ul class="doc-list-generic doc-text-body">
```

heatherlarsen (approved):
This is a good compromise, thank you. 🚀

didoo (approved):
👍

PR #1032, opened 2023-01-09
HDS-1383 Add meta tag for twitter site

KristinLBradley:
### :pushpin: Summary

Fix image preview for website links shared in Twitter.

**Preview:** https://hds-website-git-hds-1383-twitter-preview-hashicorp.vercel.app/
**Check Twitter share image:** https://socialsharepreview.com/

### :link: External links

Jira ticket: [HDS-1383](https://hashicorp.atlassian.net/browse/HDS-1383)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable
- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-1383]: https://hashicorp.atlassian.net/browse/HDS-1383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1155, opened 2023-01-30
New docs website / Fix card image fix #2

didoo:
### :pushpin: Summary

Since the previous fix #1032 didn't work (the image is not fingerprinted) let's try to use a static value instead of a config value.

***

### 👀 Reviewer's checklist:

### DIFF
--- a/website/app/templates/head.hbs
+++ b/website/app/templates/head.hbs
@@ -38,19 +38,15 @@ We've made some small customizations to simplify this compared to the default te
   <meta property="og:description" content={{this.model.description}} />
 {{/if}}
 
-{{#if this.model.imgSrc}}
-  <meta property="og:image" content={{this.model.imgSrc}} />
-{{/if}}
+<meta property="og:image" content="https://helios.hashicorp.design/assets/logos/share-card.jpg" />
 
 {{! END OPENGRAPH }}
 
 {{! BEGIN TWITTER }}
 
 <meta name="twitter:card" content="summary_large_image" />
-
-{{#if this.model.imgSrc}}
-  <meta name="twitter:image" content={{this.model.imgSrc}} />
-{{/if}}
+<meta name="twitter:site" content="@HashiCorp" />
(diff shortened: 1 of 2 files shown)


KristinLBradley (approved):

website/config/environment.js
```diff
@@ -34,7 +34,6 @@ module.exports = function (environment) {
       siteName: 'Helios Design System',
       title: 'Helios Design System',
       url: 'https://helios.hashicorp.design',
-      imgSrc: 'https://helios.hashicorp.design/assets/logos/share-card.jpg',
       twitterUsername: '@HashiCorp',
```
MelSumner, line 37:
```suggestion
```

Since we are hardcoding the values we could probably take this out too

didoo: Nice catch! 🤦‍♂️

didoo:
Seems (in theory) to have worked
screenshot_2309
/cc @KristinLBradley @Dhaulagiri

website/app/templates/head.hbs
```diff
@@ -38,19 +38,16 @@ We've made some small customizations to simplify this compared to the default te
   <meta property="og:description" content={{this.model.description}} />
 {{/if}}
 
-{{#if this.model.imgSrc}}
-  <meta property="og:image" content={{this.model.imgSrc}} />
-{{/if}}
+<meta property="og:image" content="https://helios.hashicorp.design/assets/logos/share-card.jpg" />
```
Dhaulagiri, line 41: A side effect of this change is that we are no longer using the logic defined [here](https://github.com/hashicorp/design-system/blob/main/website/app/services/head-data.js#L20) to update the image where we have a more specific one defined for things like components or foundations

didoo: Opened #1158 to take in account also this logic.

didoo: @Dhaulagiri on a side note: that's why I didn't like the idea of having the config data in the `environment.js` file (even it's an Ember pattern). Because now things/logic/definitions are scattered around in multiple places, and one needs to know it, and know where, otherwise some pieces are lost (like in this case).

PR #1156, opened 2023-01-31
Skip flakey test in modal component

alex-ju:
### :pushpin: Summary

Unfortunately one of the callback tests in our Modal component is flakey (it became stable for a while but it now started to become a problem again) so we're going to skip it until we find a stable way to test it.

As reported here: https://hashicorp.slack.com/archives/C025N5V4PFZ/p1675112561810589


MelSumner (approved):
Do we have an issue in the backlog to re-visit?

alex-ju:
> Do we have an issue in the backlog to re-visit?

[Added one now](https://hashicorp.atlassian.net/browse/HDS-1493) – looked into a few skipped tests a while back and now this seems to be the only one around.

PR #1157, opened 2023-01-31
Table: Resolves issue where some CSS classes were malformed

MelSumner:
### :pushpin: Summary

If merged, this PR resolves HDS-1491, fixing an issue where some CSS classes on the table component were malformed.

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/six-snakes-reply.md
+++ b/.changeset/six-snakes-reply.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Fixes malformed CSS classes in table component
--- a/packages/components/addon/components/hds/table/td.js
+++ b/packages/components/addon/components/hds/table/td.js
@@ -37,7 +37,7 @@ export default class HdsTableTdComponent extends Component {
 
     // add a class based on the @align argument
     if (this.align) {
-      classes.push(`.hds-table__td--text-${this.align}`);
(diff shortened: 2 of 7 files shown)


alex-ju (commented):
Is it possible to have a test to check for these classes to prevent future regression?

MelSumner:
I'm thinking about a linting rule that could be applied universally, but I don't have a well-formed solution _just_ yet.

MelSumner:
ahh yes, I suppose that would address this specific issue, but I guess I was thinking more broadly.

Dhaulagiri:
Shouldn't the showcase catch this as well?

alex-ju (approved):
💯

alex-ju:
> Shouldn't the showcase catch this as well?

@Dhaulagiri it was reflected in the showcase – I approved the VRT which showed [the difference in alignment](https://percy.io/HashiCorp/ds-components/builds/24706836/changed/1379864778?browser=chrome&browser_ids=33%2C34&subcategories=approved&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280)

Dhaulagiri:
@alex-ju thanks for that context! More an issue that we didn't notice when the initial showcase was created 👌

PR #1160, opened 2023-01-31
`Hds::Table` - Update API to support custom sorting

MelSumner:
### :pushpin: Summary

If merged, this PR updates the table component so it supports custom sort criteria and adjusts some other small items (per review)

- Adds `sortingFunction` support in `@columns` declaration
- Adds `sortedMessageText` support for custom sorting message
- Updates the `aria-sort` to fallback to "none" instead of "null" (per spec)
- Adds extra showcase examples
- Adds support for the `{{each}}` loop's `key` to be defined with `identityKey` (optional; falls back to the Ember default if none is provided)

Documentation will be added in a separate PR.

### :link: External links

Jira ticket: [HDS-1414](https://hashicorp.atlassian.net/browse/HDS-1414)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-1414]: https://hashicorp.atlassian.net/browse/HDS-1414?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


MelSumner:
I think that I have myself running around in circles and reinventing things that already exist, so I'm trying to take a step back. 

I was able to make the table sort by importance of badge color: 
image

and I did that by adding a key/name to `badge-color`: 
image

Then the invocation looked like this: 
image

We should see if this would work for our consumers, and go from there, because I'm finding myself re-creating functionality that already exists and I think that's a clue that this is going in an incorrect direction.

MelSumner:
I also have some examples where users can craft their own `{{#each}}` in the table body: 

1. this gives them the ability to craft their own [combination of helpers](https://github.com/DockYard/ember-composable-helpers)
2. the tradeoff seems to be that they have to provide their own table header (but they can still build it using the HDS components)

Here's an example from a pairing session during a component intro session: https://github.com/hashicorp/cloud-ui/pull/4410/files

packages/components/tests/integration/components/hds/table/index-test.js
```diff
@@ -243,6 +261,66 @@ module('Integration | Component | hds/table/index', function (hooks) {
...
+  test('it renders a custom sortedMessageText if supplied', async function (assert) {
+    setSortableTableData(this);
+    this.set('sortedMessageText', 'Melanie will sort it');
+
+    await render(hbsSortableTable);
+    await click('#data-test-table .hds-table__th-sort:nth-of-type(1) button');
+    assert.dom('#data-test-table caption').hasText('Melanie will sort it');
+
+    await click('#data-test-table .hds-table__th-sort:nth-of-type(1) button');
+    assert.dom('#data-test-table caption').hasText('Melanie will sort it');
```
didoo, line 273: @MelSumner setting the `sortedMessageText` this way, makes the argument always true, so it's always rendered no matter what. We can test it only once (unless we want to test the relation of the message with the sorting and sorting order, in which case we have to change the way we build the message and make it dependant on the table state).

```suggestion
    await render(hbsSortableTable);
    assert.dom('#data-test-table caption').hasText('Melanie will sort it');
```

packages/components/tests/integration/components/hds/table/index-test.js
```diff
@@ -243,6 +261,66 @@ module('Integration | Component | hds/table/index', function (hooks) {
...
+    await click('#data-test-table .hds-table__th-sort:nth-of-type(1) button');
+    assert
+      .dom('#data-test-table caption')
+      .hasText('A custom caption. Melanie will sort it!');
+
+    await click('#data-test-table .hds-table__th-sort:nth-of-type(1) button');
+    assert
+      .dom('#data-test-table caption')
+      .hasText('A custom caption. Melanie will sort it!');
+  });
```
didoo, line 291: @MelSumner same as above
```suggestion
    await render(hbsSortableTable);
    assert
      .dom('#data-test-table caption')
      .hasText('A custom caption. Melanie will sort it!');
```

packages/components/tests/integration/components/hds/table/index-test.js
```diff
@@ -35,7 +35,12 @@ const setSortableTableData = (context) => {
   ]);
   context.set('columns', [
     { key: 'artist', label: 'Artist', isSortable: true },
-    { key: 'album', label: 'Album', isSortable: true },
+    {
+      key: 'album',
+      label: 'Album',
+      isSortable: true,
+      sortingFunction: 'mySortingFunction',
```
didoo, line 42: @MelSumner this will not work, because you're passing a string, not a function (try to change it to `'foo'` and you will see tests pass anyway)

packages/components/tests/integration/components/hds/table/index-test.js
```diff
@@ -243,6 +261,66 @@ module('Integration | Component | hds/table/index', function (hooks) {
...
+
+    await click('#data-test-table .hds-table__th-sort:nth-of-type(1) button');
+    assert
+      .dom('#data-test-table caption')
+      .hasText('A custom caption. Melanie will sort it!');
+  });
+
+  test('it uses a custom sort function if one is supplied', async function (assert) {
+    const mySortingFunction = (a, b) => {
+      if (a.album < b.album) {
```
didoo, line 295: @MelSumner this is the standard sorting, so it will not detect if it works or not (in fact it was not used by the test, but the tests were passing anyway). I would suggest to use a different (custom) method, like length of the string, or last three characters)

MelSumner: I had a feeling this would be the case. Will update.

packages/components/tests/integration/components/hds/table/index-test.js
```diff
@@ -243,18 +260,80 @@ module('Integration | Component | hds/table/index', function (hooks) {
...
+    this.set('caption', 'A custom caption.');
+    this.set('sortedMessageText', 'Melanie will sort it!');
+
+    await render(hbsSortableTable);
+    assert
+      .dom('#data-test-table caption')
+      .hasText('A custom caption. Melanie will sort it!');
+  });
+
+  test('it uses a custom sort function if one is supplied', async function (assert) {
```
MelSumner, line 282: @hashicorp/hds-engineering this is the test I'm struggling with right now. 

1. It seems kind of useless to write a test that just copies what's in the showcase, because the showcase rendering indicates that the custom function did actually work. Trying to find the right approach for what should be tested.
2. Brian mentioned using Sinon, do we want to add another package? Will it add value?
3. Should we even be testing this?

Really stuck here so feedback is welcome.

didoo: Some considerations:
1. the showcase (in terms of testing) is mainly for visual regression testing, not functional testing which is done via integration and acceptance tests
2. I don't know about sinon, so I can't help with it (my instinct tells me there's a risk to go down a rabbit hole with it)
3. yes, I think so because it's a complex functionality, and we want to make sure that is covered as much as possible by automatic tests, so that if in the future if we have to touch the code of the `Table` we have some safeguard in place

My suggestion is:
- the code I saw this morning, was a good direction apart from the two things I mentioned in my PR: the sorting function provided to the `@columns` should be a function, and it should be a "sorting" function that returns +1/-1/0 depending on the arguments (you can see a hint of it [in the `sort-by` source code](https://github.com/DockYard/ember-composable-helpers/blob/62fb18314e563d70c6588d733df4d26fd4b14aa1/addon/helpers/sort-by.js#L33)) and for this you can use as model the function we've already written for the custom sorting in the showcase: https://github.com/hashicorp/design-system/blob/melsumner/HDS-1414-table-custom-sorting/packages/components/tests/dummy/app/controllers/components/table.js#L28-L38

If you want, I can pair with you and see how to get this test working.

MelSumner: Sure, let's look at it again tomorrow, I think I managed to figure it out, but see what you think.

didoo: Works for me. I'll leave it to you to decide if you want to go with the duplication of code, or use the initial implementation that leveraged the `setSortableTableData` method.

packages/components/app/styles/components/table.scss
```diff
@@ -119,23 +119,31 @@ $hds-table-cell-padding-tall: 20px 16px;
 }
 
 // TABLE DENSITY (TBODY ROW HEIGHT)
+.hds-table--density-short .hds-table__tbody th,
```
didoo, line 122: @MelSumner can you please move these changes to a separate PR, so they're isolated? Thanks

MelSumner: Sure! I thought from your comments elsewhere that you would want them combined. Sorry about that!

packages/components/tests/dummy/app/templates/components/table.hbs
```diff
@@ -169,6 +169,26 @@
       </B.Tr>
     </:body>
   </Hds::Table>
+  <h4 class="dummy-h4">
```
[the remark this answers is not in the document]
MelSumner: oh  ok. Will remove.

packages/components/tests/integration/components/hds/table/index-test.js
didoo: Fixed together in pairing session

didoo (approved):
🚢

PR #1163, opened 2023-02-01
`SideNav` component (HDS-1143)

KristinLBradley:
### :pushpin: Summary
If merged, this PR will add the PageNav component to the design system.

**Showcase Preview:** https://hds-showcase-git-hds-1143-sidenav-component-hashicorp.vercel.app/components/side-nav

* JSON tokens will be released in a later iteration as the potential “dark” palette for the Dropdown hasn't been finalized. (Dropdown content will use current lighter palette for now.)
* Documentation will be released after components are merged.

### :link: External links
- Jira tickets: 
  - [HDS-1143](https://hashicorp.atlassian.net/browse/HDS-1143)
  - [HDS-1578](https://hashicorp.atlassian.net/browse/HDS-1578)
- [Figma file](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/kGo6tmw1Jk4M94x7qJd4FV/HDS-Product---Components?node-id=26174%3A59182&t=iFpO3CCT8qaLo5FN-1)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-143]: https://hashicorp.atlassian.net/browse/HDS-143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HDS-1143]: https://hashicorp.atlassian.net/browse/HDS-1143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HDS-1578]: https://hashicorp.atlassian.net/browse/HDS-1578?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/addon/components/hds/side-nav/index.hbs
```diff
@@ -0,0 +1,39 @@
...
+
+  * SideNav (parent wrapper component)
+    * Header (yielded wrapper component)
+      - Logo / Home Button (child) -- we should probably provide
+      - Help Dropdown (child) -- should we provide? Perhaps allow users just to add custom links?
+      - User Dropdown (child) -- should we provide? Seems like we should
+
+      - Hamburger Button / Close Button -- Need to know more about our planned design
+    * Body (yielded wrapper component)
+      - NavGroup (1 or more, wrapper/parent component)
```
KristinLBradley, line 29: I'll probably rename these "NavGroup" child components because the name is a bit redundant.

packages/components/tests/dummy/app/templates/components/side-nav.hbs
```diff
@@ -0,0 +1,144 @@
...
+  <h3 class="dummy-h3" id="accessibility"><a href="#accessibility" class="dummy-link-section">§</a> Accessibility</h3>
+
+  {{! ADD YOUR CONTENT HERE }}
+
+</section>
+
+<section data-test-percy>
+  <h3 class="dummy-h3" id="showcase"><a href="#showcase" class="dummy-link-section">§</a> Showcase</h3>
+
+  {{! WIP: New SideNav component }}
```
KristinLBradley, line 61: This is the start of what I have working so far:

packages/components/tests/dummy/app/templates/components/side-nav.hbs
```diff
@@ -0,0 +1,144 @@
...
+      <Hds::SideNav::Section @header="Services" as |S|>
+        <S.Item @label="Boundary" @icon="boundary" @href="#" />
+        <S.Item @label="Consul" @icon="consul" @href="#" />
+        <S.Item @label="Packer" @icon="packer" @href="#" />
+        <S.Item @label="Vault" @icon="vault" @href="#" />
+        <S.Item @label="Vault Secrets" @icon="lock" @href="#" />
+        <S.Item @label="Terraform" @icon="terraform" @href="#" />
+        <S.Item @label="Vagrant" @icon="vagrant" @badge="Alpha" @href="#" />
+
+        {{! TODO: Waypoint should have nested links: }}
```
KristinLBradley, line 96: Below are different ideas of what I was considering. I think it probably makes snese to have 2 different types of Items/Link components. One component for a simple link and a different one for a nested link.

I'm not sure yet how the nested link component should work exactly though.

packages/components/addon/components/hds/side-nav/section/item.hbs
```diff
@@ -0,0 +1,27 @@
+<li class="hds-side-nav__list-item">
+  <Hds::Interactive
+    @href={{@href}}
+    @route={{@route}}
+    @models={{hds-link-to-models @model @models}}
+    @query={{hds-link-to-query @query}}
+    @icon={{@icon}}
+    @isRouteExternal={{@isRouteExternal}}
+    @current-when={{@current-when}}
+    class="hds-side-nav__link hds-typography-body-200 hds-font-weight-medium"
```
didoo, line 10: this means that theoretically the font styling _may_ leak in the `Hds::Badge` which we don't want, so to make 100% sure we avoid this, the typographic styling should be applied to a wrapping element (`span` or `div` or something else) of the `@label` text.

In a component world, we should try to avoid as much an possible to rely on inheritance, and prefer a more declarative approach.

packages/components/tests/dummy/app/templates/components/side-nav.hbs
didoo: I am not sure where this is happening in Cloud UI, can you point to a specific page (or screenshot) or code in their codebase where this is a real case? seems so weird to me that they implemented it this way (with nested links)

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,142 @@
...
+  --nav-item-padding-horizontal: 12px;
+}
+
+
+// * SideNav Parent (wrapper) component
+// -------------------------------------------------------------------
+
+.hds-side-nav {
+  // NOTE: Cloud-ui uses grid layout in non-mobile
+  position: fixed; // TODO: should apply only to "mobile" view
```
didoo, line 16: TBD if we should take care of this, or it's their concern to implement the positioning and responsiveness of the component (probably it's ours, but that's not 100% sure yet).
That's why I suggested to start from low-level components (atoms) and worry about the top-level positioning and responsiveness only later. It's easier to solve smaller problems, to gain knowledge on the domain/problem space, and use it later to solve bigger problems.

packages/components/tests/dummy/app/templates/components/side-nav.hbs
didoo: Strong suggestion: add also the single items (the atoms); probably it's something I would do from the very beginning, to make sure I am covering all the possible use cases, variants, states, etc
This also allows to gradually expand the knowledge of how things should work (for example, the "nested links" problem, needs to be solved early to understand what the API of the "item" should look like)

packages/components/addon/components/hds/side-nav/section/item.hbs
```diff
@@ -0,0 +1,27 @@
...
+    ...attributes
+  >
+    {{#if @icon}}
+      <FlightIcon class="hds-side-nav__list-icon" @name={{@icon}} @stretched={{true}} />
+    {{/if}}
+
+    {{@label}}
+
+    {{#if @badge}}
+      <Hds::Badge @text={{@badge}} @type="inverted" @size="small" />
```
didoo, line 20: the badge need to support multiple color variants (you can see it in the Cloud UI codebase)

packages/components/addon/components/hds/side-nav/section/item.hbs
```diff
@@ -0,0 +1,27 @@
...
+    {{/if}}
+
+    {{#if @count}}
```
didoo, line 23: according to [the design specifications](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/kGo6tmw1Jk4M94x7qJd4FV/HDS-Product---Components?node-id=27214%3A64617&t=V4bjvEgUYsqXqFAX-1), the badge count should come before the normal badge: 
screenshot_2349

packages/components/addon/components/hds/side-nav/home-link.hbs
```diff
@@ -0,0 +1,14 @@
+<Hds::Button
```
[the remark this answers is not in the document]
KristinLBradley: Ok thanks I will change it

packages/components/addon/components/hds/side-nav/section/item.hbs
KristinLBradley: ok

packages/components/addon/components/hds/side-nav/section/item.hbs
KristinLBradley: What I tried first was just using a yield in the component which would allow users to add any content including badges inside. Would that be better? If so I can change it back. (I changed the API to add the arguments instead because I thought it would make it easier to use but I wasn't sure what was best.)

packages/components/addon/components/hds/side-nav/section/item.hbs
KristinLBradley: Ok I'll change it

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,142 @@
...
+
+// * Sub-component: Header
+// ---------------------------------------
+.hds-side-nav__header {
+  display: flex;
+  gap: 8px;
+  align-items: center;
+  padding: 12px;
+
+  .hds-side-nav__header-logo {
```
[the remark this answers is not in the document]
KristinLBradley: ok

packages/components/tests/dummy/app/templates/components/side-nav.hbs
KristinLBradley: Cloud ui code isn't set up as a reusable component. I'll look at their code more closely again.

packages/components/app/styles/components/side-nav.scss
KristinLBradley: I commented the positioning out for now. Will probably delete before release but may need to use the positioning to test things like overflowing content.

KristinLBradley:
> @KristinLBradley added a few comments and suggestions.
> 
> what is perplexing me is
> 
> * why we're doing all this work in "isolation" instead of doing it in context (eg in Cloud UI)
> * why we're re-building everything from scratch, while the code already exists and works in Cloud UI
> 
> That's why I suggested at the beginning of the task to continue to work on the playground PR I opened in Cloud UI ([hashicorp/cloud-ui#4701](https://github.com/hashicorp/cloud-ui/pull/4701)) so that we could leverage the existing code, see things in context (and immediately see if the approach taken was working or not), and probably work incrementally (in this special case) using the existing implementation as baseline, and slowly porting it to our HDS conventions and standards
> 
> in any case, I've set up a meeting for later today so we can go through these points and see what is the best way forward

Ok I'll try to go back to work within Cloud-ui.

packages/components/tests/dummy/app/templates/components/side-nav.hbs
KristinLBradley: I'll just start over on this part. I was trying to come up with an API that would be more easy to use and reusable.

packages/components/addon/components/hds/side-nav/section/item.hbs
KristinLBradley: I will redo this to support different variants.

packages/components/tests/dummy/app/templates/components/side-nav.hbs
KristinLBradley: I have a really hard time starting from tiny details and building up. I like to get an overall picture of basic structure then drill down the details as I discover what makes sense.

KristinLBradley:
@didoo Did I mess up the listing of components in the showcase?

https://hds-showcase-git-hds-1143-sidenav-component-hashicorp.vercel.app/

I see just this:
image

packages/components/tests/dummy/app/templates/components/side-nav.hbs
```diff
@@ -0,0 +1,297 @@
+{{page-title "SideNav Component"}}
+
+<Shw::Text::H1>SideNav</Shw::Text::H1>
+
+<section data-test-percy>
+  <Shw::Text::H2>Standalone cases:</Shw::Text::H2>
+
+  <table class="shw-side-nav-table-layout">
```
KristinLBradley, line 8: I'll replace these tables. I just wanted to get everything copied over from Cloud-ui and be sure it was working first.

didoo:
> I'm having a hard time figuring out the scope of exactly what we want to do with the SideNav component. I'm not sure yet how the nested navigation links should work or what kind of API would make sense.

@KristinLBradley I've created a new ticket for this, where I have specified the scope in detail for the first iteration/release:
https://hashicorp.atlassian.net/browse/HDS-1578

> I'm also not sure what we want to do for the responsive behavior. Christian is still working on some of the design.
Out of scope, for this first iteration. 

> Could use some ideas on how to approach this.

Let me know if it's still not clear.

> @didoo Did I mess up the listing of components in the showcase?

No worries, I've pushed a fix.

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+  // Non-active "Links"
+  --token-side-nav-surface-default: #0c0c0e;
+  --token-side-nav-surface-hover: #3b3d45;
+  --token-side-nav-surface-active: #656a76;
+
+  // "Active" "Links"
+  --token-side-nav-selected-surface-default: #656a76;
+  --token-side-nav-selected-surface-hover: #3b3d45;
+  --token-side-nav-selected-surface-active: #656a76;
+}
```
KristinLBradley, line 26: @didoo  I added color tokens based on what @cveigt proposed. I can modify these names if we want changes. 

I still need to add more tokens for margins, padding, etc.

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+// Child "wrapper" blocks & child components:
+
+// * Sub-component: Header
+// ---------------------------------------
+.hds-side-nav__header {
+  display: flex;
+  gap: 8px; // TODO: replace with margins
+  align-items: center;
+  padding: 12px;
+  background: var(--token-side-nav-surface-default); // Q: Should this be added here or should it just be transparent?
```
KristinLBradley, line 50: I started adding these background colors to the various components so they could be displayed more easily as stand alone components in the showcase. I'm not sure if that's a good idea to repeat on all the child elements though vs. relying on the background color of the sidebar container itself.

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+}
+
+.hds-side-nav__list-item {
+  list-style-type: none; // Q: Should ul have role="list" added for accessibility?
+
+  + .hds-side-nav__list-item { margin-top: 4px; }
+}
+
+// generic item
+.hds-side-nav__item {
```
KristinLBradley, line 137: I need to think through some of this naming more. Currently I think it's a bit confusing...

packages/components/app/styles/components/side-nav.scss
didoo: the `__header` should be responsible only of the layout

didoo: re. the showcase, if we need a background for contrast we have to add it via custom CSS in the  CSS for the showcase page

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+  border-radius: 5px;
+
+  &:focus,
+  &.mock-focus {
+    @include hds-focus-ring-with-pseudo-element();
+  }
+}
+
+// Some styles in cloud-ui, interfere so adding additional selector to increase specificity
+body .hds-side-nav__header-logo {
```
didoo, line 81: this code is not anymore in Cloud UI, so we can remove this specificity "hacks"[^1] and move all the declarations under the same CSS selector (`.hds-side-nav__header-logo`)

[^1] this will be done in a different way, using this special Sass variable: https://github.com/hashicorp/structure/blob/main/packages/pds-ember/app/styles/pds/_overrides.scss#L8 which is used then in this way: https://github.com/hashicorp/cloud-ui/blob/main/packages/cloud-ui-core/addon/styles/_overrides.scss#L35)

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+// * Sub-component: Body
+// ---------------------------------------
+.hds-side-nav__body {
+  flex: 1;
+  padding: 12px;
+  overflow: auto;
+}
+
+.hds-side-nav__section {
+  color: var(--token-side-nav-foreground-primary); // Delete?
```
didoo, line 111: yes, for me should be deleted (we don't rely on inheritance in components)

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+  background: var(--token-side-nav-surface-default); // Q: Should this be added here or should it just be transparent?
+
+  + .hds-side-nav__section { margin-top: 28px; }
```
didoo, line 114: easier to read/interpret (otherwise one could have a doubt that it's `.a + .b` instead of `.a + .a`
```suggestion
  & + & { margin-top: 28px; }
```

didoo: plus, in Figma seems `24px` to me

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+}
+
+.hds-side-nav__section-title {
+  padding: 0 var(--token-side-nav-item-padding-horizontal);
+  color: var(--token-side-nav-foreground-faint);
```
didoo, line 119: **this** is the right place to define the color

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+  background: var(--token-side-nav-surface-default); // Q: Should this be added here or should it just be transparent?
```
didoo, line 120: yes, to be removed

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+
+  + .hds-side-nav__list { margin-top: 12px; }
```
didoo, line 122: can we imagine a use case in which the section has just a title? probably not.
so I think we're safe to just apply a `margin-bottom` of `12px` to the `title` itself

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+}
+
+.hds-side-nav__list {
+  margin: 0;
+  padding: 0;
+}
+
+.hds-side-nav__list-item {
+  list-style-type: none; // Q: Should ul have role="list" added for accessibility?
```
didoo, line 131: should this declaration be moved to the `.hds-side-nav__list` directly?

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+
+  + .hds-side-nav__list-item { margin-top: 4px; }
```
didoo, line 133:
```suggestion
  & + & { margin-top: 4px; }
```

packages/components/app/styles/components/side-nav.scss
didoo: I've set up a meeting for later, in which we can come up with a shared naming convention

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// generic item
+.hds-side-nav__item {
+  display: flex;
+  align-items: center;
+  width: 100%;
+  min-height: 2.25rem;
```
didoo, line 141: let's not use `rem` for sizing element. the only place where we're using `rem` is in the typographic design tokens.

packages/components/tests/dummy/app/styles/pages/db-side-nav.scss
```diff
@@ -0,0 +1 @@
+// SIDE-NAV
```
didoo, line 1: this file can be deleted

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+  padding: 0 var(--token-side-nav-item-padding-horizontal);
+  color: var(--token-side-nav-foreground-primary);
+  background: var(--token-side-nav-surface-default); // Q: Should this be added here or should it just be transparent?
+}
+
+// interactive "link" item
+.hds-side-nav__item--link {
+  text-decoration: none;
+  overflow-wrap: break-word;
```
didoo, line 150: 🤔 is this needed?

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+  border-radius: 5px;
+
+  .hds-badge,
+  .hds-badge-count {
+    margin-left: 8px;
+  }
+}
+
+// Some styles in cloud-ui, interfere so adding additional selector to increase specificity
+body .hds-side-nav__item--link:hover,
```
didoo, line 161: same as above, we don't need these specificity hacks in our CSS code, you can safely remove them

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+body .hds-side-nav__item--link:focus,
+body .hds-side-nav__item--link.mock-focus {
+  @include hds-focus-ring-with-pseudo-element();
+}
+
+.hds-side-nav-label { color: var(--token-side-nav-foreground-primary); }
+
+.hds-side-nav__list-icon {
+  width: 1rem;
+  min-width: 1rem;
```
didoo, line 197: no `rem`; no need to use `min-width` (all icons are square).
we can just use explicit `width/height` declarations

packages/components/tests/dummy/app/templates/components/side-nav.hbs
```diff
@@ -0,0 +1,290 @@
+{{page-title "SideNav Component"}}
+
+<Shw::Text::H1>SideNav</Shw::Text::H1>
```
didoo, line 3: I am not reviewing the showcase for now

didoo (commented):
@KristinLBradley added a bunch of comments. we can discuss them later if that's OK for you (can you set up a meeting for today or Monday, when it's better for you? thanks)

packages/components/addon/components/hds/side-nav/section/link.hbs
```diff
@@ -0,0 +1,31 @@
...
+
+    {{#if @count}}
+      <Hds::BadgeCount @text={{@count}} @type="inverted" @size="small" />
+    {{/if}}
+
+    {{#if @badge}}
+      <Hds::Badge @text={{@badge}} @type="inverted" @size="small" />
+    {{/if}}
+
+    {{yield}}
```
didoo, line 29: @KristinLBradley I've just noticed that this component is missing one icon:
modified

didoo (commented):
added another important comment

packages/components/addon/components/hds/side-nav/header.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="hds-side-nav__header">
```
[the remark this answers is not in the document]
KristinLBradley: Sorry I thought I had refactored it previously as you said but I think I lost part of what I did in Cloud-ui with the merge issues I was having and didn't get everything finished in this design system branch yet. I'll look at your other comments, thank you.

(I had also made other changes such as to separate "Link" and "Item" components in Cloud-ui which I lost and redid here.)

packages/components/addon/components/hds/side-nav/home-link.hbs
```diff
@@ -0,0 +1,18 @@
+<Hds::Interactive
+  @href={{@href}}
```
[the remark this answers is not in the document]
KristinLBradley: The link doesn't work. Could you please tell me where to look? Then I'll make a reference for myself.

packages/components/addon/components/hds/side-nav/index.hbs
```diff
@@ -0,0 +1,11 @@
+<div class="hds-side-nav" ...attributes>
```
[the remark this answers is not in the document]
KristinLBradley: I considered it to be similar to the "Pagination" components in which the main component is primarily a wrapper (although the Pagination parent components handle other functionality too.)

packages/components/addon/components/hds/side-nav/section/index.hbs
```diff
@@ -0,0 +1,11 @@
+<nav class="hds-side-nav__section" ...attributes>
+  {{yield (hash Title=(component "hds/side-nav/section/title"))}}
+
+  {{! TODO: If there are no links or items, ul should not be rendered (should consumer add it themselves as a component?) }}
```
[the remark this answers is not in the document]
KristinLBradley: Thanks. I don't want to render an empty ul to the DOM if I can avoid it. I'll look at it more.

packages/components/addon/components/hds/side-nav/section/item.hbs
```diff
@@ -0,0 +1,7 @@
+<li class="hds-side-nav__list-item">
+  <span class="hds-side-nav__item" ...attributes>
+    <span class="hds-side-nav-label hds-typography-body-200">
+      {{yield}}
+    </span>
+  </span>
+</li>
```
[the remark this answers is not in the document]
KristinLBradley: Thanks, yeah I need to look at this one more. Should consumers have to use `SideNav::List::Item` or could it just be an internal component? 

(I was trying to avoid consumers needing to use a separate component to contain the Link, etc. content themselves...)

packages/components/addon/components/hds/side-nav/section/link.hbs
```diff
@@ -0,0 +1,31 @@
+<li class="hds-side-nav__list-item">
+  <Hds::Interactive
+    @href={{@href}}
```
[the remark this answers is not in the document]
KristinLBradley: Could we possibly document the order somewhere it's easy to refer to?

packages/components/addon/components/hds/side-nav/section/link.hbs
```diff
@@ -0,0 +1,31 @@
...
+  >
+    {{#if @icon}}
+      <FlightIcon class="hds-side-nav__list-icon" @name={{@icon}} @stretched={{true}} />
+    {{/if}}
+
+    <span class="hds-side-nav-label hds-typography-body-200 hds-font-weight-medium">
+      {{@label}}
+    </span>
+
+    {{#if @count}}
```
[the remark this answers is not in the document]
KristinLBradley: I was trying to follow what I had done in [Tabs](https://helios.hashicorp.design/components/tabs?tab=code).

packages/components/addon/components/hds/side-nav/section/link.hbs
KristinLBradley: I hadn't looked at how to handle nested link related elements yet. I can add that. It seems like it should be handled with an extra argument? I'll look at it.

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+
+
+// * SideNav Parent (wrapper) component
+// -------------------------------------------------------------------
+
+.hds-side-nav {
+  display: flex;
+  flex-direction: column;
+  width: 280px;
+  height: 100vh;
```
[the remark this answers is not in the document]
KristinLBradley: Ok, how should the width and height be set then?

packages/components/addon/components/hds/side-nav/section/link.hbs
didoo: The component itself (the `::Link` I mean) doesn't know anything about it being nested or not, is just how it's used (for nesting). And yes, it's just a matter of adding another argument (name TBD) and make sure the layout still works.

packages/components/app/styles/components/side-nav.scss
KristinLBradley: I added it on the `li` item because I thought that's what you normally preferred vs adding it to the `ul` which I normally prefer myself. I can change it.

packages/components/app/styles/components/side-nav.scss
KristinLBradley: Ok, I think I got that from Cloud-ui.

packages/components/addon/components/hds/side-nav/home-link.hbs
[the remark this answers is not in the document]
didoo: Mmm, can yo try again? The link works for me (or you can just search for `<Hds::Interactive` in the codebase and see the occurrences there, which is what the link does, only in GitHub)

packages/components/app/styles/components/side-nav.scss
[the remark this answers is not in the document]
didoo: In the showcase, via a custom class applied in the parent container.
In the actual production code, will be set by the parent container wrapper that handles the whole page frame.

packages/components/addon/components/hds/side-nav/section/link.hbs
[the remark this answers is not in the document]
didoo: I see. But probably in this case, since there is also `@badge`, I would make it more obvious on what is what.
Let's discuss anyway later, in the naming session.

packages/components/addon/components/hds/side-nav/section/link.hbs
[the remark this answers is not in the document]
didoo: It's in the codebase, in practically every instance of the `<Hds::Interactive`

packages/components/addon/components/hds/side-nav/home-link.hbs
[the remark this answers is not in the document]
KristinLBradley: I tried a number of times.
image

packages/components/addon/components/hds/side-nav/index.hbs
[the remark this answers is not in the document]
didoo: But in case of the `Pagination`, we have a top-level component that can be used as a single drop-in component handling everything (layout **and** logic). At the moment, is unlikely this will happen for the `SideBar`.
And even if it will be possible, considered the animation + responsiveness, it's very likely that there will be a more general over-arching component (not this one) that will handle the whole.

packages/components/app/styles/components/side-nav.scss
didoo: Mmm. Yeah, I see your point. Let's decide when we figure out how to split the `Item` vs `Link`

packages/components/addon/components/hds/side-nav/home-link.hbs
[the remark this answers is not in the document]
KristinLBradley: Adding for reference:
image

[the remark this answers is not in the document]
alex-ju: Oddly, the link works just fine for me 🤷

packages/components/addon/components/hds/side-nav/section/link.hbs
KristinLBradley: I understand. I'll add that next.

packages/components/addon/components/hds/side-nav/home-link.hbs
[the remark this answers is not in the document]
KristinLBradley: ¯\_(ツ)_/¯

Maybe I'm missing some permission or something.

packages/components/addon/components/hds/side-nav/section/link.hbs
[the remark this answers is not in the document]
KristinLBradley: Going to keep this as @count for now per my understanding from our discussions. We can revisit the question before merging if need be.

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+  color: var(--token-side-nav-foreground-primary); // Delete?
```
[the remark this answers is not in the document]
KristinLBradley: I think we discussed setting a default color here in our meeting so I will leave for now. We can review again before merge. I will clean up the other excess color declarations thought.

packages/components/app/styles/components/side-nav.scss
KristinLBradley: I'll leave for now. Going to focus on making related changes to those at same time to try to avoid confusing myself.

packages/components/app/styles/components/side-nav.scss
KristinLBradley: I'll just delete it

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+  --token-side-nav-item-padding-horizontal: 12px;
+
+  /* Colors (WIP): */
+  --token-side-nav-foreground-primary: #dedfe3;
+  --token-side-nav-foreground-primary-strong: #fff;
+  --token-side-nav-foreground-faint: #8c909c;
+
+  // Non-active "Links"
+  --token-side-nav-surface-default: #0c0c0e;
+  --token-side-nav-surface-hover: #3b3d45;
```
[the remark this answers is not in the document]
KristinLBradley: Began updating color token names along the lines we discussed. Some may still need tweaking.

packages/components/addon/components/hds/side-nav/header.hbs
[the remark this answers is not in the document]
KristinLBradley: This is the current structure of the "header" area (leaving out or simplifying other content):

```
<Hds::SideNav>
	<:header>
		<Hds::SideNav::Header>
			<Hds::SideNav::HomeLink />
			Dropdown
			Dropdown
		</Hds::SideNav::Header>
	</:header>
	
	Extra optional generic content such as search, etc.
</Hds::SideNav>
```

**Questions:**
* Should `<Hds::SideNav>` be renamed to `<Hds::SideNavWrapper>` or similar?
* Should `<Hds::SideNav::Header>` be renamed to `<Hds::SideNav::HeaderLayout>` or similar?
* Should any extra markup or components be added in the header area such as a div wrapper around the `:header` named block to add the padding there instead, etc? 

Current structure of SideNav wrapper component for reference:
```
<div class="hds-side-nav__wrapper" ...attributes>
  {{yield to="header"}}

  <div class="hds-side-nav__body">
    {{yield to="body"}}
  </div>

  <div class="hds-side-nav__footer">
    {{yield to="footer"}}
  </div>
</div>
```

packages/components/app/styles/components/side-nav.scss
KristinLBradley: I've begun refactoring class naming a bit as part of other related changes to the Item & List components but still need to focus on it more next.

packages/components/addon/components/hds/side-nav/section/item.hbs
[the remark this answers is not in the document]
KristinLBradley: I refactored this so that the "Item" component is now the same as the HTML list-item (li) element. It must now be used to wrap Link components.

I still need to work on the "BackLink" component after this.

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,205 @@
...
+    margin-right: auto; // Q: Make more generic?
+  }
+
+  // TEMP HACK ONLY! (TODO: replace)
+  .hds-disclosure__toggle:not(:hover) * {
+    color: var(--token-side-nav-foreground-primary) !important;
+  }
+}
+
+.hds-side-nav__header-logo {
```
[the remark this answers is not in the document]
KristinLBradley: I could use your feedback on related questions I had related to the header structure: https://github.com/hashicorp/design-system/pull/1163#discussion_r1117812402

packages/components/addon/components/hds/side-nav/section/index.hbs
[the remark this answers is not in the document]
KristinLBradley: I couldn't find anyway to do this in Ember.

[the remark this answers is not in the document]
didoo: This is now a non-problem with the new "section" Structure (there's likely always at least one item in the `nav/list`.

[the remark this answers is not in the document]
KristinLBradley: Btw, I held off making those updates to the structure for now until we hear from Matthew Irish again.

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,238 @@
...
+  display: block;
+  align-items: center;
+  justify-content: center;
+  width: 48px;
+  height: 48px;
+}
+
+.hds-side-nav-header__actions {
+  display: flex;
+  gap: 16px;
```
KristinLBradley, line 87: @cristiano Should I replace this `gap` style with margins?

didoo: yes, for now.

KristinLBradley:
@didoo I changed `SideNav::Section::Title` and `SideNav::BackLink` to be wrapped with `SideNav::Section::Item` (the `<li>`). I'm confused about other refactors needed to work with Cloud-Ui's structure now though to make everything part of a list. I guess I need to get rid of/replace `Hds::SideNav::Section` entirely?

didoo:
> @didoo I changed `SideNav::Section::Title` and `SideNav::BackLink` to be wrapped with `SideNav::Section::Item` (the `<li>`). I'm confused about other refactors needed to work with Cloud-Ui's structure now though to make everything part of a list. I guess I need to get rid of/replace `Hds::SideNav::Section` entirely?

Mmm, no. As things are standing now, the idea was that the `Section` (that we may want to rename to `List`?) becomes responsible of providing two semantic elements:
- the `<nav>` element (that needs to be moved out of the `wrapper`) 
- the `<ul>` element

These elements will contain the "items" (be it a generic `item`, a `link`, a `title`, or a `backlink`).

packages/components/app/styles/components/side-nav.scss
KristinLBradley: I replaced it with a in-between margin style. Let me know if you don't like the general selectors I used though.

KristinLBradley:
> > @didoo I changed `SideNav::Section::Title` and `SideNav::BackLink` to be wrapped with `SideNav::Section::Item` (the `<li>`). I'm confused about other refactors needed to work with Cloud-Ui's structure now though to make everything part of a list. I guess I need to get rid of/replace `Hds::SideNav::Section` entirely?
> 
> Mmm, no. As things are standing now, the idea was that the `Section` (that we may want to rename to `List`?) becomes responsible of providing two semantic elements:
> 
> * the `<nav>` element (that needs to be moved out of the `wrapper`)
> * the `<ul>` element
> 
> These elements will contain the "items" (be it a generic `item`, a `link`, a `title`, or a `backlink`).

I pushed up a bunch of changes from refactoring items we discussed. I decided to leave Link and BackLink separate for now because I was worried about overloading the Link component and making it harder to use. Let me know if you think it would be better to combine them after all though.

When you have time, I'd appreciate feedback on everything so far.

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,243 @@
...
+.hds-side-nav__list-item-link {
+  display: flex;
+  align-items: center;
+  width: 100%;
+  min-height: 36px;
+  padding: var(--token-side-nav-body-list-item-padding-vertical) var(--token-side-nav-body-list-item-padding-horizontal);
+  color: var(--token-side-nav-foreground-primary);
+  text-decoration: none;
+  background: var(--token-side-nav-surface);
+  border-color: transparent;
```
didoo, line 179: since we're not setting anywhere the border width, I imagine this is needed because it's a button in some cases, right? in that case, probably better to make it clear/explicit adding a comment

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,243 @@
...
+  border-radius: 5px;
+
+  .hds-badge,
+  .hds-badge-count {
+    margin-left: var(--token-side-nav-body-list-item-content-spacing-horizontal);
+  }
```
didoo, line 185: personally I would move these declarations after the status ones (they refer to the `link` children)

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,243 @@
...
+    background: var(--token-side-nav-interactive-surface-hover);
+    border-color: transparent;
+
+    .hds-side-nav__list-item-label,
+    .hds-side-nav__list-item-icon {
+      color: var(--token-color-foreground-high-contrast);
+    }
+  }
+
+  &:active,
```
didoo, line 203: TBD it may be that later we will need to add also a `.active` class here, but is something we will discuss later with Matthew too

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,243 @@
...
+.hds-side-nav__list-item-icon-chevron {
+  display: block;
+  flex: none;
+  margin-left: var(--token-side-nav-body-list-item-content-spacing-horizontal);
+}
+
+// * Footer child elements
+// ---------------------------------------
+
+// TODO ?
```
didoo, line 243:
```suggestion
```

didoo (commented):
added a bunch of comments. nothing major, so we're in a good place now for targeting a release soon

packages/components/addon/components/hds/side-nav/list/link.hbs
```diff
@@ -0,0 +1,38 @@
...
+    @href={{@href}}
+    @isHrefExternal={{@isHrefExternal}}
+    ...attributes
+  >
+    {{#if @icon}}
+      <FlightIcon class="hds-side-nav__list-item-icon" @name={{@icon}} />
+    {{/if}}
+
+    <span class="hds-side-nav__list-item-label hds-typography-body-200 hds-font-weight-medium">
+      {{@label}}
```
didoo, line 19: @KristinLBradley something I forgot in my review: @meirish suggested to use `@text` instead of `@label` to be more consistent with other HDS components. Makes sense to me. What do you think?

KristinLBradley: That's fine. I'll update it.

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,243 @@
...
+    background: var(--token-side-nav-interactive-surface-hover);
+  }
+
+  &:active,
+  &.mock-active {
+    background: var(--token-side-nav-interactive-surface-active);
+  }
+}
+
+// class name only, no component:
```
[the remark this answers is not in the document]
KristinLBradley: I'll update it to clarify. It means that there isn't a special SideNav Dropdown component, it just gets added to the existing Dropdown component as a classname.

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,243 @@
...
+  > * + * { margin-left: var(--token-side-nav-header-actions-spacing); }
+}
+
+.hds-side-nav__home-link {
+  display: block;
+  width: 100%;
+  height: 100%;
+  padding: var(--token-side-nav-header-home-link-padding); // by design
+  // TODO discuss with Cristiano/Christian if this needs to become a component-level token
+  color: var(--token-color-foreground-high-contrast);
```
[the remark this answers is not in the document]
KristinLBradley: I'm going to use: `--token-side-nav-foreground-primary-strong` which makes sense to me. We can discuss and change it later if desired.

packages/components/addon/components/hds/side-nav/list/index.hbs
```diff
@@ -0,0 +1,11 @@
+<nav class="hds-side-nav__list-wrapper" ...attributes>
+  <ul class="hds-side-nav__list">
```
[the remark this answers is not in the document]
KristinLBradley: @MelSumner The li's of this list are styled with `list-style-type: none;`. Should `role="list" be added to the wrapping `ul`? And/or an aria role?`

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,243 @@
...
+}
+
+// * Body child elements
+// ---------------------------------------
+.hds-side-nav__body {
+  flex: 1;
+  overflow-y: auto;
+}
+
+.hds-side-nav__list-wrapper {
```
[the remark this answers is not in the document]
KristinLBradley: Ok I will update.

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,243 @@
...
+  // TODO: this doesn't work because Ember adds whitespace inside the ul
+  &:empty { display: none; }
+}
+
+.hds-side-nav__list-item {
+  display: flex;
+  align-items: center;
+  list-style-type: none; // Q: Should ul have role="list" added for accessibility?
+
+  &:last-child { margin-bottom: 0; }
```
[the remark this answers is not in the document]
KristinLBradley: You're right. This is left over from an earlier iteration.

packages/components/addon/components/hds/side-nav/list/back-link.hbs
```diff
@@ -0,0 +1,19 @@
+<Hds::SideNav::List::Item>
+  <Hds::Interactive
+    class="hds-side-nav__list-item-link hds-side-nav__list-item-link--back-link"
```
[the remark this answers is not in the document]
KristinLBradley: Would like to discuss more if this needs any changes. The Cloud-ui treatment is very confusing...

packages/components/tests/integration/components/hds/side-nav/link-test.js
```diff
@@ -0,0 +1,52 @@
...
+      hbs`<Hds::SideNav::List::Link @icon="boundary" @label="Boundary" @count="3" @badge="Alpha" @hasSubItems={{true}} />`
+    );
+    assert.dom('.flight-icon-boundary').exists();
+    assert.dom('.hds-side-nav__list-item-label').hasText('Boundary');
+    assert.dom('.hds-badge-count').hasText('3');
+    assert.dom('.hds-badge').hasText('Alpha');
+    assert.dom('.hds-side-nav__list-item-icon-chevron-wrapper').exists();
+  });
+
+  test('it renders the passed in custom content', async function (assert) {
```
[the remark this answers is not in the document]
KristinLBradley: I simplified it a bit but would like to keep more realistic passed in content.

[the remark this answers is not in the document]
didoo: can you expand on why? If a component yields, it yields whatever children we pass, be it a simple empty `<div>`, be it a complex set of components. For me we're reducing readability and adding complexity, for no real benefit (from a computer science perspective).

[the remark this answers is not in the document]
KristinLBradley: I'll just change it.

packages/components/addon/components/hds/side-nav/list/title.hbs
```diff
@@ -0,0 +1,5 @@
+<Hds::SideNav::List::Item>
+  <div class="hds-side-nav__list-title hds-typography-body-100 hds-font-weight-semibold">
+    {{yield}}
+  </div>
```
didoo, line 4: @KristinLBradley something I noticed in the Cloud UI version of the `SideNav`: they strip away the whitespace for the "header/title", probably better if we do the same (they must have had good reasons)

screenshot_2462

```suggestion
  <div class="hds-side-nav__list-title hds-typography-body-100 hds-font-weight-semibold">{{~yield~}}</div>
```

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,259 @@
...
+  // Interactive elements:
+  // base:
+  --token-side-nav-interactive-surface-hover: #3b3d45;
+  --token-side-nav-interactive-surface-active: #656a76;
+
+  // selected / current:
+  --token-side-nav-interactive-selected-surface-default: #656a76;
+  --token-side-nav-interactive-selected-surface-hover: #3b3d45;
+  --token-side-nav-interactive-selected-surface-active: #656a76;
+}
```
KristinLBradley, line 44: @cveigt Above are the tokens I defined if you would like to review. Probably the most relevant ones for you are the colors.

packages/components/addon/components/hds/side-nav/list/link.hbs
```diff
@@ -0,0 +1,38 @@
...
+      <Hds::BadgeCount @text={{@count}} @type="inverted" @size="small" />
+    {{/if}}
+
+    {{#if @badge}}
+      <Hds::Badge @text={{@badge}} @type="inverted" @size="small" />
+    {{/if}}
+
+    {{yield}}
+
+    {{#if @hasSubItems}}
```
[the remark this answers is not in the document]
KristinLBradley: @cveigt We should discuss this

packages/components/addon/components/hds/side-nav/list/index.hbs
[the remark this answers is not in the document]
MelSumner: @KristinLBradley correct, it should have a role added when the list-style is changed. This is so Safari will identify it correctly as a list.

packages/components/addon/components/hds/side-nav/list/link.hbs
```diff
@@ -0,0 +1,38 @@
...
+    class="hds-side-nav__list-item-link"
+    @current-when={{@current-when}}
+    @models={{hds-link-to-models @model @models}}
+    @query={{hds-link-to-query @query}}
+    @replace={{@replace}}
+    @route={{@route}}
+    @isRouteExternal={{@isRouteExternal}}
+    @href={{@href}}
+    @isHrefExternal={{@isHrefExternal}}
+    ...attributes
```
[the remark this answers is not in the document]
MelSumner: I think that Ember already adds an active class automatically 🤔

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,243 @@
...
+  }
+
+  &:active,
+  &.mock-active {
+    background: var(--token-side-nav-interactive-surface-active);
+  }
+}
+
+// class name only, no component:
+.hds-side-nav__dropdown {
```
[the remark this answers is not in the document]
MelSumner: Is there a way to confirm that if this component is on the same page as a regular dropdown, we don't have style conflicts?

packages/components/addon/components/hds/side-nav/list/index.hbs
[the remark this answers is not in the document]
MelSumner: Also a suggestion to add a test that ensures the role stays there (just in case of refactoring later by someone else)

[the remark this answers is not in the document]
KristinLBradley: Updated to add the role to the `ul` and also to add tests for the List component including a check that it has `role="list"`

packages/components/addon/components/hds/side-nav/list/back-link.hbs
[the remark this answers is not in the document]
didoo: This has now been fixed.

packages/components/addon/components/hds/side-nav/button.hbs
```diff
@@ -0,0 +1,14 @@
...
+  class="hds-side-nav__button"
+  @current-when={{@current-when}}
+  @models={{hds-link-to-models @model @models}}
+  @query={{hds-link-to-query @query}}
+  @replace={{@replace}}
+  @route={{@route}}
+  @isRouteExternal={{@isRouteExternal}}
+  @href={{@href}}
+  @isHrefExternal={{@isHrefExternal}}
+  ...attributes
```
didoo, line 11: @KristinLBradley if you look at the `Hds::Button`, you will see that when it's icon-only then a `<span>` containing required `@text` is displayed for screen-readers only. please sync with @MelSumner if she prefers this approach here too, or a (required) `aria` attribute is enough to describe the button for screen-readers

packages/components/addon/components/hds/side-nav/list/link.hbs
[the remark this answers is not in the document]
didoo: yes, but this is in case in which they need to apply an "active" state programmatically (eg. it's a button, so Ember doesn't apply the `active` class; or the `current-when` doesn't work for some reasons)

packages/components/addon/components/hds/side-nav/list/link.hbs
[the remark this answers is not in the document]
didoo: done, resolving this

packages/components/app/components/hds/side-nav/button.js
```diff
@@ -0,0 +1 @@
+export { default } from '@hashicorp/design-system-components/components/hds/side-nav/button';
```
didoo, line 1: TBD if we want to have it under `header` to avoid confusion (otherwise consumers may think it can be used in any part of the sidenav)

packages/components/app/styles/components/side-nav.scss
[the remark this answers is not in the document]
didoo: in the current implementation, the class `hds-side-nav__dropdown` is applied directly to the Dropdown element (as additional class); later once the "dark" styles are finalized, we will use the parent `hds-side-nav-header__actions` container as selector, to avoid the consumers having to manually add the class themselves

in both cases, no risks of style leaking to other dropdowns

packages/components/app/styles/components/side-nav.scss
didoo: done, resolving this

packages/components/addon/components/hds/side-nav/button.hbs
```diff
@@ -0,0 +1,14 @@
...
+>
+  <FlightIcon @name={{@icon}} @color={{@color}} @stretched={{true}} />
```
didoo, line 13: important: the icon is large, so we need a large glyph

```suggestion
  <FlightIcon @name={{@icon}} @color={{@color}} @stretched={{true}} @size="24" />
```

packages/components/app/components/hds/side-nav/button.js
KristinLBradley: Would it necessarily be bad if they used it in another part of the SideNav? What if, for example, they didn't use a `List` in the body section and instead wanted custom content including a dropdown for some reason?

packages/components/addon/components/hds/side-nav/button.hbs
KristinLBradley: It looks like `Hds::Button` just uses an aria-label to include text when it's icon-only unless I'm looking at it wrong...

Should I make the aria-label required? (I think it should be. Or else a hidden span with accessible text.)

packages/components/app/components/hds/side-nav/button.js
didoo: Decided makes sense to leave it as more generic element

packages/components/addon/components/hds/side-nav/button.hbs
didoo: Fixed

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,267 @@
...
+  width: 100%;
+  height: 100%;
+  color: var(--token-side-nav-foreground-primary); // Default color
+  background: var(--token-side-nav-surface);
+}
+
+.hds-side-nav__wrapper-header {
+  padding-top: var(--token-side-nav-wrapper-padding-vertical);
+  padding-right: var(--token-side-nav-wrapper-padding-horizontal);
+  padding-bottom: 8px; // by design
```
KristinLBradley, line 94: @didoo Should a token be added for the the padding-bottom value?

KristinLBradley: I decided to go ahead and just add a token since to me it makes sense.

cveigt:
@KristinLBradley, I got a few minor comments:

- Spacing between the dropdowns/actions should be 8px. Currently, it's set to 10px.

Screen Shot 2023-03-12 at 11 56 35 PM

- Spacing/margin between list items should be 2px to align with TFC/HCP and the Figma component.

Screen Shot 2023-03-13 at 12 00 51 AM

- Regarding the font-weight looking bolder than in Figma. Would it be possible to try applying font-smoothing or some other property that could help smooth the font?

KristinLBradley:
> 

I updated the spacing. Regarding the font-weight, I'm applying these 2 classes to style the ext currently: 
- `hds-typography-body-200` 
- `hds-font-weight-medium`

I really wouldn't want to add one-off styling to these items. Have we had this issue before or is there something I could possibly be missing? (@didoo could you please take a look also?)

packages/components/app/styles/components/side-nav.scss
KristinLBradley: @didoo @cveigt I'd like to confirm with you both that these tokens are what we want. Please let me know also if there are any others I should add.

After these are confirmed I will create the JSON file for them.

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,274 @@
...
+  --token-side-nav-foreground-faint: #8c909c;
+  --token-side-nav-surface: #0c0c0e;
+
+  // Interactive elements:
+  // base:
+  --token-side-nav-interactive-surface-hover: #3b3d45;
+  --token-side-nav-interactive-surface-active: #656a76;
+
+  // selected / current:
+  --token-side-nav-interactive-selected-surface-default: #656a76;
```
didoo, line 43: probably I would use the `active` term, at this point, instead of `selected` (which is not used anywhere else, as far as I remember)

packages/components/app/styles/components/side-nav.scss
```diff
@@ -0,0 +1,274 @@
...
+  --token-side-nav-body-list-item-content-spacing-horizontal: 8px;
+  --token-side-nav-body-list-item-border-radius: 5px;
+
+  // Colors:
+
+  // SideNav container:
+  --token-side-nav-foreground-primary: #dedfe3;
+  --token-side-nav-foreground-primary-strong: #fff;
+  --token-side-nav-foreground-faint: #8c909c;
+  --token-side-nav-surface: #0c0c0e;
```
didoo, line 35: not a big fan of this token (too generic). maybe `surface-primary` for consistency?

packages/components/app/styles/components/side-nav.scss
didoo: @KristinLBradley added a couple of comments, but nothing that would block the release (we can always update the names later, if needed)

didoo:
> I really wouldn't want to add one-off styling to these items. Have we had this issue before or is there something I could possibly be missing? (@didoo could you please take a look also?)

@cveigt @KristinLBradley in _theory_ we could apply the font smoothing to the `hds-side-nav__list-item-text` element/selector, but I would prefer not to, unless strictly necessary: this functionality [is not well supported](https://caniuse.com/?search=font%20smooth) by the browsers, and leads to different rendering for different browsers, it's not something on which we have a real control.

screenshot_2485

packages/components/app/styles/components/side-nav.scss
KristinLBradley: I updated the token names but I realized I don't actually use these tokens anywhere currently. Where should I be using them. For just the `.hds-side-nav__list-item-link` states?

didoo: let's just delete them for now, we will review anyway all the tokens (I suspect) once there's more clarity on the dark/override colors for the dropdown

KristinLBradley:
> > I really wouldn't want to add one-off styling to these items. Have we had this issue before or is there something I could possibly be missing? (@didoo could you please take a look also?)
> 
> @cveigt @KristinLBradley in _theory_ we could apply the font smoothing to the `hds-side-nav__list-item-text` element/selector, but I would prefer not to, unless strictly necessary: this functionality [is not well supported](https://caniuse.com/?search=font%20smooth) by the browsers, and leads to different rendering for different browsers, it's not something on which we have a real control.
> 
> screenshot_2485

What have we done in the past for this?

As for myself, I don't usually expect image editing software to render fonts the same as in the browser so in the past I usually didn't worry too much about minor differences. I don't know what we normally do in regards to issues like that though. 

(I found this related link btw: https://forum.figma.com/t/why-does-a-font-weight-in-figma-seem-lighter-than-the-same-weight-in-the-browser/2207)

cveigt (approved):
@KristinLBradley it looks great 🙌  I only have a minor piece of feedback.
Could we change the size of the flight icon inside `hds-side-nav-header__logo` from 42px to 40px? The size of the container and target are correct, it's only the logo (HCP icon) that will scale down

Screen Shot 2023-03-14 at 4 28 36 PM

didoo (approved):
Good to go! 🚢 🚀

PR #1140, opened 2023-01-25
Allow deep linking

alex-ju:
### :pushpin: Summary

Allow deep linking – linking to headings or sections using URI fragments.

Preview link: https://hds-website-git-alex-ju-allow-fragment-navigation-hashicorp.vercel.app/components/alert#usage

### :hammer_and_wrench: Detailed description

There are two commits in this PR:
 - one that identifies when the fragment gets removed (there's a tradeoff here)
 - another one that proposes a solution for automatic scroll to heading/section, if the `id` exists on the page

Note: code is rather snappy, but wanted to get it up to get feedback

### :link: External links

Jira ticket: [HDS-1419](https://hashicorp.atlassian.net/browse/HDS-1419)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable
- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-1419]: https://hashicorp.atlassian.net/browse/HDS-1419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1167, opened 2023-02-02
New docs website - Post launch / Add inline anchor links for headers

didoo:
### :pushpin: Summary

This PR, built on top of #1140, adds visible (on hover/focus) anchor links to the headings in the page.

In this PR I have:
- updated logic that handles the generation of the sections TOCs at runtime
    - now we add anchor links to levels H2-H5, but we list only H2-H3 in the sidecar
- applied styling to the injected anchor links

👉 👉 👉 **Preview**: https://hds-website-git-post-launch-add-inline-headers-5ce71a-hashicorp.vercel.app/components/alert

Notice: after meeting with @heatherlarsen and @MelSumner these decision were taken:
- we will keep the anchor appear only on mouse over (or tap on the heading, on touch devices) - reason: having it always visible would clutter too much the page (we have also received feedback about this on the "scrappy" website)
- we will keep the size of the anchor element fixed (and not relative to the font-size via `em`) and we will define specific margin-top values for the different headings level to define the vertical/optical alignment with the text (not perfect, but acceptable); in this way we can be more in line with the design specs (and to Developer implementation)
- still to decide what to do on small viewports with touch (smaller anchor size or anchor shifting the text to the right)

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-1221

***

### 👀 Reviewer's checklist:

### DIFF
--- a/website/app/controllers/show.js
+++ b/website/app/controllers/show.js
@@ -8,18 +8,33 @@ import { inject as service } from '@ember/service';
 
 import { showdownConfig } from '../shared/showdown-config';
 
-const getTOCs = (container) => {
+const getAnchoredHeadings = (container) => {
   let headings = [];
-  container.querySelectorAll(`:scope > h2, :scope > h3`).forEach((element) => {
-    // we need to add a class to avoid the element being hidden behind the fixed top header
-    element.classList.add('doc-page-sidecar-scroll-margin-top');
-    // we add it to the list of headings used as TOC in the sidecar
-    headings.push({
-      target: element.id,
-      text: element.innerText,
-      depth: element.tagName.replace(/h/i, ''),
+  container
+    .querySelectorAll(`:scope > h2, :scope > h3, :scope > h4, :scope > h5`)
+    .forEach((element) => {
+      // we add an anchor link to the heading
+      if (
+        element.id &&
+        // notice: we use the class name as a way to detect if the anchor is already been added
+        // reason: `didInsertContent` is run at each tab click (something we probably want to look into in a separate ticket)
+        !element.classList.contains('doc-page-heading-scroll-margin-top')
+      ) {
+        const anchor = document.createElement('a');
+        anchor.href = `#${element.id}`;
+        anchor.className = 'doc-page-heading-link';
+        anchor.setAttribute('aria-label', element.innerText);
+        element.prepend(anchor);
+      }
+      // we need to add a class to avoid the element being hidden behind the fixed top header
+      element.classList.add('doc-page-heading-scroll-margin-top');
+      // we add it to the list of headings (a subset will be used as TOC in the sidecar)
+      headings.push({
+        target: element.id,
+        text: element.innerText,
+        depth: element.tagName.replace(/h/i, ''),
+      });
     });
-  });
   return headings;
 };
 
@@ -119,7 +134,8 @@ export default class ShowController extends Controller {
         tocs.push({
           index,
           id: `toc-${name}`,
-          list: getTOCs(section),
+          // we show only the headings level 2 and 3 in the sidecar
+          list: getAnchoredHeadings(section).filter((item) => item.depth <= 3),
         });
       });
     } else {
@@ -127,7 +143,7 @@ export default class ShowController extends Controller {
       tocs.push({
         index: 0,
         id: 'toc-all',
-        list: getTOCs(container),
+        list: getAnchoredHeadings(container),
       });
     }
 
--- a/website/app/styles/pages/application/content.scss
+++ b/website/app/styles/pages/application/content.scss
@@ -57,21 +57,54 @@
 }
 
 .doc-page-heading-link {
+  $box-size: 24px;
+  $icon-size: 16px;
+  $gap: 8px;
   display: inline-block;
-  width: 24px;
-  height: 1em;
-  margin-top: 0.15em;
-  margin-left: -24px;
+  width: $box-size;
+  height: $box-size;
+  margin-right: $gap;
+  margin-left: -($box-size + $gap);
   vertical-align: text-top;
-  // background-color: red;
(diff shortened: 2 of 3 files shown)


heatherlarsen:
@didoo Can you provide more detail about what the issue or concern is with the current designs for this? Or would you rather I set up a sync to discuss? I'm seeing both placement and visual design were not met so hoping we can get a little bit closer to the proposed design before moving forward.

website/app/controllers/show.js
```diff
@@ -8,18 +8,27 @@ import { inject as service } from '@ember/service';
 
 import { showdownConfig } from '../shared/showdown-config';
 
-const getTOCs = (container) => {
+const getAnchoredHeadings = (container) => {
```
MelSumner, line 11: [a11y, blocker] links must have discernible text (accessible name) 

(FWIW, this was done incorrectly on the scrappy site too)

website/app/styles/pages/application/content.scss
```diff
@@ -43,3 +43,30 @@
...
+  height: 1em;
+  margin-top: 0.15em;
+  margin-left: -24px;
+  vertical-align: text-top;
+  // background-color: red;
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='16' height='16' fill='none' viewBox='0 0 16 16'%3E%3Cg fill='%232264D6'%3E%3Cpath d='M11.141 1a3.893 3.893 0 00-2.738 1.093L7.314 3.175A.75.75 0 008.372 4.24l1.077-1.07a2.393 2.393 0 013.384 3.382l-1.881 1.88a2.393 2.393 0 01-3.608-.258.75.75 0 00-1.202.899 3.893 3.893 0 005.87.42l1.886-1.886.01-.009A3.893 3.893 0 0011.14 1z'/%3E%3Cpath d='M7.019 5.365a3.893 3.893 0 00-3.032 1.13L2.102 8.382l-.01.01a3.893 3.893 0 005.505 5.504l1.084-1.084a.75.75 0 00-1.06-1.06l-1.07 1.07a2.393 2.393 0 01-3.384-3.384l1.881-1.88a2.393 2.393 0 013.609.258.75.75 0 101.2-.899A3.893 3.893 0 007.02 5.365z'/%3E%3C/g%3E%3C/svg%3E");
+  background-repeat: no-repeat;
+  background-position: center center;
+  background-size: 16px 16px;
+  opacity: 0;
```
MelSumner, line 66: [a11y, blocker] Interactive elements cannot be hidden.

MelSumner (changes requested):

MelSumner:
I think something like this could work re: a11y

```html
<div class="section-title">
  <a href="#section-accessibility" aria-label="accessibility"><span aria-hidden="true">§</span></a>
  <h2 id="section-accessibility">Accessibility</h2>
</div>
```

MelSumner:
@didoo if you keyboard navigate a components page (I loaded the page, pressed ENTER on the "Skip to content" link, and TAB navigated from there) you will see this: 
CleanShot 2023-02-22 at 16 02 44

didoo:
> @didoo if you keyboard navigate a components page (I loaded the page, pressed ENTER on the "Skip to content" link, and TAB navigated from there) you will see this: CleanShot 2023-02-22 at 16 02 44

@MelSumner this is how the browser renders the content, because they're in the DOM, in the `tabpanel`, but with `opacity = 0`. In fact if you change slightly the position you will see this:
screenshot_2402
I don't think there's much we can do, unless we change from `opacity=0/1` to `display=none/inline-block`, but in that case I imagine they would not be accessible.

KristinLBradley:
> > @didoo if you keyboard navigate a components page (I loaded the page, pressed ENTER on the "Skip to content" link, and TAB navigated from there) you will see this
> 
> @MelSumner this is how the browser renders the content, because they're in the DOM, in the `tabpanel`, but with `opacity = 0`. In fact if you change slightly the position you will see this: screenshot_2402 I don't think there's much we can do, unless we change from `opacity=0/1` to `display=none/inline-block`, but in that case I imagine they would not be accessible.

@melanie @didoo I think it would help to define the desired expected visual behavior. For example, should the entire Panel have a blue outline when you navigate to it via keyboard? I believe when I was working on HDS Tabs that we didn't want that visual behavior... So maybe we could use HDS Tabs as an example of the desired behavior?

heatherlarsen:
@KristinLBradley Good call out. I would expect this to behave the same way the HDS `tabs` component behaves, where the first interactive element receives focus, not the whole panel.

MelSumner (approved):
Noting that it's not the best a11y that interactive elements are hidden, but we've discussed why we are choosing not to do it at this time. 

Otherwise, this seems good to me. the a11y seems to otherwise checkout too. 👍

didoo:
@heatherlarsen @MelSumner @KristinLBradley I have reverted the second version and polished the first version, which from what I understand from the conversation in Figma is what we want. can you please check it and see if it's OK for you?
https://hds-website-git-post-launch-add-inline-headers-5ce71a-hashicorp.vercel.app/components/alert

some considerations:
- this may not perfect, but is what we can do in our context (other suggested solutions had different DOM structures, essentially based on a parent container that we don't have in our case, so the icon can stay or on the left, or on the right, but not in both places depending on the viewport)
- for the "mobile" viewport, I am hiding the focus outline, because we're likely in a touch environment, so there's no concept of focus (or hover)

> @KristinLBradley Good call out. I would expect this to behave the same way the HDS `tabs` component behaves, where the first interactive element receives focus, not the whole panel.

Yes, this will be done in a distinct ticket/task as follow-up.

/cc @Dhaulagiri

heatherlarsen:
@didoo I worry about how early we're dropping the visible focus, considering folks may shrink the browser down to place windows side by side. Can we lower that threshold to something more like 414px or 428px ([ref](https://gs.statcounter.com/screen-resolution-stats/mobile/worldwide))? I realize this results in there being no space between the anchor and the heading, but I think that's fine, given we've exhausted all other options. 

Beyond that, I'm ok with this direction from a visual and interaction perspective. I'm not sure how accessible it is though. cc @MelSumner

didoo:
> @didoo I worry about how early we're dropping the visible focus, considering folks may shrink the browser down to place windows side by side. Can we lower that threshold to something more like 414px or 428px ([ref](https://gs.statcounter.com/screen-resolution-stats/mobile/worldwide))? I realize this results in there being no space between the anchor and the heading, but I think that's fine, given we've exhausted all other options.

@heatherlarsen I'm dropping the visible focus when the padding on the sides of the content goes from `24px` to `16px` otherwise, this happens:

localhost_3001_components_alert

Let me know if that's OK for you, in which case I can always leave the focus visible, even on small screens.

heatherlarsen:
@didoo Yup, all good. Let's just keep it visible for all sizes, then.

MelSumner (approved):

PR #1181, opened 2023-02-07
`Tabs::Panel` - add style to prevent accidental display of hidden panels (HDS-1525)

KristinLBradley:
### :pushpin: Summary
If merged, this PR adds a style to avoid consumers accidentally displaying hidden Tabs Panels.

### :link: External links
- Jira ticket: [HDS-1525](https://hashicorp.atlassian.net/browse/HDS-1525)
- Slack discussion: https://hashicorp.slack.com/archives/C7KTUHNUS/p1675731491282229

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/app/styles/components/tabs.scss
+++ b/packages/components/app/styles/components/tabs.scss
@@ -112,3 +112,8 @@
     transition-property: left, width;
   }
 }
+
+// Prevent consumers from accidentally overriding default styling of HTML “hidden” attribute
+.hds-tabs__panel[hidden] {
+  display: none !important;
+}


packages/components/app/styles/components/tabs.scss
```diff
@@ -112,3 +112,8 @@
     transition-property: left, width;
   }
 }
+
+// Prevent consumers from accidentally overriding default styling of HTML “hidden” attribute
+.hds-tabs__panel[hidden] {
+  display: none !important;
+}
```
alex-ju, line 119: would we be able to solve this specific case by increasing specificity, either by chaining with parent `.hds-tabs.hds-tabs__panel[hidden]` or chaining with itself `.hds-tabs__panel.hds-tabs__panel`?

KristinLBradley: I deleted the !important. The person having the issue wasn't able to track down the exact cause but we can try this and if it doesn't fix the issue then try chaining together selectors.

KristinLBradley: The Slack thread link has details of what I discussed with them: https://hashicorp.slack.com/archives/C7KTUHNUS/p1675731491282229

She wasn't able to track down the exact code causing the issue but she provided what she used to fix the issue on her side.

didoo (approved):
Discussed with @ker-an and all is good now:
- she removed the `!important` from her CSS declaration, it was not needed (and she will add a comment to remove once the new version of HDS is adopted in their codebase)
- we can approve this PR, it fixes the problem (we tested the selector in her browser, and it worked as expected)
  - the issue in her specific case was that there was a global declaration (probably a CSS reset file) that assigned a `display: block` to the `section, article, header` etc, so the specificity was higher than the simple `hidden` attribute in the HTML (I imagine) 

/cc @KristinLBradley @alex-ju

PR #1184, opened 2023-02-09
`Hds:Table` - Fix documentation

didoo:
### :pushpin: Summary

During a recent discussion with @aklkv about the Table API and some difficulties he was having in adopting the component in his codebase, we have realized that one important factor was that the existing documentation was not correct, in particular the example of "Simple Table with model defined (non-sortable)" is taken as is it would not work.

This PR updates the documentation to make it clearer how the component should be used and declared in that specific case.

👉 👉 👉 **Preview**: https://hds-website-git-fix-table-docs-hashicorp.vercel.app/components/table?tab=code#simple-table-with-model-defined-non-sortable

_Notice: we will review the whole documentation once the work around the component API extension will be concluded._

In this PR I have:
- renamed an internal variable to `record` for clarity (too many `data` variables moving around, it was hard to understand what was what)
- added comment in code, to help other developers understand what's happening in that specific loop
- updated (and simplified in some parts) the "Simple Table with model defined (non-sortable)" example in the “How to use” section of the documentation

The existing documentation:
screenshot_2358

This example will not work, because the table needs the `@columns` argument to loop over the model (this is something we may want to reconsider, but if we decide to make this change it will be in a separate task).

***

### 👀 Reviewer's checklist:

### DIFF
--- a/packages/components/addon/components/hds/table/index.hbs
+++ b/packages/components/addon/components/hds/table/index.hbs
@@ -32,13 +32,13 @@
   </thead>
   <tbody class="hds-table__tbody">
     {{#if @columns}}
-      {{#each (sort-by this.getSortCriteria @model) as |data|}}
+      {{#each (sort-by this.getSortCriteria @model) as |record|}}
         {{yield
           (hash
             Tr=(component "hds/table/tr")
             Th=(component "hds/table/th" scope="row")
             Td=(component "hds/table/td" align=@align)
-            data=data
+            data=record
           )
           to="body"
         }}
--- a/packages/components/addon/components/hds/table/index.hbs
+++ b/packages/components/addon/components/hds/table/index.hbs
@@ -32,6 +32,11 @@
   </thead>
   <tbody class="hds-table__tbody">
     {{#if @columns}}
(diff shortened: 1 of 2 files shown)


packages/components/addon/components/hds/table/index.hbs
```diff
@@ -32,13 +32,18 @@
...
+        this means the consumer will *have to* use the `data` key to access it in their template
+      -------------------------------------------------------------------------------------------- }}
+      {{#each (sort-by this.getSortCriteria @model) as |record|}}
         {{yield
           (hash
             Tr=(component "hds/table/tr")
             Th=(component "hds/table/th" scope="row")
             Td=(component "hds/table/td" align=@align)
-            data=data
+            data=record
```
MelSumner, line 46: Does this mean the rest of the examples and showcase need to change as well?

didoo: Nope, it's an internal variable.

MelSumner (approved):
Once this is merged, I'll rebase some other open PRs that update the existing docs.

PR #1185, opened 2023-02-09
Extend `Dropdown::ListItem`

alex-ju:
### :pushpin: Summary

Extend `Hds::Dropdown::ListItem` with selectable items and the option for result counts to support the upcoming 'Filter' pattern.

- Add `Hds::Dropdown::ListItem::Checkmark` as a proxy for `Hds::Interactive` with a `@selected` argument (for context switching)
- Add `Hds::Dropdown::ListItem::Checkbox` as a proxy for `Hds::Form::Checkbox` (for filter)
- Add `Hds::Dropdown::ListItem::Radio` as a proxy for `Hds::Form::Radio` (for filter)
- Add `@count` argument to selectable list items (Checkmark, Checkbox, Radio) 

> **Note**
> Internal class name changes: `hds-dropdown-list-item--[variant]` → `hds-dropdown-list-item--variant-[variant]`
> - `hds-dropdown-list-item--copy-item` → `hds-dropdown-list-item--variant-copy-item`
> - `hds-dropdown-list-item--description` → `hds-dropdown-list-item--variant-description`
> - `hds-dropdown-list-item--generic` → `hds-dropdown-list-item--variant-generic`
> - `hds-dropdown-list-item--interactive` → `hds-dropdown-list-item--variant-interactive`
> - `hds-dropdown-list-item--separator` → `hds-dropdown-list-item--variant-separator`
> - `hds-dropdown-list-item--title` → `hds-dropdown-list-item--variant-title`
>
> These should not represent a breaking change as the consumers rely on the public API, but if test assertions are based on class names or extensions/overrides have been applied to these elements they may raise issues. For this reason, we flag this change as a **breaking change** to give consumers a heads-up.

[👉 Preview showcase](https://hds-showcase-git-alex-ju-dropdown-extension-hashicorp.vercel.app/components/dropdown)

Related work:
- [x] [Test the current API](https://github.com/hashicorp/cloud-ui/commit/d63bb0f6f93771f056d3e6c7efdb7b01fcd46a49) in context (`cloud-ui`s Vault insights seems like a good candidate)
- [x] Figure out if additional filter features (such as allowing result count in content items) should be part of this work or a follow-up PR – decided to group the selectable list item with the result count in this PR and leave search and actions for separate PRs so we can deliver the whole filter pattern work incrementally

<table>
<tr><td>

checkbox-before

</td><td>

checkbox-after

</td></tr>
<tr><td>

radio-before

</td><td>

radio-after

</td></tr>
</table>

### :link: External links

Jira ticket: Part of [HDS-1109](https://hashicorp.atlassian.net/browse/HDS-1109)
Figma file: https://www.figma.com/file/Kgc8BgB8jGuJXISndgTrD1/Dropdown?node-id=27919%3A58777&t=4MEZaQEQPfhRuVTW-4

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable

[HDS-1109]: https://hashicorp.atlassian.net/browse/HDS-1109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/olive-avocados-attack.md
+++ b/.changeset/olive-avocados-attack.md
@@ -1,5 +1,15 @@
 ---
-"@hashicorp/design-system-components": minor
+"@hashicorp/design-system-components": major
 ---
 
-Allow `Checkmark`, `Checkbox` and `Radio` as `Hds::Dropdown::ListItem`s
+Add `Checkmark`, `Checkbox` and `Radio` as `Hds::Dropdown::ListItem`s
+
+Rename `Hds::Dropdown::ListItem` internal CSS classes as follows:
+ - `hds-dropdown-list-item--copy-item` → `hds-dropdown-list-item--variant-copy-item`
+ - `hds-dropdown-list-item--description` → `hds-dropdown-list-item--variant-description`
+ - `hds-dropdown-list-item--generic` → `hds-dropdown-list-item--variant-generic`
+ - `hds-dropdown-list-item--interactive` → `hds-dropdown-list-item--variant-interactive`
+ - `hds-dropdown-list-item--separator` → `hds-dropdown-list-item--variant-separator`
+ - `hds-dropdown-list-item--title` → `hds-dropdown-list-item--variant-title`
+
+**Note:** If test assertions are relying on these class names, tests will fail. If extensions/overrides have been applied to these classes, they will suffer visual changes.
--- a/.changeset/olive-avocados-attack.md
+++ b/.changeset/olive-avocados-attack.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+Allow `Checkmark`, `Checkbox` and `Radio` as `Hds::Dropdown::ListItem`s
--- a/packages/components/addon/components/hds/dropdown/index.hbs
+++ b/packages/components/addon/components/hds/dropdown/index.hbs
@@ -16,6 +16,7 @@
       {{yield
         (hash
           close=c.close
+          Checkbox=(component "hds/dropdown/list-item/checkbox")
           Checkmark=(component "hds/dropdown/list-item/checkmark")
           CopyItem=(component "hds/dropdown/list-item/copy-item")
           Description=(component "hds/dropdown/list-item/description")
--- a/packages/components/addon/components/hds/dropdown/index.hbs
+++ b/packages/components/addon/components/hds/dropdown/index.hbs
@@ -16,6 +16,7 @@
       {{yield
         (hash
           close=c.close
+          Checkmark=(component "hds/dropdown/list-item/checkmark")
           CopyItem=(component "hds/dropdown/list-item/copy-item")
           Description=(component "hds/dropdown/list-item/description")
           Generic=(component "hds/dropdown/list-item/generic")
--- a/packages/components/addon/components/hds/dropdown/index.hbs
+++ b/packages/components/addon/components/hds/dropdown/index.hbs
@@ -12,7 +12,7 @@
     }}
   </:toggle>
   <:content as |c|>
-    <ul class={{this.listClassNames}} {{style width=@width}}>
+    <ul class={{this.listClassNames}} {{style width=@width}} role="listbox">
       {{yield
         (hash
           close=c.close
--- a/packages/components/addon/components/hds/dropdown/index.hbs
+++ b/packages/components/addon/components/hds/dropdown/index.hbs
@@ -22,6 +22,7 @@
           Description=(component "hds/dropdown/list-item/description")
           Generic=(component "hds/dropdown/list-item/generic")
           Interactive=(component "hds/dropdown/list-item/interactive")
+          Radio=(component "hds/dropdown/list-item/radio")
           Separator=(component "hds/dropdown/list-item/separator")
           Title=(component "hds/dropdown/list-item/title")
         )
--- a/packages/components/addon/components/hds/dropdown/index.hbs
+++ b/packages/components/addon/components/hds/dropdown/index.hbs
@@ -12,7 +12,7 @@
     }}
   </:toggle>
   <:content as |c|>
-    <ul class={{this.listClassNames}} {{style width=@width}} role="listbox">
+    <ul class={{this.listClassNames}} {{style width=@width}} {{did-insert this.didInsertList}}>
       {{yield
         (hash
           close=c.close
--- a/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
@@ -1,4 +1,4 @@
-<li class="hds-dropdown-list-item hds-dropdown-list-item--checkbox">
+<li class="hds-dropdown-list-item hds-dropdown-list-item--variant-checkbox">
   <Hds::Form::Checkbox::Base class="hds-dropdown-list-item__control" id={{this.id}} @value={{@value}} ...attributes />
   {{#if @icon}}
     <div class="hds-dropdown-list-item__icon">
--- a/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
@@ -8,7 +8,7 @@
   <label class="hds-dropdown-list-item__label hds-typography-body-200" for={{this.id}}>
     {{yield}}
   </label>
-  {{#if @resultCount}}
-    <span class="hds-dropdown-list-item__count hds-typography-body-100 hds-font-weight-medium">{{@resultCount}}</span>
+  {{#if @count}}
+    <span class="hds-dropdown-list-item__count hds-typography-body-100 hds-font-weight-medium">{{@count}}</span>
   {{/if}}
 </li>
--- a/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
@@ -1,6 +1,13 @@
 <li class="hds-dropdown-list-item hds-dropdown-list-item--checkbox">
   <Hds::Form::Checkbox::Base class="hds-dropdown-list-item__control" id={{this.id}} @value={{@value}} ...attributes />
-  <label class="hds-dropdown-list-item__label hds-typography-body-200" for={{this.id}}>{{yield}}</label>
+  {{#if @icon}}
+    <div class="hds-dropdown-list-item__icon">
+      <FlightIcon @name={{@icon}} @isInlineBlock={{false}} />
+    </div>
+  {{/if}}
+  <label class="hds-dropdown-list-item__label hds-typography-body-200" for={{this.id}}>
+    {{yield}}
+  </label>
   {{#if @resultCount}}
     <span class="hds-dropdown-list-item__count hds-typography-body-100 hds-font-weight-medium">{{@resultCount}}</span>
   {{/if}}
--- a/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
@@ -2,4 +2,7 @@
   <Hds::Form::Checkbox::Field @id={{@id}} @name={{@name}} @value={{@value}} ...attributes as |F|>
     <F.Label>{{yield}}</F.Label>
   </Hds::Form::Checkbox::Field>
+  {{#if @resultCount}}
+    <Hds::BadgeCount class="hds-dropdown-list-item__count" @size="small" @type="filled" @text={{@resultCount}} />
+  {{/if}}
 </li>
--- a/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
@@ -1,7 +1,6 @@
 <li class="hds-dropdown-list-item hds-dropdown-list-item--checkbox">
-  <Hds::Form::Checkbox::Field @id={{@id}} @name={{@name}} @value={{@value}} ...attributes as |F|>
-    <F.Label>{{yield}}</F.Label>
-  </Hds::Form::Checkbox::Field>
+  <Hds::Form::Checkbox::Base class="hds-dropdown-list-item__control" id={{this.id}} @value={{@value}} ...attributes />
+  <label class="hds-dropdown-list-item__label hds-typography-body-200" for={{this.id}}>{{yield}}</label>
   {{#if @resultCount}}
     <span class="hds-dropdown-list-item__count hds-typography-body-100 hds-font-weight-medium">{{@resultCount}}</span>
   {{/if}}
--- a/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
@@ -3,6 +3,6 @@
     <F.Label>{{yield}}</F.Label>
   </Hds::Form::Checkbox::Field>
   {{#if @resultCount}}
-    <Hds::BadgeCount class="hds-dropdown-list-item__count" @size="small" @type="filled" @text={{@resultCount}} />
+    <span class="hds-dropdown-list-item__count hds-typography-body-100 hds-font-weight-medium">{{@resultCount}}</span>
   {{/if}}
 </li>
--- a/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
@@ -1,0 +1,5 @@
+<li class="hds-dropdown-list-item hds-dropdown-list-item--checkbox">
+  <Hds::Form::Checkbox::Field @id={{@id}} @name={{@name}} @value={{@value}} ...attributes as |F|>
+    <F.Label>{{yield}}</F.Label>
+  </Hds::Form::Checkbox::Field>
+</li>
--- a/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
@@ -1,4 +1,4 @@
-{{! template-lint-disable require-context-role }}
+{{! template-lint-disable require-context-role require-presentational-children }}
 <li class={{this.classNames}}>
   <Hds::Interactive
     @current-when={{@current-when}}
@@ -19,6 +19,9 @@
     </div>
     {{#if @resultCount}}
       <span class="hds-dropdown-list-item__count hds-typography-body-100 hds-font-weight-medium">{{@resultCount}}</span>
+    {{/if}}
+    {{#if @selected}}
+      <FlightIcon class="hds-dropdown-list-item__checkmark" @name="check" @isInlineBlock={{false}} />
     {{/if}}
   </Hds::Interactive>
 </li>
--- a/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
@@ -1,0 +1,18 @@
+<li class={{this.classNames}}>
+  <Hds::Interactive
+    @current-when={{@current-when}}
+    @models={{hds-link-to-models @model @models}}
+    @query={{hds-link-to-query @query}}
+    @replace={{@replace}}
+    @route={{@route}}
+    @isRouteExternal={{@isRouteExternal}}
+    @href={{@href}}
+    @isHrefExternal={{@isHrefExternal}}
+    class="hds-dropdown-list-item__interactive"
+    ...attributes
+  >
+    <div class="hds-dropdown-list-item__interactive-text hds-typography-body-200 hds-font-weight-medium">
+      {{yield}}
+    </div>
+  </Hds::Interactive>
+</li>
--- a/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
@@ -1,3 +1,4 @@
+{{! template-lint-disable require-context-role }}
 <li class={{this.classNames}}>
   <Hds::Interactive
     @current-when={{@current-when}}
@@ -10,6 +11,8 @@
     @isHrefExternal={{@isHrefExternal}}
     class="hds-dropdown-list-item__interactive"
     ...attributes
+    role="option"
+    aria-selected={{if @selected "true" "false"}}
   >
     <div class="hds-dropdown-list-item__interactive-text hds-typography-body-200 hds-font-weight-medium">
       {{yield}}
--- a/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
@@ -20,8 +20,10 @@
     {{#if @resultCount}}
       <span class="hds-dropdown-list-item__count hds-typography-body-100 hds-font-weight-medium">{{@resultCount}}</span>
     {{/if}}
-    {{#if @selected}}
-      <FlightIcon class="hds-dropdown-list-item__checkmark" @name="check" @isInlineBlock={{false}} />
-    {{/if}}
+    <span class="hds-dropdown-list-item__checkmark">
+      {{#if @selected}}
+        <FlightIcon class="hds-dropdown-list-item__checkmark-icon" @name="check" @isInlineBlock={{false}} />
+      {{/if}}
+    </span>
   </Hds::Interactive>
 </li>
--- a/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
@@ -22,8 +22,8 @@
     <div class="hds-dropdown-list-item__interactive-text hds-typography-body-200 hds-font-weight-medium">
       {{yield}}
     </div>
-    {{#if @resultCount}}
-      <span class="hds-dropdown-list-item__count hds-typography-body-100 hds-font-weight-medium">{{@resultCount}}</span>
+    {{#if @count}}
+      <span class="hds-dropdown-list-item__count hds-typography-body-100 hds-font-weight-medium">{{@count}}</span>
     {{/if}}
     <span class="hds-dropdown-list-item__checkmark">
       {{#if @selected}}
--- a/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
@@ -1,4 +1,4 @@
-{{! template-lint-disable require-context-role }}
+{{! template-lint-disable require-context-role require-presentational-children }}
 <li class={{this.classNames}}>
   <Hds::Interactive
     @current-when={{@current-when}}
@@ -17,5 +17,8 @@
     <div class="hds-dropdown-list-item__interactive-text hds-typography-body-200 hds-font-weight-medium">
       {{yield}}
     </div>
+    {{#if @resultCount}}
+      <Hds::BadgeCount class="hds-dropdown-list-item__count" @size="small" @type="filled" @text={{@resultCount}} />
+    {{/if}}
   </Hds::Interactive>
 </li>
--- a/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
@@ -14,6 +14,11 @@
     role="option"
     aria-selected={{if @selected "true" "false"}}
   >
+    {{#if @icon}}
+      <div class="hds-dropdown-list-item__interactive-icon">
+        <FlightIcon @name={{@icon}} @isInlineBlock={{false}} />
+      </div>
+    {{/if}}
     <div class="hds-dropdown-list-item__interactive-text hds-typography-body-200 hds-font-weight-medium">
       {{yield}}
     </div>
--- a/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
@@ -1,4 +1,4 @@
-{{! template-lint-disable require-context-role require-presentational-children }}
+{{! template-lint-disable require-context-role }}
 <li class={{this.classNames}}>
   <Hds::Interactive
     @current-when={{@current-when}}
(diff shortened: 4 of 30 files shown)


packages/components/addon/components/hds/dropdown/index.hbs
```diff
@@ -8,14 +8,17 @@
     }}
   </:toggle>
   <:content as |c|>
-    <ul class={{this.listClassNames}} {{style width=@width}}>
+    <ul class={{this.listClassNames}} {{style width=@width}} role="listbox">
```
alex-ju, line 11: This role would need to be applied only when using the `Checkmark` contextual component. This raises the usual challenge of a parent component knowing about children.

MelSumner: Have you tried using a conditional here? 

```
role={{if hasCheckbox "listbox" null }}
```

Ember won't render an attribute at all if the value is `null`.

alex-ju: yes, using a boolean here would be an option, just have to make sure we either document it well or we try to determine it automatically within the component

packages/components/app/styles/components/dropdown.scss
```diff
@@ -408,8 +416,82 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+}
+
+.hds-dropdown-list-item--checkmark-selected {
+  background-image: url("data:image/svg+xml,%3Csvg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M14.78 4.28a.75.75 0 00-1.06-1.06l-7.97 7.97-3.47-3.47a.75.75 0 00-1.06 1.06l4 4a.75.75 0 001.06 0l8.5-8.5z' fill='%231060ff'/%3E%3C/svg%3E");
+  background-repeat: no-repeat;
+  background-position: top 10px right 10px;
+  background-size: var(--token-form-select-background-image-size) var(--token-form-select-background-image-size);
+
+  .hds-dropdown-list-item__interactive {
+    color: var(--token-color-foreground-action-active);
```
alex-ju, line 459:
```suggestion
    color: var(--token-color-foreground-action);
```

packages/components/app/styles/components/dropdown.scss
```diff
@@ -408,8 +416,82 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+  }
+}
+
+.hds-dropdown-list-item--checkmark-selected {
+  background-image: url("data:image/svg+xml,%3Csvg viewBox='0 0 16 16' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath d='M14.78 4.28a.75.75 0 00-1.06-1.06l-7.97 7.97-3.47-3.47a.75.75 0 00-1.06 1.06l4 4a.75.75 0 001.06 0l8.5-8.5z' fill='%231060ff'/%3E%3C/svg%3E");
+  background-repeat: no-repeat;
+  background-position: top 10px right 10px;
+  background-size: var(--token-form-select-background-image-size) var(--token-form-select-background-image-size);
+
+  .hds-dropdown-list-item__interactive {
```
didoo, line 458: maybe it's me, but may be that this rule is not applied? I don't see it in the styles when I inspect it in the browser (but may be doing something wrong).

packages/components/app/styles/components/dropdown.scss
```diff
@@ -408,8 +416,82 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+
+// HDS::DROPDOWN::LIST-ITEM::CHECKBOX & HDS::DROPDOWN::LIST-ITEM::RADIO
+
+.hds-dropdown-list-item--checkbox,
+.hds-dropdown-list-item--radio {
+  display: flex;
+  padding: 8px 16px;
+
+  .hds-form-field--layout-flag {
+    flex: 1;
```
didoo, line 471: Probably I would not override this, but use a wrapping container (or rely on the `auto` margin of the elements on the right, if possible). Not a blocker though.

packages/components/app/styles/components/dropdown.scss
```diff
@@ -408,8 +416,82 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+  }
+
+  .hds-form-label {
+    font-weight: normal;
```
didoo, line 475: In the designs in Figma, I see the color is not `form/label` but `foreground/primary`. Can you check with @heatherlarsen which one we should use?

didoo (commented):
@alex-ju did a first high-level pass at the code and added a few comments.

(didn't look at the tests, for now)

packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
```diff
@@ -0,0 +1,8 @@
+<li class="hds-dropdown-list-item hds-dropdown-list-item--checkbox">
+  <Hds::Form::Checkbox::Field @id={{@id}} @name={{@name}} @value={{@value}} ...attributes as |F|>
+    <F.Label>{{yield}}</F.Label>
+  </Hds::Form::Checkbox::Field>
+  {{#if @resultCount}}
+    <Hds::BadgeCount class="hds-dropdown-list-item__count" @size="small" @type="filled" @text={{@resultCount}} />
```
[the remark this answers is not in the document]
heatherlarsen: This is my fault. I used the BadgeCount component as a starting point in the explorations because we were looking at using them directly and for this variation of the design, I just detached and overrode the values. I'll make sure the built component is just plain text with a container. It makes sense to me that the implemented version does not use the BadgeCount component.

packages/components/app/styles/components/dropdown.scss
heatherlarsen: `foreground/primary` is correct. This is to keep the color of the label when it is within the menu, consistent with the other dropdown list item variants, which use `foreground/primary`.

didoo: In this case, another reason (not strong, just a possibility) to considering using the `Base` form element, and not the `Field`, like it's been done in the `Pagination` component (and in that case, it's likely you will not need to override the `hds-form-field--layout-flag` flex).

alex-ju: good spot @didoo and thank you for confirming @heatherlarsen; will try with the `Base` element the to minimise the overrides

packages/components/addon/components/hds/dropdown/list-item/checkbox.hbs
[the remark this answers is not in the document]
alex-ju: thank you both! will update to use a plain element then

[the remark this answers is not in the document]
alex-ju: [We're now using a plain element for the result count](https://github.com/hashicorp/design-system/pull/1185/commits/d3d7cc80906c0163b0d39d66f22f6b0932068704) (replaced `BadgeCount` with `<span>` and applied similar style)

packages/components/app/styles/components/dropdown.scss
```diff
@@ -408,8 +416,82 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+      }
+    }
+  }
+
+  .hds-dropdown-list-item__interactive-text {
+    flex: 1;
+  }
+}
+
+.hds-dropdown-list-item--checkmark-selected {
```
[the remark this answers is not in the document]
alex-ju: I went with the background image to preserve the result count position/alignment for both selected and unselected items; I've updated the code to [use FlightIcon for checkmark](https://github.com/hashicorp/design-system/pull/1185/commits/21f97281478d34c1a04b925e7754c2194385703b) for selected items, and rely on padding for unselected ones.

packages/components/app/styles/components/dropdown.scss
alex-ju: yes, it wasn't applied, because of lack of specificity – adjusted now

packages/components/app/styles/components/dropdown.scss
alex-ju: updated to [use the form base element instead of form field](https://github.com/hashicorp/design-system/pull/1185/commits/4aefe474463b81fe1b281b8b9a85a3f7c9029281)

packages/components/app/styles/components/dropdown.scss
```diff
@@ -208,18 +209,23 @@ $hds-dropdown-toggle-border-radius: 5px;
...
 .hds-dropdown-list-item--generic {
   padding-right: 16px;
   padding-left: 16px;
 }
 
-
-.hds-dropdown-list-item--interactive {
+// HDS::DROPDOWN::LIST-ITEM::INTERACTIVE & HDS::DROPDOWN::LIST-ITEM::CHECKMARK
+.hds-dropdown-list-item--interactive,
+.hds-dropdown-list-item--checkmark {
```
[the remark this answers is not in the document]
alex-ju: we still need a `--checkmark` modifier operating some overrides, so instead of composing from two modifiers (`--interactive` and `--checkmark` I thought I may use the same properties block. Some of these properties could potentially be abstracted in a mixin to simplify the code a bit.

packages/components/addon/components/hds/dropdown/list-item/checkmark.hbs
```diff
@@ -0,0 +1,24 @@
+{{! template-lint-disable require-context-role require-presentational-children }}
+<li class={{this.classNames}}>
+  <Hds::Interactive
```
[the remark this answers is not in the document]
alex-ju: `cloud-ui` doesn't do URL updating on click, but on a separate 'apply filter' button, but HashiDocs, for example, does it on click, and I imagine other products might take this approach (as I understand we're aiming to support both, by design).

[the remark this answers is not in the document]
didoo: 👍

packages/components/app/styles/components/dropdown.scss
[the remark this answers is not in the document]
didoo: 👍

packages/components/app/styles/components/dropdown.scss
```diff
@@ -408,8 +416,86 @@ $hds-dropdown-toggle-border-radius: 5px;
...
   color: var(--token-color-foreground-strong);
 }
 
+// HDS::DROPDOWN::LIST-ITEM::CHECKMARK
+
+.hds-dropdown-list-item--checkmark {
+  a.hds-dropdown-list-item__interactive,
+  button.hds-dropdown-list-item__interactive {
+    // leave room for the checkmark to align the result count on non-selected items
+    padding-right: 25px; // 9px default + 16px for the 'check' FlightIcon
```
didoo, line 432: Maybe you could use a parent container, with its own intrinsic size, and then conditionally render the icon inside it, so that you don't need to use extra padding (and calculations) to achieve the desired layout

packages/components/app/styles/components/dropdown.scss
```diff
@@ -408,8 +416,86 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+  a.hds-dropdown-list-item__interactive,
+  button.hds-dropdown-list-item__interactive {
+    padding-right: 9px; // reset to default
+    color: var(--token-color-foreground-action);
+  }
+}
+
+.hds-dropdown-list-item__checkmark {
+  align-self: flex-start;
+  margin-top: 0.125rem;
```
didoo, line 462: if you use a container you don't need this padding anymore (you can set the height directly to the desired value, which is `20px`, the line-height of the main text)

packages/components/app/styles/components/dropdown.scss
```diff
@@ -408,8 +416,86 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+    color: var(--token-color-foreground-primary);
+  }
+}
+
+// RESULT COUNT
+.hds-dropdown-list-item__count {
+  align-self: flex-start;
+  padding: 0.125rem 0.5rem;
+  color: var(--token-color-foreground-faint);
+  line-height: 1.2308; // 16px
```
didoo, line 490: maybe instead of doing this double trick of setting a padding + a line height, you can be more declarative and set the line-height to `20px` (the line-height of the main text) and no vertical padding (adding a comment explaining why this 20px value)

packages/components/tests/integration/components/hds/dropdown/list-item/checkmark-test.js
```diff
@@ -0,0 +1,92 @@
...
+
+    test('it should render the "list-item/checkmark" as a <li> element with a CSS class that matches the component name', async function (assert) {
+      await render(
+        hbs`<Hds::Dropdown::ListItem::Checkmark>Checkmark item</Hds::Dropdown::ListItem::Checkmark>`
+      );
+      assert.dom('.hds-dropdown-list-item').hasTagName('li');
+      assert
+        .dom('.hds-dropdown-list-item')
+        .hasClass('hds-dropdown-list-item--checkmark')
+        .hasClass('hds-dropdown-list-item--color-action');
```
didoo, line 26: I've mixed feeling about this. Isn't this class mainly presentational?
```suggestion
```

packages/components/tests/integration/components/hds/dropdown/list-item/checkbox-test.js
```diff
@@ -0,0 +1,66 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module(
+  'Integration | Component | hds/dropdown/list-item/checkbox',
```
didoo, line 7: I see below we are testing the "selected" state of the "checkmark" which made me think: should we test also the "checked" state of the `list-item/checkbox`? (and radio too?)

didoo: also, should we test the `getId` functionality too? (the label and control having the correct ID?)

didoo (commented):
left a few comments/suggestions

packages/components/app/styles/components/dropdown.scss
alex-ju: Followed the suggestion and refactored to use a fixed-width container, that conditionally renders the checkmark icon in [35201dd](https://github.com/hashicorp/design-system/pull/1185/commits/35201dd1eb17d523f19e16967eeac30e8dfe3875)

packages/components/tests/integration/components/hds/dropdown/list-item/checkbox-test.js
alex-ju: amended the tests in [121b639](https://github.com/hashicorp/design-system/pull/1185/commits/121b6396c33a83bca71f09c1805ad12f7de8fb43) to capture these features

packages/components/app/styles/components/dropdown.scss
alex-ju: that makes sense, especially now when the checkmark icon also uses this value – updated in [3991ef2](https://github.com/hashicorp/design-system/pull/1185/commits/3991ef22579e58c852e1f7cbadc57b66b81487a9)

alex-ju:
@didoo, @heatherlarsen, @MelSumner I rebase this to main to align with the new Showcase – this is good for another review

packages/components/app/styles/components/dropdown.scss
```diff
@@ -413,8 +421,88 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+      color: var(--token-color-foreground-action-active);
+
+      &::after {
+        background-color: inherit;
+      }
+    }
+  }
+
+  .hds-dropdown-list-item__interactive-text {
+    flex: 1;
```
didoo, line 450: have you considered moving this declaration directly to the `.hds-dropdown-list-item__interactive-text` instead of here? I _don't think_ there would be any unintended side effects, and would avoid having too many specific/custom declarations around (even the overrides of the `focus/active` states above make me nervous: was it _really_ necessary?).

packages/components/app/styles/components/dropdown.scss
```diff
@@ -413,8 +421,88 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+  }
+}
+
+.hds-dropdown-list-item--checkmark-selected {
+  a.hds-dropdown-list-item__interactive,
+  button.hds-dropdown-list-item__interactive {
+    padding-right: 9px; // reset to default
```
didoo, line 457: I _think_ you don't need this anymore, can you double check?
```suggestion
```

packages/components/app/styles/components/dropdown.scss
```diff
@@ -413,8 +421,88 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+}
+
+.hds-dropdown-list-item__checkmark {
+  align-self: flex-start;
+  width: 20px;
+  height: 20px; // replicating the resulted height of the list item
+}
+
+.hds-dropdown-list-item__checkmark-icon {
+  height: inherit;
```
didoo, line 469: too much "magic" for me here :) - I would prefer to have the parent as flexbox and use that to align the icon, it would be more explicit

packages/components/app/styles/components/dropdown.scss
```diff
@@ -413,8 +421,88 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+
+  .hds-dropdown-list-item__control {
+    flex-shrink: 0;
+    margin-top: 2px;
+    margin-right: 8px;
+  }
+
+  .hds-dropdown-list-item__label {
+    flex-grow: 1;
+    width: fit-content;
```
didoo, line 487: is this needed? I tried to turn it on/off in the devtools panel but I can't see a difference in the showcase

packages/components/app/styles/components/dropdown.scss
```diff
@@ -413,8 +421,88 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+  padding: 0 0.5rem;
+  color: var(--token-color-foreground-faint);
+  line-height: 20px; // replicating the resulted height of the list item
+}
+
+// BADGE INSIDE CHECKMARK, CHECKBOX OR RADIO
+.hds-dropdown-list-item--checkmark,
+.hds-dropdown-list-item--checkbox,
+.hds-dropdown-list-item--radio {
+  // align any `Hds::Badge` within a selectable list item to match baseline
```
didoo, line 504: 👏👏👏

packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -306,4 +306,298 @@
     </SF.Item>
   </Shw::Flex>
 
+  <Shw::Text::H3>Checkmark</Shw::Text::H3>
+  <Shw::Flex as |SF|>
+    <SF.Item>
```
didoo, line 311: I would add `@label` declarations to these (and other) items, is really hard to understand what they "showcase"

packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -306,4 +306,298 @@
...
+            {{state}}
+          </Hds::Dropdown::ListItem::Checkmark>
+        {{/each}}
+      </ul>
+    </SF.Item>
+    <SF.Item>
+      <ul class="hds-dropdown-list" role="listbox">
+        {{#each @model.ITEM_STATES as |state|}}
+          <Hds::Dropdown::ListItem::Checkmark mock-state-value={{state}}>
+            <Shw::Placeholder @text="custom content" @width="150" @height="20" @background="#eee" />
```
didoo, line 351: super-nitpick: if you use `144` instead of `150`, the items will be better aligned (and will make easier for the eye to compare them)

packages/components/app/styles/components/dropdown.scss
alex-ju: Indeed this property can be applied to the interactive text with no side effects.
Disabling the background color for focus/active states was needed to allow the use of badges inside the checkmark list item while [preserving a compliant contrast](https://www.figma.com/file/Kgc8BgB8jGuJXISndgTrD1/Dropdown?node-id=28361%3A64942&t=3jf3TMGNrOAbdpgZ-4).

packages/components/app/styles/components/dropdown.scss
alex-ju: Indeed, this was leftover – removed it

packages/components/app/styles/components/dropdown.scss
alex-ju: Update to make use of flex model

packages/components/app/styles/components/dropdown.scss
alex-ju: not needed anymore in the new flex layout – removed it

packages/components/tests/dummy/app/templates/components/dropdown.hbs
alex-ju: Good suggestion – added

packages/components/tests/dummy/app/templates/components/dropdown.hbs
alex-ju: Thanks, yes, that was the initial intention, but haven't updated the values after the layout refactor.

heatherlarsen:
@alex-ju I just looked through the showcase...this is great ✨ 

[visual bug] I tested the updates in the showcase and noticed that there's still a light blue background that appears when clicking on the list items. (example: [screencast-calendar.google.com-2023.02.27-14_00_37.webm](https://user-images.githubusercontent.com/8553306/221695941-cc6e6001-82df-4459-b1e6-a5f05be2b790.webm)) Can that be removed please?

[question] Can the custom content versions render a count as well?

alex-ju:
> @alex-ju I just looked through the showcase...this is great ✨
> 
> [visual bug] I tested the updates in the showcase and noticed that there's still a light blue background that appears when clicking on the list items. (example: [screencast-calendar.google.com-2023.02.27-14_00_37.webm](https://user-images.githubusercontent.com/8553306/221695941-cc6e6001-82df-4459-b1e6-a5f05be2b790.webm)) Can that be removed please?

Good spot! There was a declaration left that caused that behavior – removed it now.

> [question] Can the custom content versions render a count as well?

Yes, they can. I updated the custom content examples to show the count

heatherlarsen:
@alex-ju Is there an option to add an icon to the checkmark, checkbox, and radio list items? Apologies that this wasn't in the Figma file; I must've forgotten to pull those in from the graveyard, but I've added that back into the file here:  https://www.figma.com/file/Kgc8BgB8jGuJXISndgTrD1/Dropdown?node-id=29345%3A66370&t=mlu3WJnbpLdrRqAE-4

I noticed the checkmark icon isn't optically aligned with the text. In Figma, I had to add a top padding of 2px above each icon. It's definitely hacky, but it fixes most of the imbalance issues. Let me know if you want to chat more about this, though.

alex-ju:
> @alex-ju Is there an option to add an icon to the checkmark, checkbox, and radio list items? Apologies that this wasn't in the Figma file; I must've forgotten to pull those in from the graveyard, but I've added that back into the file here: https://www.figma.com/file/Kgc8BgB8jGuJXISndgTrD1/Dropdown?node-id=29345%3A66370&t=mlu3WJnbpLdrRqAE-4

Noted – I'll make sure they can render icons as well

> I noticed the checkmark icon isn't optically aligned with the text. In Figma, I had to add a top padding of 2px above each icon. It's definitely hacky, but it fixes most of the imbalance issues. Let me know if you want to chat more about this, though.

Hmm.. I remember needing to add such padding, but then I updated the approach to [use an extra container](https://github.com/hashicorp/design-system/pull/1185#discussion_r1111809166) ensuring alignment with baseline and cap height/ascender. I'm wondering if I'm missing something here 🤔  (see screen capture below)

Screenshot 2023-03-02 at 14 22 28

heatherlarsen (changes requested):
@alex-ju Ahh, you are correct; thank you! Maybe I was looking at an old link 🤔 

I've been able to go through the showcase more thoroughly for design alignment and found two more minor things: 

- there doesn't appear to be a gap between the text and the checkmark icon, so for longer strings, the text ends up bumping up right next to the icon. Is it possible to have an 8px gap between the text and the icon?
- because the checkmark icon is placed in a 20x20 container, there's an extra 4px to the right of the icon, which results in an optically larger padding on the right. The padding on the right should be a total of 10px, it currently looks like it's closer to 14px. Can the size of the container be reduced to 16px wide?

Let me know if we need to discuss these further. Thanks!

alex-ju:
> @alex-ju Ahh, you are correct; thank you! Maybe I was looking at an old link 🤔
> 
> I've been able to go through the showcase more thoroughly for design alignment and found two more minor things:
> 
> * there doesn't appear to be a gap between the text and the checkmark icon, so for longer strings, the text ends up bumping up right next to the icon. Is it possible to have an 8px gap between the text and the icon?
> * because the checkmark icon is placed in a 20x20 container, there's an extra 4px to the right of the icon, which results in an optically larger padding on the right. The padding on the right should be a total of 10px, it currently looks like it's closer to 14px. Can the size of the container be reduced to 16px wide?
> 
> Let me know if we need to discuss these further. Thanks!

Both are good spots @heatherlarsen! I made sure we have a consistent 8px spacing between the elements of a list item (icon, text, count, checkmark) and have updated the checkmark width to 16px to avoid any extra space there.

alex-ju:
> more comments :)

@didoo I have hopefully addressed the feedback from previous rounds of review at the time.

> question: do you remember we discussed the changes to the "modifiers" classes for the "list-item" variants (from `hds-dropdown-list-item--checkmark` to `hds-dropdown-list-item--variant-checkmark` for example, to avoid confusion with `hds-dropdown-list-item__checkmark`)? since it's a breaking change, I am OK to do it in a separate PR (so this is not a blocker) but let's not forget to do it. if instead you prefer to do it directly in this one, no problems for me as well (in which case you have to update the changeset file)

Indeed, I remember the conversation and have updated the modifier classes here as we're introducing some more modifiers; have also updated the changelog to reflect this change and marked it as a major (although technically it is not, we know it will cause breaking changes).

However, since the last review, a few more changes were introduced. I kept the git history untouched to help you pick up from where you left off, although may not be of much value if reviewed as a whole.

@heatherlarsen I have hopefully addressed your feedback too, including the updated vertical alignment. If you spot anything off please leave comments as necessary.

heatherlarsen (approved):
This looks great, @alex-ju! I don't see anything else that needs adjusting ✨ :shipit:

.changeset/olive-avocados-attack.md
```diff
@@ -0,0 +1,6 @@
+---
+"@hashicorp/design-system-components": major
+---
+
+Allow `Checkmark`, `Checkbox` and `Radio` as `Hds::Dropdown::ListItem`s
+Rename `Hds::Dropdown::ListItem` internal CSS classes
```
didoo, line 6: do you think it would make sense to add here a note/comment that explains that _this_ is the breaking change that causes the version bump to be major?

didoo (commented):
sorry, premature submit 🤦‍♂️

packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -306,4 +306,358 @@
...
+      <Hds::Dropdown @listPosition="left" as |dd|>
+        <dd.ToggleButton @text="Checkmark" @color="secondary" />
+        <dd.Checkmark @count="11">virtualbox</dd.Checkmark>
+        <dd.Checkmark @count="1" @selected={{true}}>vmware</dd.Checkmark>
+        <dd.Checkmark @count="10">docker</dd.Checkmark>
+        <dd.Checkmark @count="0">hyperv</dd.Checkmark>
+      </Hds::Dropdown>
+    </SF.Item>
+  </Shw::Flex>
+
```
didoo, line 424: I would probably add level-1 and level-2 separators (`<Shw::Divider />`, `<Shw::Divider @level={{2}} />`) between the different types examples, to increase readability/scan.
See for example how it's done in the [Pagination](https://hds-showcase.vercel.app/components/pagination) or [Form Base elements](https://hds-showcase.vercel.app/components/form/base-elements) showcases.

packages/components/tests/integration/components/hds/dropdown/index-test.js
```diff
@@ -119,4 +119,27 @@ module('Integration | Component | hds/dropdown/index', function (hooks) {
...
+      </Hds::Dropdown>
+    `);
+    await click('button#test-toggle-button');
+    assert.dom('#test-dropdown ul').doesNotHaveAttribute('role');
+  });
+  test('it should render a list of items with a `listbox` role if selectable items are passed in', async function (assert) {
+    await render(hbs`
+      <Hds::Dropdown id="test-dropdown" as |dd|>
+      <dd.ToggleButton @text="toggle button" id="test-toggle-button" />
+      <dd.Checkmark>Checkmark</dd.Checkmark>
```
didoo, line 139:
```suggestion
        <dd.ToggleButton @text="toggle button" id="test-toggle-button" />
        <dd.Checkmark>Checkmark</dd.Checkmark>
```

didoo (approved):
All looking (very) good to me. 👍

For the merge and release, let's sync: we probably want to merge (and release) the `SideNav` elements, which are not breaking, so we can start to adopt them, and then release these (breaking) ones, which need to be followed by one or more PRs in Cloud UI and potentially other consumers (see https://cs.github.com/?scope=org%3Ahashicorp&scopeName=hashicorp&q=hds-dropdown-list-item)

packages/components/tests/dummy/app/templates/components/dropdown.hbs
alex-ju: Added the level 2 dividers – looks much better with them!

packages/components/tests/integration/components/hds/dropdown/index-test.js
```diff
@@ -119,4 +119,27 @@ module('Integration | Component | hds/dropdown/index', function (hooks) {
     assert.dom('#test-dropdown').hasAttribute('data-test1');
     assert.dom('#test-dropdown').hasAttribute('data-test2', 'test');
   });
+
+  // ACCESSIBILITY
+
+  test('it should render a list of items without a role if no selectable items are passed in', async function (assert) {
```
MelSumner, line 125: question: I don't really understand this test, can you explain? Lists show up in the rotor and as such need an accessible name even if they don't have interactive elements.

alex-ju: > it should render a list of items without a role if no selectable items are passed in

now that I read it, I can see how the double negation doesn't help

what I'm testing here is that _if_ we don't have selectable items (checkbox, radio), we don't add `role="listbox"` to the list, relying on the implicit role for an unordered list, which is "list" (in this scenario, a screen reader would read: _my item, button, group, 5 items_)

in the following test, we check that we add `role="listbox"` if there are selectable items (in this scenario, a screen reader would read: _my item, not selected, 1 of 5_)

I'm trying to make sure the `role` is only explicitly applied to the list if it has selectable items so that users are told about the state of each item when navigating through them

alex-ju:
@didoo while testing this change in `cloud-ui` I realized that if we change the variant class names, it will also require a new `structure` release for the [override selectors](https://github.com/hashicorp/structure/blob/main/packages/pds-ember/app/styles/pds/_overrides.scss#L8).

.changeset/olive-avocados-attack.md
alex-ju: [Updated the changelog entry](https://github.com/hashicorp/design-system/pull/1185/commits/fcf80f8b846bf615aa648335793e9a7d39d78c60) to list the actual changes and the potential effects.

PR #1187, opened 2023-02-09
Exploration: add list support to doc/component-api/property component

MelSumner:
It would be useful to have a list in the component description in the api docs. What do we think?

Here it is in use to help further clarify the `@column` use for the `Hds::Table` component: 

CleanShot 2023-02-09 at 12 40 45

Some notes: 
- it'll need some tweaking for the design (the bullets should fall in line, the vertical spacing is off, maybe we want the code and/or badges to be a little different?) 
- the CSS classes aren't correct as a result, but was trying to replicate what already existed for exploration purposes
- current limitation: values can't contain any code themselves (it all breaks 😭 )


MelSumner:
Replaced with #1197

PR #1198, opened 2023-02-10
Website: Updates `Hds::Table` component documentation

MelSumner:
### :pushpin: Summary

If merged, this PR updates the existing table component API documentation for clarity and accuracy. 

Resolves HDS-1537 and HDS-1538. Replaces #1190. Contains other branches that we decided to roll up into one PR for ease of review.

#### How to use" changes

- Added a custom sort example
- Added an example of what a custom sort function could look like
- Moved "More Examples" to its own section with examples under that to be more consistent with other examples (this also helped examples show up better in the sidecar)

#### Component API changes

- Updated `columns`: 
    - initial definition
    - nested options:
        - label
        - key
        - isSortable
        - align
        - width
        - sortingFunction
- Added `identityKey`
- Added `sortedMessageText` 
- Added child component APIs
    - `Tr`
        - note about interactions
        - `...attributes`
    - `Th`
        - note about interactions
        - align
        - scope
        -  width
        - `...attributes`
    - `ThSort`
        - note about interactions
        - description
        - isSorted
        - sortBy
        - sortOrder
        - align
        - width
        - onClick
        - `...attributes` 
    - `Td`
        - note about interactions
        - align
        - `...attributes`

Preview link: https://hds-website-git-melsumner-update-existing-tabl-675d53-hashicorp.vercel.app/components/table?tab=code 

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable


website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,107 @@
...
   <C.Property @name="<:body>" @type="named block">
     This is a named block where the content for the table body (`<tbody>`) is rendered.
   </C.Property>
   <C.Property @name="model" @type="array">
-    If defined, sets the data source that gets yielded by the `:body` named block.
+    Indicates the data model to be used by the table.
   </C.Property>
   <C.Property @name="columns" @type="array">
-    Use an `array` hash to define your table columns. While `key` and `label` are required, other options include `isSortable`, `align` (for text-alignment), and `width`.
+  Use a `hash` within the array to define each column. `Key` and `label` values are both required. Add `isSortable` if you would like the column to be sortable.
```
didoo, line 18: Technically is not correct that the `key` (lowercase) is required: it is only if one needs a sortable table, but if the table is static using just  the `label` will work.

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,107 @@
...
   </C.Property>
-  <C.Property @name="onSort" @type="function">
-    Callback function invoked (if provided) when one of the sortable table headers is clicked, and a sorting is triggered. The function receives the values of `sortBy` and `sortOrder` as arguments.
-  </C.Property>
   <C.Property @name="isStriped" @type="boolean" @values={{array "false" "true" }} @default="false">
     Define on the table invocation. If set to `true`, row striping ("zebra striping") will be applied to the table.
   </C.Property>
-  <C.Property @name="isFixed" @type="boolean" @values={{array "false" "true" }} @default="false">
+  <C.Property @name="isFixedLayout" @type="boolean" @values={{array "false" "true" }} @default="false">
     If set to `true`, the `table-display`(CSS) property will be set to `fixed`, which will automatically distribute columns equally based on the total width of the table.
```
didoo, line 30: I just noticed this, while reading, and I don't think is correct: the columns are not always equally distributed, it depends if some of them have a width applied (in which case the width is attempted to be preserved)

https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout#values
`fixed` - Table and column widths are set by the widths of table and col elements or by the width of the first row of cells.

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,107 @@
...
   </C.Property>
   <C.Property @name="density" @type="enum" @values={{array "short" "medium" "tall" }} @default="medium">
     If set, determines the density, or height, of the table’s rows.
   </C.Property>
+  <C.Property @name="align" @type="enum" @values={{array "left" "center" "right" }} @default="left">
+    If set, determines the text alignment for all cell (`td`) content in a table.
```
didoo, line 36: should we also specify that doesn't apply to table headers as the one below?

(in the table body the `align=@align` is not declared on the `Th`, don't know it it's intentional or if it's a slip and should be there)

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,107 @@
...
   </C.Property>
   <C.Property @name="caption" @type="string">
-    Adds a (non-visible) caption for users with assistive technology. If set on a sortable table, the provided table caption is paired with the automatically generated sorted message text.
+    Adds a (non-visible) caption for users with assistive technology. If set on a sortable table, the provided table caption is paired with the automatically generated sorted message text. Note: it is best practice to provide a table caption for users with assistive technology.
   </C.Property>
   <C.Property @name="...attributes">
     Supported for the `Hds::Table` component.
   </C.Property>
+  <C.Property @name="onSort" @type="function">
+    Automatically applied to sortable table headers, `onSort` is the callback function that is invoked when one of the sortable table headers is clicked (or has a keyboard interaction performed). The function receives the values of `sortBy` and `sortOrder` as arguments.
```
didoo, line 48: As reader, I would not be sure what "Automatically applied to sortable table headers" exactly means.

```suggestion
    Callback function that is invoked when one of the sortable table headers is clicked (or has a keyboard interaction performed). The function receives the values of `sortBy` and `sortOrder` as arguments.
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,107 @@
...
+  </C.Property>
+</Doc::ComponentApi>
+
+### Table::Tr
+
+The `Hds::Table::Tr` component is a template-only component.
+
```
didoo, line 55: This is internal behaviour, is not relevant for consumers.

```suggestion
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,107 @@
...
+<Doc::ComponentApi as |C|>
+    <C.Property @name="...attributes">
+    This component supports the use of `...attributes`.
+  </C.Property>
+</Doc::ComponentApi>
+
+### Table::Th
+
+This component is not eligible to receive interactions (e.g., it cannot have an `onClick` event handler attached directly to it). Instead, an interactive element should be placed _inside_ of a `th` or `td` element.
```
didoo, line 68: Is this paragraph correct? I suspect is a copy of the previous block.

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,107 @@
...
+
+If a `th` is passed as the first "cell" of a table body row, it has `scope="row"` automatically applied to it for accessibility purposes.
+
+<Doc::ComponentApi as |C|>
```
didoo, line 72: These arguments are missing:
- `@width`
- `@scope`

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,107 @@
...
+    If set, determines the text alignment.
+  </C.Property>
+  <C.Property @name="...attributes">
+    This component supports the use of `...attributes`.
+  </C.Property>
+</Doc::ComponentApi>
+
+### Table::ThSort
+
+This component is not eligible to receive interactions (e.g., it cannot have an `onClick` event handler attached directly to it). Instead, an interactive element should be placed _inside_ of a `th` or `td` element.
```
didoo, line 83: Same as above, not sure this paragraph is correct.

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,107 @@
...
+
+This is the component that supports the column sorting.
```
didoo, line 85: Maybe we should explain that should be used in place of the `Th` one when sorting is needed?

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,107 @@
...
+
+<Doc::ComponentApi as |C|>
```
didoo, line 87: These arguments are missing:
- `@isSorted`
- `@sortOrder`
- `@align`
- `@width`

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,107 @@
...
+    The action handler; sets the sort to the column key.
+  </C.Property>
+  <C.Property @name="...attributes">
+    This component supports the use of `...attributes`.
+  </C.Property>
+</Doc::ComponentApi>
+
+### Table::Td
+
+This component is not eligible to receive interactions (e.g., it cannot have an `onClick` event handler attached directly to it). Instead, an interactive element should be placed _inside_ of a `th` or `td` element.
```
didoo, line 98: Same as above.

website/docs/components/table/partials/code/component-api.md
MelSumner: It's correct and it is also a copy/paste. 👍

didoo: The section is referring to the `Th` element, but the text says:

> an interactive element should be placed _inside_ of a `th` or `td` element.

Isn't it confusing? (it is for me, when I read it)

MelSumner: Yes it's correct.

website/docs/components/table/partials/code/component-api.md
MelSumner: It was the easiest/most terse way to explain it. But I can add info about the browser's algorithm if you really think it would be useful.

website/docs/components/table/partials/code/component-api.md
MelSumner: They should not be using scope at all, that's why it's not there. I'll add width.

website/docs/components/table/partials/code/component-api.md
MelSumner: sure.

website/docs/components/table/partials/code/component-api.md
MelSumner: ok, added clarification

website/docs/components/table/partials/code/component-api.md
MelSumner: After thinking about this, I think it's different depending on how you use it, I guess. When you are invoking it in a table context, it's setting it for the table. So I don't think we need to say more than that.

website/docs/components/table/partials/code/component-api.md
didoo: Since it's referring to the `Table::Th` the text 

> an interactive element should be placed _inside_ of a `th` or `td` element.

sounds strange to me. I would do something like this: 

> an interactive element should be placed _inside_ the `Th` element

(and the same for the others, when it's a `Td` I would say that the interactive element should be inside the `Th`, etc)

website/docs/components/table/partials/code/component-api.md
didoo: How is this relevant for the consumers? can you explain please?

website/docs/components/table/partials/code/component-api.md
didoo: > They should not be using scope at all

When they build a table by themselves, just using the building blocks, are they not expected to use the `@scope` for the `Th`? If they are, then we should expose the argument and explain how/when to use it. This is more in line with our principle _"Guidance over control"_.

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,122 @@
...
+</Doc::ComponentApi>
+
+### Table::ThSort
+
+This component is not eligible to receive interactions (e.g., it cannot have an `onClick` event handler attached directly to it). Instead, an interactive element should be placed _inside_ of a `th` or `td` element.
+
+This is the component that supports the column sorting, and should be used instead of `Hds::Table::Th` if creating a custom implementation.
+
+<Doc::ComponentApi as |C|>
+  <C.Property @name="onClick" @type="function">
```
didoo, line 91: We tend to have `onSomething` events after the other normal arguments (with `...attributes` last)

website/docs/components/table/partials/code/component-api.md
MelSumner: Sorry, what I'm saying is, we automatically apply it in the appropriate way. If the `Th` element is inside of the `tbody`, we add `scope="row"`, since that's the only possible combination where this application would be correct. That way, the accessibility is built-in and the consumer doesn't have to worry about that detail.

MelSumner: Also, I can add it to the documentation, it seemed like we haven't been adding things that are either private or things the consumer doesn't have to worry about. So I could go either way, WDYT?

didoo: It's a remote case, but if consumers do something like this:
```
<Hds::Table>
  ...
  <:body>
    <Hds::Table::Tr>
      <Hds::Table::Th>...</Hds::Table::Th>
      <Hds::Table::Td>...</Hds::Table::Td>
      <Hds::Table::Td>...</Hds::Table::Td>
    </Hds::Table::Tr>
  </:body>
</Hds::Table>
```
It will not be applied. So in that sense, we should be thorough and document it (with a note that if they use contextual subcomponents, it will be applied for them automatically)

But if you think is too much, no big deal, we can leave this specific prop out.

MelSumner: oh, good callout. I think I should add it then, since a lot of these sort of fall in this same category (e.g., if you use a custom invocation then you'll have to define a lot of things yourself) 👍

website/docs/components/table/partials/code/component-api.md
MelSumner: re: `align` it's kind of strange because the showcase is correct: 

CleanShot 2023-02-23 at 10 03 25

I guess I need to figure out how it's getting that if it wasn't added to the component properly?

website/docs/components/table/partials/code/component-api.md
MelSumner: Changed it! 👍

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,38 +1,50 @@
 ## Component API
 
-The Table component itself is where most of the options will be applied. However, the child components can also be used if a custom implementation is needed.
+The Table component itself is where most of the options will be applied. However, the API for the child components are also documented here, in case a custom implementation is desired.
```
KristinLBradley, line 3: Change to either:
"API for the child components are" => "API for the child components is"
-OR-
"API for the child components are" => "APIs for the child components are"

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,38 +1,50 @@
...
-- The `Hds::Table::Th` component is a template-only component. It supports use of [`...attributes`](https://guides.emberjs.com/release/in-depth-topics/patterns-for-components/#toc_attribute-ordering) and can contain interactive elements but is not eligible to receive interactions itself. It‘s unlikely you‘ll need to add interactive elements, as sorting is already provided.
-- The `Hds::Table::Td` component is a template-only component. It supports use of [`...attributes`](https://guides.emberjs.com/release/in-depth-topics/patterns-for-components/#toc_attribute-ordering) and can contain interactive elements (e.g., `<td><a href="user-info.html">User info</a></td>`) but is not eligible to receive interactions itself (e.g., it cannot have an `onClick` event handler attached directly to it).
+### Table
 
 <Doc::ComponentApi as |C|>
   <C.Property @name="<:head>" @type="named block">
-    This is a named block where the content for the table head (`<thead>`) is rendered.
+    This is a named block where the content for the table head (`<thead>`) is rendered. It is unlikely that most consumers will need to use this named block directly for most use cases.
```
KristinLBradley, line 9: This is a bit wordy: "It is unlikely that most consumers will need to use this named block directly for most use cases."

How about something more like:
"Most consumers are unlikely to need to use this named block directly."
-OR-
"For the majority of use cases, this name block is not required."

website/docs/components/table/partials/code/component-api.md
```diff
@@ -41,9 +53,85 @@ The Table component itself is where most of the options will be applied. However
...
+    <C.Property @name="...attributes">
+    This component supports the use of `...attributes`.
+  </C.Property>
+</Doc::ComponentApi>
+
+### Table::Th
+
+This component is not eligible to receive interactions (e.g., it cannot have an `onClick` event handler attached directly to it). Instead, an interactive element should be placed _inside_ of the `Th` element.
+
+If the `Th` component is passed as the first "cell" of a table body row, it has `scope="row"` automatically applied to it for accessibility purposes.
```
KristinLBradley, line 82:
```suggestion
If the `Th` component is passed as the first "cell" of a table body row, `scope="row"` is automatically applied for accessibility purposes.
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -41,9 +53,85 @@ The Table component itself is where most of the options will be applied. However
...
+
+<Doc::ComponentApi as |C|>
+  <C.Property @name="align" @type="enum" @values={{array "left" "center" "right" }} @default="left">
+    If set, determines the text alignment.
+  </C.Property>
+  <C.Property @name="scope" @type="string" @values={{array "col" "row" }} @default="col">
+    If used as the first item in a table body's row, `scope` should be set to `row` for accessibility purposes. Note: you only need to manually set this if you're creating a custom table using the child components, instead of the standard invocation.
```
KristinLBradley, line 89:
```suggestion
    If used as the first item in a table body’s row, `scope` should be set to `row` for accessibility purposes. Note: you only need to manually set this if you’re creating a custom table using the child components, instead of the standard invocation.
```

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -210,7 +284,7 @@ Here’s a table implementation that uses an array hash with localized strings f
 </Hds::Table>
 ```
 
-#### More Examples: replacing components as a pre-adoption step
+#### Replacing components as a pre-adoption step
 
 If you're not quite ready to replace your existing tables with this component, you can totally try out a pre-adoption spike with just the components themselves. It's a little more typing but it should give you an idea of what will work for you.
```
KristinLBradley, line 289:
```suggestion
If you’re not quite ready to replace your existing tables with this component, you can totally try out a pre-adoption spike with just the components themselves. It’s a little more typing but it should give you an idea of what will work for you.
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,38 +1,50 @@
...
-  <C.Property @name="onSort" @type="function">
-    Callback function invoked (if provided) when one of the sortable table headers is clicked, and a sorting is triggered. The function receives the values of `sortBy` and `sortOrder` as arguments.
-  </C.Property>
   <C.Property @name="isStriped" @type="boolean" @values={{array "false" "true" }} @default="false">
     Define on the table invocation. If set to `true`, row striping ("zebra striping") will be applied to the table.
   </C.Property>
-  <C.Property @name="isFixed" @type="boolean" @values={{array "false" "true" }} @default="false">
-    If set to `true`, the `table-display`(CSS) property will be set to `fixed`, which will automatically distribute columns equally based on the total width of the table.
+  <C.Property @name="isFixedLayout" @type="boolean" @values={{array "false" "true" }} @default="false">
+    If set to `true`, the `table-display`(CSS) property will be set to `fixed`, automatically distributing columns based on the total width of the table per the browser algorithm.
```
KristinLBradley, line 47: Suggestion: I would simplify this wording and link to the MDN docs for more details.

=> "If set to `true`, the `table-display`(CSS) property will be set to `fixed`."

I would either link part of the above sentence to https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout or else include the link in parenthesis after the sentence.

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,38 +1,50 @@
...
-- The `Hds::Table::Th` component is a template-only component. It supports use of [`...attributes`](https://guides.emberjs.com/release/in-depth-topics/patterns-for-components/#toc_attribute-ordering) and can contain interactive elements but is not eligible to receive interactions itself. It‘s unlikely you‘ll need to add interactive elements, as sorting is already provided.
-- The `Hds::Table::Td` component is a template-only component. It supports use of [`...attributes`](https://guides.emberjs.com/release/in-depth-topics/patterns-for-components/#toc_attribute-ordering) and can contain interactive elements (e.g., `<td><a href="user-info.html">User info</a></td>`) but is not eligible to receive interactions itself (e.g., it cannot have an `onClick` event handler attached directly to it).
+### Table
 
 <Doc::ComponentApi as |C|>
   <C.Property @name="<:head>" @type="named block">
-    This is a named block where the content for the table head (`<thead>`) is rendered.
+    This is a named block where the content for the table head (`<thead>`) is rendered. It is unlikely that most consumers will need to use this named block directly for most use cases.
```
MelSumner, line 9:
```suggestion
    This is a named block where the content for the table head (`<thead>`) is rendered. Note: most consumers are unlikely to need to use this named block directly.
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,38 +1,50 @@
...
-  <C.Property @name="onSort" @type="function">
-    Callback function invoked (if provided) when one of the sortable table headers is clicked, and a sorting is triggered. The function receives the values of `sortBy` and `sortOrder` as arguments.
-  </C.Property>
   <C.Property @name="isStriped" @type="boolean" @values={{array "false" "true" }} @default="false">
     Define on the table invocation. If set to `true`, row striping ("zebra striping") will be applied to the table.
   </C.Property>
-  <C.Property @name="isFixed" @type="boolean" @values={{array "false" "true" }} @default="false">
-    If set to `true`, the `table-display`(CSS) property will be set to `fixed`, which will automatically distribute columns equally based on the total width of the table.
+  <C.Property @name="isFixedLayout" @type="boolean" @values={{array "false" "true" }} @default="false">
+    If set to `true`, the `table-display`(CSS) property will be set to `fixed`, automatically distributing columns based on the total width of the table per the browser algorithm.
```
MelSumner, line 47:
```suggestion
    If set to `true`, the `table-display`(CSS) property will be set to `fixed` (see https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout for more details).
```

KristinLBradley (approved):

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,38 +1,50 @@
...
-  <C.Property @name="onSort" @type="function">
-    Callback function invoked (if provided) when one of the sortable table headers is clicked, and a sorting is triggered. The function receives the values of `sortBy` and `sortOrder` as arguments.
-  </C.Property>
   <C.Property @name="isStriped" @type="boolean" @values={{array "false" "true" }} @default="false">
     Define on the table invocation. If set to `true`, row striping ("zebra striping") will be applied to the table.
   </C.Property>
-  <C.Property @name="isFixed" @type="boolean" @values={{array "false" "true" }} @default="false">
-    If set to `true`, the `table-display`(CSS) property will be set to `fixed`, which will automatically distribute columns equally based on the total width of the table.
+  <C.Property @name="isFixedLayout" @type="boolean" @values={{array "false" "true" }} @default="false">
+    If set to `true`, the `table-display`(CSS) property will be set to `fixed` (see https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout for more details).
```
KristinLBradley, line 47:
```suggestion
    If set to `true`, the `table-display`(CSS) property will be set to `fixed`. (See https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout for more details.)
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,38 +1,50 @@
...
     This is a named block where the content for the table body (`<tbody>`) is rendered.
   </C.Property>
   <C.Property @name="model" @type="array">
-    If defined, sets the data source that gets yielded by the `:body` named block.
+    Indicates the data model to be used by the table.
   </C.Property>
   <C.Property @name="columns" @type="array">
-    Use an `array` hash to define your table columns. While `key` and `label` are required, other options include `isSortable`, `align` (for text-alignment), and `width`.
+  Use a `hash` within the array to define each column.
+    <Doc::ComponentApi as |D|>
```
didoo, line 19: Note: the `width` property is missing from this list

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,38 +1,50 @@
...
+      The column’s label; supports internationalization.
+      </D.Property>
+      <D.Property @name="isSortable" @type="boolean" @values={{array "false" "true" }} @default="false">
+      If set to `true`, indicates that a column should be sortable.
+      </D.Property>
+      <D.Property @name="sortingFunction" @type="function">
+      Callback function to provide support for a custom callback.
+      </D.Property>
+      <D.Property @name="align" @type="enum" @values={{array "left" "center" "right" }} @default="left">
+        If set, determines the text alignment for the column.
```
didoo, line 33: This reminds me that we should explain somewhere (probably in the "how to use") that the alignment in the cells works only for the text; if there are other elements (eg. icons with text, or buttons, etc) they need to use alternative methods (eg a flexbox container)

Notice: in a separate task/ticket, not here.

website/docs/components/table/partials/code/component-api.md
```diff
@@ -41,9 +53,85 @@ The Table component itself is where most of the options will be applied. However
...
-    Adds a (non-visible) caption for users with assistive technology. If set on a sortable table, the provided table caption is paired with the automatically generated sorted message text.
+    Adds a (non-visible) caption for users with assistive technology. If set on a sortable table, the provided table caption is paired with the automatically generated sorted message text. Note: it is best practice to provide a table caption for users with assistive technology.
   </C.Property>
   <C.Property @name="...attributes">
     Supported for the `Hds::Table` component.
   </C.Property>
+  <C.Property @name="onSort" @type="function">
+    Callback function that is invoked when one of the sortable table headers is clicked (or has a keyboard interaction performed). The function receives the values of `sortBy` and `sortOrder` as arguments.
+  </C.Property>
+</Doc::ComponentApi>
```
didoo, line 64: Note: these arguments are missing:
- `identityKey`
- `sortedMessageText`

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,38 +1,50 @@
...
-  <C.Property @name="onSort" @type="function">
-    Callback function invoked (if provided) when one of the sortable table headers is clicked, and a sorting is triggered. The function receives the values of `sortBy` and `sortOrder` as arguments.
-  </C.Property>
   <C.Property @name="isStriped" @type="boolean" @values={{array "false" "true" }} @default="false">
     Define on the table invocation. If set to `true`, row striping ("zebra striping") will be applied to the table.
   </C.Property>
-  <C.Property @name="isFixed" @type="boolean" @values={{array "false" "true" }} @default="false">
-    If set to `true`, the `table-display`(CSS) property will be set to `fixed`, which will automatically distribute columns equally based on the total width of the table.
+  <C.Property @name="isFixedLayout" @type="boolean" @values={{array "false" "true" }} @default="false">
+    If set to `true`, the `table-display`(CSS) property will be set to `fixed`. (See https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout for more details.)
```
didoo, line 47: This is not rendered as link, please use standard markdown syntax for links.

website/docs/components/table/partials/code/component-api.md
```diff
@@ -41,9 +53,85 @@ The Table component itself is where most of the options will be applied. However
     If set, determines the vertical alignment for all cell (`td`) content in a table. Does not apply to table headers (`th`).
   </C.Property>
   <C.Property @name="caption" @type="string">
-    Adds a (non-visible) caption for users with assistive technology. If set on a sortable table, the provided table caption is paired with the automatically generated sorted message text.
+    Adds a (non-visible) caption for users with assistive technology. If set on a sortable table, the provided table caption is paired with the automatically generated sorted message text. Note: it is best practice to provide a table caption for users with assistive technology.
   </C.Property>
   <C.Property @name="...attributes">
     Supported for the `Hds::Table` component.
```
didoo, line 59: Let's use the standard phrase:
```suggestion
    This component supports use of [`...attributes`](https://guides.emberjs.com/release/in-depth-topics/patterns-for-components/#toc_attribute-ordering).
```

didoo: (check also other occurrences in the page)

website/docs/components/table/partials/code/component-api.md
```diff
@@ -41,9 +53,85 @@ The Table component itself is where most of the options will be applied. However
...
   </C.Property>
+  <C.Property @name="onSort" @type="function">
+    Callback function that is invoked when one of the sortable table headers is clicked (or has a keyboard interaction performed). The function receives the values of `sortBy` and `sortOrder` as arguments.
+  </C.Property>
+</Doc::ComponentApi>
+
+### Table::Tr
+
+This component is not eligible to receive interactions (e.g., it cannot have an `onClick` event handler attached directly to it). Instead, an interactive element should be placed _inside_ of the `Th`, `ThSort`, or `Td` elements.
```
didoo, line 68: I am still not convinced of these phrase here (and also in the other sub-components below).
This content is the "Component API" while what we're describing here is "how to use" it, which is different. 
Maybe should go under the "how to use" block, or in the accessibility section?

website/docs/components/table/partials/code/component-api.md
```diff
@@ -41,9 +53,85 @@ The Table component itself is where most of the options will be applied. However
...
+If the `Th` component is passed as the first "cell" of a table body row, `scope="row"` is automatically applied for accessibility purposes.
+
+<Doc::ComponentApi as |C|>
+  <C.Property @name="align" @type="enum" @values={{array "left" "center" "right" }} @default="left">
+    If set, determines the text alignment.
+  </C.Property>
+  <C.Property @name="scope" @type="string" @values={{array "col" "row" }} @default="col">
+    If used as the first item in a table body’s row, `scope` should be set to `row` for accessibility purposes. Note: you only need to manually set this if you’re creating a custom table using the child components, instead of the standard invocation.
+  </C.Property>
+  <C.Property @name="width" @type="string" @valueNote="Any valid CSS" @default="left">
```
didoo, line 91:
```suggestion
  <C.Property @name="width" @type="string" @valueNote="Any valid CSS">
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -41,9 +53,85 @@ The Table component itself is where most of the options will be applied. However
...
+### Table::ThSort
+
+This component is not eligible to receive interactions (e.g., it cannot have an `onClick` event handler attached directly to it). Instead, an interactive element should be placed _inside_ of the `ThSort` element.
+
+This is the component that supports the column sorting, and should be used instead of `Hds::Table::Th` if creating a custom implementation.
+
+<Doc::ComponentApi as |C|>
+  <C.Property @name="isSorted" @type="boolean">
+    Set to true upon sort.
+  </C.Property>
```
didoo, line 108: let's use the standard way to declare a boolean argument
```suggestion
  <C.Property @name="isSorted" @type="boolean" @values={{array "false" "true" }} @default="false">
    Indicates if a column is sorted.
  </C.Property>
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -41,9 +53,85 @@ The Table component itself is where most of the options will be applied. However
...
+  <C.Property @name="sortOrder" @type="string" @values={{array "asc" "desc" }} @default="asc">
+    Use in conjunction with `sortBy`. If defined, indicates which direction the column should be pre-sorted in. If not defined, `asc` is applied by default.
+  </C.Property>
```
didoo, line 111: please review this property declaration, is not correct
- "asc" is not the default for the `ThSort`([see here](https://github.com/hashicorp/design-system/blob/main/packages/components/addon/components/hds/table/th-sort.js#L37-L43))
- there's no relation with the `sortBy` (which is a `Table` argument): in this case `sortOrder` simply determines, if provided, the direction of the "sorting" icon and nothing more.

website/docs/components/table/partials/code/component-api.md
```diff
@@ -41,9 +53,85 @@ The Table component itself is where most of the options will be applied. However
...
+  <C.Property @name="align" @type="enum" @values={{array "left" "center" "right" }} @default="left">
+    If set, determines the column’s text alignment.
```
didoo, line 113:
```suggestion
    If set, determines the cell's text alignment.
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -41,9 +53,85 @@ The Table component itself is where most of the options will be applied. However
...
+  </C.Property>
+  <C.Property @name="width" @type="string" @valueNote="Any valid CSS" @default="left">
```
didoo, line 115:
```suggestion
  <C.Property @name="width" @type="string" @valueNote="Any valid CSS">
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -41,9 +53,85 @@ The Table component itself is where most of the options will be applied. However
...
+    If set, determines the column’s width.
+  </C.Property>
+  <C.Property @name="onClick" @type="function">
+    The action handler; sets the sort to the column key.
```
didoo, line 119:
```suggestion
    Callback function invoked (if provided) when the element is clicked.
```

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -2,25 +2,23 @@ This component takes advantage of the `sort-by` helper provided in [ember-compos
 
 ## How to use this component
 
-### Non-sortable table
-
-If you don’t have or want to use a model, a basic invocation could look like:
+### Non-sortable table with no model defined, caption defined
```
didoo, line 5: I am not sure we should qualify normal tables as "non-sortable". They're just tables.
Plus, I would remove the "caption defined" from the title, for simplicity (if you want to stress the importance of adding a caption argument, I would add a specific example in the `More examples" or have a dedicated section about it in the "Accessibility" panel ("Best practices").

```suggestion
### Table with no model defined
```

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -164,7 +162,83 @@ You can optionally also indicate that the column defined in `@sortBy` should be
 </Hds::Table>
 ```
 
-#### More Examples: internationalized column headers, overflow menu dropdown
+#### Custom sort callback
+
+To implement a custom sort callback on a columns, (1) add a custom function as the value for `sortingFunction` in the column hash, and (2) include a custom `onSort` action in your table invocation. This is useful for cases where the key might not be A-Z or 0-9 sortable by default, i.e., status, and you’re otherwise unable to influence the shape of the data in the model. Here is an example, code truncated for clarity:
```
didoo, line 167:
```suggestion
To implement a custom sort callback on a columns, (1) add a custom function as the value for `sortingFunction` in the column hash, and (2) include a custom `onSort` action in your table invocation to track the sorting order and use it in the custom sorting function. This is useful for cases where the key might not be A-Z or 0-9 sortable by default, i.e., status, and you’re otherwise unable to influence the shape of the data in the model. Here is an example, code truncated for clarity:
```

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -164,7 +162,83 @@ You can optionally also indicate that the column defined in `@sortBy` should be
...
+</Hds::Table>
+```
+
+Here’s an example of what a custom sort function could look like. In this example, we are indicating that we want to sort on a status, which takes its order based on the position in the array:
+
+```javascript
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
```
didoo, line 195: we don't have to port all the controller code here, let's focus on what is relevant
```suggestion
```

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -164,7 +162,83 @@ You can optionally also indicate that the column defined in `@sortBy` should be
...
+// we use an array to declare the custom sorting order for the clusters' status
+const customSortingCriteriaArray = [
+  'failing',
+  'active',
+  'establishing',
+  'pending',
+];
+
+export default class ComponentsTableController extends Controller {
+  @tracked customSortOrderForClusterStatus = 'asc';
```
didoo, line 205:
```suggestion
// we track the sorting order, so it can be used in the custom sorting function
@tracked customSortOrderForStatus = 'asc';
```

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -164,7 +162,83 @@ You can optionally also indicate that the column defined in `@sortBy` should be
...
+```
+
+Here’s an example of what a custom sort function could look like. In this example, we are indicating that we want to sort on a status, which takes its order based on the position in the array:
+
+```javascript
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+// we use an array to declare the custom sorting order for the clusters' status
```
didoo, line 196:
```suggestion
// we use an array to declare the custom sorting order for the "status" column
```

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -164,7 +162,83 @@ You can optionally also indicate that the column defined in `@sortBy` should be
...
+      const index2 = customSortingCriteriaArray.indexOf(s2['status']);
+      if (index1 < index2) {
+        return this.customSortOrderForClusterStatus === 'asc' ? -1 : 1;
+      } else if (index1 > index2) {
+        return this.customSortOrderForClusterStatus === 'asc' ? 1 : -1;
+      } else {
+        return 0;
+      }
+    };
+  }
```
didoo, line 230:
```suggestion
// we define a "getter" that returns a custom sorting function ("s1" and "s2" are data records)
get customSortingMethodForStatus() {
  return (s1, s2) => {
    const index1 = customSortingCriteriaArray.indexOf(s1['status']);
    const index2 = customSortingCriteriaArray.indexOf(s2['status']);
    if (index1 < index2) {
      return this.customSortOrderForStatus === 'asc' ? -1 : 1;
    } else if (index1 > index2) {
      return this.customSortOrderForStatus === 'asc' ? 1 : -1;
    } else {
      return 0;
    }
  };
}
```

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -164,7 +162,83 @@ You can optionally also indicate that the column defined in `@sortBy` should be
...
+  get clustersWithExtraData() {
+    return this.model.clusters.map((record) => {
+      return {
+        ...record,
+        'status-sort-order': customSortingCriteriaArray.indexOf(
+          record['status']
+        ),
+      };
+    });
+  }
```
didoo, line 216: not needed
```suggestion
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,38 +1,50 @@
...
   </C.Property>
   <C.Property @name="columns" @type="array">
-    Use an `array` hash to define your table columns. While `key` and `label` are required, other options include `isSortable`, `align` (for text-alignment), and `width`.
+  Use a `hash` within the array to define each column.
+    <Doc::ComponentApi as |D|>
+      <D.Property @name="key" @type="string">
+      The column’s key; required if the column is sortable.
+      </D.Property>
+      <D.Property @name="label" @type="string" @required="true">
+      The column’s label; supports internationalization.
```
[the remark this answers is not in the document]
didoo: Plus, is already explained in the "how to" with an example.

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -164,7 +162,83 @@ You can optionally also indicate that the column defined in `@sortBy` should be
...
+        return 0;
+      }
+    };
+  }
+
+  @action
+  customOnSort(_sortBy, sortOrder) {
+    this.customSortOrderForClusterStatus = sortOrder;
+  }
+}
```
didoo, line 236:
```suggestion
// we define a callback function that listens to the `onSort` event in the table, and updates the tracked sort order values accordingly
@action
customOnSort(_sortBy, sortOrder) {
  this.customSortOrderForStatus = sortOrder;
}
```

website/docs/components/table/partials/code/how-to-use.md
didoo: Also, should we use a `<ol>` element here, instead of `(1)/(2)`?

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -210,9 +284,9 @@ Here’s a table implementation that uses an array hash with localized strings f
 </Hds::Table>
 ```
 
-#### More Examples: replacing components as a pre-adoption step
+#### Replacing components as a pre-adoption step
 
-If you're not quite ready to replace your existing tables with this component, you can totally try out a pre-adoption spike with just the components themselves. It's a little more typing but it should give you an idea of what will work for you.
+If you’re not quite ready to replace your existing tables with this component, you can totally try out a pre-adoption spike with just the components themselves. It’s a little more typing but it should give you an idea of what will work for you.
```
didoo, line 289: Two more considerations:
- the example below doesn't work (`@columns` is required, when there is a `@model`)
- I don't understand this example/paragraph. What makes this example a pre-adoption step? is using the `@model` (so is looping over records) and the yielded/contextual components.
Maybe you wanted to have an example where the table is generated using the non-yielded components? something like this?
```
<Hds::Table>
  <thead>
    <Hds::Table::Tr>
      <Hds::Table::Th>Artist</Hds::Table::Th>
      <Hds::Table::Th>Album</Hds::Table::Th>
      <Hds::Table::Th>Release Year</Hds::Table::Th>
    </Hds::Table::Tr>
  </thead>
  <tbody>
    {{#each this.myData as |record|}}
      <Hds::Table::Tr>
        <Hds::Table::Td>{{record.artist}}</Hds::Table::Td>
        <Hds::Table::Td>{{record.album}}</Hds::Table::Td>
        <Hds::Table::Td>{{record.year}}</Hds::Table::Td>
      </Hds::Table::Tr>
    {{/each}}
  </tbody>
</Hds::Table>
```

didoo: Created a ticket for this: https://hashicorp.atlassian.net/browse/HDS-1585

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,38 +1,50 @@
...
   <C.Property @name="model" @type="array">
-    If defined, sets the data source that gets yielded by the `:body` named block.
+    Indicates the data model to be used by the table.
   </C.Property>
   <C.Property @name="columns" @type="array">
-    Use an `array` hash to define your table columns. While `key` and `label` are required, other options include `isSortable`, `align` (for text-alignment), and `width`.
+  Use a `hash` within the array to define each column.
+    <Doc::ComponentApi as |D|>
+      <D.Property @name="key" @type="string">
+      The column’s key; required if the column is sortable.
```
MelSumner, line 21:
```suggestion
      The column’s key (one of the keys in the model's records); required if the column is sortable.
```

website/docs/components/table/partials/code/component-api.md
MelSumner: I'll double check this.

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,38 +1,50 @@
...
   <C.Property @name="<:body>" @type="named block">
     This is a named block where the content for the table body (`<tbody>`) is rendered.
   </C.Property>
   <C.Property @name="model" @type="array">
-    If defined, sets the data source that gets yielded by the `:body` named block.
+    Indicates the data model to be used by the table.
   </C.Property>
   <C.Property @name="columns" @type="array">
-    Use an `array` hash to define your table columns. While `key` and `label` are required, other options include `isSortable`, `align` (for text-alignment), and `width`.
+  Use a `hash` within the array to define each column.
```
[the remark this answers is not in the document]
MelSumner: Part 1: I agree and will update the definition of the columns
Part 2: Since we've already used array hashes in other components, I will not include additional (and, tbh, non-essential) information here.

website/docs/components/table/partials/code/component-api.md
MelSumner: Added

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,38 +1,50 @@
...
+    <Doc::ComponentApi as |D|>
+      <D.Property @name="key" @type="string">
+      The column’s key; required if the column is sortable.
+      </D.Property>
+      <D.Property @name="label" @type="string" @required="true">
```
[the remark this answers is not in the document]
MelSumner: Changed.

website/docs/components/table/partials/code/component-api.md
MelSumner: Added.

website/docs/components/table/partials/code/component-api.md
MelSumner: We worked through this in a video call but documenting here for visibility. 
The columns are unsorted by default. If a pre-sort is desired, the author should add `@sortBy` with a value; the column would then be pre-sorted, ascending, by default. However, if the author desires to pre-sort the column in a descending order, they would add the `@sortOrder` as well.

website/docs/components/table/partials/code/component-api.md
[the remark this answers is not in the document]
didoo: regarding the use of `hash` as a way to explain what consumers should do: in this implementation, where is the `hash`? https://github.com/hashicorp/cloud-ui/blob/main/engines/consul/addon/components/cluster-peering/peers-table/index.hbs

Not every consumer will use an inline `hash` helper in the handlebars template, so we have to write our documentation accordingly, to take in account these two different ways of providing the `@columns` argument:
- `hash` helper in the template 
- object computed in the backing class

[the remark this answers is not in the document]
MelSumner: We've taken really clear documentation and slowly mutated it to be very confusing. 

There are no clear lines between two very distinct approaches: 

1. What someone needs to understand in order to use the table the way we designed it to be used
2. All of the other things that will be useful if they are one of the edge cases that builds their own table from our component. 

My concern is that we have increased the risk of making all of this really confusing, when all I set out to do was correct outdated documentation. 

I also don't think I know _where_ exactly it started going off the rails, so I don't know how to get us back to something simpler and redirect future work to JIRA tickets. 

WDYT?

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,149 @@
...
   </C.Property>
   <C.Property @name="valign" @type="enum" @values={{array "top" "middle" }} @default="top">
     If set, determines the vertical alignment for all cell (`td`) content in a table. Does not apply to table headers (`th`).
   </C.Property>
   <C.Property @name="caption" @type="string">
-    Adds a (non-visible) caption for users with assistive technology. If set on a sortable table, the provided table caption is paired with the automatically generated sorted message text.
+    Adds a (non-visible) caption for users with assistive technology. If set on a sortable table, the provided table caption is paired with the automatically generated sorted message text. Note: it is best practice to provide a table caption for users with assistive technology.
+  </C.Property>
+  <C.Property @name="identityKey" @type="string" @default="@identity">
+    Option to pass a custom key; Ember’s default value is used if not defined.
```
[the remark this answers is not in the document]
MelSumner: I took that part out entirely, see if you prefer the truncated version.

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,149 @@
...
+  </C.Property>
+  <C.Property @name="sortedMessageText" @type="string">
+    Empty string on un-sorted column; once a sort is applied, the default message is “Sorted by (column label), (asc/desc)ing”. This is also customizable.
```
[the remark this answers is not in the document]
MelSumner: I've re-worked this section a bit.

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,149 @@
...
+        If set to `true`, indicates that a column should be sortable.
+      </D.Property>
+      <D.Property @name="align" @type="enum" @values={{array "left" "center" "right" }} @default="left">
+        If set, determines the text alignment for the column.
+      </D.Property>
+      <D.Property @name="width" @type="string" @valueNote="Any valid CSS">
+        If set, determines the column’s width.
+      </D.Property>
+      <D.Property @name="sortingFunction" @type="function">
+        Callback function to provide support for a custom callback.
```
MelSumner, line 36:
```suggestion
        Callback function to provide support for custom sorting logic. It should implement a typical bubble-sorting algorithm using two elements and comparing them. For more details, see the example of custom sorting in the How To Use section.
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -1,49 +1,149 @@
...
+<Doc::ComponentApi as |C|>
+  <C.Property @name="isSorted" @type="boolean" @values={{array "false" "true" }} @default="false">
+    Indicates if a column is sorted.
+  </C.Property>
+  <C.Property @name="sortBy" type="string" @value="column key">
+    Use to indicate which column should be pre-sorted.
+  </C.Property>
+  <C.Property @name="sortOrder" @type="string" @values={{array "asc" "desc" }} @default="asc">
+    Use in conjunction with `sortBy`. If defined, indicates which direction the column should be pre-sorted in. If not defined, `asc` is applied by default.
+  </C.Property>
```
MelSumner, line 123:
```suggestion
  <C.Property @name="sortOrder" @type="string" @values={{array "asc" "desc" }}>
    If defined, indicates which direction the column should be sorted. Controls the sort icon indicator and the `aria-sort` value.
  </C.Property>
```

didoo (approved):
🚢

PR #1203, opened 2023-02-13
`Hds::Table` - Update API to support custom sorting - Extra

didoo:
### :pushpin: Summary

This PR builds on top of #1160 and introduces some more changes to it (so they can be reviewed separately).

In this PR I have:
- removed all non-showcase content from the “dummy” page to reduce surface area of changes (will be removed anyway soon)
- refactored the data retrieval logic for the `Table` component so that I can add a new type of data structure to the model
- added a “Clusters” type of data structure to the `Table` model
- updated showcase examples for custom sorting using the `clusters` data
  - sortable table with custom sorting based on extra key added to the model
  - sortable table with custom sorting function added in the `@columns` description
- reverted changes to the `folk` data source (not needed anymore)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


MelSumner (approved):
I have no objections to this approach.

PR #1211, opened 2023-02-20
Update dependencies (patches and minors)

alex-ju:
### :pushpin: Summary

`yarn upgrade` (for patches and minor releases)

As we don't have dependabot setup for all dependencies, I took some time to bring most of the ones with a patch or minor up to date – for major releases (breaking changes) may not batch them all up in a single PR.

Left each update/set of updates per commit so we can revert if needed.

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


didoo (approved):
OK, I suppose :)

Dhaulagiri:
👏  thanks for tackling this! A couple points of feedback:

* Since we're changing dependencies in `dependencies` of the components package this would require a changeset
* There's nothing wrong per se with the updates to the different ember dependencies, but it _is_ usually useful to do them as part of a larger `npx ember-cli-update` process so we get the full breadth of updates that typically come along with each incrementally version change of the ember-cli ecosystem. I think what you have here is ok to ship but would be useful to do the full process at some point.

alex-ju:
> 👏 thanks for tackling this! A couple points of feedback:
> 
> * Since we're changing dependencies in `dependencies` of the components package this would require a changeset
> * There's nothing wrong per se with the updates to the different ember dependencies, but it _is_ usually useful to do them as part of a larger `npx ember-cli-update` process so we get the full breadth of updates that typically come along with each incrementally version change of the ember-cli ecosystem. I think what you have here is ok to ship but would be useful to do the full process at some point.

Yes, the changelog is relevant and I will be add it
Good point about the `ember-cli`, I can move those commits to a separate PR and tackle the entire ecosystem separately via `ember-cli-update`

MelSumner (approved):
Nice, thank you for doing this!

alex-ju:
> > 👏 thanks for tackling this! A couple points of feedback:
> > 
> > * Since we're changing dependencies in `dependencies` of the components package this would require a changeset
> > * There's nothing wrong per se with the updates to the different ember dependencies, but it _is_ usually useful to do them as part of a larger `npx ember-cli-update` process so we get the full breadth of updates that typically come along with each incrementally version change of the ember-cli ecosystem. I think what you have here is ok to ship but would be useful to do the full process at some point.
> 
> Yes, the changelog is relevant and I will be add it Good point about the `ember-cli`, I can move those commits to a separate PR and tackle the entire ecosystem separately via `ember-cli-update`

@Dhaulagiri I found it easier to apply the ember-cli-update here in [155f8a5](https://github.com/hashicorp/design-system/pull/1211/commits/155f8a56b57b3aca90705d28d3f05e3939770b03). one thing I haven't applied from their suggestions was a major on `ember-template-lint` which requires further lint updates, so I'd leave it for a separate PR.

PR #1212, opened 2023-02-21
Extend `Dropdown` content

alex-ju:
### :pushpin: Summary

Extend `Hds::Dropdown` with two generic containers `Hds::Dropdown::Header` and `Hds::Dropdown::Footer` to support the upcoming 'Filter' pattern.

Add `Hds::Dropdown::Header` and `Hds::Dropdown::Footer` as generic containers, before and after the dropdown list, respectively
- Add a `hds-dropdown__content` wrapper to the `hds-dropdown__list` and apply the positioning, width, and the associated style to this new wrapper element
- Prepend the `Header` element and append the `Footer` element to the dropdown list
- Allow custom `@height` to the `hds-dropdown__content` element

> **Note**
> Internal class name change: `hds-dropdown-list` → `hds-dropdown__list`
> The following modifiers will be applied to the `hds-dropdown__content` element instead of `hds-dropdown-list`
> - `hds-dropdown-list--fixed-width` → `hds-dropdown__content--fixed-width`
> - `hds-dropdown-list--position-left` → `hds-dropdown__content--position-left`
> - `hds-dropdown-list--position-right` → `hds-dropdown__content--position-right`
>
> These should not represent a breaking change as the consumers rely on the public API, but if test assertions are based on class names, nesting order, element types, or extensions/overrides have been applied to these elements they may raise issues. For this reason, we flag this change as a **breaking change** to give consumers a heads-up.

[👉 Preview showcase](https://hds-showcase-git-alex-ju-dropdown-extension-2-hashicorp.vercel.app/components/dropdown)

### :link: External links

Jira ticket: [HDS-1009](https://hashicorp.atlassian.net/browse/HDS-1109)
Figma file: https://www.figma.com/file/Kgc8BgB8jGuJXISndgTrD1/Dropdown?node-id=28361%3A64950&t=cv28452FkTz2LkFo-4

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable

[HDS-1009]: https://hashicorp.atlassian.net/browse/HDS-1009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/addon/components/hds/dropdown/footer.hbs
+++ b/packages/components/addon/components/hds/dropdown/footer.hbs
@@ -1,3 +1,3 @@
-<div class="hds-dropdown__footer {{if @hasSeparator 'hds-dropdown__footer--with-separator'}}" ...attributes>
+<div class="hds-dropdown__footer {{if @hasDivider 'hds-dropdown__footer--with-divider'}}" ...attributes>
   {{yield}}
 </div>
--- a/packages/components/addon/components/hds/dropdown/footer.hbs
+++ b/packages/components/addon/components/hds/dropdown/footer.hbs
@@ -1,3 +1,3 @@
-<div class="hds-dropdown__footer" ...attributes>
+<div class="hds-dropdown__footer{{if @hasSeparator ' hds-dropdown__footer--separator'}}" ...attributes>
   {{yield}}
 </div>
--- a/packages/components/addon/components/hds/dropdown/footer.hbs
+++ b/packages/components/addon/components/hds/dropdown/footer.hbs
@@ -1,3 +1,3 @@
-<div class="hds-dropdown__footer{{if @hasSeparator ' hds-dropdown__footer--separator'}}" ...attributes>
+<div class="hds-dropdown__footer {{if @hasSeparator 'hds-dropdown__footer--with-separator'}}" ...attributes>
   {{yield}}
 </div>
--- a/packages/components/addon/components/hds/dropdown/footer.hbs
+++ b/packages/components/addon/components/hds/dropdown/footer.hbs
@@ -1,0 +1,3 @@
+<div class="hds-dropdown__footer" ...attributes>
+  {{yield}}
+</div>
--- a/packages/components/addon/components/hds/dropdown/header.hbs
+++ b/packages/components/addon/components/hds/dropdown/header.hbs
@@ -1,3 +1,3 @@
-<div class="hds-dropdown__header" ...attributes>
+<div class="hds-dropdown__header{{if @hasSeparator ' hds-dropdown__header--separator'}}" ...attributes>
   {{yield}}
 </div>
--- a/packages/components/addon/components/hds/dropdown/header.hbs
+++ b/packages/components/addon/components/hds/dropdown/header.hbs
@@ -1,0 +1,3 @@
+<div class="hds-dropdown__header" ...attributes>
+  {{yield}}
+</div>
--- a/packages/components/addon/components/hds/dropdown/header.hbs
+++ b/packages/components/addon/components/hds/dropdown/header.hbs
@@ -1,3 +1,3 @@
-<div class="hds-dropdown__header {{if @hasSeparator 'hds-dropdown__header--with-separator'}}" ...attributes>
+<div class="hds-dropdown__header {{if @hasDivider 'hds-dropdown__header--with-divider'}}" ...attributes>
   {{yield}}
 </div>
--- a/packages/components/addon/components/hds/dropdown/header.hbs
+++ b/packages/components/addon/components/hds/dropdown/header.hbs
@@ -1,3 +1,3 @@
-<div class="hds-dropdown__header{{if @hasSeparator ' hds-dropdown__header--separator'}}" ...attributes>
+<div class="hds-dropdown__header {{if @hasSeparator 'hds-dropdown__header--with-separator'}}" ...attributes>
   {{yield}}
 </div>
--- a/packages/components/addon/components/hds/dropdown/index.hbs
+++ b/packages/components/addon/components/hds/dropdown/index.hbs
@@ -12,21 +12,25 @@
     }}
   </:toggle>
   <:content as |c|>
-    <ul class={{this.listClassNames}} {{style width=@width}} {{did-insert this.didInsertList}}>
-      {{yield
-        (hash
-          close=c.close
-          Checkbox=(component "hds/dropdown/list-item/checkbox")
-          Checkmark=(component "hds/dropdown/list-item/checkmark")
-          CopyItem=(component "hds/dropdown/list-item/copy-item")
-          Description=(component "hds/dropdown/list-item/description")
-          Generic=(component "hds/dropdown/list-item/generic")
-          Interactive=(component "hds/dropdown/list-item/interactive")
-          Radio=(component "hds/dropdown/list-item/radio")
-          Separator=(component "hds/dropdown/list-item/separator")
-          Title=(component "hds/dropdown/list-item/title")
-        )
-      }}
-    </ul>
+    <div class={{this.classNames}} {{style width=@width height=@height}}>
+      {{yield (hash Header=(component "hds/dropdown/header"))}}
+      <ul class="hds-dropdown__list" {{did-insert this.didInsertList}}>
+        {{yield
+          (hash
+            close=c.close
+            Checkbox=(component "hds/dropdown/list-item/checkbox")
+            Checkmark=(component "hds/dropdown/list-item/checkmark")
+            CopyItem=(component "hds/dropdown/list-item/copy-item")
+            Description=(component "hds/dropdown/list-item/description")
+            Generic=(component "hds/dropdown/list-item/generic")
+            Interactive=(component "hds/dropdown/list-item/interactive")
+            Radio=(component "hds/dropdown/list-item/radio")
+            Separator=(component "hds/dropdown/list-item/separator")
+            Title=(component "hds/dropdown/list-item/title")
+          )
+        }}
+      </ul>
+      {{yield (hash close=c.close Footer=(component "hds/dropdown/footer"))}}
+    </div>
   </:content>
 </Hds::Disclosure>
--- a/packages/components/addon/components/hds/dropdown/index.js
+++ b/packages/components/addon/components/hds/dropdown/index.js
@@ -31,19 +31,19 @@ export default class HdsDropdownIndexComponent extends Component {
   }
 
   /**
-   * Get the class names to apply to the "list"
-   * @method DropdownIndex#listClassNames
-   * @return {string} The "class" attribute to apply to the "list" element
+   * Get the class names to apply to the content
+   * @method classNames
+   * @return {string} The "class" attribute to apply to the disclosed content
    */
-  get listClassNames() {
-    let classes = ['hds-dropdown-list'];
+  get classNames() {
+    let classes = ['hds-dropdown__content'];
 
     // add a class based on the @listPosition argument
-    classes.push(`hds-dropdown-list--position-${this.listPosition}`);
+    classes.push(`hds-dropdown__content--position-${this.listPosition}`);
 
     // add a class based on the @width argument
     if (this.args.width) {
-      classes.push('hds-dropdown-list--fixed-width');
+      classes.push('hds-dropdown__content--fixed-width');
     }
 
     return classes.join(' ');
--- a/packages/components/app/styles/components/dropdown.scss
+++ b/packages/components/app/styles/components/dropdown.scss
@@ -156,7 +156,7 @@ $hds-dropdown-toggle-border-radius: 5px;
   }
 }
 
-.hds-dropdown__header--separator {
+.hds-dropdown__header--with-separator {
   border-bottom: 1px solid var(--token-color-border-primary);
 }
 
@@ -180,7 +180,7 @@ $hds-dropdown-toggle-border-radius: 5px;
   }
 }
 
-.hds-dropdown__footer--separator {
+.hds-dropdown__footer--with-separator {
   border-top: 1px solid var(--token-color-border-primary);
 }
 
--- a/packages/components/app/styles/components/dropdown.scss
+++ b/packages/components/app/styles/components/dropdown.scss
@@ -115,37 +115,69 @@ $hds-dropdown-toggle-border-radius: 5px;
 // UL ELEMENT
 // GOES INSIDE HDS::DISCLOSURE's :content block
 
-.hds-dropdown-list {
+.hds-dropdown__content {
+  display: flex;
+  flex-direction: column;
   width: max-content; // notice: this is important because being in a position absolute means the layout algorithm assigns a width of 0 and this impacts on the flex algorithm of the children (in some cases they don't use the full width)
   min-width: 200px;
   max-width: 400px;
-  margin: 0;
-  padding: 4px 0;
-  list-style: none;
   background-color: var(--token-color-surface-primary);
   border-radius: 6px;
   box-shadow: var(--token-surface-high-box-shadow);
 }
 
-.hds-dropdown-list--fixed-width {
+.hds-dropdown__content--fixed-width {
   min-width: initial;
   max-width: initial;
 }
 
-.hds-dropdown-list--position-right {
+.hds-dropdown__content--position-right {
   position: absolute;
   top: calc(100% + 4px);
   right: 0;
   z-index: 2; // https://github.com/hashicorp/design-system/issues/114
 }
 
-.hds-dropdown-list--position-left {
+.hds-dropdown__content--position-left {
   position: absolute;
   top: calc(100% + 4px);
   left: 0;
   z-index: 2; // https://github.com/hashicorp/design-system/issues/114
 }
 
+.hds-dropdown__header {
+  flex: none;
+  padding: 6px 8px;
+}
+
+.hds-dropdown__list {
+  flex: 1 1 auto;
+  margin: 0;
+  padding: 4px 0;
+  overflow-y: auto;
+  list-style: none;
+  overscroll-behavior: contain;
+}
+
+.hds-dropdown__footer {
+  position: relative;
+  flex: none;
+  padding: 8px 8px 6px;
+
+  .hds-link-standalone {
+    width: initial;
+    padding: 5px 8px;
+  }
+
+  &::after {
+    position: absolute;
+    top: 0;
+    right: 6px;
+    left: 6px;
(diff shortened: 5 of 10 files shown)


didoo:
> Note: these changes should not represent a breaking change if the consumers rely on the public API, but if test assertions are based on class names or element types or extensions/overrides have been applied to dropdown elements they may raise issues.

@alex-ju I can definitely confirm you that this has already happened (a lot) in Cloud UI, and so unfortunately we will have to:
- consider this release as a breaking change (even if technically is not)
- once released, open a PR in Cloud UI to bump the `@hashicorp/design-system-components` version in Cloud UI (and potentially other consumers?) 
  - in that PR (or in other PRs, depends on the complexity) we will have to fix all the cases where the changes introduced here (and potentially in #1185) are breaking selectors in CSS or tests (or CSS class names applied to markup elements)

As you can see, this will add a lot of overhead and extra work for us, that was not initially planned. For this reason we should involve @Dhaulagiri an Misha and plan the work with them (we will need the support of the consuming teams, and their approvals of the PRs we will open to fix all the issues we will find).
We should also raise this problem with the engineering ambassadors, and discuss with them how to mitigate this _"bad usage if HDS class names"_ in the future.

/cc @hashicorp/hds-engineering

packages/components/addon/components/hds/dropdown/index.hbs
```diff
@@ -8,21 +8,25 @@
...
+            Description=(component "hds/dropdown/list-item/description")
+            Generic=(component "hds/dropdown/list-item/generic")
+            Interactive=(component "hds/dropdown/list-item/interactive")
+            Radio=(component "hds/dropdown/list-item/radio")
+            Separator=(component "hds/dropdown/list-item/separator")
+            Title=(component "hds/dropdown/list-item/title")
+          )
+        }}
+      </ul>
+      {{yield (hash close=c.close Footer=(component "hds/dropdown/footer"))}}
```
didoo, line 29: regarding the discussion we had yesterday, about inverting the order of the elements when the dropdown appears on top (like in the `org selector` in the sidebar navigation) there's a much simpler solution: just use the `footer` instead of the `header` :) 

so, all good here!

packages/components/addon/components/hds/dropdown/index.js
```diff
@@ -26,18 +26,18 @@ export default class HdsDropdownIndexComponent extends Component {
 
   /**
    * Get the class names to apply to the "list"
-   * @method DropdownIndex#listClassNames
+   * @method DropdownIndex#contentClassNames
```
didoo, line 29:
```suggestion
   * @method contentClassNames
```

packages/components/addon/components/hds/dropdown/footer.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="hds-dropdown-footer" ...attributes>
```
didoo, line 1: should we use a more BEM-y notation here? `hds-dropdown__footer` (no strong opinions, I see benefits also in keeping it as is, to differentiate from the other `hds-dropdown__***` sub-elements)

packages/components/app/styles/components/dropdown.scss
```diff
@@ -110,37 +110,64 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+  list-style: none;
+  overscroll-behavior: contain;
+}
+
+.hds-dropdown-footer {
+  position: relative;
+  flex: none;
+  padding: 10px 8px 6px;
+
+  &::after {
```
didoo, line 162: I don't think this should be implemented this way. My understanding, looking at the Figma designs, is that the line above the content of the footer is a `separator` (though, with a different size).

Plus, I suggest to review and discuss the Figma implementation with @heatherlarsen (or @jorytindall if there's already been an handover) because in Figma they're using a `Generic item` to mimic the footer (and maybe the header too) but they should be separate generic containers, with their own padding specifications to avoid misalignments between code and design.

packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -116,46 +116,52 @@
...
-        <Hds::Dropdown::ListItem::Separator />
-        <Hds::Dropdown::ListItem::Interactive
-          @route="index"
-          @text="A longer item that could span multiple lines if the characters surpass a certain length"
-        />
-      </ul>
-      {{! template-lint-enable no-inline-styles }}
+      <div class="hds-dropdown-content">
+        {{! template-lint-disable no-inline-styles }}
+        <ul class="hds-dropdown-list" style="width: 250px">
```
didoo, line 150: I _think_ the width should be applied to the content now, not the list, right?

packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -259,112 +283,136 @@
     <div>
       <span class="dummy-text-small">base</span>
       {{! Notice: we want to emulate the case of a fixed width list }}
-      <ul class="hds-dropdown-list" style="width: 250px">
-        <Hds::Dropdown::ListItem::CopyItem @text="91ee1e8ef65b337f0e70d793f456c71d" />
-      </ul>
+      <div class="hds-dropdown-content">
+        <ul class="hds-dropdown-list" style="width: 250px">
```
didoo, line 287: same here?

packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -259,112 +283,136 @@
...
     <div>
       <span class="dummy-text-small">with copyItemTitle</span>
-      <ul class="hds-dropdown-list" style="width: 250px">
-        <Hds::Dropdown::ListItem::CopyItem
-          @copyItemTitle="Lorem ipsum dolor"
-          @text="91ee1e8ef65b337f0e70d793f456c71d"
-        />
-      </ul>
+      <div class="hds-dropdown-content">
+        <ul class="hds-dropdown-list" style="width: 250px">
```
didoo, line 295: same here?

packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -259,112 +283,136 @@
...
     {{! template-lint-enable no-inline-styles }}
   </div>
   <h6 class="dummy-h6">States</h6>
   <div class="dummy-dropdown-list-items-base-sample">
     {{! Notice: we want to emulate the case of a fixed width list }}
     {{! template-lint-disable no-inline-styles }}
-    <ul class="hds-dropdown-list" style="width: 250px">
-      {{#each @model.ITEM_STATES as |state|}}
+    <div class="hds-dropdown-content">
+      <ul class="hds-dropdown-list" style="width: 250px">
```
didoo, line 310: same here?

didoo (commented):
@alex-ju as discussed, I had a first pass at the changes, and added a few comments

packages/components/app/styles/components/dropdown.scss
didoo: For reference: https://www.figma.com/file/U5cpVPoELaUktOex3ByG1j?node-id=0:1#375922763

(plus, it came to mind that the separator internally uses an `<li>` element, so... we have to find a way to or use it directly (with a different tag) or make its styling shareable).

(maybe we could consider an optional `@hasSeparator` argument for the `header/footer` elements?)

packages/components/addon/components/hds/dropdown/footer.hbs
alex-ju: Yeah, good call here. I think the element notation would make more sense here, indeed, so I've updated to `hds-dropdown__footer` as we don't expect this to be invoked on its own, we've only placed it in a separate file to allow splattributes.

packages/components/tests/dummy/app/templates/components/dropdown.hbs
didoo, line 1: I imagine the Helios website documentation will be updated in a follow-up ticket?

packages/components/app/styles/components/dropdown.scss
```diff
@@ -115,37 +115,68 @@ $hds-dropdown-toggle-border-radius: 5px;
...
+  list-style: none;
+  overscroll-behavior: contain;
+}
+
+.hds-dropdown__footer {
+  position: relative;
+  flex: none;
+  padding: 8px 8px 6px;
+
+  .hds-link-standalone {
```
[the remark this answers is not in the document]
alex-ju: The only alternative I could think of to avoid this, would be to have the link as contextual component for the footer subcomponent.

So instead of 
```hbs
<Hds::Dropdown::Footer>
  <Hds::Link::Standalone @icon="list" @text="Organizations" @color="secondary" @href="#" />
</Hds::Dropdown::Footer>
```

we'd have
```hbs
<Hds::Dropdown::Footer as |DF|>
  <DF.Link @icon="list" @text="Organizations" @color="secondary" @href="#" />
</Hds::Dropdown::Footer>
```

The tradeoff would be that for generic content, instead of

```hbs
<Hds::Dropdown::Footer>
  <Hds::ButtonSet>
    <Hds::Button @text="Apply" @isFullWidth={{true}} @size="small" />
    <Hds::Button @text="Cancel" @color="secondary" @isFullWidth={{true}} @size="small" />
  </Hds::ButtonSet>
</Hds::Dropdown::Footer>
```

we'd have an additional level of nesting

```hbs
<Hds::Dropdown::Footer as |DF|>
  <DF.Generic>
    <Hds::ButtonSet>
      <Hds::Button @text="Apply" @isFullWidth={{true}} @size="small" />
      <Hds::Button @text="Cancel" @color="secondary" @isFullWidth={{true}} @size="small" />
    </Hds::ButtonSet>
  </DF.Generic>
</Hds::Dropdown::Footer>
```

packages/components/app/styles/components/dropdown.scss
alex-ju: I remember discussing this with @heatherlarsen last week, making the point that in code using a `list-item/separator` doesn't make sense semantically, and a simple border makes more sense (which is the current implementation). Not sure how feasible it is for this to be done in Figma though

packages/components/addon/components/hds/dropdown/index.hbs
```diff
@@ -12,18 +12,22 @@
...
-          CopyItem=(component "hds/dropdown/list-item/copy-item")
-          Description=(component "hds/dropdown/list-item/description")
-          Generic=(component "hds/dropdown/list-item/generic")
-          Interactive=(component "hds/dropdown/list-item/interactive")
-          Separator=(component "hds/dropdown/list-item/separator")
-          Title=(component "hds/dropdown/list-item/title")
-        )
-      }}
-    </ul>
+    <div class={{this.contentClassNames}} {{style width=@width height=@height}}>
```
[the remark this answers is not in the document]
alex-ju: Updated to `classNames` to have consistency across components.

packages/components/addon/components/hds/dropdown/index.hbs
```diff
@@ -12,18 +12,22 @@
...
+            CopyItem=(component "hds/dropdown/list-item/copy-item")
+            Description=(component "hds/dropdown/list-item/description")
+            Generic=(component "hds/dropdown/list-item/generic")
+            Interactive=(component "hds/dropdown/list-item/interactive")
+            Separator=(component "hds/dropdown/list-item/separator")
+            Title=(component "hds/dropdown/list-item/title")
+          )
+        }}
+      </ul>
+      {{yield (hash close=c.close Footer=(component "hds/dropdown/footer"))}}
```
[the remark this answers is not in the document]
alex-ju: Gave it another go now. Unfortunately, if `close` is not forwarded via hashing it can't be called by elements in the footer as in the example below.

```hbs
<Hds::Dropdown @listPosition="left" as |dd|>
  <dd.ToggleButton @text="Checkbox" @color="secondary" />
  <dd.Footer>
    <Hds::Button @text="Apply filters" @isFullWidth={{true}} {{on "click" dd.close}} @size="small" />
  </dd.Footer>
</Hds::Dropdown>
```

packages/components/app/styles/components/dropdown.scss
alex-ju: @didoo, I've added the `@hasSeparator` argument, as suggested, so this should be good for another review

packages/components/addon/components/hds/dropdown/footer.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="hds-dropdown__footer{{if @hasSeparator ' hds-dropdown__footer--separator'}}" ...attributes>
```
didoo, line 1:
```suggestion
<div class="hds-dropdown__footer {{if @hasSeparator 'hds-dropdown__footer--separator'}}" ...attributes>
```

packages/components/addon/components/hds/dropdown/header.hbs
```diff
@@ -0,0 +1,3 @@
+<div class="hds-dropdown__header{{if @hasSeparator ' hds-dropdown__header--separator'}}" ...attributes>
```
didoo, line 1:
```suggestion
<div class="hds-dropdown__header {{if @hasSeparator 'hds-dropdown__header--separator'}}" ...attributes>
```

packages/components/addon/components/hds/dropdown/footer.hbs
didoo, line 1:
```suggestion
<div class="hds-dropdown__footer {{if @hasSeparator 'hds-dropdown__footer--separator'}}" ...attributes>
```

packages/components/app/styles/components/dropdown.scss
```diff
@@ -115,37 +115,69 @@ $hds-dropdown-toggle-border-radius: 5px;
...
   z-index: 2; // https://github.com/hashicorp/design-system/issues/114
 }
 
+.hds-dropdown__header {
+  position: relative;
+  flex: none;
+  padding: 6px 8px;
+}
+
+.hds-dropdown__header--separator {
```
didoo, line 154: not a big issue, but probably sounds better to read as `--with-separator` or `--has-separator` (same for the footer)

packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -287,23 +315,176 @@
...
-          mock-state-selector="button"
-        />
-      </ul>
+        </ul>
+      </div>
       {{! template-lint-enable no-inline-styles }}
     </SF.Item>
   </Shw::Flex>
 
+  <Shw::Text::H4>Header and footer</Shw::Text::H4>
```
didoo, line 339: I think we're missing one example where the header has a separator

didoo (commented):
@alex-ju added a few minor comments

packages/components/tests/dummy/app/templates/components/dropdown.hbs
alex-ju: Updated the examples to showcase what they would look like with or without separators

packages/components/app/styles/components/dropdown.scss
alex-ju: went with `--with-separator` for now; I've looked for similar examples of translating a boolean into a modifier but couldn't find anything at hand (`--with-[]` or `--has-[]`)

packages/components/addon/components/hds/dropdown/header.hbs
alex-ju: I tend to do this to have clean mark-up (`"hds-dropdown__header"` instead of ``"hds-dropdown__header "``, but I appreciate it's easier to read in the suggested form so I've updated accordingly

didoo (approved):
Good to go! ✅

heatherlarsen:
@alex-ju I've updated the padding of all examples in the header and footer of the List in the Figma component based on our conversation today (tagged you in Figma as well).

- Generic blocks - `padding: 0px 8px`
- Search (within the Header) - `padding: 8px`
- Button (within the Footer) - `padding: 8px 8px 4px 8px`
- ButtonSet (within the Footer) - `padding: 8px 8px 4px 8px`, gap between buttons: 8px (instead of the standard 16px as the buttons are smaller and the overall container is smaller, so the standard gap is too large)
- Standalone Link - `padding: 12px 16px 8px 16px`

This is with the understanding that we're allowing them to put whatever they want into the header and footer but that in Figma, we'll provide a few pre-defined options to help ensure consistent spacing for those common elements.

Additionally, I've updated the property for the divider to `hasDivider` instead of `hasSeparator` as that seems to be what we favor in other components. 

I believe this is in alignment with what we discussed earlier, but let me know if I misunderstood or if you have concerns about these. Thanks!

alex-ju:
> @alex-ju I've updated the padding of all examples in the header and footer of the List in the Figma component based on our conversation today (tagged you in Figma as well).
> 
> * Generic blocks - `padding: 0px 8px`
> * Search (within the Header) - `padding: 8px`
> * Button (within the Footer) - `padding: 8px 8px 4px 8px`
> * ButtonSet (within the Footer) - `padding: 8px 8px 4px 8px`, gap between buttons: 8px (instead of the standard 16px as the buttons are smaller and the overall container is smaller, so the standard gap is too large)
> * Standalone Link - `padding: 12px 16px 8px 16px`

I tried my best to keep in sync with the design (and the 8px spacing unit in general), but often these values are not always reflected in CSS for various reasons (e.g. the Standalone link has a transparent border that needs to be subtracted from the padding) although the final layout is a perfect match.

> This is with the understanding that we're allowing them to put whatever they want into the header and footer but that in Figma, we'll provide a few pre-defined options to help ensure consistent spacing for those common elements.

Correct, that's the approach we're going to take on the code side

> Additionally, I've updated the property for the divider to `hasDivider` instead of `hasSeparator` as that seems to be what we favor in other components.

Updated

> I believe this is in alignment with what we discussed earlier, but let me know if I misunderstood or if you have concerns about these. Thanks!

Indeed very much in alignment.

One thing that we discussed and I wasn't able to deliver on is the top/bottom padding to be applied to the container (not the `<ul>` of interactive items). I know this is not reflected in Figma, but wanted to flag it as it has implications for how the calculations needed to be different in code (as pointed out above). Applying this change would have effects on the focus states of interactive items (as shown below)

<table>
<tr><th>Padding applied to the list (current)</th><th>Padding applied to the container</th></tr>
<tr><td>

Screenshot 2023-03-24 at 12 45 41

</td><td>

Screenshot 2023-03-24 at 11 41 31


</td></tr>
</table>

heatherlarsen (approved):
Excited to see all these moving pieces coming together :shipit:

PR #1218, opened 2023-02-22
Remove visual regression testing from `ember-flight-icons`

alex-ju:
### :pushpin: Summary

Remove visual regression testing from `ember-flight-icons`

### :link: External links

Jira ticket: [HDS-1550](https://hashicorp.atlassian.net/browse/HDS-1550)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-1550]: https://hashicorp.atlassian.net/browse/HDS-1550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/ci-ember-flight-icons.yml
```diff
@@ -31,12 +31,6 @@ jobs:
       - name: Lint
         run: yarn run lint
         working-directory: packages/ember-flight-icons
-      - name: Run Tests
-        run: yarn run test:ember:percy
-        working-directory: packages/ember-flight-icons
```
Dhaulagiri, line 36: We need to keep this section so the tests still run, but remove the `:percy` section from the yarn command

Dhaulagiri (changes requested):
Overall looks good, but we need to put back the test command so the non-percy tests still run

.github/workflows/ci-ember-flight-icons.yml
Dhaulagiri: or do we 🤔 ? They're also running as part of the ember-try scenarios so maybe this is correct and I was off base with my initial response.

alex-ju: Right, we still have a few ember tests running

alex-ju: yeah, I think the matrix test should cover this too

Dhaulagiri: I guess the only reason to keep it is that the version of ember specified in our project could be different than the `ember-release` scenario.

I think what you have here is good now but it would be worth discussing with the team whether it makes sense for us to run these tests 3 times.

Dhaulagiri (approved):

.github/workflows/ci-ember-flight-icons.yml
alex-ju: @hashicorp/hds-engineering

didoo: could we imagine (would it be possible) to move these tests inside the `packages/components` test suite, like I did for the Percy test? Or because it's an external component (coming from an npm package, I mean) they would not work?

Dhaulagiri: I think technically it would work since one depends on the other already (it'd be easy to try and see). It's definitely _unorthodox_ to have tests for a package inside another package, but given how our packages are effectively tied together and we never change the ember-flight-icons package the risks seem minimal.

This opens up another can of works around whether or not it make sense for these to be separate packages at all, but we could consider that separately...

MelSumner: > This opens up another can of works around whether or not it make sense for these to be separate packages at all, but we could consider that separately...

It doesn't make sense to make them separate, I don't think it's a can of worms, and we should make a story to combine them :)

alex-ju: It sounds like we're in agreement that running the `ember-flight-icons` tests separately doesn't make much sense, so I [moved the icons tests to our components package](https://github.com/hashicorp/design-system/pull/1218/commits/48a3dec640c2e9710eaad2aa2d230345db65e9a9) and [removed the `ember-flight-icons` CI testing](https://github.com/hashicorp/design-system/pull/1218/commits/a477f989dcb2f0fb4893fd2b4a210afe36979025). 

I can also record a card to look into/investigate the impact of moving the `ember-flight-icons` into `components`.

didoo (approved):
nice job!

PR #1222, opened 2023-02-24
Add “aliases” information to the tokens list in the website documentation

didoo:
### :pushpin: Summary

During conversation with Jake Herschman, he suggested to add the information if which aliases there were associated to a token. Since it's a good idea, and it was simple to do, I've made the changes straightaway (he's in the middle of a large work of mapping colors between Structure and HDS, and this can help him).

Preview: https://hds-website-git-docs-tokens-aliases-hashicorp.vercel.app/foundations/tokens?searchQuery=token-color-foreground-faint

screenshot_2410

***

### 👀 Reviewer's checklist:

### DIFF
--- a/website/app/components/doc/meta-row.hbs
+++ b/website/app/components/doc/meta-row.hbs
@@ -6,12 +6,24 @@
 <div class="doc-meta-row" ...attributes>
   <div class="doc-meta-row__label">{{@label}}</div>
   <div class="doc-meta-row__value">
+    {{! when we pass a single value, we have two different use cases to support }}
     {{#if @copyable}}
       <Doc::CopyButton @textToCopy={{@valueToCopy}} @textToShow={{(or @valueToShow @valueToCopy)}} @type="ghost" />
-    {{else}}
+    {{/if}}
+    {{#if @valueToShow}}
       <code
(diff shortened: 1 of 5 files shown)


KristinLBradley:
Would we want to add the "--" at the beginning of the variable? https://developer.mozilla.org/en-US/docs/Web/CSS/--*

To me that would make sense and it would help me to recognize more easily at a glance that those are CSS custom properties vs. some other kind of variable or token names.

didoo:
> Would we want to add the "--" at the beginning of the variable? https://developer.mozilla.org/en-US/docs/Web/CSS/--*
> 
> To me that would make sense and it would help me to recognize more easily at a glance that those are CSS custom properties vs. some other kind of variable or token names.

I'm not sure why I didn't add it before (it was done in a previous PR). Fixed now.

MelSumner:
I don't object but perhaps @heatherlarsen would be more appropriate for reviewing this change.

heatherlarsen:
I don't know that this is the most scalable approach, as I can see the alias lists becoming quite lengthy, particularly for neutral values, but design doesn't have the bandwidth to come up with an alternative solution at the moment. Feel free to ship this, and we can revisit improvements later.

heatherlarsen (approved):

PR #1226, opened 2023-02-27
`Table` - Change internal logic for `Thsort` arguments conditionals

didoo:
### :pushpin: Summary

While working on the documentation update for the `Table` component (#1198) we have realized that the internal logic for how the `ThSort` component is rendered based on the sorting arguments was working, but technically not correct (the `isSorted` argument was redundant, and when used as standalone component the `ThSort` element could not render the icons states just on the base of the `sortOrder` argument).

(A follow-up PR will also update the showcase to collect more use cases for the `Table` base elements).

In this PR I have:

- updated the internal logic of the `ThSort` to rely only on the `sortOrder` argument to render its sorting state

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/table/th-sort.js
+++ b/packages/components/addon/components/hds/table/th-sort.js
@@ -19,11 +19,14 @@ export default class HdsTableThSortComponent extends Component {
    * @description Sets the aria-sort attribute based on the sort order defined; acceptable values are ascending, descending, none(default) and other. Authors SHOULD only apply this property to table headers or grid headers. If the property is not provided, there is no defined sort order. For each table or grid, authors SHOULD apply aria-sort to only one header at a time.
    */
   get ariaSort() {
-    if (this.args.isSorted) {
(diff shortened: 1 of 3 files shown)


MelSumner:
Acknowledging this PR, need to test it out.

packages/components/addon/components/hds/table/th-sort.js
```diff
@@ -35,11 +38,20 @@ export default class HdsTableThSortComponent extends Component {
...
-      return this.args.sortOrder === 'asc' ? 'arrow-up' : 'arrow-down';
-    } else {
-      return 'swap-vertical';
+    switch (this.args.sortOrder) {
+      case 'asc':
+        return 'arrow-up';
+      case 'desc':
+        return 'arrow-down';
+      default:
+        // none is the default per the spec.
```
MelSumner, line 47:
```suggestion
```

MelSumner (approved):

PR #1227, opened 2023-02-27
`Table` - Update showcase for "base elements"

didoo:
### :pushpin: Summary

This is a follow-up of #1226 where we add more use cases to the showcase page for the `Table` component, to cover the "base elements" of the table.

Preview: https://hds-showcase-git-table-showcase-hashicorp.vercel.app/components/table (see bottom of the page)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/tests/dummy/app/templates/components/table.hbs
+++ b/packages/components/tests/dummy/app/templates/components/table.hbs
@@ -456,50 +456,9 @@
     </:body>
   </Hds::Table>
 
-  <Shw::Text::H2>Table elements</Shw::Text::H2>
+  <Shw::Divider @level={{2}} />
 
-  <Shw::Text::H3>Head</Shw::Text::H3>
-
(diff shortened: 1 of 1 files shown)


MelSumner:
It seems to me that quite a few of these new showcase elements are showcasing things that are not the responsibility of the table component. What was the rationale for adding these? Thinking about other components where we yield out the contents and the consumer is responsible for them.

didoo:
> It seems to me that quite a few of these new showcase elements are showcasing things that are not the responsibility of the table component. What was the rationale for adding these? Thinking about other components where we yield out the contents and the consumer is responsible for them.

I've added only showcases for the "base elements". If they are used as yielded component or directly, is not relevant in terms of showcasing how they behave, and capture their variants. No?
I am not sure what is different from the showcase of the `Pagination` base components, of the showcase of the `Dropdown` base elements.

MelSumner (approved):

PR #1228, opened 2023-02-27
fix analytics bug: add url override to head-data overrides

Dhaulagiri:
### :pushpin: Summary

This will update the `url` property to include the actual full url for each page on the site, fixing a bug with our `canonical` tag that adversely impacts our analytics.

[The Fathom docs hint at this being the root cause](https://usefathom.com/docs/troubleshooting/canonicals):
> Misusing canonical meta tags typically happens when you have a canonical meta tag present on every page, telling analytics and search engines that every page on your site is the homepage (this is obviously incorrect).
>
> If that tag is present on every page, then you’re telling everyone that every page on your site is your homepage. Once you fix this, your stats will show data for all pages...

This tracks with the changes in https://github.com/hashicorp/design-system/pull/1071 landing right around the time we lost this data granularity. This change has meant that the `canonical` tag for **all** pages on the site is https://helios.hashicorp.design/. Oops!

This PR attempts to manually set `url` correctly based on the current route. I did some manual smoke testing of this but please have a 2nd look if you can. 

#### How to test

Visit pages on the site and view source to see the value `link rel="canonical"`

#### Before
image-20230224-203543

#### After (in  a test account)
Screenshot 2023-02-27 at 10 12 57

### :link: External links

### DIFF
--- a/website/app/services/head-data.js
+++ b/website/app/services/head-data.js
@@ -5,8 +5,11 @@
 
 import HeadDataService from 'ember-meta/services/head-data';
 import config from 'ember-get-config';
+import { inject as service } from '@ember/service';
 
 export default class CustomHeadDataService extends HeadDataService {
+  @service router;
+
   get title() {
     return (
       this.currentRouteMeta?.title ??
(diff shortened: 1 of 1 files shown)


website/app/services/head-data.js
```diff
@@ -28,4 +31,8 @@ export default class CustomHeadDataService extends HeadDataService {
       : 'assets/logos/share-card.jpg';
     return `${config['ember-meta'].url}/${previewImage}`;
   }
+
+  get url() {
+    return `${config['ember-meta'].url}/${this.router.currentURL}`;
```
Dhaulagiri, line 36: should figure out how to not do this if we are on the home page (it adds `//` currently)

MelSumner (approved):

PR #1229, opened 2023-02-28
Website: add accessible names to lists on home page

MelSumner:
### :pushpin: Summary

If merged, this PR adds accessible names to the lists on the home page. It makes no visible changes to the UI, just provides necessary information to the user with a screen-reader. Closes HDS-1435.

- added an `aria-labelledby` attribute to the `ul` element inside of the primary nav
- added an `aria-label` to the list of four cards

Before (neither list has an accessible name): 
CleanShot 2023-02-28 at 14 49 54

After the changes were made, it's clear that the "list" elements have accessible names by inspecting the accessibility tree:
CleanShot 2023-02-28 at 14 36 47

### :link: External links

Jira ticket: [HDS-1435](https://hashicorp.atlassian.net/browse/HDS-1435)
Figma file: [if it applies]

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-1435]: https://hashicorp.atlassian.net/browse/HDS-1435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/templates/index.hbs
+++ b/website/app/templates/index.hbs
@@ -26,7 +26,7 @@
         />
       </div>
     </div>
-    <ul class="doc-page-home__cards" role="list">
+    <ul class="doc-page-home__cards" role="list" aria-label="the cornerstones of the helios design system">
       {{#each this.cards as |card|}}
         <li class="doc-page-home__card">
           <p class="doc-page-home__card-title">{{capitalize card.title}}</p>
--- a/website/app/components/doc/page/header/index.hbs
+++ b/website/app/components/doc/page/header/index.hbs
@@ -8,8 +8,8 @@
   <LinkTo @route="index" class="doc-page-header__logo" aria-label="home page">
     <Doc::Logo::DesignSystem />
   </LinkTo>
-  <nav class="doc-page-header__nav-menu" aria-label="primary navigation">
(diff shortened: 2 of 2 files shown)


website/app/templates/index.hbs
```diff
@@ -26,7 +26,7 @@
         />
       </div>
     </div>
-    <ul class="doc-page-home__cards" role="list">
+    <ul class="doc-page-home__cards" role="list" aria-label="the cornerstones of the helios design system">
```
didoo, line 29: Just a question: is this all lowercase on purpose?

didoo (approved):

website/app/templates/index.hbs
MelSumner: It's lowercase because it doesn't matter in the case of a screen-reader. It would matter if there was an acronym (those should be all uppercase and are read out separately). But in this instance, it is fine. 👍

didoo: yeah, what I imagined (but wasn't sure). thanks for the clarification

PR #1232, opened 2023-03-03
Make `ember-body-class` a dev dependency in components

### LINKED ISSUE #1231: 1.7.0 includes `ember-body-class` as a dependency.
## IS: Current behavior
When updating, we found that [`ember-body-class`](https://github.com/mansona/ember-body-class) ended up in our app dependencies.

It looks like it was added here on accident: https://github.com/hashicorp/design-system/pull/1170

This is a blocker to upgrade to this version because it re-opens Ember's Route class to add a class to `<body>` based on the name of the current route. 

Design system upgrades should not include a dependency on `ember-body-class`.

alex-ju:
### :pushpin: Summary

Make `ember-body-class` a dev dependency in `packages/components`.

This dependency is used in `website`. Fixes #1231

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


Dhaulagiri (approved):
approved if you also add a patch changeset

Dhaulagiri:
although it looks like the showcase does rely on this so we should move it to `devDependencies`?

alex-ju:
> although it looks like the showcase does rely on this so we should move it to `devDependencies`?

yes, it's more complex that I thought 😞 will keep digging.

alex-ju:
> > although it looks like the showcase does rely on this so we should move it to `devDependencies`?
> 
> yes, it's a bit more complex than I thought 😞 will keep digging.

ok, I followed the logic here – dev dependency would be the way forward – we can't completely get rid of it. added a changelog entry as well

Dhaulagiri (approved):
:shipit:

PR #1238, opened 2023-03-06
Ignore patch releases for Percy

alex-ju:
### :pushpin: Summary

Ignore patch releases for Percy

Further reducing dependabot activity.

Percy is the only dependency where we received automated updates via dependabot. Until we review the overall dependabot activity, we can skip fixes on this dependency and only receive minor and major updates.

[Related Slack thread](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1678113697385509)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,7 @@ updates:
       interval: "weekly"
     allow: 
       - dependency-name: "@percy/*"
+    ignore:
+      # For Percy, ignore all patch updates
+      - dependency-name: "@percy/*"
+        update-types: ["version-update:semver-patch"]


MelSumner (approved):
Is this our way of indicating that all patch releases should not contain any visual changes?

alex-ju:
> Is this our way of indicating that all patch releases should not contain any visual changes?

Not, it's not related to the actual visual regression tests themselves.
Percy is rather active in updating their CLI (often without any impact on consumers) and Brian [suggested](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1678113697385509) we may not need to be so thorough in keeping it up-to-date, at least not more than we do with other dependencies.

PR #1245, opened 2023-03-10
Website / Remove focus from tab panel

didoo:
### :pushpin: Summary

This PR removes the `tabindex` attribute from the tab panels in the website, to avoid them receiving focus and ending in this weird result:

screenshot_2474

(This is a follow-up of a comment in https://github.com/hashicorp/design-system/pull/1167#issuecomment-1442058154)

👉 👉 👉 **Preview**: https://hds-website-git-website-remove-focus-from-tab-panel-hashicorp.vercel.app/components/alert

***

### 👀 Reviewer's checklist:

### DIFF
--- a/website/app/controllers/show.js
+++ b/website/app/controllers/show.js
@@ -118,7 +118,6 @@ export default class ShowController extends Controller {
         const id = section.id;
         const name = section.getAttribute('data-tab');
         section.setAttribute('role', 'tabpanel');
-        section.setAttribute('tabindex', '0');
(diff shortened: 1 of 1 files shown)


MelSumner (approved):
> >  it is recommended that all tabpanel elements in a tab set are focusable if there are any panels in the set that contain content where the first element in the panel is not focusable.

For _most_ of our tabpanels, there _is_ an interactive element as the first item. So we should work on that if we're going to take out the focusability of the tabpanel itself.

PR #1261, opened 2023-03-21
Update Component API documentation pages to standardize the `LinkTo` arguments

didoo:
### :pushpin: Summary

Small PR to standardize the way we declare the `route/models/model/query/current-when/replace` properties (forwarded to the `Hds::Interactive` component) in the Component API documentation for our components.

(reason: I was searching for an example, and I found only one, and I thought it was the only one; actually there are several ones, only they were with different format, and in different order)

Preview: https://hds-website-git-standardize-component-api-inte-44a682-hashicorp.vercel.app/components/link/inline?tab=code#component-api

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


MelSumner (approved):
looks good. I bet we could write a linting rule for this sort of thing. 🤔

PR #1262, opened 2023-03-22
Add new `@listPosition`s for `Dropdown`

alex-ju:
### :pushpin: Summary

Add top positions for dropdown list.

Extend the dropdown `@listPosition` API by adding support to `top-left` and `top-right`.

~~The existing properties are `left` and `right` (default); while there is an opportunity to align values by updating these to `bottom-left` and `bottom-right` it would mean a breaking change for the consumers which may arguably be worth pursuing.~~

Update: we decided to add `bottom-left` and `bottom-right` (default) at this time, with the intention to retire `left` and `right` in a future breaking release, after updating the consumers code (potentially via codemod).

[👉 Preview changes](https://hds-showcase-git-alex-ju-dropdown-extension-4-hashicorp.vercel.app/components/dropdown)

### :link: External links

Jira ticket: [HDS-1684](https://hashicorp.atlassian.net/browse/HDS-1684)
Figma file: https://www.figma.com/file/Kgc8BgB8jGuJXISndgTrD1/Dropdown?node-id=5793-21152&t=iOJ801rNqXMhUrby-4

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable
- [x] Dependant on #1212 (to apply the modifiers to the content, instead of list)

[HDS-1684]: https://hashicorp.atlassian.net/browse/HDS-1684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/addon/components/hds/dropdown/index.js
+++ b/packages/components/addon/components/hds/dropdown/index.js
@@ -8,7 +8,7 @@ import { action } from '@ember/object';
 import { assert } from '@ember/debug';
 
 export const DEFAULT_POSITION = 'right';
-export const POSITIONS = ['right', 'left'];
+export const POSITIONS = ['right', 'left', 'top-left', 'top-right'];
 
 export default class HdsDropdownIndexComponent extends Component {
   /**
--- a/packages/components/addon/components/hds/dropdown/index.js
+++ b/packages/components/addon/components/hds/dropdown/index.js
@@ -8,6 +8,7 @@ import { action } from '@ember/object';
 import { assert } from '@ember/debug';
 
 export const DEFAULT_POSITION = 'bottom-right';
+// TODO: retire 'right' and 'left' in favor of `bottom-right` and `bottom-left` https://github.com/hashicorp/design-system/pull/1262
 export const POSITIONS = [
   'right',
   'left',
--- a/packages/components/addon/components/hds/dropdown/index.js
+++ b/packages/components/addon/components/hds/dropdown/index.js
@@ -7,14 +7,21 @@ import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { assert } from '@ember/debug';
 
-export const DEFAULT_POSITION = 'right';
-export const POSITIONS = ['right', 'left', 'top-left', 'top-right'];
+export const DEFAULT_POSITION = 'bottom-right';
+export const POSITIONS = [
+  'right',
+  'left',
+  'bottom-left',
+  'bottom-right',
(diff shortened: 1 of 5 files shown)


packages/components/app/styles/components/dropdown.scss
```diff
@@ -132,20 +132,36 @@ $hds-dropdown-toggle-border-radius: 5px;
   max-width: initial;
 }
 
+.hds-dropdown-list--position-bottom-right,
 .hds-dropdown-list--position-right {
   position: absolute;
   top: calc(100% + 4px);
   right: 0;
   z-index: 2; // https://github.com/hashicorp/design-system/issues/114
```
didoo, line 140: I think you can remove these comments, are not relevant anymore
```suggestion
  z-index: 2;
```

packages/components/addon/components/hds/dropdown/index.js
```diff
@@ -6,14 +6,21 @@
 import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
 
-export const DEFAULT_POSITION = 'right';
-export const POSITIONS = ['right', 'left'];
+export const DEFAULT_POSITION = 'bottom-right';
+export const POSITIONS = [
+  'right',
```
didoo, line 11: maybe add a comment (and a link to this PR) to explain why we still show these old values

didoo (commented):
added a couple of minor comments, but for me is good to go.
let me know when it's not anymore in draft, so I can approve it

didoo (approved):

didoo (commented):
A terrible doubt (sorry, it just occurred me, see below why): when the dropdown opens above (`top-[...]`) should the chevron on the right point to the top, not the bottom?

Reason being, I realized that the "context switcher" in the sidebar navigation (in production) uses the [caret](https://helios.hashicorp.design/icons/library?searchQuery=caret&selectedIconSize=24) icon
screenshot_2529
but currently the dropdown "toggle" doesn't support this option. What are we going to suggest the consumers to do, in that case?

/cc @heatherlarsen @jorytindall

heatherlarsen:
@didoo What @alex-ju indicated above is what was discussed during the original design phase for the dropdown component. What's important is that there's consistency in the direction of the chevrons to indicate open/close and that it's distinct from the `select`. This is to ensure users have a consistent experience regardless of where the `dropdown` is on the screen.

PR #1270, opened 2023-03-27
NOT FOR MERGE demo: split button

MelSumner:
### :pushpin: Summary

This PR is a simple demo that shows how a split-button could be created with the button component and the dropdown component with a little extra CSS. Some extra work would probably need to be done around the focus rings and maybe active states(?) but this PR is just more of a demo to show potential.

closed state

open state


MelSumner:
closing since we've discussed.

PR #1293, opened 2023-04-07
ember-maybe-in-element 2.1.0

Dhaulagiri:
### :pushpin: Summary

This fixes the noise we were seeing in local website dev related to `broccoli-babel-transpiler is opting out of caching...`.

To verify, run the website locally with and without this branch and compare the difference.

### :link: External links

* [Related issue](https://github.com/DockYard/ember-maybe-in-element/pull/54) via [ember discuss](https://discuss.emberjs.com/t/broccoli-babel-transpiler-is-opting-out-of-caching-due-to-a-plugin-that-does-not-provide-a-caching-strategy/19959)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -9947,7 +9947,7 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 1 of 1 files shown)


MelSumner (approved):
Thank you for fixing this!

PR #1299, opened 2023-04-12
Fix sidecar levels shown in single markdown files 

didoo:
### :pushpin: Summary

This PR fixes an oversight in the logic that handles the headings levels shown in the "sidecar" for single markdown files.

([context](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1681247451576349))

👉 👉 👉 **Preview**: https://hds-website-git-fix-sidecar-levels-hashicorp.vercel.app/getting-started/for-designers

screenshot_2598

***

### 👀 Reviewer's checklist:

### DIFF
--- a/website/app/controllers/show.js
+++ b/website/app/controllers/show.js
@@ -142,7 +142,8 @@ export default class ShowController extends Controller {
       tocs.push({
         index: 0,
         id: 'toc-all',
-        list: getAnchoredHeadings(container),
+        // we show only the headings level 2 and 3 in the sidecar
+        list: getAnchoredHeadings(container).filter((item) => item.depth <= 3),
       });
     }
 


website/app/controllers/show.js
```diff
@@ -35,6 +35,7 @@ const getAnchoredHeadings = (container) => {
         depth: element.tagName.replace(/h/i, ''),
       });
     });
+  console.log(headings);
```
alex-ju, line 38: don't forget to remove this debug line

website/app/controllers/show.js
```diff
@@ -142,7 +143,8 @@ export default class ShowController extends Controller {
       tocs.push({
         index: 0,
         id: 'toc-all',
-        list: getAnchoredHeadings(container),
+        // we show only the headings level 2 and 3 in the sidecar
+        list: getAnchoredHeadings(container).filter((item) => item.depth <= 3),
```
alex-ju, line 147: works for a hotfix; alternatively, we could update the `getAnchoredHeadings` to get a second argument (specifying the header limit or interval) but it would require a bit of refactoring within the function too

alex-ju (approved):

website/app/controllers/show.js
didoo: 🤦‍♂️

website/app/controllers/show.js
didoo: Yeah, you're right. I like your idea. But I've looked at the code, and the `getAnchoredHeadings` does what it says on the tin, is pretty agnostic, so adding logic for filtering the returned array based on a second argument would not be too great. I'll live with the code (filter) duplication for now (hopefully the double comment will make it more visible that both need to be updated)

MelSumner (approved):

heatherlarsen (approved):
Thanks for the quick fix! :shipit:

PR #1303, opened 2023-04-13
Update HOW-TO-TEST-A-COMPONENT-IN-CLOUD-UI.md

didoo:
### :pushpin: Summary

Updated the "How to test a component in Cloud UI" file, to include instructions on how to install a specific branch in GitHub as package.

Thanks @alex-ju for the explanation of how to do it.

***

### 👀 Reviewer's checklist:

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #1308, opened 2023-04-14
Update HOW-TO-TEST-A-COMPONENT-IN-CLOUD-UI.md

didoo:
### :pushpin: Summary

Clarified process to use remote GitBranch (follow-up of #1303)

***

### 👀 Reviewer's checklist:

### DIFF
--- a/packages/components/HOW-TO-TEST-A-COMPONENT-IN-CLOUD-UI.md
+++ b/packages/components/HOW-TO-TEST-A-COMPONENT-IN-CLOUD-UI.md
@@ -74,4 +74,12 @@ You can install the "remote" package directly via `yarn` using the following com
 yarn add '@hashicorp/design-system-components@https://github.com/hashicorp/design-system#head=BRANCH-NAME&workspace=@hashicorp/design-system-components'
 ```
 
+Now, for how things work and are wired up in Cloud UI, this should be done for every engine (there are multiple `package.json` files declaring `@hashicorp/design-system-components` as dependency). So you have to find a way to replace them all at once. One option is to follow these steps:
+
+- run the command above in the roof of the repo; this will add a new entry in the `dependency` block of the `/package.json` file (you will remove it later)
+- copy the line just added (eg. "@hashicorp/design-system-components": "https://github.com/hashicorp/design-system#head=app-sidenav-component&workspace=@hashicorp/design-system-components"`)
+- find & replace in **all** the `package.json` files of the repository the row declaring the `@hashicorp/design-system-components` dependency (eg. `"@hashicorp/design-system-components": "^2.0.0"`) with the line you've just copied, and save all the changes
+- undo the changes in the root `/package.json` file (they were meant only for copying the right string)
+- run `yarn install` to update the `node_modules` and use the "remote" version of the package
+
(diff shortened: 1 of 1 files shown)


alex-ju (approved):
Good spot! Indeed, it needs to be done for each engine (or at least the engines we're testing in) in order to resolve with the branch

packages/components/HOW-TO-TEST-A-COMPONENT-IN-CLOUD-UI.md
```diff
@@ -74,4 +74,12 @@ You can install the "remote" package directly via `yarn` using the following com
 yarn add '@hashicorp/design-system-components@https://github.com/hashicorp/design-system#head=BRANCH-NAME&workspace=@hashicorp/design-system-components'
 ```
 
+Now, for how things work and are wired up in Cloud UI, this should be done for every engine (there are multiple `package.json` files declaring `@hashicorp/design-system-components` as dependency). So you have to find a way to replace them all at once. One option is to follow these steps:
+
+- run the command above in the roof of the repo; this will add a new entry in the `dependency` block of the `/package.json` file (you will remove it later)
```
Dhaulagiri, line 79:
```suggestion
- run the command above in the root of the repo; this will add a new entry in the `dependency` block of the `/package.json` file (you will remove it later)
```

small typo

packages/components/HOW-TO-TEST-A-COMPONENT-IN-CLOUD-UI.md
```diff
@@ -74,4 +74,12 @@ You can install the "remote" package directly via `yarn` using the following com
 yarn add '@hashicorp/design-system-components@https://github.com/hashicorp/design-system#head=BRANCH-NAME&workspace=@hashicorp/design-system-components'
 ```
 
+Now, for how things work and are wired up in Cloud UI, this should be done for every engine (there are multiple `package.json` files declaring `@hashicorp/design-system-components` as dependency). So you have to find a way to replace them all at once. One option is to follow these steps:
+
+- run the command above in the root of the repo; this will add a new entry in the `dependency` block of the `/package.json` file (you will remove it later)
+- copy the line just added (eg. "@hashicorp/design-system-components": "https://github.com/hashicorp/design-system#head=app-sidenav-component&workspace=@hashicorp/design-system-components"`)
+- find & replace in **all** the `package.json` files of the repository the row declaring the `@hashicorp/design-system-components` dependency (eg. `"@hashicorp/design-system-components": "^2.0.0"`) with the line you've just copied, and save all the changes
```
meirish, line 81: There's a shortcut for this in `cloud-ui` - if you run `yarn up @hashicorp/design-system-components@^2.0.0` from the root, it will upgrade all of the packages that already have it installed to that version, and update all of the `package.json` files in one step.

didoo: > Alternatively, you could use the resolutions block in the root package.json file to force all versions of the HDS package to install one version.

I like this one! Let me test it and if it works I'll update the explanation accordingly. Thanks for the tip @meirish ! 🙏

PR #1315, opened 2023-04-21
Add a `isInlineBlock` argument to the `Dropdown` component

didoo:
### :pushpin: Summary

As previously discussed with @alex-ju and @KristinLBradley, we want to add an option for the `Dropdown` (and use the same approach for the `Tooltip`) so that users can achieve specific layout more easily (eg, aligning a dropdown inside a table cell which is right-aligned).

_Notice: this is a code-only change, there's no impact or work that needs to be done on the design side._

In this PR I have:
- updated the `Dropdown` component to support an `@isInlineBlock` argument
- updated the showcase for the `Dropdown` to include a use case for `inline`
- updated the website documentation for the `Dropdown` component and the `Table` component

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/app/styles/components/dropdown.scss
```diff
@@ -14,6 +14,20 @@
 
 $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
 
+// DROPDOWN
+
+// TODO: which one of these?
+// .hds-dropdown--display-inline { // same as flight-icon
+// .hds-dropdown--is-inline-block {
```
alex-ju, line 21: I would probably go with `--is-inline-block` if only to match the component argument, but no strong feelings.

alex-ju (approved):
Looking great! Do want a test to make sure the class is applied when the argument is set?

packages/components/app/styles/components/dropdown.scss
didoo: 👍

PR #1316, opened 2023-04-21
Add `SegmentedGroup` component

alex-ju:
### :pushpin: Summary

Add `SegmentedGroup` component

The `SegmentedGroup` component is essentially a wrapper that can contain a set of pre-defined contextual components (Button, Dropdown, TextInput, Select, and Generic – also referred to as 'segments'). The component respects the order in which these contextual components are invoked, allowing a custom order and any number of elements.

The API of the `SegmentedGroup`
```hbs
<Hds::SegmentedGroup as |S|>
  <S.Button/>
  <S.Dropdown/>
  <S.TextInput/>
  <S.Select/>
  <S.Generic/>
</Hds::SegmentedGroup>
```

We expect this component to be used for filters but also in forms, where a consumer can rely on composition to create a bespoke form field, as shown below and reflected in the Showcase:

```hbs
<Hds::Form::Field @layout="vertical" as |F|>
  <F.Label>This is the label</F.Label>
  <F.HelperText>This is the helper text</F.HelperText>
  <F.Control>
    <Hds::SegmentedGroup as |S|>
      <S.TextInput id={{F.id}} aria-describedby={{F.ariaDescribedBy}} />
      <S.Button @color="secondary" @text="Button" />
    </Hds::SegmentedGroup>
  </F.Control>
  <F.Error>This is the error</F.Error>
</Hds::Form::Field>
```

[👉 Preview showcase](https://hds-showcase-git-alex-ju-segmented-group-hashicorp.vercel.app/components/segmented-group)

**Note**: The documentation (API, How to use, Accessibility) will be added in a separate PR.

### :link: External links

Jira ticket: [HDS-1686](https://hashicorp.atlassian.net/browse/HDS-1686)
Figma file: https://www.figma.com/file/xv7PLLOif0QcHUMTF4jaAi/InputGroup?node-id=30583%3A51146&t=WIEs1rOZYPTEPLAJ-4

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable

[HDS-1686]: https://hashicorp.atlassian.net/browse/HDS-1686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/addon/components/hds/segmented-group/index.hbs
```diff
@@ -0,0 +1,15 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<div class={{this.classNames}} ...attributes>
+  {{yield
+    (hash
+      Button=(component "hds/button" contextualClass="hds-button--segmented")
```
MelSumner, line 8: Are we sure we want to do this? It's not compatible with embroider. (cc @Dhaulagiri since we talked about it earlier today)

Dhaulagiri: Can you say more about what aspect of this is not compatible? What is recommended as an alternative?

MelSumner:
@alex-ju can we add this to the next team agenda to discuss? I'm not sure I understand what this component really does, I'm sure I'm just missing something. I read through the CRD that is linked in the JIRA ticket but I guess I still don't really understand why it needs to be a component.

MelSumner (approved):

alex-ju:
@didoo I tried [applying the `hds-form-field__control` class as a wrapper and have updated the tests accordingly](https://github.com/hashicorp/design-system/pull/1316/commits/2ab072ed88b8101bc19f649342f82d23e8bbdd11). It does seem to introduce [some regression looking at our tests](https://percy.io/HashiCorp/ds-components/builds/26890900?utm_campaign=HashiCorp&utm_content=ds-components&utm_source=github_status_private), and it would also be a breaking change (as we'll apply the class to a wrapper element instead of the control itself).

Regardless of the approach we take for the above class, there are a few other things I made note of following our sync review, that I will address in a follow-up commit (do let me know if you think I've missed something):
 - move the Generic block to the end
 - remove the js file associated with the component (as it doesn't contain any logic)
 - use a single class (thinking of `hds-segmented-group__item` or `hds-segmented-group__segment`) across segments

alex-ju:
@didoo I have [addressed the visual regression](https://github.com/hashicorp/design-system/pull/1316/commits/4867ce7469876036ff5eec4054ed127a90cf27c7) as discussed. However, I needed to use `flex` (could also be `inline-flex`) to remove some computed spacing after inputs.

<table>
<tr><th>Block</th><th>Flex</th></tr>
<tr><td>

Screenshot 2023-04-26 at 15 54 51


</td><td>

Screenshot 2023-04-26 at 15 54 43


</td></tr>
</table>

Have also addressed the other points mentioned above.

packages/components/tests/integration/components/hds/segmented-group/index-test.js
```diff
@@ -0,0 +1,38 @@
...
+  test('it should render the component with a CSS class that matches the component name', async function (assert) {
+    await render(hbs`<Hds::SegmentedGroup id="test-segmented-group" />`);
+    assert.dom('#test-segmented-group').hasClass('hds-segmented-group');
+  });
+
+  // CONTEXTUAL COMPONENTS
+
+  test('it renders the contextual components with CSS modifier classes', async function (assert) {
+    await render(
+      hbs`<Hds::SegmentedGroup as |S|>
```
didoo, line 23: I think the "generic" container is missing

didoo (commented):
First part of the code review.

packages/components/tests/dummy/app/templates/components/segmented-group.hbs
didoo, line 1: Probably I would organize the content in this way, to follow how other components are organized:

- Content
  - Two segments
    - Button/Trailing + Button/Leading + Dropdown/Trailing + Dropdown/Leading
    - With different input types
  - Multiple segments
  - With generic content
    - trailing + leading + trailing/leading
  - Inside a form field
- States
  - trailing
  - leading

packages/components/app/styles/components/form/field.scss
```diff
@@ -32,6 +32,9 @@
   }
 
   .hds-form-field__control {
+    display: flex; // to avoid space between elements
```
didoo, line 35:
```suggestion
    display: flex; // to avoid extra whitespace if the content is an <input> or <textarea> (they're inline elements)
```

packages/components/app/styles/components/form/field.scss
```diff
@@ -70,6 +73,7 @@
...
```
didoo, line 76:
```suggestion
    display: flex; // to avoid extra whitespace if the content is an <input> or <textarea> (they're inline elements)
```

didoo (approved):
Added a few suggestions, but apart from that the code is **great**!

packages/components/tests/dummy/app/templates/components/segmented-group.hbs
alex-ju: Updated as suggested ✨

alex-ju:
After merging this work in, we found an edge-case regression in Firefox, on `Form::Select::Field`  when used in a constrained container (regardless of the container layout).

<table>
<tr><td>

contrained-select-firefox-before


</td><td>

contrained-select-firefox-after


</td></tr>
</table>

We identified this issue as caused by a difference in how the select implicit styles are set on Firefox (including `overflow`, and `white-space`). Potential solutions to prevent this regression from happening are to set `overflow: auto` or `display: grid` on `hds-form-field__control`, both with potential unexpected consequences.

After weighing the pros and cons we decided not to address this regression, but we're documenting it here in case we need to reconsider the decision.

PR #1304, opened 2023-04-13
`SideNav` component - Porting of advanced features from `HcNav`

didoo:
### :pushpin: Summary

This is a large PR that takes the existing `Hds::SideNav` implementation and extends it to include advanced features like responsiveness (animation/transition) and content "portaling", by adopting (and adapting) the implementation in Cloud UI (then extracted and refactored by @meirish as reusable/standalone component in https://github.com/hashicorp/ember-shared-components).

### :hammer_and_wrench: Detailed description

Unfortunately, it's not been possible to break down the PR in smaller commits: some of the work has been done in other PRs (see related PRs below) and ported in this PR all together, other work involved large renaming/refactoring of the main component (`SideNav`) as well as its children/sub-components. I'll try to give an overview of what the code changes are, and why they've been done.

- added new dependencies
  - [`ember-a11y-refocus`](https://github.com/ember-a11y/ember-a11y-refocus) - used to add a navigator narrator to the `SideNav` (it's optionally controlled by the consumers)
  - [`ember-stargate`](https://github.com/simonihmig/ember-stargate) - used for content "portaling" (see the Cloud UI implementation to understand how this works
- updated `SideNav` component porting advanced features from `HcNav` (responsiveness/animation/portaling)
  - `SideNav`
    - "merged" the layout provided by the `SideNav::Wrapper` container, with the high-level functionalities provided by the `HcNav` implementation:
      - accessibility improvements
        - added support for `a11y-refocus` `<NavigatorNarrator>` element
          - it's added by default but can be disabled via `hasA11yRefocus` argument
          - we expose a set of arguments to customise its behaviour
        - added focus trapping for the navigation (when is "responsive", is in "mobile" mode and is "expanded")
      - responsiveness
        - made the "responsive" behaviour optional (`true` by default)
          - it adds a button that toggles the minimized/expanded states when in "mobile" viewport (the button itself changes visual state)
        - exposed the `header/body/footer` containers of the "sidenav" as named blocks and exposed the `isMinimized` tracked property so it can be used for custom behaviours if necessary
        - the consumers can "override" some of the default properties (eg. breaking point between "mobile" and "desktop" views) using CSS custom props (eg. `--hds-app-desktop-breakpoint`)
        - when in "mobile" + "expanded" mode
          - an overlay is shown (if clicked it minimizes the sidenav)
          - a keyboard event listener for `esc` is added (if triggered it minimizes the sidenav)
[...]

PR #1321, opened 2023-04-24
`SideNav` - Extract a "base" component

didoo:
### :pushpin: Summary

This is a follow-up of https://github.com/hashicorp/design-system/pull/1304 in which I have extracted the "basic" layout functionality of the `SideNav` in a dedicated component, so that it can be used independently from the full-fledged one, to achieve basic navigations (in which only the layout wrapper is needed, not all the responsiveness/animation etc).

_The idea for this change has materialised when I was starting to write the documentation for "how to use" the sidebar navigation, and I've realised that in cases where a consumer would need just a basic navigation, the implemented `SideNav` could be an overkill (especially for the fact that there's a lot of JavaScript involved, that in the case of a basic navigation may not be necessary at all). So I decided to do test, see how the code would look like, and at the end of the process I saw that it wasn't bad at all, on the contrary: made sense. So I decided to open the PR and gather feedback from the other HDS engineers._

In this PR I have:
- added a `SideNav::Base` component (copy&pasted the existing `SideNav` and stripped away all the non-basic-layout stuff)
- updated the `SideNav` to use (internally) the `SideNav::Base` subcomponent
- added showcase example for `SideNav::Base` + moved `SideNav` examples in the showcase up in the page
- updated the “Component API” section of the website
- added integration tests for `SideNav::Base`

### :link: External links

- https://github.com/hashicorp/design-system/pull/1304

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable

### DIFF
--- a/website/docs/components/side-nav/partials/code/component-api.md
+++ b/website/docs/components/side-nav/partials/code/component-api.md
@@ -25,7 +25,7 @@ This is the basic component (layout only).
 
 <Doc::ComponentApi as |C|>
   <C.Property @name="<:root>" @type="named block">
-    A named block that can be used to render content that needs to be rendered outside of the "header/body/footer" containers of the SideNav.
+    A named block for rendering content outside of the "header/body/footer" containers of the SideNav.
   </C.Property>
   <C.Property @name="<:header>" @type="named block">
     A named block where the content for the “header” area of the SideNav is rendered. The `SideNav::Header` component should be added here.
--- a/website/docs/components/side-nav/partials/code/component-api.md
+++ b/website/docs/components/side-nav/partials/code/component-api.md
@@ -2,7 +2,31 @@
 
 ### SideNav
 
+This is the full-fledged component (responsive and animated).
+
 <Doc::ComponentApi as |C|>
+  <C.Property @name="<:header>" @type="named block">
+    A named block where the content for the “header” area of the SideNav is rendered. The `SideNav::Header` component should be added here. It yields the value of `isMinimized` too.
+  </C.Property>
+  <C.Property @name="<:body>" @type="named block">
+    A named block where the content for the “body” or main content of the SideNav is rendered. The `SideNav::List` and `SideNav::PortalTarget` components should be added here when used. It yields the value of `isMinimized` too.
+  </C.Property>
+  <C.Property @name="<:footer>" @type="named block">
+    A named block where the content for the “footer” section of the SideNav is rendered. It yields the value of `isMinimized` too.
+  </C.Property>
+  <C.Property @name="...attributes">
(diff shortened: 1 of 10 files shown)


website/docs/components/side-nav/partials/code/component-api.md
```diff
@@ -2,7 +2,31 @@
...
+  </C.Property>
+</Doc::ComponentApi>
+
+### SideNav::Base
+
+This is the basic component (layout only).
+
 <Doc::ComponentApi as |C|>
+  <C.Property @name="<:root>" @type="named block">
+    A named block where content that lives in the "root" of the SideNav is rendered (the content should not cause a layout height).
```
KristinLBradley, line 28: "the content should not cause a layout height" could you please clarify this part? Does it mean the content shouldn't include a fixed height or something else?

KristinLBradley (commented):
Looks good. I just added a comment on one part that seemed unclear to me.

website/docs/components/side-nav/partials/code/component-api.md
didoo: Yeah, you're right @KristinLBradley, it's not clear (it's hard to explain). Let me have a new go at tomorrow, with a fresh mind, and see how to improve the description.

didoo: @KristinLBradley I've made a change that removes the issue at the "root" (pun intended) :)

website/docs/components/side-nav/partials/code/component-api.md
```diff
@@ -2,7 +2,31 @@
...
+  </C.Property>
+</Doc::ComponentApi>
+
+### SideNav::Base
+
+This is the basic component (layout only).
+
 <Doc::ComponentApi as |C|>
+  <C.Property @name="<:root>" @type="named block">
+    A named block that can be used to render content that needs to be rendered outside of the "header/body/footer" containers of the SideNav.
```
alex-ju, line 28: suggestion – no strong opinion
```suggestion
    A named block for rendering content outside of the "header/body/footer" containers of the SideNav.
```

alex-ju (approved):
Looks great, could spot anything off!

Not part of this change, but just a note that in the 'yielded content' example we have multiple `<nav>` items, whereas in the portal we have a single element. For parity, we could have a single `Hds::SideNav::List` for the 'yielded content' (if only to show there's no difference in the end result).

didoo:
> Looks great, could spot anything off!
👍

> Not part of this change, but just a note that in the 'yielded content' example we have multiple `<nav>` items, whereas in the portal we have a single element. For parity, we could have a single `Hds::SideNav::List` for the 'yielded content' (if only to show there's no difference in the end result).
@alex-ju it was intentional, to test what happens when/if a consumers uses two `List`s in the `<:body>`, to see how the SideNav component reacts (expecially in terms of vertical spacing).

PR #1344, opened 2023-05-04
Sidenav fix flickering 3

cbfx:
### :pushpin: Summary

 If merged, this PR will remove the necessity of debouncing the animation and remove a flicker if a panel has already been rendered as the next panel to be animated.

### :link: External links

Figma file: [if it applies]

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/side-nav/portal/target.js
+++ b/packages/components/addon/components/hds/side-nav/portal/target.js
@@ -9,12 +9,12 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { DEBUG } from '@glimmer/env';
 import Ember from 'ember';
-import { debounce } from '@ember/runloop';
 
 export default class SidenavPortalTarget extends Component {
   @service router;
 
   @tracked numSubnavs = 0;
+  @tracked lastPanelEl = null;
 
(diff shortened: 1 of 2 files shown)


packages/components/addon/components/hds/side-nav/portal/target.js
```diff
@@ -38,7 +39,7 @@ export default class SidenavPortalTarget extends Component {
 
   @action
   didUpdateSubnav(element, [count]) {
-    debounce(this, 'animateSubnav', element, [count], 100);
+    debounce(this, 'animateSubnav', element, [count], 0);
```
didoo, line 42: can you add a comment on why we are leaving the `debounce`, but with a value of `0`?

didoo (approved):
@cbfx all good for me, left only a minor comment. 

can you let me know when it's merged, so I can do a new release and you can bump the package version in your PR/branch?

packages/components/addon/components/hds/side-nav/portal/target.js
cbfx: @didoo i left it out of caution incase we wanted to provide a way to configure that value. i updated my PR to just simply remove the debounce.

PR #1355, opened 2023-05-12
Move utility functions from `form/utils` to `addon/helpers`

alex-ju:
### :pushpin: Summary

Move the following utility functions  from `form/utils` to `addon/helpers` as recommended in [HDS-942](https://hashicorp.atlassian.net/browse/HDS-942)

`getElementId`
 - a utility getter that returns `this.args.id` if it exists, otherwise generates a new id using Ember's `guidFor`
 - used by `dropdown/list-item/[checkbox|radio]`, `form/[field|fieldset]`, `flyout` and `modal`
`setAriaDescribedBy`
 - a utility setter that appends `this.args.extraAriaDescribedBy` to `this.descriptors`
 - used by `form/[field|fieldset|radio-card]`

**Note**
While reviewing our usage of `guidFor` we identified 2 patterns in our codebase:
 - we use Ember's `guidFor` directly when the `id` is always generated (in `pagination`, `tabs`)
 - we use `getElementId` when we want to preserve the passed-in value for `id` argument (in `form` components, `flyout`, `modal`)

I believe there is not much value in trying to converge to using a single pattern as they address different scenarios, but open to having a conversation around this.

### :link: External links

Jira ticket: [HDS-942](https://hashicorp.atlassian.net/browse/HDS-942)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-942]: https://hashicorp.atlassian.net/browse/HDS-942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/addon/components/hds/dropdown/list-item/checkbox.js
```diff
@@ -4,7 +4,7 @@
  */
 
 import Component from '@glimmer/component';
-import { getElementId } from '../../form/utils/getElementId';
+import { getElementId } from '../../../../helpers/hds-get-element-id';
```
Dhaulagiri, line 7:
```suggestion
import { getElementId } from '@hashicorp/design-system-components/helpers/hds-get-element-id';
```
it's slightly more typing, but my experience is that this other syntax utilizing ember's resolver is a little more digestible/maintaible than trying to figure out how many times you have to traverse up the file tree to find this file

alex-ju: I was looking for a quicker/easier to read resolver on this, but couldn't figure it out

didoo (approved):
All good for me.

The only concern I have is that - from what I understand - these are not Ember "helpers", _technically_  speaking, just javascript helper functions:

> Helper functions are JavaScript functions that you can call from your template.

https://guides.emberjs.com/release/components/helper-functions/

but they live in the `addons/helpers` folder, so one could be lead to think they could be used as classic "Ember helpers" in the HBS templates, while in reality they're imported in the backing classes.

That said, I don't have an alternative solution (a "shared" folder?). Maybe @MelSumner can suggest a more idiomatic way where to save these files?

alex-ju:
> All good for me.
> 
> The only concern I have is that - from what I understand - these are not Ember "helpers", _technically_ speaking, just javascript helper functions:
> 
> > Helper functions are JavaScript functions that you can call from your template.
> 
> https://guides.emberjs.com/release/components/helper-functions/
> 
> but they live in the `addons/helpers` folder, so one could be lead to think they could be used as classic "Ember helpers" in the HBS templates, while in reality they're imported in the backing classes.
> 
> That said, I don't have an alternative solution (a "shared" folder?). Maybe @MelSumner can suggest a more idiomatic way where to save these files?

I share the concern here – I'd be keen to find them a place that avoids this confusion.

MelSumner:
We could also consider https://api.emberjs.com/ember/4.12/classes/Ember.Templates.helpers/methods/unique-id?anchor=unique-id and document that users should use it. 

One thing we'd have to probably check is "how does this work in Fastboot" because IIRC that hasn't been fully tested yet.

alex-ju:
I did a bit of digging into the Ember world, and the idiomatic way to share functions seems to be via a folder named `utils` – 
there's even a generator: `ember g util my-util` – hence I moved them there, to avoid confusion with actual `helper` functions. Let me know if that's ok with you all.

didoo:
> I did a bit of digging into the Ember world, and the idiomatic way to share functions seems to be via a folder named `utils` – there's even a generator: `ember g util my-util` – hence I moved them there, to avoid confusion with actual `helper` functions. Let me know if that's ok with you all.

@alex-ju works for me! 👍

PR #1242, opened 2023-03-09
Revert ember-keyboard & ember-truth-helpers bumps

Dhaulagiri:
### :pushpin: Summary

At least for now, this relaxes a couple of updates we did that were impacting the ability of certain consumers from installing 1.7.x. We'll want to figure out a more sustainable longer term plan for this, but since we didn't _need_ these, relaxing and allowing consumers to install previous versions seems like the pragmatic choice.

### :link: External links

Jira ticket: [HDS-1671](https://hashicorp.atlassian.net/browse/HDS-1671)

***

### 👀 Reviewer's checklist:

- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-1671]: https://hashicorp.atlassian.net/browse/HDS-1671?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1360, opened 2023-05-17
Upgrade forward-compatible packages

alex-ju:
### :pushpin: Summary

Bump patch and minor versions of our dependencies using `yarn upgrade-interactive`

Did this while writing https://go.hashi.co/rfc/ds-050

### :link: External links

Jira ticket: [HDS-1925](https://hashicorp.atlassian.net/browse/HDS-1925)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-1925]: https://hashicorp.atlassian.net/browse/HDS-1925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


didoo (commented):
I remember we had issues (with some consumers) when we changed the version of a package to the latest, and they were on an older version in their codebase. Could this be the case again, bumping all our dependencies to the latest?

didoo:
another question: should we have a changelog in this case?

packages/components/package.json
```diff
@@ -42,29 +42,29 @@
...
+    "ember-auto-import": "^2.6.3",
     "ember-cached-decorator-polyfill": "^0.1.4",
     "ember-cli-babel": "^7.26.11",
     "ember-cli-htmlbars": "^6.2.0",
     "ember-cli-sass": "^10.0.1",
     "ember-composable-helpers": "^4.5.0",
-    "ember-focus-trap": "^1.0.1",
-    "ember-keyboard": "^8.1.0",
+    "ember-focus-trap": "^1.0.2",
+    "ember-keyboard": "^8.2.0",
```
alex-ju, line 52: Needs to be pinned to `8.1.0` due to [conflicting dependencies in Terraform](https://github.com/hashicorp/design-system/pull/1242).

MelSumner: Question: aren't minor versions supposed to be compatible with each other? Just curious.

alex-ju: In theory, yes. The trouble, from the threads that I've read, is that `ember-keyboard@8.2.0` bumps `ember-modifier` from `3.x` to `4.x`, somehow/sometimes imposing this latest version of `ember-modifier` to our consumers. I'm still checking facts around this, but that's the gist as I get it.

PR #1361, opened 2023-05-18
HDS-960: Add support for aria-label customization

MelSumner:
### :pushpin: Summary

If merged, this PR resolves issue HDS-960 by adding aria-label localization support. 

### :hammer_and_wrench: Detailed description

- changed `aria-label` value from "breadcrumbs" to `{{{this.ariaLabel}}` which returns the custom text if defined, otherwise falls back to "breadcrumbs" 
- updated tests

### :camera_flash: Screenshots

CleanShot 2023-05-18 at 10 29 01

### :link: External links

Jira ticket: [HDS-960](https://hashicorp.atlassian.net/browse/HDS-960)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable
- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-960]: https://hashicorp.atlassian.net/browse/HDS-960?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1366, opened 2023-05-19
Website `Component API` properties - Allow to show a "default" value even when `@values` is not provided

didoo:
### :pushpin: Summary

While reviewing @MelSumner PR #1361 I noticed something that found missing before: the possibility to show what the default value of an argument is, even when there is no `@value` argument.

In this PR I am proposing to use the existing `@default` argument to achieve this result.

The actual changes to the documentation will be done in a follow-up PR: https://github.com/hashicorp/design-system/pull/1369

In this PR I have:

- allowed the “Component API” `property` block to show a `@default` value even if a list of `@values` is not provided

screenshot_2700

***

### 👀 Reviewer's checklist:

### DIFF
--- a/website/app/components/doc/component-api/property.hbs
+++ b/website/app/components/doc/component-api/property.hbs
@@ -35,6 +35,16 @@
         {{@valueNote}}
       {{/if}}
     </dd>
+  {{else if @default}}
+    <dt class="doc-component-api__term doc-component-api__term--values">Default</dt>
+    <dd class="doc-component-api__description doc-component-api__description--values">
+      <ul class="doc-component-api__property-values" role="list">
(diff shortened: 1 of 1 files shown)


didoo:
@MelSumner regarding the HTML/DOM structure, I have simply duplicated the existing one, only in this case we have a list with a single item. If you think is too ugly or not accessible, let me know (or even better suggest an alternative structure) and I'll fix it.

MelSumner (approved):
I've been wanting this feature for a while now, thanks for adding it! Looks like there's some prettier complaints to fix but I don't see any glaring issues at this time. I look forward to trying this out on some of our other component APIs where I think it will prove useful.

alex-ju (approved):

PR #1373, opened 2023-05-23
Adds args with string fallback to aria-labels in components

MelSumner:
### :pushpin: Summary

If merged, this PR resolves instances where a bare string was being set as the value for the `aria-label` attribute. It adds support for an arg with a string fallback. 

Why? 

First and foremost, this is nothing new. We're already doing this, just inconsistently. This PR fixes that. 

While we could just make sure each component supports `...attributes` and leave it up to the consumer to remember to use `aria-label`, explicitly adding it to the appropriate components will have a few benefits: 

1. as a side effect of "because it's a supported arg, it's in the API docs" may help developers more easily remember to add it
2. the fallback ensures that it's there even if they forget, but can be localized if they remember and want to do that (I think also some products will have some level of linting enabled that should yell at them for [no-bare-strings]())
3. there are some places where it's (by necessity) in a composed component but not one exposed to the consumer directly. By ensuring that it can be set every time by consumers, it's more convenient to remember because it's consistent across the board (and consistent with how we've tried to approach components in general)

You may also notice that there are a couple of instances where a component-class was added to support this, instead of just using a conditional statements. While we do use conditional statements in some cases, our eng team felt as though it would be easier to search for `get ariaLabel` and get all of the results, and would be easier than trying to remember that it was also potentially a conditional statement. 

Components changed: 
- breadcrumb/truncation
- pagination/compact
- pagination/numbered
- side-nav/index
- dismiss-button
- tag

Component API files updated: 
- breadcrumb
- pagination
- side-nav
- tag
- utils/dismiss-button

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/clean-frogs-dance.md
+++ b/.changeset/clean-frogs-dance.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Updates aria-label support for consistency
--- a/.changeset/clean-frogs-dance.md
+++ b/.changeset/clean-frogs-dance.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-Updates aria-label support for consistency
+Updates aria-label support for consistency. Consumers can now see in the component API docs where `@ariaLabel` is supported for a custom value, and what the fallback value is.
--- a/website/docs/components/breadcrumb/partials/code/component-api.md
+++ b/website/docs/components/breadcrumb/partials/code/component-api.md
@@ -46,6 +46,9 @@ The Breadcrumb component is composed of three different parts, each with their o
 ### Breadcrumb::Truncation
 
 <Doc::ComponentApi as |C|>
+  <C.Property @name="ariaLabel" @type="string">
+    Set on the truncation toggle button. Accepts a localized string; the fallback is set to `show more`.
+  </C.Property>
   <C.Property @name="yield">
     Elements passed as children of this child component are yielded to the content of the [Disclosure](/utilities/disclosure) component (used to show/hide the yielded Breadcrumb Items via a "toggle" button).
   </C.Property>
--- a/website/docs/components/breadcrumb/partials/code/component-api.md
+++ b/website/docs/components/breadcrumb/partials/code/component-api.md
@@ -46,8 +46,8 @@ The Breadcrumb component is composed of three different parts, each with their o
 ### Breadcrumb::Truncation
 
 <Doc::ComponentApi as |C|>
-  <C.Property @name="ariaLabel" @type="string">
-    Set on the truncation toggle button. Accepts a localized string; the fallback is set to `show more`.
+  <C.Property @name="ariaLabel" @type="string" @default="show more">
+    Set on the truncation toggle button. Accepts a localized string.
   </C.Property>
   <C.Property @name="yield">
     Elements passed as children of this child component are yielded to the content of the [Disclosure](/utilities/disclosure) component (used to show/hide the yielded Breadcrumb Items via a "toggle" button).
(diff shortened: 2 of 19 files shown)


website/docs/components/breadcrumb/partials/code/component-api.md
```diff
@@ -46,6 +46,9 @@ The Breadcrumb component is composed of three different parts, each with their o
 ### Breadcrumb::Truncation
 
 <Doc::ComponentApi as |C|>
+  <C.Property @name="ariaLabel" @type="string">
+    Set on the truncation toggle button. Accepts a localized string; the fallback is set to `show more`.
+  </C.Property>
```
didoo, line 51:
```suggestion
  <C.Property @name="ariaLabel" @type="string" @default="show more">
    Set on the truncation toggle button. Accepts a localized string.
  </C.Property>
```

website/docs/components/pagination/partials/code/component-api.md
```diff
@@ -22,6 +22,9 @@ These pagination sub-elements may be used directly if you need to cover a very s
 ### Pagination::Numbered
 
 <Doc::ComponentApi as |C|>
+<C.Property @name="ariaLabel" @type="string">
+    Accepts a localized string; the fallback is set to `Pagination navigation`.
+</C.Property>
```
didoo, line 27:
```suggestion
<C.Property @name="ariaLabel" @type="string" @default="Pagination navigation">
    Accepts a localized string.
</C.Property>
```

website/docs/components/pagination/partials/code/component-api.md
```diff
@@ -63,6 +66,9 @@ This component supports use of [`...attributes`](https://guides.emberjs.com/rele
 ### Pagination::Compact
 
 <Doc::ComponentApi as |C|>
+<C.Property @name="ariaLabel" @type="string">
+    Accepts a localized string; the fallback is set to `Pagination navigation`.
+</C.Property>
```
didoo, line 71:
```suggestion
<C.Property @name="ariaLabel" @type="string" @default="Pagination navigation">
    Accepts a localized string.
</C.Property>
```

website/docs/utilities/dismiss-button/partials/code/component-api.md
```diff
@@ -1,6 +1,9 @@
 ## Component API
 
 <Doc::ComponentApi as |C|>
+  <C.Property @name="ariaLabel" @type="string">
+    Accepts a localized string; the fallback is set to `dismiss`.
+  </C.Property>
```
didoo, line 6:
```suggestion
  <C.Property @name="ariaLabel" @type="string" @default="dismiss">
    Accepts a localized string.
  </C.Property>
```

didoo (approved):
LGTM 👍

I've left a few suggestions to use the recently updated `@default` field.

.changeset/clean-frogs-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Updates aria-label support for consistency
```
Dhaulagiri, line 5: It seems like it would be useful to be a little more explicit and detailed here so consumers of the changelog understand they can now use `@ariaLabel` in more places than before?

MelSumner, line 5:
```suggestion
Updates aria-label support for consistency. Consumers can now see in the component API docs where `@ariaLabel` is supported for a custom value, and what the fallback value is.
```

How about this, @Dhaulagiri ?

Dhaulagiri: 🚢 looks good!

PR #1377, opened 2023-05-24
Add `PageHeader` Component

jorytindall:
### :pushpin: Summary

If merged this PR adds the `PageHeader` component to Helios as a collaboration/contribution from the HCP design team and multiple stakeholders across product lines.

I added a variety of different use cases and edge cases to the showcase as a means to capture specific questions and "gotchas" with this component. Once approved, it probably makes sense to remove some of them, though I think there is value in including some of the contextual examples since this tows the line between a component and pattern. 

### 🔧  Details

There are several areas that I would like specific feedback on including:

* Potential issues with the Dropdown `listPosition` and lack of collision awareness
* Button order and alignment at smaller viewport sizes
* Responsive strategy
  * Using container queries; are able to use these yet, does the support exist?.
  * I've used `flex-wrap` with some of the inline elements, making the point at which they wrap variable. Is this something we want more control over?
  * Is there a more elegant way to handle the breakpoints without using static values? Considering the method in the side-nav to conditionally apply classes with JavaScript, is that a recommended method here? Should the classes be exposed to be overwritten?
  * The linter does not recognize container queries yet, is that an adjustment that needs to be made?
* I've written a few simple tests for this component, are more necessary? If so, I believe I can add them without too much trouble given that this is largely a presentational component.

### :link: External links

Jira ticket: [HDS-XXX](https://hashicorp.atlassian.net/browse/HDS-1892)
Figma file: [PageHeader](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/Urm6MreKfPoJMMW1sWTR08/HDS-Product---Components?type=design&node-id=34664%3A70976&t=isQ1cpOW4GbGpAmh-1)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/addon/components/hds/page-header/index.hbs
```diff
@@ -0,0 +1,30 @@
+{{!
+Copyright (c) HashiCorp, Inc.
+SPDX-License-Identifier: MPL-2.0
+}}
+<div class="hds-page-header" ...attributes>
```
MelSumner, line 5: is it possible to have the header be the outermost containing element?

We want to try to have all content contained in landmark elements; users with a screen reader can browse a page via landmarks, and a bookmark component outside of the header (one of the landmarks) is not ideal.

jorytindall: The reason the header isn't the most outermost element is because it often wraps breadcrumbs which render a `nav` element. The intention was to not nest a landmark nav inside of a `header`.

packages/components/addon/components/hds/page-header/index.hbs
```diff
@@ -0,0 +1,30 @@
...
+  {{yield (hash Breadcrumb=(component "hds/yield"))}}
+  <header class="hds-page-header__wrapper">
+    {{yield (hash IconTile=(component "hds/icon-tile" size="medium"))}}
+    <div class="hds-page-header__main">
+      <div class="hds-page-header__content">
+        <div class="hds-page-header__heading">
```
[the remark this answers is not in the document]
jorytindall: I believe this was the only way to account for the badge coupled with the title/heading, while also justifying the actions to the end of the container. It's a little bit div heavy though, I'll admit.

KristinLBradley:
RE: Container queries

If we go by the [browser versions that Consul supports](https://github.com/hashicorp/consul/blob/main/ui/packages/consul-ui/config/targets.js), then it appears we can't use them. See: https://caniuse.com/css-container-queries

You could potentially ask if they would be willing to update their list of supported browsers to more recent ones though...

MelSumner:
DOM order needs to match visual order; it's a known issue with flex-direction: reverse-row. I can add some extra links to this CRD so our team can read through. 👍

packages/components/addon/components/hds/page-header/index.hbs
alex-ju: This was my recommendation, Melanie, to avoid nesting them, but if you think nesting the two landmarks is not a concern we can revert and apply the header element at the top level of this component.

.changeset/light-seas-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+Adds PageHeader component
```
alex-ju, line 5:
```suggestion
Add `PageHeader` component
```

alex-ju (commented):
> * Potential issues with the Dropdown `listPosition` and lack of collision awareness

While I see the issue here, I think this should be a concern of our Dropdown component, and something we can consider as a potential enhancement.

> * Button order and alignment at smaller viewport sizes

To me, having the same order makes sense regardless of their positioning. `flex-direction` it's certainly a no, as it will fail WCAG 1.3.2 Meaning sequence.

> * Responsive strategy
>   * Using container queries; are able to use these yet, does the support exist?

Kristin made a good point about checking in browser support on products. Although I suggest we see the usage of container queries in this component as a progressive enhancement – the layout doesn't break if the feature is not supported, it just looks slightly differently on narrow viewports.

>   * I've used `flex-wrap` with some of the inline elements, making the point at which they wrap variable. Is this something we want more control over?

Good question – curious to see others' views on this

>   * Is there a more elegant way to handle the breakpoints without using static values? Considering the method in the side-nav to conditionally apply classes with JavaScript, is that a recommended method here? Should the classes be exposed to be overwritten?

I _think_ that in the case of SideNav the applied classes were a way to instrument the component from outside, not necessarily a pattern that we need to replicate here.

>   * The linter does not recognize container queries yet, is that an adjustment that needs to be made?

Doesn't recognize them, but doesn't complaining either, so I say it's a non-issue.

> * I've written a few simple tests for this component, are more necessary? If so, I believe I can add them without too much trouble given that this is largely a presentational component.

Yes, I suggest we test the assertion (the error we throw when no title is provided), you'll be able to find some similar examples in other components.

packages/components/addon/components/hds/page-header/index.hbs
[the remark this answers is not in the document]
alex-ju: This was my recommendation, Melanie, to avoid nesting them, but if you think nesting the two landmarks is not a concern we can revert and apply the header element at the top level of this component.

heatherlarsen:
> While I see the issue here, I think this should be a concern of our Dropdown component, and something we can consider as a potential enhancement.

@alex-ju I created [HDS-1995](https://hashicorp.atlassian.net/browse/HDS-1995) so we don't lose track of this work.

heatherlarsen (approved):
I know you didn't ask for a designer review on this, but I figured it could use some extra eyes since you're the original designer. This looks really solid and aligns with the Figma component well! :shipit:

jorytindall:
> I know you didn't ask for a designer review on this, but I figured it could use some extra eyes since you're the original designer. This looks really solid and aligns with the Figma component well! :shipit:

Probably begs a question about who the approvers are in this process. I guess I'm my own stakeholder 😂

didoo (commented):
Done a first pass and added a few comments (some of them may be blockers). I'll do a second review after the first round of reviews (I imagine there will be some tweaks to do)

Re. your request for feedback: 
- Potential issues with the Dropdown listPosition and lack of collision awareness
  - I agree with @alex-ju it's a separate issue, to be solved in another effort
- Using container queries
  - I agree with @alex-ju also on this: we should consider this as progressive enhancement, so I would not worry too much about it (in this very specific case, which is mostly presentational)
  - Is there a more elegant way to handle the breakpoints without using static values?
    - Not that I know/can think of
- Considering the method in the side-nav to conditionally apply classes with JavaScript, is that a recommended method here?
  - Nope :)
- The linter does not recognize container queries yet, is that an adjustment that needs to be made?
  -   Yes, but should be done in a separate ticket (it's about the tool we're using)

packages/components/addon/components/hds/page-header/index.hbs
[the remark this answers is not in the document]
MelSumner: It's acceptable to nest landmark elements directly 👍

packages/components/addon/components/hds/page-header/metadata.hbs
```diff
@@ -0,0 +1,13 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<div class="hds-page-header__metadata" ...attributes>
+  {{yield
+    (hash
+      Subtitle=(component "hds/pagetitle/subtitle")
```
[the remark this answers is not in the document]
jorytindall: Yes, this was initially intended to be a wrapper around the the metadata with other nested contextual components, but did not end up using that strategy. Good catch!

packages/components/app/styles/components/page-header.scss
```diff
@@ -0,0 +1,99 @@
...
+}
+
+.hds-page-header__metadata {
+  display: flex;
+  flex-direction: column;
+  gap: 0.25rem;
+
+  &:empty {
+    display: none;
+  }
```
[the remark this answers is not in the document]
jorytindall: Ah, we had originally tried this to account for not rendering an empty `div` when no metadata is yielded, but alas it doesn't work. I'll remove it.

packages/components/app/styles/components/page-header.scss
```diff
@@ -0,0 +1,99 @@
...
+  flex-direction: row;
+  flex-wrap: wrap;
+  gap: 1rem;
+  align-items: center;
+}
+
+.hds-page-header__header-content {
+  color: var(--token-color-foreground-strong);
+}
+
```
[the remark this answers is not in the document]
jorytindall: Ok, I've reworked these a bit to hopefully be a bit more semantic. Definitely open to suggestions if you have a recommendation, but there are less repetitive instances of "header/heading"

packages/components/addon/components/hds/page-header/index.hbs
```diff
@@ -0,0 +1,30 @@
...
+          <h1 class="hds-page-header__heading-content hds-typography-display-500">
+            {{this.title}}
+          </h1>
+          {{yield (hash Badges=(component "hds/page-header/badges"))}}
+        </div>
+        <div class="hds-page-header__metadata">
+          {{yield
+            (hash
+              Subtitle=(component "hds/page-header/subtitle")
```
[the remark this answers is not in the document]
jorytindall: I was considering that, and upon further testing it _seems_ like it would make sense to use the same strategy as the rest of the yielded content and not pass the title as an argument. This is also the method used in the `Modal` and `Flyout`, and will make adoption easier. I'll make some adjustments and update the hierarchy.

packages/components/app/styles/components/page-header.scss
```diff
@@ -0,0 +1,79 @@
...
+  flex-direction: column;
+  gap: 1rem;
+  container-type: inline-size;
+}
+
+.hds-page-header__body {
+  display: flex;
+  flex-direction: column;
+  gap: 0.5rem 1rem;
+  container-type: inline-size;
```
didoo, line 21: I see there are two `container-type: inline-size` declaration (one here, and one in the parent). Are you sure you need this one here? If I comment it out in the devtools I don't see any effect 🤔

packages/components/app/styles/components/page-header.scss
```diff
@@ -0,0 +1,79 @@
...
+  display: flex;
+  flex-direction: column;
+  gap: 1rem;
+  container-type: inline-size;
+}
+
+.hds-page-header__body {
+  display: flex;
+  flex-direction: column;
+  gap: 0.5rem 1rem;
```
didoo, line 20: Niptick: some values are expressed as `rem`, others as `px`. Since the convention is to use `px` unless it's a font-size, can we convert all the `rem`s to `px`?

packages/components/tests/dummy/app/styles/app.scss
didoo, line 1: @jorytindall can you undo these changes? The quotes changes, I mean: in CSS we use double quotes for imports.

packages/components/tests/dummy/app/templates/components/page-header.hbs
```diff
@@ -0,0 +1,323 @@
...
+          <Hds::Button @text="Get started" @icon="terminal-screen" @iconPosition="trailing" />
+        </PH.Actions>
+      </Hds::PageHeader>
+
+      <Shw::Divider @level={{2}} />
+
+      <Shw::Text::H3>Component wrapping</Shw::Text::H3>
+
+      <Shw::Text::Body @tag="p">The component uses container queries to determine two breakpoints at
+        <code>700px</code>
```
didoo, line 288: could it be that it's now `768px`?

packages/components/tests/dummy/app/templates/components/page-header.hbs
```diff
@@ -0,0 +1,323 @@
...
+            </PH.Badges>
+            <PH.Subtitle>Forced small viewport</PH.Subtitle>
+            <PH.Actions>
+              <Hds::Button @text="Edit" @color="secondary" />
+              <Hds::Button @text="Create" @color="primary" />
+            </PH.Actions>
+          </Hds::PageHeader>
+        </SG.Item>
+
+      </Shw::Grid>
```
didoo, line 317: Maybe instead of using a grid with three columns, what you could do is to have the same component inside three different containers, with different widths. Something like this:

```suggestion
      <Shw::Flex @direction="column" as |SF|>
        {{#let (array "399px" "767px" "100%") as |widths|}}
          {{#each widths as |width|}}
            <SF.Item @label="container width = {{width}}">
              <Shw::Outliner {{style width=width}}>
                <Hds::PageHeader as |PH|>
                  <PH.Title>Title of the page</PH.Title>
                  <PH.IconTile @icon="smartphone" />
                  <PH.Breadcrumb>
                    <Hds::Breadcrumb>
                      <Hds::Breadcrumb::Item @text="Organization" />
                      <Hds::Breadcrumb::Item @text="User" />
                    </Hds::Breadcrumb>
                  </PH.Breadcrumb>
                  <PH.Badges>
                    <Hds::Badge @text="Small viewport" @icon="tablet" />
                  </PH.Badges>
                  <PH.Subtitle>Forced small viewport</PH.Subtitle>
                  <PH.Actions>
                    <Hds::Button @text="Edit" @color="secondary" />
                    <Hds::Button @text="Create" @color="primary" />
                  </PH.Actions>
                </Hds::PageHeader>
              </Shw::Outliner>
            </SF.Item>
          {{/each}}
        {{/let}}
      </Shw::Flex>
```

didoo (commented):
@jorytindall did a second pass on the code, added some more comments

packages/components/app/styles/components/page-header.scss
jorytindall: Yes, looks like you're correct. Still figuring out the details of containers, but looks like declaring it once at the top of the component is sufficient.

packages/components/tests/dummy/app/styles/app.scss
jorytindall: Whoops, yes, sorry about that!

packages/components/tests/dummy/app/templates/components/page-header.hbs
jorytindall: That's much more elegant!

packages/components/app/styles/components/page-header.scss
didoo: @jorytindall in case you didn't see it https://developer.mozilla.org/en-US/docs/Web/CSS/container-name (in case you need multiple "containers"; but in this case one is enough)

packages/components/app/styles/components/page-header.scss
```diff
@@ -0,0 +1,79 @@
...
+}
+
+.hds-page-header__content {
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+  gap: 8px;
+}
+
+.hds-page-header__title-main {
```
didoo, line 52: see comment above

```suggestion
.hds-page-header__title-badges-wrapper {
```

packages/components/tests/dummy/app/styles/app.scss
```diff
@@ -23,7 +23,6 @@
 @import "./showcase-components/outliner";
 @import "./showcase-components/placeholder";
 
-
```
didoo, line 26: the two lines were on purpose :|

packages/components/tests/dummy/app/templates/components/page-header.hbs
```diff
@@ -0,0 +1,358 @@
...
+      </Hds::PageHeader>
+
+      <Shw::Divider @level={{2}} />
+
+    </SF.Item>
+
+    <SF.Item @label="Long title + badge">
+
+      <Hds::PageHeader as |PH|>
+        <PH.Title>Long page title with a badge ipsum dolor sit amet</PH.Title>
```
didoo, line 300: We want to capture the wrapping when the page is snapshotted by Percy (to detect possible regressions in future) so the text needs to wrap at full-page width

```suggestion
        <PH.Title>Page title with a very very very very long string that can span multiple lines</PH.Title>
```

didoo (approved):
Left a few minor suggestions/comments, but apart from that it's good to go. Great job! 👍

packages/components/tests/dummy/app/templates/components/page-header.hbs
```diff
@@ -0,0 +1,358 @@
...
+
+    </SF.Item>
+
+    <SF.Item @label="Kitchen sink example">
+
+      <Hds::PageHeader as |PH|>
+        <PH.Title>Users</PH.Title>
+        <PH.Subtitle>Access control (IAM)</PH.Subtitle>
+        <PH.Description>Manage organization users, roles, and access control.
+        </PH.Description>
```
didoo, line 215:
```suggestion
        <PH.Description>Manage organization users, roles, and access control.</PH.Description>
        <PH.Description>And this is a second description, in case multiple paragraphs are needed.</PH.Description>
```

packages/components/tests/dummy/app/styles/app.scss
jorytindall: I believe this occurred at the same time as my changing the the double quotes to single quotes, which I think was prettier being a bit to heavy handed. I've been having issues with it especially related to markdown, let me take a look at my config.

packages/components/tests/dummy/app/templates/components/page-header.hbs
```diff
@@ -0,0 +1,361 @@
...
+
+<section data-test-percy>
+
+  <Shw::Text::H2>Basic examples</Shw::Text::H2>
+
+  <Shw::Flex @direction="column" as |SF|>
+
+    <SF.Item @label="Only title">
+
+      <Hds::PageHeader as |PH|>
```
KristinLBradley, line 18: Super nit: If content is indented it doesn't also need blank lines above and below the indented portion.

PR #1380, opened 2023-05-25
Removes aria-label from stepper/indicator component

MelSumner:
### :pushpin: Summary

If merged, this PR removes the `aria-label` attribute from the hds/stepper/indicator and hds/task/indicator components, which was there incorrectly (a div is not eligible to receive the aria-label attribute)

Here's a failing test from a consuming app: 
CleanShot 2023-05-25 at 14 07 15@2x

### :link: External links

Jira ticket to address the issue of our own tests not catching this: [HDS-1965](https://hashicorp.atlassian.net/browse/HDS-1965)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-1965]: https://hashicorp.atlassian.net/browse/HDS-1965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/addon/components/hds/stepper/step/indicator.hbs
+++ b/packages/components/addon/components/hds/stepper/step/indicator.hbs
@@ -2,7 +2,7 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<div class={{this.classNames}} ...attributes aria-label={{@status}}>
+<div class={{this.classNames}} ...attributes>
   <div class="hds-stepper-indicator-step__svg-hexagon">
     <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg" aria-hidden="true">
       <path
--- a/.changeset/tiny-suns-share.md
+++ b/.changeset/tiny-suns-share.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+A11y Fix: removes `aria-label` from stepper/indicator
--- a/.changeset/tiny-suns-share.md
+++ b/.changeset/tiny-suns-share.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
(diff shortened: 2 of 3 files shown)


packages/components/addon/components/hds/stepper/step/indicator.hbs
```diff
@@ -2,7 +2,7 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<div class={{this.classNames}} ...attributes aria-label={{@status}}>
```
KristinLBradley, line 5: Is the status information valuable? Should it be moved to somewhere else?

didoo (approved):
OK for me.

A couple of question: 
- would it be possible to link (or take a screenshot) of the failing tests in TFC, and add it to the ticket description, for future reference? Thanks
- since it's already used in Cloud UI, how's that a11y tests didn't fail there?

didoo:
@MelSumner there are two `Stepper` indicators, `Step::Indicator` and `Task::Indicator` but you've removed the aria label only from the "step" one. Was it intentional?

MelSumner:
@didoo I will check for other locations that might have this, I was going off the failing TFC test and immediately switching to the task of figuring out why _our_ tests didn't fix this. I can do a quick scan for "is this happening in another component too" though.

packages/components/addon/components/hds/stepper/step/indicator.hbs
MelSumner: By removing the incorrect `aria-label` we will at least stop their tests from failing. 

I'm looking at the component to see if there's a better way to do this but IIRC this was supposed to be integrated into the consumer's components and not used on its own; I'll need to do a little more checking to see what the right next fix is. 👍

MelSumner:
Since the tests are passing and I'm out for the long weekend, and I know we have team consensus, I'm going to merge so our team can ship this for consumers.

PR #1381, opened 2023-05-25
`Reveal` component (HDS-1911)

KristinLBradley:
### :pushpin: Summary

_**STATUS:** Code is ready for release pending completion of ([Web docs](https://github.com/hashicorp/design-system/pull/1386))_

If merged, this PR will add the `Reveal` component to the design system.

- **Showcase preview:** https://hds-showcase-git-hds-1911-reveal-component-hashicorp.vercel.app/components/reveal

**Current status details:**
* `Reveal` component implementation:
 * Styling
 * API implementation
 * Integration tests
* Showcase examples including states, options, examples of different use cases taken from Cloud-ui, etc
* Updated `Hds::DisclosurePrimitive` to add `@isOpen` prop to control status from outside (so DisclosurePrimitive can optionally be opened on render)
* [Documentation branch](https://github.com/hashicorp/design-system/pull/1386) with tech & accessibility docs.
* changeset added

**_Design update to add minimal padding and no offset for toggle buttons:_**
image

### :link: External links
- Jira ticket: [HDS-1911](https://hashicorp.atlassian.net/browse/HDS-1911)
- [Figma file](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/ZXJVYXcw09ZL2Xl2Mfcmic/HDS-Product---Components?type=design&node-id=35103-70091&t=ebnGw7Nz4paIPK7c-0)
- [CRD](https://docs.google.com/document/d/1TKgRk8OI8XKM4WPQ9_lYG7yGsKHOZYx0DIp2Pe5DCBM/edit#heading=h.jxhqdjc34uui)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-1911]: https://hashicorp.atlassian.net/browse/HDS-1911?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/app/styles/components/reveal.scss
```diff
@@ -0,0 +1,40 @@
...
+//
+// REVEAL COMPONENT
+//
+
+@use "../mixins/button" as *;
+
+.hds-reveal {
+  display: inline-block; // discuss if desired
+  // outline: 1px solid limegreen;
+  // outline-offset: -1px;
```
KristinLBradley, line 15: Note: Outlines are temporary for visualizing and testing the layout

packages/components/app/styles/components/reveal.scss
```diff
@@ -0,0 +1,40 @@
...
+}
+
+.hds-reveal__button {
+  position: relative;
+  left: -0.9375rem;
+  margin-right: (-0.9375rem * 2); // not sure if this is a good idea (trying to account for space to left and right of button)
```
KristinLBradley, line 21: Let me know if you have a better ides for this or feel it should be modified in some way. (Moved to parent maybe?)

packages/components/app/styles/components/reveal.scss
KristinLBradley, line 1: This is just rough preliminary CSS. Still refining...

packages/components/tests/dummy/app/templates/components/reveal.hbs
KristinLBradley, line 1: I could use suggestions on different helpful use cases to include in the Showcase

packages/components/tests/dummy/app/templates/components/reveal.hbs
```diff
@@ -0,0 +1,114 @@
...
+  <Shw::Flex @direction="column" {{style gap="2rem"}} as |SF|>
+    <SF.Item @label="Used inline with text (edge case)">
+      <p class="hds-typography-body-300">
+        Lorem ipsum dolor sit amet consectetur
+        <Hds::Reveal @text="More options">
+          Additional content
+        </Hds::Reveal>
+        adipisicing elit. Doloremque blanditiis sapiente iste beatae voluptates voluptatum.
+      </p>
+    </SF.Item>
```
KristinLBradley, line 51: This example is an edge case. May want to move or even delete if it's not helpful.

packages/components/tests/integration/components/hds/reveal/index-test.js
KristinLBradley, line 1: Will add tests as an upcoming step

KristinLBradley:
This is still in a rough preliminary stage. I could use overall feedback on the API, functionality I'm missing, helpful Showcase examples to add, or anything that currently looks problematic.

* CSS is still rough but could use basic feedback on what's there so far
* I haven't worked on adding tests yet. Will add once API is more set.

packages/components/app/styles/components/reveal.scss
```diff
@@ -0,0 +1,45 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+//
+// REVEAL COMPONENT
+//
+
+:root {
+  // Spacing
+  --token-reveal-content-margin-top: 4px;
```
KristinLBradley, line 12: Let me know if you think other values should be tokenized. (I will create JSON tokens file closer to end of development.)

packages/components/tests/dummy/app/templates/components/reveal.hbs
```diff
@@ -0,0 +1,163 @@
...
+      </p>
+      <Hds::Reveal @text="More options">
+        Additional content
+      </Hds::Reveal>
+      <p class="hds-typography-body-300">
+        adipisicing elit. Doloremque blanditiis sapiente iste beatae voluptates voluptatum.
+      </p>
+    </SF.Item>
+
+    <SF.Item @label="Used inline with text (edge case)">
```
MelSumner, line 104: I would think that we want to explicitly not include this, maybe even have documentation for this as a "don't".

packages/components/app/styles/components/reveal.scss
MelSumner: Have we thought about this always having some kind of border/visual indicator? 

If it doesn't, then you have to use negative margins and it looks bad when it's open: 
CleanShot 2023-05-30 at 12 00 03

If it did, maybe everything would stay aligned whether opened or closed? It's entirely possible everyone has had this discussion ad nauseam but I thought I'd at least ask about it.

packages/components/addon/components/hds/reveal/toggle/button.hbs
```diff
@@ -0,0 +1,14 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<Hds::Button
```
MelSumner, line 6: I see that you have `aria-expanded` here, I think you may need to also add `aria-controls` as well. Here, a widget is just "something else is revealed or hidden": 

> A button that opens a widget should have `aria-controls` set to the `id` of the expandable widget and `aria-expanded` set to the current state of the widget.

packages/components/app/styles/components/reveal.scss
KristinLBradley: I agree it looks a little awkward when the borders are visible on hover or focus but this is my understanding of the desired design. This trade-off was so that the text content aligns with other content when the borders aren't visible.

@cveigt Could you confirm the desired design?

packages/components/tests/dummy/app/templates/components/reveal.hbs
KristinLBradley: I included it partially just to see how content right next to it looks in relation although I don't think it would ever be practically used this way. Maybe I could move it to another more clear "Edge case" section? 

@didoo @alex-ju What do you think? Would it be helpful to keep this example around or should I just delete it?

packages/components/tests/dummy/app/templates/components/reveal.hbs
alex-ju, line 90: thought: the fact that this component can be used inline (at least at this point, considering the `inline-block` property) makes this example very much relevant, as this is how a consumer would see it rendered in their application next to any other inline element. Moving to a `block` element would make this obsolete, indeed.

packages/components/addon/components/hds/reveal/toggle/button.hbs
KristinLBradley: Pushed up changes, thanks

packages/components/tests/dummy/app/templates/components/reveal.hbs
alex-ju: Generally, I try to make sure I accommodate existing use cases and example provided in the design reference (to ease the visual review)

alex-ju (commented):
First-look feedback on API and showcase: API looks intuitive and showcase seems relevant to me considering the current implementation.

packages/components/tests/dummy/app/templates/components/reveal.hbs
KristinLBradley: I added a number of examples from Cloud-ui since I first added this comment. If anyone has suggestions for additional helpful examples though, please let me know.

packages/components/tests/dummy/app/templates/components/reveal.hbs
```diff
@@ -0,0 +1,167 @@
...
+
+    <SF.Item @label="With interactive content">
+      <Hds::Reveal @text="More options">
+        See here for more info:
+        <a href="https://www.hashicorp.com/" target="_blank" rel="noopener noreferrer">HashiCorp</a>
+      </Hds::Reveal>
+    </SF.Item>
+  </Shw::Flex>
+
+  <Shw::Text::H3>Used next to other content</Shw::Text::H3>
```
heatherlarsen, line 69: @cveigt Have you guys considered/discussed setting a default margin above/below the `Reveal` to ensure elements have some space between them? This would be based on the guidance you intend to include in the documentation, as was noted in Figma. I could see this being helpful so long as they can easily override the value for special use cases.

packages/components/addon/components/hds/reveal/index.hbs
```diff
@@ -0,0 +1,21 @@
...
+    <Hds::Reveal::Toggle::Button
+      aria-controls={{this.contentId}}
+      @text={{if (and t.isOpen @textWhenOpen) @textWhenOpen @text}}
+      @isOpen={{t.isOpen}}
+      {{on "click" t.onClickToggle}}
+    />
+  </:toggle>
+
+  <:content>
+    <div class="hds-reveal__content hds-typography-body-300" id={{this.contentId}}>
```
heatherlarsen, line 17: @KristinLBradley I see you're using the term "content" here, but in the `Tabs` component, we used the term "panel." Should these be consistent?

packages/components/tests/dummy/app/templates/components/reveal.hbs
```diff
@@ -0,0 +1,167 @@
...
+    <SF.Item @label="Used inline with text">
+      <p class="hds-typography-body-300">
+        Lorem ipsum dolor sit amet consectetur
+        <Hds::Reveal @text="More options">
+          Additional content
+        </Hds::Reveal>
+        adipisicing elit. Doloremque blanditiis sapiente iste beatae voluptates voluptatum.
+      </p>
+    </SF.Item>
+  </Shw::Flex>
```
heatherlarsen, line 117: @KristinLBradley I see that this says "Edge case," but has there been an example of this somewhere? If so, this makes me worried about our approach to the overhanging states because they end up covering the text when hovered. When opened, it also pushes the original content down, which is quite an odd experience. I would expect the original content to remain in place and the additional (hidden) content to open below.  

cc @cveigt

cveigt: This is not a real use case on our product. But even if it was, I would expect the behavior to be as @heatherlarsen  describes 

> I would expect the original content to remain in place and the additional (hidden) content to open below.

One thing to note is that the showcase is an internal tool (at least for now) and is used to test different scenarios. We should not expect this to be anywhere on the documentation website.

packages/components/app/styles/components/reveal.scss
cveigt: It's definitely not ideal. In general, this is a trade-off that comes with using the tertiary button. Now, what @MelSumner outlines is a real use case, and it looks particularly bad when used on a form. How much updating the style or building a custom toggle would affect the scope?

packages/components/tests/dummy/app/templates/components/reveal.hbs
cveigt: I didn't consider having margin built into the component, if that's what you mean. Have we done that in other components? cc @KristinLBradley

packages/components/app/styles/components/reveal.scss
```diff
@@ -0,0 +1,45 @@
...
+}
+
+.hds-reveal {
+  display: inline-block; // discuss if desired
```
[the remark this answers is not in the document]
cveigt: @KristinLBradley, what would it be used for?

packages/components/addon/components/hds/reveal/toggle/button.js
```diff
@@ -0,0 +1,60 @@
...
+  get text() {
+    let { text } = this.args;
+
+    assert(
+      '@text for "Hds::Reveal::Toggle::Button" must have a valid value',
+      text !== undefined
+    );
+
+    return text;
+  }
```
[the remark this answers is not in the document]
didoo: OK, now I understand what you intended to do. But in this case I would argue that this assertion needs to live in the `Hds::Reveal` backing class, not here.

didoo (changes requested):
Added a few comments, some of them I think are blockers

didoo (changes requested):
Added a few comments, some of them I think are blockers

packages/components/app/styles/components/reveal.scss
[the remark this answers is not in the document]
KristinLBradley: I would personally prefer to keep it as inline-block to mimic the default layout of an HTML button and leave layout in relation to other content to the consumer. I can delete it though if you want it to display as block instead.

packages/components/addon/components/hds/reveal/index.hbs
KristinLBradley: To me, the content which is hidden and shown by `Reveal` is more generic so just using "content" makes sense. (In `Tabs` there are a series of "panels" which are hidden and shown vs. one chunk of "content" being toggled between hiding/showing) Also it fits the naming used by the primitive as @didoo said.

packages/components/app/styles/components/reveal.scss
```diff
@@ -0,0 +1,42 @@
...
+.hds-reveal__toggle-button--is-open {
+  .flight-icon-chevron-down {
+    transform: rotate(-180deg);
+  }
+}
+
+.hds-reveal__content {
+  margin-top: var(--token-reveal-content-margin-top);
+
+  img { max-width: 100%; }
```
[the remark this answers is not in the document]
KristinLBradley: Ok but then consumers have to write custom styles for a common usage pattern within Cloud-ui. Anyway I can delete it if you like.

packages/components/app/styles/components/reveal.scss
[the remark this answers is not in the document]
KristinLBradley: Anyway I'll just delete it.

packages/components/addon/components/hds/reveal/index.js
```diff
@@ -0,0 +1,30 @@
...
+   * @return {string} The "class" attribute to apply to the component.
+   */
+  get classNames() {
+    let classes = ['hds-reveal'];
+
+    // add a class based on the @xxx argument
+    // classes.push(`hds-reveal--[variant]-${this.xxx}`);
+
+    return classes.join(' ');
+  }
```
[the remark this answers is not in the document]
KristinLBradley: Thanks, I usually leave this until the end to delete in case I need to make use of it to control additional class names. I'll go ahead and delete it.

packages/components/addon/components/hds/reveal/index.hbs
heatherlarsen: Should `Tabs` be updated to "content" then? :homerfade:

packages/components/addon/components/hds/reveal/index.hbs
```diff
@@ -0,0 +1,21 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<Hds::DisclosurePrimitive class={{this.classNames}} @isOpen={{@isOpen}} ...attributes>
+  <:toggle as |t|>
+    <Hds::Reveal::Toggle::Button
+      aria-controls={{this.contentId}}
+      @text={{if (and t.isOpen @textWhenOpen) @textWhenOpen @text}}
```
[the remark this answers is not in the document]
KristinLBradley: I can change it but to me this seems simpler and less convoluted.

@MelSumner @alex-ju do either of you have a preference for how to handle something like this?

packages/components/addon/components/hds/reveal/toggle/button.js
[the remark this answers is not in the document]
KristinLBradley: Thanks I wasn't sure if it was helpful to include the assertion here or not.

packages/components/addon/components/hds/reveal/toggle/button.js
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * @param onClick
+   * @type {function}
+   * @default () => {}
+   */
+  get onClick() {
```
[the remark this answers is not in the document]
KristinLBradley: I am actually using the toggle button as a standalone element in the showcase though which is what the check guards for (see comment, I copied this from a similar example I found)...

packages/components/addon/components/hds/reveal/index.hbs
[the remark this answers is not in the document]
KristinLBradley: I'd rather just keep it as is. When I try changing it then I'm having trouble getting the text to change because the tracked state of isOpen doesn't seem to work properly (at least so far in what I tried). Anyway changing it introduces more complexity...

packages/components/addon/components/hds/reveal/index.hbs
KristinLBradley: No, `Tabs` is a more complex component and there are a series of full width "panels" that are controlled by activating the individual tabs. `Reveal` is much simpler and just toggles on and off the display of generic content. The content isn't full width and doesn't swap display between other content.

At least that's how I think of it...

packages/components/tests/dummy/app/templates/components/reveal.hbs
KristinLBradley: This has since been deleted so I'm resolving this thread, thanks

cveigt:
@KristinLBradley we have a `content type` variant in Figma with `default` and `custom` as possible values (`default` being basically a text box/paragraph element).

To support it in code, I see two options:
1. Add default styling to the content. In this case, it'd be the typography styles for standard body text, which is what we discussed.
2. Correct me if I'm wrong, but we can also create a property on the API for a content type that can be set to default or custom/generic.

I'm curious to know why one would work better than the other.

KristinLBradley:
> 



> @KristinLBradley we have a `content type` variant in Figma with `default` and `custom` as possible values (`default` being basically a text box/paragraph element).
> 
> To support it in code, I see two options:
> 
> 1. Add default styling to the content. In this case, it'd be the typography styles for standard body text, which is what we discussed.
> 2. Correct me if I'm wrong, but we can also create a property on the API for a content type that can be set to default or custom/generic.
> 
> I'm curious to know why one would work better than the other.

I don't understand the 2nd option. All I am doing currently is setting the "hds-typography-body-300" class on the content container so if consumers just add pain text as the content it will have styling. If consumers choose to nest other tags inside the content div then they can add other class names to override the container styling as they choose. (Or if they include components which come with font styling then those styles would also override the container styles.)

@didoo Would appreciate if you could weigh in on this also. Should I include the "hds-typography-body-300" class on the content container or just leave it off so that consumers would always have to style text content themselves? (Or if they use global font styles then those would just get inherited.)

packages/components/app/styles/components/reveal.scss
cveigt: @KristinLBradley @didoo I was looking into this issue and I don’t see many routes we can take using the tertiary button without making significantly updating its styling and changing the current states.

Here are two options I consider could work.

Screen Shot 2023-05-31 at 11 26 22 PM

**Option A** removes the offset added to the toggle, so, it is basically the button as is. I don’t like it because it looks okay (not great) in some cases and pretty off in others, which leaves us in a similar situation.

**Option B** removes the current offset and modifies the internal padding on all sides of the button. This looks much better and doesn't seem unreasonable, IMO. If this is not feasible, **option C** would be building a custom toggle/button because this is a real use case, and I worry that we may start to get comments and feedback from both consumers and other stakeholders or maybe even users/customers that it looks like a bug or a glitch in the UI, similarly to how happened with the Button/Link.

@heatherlarsen would appreciate your thoughts on this as well.

cveigt: @KristinLBradley I looked into your suggestion

> It's not that difficult to just add a border that's always visible

This is a good idea but it would look almost like a secondary button, which is not intended, and when used near other buttons or form controls, it is a bit confusing as they look similar.

packages/components/tests/dummy/app/templates/components/reveal.hbs
didoo: Adding default margins to components is an anti-pattern in design systems. Let's try to avoid it as much as possible, leads to a lot of problems (that's why it has become an anti-pattern, because people thought it was a good idea, tried it, and found out it was not).

packages/components/addon/components/hds/reveal/index.hbs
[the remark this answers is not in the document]
didoo: OK, I see now why it needs to live in the template, because the `isOpen` in the `Reveal` is technically a `isInitiallyOpen` while the `isOpen` in the `Reveal::Toggle::Button` is actually the toggle state (the fact that both have the same name lead me to this confusion). Leave it as is, then.

didoo:
> > 
> 
> > @KristinLBradley we have a `content type` variant in Figma with `default` and `custom` as possible values (`default` being basically a text box/paragraph element).
> > To support it in code, I see two options:
> > 
> > 1. Add default styling to the content. In this case, it'd be the typography styles for standard body text, which is what we discussed.
> > 2. Correct me if I'm wrong, but we can also create a property on the API for a content type that can be set to default or custom/generic.
> > 
> > I'm curious to know why one would work better than the other.
> 
> I don't understand the 2nd option. All I am doing currently is setting the "[hds-typography-body-300](https://hashicorp.atlassian.net/browse/HDS-typography-body-300)" class on the content container so if consumers just add pain text as the content it will have styling. If consumers choose to nest other tags inside the content div then they can add other class names to override the container styling as they choose. (Or if they include components which come with font styling then those styles would also override the container styles.)
> 
> @didoo Would appreciate if you could weigh in on this also. Should I include the "[hds-typography-body-300](https://hashicorp.atlassian.net/browse/HDS-typography-body-300)" class on the content container or just leave it off so that consumers would always have to style text content themselves? (Or if they use global font styles then those would just get inherited.)

@cveigt FYI @KristinLBradley is already doing the option 1, only she's using `body-300` instead of `body-200`

@KristinLBradley I think you just have to change the class name associated with the content and use `body-200` (that's the style I see in Figma)
modified

packages/components/addon/components/hds/reveal/toggle/button.js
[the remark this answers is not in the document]
didoo: > as a standalone element in the showcase

Yes, but this doesn't mean that we should change a sub-component behaviour only for the showcase, if we need to show it in isolation for demo purpose, but we know it's not a real use case. 

The showcase is helpful to identify _real_ use cases we didn't think of, not the other way round :)

packages/components/app/styles/components/reveal.scss
didoo: probably for such a simple component I would not bother to create a single token and just use an inlined value in the declaration.

packages/components/app/styles/components/reveal.scss
MelSumner: @cveigt since it _is_ a button from a functional perspective, I wonder if the differentiation should be in the pattern guidance? like, how to use a reveal when you also have some form buttons. Just a thought.

heatherlarsen: @cveigt Thanks for putting this side by side together! Option B feels like the right fit. That would be my vote if it's feasible from a technical perspective and doesn't add too much complexity to the project.

packages/components/tests/dummy/app/templates/components/reveal.hbs
KristinLBradley: Going to go ahead and resolve this thread as I agree it adds more potential problems than it solves to add default margins to this type of component.

packages/components/addon/components/hds/reveal/index.hbs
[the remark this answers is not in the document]
KristinLBradley: Do you have suggestions for changing any of the names to help prevent confusion?

packages/components/app/styles/components/reveal.scss
KristinLBradley: Ok I just went ahead and removed the CSS custom property and added the value in the styles directly. Let me know if I misunderstood what you meant though.

packages/components/app/styles/components/reveal.scss
KristinLBradley: Going to try Option B (remove button side padding) and see what @cveigt and others think...

cveigt: The latest changes look good, @KristinLBradley

KristinLBradley: Thanks! Will close this thread as resolved then.

cveigt (approved):
Changes look good 👍  Thanks, Kristin!

packages/components/app/styles/components/reveal.scss
```diff
@@ -0,0 +1,30 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+//
+// REVEAL COMPONENT
+//
+
+.hds-reveal__toggle-button {
+  min-height: 1.75rem; // 28px, Override default so padding difference will be visible
```
didoo, line 11: I would add a comment here, highlighting that we're overriding some styles of the tertiary button, per design specs

packages/components/app/styles/components/reveal.scss
```diff
@@ -0,0 +1,30 @@
...
+
+.hds-reveal__toggle-button--is-open {
+  .flight-icon-chevron-down {
+    transform: rotate(-180deg);
+  }
+}
+
+.hds-reveal__content {
+  margin-top: 4px;
+  color: var(--token-color-foreground-primary);
```
didoo, line 29: consider applying the class `hds-foreground-primary` to the element with class `hds-reveal__content hds-typography-body-200` instead of using a token here (so all the styling is in the same place)

packages/components/tests/integration/components/hds/reveal/index-test.js
```diff
@@ -0,0 +1,127 @@
...
+    await click('.hds-reveal__toggle-button');
+    assert.dom('.hds-reveal__content').exists().hasText('Additional content');
+    // Test that content is hidden after the toggle is triggered again
+    await click('.hds-reveal__toggle-button');
+    assert.dom('.hds-reveal__content').doesNotExist();
+  });
+
+  // A11Y
+
+  test('it displays the correct value for aria-expanded when closed vs.open', async function (assert) {
```
didoo, line 52:
```suggestion
  test('it displays the correct value for aria-expanded when closed vs open', async function (assert) {
```

packages/components/tests/dummy/app/templates/components/reveal.hbs
```diff
@@ -0,0 +1,153 @@
...
+        <F.HelperText>Ensure that the private subnets have access to the external network.
+        </F.HelperText>
+      </Hds::Form::TextInput::Field>
+      <Hds::Reveal @text="Where can I find this?">
+        <Hds::Alert @type="compact" as |A|>
+          <A.Description>Navigate to Subnets(1), if you already have subnet created for the VPC of choice, copy the
+            Subnet ID(2). Otherwise, Create subnet(3) for chosen VPC.</A.Description>
+        </Hds::Alert>
+        <p>
+          <img
```
didoo, line 84: this image should have a special "local" class (specific for this page) that sets the max width to 100% (same as it would happen on the consumers side)

didoo (approved):
All good for me. I've only left a bunch of minor comments/suggestions

packages/components/tests/integration/components/hds/reveal/index-test.js
```diff
@@ -0,0 +1,127 @@
...
+    assert.dom('.hds-reveal__content').doesNotExist();
+    // Test that content is displayed after the toggle is triggered
+    await click('.hds-reveal__toggle-button');
+    assert.dom('.hds-reveal__content').exists().hasText('Additional content');
+    // Test that content is hidden after the toggle is triggered again
+    await click('.hds-reveal__toggle-button');
+    assert.dom('.hds-reveal__content').doesNotExist();
+  });
+
+  // A11Y
```
alex-ju, line 50: under this section, should we test for `aria-controls` as well?

alex-ju (approved):
Looking great! I think we should add a test to avoid accessibility regression on `aria-controls` otherwise good to be merged

KristinLBradley:
> Looking great! I think we should add a test to avoid accessibility regression on `aria-controls` otherwise good to be merged

Thanks! Btw, do you know what test I can use to check if the value of the toggle button aria-controls matches the id of the content div? I'm looking through different possible tests but not finding an example yet...

heatherlarsen (approved):
:shipit:

packages/components/tests/integration/components/hds/reveal/index-test.js
KristinLBradley: Added a test, thanks

MelSumner (approved):

website/docs/components/reveal/partials/guidelines/guidelines.md
```diff
@@ -0,0 +1,66 @@
...
+!!!
+
+!!! Do
+
+![Reveal example with simple content](/assets/components/reveal/content-type-do.png =604x*)
+!!!
+
+## Spacing
+
+- When using the Reveal to contain information related to a single element within a group or section, we recommend to leave 8px between the Reveal and the element it relates to.
```
alex-ju, line 48:
```suggestion
- When using the Reveal to contain information related to a single element within a group or section, we recommend leaving 8px between the Reveal and the element it relates to.
```

website/docs/components/reveal/partials/guidelines/guidelines.md
```diff
@@ -0,0 +1,66 @@
...
+
+![Reveal example within a section](/assets/components/reveal/spacing-guidance-single-element.png =591x*)
+
+- When used to show and hide information not directly associated to a single element but an entire section or group, we recommend to leave 24px between the Reveal and the section it relates to.
```
alex-ju, line 52:
```suggestion
- When used to show and hide information not directly associated with a single element but an entire section or group, we recommend leaving 24px between the Reveal and the section it relates to.
```

website/docs/components/reveal/partials/code/how-to-use.md
```diff
@@ -0,0 +1,29 @@
+## How to use this component
+
+The Reveal component renders a button which triggers the display of additional content. The additional content can consist of either plain text or HTML and can include interactive elements such as links.
```
alex-ju, line 3:
```suggestion
The Reveal component renders a button that triggers the display of additional content. The additional content can consist of either plain text or HTML and can include interactive elements such as links.
```

website/docs/components/reveal/partials/code/how-to-use.md
```diff
@@ -0,0 +1,29 @@
...
+
+```handlebars
+<Hds::Reveal @text="Open me" @textWhenOpen="Close me">
+  Additional content
+</Hds::Reveal>
+```
+
+### isOpen
+
+Set `isOpen` to true in order to display the content on page load instead of initially hiding it.
```
alex-ju, line 23:
```suggestion
Set `isOpen` to `true` to display the content on page load instead of initially hiding it.
```

website/docs/components/reveal/partials/code/component-api.md
```diff
@@ -0,0 +1,16 @@
+## Component API
+
+<Doc::ComponentApi as |C|>
+  <C.Property @name="text" @required="true" @type="string">
+    Plain text string which will appear on the toggle button which controls hiding and showing of the content. If no text value is defined an error will be thrown.
+  </C.Property>
+  <C.Property @name="textWhenOpen" @type="string">
+    Plain text which displays on the toggle button while the content is displayed.
+  </C.Property>
+  <C.Property @name="isOpen" @values={{array "false" "true" }} @default="false" @type="boolean">
```
alex-ju, line 10: I think [we now consider boolean values redundant](https://github.com/hashicorp/design-system/pull/1369/files), so only having the default should do.

alex-ju (commented):
A few suggestions (non-blocking) before getting this in

website/docs/components/reveal/partials/code/component-api.md
KristinLBradley, line 10:
```suggestion
  <C.Property @name="isOpen" @default="false" @type="boolean">
```

PR #1383, opened 2023-05-26
remove version caveat

Dhaulagiri:
This was added in https://github.com/hashicorp/design-system/pull/670 but now that all our consumer are on >= 3.28 this caveat isn't really necessary


alex-ju (approved):
Amended your commit to removed it from `ember-flight-icons` too

Dhaulagiri:
🤦  thank you!

PR #1388, opened 2023-05-30
Fix `SideNav` issue with content resizing when transitioning its `width`

didoo:
### :pushpin: Summary

This PR solves an issue around the resizing of the content when the width of the `SideNav` is animated/transitioned:

https://github.com/hashicorp/design-system/assets/686239/af7a4cef-2921-4e19-826a-de6d915f90d7
(the animation is slowed down for debugging)

This is a regression from the initial implementation and it's due to the changes introduced in https://github.com/hashicorp/design-system/pull/1338/

With these changes, I've move the target of the `hds-side-nav-hide-when-minimized` from the "panels" container to the "panel" elements. The intention was to avoid the navigational elements to be interactive while not fully visible yet (the visibility of an element is not inherited, a child can be set to visible even if the parent is not visible) because their opacity was transitioning. The problem with this change is that the single panels **also** receive some CSS properties' changes via JavaScript (to control the sliding and appearence of the panels) and so there were conflicts with the `visibility` property.

The final effect was hard to spot, because the entire transition is quite quick, but still perceivable.

So in this PR I have reverted the changes, and added a new special class that disables the interactivity of the content while the sidenav is transitioning states, solving the problem at the root.

For context, while debugging I've another issue, almost undetectable, a flickering of the panels when injected (sub-navigation). This will be solved in a separate ticket ([HDS-1971](https://hashicorp.atlassian.net/browse/HDS-1971)), because it needs some non-trivial JS refactoring.

In this PR I have:
- moved back target of "hide when minimized" class to the "panels" container
    - reverted change done in https://github.com/hashicorp/design-system/pull/1338/
- added special class `hds-side-nav--is-animating` to disable interaction while animating/transitioning
  - I've used the same technique used in the previous `hc-nav` implementation:
     - https://github.com/hashicorp/cloud-ui/blob/de3a8b71f7cb5b87d1945866bf20cac55fcc2b7e/addons/hc-nav/addon/components/hc-nav/container/index.hbs#L7-L8
     - https://github.com/hashicorp/cloud-ui/blob/de3a8b71f7cb5b87d1945866bf20cac55fcc2b7e/addons/hc-nav/addon/components/hc-nav/container/index.js#L62-L73
- improved definition of panel’s default width (not visible to users, but technically more correct)

A frame captured in the middle of the animation:
screenshot_2722

### :link: External links

Slack conversation: https://hashicorp.slack.com/archives/C7KTUHNUS/p1685041498766739
Jira ticket: https://hashicorp.atlassian.net/browse/HDS-1970

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable

[HDS-1971]: https://hashicorp.atlassian.net/browse/HDS-1971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/nice-clouds-sort.md
+++ b/.changeset/nice-clouds-sort.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Fixed `SideNav` issue with content resizing when transitioning its `width`
--- a/packages/components/addon/components/hds/side-nav/index.hbs
+++ b/packages/components/addon/components/hds/side-nav/index.hbs
@@ -3,7 +3,13 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-<Hds::SideNav::Base class={{this.classNames}} ...attributes {{focus-trap isActive=this.shouldTrapFocus}}>
+<Hds::SideNav::Base
+  class={{this.classNames}}
+  ...attributes
+  {{on "transitionstart" (fn this.setTransition "start")}}
+  {{on "transitionend" (fn this.setTransition "end")}}
+  {{focus-trap isActive=this.shouldTrapFocus}}
+>
   <:root>
     {{#if this.hasA11yRefocus}}
       <NavigationNarrator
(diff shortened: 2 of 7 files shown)


alex-ju (approved):
Tested this and works as expected; code looks good too. Don't forget the changelog entry.

PR #1391, opened 2023-05-31
Embed page scroll management into `Modal` and `Flyout`

alex-ju:
### :pushpin: Summary

Embed page scroll management into `Modal` and `Flyout`

We currently [recommend consumers implement the page scroll management](https://helios.hashicorp.design/components/modal?tab=code#page-scroll) (disabled page scroll when open and enable when closed) for `Modal` and `Flyout` using the `onOpen` and `onClose` callbacks. The reason behind this initial approach was to keep components encapsulated and avoid interfering with other elements on the page (in this case the `<body>` element).

Based on feedback from Terraform/Atlas and after observing the above recommendation is not followed in practice we decided to embed the logic into the `Modal` and `Flyout` components and remove the recommendation from the associated docs.

Note: have tried a few approaches for testing this new logic, but the way qunit tests are set up does not allow access to the `<body>` element

### :link: External links

Jira ticket: [HDS-1963](https://hashicorp.atlassian.net/browse/HDS-1963)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/flyout/index.js
+++ b/packages/components/addon/components/hds/flyout/index.js
@@ -60,8 +60,14 @@ export default class HdsFlyoutIndexComponent extends Component {
 
   @action
   didInsert(element) {
-    // Store a reference of the `<dialog>` element
+    // Store references of `<dialog>` and `<body>` elements
     this.element = element;
+    this.body = document.body;
+
+    if (this.body) {
+      // Store the initial `overflow` value of `<body>` so we can reset to it
+      this.bodyInitialOverflowValue = this.body.style.overflow;
+    }
 
     // Register `<dialog>` element for polyfilling if no native support is available
     if (!element.showModal) {
@@ -98,6 +104,9 @@ export default class HdsFlyoutIndexComponent extends Component {
     this.element.showModal();
     this.isOpen = true;
 
+    // Prevent page from scrolling when the dialog is open
+    if (this.body) this.body.style.overflow = 'hidden';
+
     // Call "onOpen" callback function
     if (this.args.onOpen && typeof this.args.onOpen === 'function') {
       this.args.onOpen();
@@ -108,5 +117,8 @@ export default class HdsFlyoutIndexComponent extends Component {
   onDismiss() {
     // Make flyout dialog invisible using the native `close` method
     this.element.close();
+
+    // Reset page `overflow` property
+    if (this.body) this.body.style.overflow = this.bodyInitialOverflowValue;
   }
 }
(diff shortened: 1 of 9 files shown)


didoo (approved):
All good for me. All very neat and elegant, as usual :)

I have only one doubt and a question: 
- since we're adding a default behaviour that _hypothetically_ is altering how the component interacts with the context in which it is used, and interfere with other existing behaviours, should we _ideally_ consider this a major and not a minor? (mixed feelings on this, I just wanted to hear your point of view)
- what will happen on the consumers side once we release the changes? should they make some code updates on their side too? if that is the case, what is the plan? we'll open PRs for them, or we'll just tell them what to do?

packages/components/addon/components/hds/flyout/index.js
```diff
@@ -98,6 +104,9 @@ export default class HdsFlyoutIndexComponent extends Component {
     this.element.showModal();
     this.isOpen = true;
 
+    // Prevent page from scrolling when the dialog is open
+    if (this.body) this.body.style.overflow = 'hidden';
```
meirish, line 108: I _think_ setting a style property directly like this might be incompatible with apps that use a strict Content Security Policy (this would violate inline-style directives). `[style.setProperty](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty)` safely updates the CSSOM (this is the API `ember-style-modifier` uses to safely add inline styles).

alex-ju: Indeed, that's something worth investigating. I guess an alternative would be to add/remove a CSS class on the `<body>` element – not elegant either, but more in line with stricter CSPs

alex-ju: ah, right, setters and getters are fine – thanks for the heads-up, will update and re-test

alex-ju:
@didoo, I touched the code since your review to include Matthew's suggestion – would you be able to have another look before merging?

packages/components/addon/components/hds/flyout/index.js
```diff
@@ -98,6 +105,9 @@ export default class HdsFlyoutIndexComponent extends Component {
     this.element.showModal();
     this.isOpen = true;
```
MelSumner, line 106: (non-blocking comment) fwiw, I usually see the things that are tracked first inside the action, so I'd probably switch lines 98 and 99. I don't know if this is just a personal preference, or how I learned to do it, or if there's some reason for it but I thought that I'd mention it for your consideration.

didoo (approved):
Re-reviewed. All good 👍

(good catch @meirish 🙌)

PR #1403, opened 2023-06-06
WIP: new GitHub action: trying ember-a11y-testing as nightly cron job

MelSumner:
### :pushpin: Summary

If merged, this PR adds a GitHub action that runs `ember-a11y-testing` on `packages/components` as a nightly cron job.

#### Slower test runs as the number of tests grows

It was fine to run a11y tests globally when we didn't have a lot of tests. Now we have 715+ tests, and it takes a very long time to run a11y tests at the same time as all of the other tests. 

The full test suite takes about 28 minutes to run, vs a little over 1 minute without the a11y test suite. 

To try this yourself, navigate to the `packages/components` folder locally, and run `ember test --serve`. Open the browser URL it gives you, and note about how long this takes (it's not long).

After that finishes, go up to the URL bar, and where it says `?hidepassed` add `&enableA11yAudit` as a second query param, then press enter. Now, you'll be able to see first-hand how long the tests take. 

Our team has cared about how long it takes our tests to run, but I also want us to have a11y tests, so we can really be sure that we're delivering an accessible product. 

I've added a temporary file for discussion - `test-output.txt` so we can see the tests that are false positives. 

- adds a new github workflow file
- adds a new script definition to package.json
- updates the test helpers global rule definitions

TODO
- can we skip the acceptance tests? These are for our dummy app and are for our visual regression tests, so I'd like to skip the whole group of acceptance tests and only focus on integration tests. BUT I might just need to fix the failing acceptance test.
- ~~figure out why the color contrast checks are failing for so many tests; is this a race condition, or a state where color-contrast isn't required?~~ this is because the ACT rules were turned on and they include AAA rules. 
- figure out how to validate the syntax
- confirm the output approach (maybe we want to send the results to an s3 bucket instead of a folder in the repo?)

### :link: External links

Figma file: [if it applies]

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


MelSumner:
Proof of concept, closing PR.

.github/workflows/ci-components-nightly.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Install Dependencies
+        run: yarn install --immutable
+      - name: Run Tests
+        run: yarn run test:a11y
+        working-directory: packages/components
+      - name: Archive a11y test results
+        uses: actions/upload-artifact@v3
+        with:
+          name: a11y-test-report
+          path: packages/components/test-output/a11y-test-results.json
```
MelSumner, line 37: @alex-ju I am not sure if it will format to json correctly, do you happen to know? I'll hunt for other examples in the meantime.

alex-ju: my understanding of `actions/upload-artifact` is that it takes a local or temporary file and uploads it to GitHub as an artifact of the current flow. what I'm not sure about is how do we get to create `packages/components/test-output/a11y-test-results.json` – running `yarn run test:a11y` only outputs the tests to the terminal, so we might need to update the action to something like this `run: yarn run test:a11y > a11y-test-results.txt`

In regards to the format, I'm wondering if we need to take [the middleware approach](https://github.com/ember-a11y/ember-a11y-testing#test-middleware) to be able to generate it in a json format

MelSumner, line 37:
```suggestion
          path: packages/components/test-output/a11y-test-results.txt
```

MelSumner: The qUnit website had some hints, but I've changed it to .txt for now so we can at least see if the job works.

alex-ju (approved):
Happy to give it a try, we can always revert

PR #1406, opened 2023-06-08
WIP: update test config for a11y tests

MelSumner:
### :pushpin: Summary

If merged, this PR tweaks the configuration of our global a11y tests. 

- updates to the latest version of ember-a11y-testing
- updated the package.json to include a run option that explicitly enables a11yAudits (`test:ember:a11y`)
- updated the ci-components.yml (CI config to include a test run (not included in ember try) that are the tests + the a11y audit
- removes ACT from the `setRunOptions` because it was including AAA tests
- excludes (for now), showcase pages and the icon sprite
- a few other minor tweaks to accommodate some of the axe-core magic

This option took approx 16 minutes to run, and there should be 191 failed tests.

This option demonstrates that our tests would need to be tidied up- the invocations in the tests should more accurately reflect a complete invocation. 

For example:

```js
test('it should render with a CSS class that matches the component name', async function (assert) {
    await render(hbs`<Hds::Table::Th id="data-test-table-th" />`);
    assert.dom('#data-test-table-th').hasClass('hds-table__th');
  });
```

This tests that the component renders and that it has the correct CSS class. However, it fails an a11y audit because it does not have an accessible name, and TH elements are required to have an accessible name. 

To fix, the invocation would need to be updated slightly:

```js
test('it should render with a CSS class that matches the component name', async function (assert) {
    await render(hbs`<Hds::Table::Th id="data-test-table-th">Artist</Hds::Table::Th>`);
    assert.dom('#data-test-table-th').hasClass('hds-table__th');
  });
```

### :link: External links

Figma file: [if it applies]

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.github/workflows/ci-components.yml
+++ b/.github/workflows/ci-components.yml
@@ -36,6 +36,27 @@ jobs:
         working-directory: packages/components
         env:
           PERCY_TOKEN: ${{ secrets.PERCY_TOKEN_COMPONENTS }}
+  test-with-a11y:
+    name: "Tests with A11y"
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+      - name: Install Node
+        uses: actions/setup-node@64ed1c7eab4cce3362f8c340dee64e5eaeef8f7c # v3.6.0
+        with:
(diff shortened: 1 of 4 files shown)


MelSumner:
Proof of concept, closing PR

PR #1412, opened 2023-06-09

hashibot-hds:
If merged, this PR would add the ampersand icon to the icon set.


Dhaulagiri (approved):
I went ahead and added a changeset for this as well

PR #1415, opened 2023-06-12
`Hds::Table` - Improve columns `@width` handling and documentation

didoo:
### :pushpin: Summary

This PR is the result of a [thread in Slack](https://hashicorp.slack.com/archives/C7KTUHNUS/p1682372763007829?thread_ts=1682369934.519929&cid=C7KTUHNUS) and a follow-up meeting and discussion with @emailnitram about how the `HDS::Table` component behaves when the user needs a predefined column width, in the specific case where the table has a lot (**a lot**) of columns.

Now, this is more of an edge case, but still a valid case in the context of the product needs ([see here](https://github.com/hashicorp/atlas/blob/main/frontend/atlas/app/templates/v2/organization/explorer/index.hbs)). 
Essentially the need is:
- to have a table that contains a large number of columns (hence the content can't fit in the container) 
- for some of these columns be able to apply a specific width and have the browser respect that width (without having to specify the width for all the columns)
- have a parent container for the table that allows horizontal scrolling

After some exploration in the showcase, I have realized that:
- the way in which the browser renders a table is not always what I was expecting (probably based on heuristic more than strict rules)
- the combination of `table-layout: auto|fixed` + column `width: auto|px|%` leads to different behaviour of the browser layout engine
- it's not possible to satisfy the requirements described above, unless we set **both** the `width` and the `min-width` of the columns (this forces the table to respect the width and not try to collapse the content anyway)

_Notice: using just `table-layout: fixed` won't do the trick, because in that case all the columns that don't have an explicit width will automatically receive the same residual width, no matter what their content is)._

At the end I have left only the "meaningful" examples of combinations in the showcase, and added a couple of useful examples in the "how to use" section of the documentation, to cover the topic of "scrollable tables".

In this PR I have:
- updated the `Table` showcase to include variants/use cases for different layouts
    - combination of table-layout + column widths (px/%)
    - large table with overflow scrollable
- updated `Th/ThSort` subcomponents to add also `min-width` inline style
    - this caused a few visual regression in the existing table, but technically they are correct (now the columns width is exactly as declared by the user, while before it was not)
- updated the "How to use" section of the Helios website adding a couple of examples of `Scrollable table` 
- updated a couple of integration tests for `@width` argument

👉 👉 👉 **Previews**:
- **website**: https://hds-website-git-hds-table-fixed-columns-width-hashicorp.vercel.app/components/table?tab=code#scrollable-table
- **showcase**: https://hds-showcase-git-hds-table-fixed-columns-width-hashicorp.vercel.app/components/table (see "Layout" section)

screenshot_2755
screenshot_2754

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-1829

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/orange-crews-applaud.md
+++ b/.changeset/orange-crews-applaud.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Hds::Table` - Changed the way in which the column `@width` defined by the user is applied
--- a/packages/components/addon/components/hds/table/th-sort.hbs
+++ b/packages/components/addon/components/hds/table/th-sort.hbs
@@ -2,7 +2,13 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<th class={{this.classNames}} aria-sort={{this.ariaSort}} {{style width=@width}} ...attributes scope="col">
+<th
+  class={{this.classNames}}
+  aria-sort={{this.ariaSort}}
+  {{style width=@width minWidth=@width}}
+  ...attributes
+  scope="col"
+>
   <button type="button" {{on "click" this.onClick}}>
     <div class="hds-table__th-sort--button-content">
       <span class="hds-typography-body-200 hds-font-weight-semibold">{{yield}}</span>
--- a/packages/components/tests/dummy/app/templates/components/table.hbs
+++ b/packages/components/tests/dummy/app/templates/components/table.hbs
@@ -503,6 +503,177 @@
 
   <Shw::Divider />
 
+  <Shw::Text::H2>Layout</Shw::Text::H2>
+
+  <Shw::Text::H3>Interaction between table layout and columns width</Shw::Text::H3>
+
+  <Shw::Text::H4>Width in <code>px</code> + Table-layout = 'auto'</Shw::Text::H4>
+
+  <Hds::Table
+    @model={{this.model.manycolumns}}
+    @columns={{array
+      (hash key="first_name" label="First Name" isSortable="true" width="200px")
+      (hash key="last_name" label="Last Name" isSortable="true" width="200px")
+      (hash key="age" label="Age" isSortable="true")
+      (hash key="email" label="Email")
+      (hash key="bio" label="Biography" width="350px")
+    }}
+  >
+    <:body as |B|>
+      <B.Tr>
+        <B.Td>{{B.data.first_name}}</B.Td>
+        <B.Td>{{B.data.last_name}}</B.Td>
+        <B.Td>{{B.data.age}}</B.Td>
+        <B.Td>{{B.data.email}}</B.Td>
+        <B.Td>{{B.data.bio}}</B.Td>
+      </B.Tr>
+    </:body>
+  </Hds::Table>
+
+  <Shw::Text::H4>Width in <code>px</code> + Table-layout = 'fixed'</Shw::Text::H4>
+
+  <Hds::Table
+    @model={{this.model.manycolumns}}
+    @columns={{array
+      (hash key="first_name" label="First Name" isSortable="true" width="200px")
+      (hash key="last_name" label="Last Name" isSortable="true" width="200px")
+      (hash key="age" label="Age" isSortable="true")
+      (hash key="email" label="Email")
+      (hash key="bio" label="Biography" width="350px")
+    }}
+    @isFixedLayout={{true}}
+  >
+    <:body as |B|>
+      <B.Tr>
+        <B.Td>{{B.data.first_name}}</B.Td>
+        <B.Td>{{B.data.last_name}}</B.Td>
+        <B.Td>{{B.data.age}}</B.Td>
+        <B.Td>{{B.data.email}}</B.Td>
+        <B.Td>{{B.data.bio}}</B.Td>
+      </B.Tr>
+    </:body>
+  </Hds::Table>
+
+  <Shw::Text::H4>Width in <code>%</code> + Table-layout = 'auto'</Shw::Text::H4>
+
+  <Hds::Table
+    @model={{this.model.manycolumns}}
+    @columns={{array
+      (hash key="first_name" label="First Name" isSortable="true" width="20%")
+      (hash key="last_name" label="Last Name" isSortable="true" width="20%")
+      (hash key="age" label="Age" isSortable="true")
+      (hash key="email" label="Email")
+      (hash key="bio" label="Biography" width="35%")
+    }}
+  >
+    <:body as |B|>
+      <B.Tr>
+        <B.Td>{{B.data.first_name}}</B.Td>
+        <B.Td>{{B.data.last_name}}</B.Td>
+        <B.Td>{{B.data.age}}</B.Td>
+        <B.Td>{{B.data.email}}</B.Td>
+        <B.Td>{{B.data.bio}}</B.Td>
+      </B.Tr>
+    </:body>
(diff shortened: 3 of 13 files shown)


.changeset/orange-crews-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
```
didoo, line 2: I have set the semver change to _patch_ but happy to hear alternative opinions.

packages/components/addon/components/hds/table/th-sort.hbs
```diff
@@ -2,7 +2,13 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<th class={{this.classNames}} aria-sort={{this.ariaSort}} {{style width=@width}} ...attributes scope="col">
+<th
+  class={{this.classNames}}
+  aria-sort={{this.ariaSort}}
+  {{style width=@width minWidth=@width}}
```
didoo, line 8: FYI from my experiments I've seen that we can't set just the `min-width` or in some cases the table will not respect the desired width, we have to declare both

packages/components/tests/dummy/app/templates/components/table.hbs
```diff
@@ -503,6 +503,177 @@
...
+        <B.Td>{{B.data.first_name}}</B.Td>
+        <B.Td>{{B.data.last_name}}</B.Td>
+        <B.Td>{{B.data.age}}</B.Td>
+        <B.Td>{{B.data.email}}</B.Td>
+        <B.Td>{{B.data.bio}}</B.Td>
+      </B.Tr>
+    </:body>
+  </Hds::Table>
+
+  <Shw::Text::H4>Width in <code>px</code> + Table-layout = 'fixed'</Shw::Text::H4>
```
didoo, line 533: I know this example looks "broken" but I've left it on purpose to show what happens when a fixed layout is applied to the table, with only some columns having a width declared.

didoo: (Maybe I should add a comment in the page to make this explicit?)

alex-ju (approved):
Code looks good, did some cross-browser testing as well. Left a few suggestions on the docs. In terms of versioning, it looks like a fix/patch to me – we're not altering the API, we're just fixing the behavior we expect to happen by setting a custom `width`.

packages/components/addon/components/hds/table/th-sort.hbs
MelSumner: I would believe this just because of the other weird CSS things that apply to table elements. 😞

didoo: @MelSumner can you expand? I don't understand the comment

MelSumner (approved):
I still don't like that we're supporting inline styles but have no objections otherwise.

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -380,6 +380,95 @@ To create a column that has right-aligned content, set `@align` to `right` on bo
 </Hds::Table>
 ```
 
+### Scrollable table
+
+As a general principle, tables should contain as little information as possible. Consuming a large amount of data in a tabular format leads to an intense cognitive load for the user. We recommend using functionalities like [pagination](/components/pagination), [sorting](/components/table?tab=code#sortable-table), and [filtering](/patterns/filter-patterns) to reduce this load.
```
jorytindall, line 385: > As a general principle, tables should contain as little information as possible.

I think I know what you're trying to say here, but I'm not sure if this lands with me. I read this and think "well, that doesn't make sense, a table can display potentially complex records and data which could mean a ton of information."

Maybe a better way to say this is by leading with the point about usability:

> Consuming a large amount of data in a tabular format can lead to an intense cognitive load for the user. As a general principle, care should be taken to simplify the information within a table as much as possible.

That isn't perfect, but it seems like the guidance should be around simplifying the information and making it human-readible.

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -380,6 +380,95 @@ To create a column that has right-aligned content, set `@align` to `right` on bo
...
+
+That said, there may be cases in which it's necessary to show a table with a large number of columns, and allow the user to scroll horizontally. In this case the consumer can use different approaches, depending on their context, needs and design specs.
```
jorytindall, line 387:
```suggestion
That said, there may be cases when it's necessary to show a table with a large number of columns and allow the user to scroll horizontally. In this case, the consumer can use different approaches depending on their context, needs, and design specs.
```

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -380,6 +380,95 @@ To create a column that has right-aligned content, set `@align` to `right` on bo
...
+
+Below we show a couple of examples of how a scrollable table could be implemented: use them as starting point (your mileage may vary).
+
+#### Using a container with `overflow: auto`
+
+In most cases, wrapping the table with a container that has `overflow: auto` does the trick.
+
+The table layout is `auto` (default) which means the browser will try to optimize the width of the columns to fit their different content. In some cases, this will mean the content may wrap (see the `Phone` column as an example) in which case you may want to apply a `width` to suggest to the browser to apply a specific width to a column (see the `Biography` column).
```
jorytindall, line 395:
```suggestion
The default table layout is `auto` which means the browser will try to optimize the width of the columns to fit their different content. In some cases, this will mean the content may wrap (see the `Phone` column as an example) in which case you may want to apply a `width` to apply a specific width to a column (see the `Biography` column).
```

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -380,6 +380,95 @@ To create a column that has right-aligned content, set `@align` to `right` on bo
...
+        <B.Td>{{B.data.occupation}}</B.Td>
+      </B.Tr>
+    </:body>
+  </Hds::Table>
+</div>
+```
+
+#### Using a container with `overflow: auto` and a sub-container with `width: max-content`
+
+If you want to specify the column width of some of the columns and leave some others adapt automatically to their content, but at the same time avoid the wrapping of the content within the cells, you have to introduce a second wrapping element around the table, that has its `width` set to ` max-content`.
```
jorytindall, line 432: This reads a bit weird to me, maybe consider this:

"Introducing a secondary wrapping element around the table with its `width` set to `max-content` will allow you to specify the width of some columns, leave others to adapt to their content automatically, and avoid the wrapping of content within cells."

jorytindall (approved):
Left a couple of non-blocking comments in the documentation. Some of it reads a little weird for me, but understanding that it's pretty complex content and a bit verbose.

website/docs/components/table/partials/code/how-to-use.md
didoo: @jorytindall I am going to leave `to suggest to the browser` because technically that's what's happening (hence this ticket): the browser has its own heuristics to determine if apply that width or not, depending on the content on the table. I've added a link to this page/content https://www.w3.org/TR/WD-CSS2-971104/tables.html#h-17.2 so maybe it's clearer what I mean

website/docs/components/table/partials/code/how-to-use.md
didoo: I've made a mix of the two :)

emailnitram:
thanks for this improvement! updated atlas with these changes

PR #1416, opened 2023-06-12
`Hds::Table`- Update `ThSort` Component API docs + Other tweaks

didoo:
### :pushpin: Summary

While playing with the code of the `Hds::Table` and in particular with the `ThSort` subcomponent ([context](https://hashicorp.slack.com/archives/C7KTUHNUS/p1686254730931569)) I have noticed that in the `Component API` documentation we were describing a `@isSorted` argument that in reality doesn't exist in the code implementation.

This PR removes this argument from the documentation (and the tests) and updates other parts of the documentation with small tweaks.

In this PR I have:
- removed `@isSorted` argument from the “Component API” documentation
- removed `@isSorted` argument from the integration tests
- done small tweaks to the `Component API` documentation for `Tr`/`Th`/`ThSort`/`Td`/
  - added a `yield` argument where it applied
  - removed some misleading text 

Preview: 

***

### DIFF
--- a/website/docs/components/table/partials/code/component-api.md
+++ b/website/docs/components/table/partials/code/component-api.md
@@ -74,12 +74,15 @@ The Table component itself is where most of the options will be applied. However
 
 ### Table::Tr
 
-Note: This component is not eligible to receive interactions (e.g., it cannot have an `onClick` event handler attached directly to it). Instead, an interactive element should be placed _inside_ of the `Th`, `ThSort`, or `Td` elements.
+Note: This component is not eligible to receive interactions (e.g., it cannot have an `onClick` event handler attached directly to it). Instead, an interactive element should be placed _inside_ of the `Th`, `Td` elements.
 
 This component can contain `Hds::Table::Th`, `Hds::Table::ThSort`, or `Hds::Table::Td` components.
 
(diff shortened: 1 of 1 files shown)


website/docs/components/table/partials/code/component-api.md
```diff
@@ -100,21 +103,19 @@ If the `Th` component is passed as the first cell of a table body row, `scope="r
...
+    Elements passed as children of this component are yielded inside the `<th>` element.
+  </C.Property>
   <C.Property @name="...attributes">
     This component supports use of [`...attributes`](https://guides.emberjs.com/release/in-depth-topics/patterns-for-components/#toc_attribute-ordering).
   </C.Property>
 </Doc::ComponentApi>
 
 ### Table::ThSort
 
-Note: This component is not eligible to receive interactions (e.g., it cannot have an `onClick` event handler attached directly to it). Instead, an interactive element should be placed _inside_ of the `ThSort` element.
```
didoo, line 110: @MelSumner I have removed this entirely, because it doesn't make sense in the context of the `ThShort` (just below, we tell the user the component supports an `onClick` event handler) plus this component should **not** have interactive elements as child.

PR #1419, opened 2023-06-13
Update Ember test matrix

alex-ju:
### :pushpin: Summary

Update Ember test matrix as proposed in RFC [DS-051](https://go.hashi.co/rfc/ds-051).

Adding the latest LTS version (currently 4.8)
Removing `ember-release` (this resolves in 4.12.1, which is the version that we're going to be using across repo so we already have it covered by our CI tests)

Screenshot 2023-06-13 at 12 22 44

### :link: External links

Jira ticket: [HDS-1924](https://hashicorp.atlassian.net/browse/HDS-1924)

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable

[HDS-1924]: https://hashicorp.atlassian.net/browse/HDS-1924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci-components.yml
+++ b/.github/workflows/ci-components.yml
@@ -48,7 +48,8 @@ jobs:
         try-scenario:
          # - ember-lts-3.24
           - ember-lts-3.28
-          - ember-release
+          - ember-lts-4.8
+         # - ember-release
           - ember-beta
          # - ember-canary
          # - ember-classic
--- a/.github/workflows/ci-components.yml
+++ b/.github/workflows/ci-components.yml
@@ -52,7 +52,6 @@ jobs:
          # - ember-release
           - ember-beta
          # - ember-canary
-         # - ember-classic
          # - embroider-safe
          # - embroider-optimized
 
--- a/packages/components/config/ember-try.js
+++ b/packages/components/config/ember-try.js
@@ -19,18 +19,18 @@ module.exports = async function () {
     },
     scenarios: [
       {
-        name: 'ember-lts-3.24',
-        npm: {
-          devDependencies: {
-            'ember-source': '~3.24.3',
-          },
-        },
-      },
-      {
         name: 'ember-lts-3.28',
         npm: {
           devDependencies: {
             'ember-source': '~3.28.0',
+          },
+        },
+      },
+      {
+        name: 'ember-lts-4.8',
+        npm: {
+          devDependencies: {
+            'ember-source': '~4.8.0',
           },
         },
       },
--- a/packages/components/config/ember-try.js
+++ b/packages/components/config/ember-try.js
@@ -22,7 +22,7 @@ module.exports = async function () {
         name: 'ember-lts-3.28',
         npm: {
           devDependencies: {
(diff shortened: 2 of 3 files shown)


.github/workflows/ci-components.yml
```diff
@@ -48,7 +48,8 @@ jobs:
         try-scenario:
          # - ember-lts-3.24
```
Dhaulagiri, line 49:
```suggestion
```

no need to keep this around

.github/workflows/ci-components.yml
```diff
@@ -48,7 +48,8 @@ jobs:
...
           - ember-lts-3.28
-          - ember-release
+          - ember-lts-4.8
+         # - ember-release
```
MelSumner, line 52:
```suggestion
          - ember-release
```

This should def stay in, OR we should add couple more of the 4.X releases.

.github/workflows/ci-components.yml
```diff
@@ -48,7 +48,8 @@ jobs:
...
           - ember-beta
          # - ember-canary
          # - ember-classic
```
MelSumner, line 55:
```suggestion
```

We don't need this in there at all.

packages/components/config/ember-try.js
```diff
@@ -19,18 +19,18 @@ module.exports = async function () {
     },
     scenarios: [
       {
-        name: 'ember-lts-3.24',
+        name: 'ember-lts-3.28',
         npm: {
           devDependencies: {
-            'ember-source': '~3.24.3',
+            'ember-source': '~3.28.0',
```
MelSumner, line 25:
```suggestion
            'ember-source': '~3.28.0',
```
(suggestion) Might be more accurate to use whatever the last point release of 3.28 was

MelSumner (changes requested):

.github/workflows/ci-components.yml
alex-ju: > This should def stay in

Apologies if the language in [RFC DS-051](https://docs.google.com/document/d/1u1-lTpiLN791DtXCZsNdFayt8NaDsaFSultBcec6TAA/edit#bookmark=id.dzbnfq85y6gq) was not clear. The proposition to remove `ember-release` (currently 4.12.1) is because it is already covered by our default tests (outside the `ember-try` suite), and we're committing to keeping up with it. I tried to highlight this in the PR description too. What do you think would be the value of running our tests with 4.12.1 twice?

> OR we should add couple more of the 4.X releases

The only other 4.x LTS is 4.4, and considering Ember's commitment to backward compatibility, I'm not sure of the value of testing with both 4.4 and 4.8.

I think it's important to keep our matrix down to a reasonable number, as each entry has a considerable impact on the test duration for each PR.

packages/components/config/ember-try.js
alex-ju: sure, will update to 3.28.11

PR #1420, opened 2023-06-13
Fix color for connection-gateway icon

hashibot-hds:
This PR updates the 16px `connection-gateway` icon color to use the dynamic value instead of #000.


Dhaulagiri:
Rebased this to drop the changes that were solely dropping the copywrite headers and added a changeset

PR #1423, opened 2023-06-14
`Accordion` component (HDS-2034)

KristinLBradley:
_**Status:** Ready for release_

### :pushpin: Summary
If merged, this PR will add the `Accordion` component to the design system.

- **Showcase Preview:** https://hds-showcase-git-hds-2034-accordion-component-hashicorp.vercel.app/components/accordion

### :hammer_and_wrench: Additional details
Tested in Cloud-ui, comparing to current implementation: https://hashicorp.atlassian.net/browse/HDS-2036

### :link: External links
- Jira ticket: [HDS-2034](https://hashicorp.atlassian.net/browse/HDS-2034)
- [Figma file](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/nGGr1ZOkVEPvzmq4HkASFn/HDS-Product---Components?node-id=36870%3A71031&mode=dev)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-2034]: https://hashicorp.atlassian.net/browse/HDS-2034?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/addon/components/hds/accordion/row/index.js
```diff
@@ -0,0 +1,33 @@
...
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+
+export default class HdsAccordionRowIndexComponent extends Component {
+  @tracked isOpen = this.args.isOpen ?? false;
+
+  @action
+  toggle() {
+    this.isOpen = !this.isOpen;
+  }
```
KristinLBradley, line 16: I added this to be able to set a "hds-accordion-row--is-open" class name on the component parent for more flexibility in styling the open vs. closed states.

packages/components/tests/dummy/app/templates/components/accordion.hbs
```diff
@@ -0,0 +1,185 @@
...
+  <Shw::Flex {{style gap="1.5rem"}} @direction="column" as |SF|>
+    <SF.Item @label="One row">
+      <Hds::Accordion>
+        <Hds::Accordion::Row>
+          <:toggle>Item one</:toggle>
+          <:content>
+            <Shw::Placeholder @text="generic content" @height="40" @background="#eee" />
+          </:content>
+        </Hds::Accordion::Row>
+      </Hds::Accordion>
```
KristinLBradley, line 24: I tried this syntax but it was causing the Accordion Row not to render...

```
<Hds::Accordion as |A|>
  <A.Row>
    <:toggle>Item one</:toggle>
    <:content>
      <Shw::Placeholder @text="generic content" @height="40" @background="#eee" />
    </:content>
  </A.Row>
</Hds::Accordion>
```

KristinLBradley: Never mind, figured out what I did wrong

packages/components/addon/components/hds/accordion/row/index.hbs
```diff
@@ -0,0 +1,23 @@
...
+  <:toggle as |t|>
+    <div class="hds-accordion-row__toggle">
+      <div class="hds-accordion-row__toggle-content hds-typography-body-200 hds-foreground-primary">
+        {{yield to="toggle"}}
+      </div>
+      <button
+        class="hds-accordion-row__toggle-button"
+        type="button"
+        {{on "click" t.onClickToggle}}
+        aria-label="Toggle display"
```
KristinLBradley, line 11: TODO: I need to expose this value for translation

cveigt:
I'm curious about the term `row` used to name the single container. It makes me think of a table. Have we considered `item`? So the `AccordionItem` is the single container and `Accordion` is the group of items.

KristinLBradley:
> I'm curious about the term `row` used to name the single container. It makes me think of a table. Have we considered `item`? So the `AccordionItem` is the single container and `Accordion` is the group of items.

I also considered "bar". The naming is still a WIP, not final. I'm not ready for anything like a final review yet.

packages/components/app/styles/components/accordion.scss
KristinLBradley, line 1: **Note:** The styles are in a _very_ rough state since I'm still experimenting with the design for "clickable" vs. "non-clickable" AccordionItems...

packages/components/addon/components/hds/accordion/item/index.hbs
```diff
@@ -0,0 +1,25 @@
...
+        class="hds-accordion-item__toggle-button"
+        type="button"
+        {{on "click" t.onClickToggle}}
+        aria-label={{this.ariaLabel}}
+        aria-expanded={{if this.isOpen "true" "false"}}
+        aria-controls={{this.contentId}}
+      >
+        <FlightIcon @name="chevron-down" @size="24" @isInlineBlock={{false}} />
+      </button>
+      <div class="hds-accordion-item__toggle-content hds-typography-body-200 hds-foreground-primary">
```
cveigt, line 14: In the Toggle, the class should be `display-200-semibold`.

Can we also set the color to `--token-color-foreground-strong` instead of `foreground-primary` since it'll be a heading/title?

KristinLBradley: Thanks I can look at this next.

KristinLBradley: Pushed updates

packages/components/addon/components/hds/accordion/item/index.js
```diff
@@ -0,0 +1,67 @@
...
+   */
+  get classNames() {
+    let classes = ['hds-accordion-item'];
+
+    // add a class based on the @isOpen argument
+    if (this.isOpen) {
+      classes.push('hds-accordion-item--is-open');
+    }
+
+    if (this.isClickable) {
```
KristinLBradley, line 57: Still not sure if "isClickable" is the best naming so open to other ideas.

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,139 @@
...
+
+    &.mock-focus {
+      > .hds-disclosure-primitive__toggle .hds-accordion-item__toggle {
+        @include hds-focus-ring-with-pseudo-element();
+
+        &::before {
+          box-shadow: var(--token-focus-ring-action-box-shadow);
+        }
+      }
+    }
```
alex-ju, line 51: I know the focus of this review is meant to be on the API, but this should not be needed. You can use attributes to specify on which item you want the 'mocking' to happen using `mock-state-value` and `mock-state-selector` (see [tag.hbs](https://github.com/hashicorp/design-system/blob/main/packages/components/tests/dummy/app/templates/components/tag.hbs#L39)).

It currently resolves in 'double' focus
Screenshot 2023-06-26 at 19 01 45

packages/components/addon/components/hds/accordion/item/index.hbs
```diff
@@ -0,0 +1,25 @@
+<Hds::DisclosurePrimitive class={{this.classNames}} @isOpen={{@isOpen}} @toggle={{this.toggle}} ...attributes>
```
alex-ju, line 1: Would `@onClickToggle=this.onClickToggle` bring it more in line with what we have elsewhere (the disclosure primitive, I think)? The `on` prefix, also speaks about it being an event.

alex-ju: Ah, I see we already have an `onClickToggle`, wired with the Button – do we need both? I'm not sure I follow the difference between the two.

packages/components/addon/components/hds/accordion/item/index.js
```diff
@@ -0,0 +1,67 @@
...
+  get ariaLabel() {
+    return this.args.ariaLabel ?? 'Toggle display';
+  }
+
+  /**
+   * @param isClickable
+   * @type {boolean}
+   * @default true
+   */
+  get isClickable() {
```
alex-ju, line 35: `isInteractive` may be more broad (not to tie it to a mouse interaction) – otherwise happy to stick to `isClickable` as it aligns with the existing `onClickToggle`
```suggestion
  get isClickable() {
```

alex-ju (commented):
Couldn't spot anything off with the API. As requested, I left some feedback on the naming of arguments and child components. Let me know when you'd like me to review the CSS, tests etc.

packages/components/app/styles/components/accordion.scss
KristinLBradley: Thanks. I wasn't sure how to do that. I'll take a look at the example you linked.

packages/components/addon/components/hds/accordion/item/index.js
MelSumner: +1 to `isInteractive` 
but maybe we can also include a hint that it's fullWidth? 

`isFullWidthInteractive` or something similar?

packages/components/app/styles/components/accordion.scss
KristinLBradley: Pushed up some fixes.

packages/components/addon/components/hds/accordion/item/index.js
KristinLBradley: The "interactive" wording is good I think. I can use that.

You don't think `isFullWidthInteractive` is a bit long?

Maybe I'll try changing it to just `isInteractive` to start. Then if that still doesn't seem clear enough I can change it again before release.

MelSumner: I will _always_ vote for more clarity over terseness, heh. Totally fine with whatever we decide though!

KristinLBradley: Ok well I updated it to `isInteractive` for now. We can think about it a bit more and change again before release if folks don't think it's clear enough.

packages/components/addon/components/hds/accordion/item/index.hbs
KristinLBradley: I used the `@toggle` to change the tracked `@isOpen` state on the item so that I can update the class name when the `onClickToggle` on the button is called.

Is there better naming I could use to make that clearer? Or perhaps some other way to do it that's better?

packages/components/tests/integration/components/hds/reveal/index-test.js
```diff
@@ -118,7 +118,7 @@ module('Integration | Component | hds/reveal/index', function (hooks) {
   test('it should spread all the attributes passed to the component on the element', async function (assert) {
     await render(
       hbs`
-      <Hds::Reveal @text="More options" id="test-reveal" class="my-class" data-test1 data-test2="test" id="test-reveal">
+      <Hds::Reveal @text="More options" id="test-reveal" class="my-class" data-test1 data-test2="test">
```
KristinLBradley, line 121: Came across this duplicate id so I just went ahead and deleted it.

KristinLBradley:
> 



> Couldn't spot anything off with the API. As requested, I left some feedback on the naming of arguments and child components. Let me know when you'd like me to review the CSS, tests etc.

Thanks, I think it's at the stage now where more in depth code feedback would be helpful.

packages/components/tests/dummy/app/templates/components/accordion.hbs
```diff
@@ -0,0 +1,330 @@
...
+          </:content>
+        </A.Item>
+      </Hds::Accordion>
+    </SF.Item>
+  </Shw::Flex>
+
+  <Shw::Text::H3>AccordionItem @isInteractive</Shw::Text::H3>
+
+  <Shw::Flex {{style gap="1.5rem"}} @direction="column" as |SF|>
+    <SF.Item @label="Default, toggleContainsInteractive=true">
```
didoo, line 308: `toggleContainsInteractive` ?

packages/components/tests/dummy/app/templates/components/accordion.hbs
```diff
@@ -0,0 +1,330 @@
...
+          <:toggle>Item one</:toggle>
+          <:content>
+            <Shw::Placeholder @text="generic content" @height="40" @background="#eee" />
+          </:content>
+        </A.Item>
+      </Hds::Accordion>
+    </SF.Item>
+  </Shw::Flex>
+
+  <Shw::Text::H3>AccordionItem @isInteractive</Shw::Text::H3>
```
didoo, line 305:
```suggestion
  <Shw::Text::H3>isInteractive</Shw::Text::H3>
```

packages/components/tests/integration/components/hds/accordion/index-test.js
```diff
@@ -0,0 +1,176 @@
...
+      </Hds::Accordion>
+    `);
+    assert.dom('.hds-accordion .hds-accordion-item').exists({ count: 2 });
+  });
+
+  test('it renders the passed in content in the Accordion Item', async function (assert) {
+    await render(hbs`
+      <Hds::Accordion as |A|>
+        <A.Item>
+          <:toggle><strong>Item one</strong></:toggle>
```
didoo, line 41: suggestion: since we're just testing the rendering, we don't need to use tags to target the elements, simply use DOM IDs. see for example here: https://github.com/hashicorp/design-system/blob/main/packages/components/tests/integration/components/hds/side-nav/base-test.js#L21-L42

packages/components/tests/dummy/app/templates/components/accordion.hbs
```diff
@@ -0,0 +1,330 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+{{page-title "Accordion Component"}}
```
didoo, line 6: general consideration: maybe I would try to see if makes sense to showcase the "base elements" in isolation too, similar to what we do for example for the [Pagination](https://hds-showcase.vercel.app/components/pagination), [Tabs](https://hds-showcase.vercel.app/components/tabs),

didoo (commented):
@KristinLBradley I've done a first pass, and added a bunch of comments

overall I think is **very** good, we just have to iron out a few details

packages/components/app/styles/components/accordion.scss
KristinLBradley: Sorry I was confusing myself about where the focus was coming from. I pushed up another fix that should be correct now.

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,116 @@
...
+.mock-focus {
+  .hds-accordion-item__button { opacity: 0.2; }
+}
+
+.hds-accordion-item__button {
+  @include hds-focus-ring-with-pseudo-element();
+  position: static;
+  padding: 0;
+  background: transparent;
+  border: 0;
```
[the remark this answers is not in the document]
KristinLBradley: I was thinking perhaps I should add a separate class. Was also thinking I should use a transparent border but wanted to look more closely at the overall height of the component at the same time. I'll work on refactoring all that...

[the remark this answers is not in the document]
KristinLBradley: I updated & refactored class names, borders, and where the box-shadow is applied to fix issues with heights, etc.

packages/components/addon/components/hds/accordion/item/index.hbs
```diff
@@ -0,0 +1,25 @@
...
+    <div class="hds-accordion-item__toggle">
+      <Hds::Accordion::Item::Button
+        @isOpen={{t.isOpen}}
+        @onClickToggle={{t.onClickToggle}}
+        @contentId={{this.contentId}}
+        @ariaLabel={{this.ariaLabel}}
+        @isInteractive={{@isInteractive}}
+      />
+
+      <div
```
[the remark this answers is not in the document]
KristinLBradley: If I change the z-index of the other content to overlap the interactive area then the functionality won't work when you click that overlapping area.

packages/components/addon/components/hds/accordion/item/index.js
KristinLBradley: "expandInteractiveArea" is similar to how I usually describe the technique I'm doing with the button so that could be good. Anyway I'll think about it more and we can wait to make a decision.

packages/components/addon/components/hds/disclosure-primitive/index.js
```diff
@@ -14,6 +14,10 @@ export default class HdsDisclosurePrimitiveComponent extends Component {
   @action
   onClickToggle() {
     this.isOpen = !this.isOpen;
+
+    if (this.args.toggle && typeof this.args.toggle === 'function') {
+      this.args.toggle();
+    }
```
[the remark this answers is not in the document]
KristinLBradley: Updated, thanks

packages/components/addon/components/hds/accordion/item/index.js
```diff
@@ -0,0 +1,67 @@
...
+  /**
+   * Get the class names to apply to the component.
+   * @method classNames
+   * @return {string} The "class" attribute to apply to the component.
+   */
+  get classNames() {
+    let classes = ['hds-accordion-item'];
+
+    // add a class based on the @isOpen argument
+    if (this.isOpen) {
```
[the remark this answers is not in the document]
KristinLBradley: Updated, thanks

packages/components/addon/components/hds/accordion/item/index.hbs
KristinLBradley: Thanks I updated to remove the related unneeded code.

packages/components/app/styles/components/accordion.scss
KristinLBradley: Unresolving...

The changes I pushed up previously messed up the box-shadow position so it was only around the toggle and not around the content too so I undid that change.

So I still have the problem that the box shadow makes the AccordionItem look a bit taller since it's outside the containing element (also due to the transparent border I added to the match the gray border around the other variant). The mock-hover state is also no longer working in the Showcase. So I need to try to figure out how to fix those items...

packages/components/tests/dummy/app/templates/components/accordion.hbs
KristinLBradley: Left over, will update

packages/components/tests/dummy/app/templates/components/accordion.hbs
```diff
@@ -0,0 +1,330 @@
...
+
+<Shw::Text::H1>Accordion</Shw::Text::H1>
+
+<section data-test-percy>
+  <Shw::Text::H2>Content</Shw::Text::H2>
```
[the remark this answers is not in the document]
KristinLBradley: Adding a basic example for now, thanks.

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,116 @@
...
+  outline: none;
+
+  &:hover { cursor: pointer; }
+
+  // expand button target to cover entire AccordionItem Toggle block
```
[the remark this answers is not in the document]
KristinLBradley: Moved this and updated the comment

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,116 @@
...
+}
+
+// ACCORDION COMPONENT (wrapper)
+
+.hds-accordion {
+  display: flex;
+  flex-direction: column;
+  gap: 12px;
+
+  // override default .hds-disclosure-primitive style (Q: Should primitive be refactored?)
```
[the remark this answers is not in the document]
alex-ju: I think this is a good opportunity to address this in the disclosure primitive. Thank you for scouting the codebases Cristiano! It looks like we only have one instance in the consumers' codebase that we can investigate to understand if it requires adjustments (at a glance, it seems to be an accordion, so likely migrated to this component at some point).

packages/components/addon/components/hds/accordion/item/index.hbs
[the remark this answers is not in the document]
didoo: Yeah, right! They're siblings, not parent/child. Mmm 🤔

The only solution would be to add an `onClick` event listener to the `hds-accordion-item__toggle-content` which is.. not great.  But still, I think consumers sooner or later will complain that text in the toggle is not copyable (may contain some information that the end user wants to copy&paste, like the name of a resource, a UUID, a time/date, etc.

Let's see what @cveigt thinks.

packages/components/addon/components/hds/accordion/item/index.hbs
```diff
@@ -0,0 +1,25 @@
...
+        class="hds-accordion-item__toggle-content hds-typography-display-200 hds-foreground-strong hds-font-weight-semibold"
```
didoo, line 13: while doing some tests, I realized this element is not 100% in width.
I _think_ it should, otherwise consumers will not be able to achieve something like this:
screenshot_2818

packages/components/tests/dummy/app/templates/components/accordion.hbs
```diff
@@ -0,0 +1,348 @@
...
+      </Hds::Accordion>
+    </SF.Item>
+
+    <SF.Item @label="With interactive content in toggle and in toggled content">
+      <Hds::Accordion as |A|>
+        <A.Item @isInteractive={{false}}>
+          <:toggle>
+            Related text
+            <a href="https://www.hashicorp.com/">A link</a>
+            <Hds::Button @text="HDS Button" @size="small" />
```
didoo, line 149: suggestion:
maybe this could be made more similar to something like this?
screenshot_2819

reference: 
screenshot_2818

didoo: 2 cents tip: trying to have examples that are more similar to production at times helps to uncover limitations in the component that are not obvious at first sight, for example the width not 100% (see comment above)

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,112 @@
...
+  margin: -1px; // offset border size
+  padding: 0;
+
+  &:hover { cursor: pointer; }
+
+  // entire toggle area is interactive
+  &.hds-accordion-item__button--is-not-interactive {
+    position: static;
+    background: transparent;
+    border: 1px solid transparent;
```
didoo, line 74: Yeah, I still see issues with the border making the code much more convoluted. 

@MelSumner do you think would be acceptable to have no borders for this specific button, that has some strict design implementations? (the button is 24px, the icon is 24px, but it has an inner border too, that we can "emulate" but we can't have both at 24px with a 1px transparent border)

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,112 @@
...
+      inset: 0;
+    }
+  }
+
+  // only chevron button area is interactive
+  &.hds-accordion-item__button--is-interactive {
+    @include hds-button();
+    @include hds-button-color-secondary();
+
+    &:focus { border-color: var(--token-color-border-strong); }
```
didoo, line 91: Are you sure you need this? In which case?

didoo (commented):
Did a second pass and added a few more comments. Happy that most of the other issues have been closed, too.

packages/components/tests/integration/components/hds/accordion/index-test.js
didoo: maybe this was lost?

packages/components/addon/components/hds/accordion/item/index.hbs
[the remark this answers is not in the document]
cveigt: But when placing information that is meant to be copied, consumers can use the interactive variant and add a copy button.

packages/components/addon/components/hds/accordion/item/index.js
```diff
@@ -0,0 +1,67 @@
...
+    return this.args.ariaLabel ?? 'Toggle display';
+  }
+
+  /**
+   * @param isInteractive
+   * @type {boolean}
+   * @default true
+   */
+  get isInteractive() {
+    return this.args.isInteractive ?? true;
```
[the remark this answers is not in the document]
cveigt: This doesn't sound very clear to me, as it could be interpreted as confused with the content, which is what expands.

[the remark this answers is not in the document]
didoo: I see your point. Then my candidate is `isFullAreaInteractive` :)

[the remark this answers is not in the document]
KristinLBradley: Just the toggle block area is interactive though. So could using `isFullAreaInteractive` still be potentially confusing?

packages/components/app/styles/components/accordion.scss
KristinLBradley: I'm going to work on more refactors to the styles today because I just noticed some details of the designs for the 2 variations changed from the original explorations that I hadn't noticed so I'll go through and check things like this again as part of that.

KristinLBradley: I double-checked this line. I'm seeing the border turn blue after I click on the button to activate it:
image

didoo: there may be an issue with `:focus` vs `:focus-visible` (in the implementation, I mean)?

KristinLBradley: I've been trying different things and so far can't figure out the cause. I'll experiment more after addressing other issues.

I pushed up some other styles fixes to make the 2 variants match the latest mocks and also figured out how to get the hover state to display in the Showcase

packages/components/addon/components/hds/accordion/item/index.hbs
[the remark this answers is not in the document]
KristinLBradley: If @cveigt is ok with the current behavior then I'm leaning towards marking this as resolved. If consumers have complaints then we could open a new ticket to modify it to use an `onClick` instead later on.

Does that sound acceptable? @didoo

packages/components/addon/components/hds/accordion/item/index.hbs
KristinLBradley: It seems to be full width to me but I can add an explicit width style.

KristinLBradley: Sorry I was looking at the wrong element... Re-checking.

KristinLBradley: Alright, pushed a fix for the correct element this time.

packages/components/app/styles/components/accordion.scss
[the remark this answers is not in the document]
KristinLBradley: I posted in #cloud-frontend to ask them about that usage: https://hashicorp.slack.com/archives/C04TTMDBPNV/p1687975977736889

If changing the style doesn't cause them a problem then I can refactor it as part of this PR.

packages/components/app/styles/components/accordion.scss
KristinLBradley: Adding `margin: -1px;` was the only thing I could come up with. It doesn't seem to cause issues although not ideal...

packages/components/app/styles/components/accordion.scss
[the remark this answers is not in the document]
KristinLBradley: Confirmed removing it won't cause Cloud-ui an issue.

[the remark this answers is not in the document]
KristinLBradley: I'll need to update Reveal styles though although it's a minor change. Are you ok if I do that as part of this PR or should I open a new separate one?
@didoo @alex-ju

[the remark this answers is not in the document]
KristinLBradley: @alex-ju Would you like to open a separate PR for this? Or I'm fine with opening one myself.

packages/components/tests/dummy/app/templates/components/accordion.hbs
KristinLBradley: I added base elements to the Showcase and made some changes to the button component to get states, etc. working. Let me know if you'd like me to reorganize the Showcase to move any of the sections or anything.

Btw, the standalone examples of `Accordion::Item::Button` states don't work great since they require a container with position:relative to contain the pseudo element for the focus ring so let me know if I should make changes to those examples or perhaps just delete them.

packages/components/tests/dummy/app/templates/components/accordion.hbs
KristinLBradley: I updated the example to improve it. 

(I wasn't sure of a good example when I added it but had intended to improve it before releasing. Also part of why I just added that overly simple example to start though was that it bothers me to write one-off custom layout styles.)

packages/components/app/styles/components/accordion.scss
KristinLBradley: Ok this should now finally be resolved. Part of my issue earlier was I didn't realize the designs for the 2 versions (`@isInteractive=true/false`) had been updated to both use a box-shadow vs. one having a border instead.

packages/components/addon/components/hds/accordion/item/index.js
KristinLBradley: I'm now leaning toward either `isFullWidthInteractive` or `isFullAreaInteractive` but still open to other suggestions. Maybe we should take a vote?

@MelSumner @alex-ju @didoo @cveigt

packages/components/app/styles/components/accordion.scss
KristinLBradley: This is my main remaining issue I think now. Still haven't been able to figure it out but I will focus on it more next.

packages/components/tests/integration/components/hds/accordion/index-test.js
KristinLBradley: I replaced some class name selectors with ids. Not sure if there were more you thought should be changed.

packages/components/app/styles/components/accordion.scss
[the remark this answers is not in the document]
alex-ju: I'm fine with the refactor being made in this PR – I think you have a better understanding of the required change.

packages/components/addon/components/hds/accordion/item/index.js
KristinLBradley: Another naming option to consider:

`isNestedInteractive` Default would be `false`

packages/components/app/styles/components/accordion.scss
alex-ju: The border of the button _should_ be blue when focussed, to indicate the active element on the page (just as the button component does). What is missing is the additional light-blue border that resides in `hds-button-state-focus`.

The way I see it, the `hds-accordion-item__button` class should look like this (happy to provide a patch if more helpful).

```scss
.hds-accordion-item__button {
  padding: 0;

  &:hover { cursor: pointer; }

  // entire toggle area is interactive
  &.hds-accordion-item__button--is-not-interactive {
    @include hds-focus-ring-with-pseudo-element();

    position: static;
    margin: -1px; // offset border size
    background: transparent;
    border: 1px solid transparent;

    // expand button target to cover entire AccordionItem Toggle block (depending on the `@isInteractive` argument)
    &::after {
      position: absolute;
      display: block;
      border-radius: 6px;
      content: "";
      inset: 0;
    }
  }

  // only chevron button area is interactive
  &.hds-accordion-item__button--is-interactive {
    @include hds-button();

    &:focus,
    &.mock-focus {
      @include hds-button-state-focus();
    }

    // `hds-button-color-secondary` determines the focus color and needs to be placed after `hds-button-state-focus`
    @include hds-button-color-secondary();
  }

  // animate chevron icon
  &.hds-accordion-item__button--is-open {
    .flight-icon-chevron-down {
      transform: rotate(-180deg);
    }
  }

  .flight-icon-chevron-down {
    @media (prefers-reduced-motion: no-preference) {
      transition: transform 0.3s;
    }
  }
}
```

alex-ju: ### Before 
accordion-focus-before

### After
accordion-focus-after

KristinLBradley: Thank you! @alex-ju Let me see...

KristinLBradley: Works great! I pushed up the changes. @alex-ju

packages/components/tests/dummy/app/templates/components/accordion.hbs
KristinLBradley: Made a number of follow on improvements to Showcase after pairing session w/ @didoo

packages/components/app/styles/components/accordion.scss
KristinLBradley: @didoo Do you have any ideas for simplifications/improvements or do you think it is ok to leave as is?

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,116 @@
...
+//
+// ACCORDION
+//
+
+@use "../mixins/button" as *;
+@use "../mixins/focus-ring" as *;
+
+:root {
+  --token-accordion-item-padding: 16px;
+  --token-accordion-item-padding-right: 12px;
```
alex-ju, line 15: Is the intention to move these to the design tokens package? If not, we seem to be using SASS variables for other components, potentially a topic for our engineering sync to decide the approach forward.

alex-ju (approved):
Looking good! Don't forget to add a changelog entry

packages/components/addon/components/hds/accordion/item/button.js
```diff
@@ -0,0 +1,30 @@
...
+  get classNames() {
+    let classes = ['hds-accordion-item__button'];
+
+    // add a class based on the @isOpen argument
+    if (this.args.isOpen) {
+      classes.push('hds-accordion-item__button--is-open');
+    }
+
+    // if toggle is NOT interactive, then button IS
+    if (this.args.isInteractive === false) {
```
didoo, line 23: I find this confusing. We're testing against an exact value (`false`) which means `true` and `undefined` fall under the same category, which is weird.

What about having a default value (that I imagine is the `true` condition, given that it's `true` in the parent element) via a getter, and avoid this strange conditional?

Something like this:
```
get parentContainsInteractive() {
    return this.args.parentContainsInteractive ?? [true|false]; // depends on how you define the default in the parent
}
```
and then

```
if (this.parentContainsInteractive) {
    classes.push('hds-accordion-item__button--parent-contains-interactive');
} else {
    classes.push('hds-accordion-item__button--parent-does-not-contain-interactive');
}
```

packages/components/app/styles/components/accordion.scss
didoo: +1 in converting these to Sass variables

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,116 @@
...
+.hds-accordion {
+  display: flex;
+  flex-direction: column;
+  gap: 12px;
+}
+
+// ACCORDION ITEM COMPONENT (nested child)
+
+.hds-accordion-item {
+  width: 100%;
```
didoo, line 29: area you sure you need this? if I comment it out, I don't see differences 🤔

packages/components/app/styles/components/accordion.scss
didoo: I think is OK as it is now. It's clear enough now that we have split the "logic" under two clear class names. Plus, I don't think there's an easy solution that would simplify even more the code. It's the best one can do given the constraints 👍

didoo (commented):
added (probably) the last comments. I'll wait that you rename the `isInteractive` arguments and related class names, and I can do a final pass (and approve)

packages/components/addon/components/hds/accordion/item/index.js
KristinLBradley: Thanks for weighing in on naming. Going to go with `containsInteractive` from Ambassador votes and our overall team feedback.

packages/components/app/styles/components/accordion.scss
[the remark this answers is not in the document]
KristinLBradley: The changes were small so have been made in this branch.

packages/components/app/styles/components/accordion.scss
KristinLBradley: I will convert to Sass variables, thanks.

packages/components/app/styles/components/accordion.scss
KristinLBradley: Oh, I probably don't now. I'll remove, thanks.

packages/components/tests/dummy/app/templates/components/accordion.hbs
```diff
@@ -0,0 +1,392 @@
...
+    </SG.Item>
+  </Shw::Grid>
+
+  <Shw::Divider @level={{2}} />
+
+  <Shw::Text::H3>Complex</Shw::Text::H3>
+
+  <Shw::Text::H4>With nested Accordions</Shw::Text::H4>
+
+  <Shw::Flex {{style gap="2rem"}} as |SF|>
```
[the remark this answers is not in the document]
KristinLBradley: Ok I'm adding an additional Accordion. It was just one Accordion with 2 items originally.

packages/components/tests/dummy/app/templates/components/accordion.hbs
```diff
@@ -0,0 +1,392 @@
...
+            />
+          </div>
+        </SF.Item>
+      </Shw::Flex>
+    {{/each}}
+  {{/let}}
+
+  <Shw::Divider />
+
+  <Shw::Text::H2>Options</Shw::Text::H2>
```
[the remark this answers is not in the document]
KristinLBradley: I'm moving it, thanks.

packages/components/addon/components/hds/accordion/item/button.js
KristinLBradley: I'm fixing the issue with the logic but I don't think it needs a getter as we're setting a default in the AccordionItem parent. (See what you think after I push up the changes.)

cveigt:
This looks great Kristin. I only have a few minor comments.
- The containsInteractive variants shouldn't have elevation; instead, they should have border primary applied.

Screen Shot 2023-06-30 at 5 28 11 PM

- I noticed a minor discrepancy that I couldn't catch why is happening. The button's size when `containsInteractive` is slightly bigger than in the Figma component. I wonder if it could be something in the accordion item button but I couldn't find any extra padding or incorrect sizing on the button or icon.

Screen Shot 2023-06-30 at 5 32 22 PM

- Can we make the chevron's fill color always be foreground primary, not only when `containsInteractive`?

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,122 @@
...
+
+//
+// ACCORDION
+//
+
+@use "../mixins/button" as *;
+@use "../mixins/focus-ring" as *;
+
+$accordion-item-padding: 16px;
+$accordion-item-medium-border-radius: 6px;
```
didoo, line 14: I don't understand what `medium` means in this context.

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,122 @@
...
+$accordion-gap: 12px;
```
didoo, line 15: the fact that the vertical spacing between items and the spacing between the content and the chevron button have the same value, it's coincidental. what I suggest to to is to remove this variable and use directly the value in the two declarations below

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,122 @@
...
+    box-shadow: var(--token-surface-mid-box-shadow);
+
+    &:hover,
+    &.mock-hover {
+      box-shadow: var(--token-surface-high-box-shadow);
+    }
+  }
+
+  &.hds-accordion-item--contains-interactive {
+    border: 1px solid var(--token-color-border-primary);
```
didoo, line 41: there is still an issue with the border. I'll show you in a call (its easy to fix).
essentially, you should add a transparent border in any case, and here change only the color to primary

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,122 @@
...
+.hds-accordion-item__toggle {
+  position: relative;
+  display: flex;
+  gap: $accordion-gap;
+  align-items: center;
+  padding:
+    $accordion-item-padding
+    $accordion-item-padding
+    $accordion-item-padding
+    12px;
```
didoo, line 56: maybe a quick comment to explain why it's different?
```suggestion
    12px; // by design
```

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,122 @@
...
+
+  // entire toggle area is interactive
+  &.hds-accordion-item__button--is-not-interactive {
+    @include hds-focus-ring-with-pseudo-element();
+    position: static;
+    margin: -1px 0;
+    background: transparent;
+    border: 1px solid transparent;
+
+    // expand button target to cover entire AccordionItem Toggle block (depending on the `@containsInteractive` argument)
```
didoo, line 74:
```suggestion
    // expand button target to cover entire AccordionItem Toggle block (depending on the `@containsInteractive/@parentContainsInteractive` argument)
```

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  .flight-icon-chevron-down {
+    @media (prefers-reduced-motion: no-preference) {
+      transition: transform 0.3s;
+    }
+  }
+}
+
+// Consumer added content that apperas next to the chevron button:
```
didoo, line 113:
```suggestion
// Consumer added content that appears next to the chevron button:
```

packages/components/app/styles/components/accordion.scss
```diff
@@ -0,0 +1,122 @@
...
+    $accordion-item-padding
+    12px;
+}
+
+.hds-accordion-item__button {
+  padding: 0;
+
+  &:hover { cursor: pointer; }
+
+  .flight-icon-chevron-down { color: var(--token-color-foreground-primary); }
```
didoo, line 64: I have mixed feelings about this here: on the `parent-contains-interactive` this color is already set by the secondary button mixin, so maybe we could think of moving this under the `hds-accordion-item__button--is-not-interactive` block as `color` property, and let the SVG just inherit it via `currentColor`?

didoo (commented):
There are a few minor things and then we're good to go.

packages/components/app/styles/components/accordion.scss
KristinLBradley: I called it "medium" because that's how that size (6px) is referred to in the documentation here: https://helios.hashicorp.design/foundations/border

I'll remove it though

packages/components/app/styles/components/accordion.scss
KristinLBradley: I want to wait on changes related to this until we can get more feedback from Christian. 

(I experimented with making changes to the border, padding, and adding a margin as we discussed but it was getting very complex and still didn't seem quite right.)

cveigt: @KristinLBradley visually the accordion items look correct, as you said. I build an accordion using the showcase and the extra pixels add up when there are 3, 5, 6 ... items, so that difference matters. I'd rather reduce the padding or margin if possible.

Other than that, I don't have any more feedback. I reviewed everything again and everything works and looks pretty good.

cveigt: We can pair and try the different options you experimented with directly on the browser if you want. Maybe it's easier or faster to try and review.

KristinLBradley: @cveigt Updated the one which had a border to now use --token-surface-base-box-shadow as you messaged. I compared to Figma and the heights and spacing now look good to me but please doublecheck.

didoo (approved):
Good to go. Great job 🙌 🙌 🙌

cveigt (approved):
Thanks for making those last few updates. Everything looks great! 🚀🚀🚀

website/docs/components/accordion/partials/guidelines/guidelines.md
```diff
@@ -0,0 +1,102 @@
...
+
+![Example of toggle content](/assets/components/accordion/toggle-content-text.png =690x*)
+
+### Generic content
+
+![Example of toggle content](/assets/components/accordion/toggle-content-custom.png =690x*)
+
+### Interactive content
+
+Use `isInteractive` when nesting interactive content in the toggle. This lets users using assistive technology, access and interact with the nested element.
```
heatherlarsen, line 72: [blocking] 

```suggestion
Use `containsInteractive` when nesting interactive content in the toggle. This lets users using assistive technology, access and interact with the nested element.
```

website/docs/components/accordion/partials/guidelines/guidelines.md
```diff
@@ -0,0 +1,102 @@
...
+
+When using this property, it should be applied to all items. Avoid mixing with default containers.
```
heatherlarsen, line 74: I'm not sure "default" is the right word here either, but I don't have a good alternative at the moment. We can always update that one later if there's any confusion.

```suggestion
When using this property, it should be applied to all items. Avoid mixing with default toggles.
```

heatherlarsen (changes requested):

heatherlarsen (approved):
:shipit:

PR #1427, opened 2023-06-16
Disable the "minifyCSS" advanced option in the “components” package

didoo:
### :pushpin: Summary

I noticed that in the showcase website, Ember (or its compiler) is doing that strange thing for which is messing up the CSS order of declarations:

screenshot_2779

In this PR I have:
- disabled the `minifyCSS` advanced option in the “components” package

(I am waiting to discuss with you all if we want to merge this change, and in that case what kind of semver to use)

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/ember-cli-build.js
+++ b/packages/components/ember-cli-build.js
@@ -20,6 +20,12 @@ module.exports = function (defaults) {
     'ember-power-select': {
       theme: false,
     },
+    // we need to add this or Ember Sass compilation will mess up the generated CSS
+    minifyCSS: {
+      options: {
+        advanced: false,
+      },
+    },
   });
 
   /*


didoo:
@Dhaulagiri @MelSumner @alex-ju what do you think of this change? will impact only the "build" of the local "dummy/app" used for the showcase, or also our consumers? in that case, how will it impact them?

Dhaulagiri:
this config is local to the dummy app so it should not impact consumers of our published package

alex-ju:
as @Dhaulagiri mentioned, this should only impact our showcase app, so I'm happy to get it in. still intrigued by the CSS 'mess up', and whether it is affecting other consumers though

didoo:
> this config is local to the dummy app so it should not impact consumers of our published package
> 
> edit: [adding a link to the docs for reference](https://cli.emberjs.com/release/writing-addons/#emberclibuildjs)
> 
> > The ember-cli-build.js file in your addon is only used to configure the dummy application found in tests/dummy/. It is never referenced by applications which include the addon.

OK, then I'll move it to review. Thanks for the clarification.

> as @Dhaulagiri mentioned, this should only impact our showcase app, so I'm happy to get it in. still intrigued by the CSS 'mess up', and whether it is affecting other consumers though

Most of them have already added this config in their setup (it's specifically described as necessary step in the [Getting started for Engineers](https://helios.hashicorp.design/getting-started/for-engineers#import-component-styles) docs.

alex-ju:
> Most of them have already added this config in their setup (it's specifically described as necessary step in the [Getting started for Engineers](https://helios.hashicorp.design/getting-started/for-engineers#import-component-styles) docs.

Indeed, I remember now

alex-ju (approved):

MelSumner (commented):
IIRC this file only has an effect on the dummy app, not what's shipped to components. 

Also not sure why we would want to ship un-minified CSS, unless something about that has changed recently.

didoo:
> Also not sure why we would want to ship un-minified CSS, unless something about that has changed recently.

@MelSumner this doesn't disable minification, it disables the "advanced" (cit) minification, that unfortunately changes the order of the CSS declarations.

PR #1432, opened 2023-06-19
Browser support matrix - Added  browser support table to “Getting started / For engineers"

didoo:
### :pushpin: Summary

This PR adds a browser support table to the “Getting started / For engineers” page, according to the [[DS-054] Introduce a browser support matrix for the design system](https://docs.google.com/document/d/1NkHOt4ULP5pTLb1B4iWCr9msUCpwzG-4_BclSkgskZ4/edit) RFC

In this PR I have:
- added  browser support table to the “Getting started / For engineers” page

👉 👉 👉 **Preview**: https://hds-website-git-browser-support-matrix-update-website-hashicorp.vercel.app/getting-started/for-engineers#browser-support

screenshot_2781

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2106

***

### 👀 Reviewer's checklist:

### DIFF
--- a/website/docs/getting-started/for-engineers.md
+++ b/website/docs/getting-started/for-engineers.md
@@ -129,3 +129,14 @@ Import CSS helper classes by adding any of the following lines to the main Sass 
 ```
 
 For more examples and guidelines read [the tokens documentation](/foundations/tokens).
+
+## Browser support
+
+The code of our styles, components and icons supports the following browsers:
+
+| Browser        | Version         |
+|----------------|-----------------|
+| Chrome         | last 2 versions |
+| Safari         | last 2 versions |
+| Firefox        | last 2 versions |
+| Microsoft Edge | last 2 versions |
--- a/website/docs/getting-started/for-engineers.md
+++ b/website/docs/getting-started/for-engineers.md
@@ -132,7 +132,7 @@ For more examples and guidelines read [the tokens documentation](/foundations/to
 
 ## Browser support
 
-The code of our styles, components and icons supports the following browsers:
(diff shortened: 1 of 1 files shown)


website/docs/getting-started/for-engineers.md
```diff
@@ -129,3 +129,14 @@ Import CSS helper classes by adding any of the following lines to the main Sass
 ```
 
 For more examples and guidelines read [the tokens documentation](/foundations/tokens).
+
+## Browser support
+
+The code of our styles, components and icons supports the following browsers:
```
alex-ju, line 135: It almost feels like the causality is the other way around – browsers support features rather than the other way around – but I can't think of better wording.

alex-ju (approved):

website/docs/getting-started/for-engineers.md
didoo: @alex-ju what do you think of this alternative?
```
Our styles, components and icons are supported by the following browsers:
```

alex-ju: yep, that reads better in my opinion

didoo, line 135:
```suggestion
Our styles, components and icons are supported by the following browsers:
```

MelSumner:
We should probably make a note that this support matrix may differ from the version of Ember being used in the add-on, since that is likely to be confusing for some.

didoo:
> We should probably make a note that this support matrix may differ from the version of Ember being used in the add-on, since that is likely to be confusing for some.

Not sure what you mean. Can you make a suggestion in the PR? thanks

PR #1433, opened 2023-06-19
[WIP] Update CSS of our components to use flex `gap`

didoo:
### :pushpin: Summary

This is a follow-up of this RFC: [[DS-053] Usage of flexbox `gap` CSS property in HDS components](https://docs.google.com/document/d/16tlSX7IC406_llmDnjKfFBGVZMJ9-x3UqISdeNCNBww/edit) in which we have decided it's now safe to use the flex `gap` property in our CSS styles.

In this PR I have
- TODO

### :link: External links

Jira ticket: TODO
RFC: [[DS-053] Usage of flexbox `gap` CSS property in HDS components](https://docs.google.com/document/d/16tlSX7IC406_llmDnjKfFBGVZMJ9-x3UqISdeNCNBww/edit)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/addon/components/hds/side-nav/list/link.hbs
```diff
@@ -20,9 +20,11 @@
       <FlightIcon class="hds-side-nav__list-item-icon-leading" @name={{@icon}} />
     {{/if}}
 
-    <span class="hds-side-nav__list-item-text hds-typography-body-200 hds-font-weight-medium">
-      {{@text}}
-    </span>
+    {{#if @text}}
```
MelSumner, line 23: Doesn't text need to be here? It should be used as the value for the `aria-label` on the link if it's not visible text.

didoo: @MelSumner this is an edge case, in which consumers don't use the `@text` but will add `yield`ed content, so the actual content will be determined by them (at the moment the only case we've seen used it is with a `<time>` element

packages/components/app/styles/components/button.scss
```diff
@@ -64,14 +64,6 @@
...
 .hds-button__text {
   flex: 1 0 0;
-
-  .hds-button__icon + & {
-    margin-left: 0.375rem;
-  }
-
-  & + .hds-button__icon {
-    margin-left: 0.375rem;
-  }
```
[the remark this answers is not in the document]
didoo: The gap is added in the parent, which is the `hds-button` and I've added there the gap (see the `_button` mixin file: https://github.com/hashicorp/design-system/pull/1433/files/5a7f202bf6a067f13f46f7db0814a8f94208b7b2#diff-5b33d9668c2711a31d4a79c98ab40233762f53389870b5b834b6cd40df2960f9)

alex-ju (approved):
Regarding the versioning: I would say it's a fix/patch – as it's a code refactor with, indeed, some minor visual changes.

KristinLBradley (approved):

cveigt (approved):

PR #1437, opened 2023-06-20
Updating contextual examples of `Link::Inline`

jorytindall:
### :pushpin: Summary

If merged, this PR will add examples in the `How to use` section of the docs in which the component is inside of a block of text, similarly to how the component is [showcased](https://hds-showcase.vercel.app/components/link/inline).

I considered a couple of different options in this in an attempt to showcase the component in a relevant and semantic manner:

I initially used several different examples of body copy that was relevant to HashiCorp and to HashiCorp products, but ended up going the route of `lorem ipsum` as the generic approach seemed better in this scenario.

I initially wrapped the text in different semantic elements (`<p>` and `<span>`) and passed the typography class name to the element, but this resulted in the website text styling (`doc-markdown-p`) overriding the token in the cascade. In the end I opted to wrap the content in a `<div>` instead which matches the strategy used in the showcase.

All of these examples use the same text example for visual consistency and to more effectively highlight the different properties of the component.

Additionally, I performed some small updates in the prose for increased consistency with other areas of the documentation; namely, referring to position as `start/end` vs. `left/right`.

### :link: External links

Jira ticket: [HDS-1883](https://hashicorp.atlassian.net/browse/HDS-1883)

***

[HDS-1883]: https://hashicorp.atlassian.net/browse/HDS-1883?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/docs/components/link/inline/partials/code/how-to-use.md
+++ b/website/docs/components/link/inline/partials/code/how-to-use.md
@@ -3,9 +3,9 @@
 The most basic invocation requires some content to be passed as children and either an `@href` or `@route` argument.
 
 ```handlebars
-<div class="hds-typography-body-300">
+<span class="hds-typography-body-300">
   Lorem <Hds::Link::Inline @href="...">ipsum dolor</Hds::Link::Inline> sit amet consectetur adipiscing elit.
-</div>
+</span>
 ```
 
 ### Color
@@ -13,15 +13,15 @@ The most basic invocation requires some content to be passed as children and eit
 There are two available colors for an Inline Link: `primary` and `secondary`. The default is `primary`. To use a different color, set `@color` to `secondary`.
 
 ```handlebars
-<div class="hds-typography-body-300">
+<span class="hds-typography-body-300">
   Lorem <Hds::Link::Inline @color="primary" @href="...">ipsum dolor</Hds::Link::Inline> sit amet consectetur adipiscing elit.
-</div>
+</span>
 ```
 
 ```handlebars
-<div class="hds-typography-body-300">
+<span class="hds-typography-body-300">
   Lorem <Hds::Link::Inline @color="secondary" @href="...">ipsum dolor</Hds::Link::Inline> sit amet consectetur adipiscing elit.
-</div>
+</span>
 ```
 
 ### Icon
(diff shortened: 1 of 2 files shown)


website/docs/components/link/inline/partials/code/how-to-use.md
```diff
@@ -3,37 +3,47 @@
...
 ```
 
 ### Color
 
-There are two available colors for a Inline Link: `primary` and `secondary`. The default is `primary`. To use a different color, set `@color` to `secondary`.
+There are two available colors for an Inline Link: `primary` and `secondary`. The default is `primary`. To use a different color, set `@color` to `secondary`.
 
 ```handlebars
-<Hds::Link::Inline @color="primary" @href="...">Read tutorial</Hds::Link::Inline>
+<div class="hds-typography-body-300">
```
didoo, line 16: Probably to avoid users thinking that they should use a `<div>` for text (in general) I would use a `<span>` instead (the visual presentation results to be the same).

What do you think?

website/docs/components/link/inline/partials/code/how-to-use.md
```diff
@@ -3,37 +3,47 @@
...
 
 ### Icon position
 
-By default, if you define an icon, it‘s placed in the trailing (right) position. If you would like to position the icon in the leading (left) position, define `@iconPosition`.
+By default, if you define an icon, it‘s placed in the trailing (end) position. If you would like to position the icon in the leading (start) position, define `@iconPosition`.
 
 ```handlebars
-<Hds::Link::Inline @href="..." @icon="film" @iconPosition="leading">Watch tutorial video</Hds::Link::Inline>
+<div class="hds-typography-body-300">
+  Lorem <Hds::Link::Inline @color="primary" @icon="external-link" @iconPosition="leading" @href="...">ipsum dolor</Hds::Link::Inline> sit amet consectetur adipiscing elit.
```
didoo, line 45: What do you think of using a different icon, that works better in a leading position?

didoo (approved):
Added a couple of suggestions. Apart from that, all good for me.

website/docs/components/link/inline/partials/code/how-to-use.md
jorytindall: Yep, that works for me. I thought I had tried that initially but you're right it does present the same.

website/docs/components/link/inline/partials/code/how-to-use.md
jorytindall: The `globe` example does seem to be a better solution visually here. I'll update that.

heatherlarsen (approved):
:shipit:

website/docs/components/link/inline/partials/code/how-to-use.md
KristinLBradley: Sorry I didn't see this in time. To me a `span` doesn't make any more sense than a `div`. I most likely would have used `p` tags. Anyway I guess it doesn't make a huge difference...

PR #1438, opened 2023-06-21
`TooltipButton` - Text alignment

didoo:
### :pushpin: Summary

This ticket https://hashicorp.atlassian.net/browse/HCPIE-399 has highlighted a possible "problem" with the "button" version of our tooltip. When used in a container that expresses some for of alignment, if the text contained in the tooltip container wraps (eg in a table cell) it becomes evident that the alignment is not "correct" in the sense that it's always centered, because it's within a `<button>` element.

I suggest we add a `text-align: inherit` to the `Hds::TooltipButton` element to avoid this problem.

added use cases of a `Tooltip` in a table cell to the showcase

In this PR I have: 
- added use cases of a `Tooltip` in a table cell to the showcase

You can see the problem in action here: 

### :link: External links

Jira ticket: TODO

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


KristinLBradley:
It looks like adding `text-align: inherit;` to .`hds-tooltip-button` should do the trick.

MelSumner:
What do you think about scoping this to a table cell? It could be a patch then, and very unlikely to break anything but rather bugfix.

didoo:
> What do you think about scoping this to a table cell? It could be a patch then, and very unlikely to break anything but rather bugfix.

But the problem is not only for table cells. That was a case that made the "issue" evident. There may be other cases in the future where it's used in a parent container that has intrinsic text alignment and the text wraps (eg in a small viewport context) and the issue would be the same. I prefer to fix the problem at the root (which is that we're using a button, and we forgot that has text aligned at the center by default).

KristinLBradley (approved):

alex-ju (approved):
Looks good! Version-wise, it looks like a fix/patch to me – we're not altering the API, we're fixing the visual appearance (alignment) of `TooltipButton` to behave as expected when setting alignment on the parent element.

MelSumner (approved):

PR #1441, opened 2023-06-22
Fix position relative on focus ring mixin

alex-ju:
### :pushpin: Summary

Fix position relative on focus ring mixin

We are updating `hds-focus-ring-with-pseudo-element` to set `position: relative` to the element only when focused.

This is because `position: relative` adds the element to the level stack, meaning that it can interfere with other elements on the stack (with a `z-index` value). Below there's an example of `Hds::Tooltip` (that uses `hds-focus-ring-with-pseudo-element`) interfering with `Hds::Dropdown` (part of the `PageHeader`). All the elements that have a tooltip pear 'in front' of the dropdown.

#### Alternatives

When looking into options we determined that potential alternatives to this solution would be:
 - to set `position: relative; z-index: 1;` on `Hds::PageHeader`
   - this would resolve the immediate issue shown below but does not address the root problem and introduces even more elements on the level stack
 - look into alternatives to the pseudo-element implementation (such as multiple `box-shadow`s or so)
   -  this would reduce the number of elements on the levels stack, but would require minor visual adjustments on multiple components
 - refactor the `PageHeader` component to avoid using container queries

#### Before
Screenshot 2023-06-21 at 20 33 02

#### After
Screenshot 2023-06-21 at 20 33 27

### :link: External links

Jira ticket: [HDS-2114](https://hashicorp.atlassian.net/browse/HDS-2114)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-2114]: https://hashicorp.atlassian.net/browse/HDS-2114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/app/styles/components/tabs.scss
+++ b/packages/components/app/styles/components/tabs.scss
@@ -66,10 +66,10 @@
 
 .hds-tabs__tab-button {
   @include hds-focus-ring-with-pseudo-element(
-    $top: var(--token-tabs-tab-focus-inset),
-    $right: var(--token-tabs-tab-focus-inset),
-    $bottom: var(--token-tabs-tab-focus-inset),
-    $left: var(--token-tabs-tab-focus-inset),
+    $top: -2px,
+    $right: -6px,
+    $bottom: -2px,
+    $left: -6px,
   );
   position: static;
   display: flex;
--- a/packages/components/app/styles/mixins/_focus-ring.scss
+++ b/packages/components/app/styles/mixins/_focus-ring.scss
@@ -32,20 +32,24 @@
 }
 
 @mixin hds-focus-ring-with-pseudo-element($top: 0, $right: 0, $bottom: 0, $left: 0, $radius: 5px, $color: action) {
-  position: relative;
-  outline-style: solid; // used to avoid double outline+focus-ring in Safari (see https://github.com/hashicorp/design-system-components/issues/161#issuecomment-1031548656)
-  outline-color: transparent;
-  isolation: isolate; // used to create a new stacking context (needed to have the pseudo element below text/icon but not the parent container)
+  &:focus,
+  &.mock-focus,
+  &:focus-visible {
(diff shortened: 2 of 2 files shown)


packages/components/app/styles/components/tabs.scss
```diff
@@ -66,10 +66,10 @@
 
 .hds-tabs__tab-button {
   @include hds-focus-ring-with-pseudo-element(
-    $top: var(--token-tabs-tab-focus-inset),
-    $right: var(--token-tabs-tab-focus-inset),
-    $bottom: var(--token-tabs-tab-focus-inset),
-    $left: var(--token-tabs-tab-focus-inset),
+    $top: -2px,
+    $right: -6px,
```
alex-ju, line 70: I'm not entirely sure why these values need adjustments tbh, but these are the values that result in no visual regression for Tabs.

These two values can be converted to tokens (e.g. `--token-tabs-tab-focus-inset-horizontal`, `--token-tabs-tab-focus-inset-vertical`) and `--token-tabs-tab-focus-inset` can be removed.

MelSumner: Do you have this tracked in a JIRA issue?

alex-ju: Planning to address it in this PR, but wanted to check with the team first.

alex-ju:
Having reviewed this offline with Cristiano we decided not to go down this avenue as it is risky for cases where the component on which a pseudo-element is applied has a position already declared (e.g 'position: absolute;'). In such cases, there will be unpredictable element/layout movement.

As an alternative approach, we will update the `Hds::PageHeader` component with a `position: relative` and ask the consumers to set the `z-index` value that works for their product.

PR #1448, opened 2023-06-26
Update alert component integration test to add a11yAudit invocation with conditional

MelSumner:
### :pushpin: Summary

If merged, this PR adds a conditional a11yAudit to the alert component 

Added this to the alert integration test: 

```
import { a11yAudit, shouldForceAudit } from 'ember-a11y-testing/test-support';
```

and updated the default test: 

```
    await render(hbs`<Hds::Alert @type="inline" id="test-alert" />`);
    assert.dom('#test-alert').hasClass('hds-alert');

    if (shouldForceAudit()) {
      await a11yAudit();
    }
    assert.ok(true, 'no a11y errors found in default component invocation');
  });
```

This changes nothing in the browser

### :link: External links

Jira ticket: [HDS-2076](https://hashicorp.atlassian.net/browse/HDS-2076)
Figma file: [if it applies]

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/tests/integration/components/hds/alert/index-test.js
+++ b/packages/components/tests/integration/components/hds/alert/index-test.js
@@ -7,6 +7,7 @@ import { module, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 import { render, resetOnerror, setupOnerror } from '@ember/test-helpers';
 import { hbs } from 'ember-cli-htmlbars';
+import { a11yAudit, shouldForceAudit } from 'ember-a11y-testing/test-support';
 
 module('Integration | Component | hds/alert/index', function (hooks) {
   setupRenderingTest(hooks);
@@ -18,6 +19,11 @@ module('Integration | Component | hds/alert/index', function (hooks) {
   test('it should render the component with a CSS class that matches the component name', async function (assert) {
     await render(hbs`<Hds::Alert @type="inline" id="test-alert" />`);
     assert.dom('#test-alert').hasClass('hds-alert');
+
+    if (shouldForceAudit()) {
+      await a11yAudit();
+    }
+    assert.ok(true, 'no a11y errors found in default component invocation');
   });
 
   // TYPE


packages/components/tests/integration/components/hds/alert/index-test.js
```diff
@@ -18,6 +19,12 @@ module('Integration | Component | hds/alert/index', function (hooks) {
   test('it should render the component with a CSS class that matches the component name', async function (assert) {
     await render(hbs`<Hds::Alert @type="inline" id="test-alert" />`);
     assert.dom('#test-alert').hasClass('hds-alert');
+
+    if (shouldForceAudit()) {
+      // `&enableA11yAudit` set in the URL
+      await a11yAudit();
+    }
+    assert.ok(true, 'no a11y errors found in default component invocation');
```
Dhaulagiri, line 27: It seems like this should be inside the conditional otherwise no tests for a11y were actually run?

Dhaulagiri (commented):
What would the expectation be for this moving forward? That we would add this to all tests, portions of existing tests or something different? It's not clear to me how this gets at the underlying question of how we would integrate a11y testing into our normal test runs without ballooning the run time of the test suite.

packages/components/tests/integration/components/hds/alert/index-test.js
alex-ju: a11y errors are thrown within `a11yAudit` so I guess the logic here is: if no errors were thrown we can log a success message; I _think_ an `assert` inside a conditional statement wouldn't be valid anyway

packages/components/tests/integration/components/hds/alert/index-test.js
```diff
@@ -18,6 +19,12 @@ module('Integration | Component | hds/alert/index', function (hooks) {
   test('it should render the component with a CSS class that matches the component name', async function (assert) {
     await render(hbs`<Hds::Alert @type="inline" id="test-alert" />`);
     assert.dom('#test-alert').hasClass('hds-alert');
+
+    if (shouldForceAudit()) {
+      // `&enableA11yAudit` set in the URL
+      await a11yAudit();
```
alex-ju, line 25: It seems like setting the include/exclude in the `test-helpers.js` is not enough, we'd also need them here.

But instead of doing
```js
    await a11yAudit({
      include: [['#ember-testing-container']],
      exclude: [['.flight-sprite-container'], ['.shw-page-main']],
    });
```
in each test file, I'm thinking we can store and export the settings object we currently have in `test-helpers.js` in a variable and import it in each test.

MelSumner: That's what I'm working on now, it _should_ be enough to have them in the test helpers file, so I need to get some extra help figuring this out. I must not be doing some syntax correctly.

MelSumner:
ok so I am submitting a new branch which will allow us to run the tests globally, and include instructions on how to filter to a specific component. This will be a lot cleaner of an implementation, and then I'll go through and tidy up the test invocations. Stay tuned! HDS-2204 is the tracking issue.

PR #1449, opened 2023-06-26
Updates test helper with global a11y invocation

MelSumner:
### :pushpin: Summary

If merged, this PR adds a commented out line in our test-helper.js to globally enable `a11yAudit`

Added to test-helper.js: 

```
// uncomment this next line to turn the tests on globally
// setEnableA11yAudit(true);
```

This is so we have the proper global invocation in the test-helper file if/when we are ready to use it. 

**Note:** Something that I still don't understand- despite the run options set in the test-helper file, if I turn on the global invocation, some tests still fail because it's trying to test the icon sprite. So I'm not sure why it's not being properly excluded, still need to troubleshoot.

***

### 👀 Reviewer's checklist:

### DIFF
--- a/packages/components/tests/test-helper.js
+++ b/packages/components/tests/test-helper.js
@@ -27,7 +27,7 @@ setupGlobalA11yHooks(() => true, {
   ],
 });
 
-// uncomment this line to turn the tests on globally
+// uncomment this next line to turn the tests on globally
 // setEnableA11yAudit(true);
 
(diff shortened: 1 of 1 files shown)


alex-ju (commented):
If we go down the route of per-test invocations (as in #1448), I'm not sure this commented setup is going to be of help in the long run. We have it documented in this PR and get back to it if we ever decide to run it globally.

PR #1453, opened 2023-06-27
Update ci-components-nightly.yml

MelSumner:
- adds step to create directory for results if it does not already exist
- adds step to create a date variable
- updates the archive file so that the date is appended to the results file
- adds ability for the job to be manually triggered if desired

### DIFF
--- a/.github/workflows/ci-components-nightly.yml
+++ b/.github/workflows/ci-components-nightly.yml
@@ -38,7 +38,7 @@ jobs:
         run: mkdir -p test-output
 
       - name: Run Tests
-        run: yarn run test:a11y
+        run: yarn run test:a11y > test-output/a11y-test-results-${{ env.DATE }}.txt
         working-directory: packages/components
         continue-on-error: true
       


.github/workflows/ci-components-nightly.yml
```diff
@@ -25,15 +26,26 @@ jobs:
...
+        
+      - name: Set up date variable
+        run: echo "DATE=$(date +'%Y-%m-%d')" >> $GITHUB_ENV
+
+      - name: Create directory for test results if one does not already exist
+        working-directory: packages/components
+        run: mkdir -p test-output
+
       - name: Run Tests
         run: yarn run test:a11y
```
alex-ju, line 41: I think we need another step here to write the results into the file (instead of printing them on the console)

Trivial example
```
run: yarn run test:a11y > packages/components/test-output/a11y-test-results-${{ env.DATE }}.txt

MelSumner: Isn't that what line 45 would be doing?

MelSumner: ok, I have added it (but note the path, since we've already indicated working directory I don't think we need it again)

PR #1456, opened 2023-06-28
`Hds::Modal` – Prevent `onClose` callback function invocation when `isDismissDisabled` is `true`

alex-ju:
### :pushpin: Summary

`Hds::Modal` – Prevent `onClose` callback function invocation when `isDismissDisabled` is `true`

`isDismissDisabled` was introduced in the Modal component, based on ambassadors' feedback/request, as a method of preventing data loss in forms. In practice, this argument and the suggested implementation were never used for such a case.

[Recent feedback from HCP Packer and Hermes](https://hashicorp.slack.com/archives/C7KTUHNUS/p1679918641970839) flags that their expectation is that the `onClose` callback function is not invoked when `isDismissDisabled` is `true`, so we're updating the code to match this consumer expectation.

### :link: External links

Jira ticket: [HDS-2146](https://hashicorp.atlassian.net/browse/HDS-2146)

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable

[HDS-2146]: https://hashicorp.atlassian.net/browse/HDS-2146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/addon/components/hds/modal/index.js
+++ b/packages/components/addon/components/hds/modal/index.js
@@ -112,7 +112,11 @@ export default class HdsModalIndexComponent extends Component {
 
     // Register "onClose" callback function to be called when a native 'close' event is dispatched
     this.element.addEventListener('close', () => {
-      if (this.args.onClose && typeof this.args.onClose === 'function') {
+      if (
+        this.args.onClose &&
+        typeof this.args.onClose === 'function' &&
+        !this.isDismissDisabled
+      ) {
         this.args.onClose();
       }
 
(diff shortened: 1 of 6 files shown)


packages/components/addon/components/hds/modal/index.js
```diff
@@ -112,7 +112,11 @@ export default class HdsModalIndexComponent extends Component {
 
     // Register "onClose" callback function to be called when a native 'close' event is dispatched
     this.element.addEventListener('close', () => {
-      if (this.args.onClose && typeof this.args.onClose === 'function') {
+      if (
+        this.args.onClose &&
+        typeof this.args.onClose === 'function' &&
+        !this.isDismissDisabled
```
didoo, line 118: nitpick: maybe this is more "semantic" ? (also avoids further evaluations if dismiss is disable)
```suggestion
        !this.isDismissDisabled &&
        this.args.onClose &&
        typeof this.args.onClose === 'function'
```

didoo (approved):
I love it: the API and the code implementation for the consumers is so much neat now! 
👏👏👏

PR #1457, opened 2023-06-28
`Text` component - Part 1 (basic API)

didoo:
### :pushpin: Summary

This is the first PR for the proposed `Text` component. It implements the following APIs:
- definition of the "variant" using the tag name + `@size` argument
- definition of the HTML tag to use (generic implementation) via the `@tag` argument
- definition of the "weight to use (with basic validation of correct combination) via the `@weight` argument

Other arguments related to alignment, color, etc will be handled in follow-up PRs.

**Preview** of the components in action: https://hds-showcase-git-text-component-2023-part-1-hashicorp.vercel.app/components/text

localhost_3000_components_text

### :link: External links

Meeting notes: https://docs.google.com/document/d/12wk3VyH2aMCI-DMlrRfrmUhrQwUzRLUGydYlw3SPuiQ/
Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2168

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/text/body.hbs
+++ b/packages/components/addon/components/hds/text/body.hbs
@@ -1,0 +1,6 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<Hds::Text @group="body" @size={{this.size}} @weight={{this.weight}} @tag={{@tag}} ...attributes>{{yield}}</Hds::Text>
--- a/packages/components/addon/components/hds/text/body.js
+++ b/packages/components/addon/components/hds/text/body.js
@@ -1,0 +1,68 @@
+import Component from '@glimmer/component';
+import { assert } from '@ember/debug';
+
+// notice: only some combinations of size + font-weight are allowed (per design specs)
+// see: https://www.figma.com/file/oQsMzMMnynfPWpMEt91OpH/HDS-Product---Foundations?node-id=1262%3A9192
+
+export const DEFAULT_SIZE = '200';
+export const AVAILABLE_SIZES = [300, 200, 100];
+
+export const DEFAULT_WEIGHT = 'regular';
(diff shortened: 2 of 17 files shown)


kaxcode (approved):
This is a good starting point 👍

aklkv (approved):

packages/components/addon/components/hds/text/body.hbs
```diff
@@ -0,0 +1,6 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<Hds::Text @group="body" @size={{this.size}} @weight={{this.weight}} @tag={{@tag}} ...attributes>{{yield}}</Hds::Text>
```
aklkv, line 6: a small nit pick, `@size` attribute might be a little confusing naming, would `@variant` be better

didoo: @aklkv here's my thinking
- until now we have always avoided using `@variant` as argument for our components, **on purpose** because it's an heavy-loaded term (in Figma a component "variant" has a specific meaning) and could lead to confusion between designers and developers, and I would prefer to continue to follow this "rule"
- in general what we're referring with `100/200/300/400/500` is the **sizing** scale for that particular style (eg. `Display`) so `@size` is the thing that mostly aligns with that semantic

Also, this specific code is internal, not external-facing. Consumers will always write something like this
```
<Hds::Text::Display @size="300" ...>
```
which I personally don't think it's too confusing: I don't see them writing something like this
```
<Hds::Text::100 @size="Display" ...>
```
or 
```
<Hds::Text::Display @size="display" ...>
```

aklkv: right ok, this makes sense 👍🏻

PR #1464, opened 2023-06-29
silence ember-string.add-package deprecation

Dhaulagiri:
### :pushpin: Summary

Per https://github.com/emberjs/ember-string/issues/370#issuecomment-1518493689 silencing seems to be the best way to "fix" this issue in the short term. There's nothing broken per se and this avoids having deprecation spew in our terminals.

This can be tested locally and see the terminal is happy again.

[HDS-2189](https://hashicorp.atlassian.net/browse/HDS-2180)


jorytindall (approved):
Approving with my limited knowledge 🙃 . I tested this locally and seems to solve the problem!

PR #1471, opened 2023-07-05
Update pull_request_template.md

MelSumner:
Updates PR template to include check that a11y tests have been run by the submitter


alex-ju (approved):
Sounds good. Could we also add this to the [component checklist](https://github.com/hashicorp/design-system/blob/main/packages/components/NEW-COMPONENT-CHECKLIST.md) (potentially under testing) including the yarn command and link to it from here?

MelSumner:
@alex-ju have a new PR with it in the eng checklist: https://github.com/hashicorp/design-system/pull/1473

PR #1474, opened 2023-07-05
Tidied up component tests for `a11yAudit`

MelSumner:
### :pushpin: Summary

If merged, this PR tidies up component tests for a11yAudit (accessibility automation tests) 

I've tried to separate out each change by commit since it felt appropriate to do in this case. 

I'm going to submit the `form` and `side-nav` test improvements each as their own PRs since they each have a lot of issues. 

The benefit of this work is that once it is complete, anyone on our team could go run the [appropriate github action](https://github.com/hashicorp/design-system/actions/workflows/ci-components-nightly.yml) and get the real state of accessibility automation results, and we'll know it's an accurate reflection of where our gaps are. 

Here are the issues that I've filed as a result of this work: 

- [HDS-2214](https://hashicorp.atlassian.net/browse/HDS-2214)
- [HDS-2213](https://hashicorp.atlassian.net/browse/HDS-2213)
- [HDS-2207](https://hashicorp.atlassian.net/browse/HDS-2207)
- [HDS-2206](https://hashicorp.atlassian.net/browse/HDS-2206)

This is internal only, it should not change anything rendered to the browser. 

### :link: External links

Jira ticket: [HDS-2204](https://hashicorp.atlassian.net/browse/HDS-2204)
Figma file: [if it applies]

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-2204]: https://hashicorp.atlassian.net/browse/HDS-2204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HDS-2214]: https://hashicorp.atlassian.net/browse/HDS-2214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HDS-2213]: https://hashicorp.atlassian.net/browse/HDS-2213?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HDS-2207]: https://hashicorp.atlassian.net/browse/HDS-2207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HDS-2206]: https://hashicorp.atlassian.net/browse/HDS-2206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -35,7 +35,7 @@
     "test:ember": "ember test",
     "test:ember-compatibility": "ember try:each",
     "test:ember:percy": "percy exec ember test",
-    "test:a11y": "ENABLE_A11Y_AUDIT=true ember test",
+    "test:a11y": "ENABLE_A11Y_AUDIT=true ember test --server",
     "test:a11y-report": "ENABLE_A11Y_MIDDLEWARE_REPORTER=true ember test"
   },
   "dependencies": {
--- a/packages/components/tests/integration/components/hds/alert/index-test.js
+++ b/packages/components/tests/integration/components/hds/alert/index-test.js
@@ -85,7 +85,7 @@ module('Integration | Component | hds/alert/index', function (hooks) {
 
   test('it should render an Hds::Button component yielded to the "actions" container', async function (assert) {
     await render(
-      hbs`<Hds::Alert @type="inline" id="test-alert" as |A|><A.Button @text="I am a button" @size="small" @color="secondary" /></Hds::Alert>`
+      hbs`<Hds::Alert @type="inline" aria-labelledby="test-alert-button" id="test-alert" as |A|><A.Button @text="I am a button" @size="small" @color="secondary" id="test-alert-button"/></Hds::Alert>`
     );
     assert
       .dom('#test-alert .hds-alert__actions button')
@@ -97,7 +97,7 @@ module('Integration | Component | hds/alert/index', function (hooks) {
   });
   test('it should render an Hds::Link::Standalone component yielded to the "actions" container', async function (assert) {
     await render(
-      hbs`<Hds::Alert @type="inline" id="test-alert" as |A|><A.Link::Standalone @icon="plus" @text="I am a link" @href="#" @size="small" @color="secondary" /></Hds::Alert>`
+      hbs`<Hds::Alert @type="inline" aria-labelledby="test-alert-link" id="test-alert" as |A|><A.Link::Standalone @icon="plus" @text="I am a link" @href="#" @size="small" @color="secondary" id="test-alert-link" /></Hds::Alert>`
     );
     assert
       .dom('#test-alert .hds-alert__actions a')
--- a/packages/components/tests/integration/components/hds/breadcrumb/item-test.js
+++ b/packages/components/tests/integration/components/hds/breadcrumb/item-test.js
@@ -12,13 +12,15 @@ module('Integration | Component | hds/breadcrumb/item', function (hooks) {
   setupRenderingTest(hooks);
 
   test('it should render the component with a CSS class that matches the component name', async function (assert) {
-    await render(hbs`<Hds::Breadcrumb::Item id="test-breadcrumb-item" />`);
+    await render(
+      hbs`<ul><Hds::Breadcrumb::Item id="test-breadcrumb-item" @text="test" /></ul>`
+    );
     assert.dom('#test-breadcrumb-item').hasClass('hds-breadcrumb__item');
   });
 
   test('it should render the correct style if the @maxWidth prop is declared', async function (assert) {
     await render(
-      hbs`<Hds::Breadcrumb::Item @maxWidth="200px" @text="test" id="test-breadcrumb-item" />`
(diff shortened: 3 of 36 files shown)


didoo (commented):
@MelSumner can you wait before merging this? I want to give it some thought. Thanks

KristinLBradley (approved):
:shipit:

didoo:
@MelSumner I've given some thought about this PR (and https://github.com/hashicorp/design-system/pull/1475).

I have some doubts about the fact that we're mixing two "concerns" under the same tests declarations:
- we're testing HDS components (**and sub-components**) in terms of their basic functionality (mainly API, `props => UI`)
- we're using the same tests (with some fixes here and there) to check their accessibility


Now, in order to have the a11y tests pass, we're adding extra props or content or attributes to the component's handlebars code, even if it would not be required per se (for that specific test to pass, in terms of functionality).
Also (more problematic for me) we're doing it also for **sub-components** that were never meant to be used in that specific way. This code doesn't pass the a11y test, but the reason why that code exists in first place is to test the sub-components **in isolation**. It's not real code, it's not production code. It's code with a very specific **scope**, intent, reason.

Now, what's the problem in adding here and there some attributes, arguments, content, one could argue? To me, makes the code harder to interpret, change, maintain. As a developer, when I look at a code snippet in a test, I don't know anymore if a prop is there because is needed, in the context of that specific test, or if it's there because otherwise an a11y test fails. We lose the idea of testing all the components as "pure" elements in isolation (which gives us a lot of confidence). Now there are also side-effects that we need to consider, when evaluating a test (to write, to read, and to pass).
Adding here and there class names like `data-test-valid-failure` makes it harder for the HDS developers to add and maintain the code: requires knowledge of which tests need to pass, which tests can be ignored (but why?), etc.

For me (and I discussed this also with @alexju) the solution would be to be able to keep the two contexts independent:
- a set of test **files** for the API/functionality of the component
  - these tests would include also tests –in isolation– for subcomponents
- a set of test **files** for the A11Y test suite to test the accessibility of the component
  - these tests would verify the accessibility of component with all its subcomponents in place (in realistic ways), and could be as thourogh as we want (potentially they could be limited in number: if you think of the `Drodown`, we could have a single component with all the item variants, and test it for all the conditions)


If we take the `Dropdown` component, imagine this file structure:
```
dropdown
├─ index-test.js 
├─ list-item 
│  ├─ checkbox-test.js 
│  ├─ checkmark-test.js 
│  ├─ copy-item-test.js 
│  ├─ description-test.js 
│  ├─ generic-test.js 
│  ├─ interactive-test.js 
│  ├─ radio-test.js 
│  ├─ separator-test.js 
│  ├─ title-test.js 
├─ toggle 
│  ├─ button-test.js 
│  ├─ icon-test.js
├─ a11y-test.js // ← the file name could be different, we need to decide the convention to use
```

Now, I don't know how to set up our test suite so that we can run the tests in two groups, according to different conditions, but I suspect there must be a way. 

For example, I see that there is this library/addon:
https://github.com/ember-cli/ember-exam
that allows to run tests (or exclude tests) based on folder/file paths (using wildcards in the paths too)
https://github.com/ember-cli/ember-exam#filtering
which is what I think we could do.

This may also solve the problem of "how do we run the a11y tests only at certain times, or certain conditions" at the root, because now there is not anymore a mixing of the two concerns.

For context and clarity: I am not suggesting to change our testing approach, I am not competent enough for it. What I am suggesting is to reconsider how we reorganize the tests in our files, because this solves some of the issues I've described above, and gives us (and you expecially) **more freedom** to do proper accessibility tests on our components, without worrying of overlapping with the fuctional/API tests.

Eager to hear your thoughts on this (and @alex-ju @KristinLBradley @Dhaulagiri too)

MelSumner:
@didoo I have a response but I won't be able to get back to this until the week of the 24th. 

The TL;DR is:
- our tests should be a complete invocation that we'd expect our users to invoke
- there are some limitations on a11y automation testing, just like there is with most all tech, but that doesn't mean we shouldn't use it

I have a more indepth response and rational for future discussions, I just can't get to it at this moment but didn't want you to think I was ignoring you or had forgotten.

alex-ju:
> Eager to hear your thoughts on this (and @alex-ju @KristinLBradley @Dhaulagiri too)

At first sight, it made sense for me to alter a few tests as a way to ensure automated a11y tests pass on our components; however, after reviewing our existing tests and discussing tradeoffs with Cristiano I lean towards having a separate file for each component for more relevant a11y tests – there are a11y nuances that are currently missing from our current tests and there are lots of tests that are not relevant to be screened by automated a11y tests.

MelSumner:
based on team discussions, it seems like we're going to go in a different direction, so I'm going to close this PR

PR #1486, opened 2023-07-10
`Alert` - Refactor text styles (HDS-1757)

KristinLBradley:
### :pushpin: Summary
If merged, this PR will refactor some typography related styles to apply them as class names instead of as CSS custom properties.

- **Showcase preview:** https://hds-showcase-git-hds-1757-alert-updates-hashicorp.vercel.app/components/alert
- **Web preview:** https://hds-website-git-hds-1757-alert-updates-hashicorp.vercel.app/

### :link: External links
- Jira ticket: [HDS-1757](https://hashicorp.atlassian.net/browse/HDS-1757)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-1757]: https://hashicorp.atlassian.net/browse/HDS-1757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KristinLBradley:
**Q's:**
- Are there any other font-related styles I should convert to class names as part of this?
- Are there any tests you think would be useful to add?
- Should I document the new child component I created in the API although it's not exposed to consumers?

packages/components/addon/components/hds/alert/index.hbs
```diff
@@ -17,10 +17,10 @@
   {{/if}}
 
   <div class="hds-alert__content">
-    <div class="hds-alert__text">
+    <Hds::Alert::Text @type={{@type}}>
```
didoo, line 20: No strong opinions, just a consideration: _in theory_ you could have also used the handlebars helpers this way:
```
<div class="hds-alert__text {{if (eq @type 'compact') 'hds-typography-body-100' 'hds-typography-body-200'}}">
```
I see pros and cons in both, so whatever works better for you, I'm OK with it.

packages/components/addon/components/hds/alert/text.js
```diff
@@ -0,0 +1,26 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+
+export default class HdsAlertTextComponent extends Component {
+  /**
+   * Get the class names to apply to the component.
+   * @method Alert#classNames
```
didoo, line 11:
```suggestion
   * @method AlertText#classNames
```

didoo (approved):
👍

(left only a couple of suggestion, but not blockers)

packages/components/addon/components/hds/alert/index.hbs
MelSumner: (+1 to template conditionals)

MelSumner (approved):

packages/components/addon/components/hds/alert/index.hbs
KristinLBradley: I probably would have done something like that but it doesn't seem to be our preferred pattern for handling conditional class names so that's why I ended up creating a new child component.

KristinLBradley: I can redo it. I just didn't think that was the way we preferred.

KristinLBradley: I refactored to remove the child component and just do conditional class names in the main template. Thanks

didoo: > I can redo it. I just didn't think that was the way we preferred.

It is not. But in this very specific case, considering the pros and cons (having to create a component only for this) using inline conditionals is probably an acceptable tradeoff.

PR #1488, opened 2023-07-11
Add `Copy/Button` and `Copy/Snippet` Components

MelSumner:
### :pushpin: Summary

If merged, this PR will add the `<Hds::Copy::Button>` and `<Hds::Copy::Snippet>` components, and replace the copy functionality in the `<Hds::Dropdown::Listitem::CopyItem>` component to use the CopySnippet. 

`<Hds::Copy::Button>` 
- [Showcase](https://hds-showcase-git-melsumner-copy-component-hashicorp.vercel.app/components/copy/button) 
- [Website/Docs](https://hds-website-git-melsumner-copy-docs-hashicorp.vercel.app/components/copy/button) (#1503)

`<Hds::Copy::Snippet>`:
- [Showcase](https://hds-showcase-git-melsumner-copy-component-hashicorp.vercel.app/components/copy/snippet) 
- [Website/Docs](https://hds-website-git-melsumner-copy-docs-hashicorp.vercel.app/components/copy/snippet) (#1503)

`<Hds::Dropdown::ListItem::CopyItem>`:
- [Showcase](https://hds-showcase-git-melsumner-copy-component-hashicorp.vercel.app/components/dropdown#:~:text=generic%20content%20here-,CopyItem,-Content) (with added examples)

- copy/button uses `Hds::Button` under the hood
- copy/snippet had too many differentiations so it's a regular button element with custom styling
- both use ember-cli-clipboard; tracking an issue/PR in the repo to resolve the deprecation warnings shown in the console
- both support an error state in code (`onError`) (even though our design doesn't officially support it yet; teams still have the hook to implement their own variation)
- we have decided not to abstract common functionality _for now_ and will re-visit when we move to single-file components
- the dropdown/listitem/copyitem has been updated to use the copysnippet but it should not change anything for consumers.

Tests passing with a11y audit:
CleanShot 2023-07-26 at 10 48 58

### :link: External links

Jira ticket: [HDS-896](https://hashicorp.atlassian.net/browse/HDS-896)
[Figma file](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/rLvpmUN7ClBOAx5iPU5PvU/HDS-Product---Components?type=design&node-id=37400-71082&mode=design&t=98esQVDQXBeNxjCm-0)

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable

[HDS-896]: https://hashicorp.atlassian.net/browse/HDS-896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/addon/components/hds/copy/button/index.hbs
```diff
@@ -0,0 +1,20 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<button
+  type="button"
+  class="{{this.classNames}} {{if this.isSuccess 'is-success'}}"
```
KristinLBradley, line 7: Since the main class name for the button is added in the js file, it's probably a good idea to keep the logic for adding the 'is-success' class in the js together with it.

Also, wouldn't we want to namespace the classname as: `'hds-copy-button--is-success'`?

packages/components/addon/components/hds/copy/button/index.hbs
```diff
@@ -0,0 +1,20 @@
...
+  aria-label={{if this.isIconOnly this.text null}}
+>
+  {{#if this.isIconOnly}}
+    <FlightIcon @name="{{if this.isSuccess 'clipboard-checked' 'clipboard-copy'}}" class="hds-copy-button__icon" />
+  {{else}}
+    <div class="hds-copy-button__text hds-typography-body-100 hds-font-weight-semibold">
+      {{this.text}}
+    </div>
+    <FlightIcon @name="{{if this.isSuccess 'clipboard-checked' 'clipboard-copy'}}" class="hds-copy-button__icon" />
+  {{/if}}
```
KristinLBradley, line 19: I would use conditionals just around the div with the text to avoid repeating the code for the `FlightIcon`.

packages/components/addon/components/hds/copy/button/index.js
```diff
@@ -0,0 +1,165 @@
...
+    );
+
+    return size;
+  }
+
+  /**
+   * @param color
+   * @type {string}
+   * @default secondary
+   * @description Determines the color of copy/button to be used; the only acceptable value is `secondary`
```
KristinLBradley, line 69: If the only acceptable color is `secondary`, then why include an argument or are you planning to add other possible values before releasing?

packages/components/addon/components/hds/copy/button/index.js
```diff
@@ -0,0 +1,165 @@
...
+
+  /**
+   * @param isFullWidth
+   * @type {boolean}
+   * @default false
+   * @description Indicates that a button should take up the full width of the parent container. The default is false.
+   */
+  get isFullWidth() {
+    return this.args.isFullWidth ?? false;
+  }
```
KristinLBradley, line 102: A lot of the arguments are duplicated from `Hds::Button`. Was there a strong reason to do that instead of using that component?

packages/components/addon/components/hds/copy/button/index.hbs
```diff
@@ -0,0 +1,16 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<Hds::Button
+  class="{{if this.isSuccess 'is-success'}}"
```
MelSumner, line 6: I'm not sure this is working but I think it _should_ work this way.

packages/components/addon/components/hds/copy/button/index.hbs
MelSumner: Resolving, have refactored draft.

packages/components/tests/dummy/app/templates/components/copy/button.hbs
```diff
@@ -0,0 +1,67 @@
...
+    </SF.Item>
+  </Shw::Flex>
+
+  <Shw::Text::H2>States</Shw::Text::H2>
+
+  <Shw::Grid @columns={{5}} as |SG|>
+    {{#each @model.SIZES as |size|}}
+      {{#each @model.STATES as |state|}}
+        <SG.Item @label="{{capitalize size}} / {{capitalize state}}">
+          <Hds::Copy::Button
```
MelSumner, line 56: I am not sure how to get the correct icon to show up on a mock state- the color is correct but the icon is not: 

CleanShot 2023-07-12 at 12 25 36

packages/components/app/styles/components/copy/button.scss
KristinLBradley, line 1: I would name this file, "copy-button" otherwise it's not clear that it's different than regular button styles.

packages/components/app/styles/components/copy/button.scss
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+//
+// COPY > BUTTON
+//
+
+// Colors + States
+
+.hds-button--color-secondary {
```
KristinLBradley, line 12: I would add a new class name to this to scope the styles instead of reusing the ".hds-button--color-secondary" class name.

packages/components/app/styles/components/copy/button.scss
```diff
@@ -0,0 +1,23 @@
...
+  .hds-button__icon {
+    color: var(--token-color-foreground-action);
+  }
+
+  &.is-success,
+  &.mock-success {
```
KristinLBradley, line 18: I thought our preference for classes like these was to use BEM style naming to scope them.

packages/components/addon/components/hds/copy/button/index.hbs
```diff
@@ -0,0 +1,17 @@
...
```
KristinLBradley, line 6: Thanks for updating to use an `Hds::Button`. 

I would add a classname to this: "hds-copy-button" then use that to scope styles which are specific to this component.

I would also use BEM style for this class:
"is-success" => "hds-copy-button--is-success"

packages/components/addon/components/hds/copy/snippet/index.hbs
```diff
@@ -0,0 +1,15 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<button
+  type="button"
+  class="{{this.classNames}} {{if this.isSuccess 'is-success'}}"
```
KristinLBradley, line 7: Is there a reason to include the isSuccess logic in here vs. in the JS file? If there is, perhaps you could add a comment above the button explaining. Otherwise, I would move the logic to the JS to keep everything together.

KristinLBradley: I would also use BEM style for this classname:

"is-success" => "hds-copy-snippet--is-success"

packages/components/app/styles/components/copy/button.scss
MelSumner: since this is what we used in copy-item in the dropdown, I kept it the same. Do you think it's safe to change?

packages/components/addon/components/hds/copy/snippet/index.hbs
MelSumner: oh, please ignore the copy/snippet for now, I wanted to see if the copy/button was closer to what we were thinking before refactoring the snippet. 👍

packages/components/app/styles/components/copy/button.scss
MelSumner: oh good point, I'll revert this change.

packages/components/app/styles/components/copy/button.scss
KristinLBradley: I don't see why it wouldn't be safe to change. I'd worry more about one of the products using a classname like "is-success" and inadvertently changing the style of the HDS components.

packages/components/tests/dummy/app/templates/components/copy/button.hbs
KristinLBradley: The other states are all native HTML button states so I would maybe list it separately and just set the value of the `isSuccess` argument to true if that would work.

packages/components/addon/components/hds/copy/button/index.js
```diff
@@ -0,0 +1,108 @@
...
+   * @return {string} The "class" attribute to apply to the component.
+   */
+  get classNames() {
+    let classes = ['hds-copy-button'];
+
+    // add a class based on the @size argument
+    classes.push(`hds-copy-button--size-${this.size}`);
+
+    // add a class based on the @color argument
+    classes.push(`hds-copy-button--color-secondary`);
```
KristinLBradley, line 62: Oh, you shouldn't need to add these class names anymore since the Hds::Button takes care of adding them.

`hds-copy-button--size-${this.size}`
`hds-copy-button--color-secondary`

packages/components/addon/components/hds/copy/button/index.hbs
```diff
@@ -0,0 +1,16 @@
...
+  @size={{this.size}}
+  @isFullWidth={{@isFullWidth}}
+  @text={{@text}}
+  @icon="{{if this.isSuccess 'clipboard-checked' 'clipboard-copy'}}"
+  {{on "click" this.copyCode}}
+  ...attributes
+  @color="secondary"
+  @iconPosition="trailing"
+  @clipboardText={{@clipboardText}}
```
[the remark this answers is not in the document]
didoo: they would not be overwritten anyway, because they're _arguments_, not _attributes_ (`...attributes` spreads "attributes")

also, `@clipboardText` is not an argument for `<Hds::Button>` so should be removed

packages/components/addon/components/hds/copy/button/index.js
```diff
@@ -0,0 +1,108 @@
...
+  @tracked isSuccess = false;
+  @tracked isError = false;
+  @tracked icon = DEFAULT_ICON;
+
+  /**
+   * @param clipboardText
+   * @type {string}
+   * @description The ID of the element containing the text to be copied. If no clipboardText value is defined `isError` will be set to true.
+   */
+  get clipboardText() {
```
didoo, line 26: I think there is some confusion between `@clipboardText` and `@clipboardTarget` (see the implementation in Cloud UI - [@clipboardText](https://github.com/search?q=repo%3Ahashicorp%2Fcloud-ui%20%40clipboardText&type=code) and [@clipboardTarget](https://github.com/search?q=repo%3Ahashicorp%2Fcloud-ui%20%40clipboardTarget&type=code) - that mimics the underlying implementation in [ember-cli-clipboard](https://github.com/jkusa/ember-cli-clipboard) / [clipboard.js](https://github.com/zenorocha/clipboard.js) where
- `@clipboardText` it's used to copy text provided as an argument (not associated to any DOM element, just in memory)
- `@clipboardTarget` it's used to copy text from a DOM node, where the "target" is a path selector for that node

In our implementation, or we support only the first one, or we support both. We can't support only the second one, because it doesn't work with `<input/textarea>` elements (see comment below)

packages/components/addon/components/hds/copy/button/index.js
didoo: +1

packages/components/addon/components/hds/copy/button/index.js
```diff
@@ -0,0 +1,108 @@
...
+    return classes.join(' ');
+  }
+
+  @action
+  async copyCode() {
+    let textToCopy;
+    if (this.args.clipboardText) {
+      let clipboardTextContent = document
+        .querySelector(this.args.clipboardText)
+        .innerHTML.trim();
```
didoo, line 73: This will not work for `<input>` and `<textarea>` elements. For example, the copying of "masked input" can't be achieved with the proposed implementation:

screenshot_2916

packages/components/addon/components/hds/copy/button/index.js
```diff
@@ -0,0 +1,108 @@
...
+      textToCopy = clipboardTextContent;
+      // leaving this in while development
+      console.log(`${textToCopy} is copied to the clipboard`);
+    } else {
+      this.isError = true;
+      // leaving this in while in development
+      console.log(
```
didoo, line 80: this happens when the argument `clipboardText` is undefined, so probably we should put guards (assertions) to make sure that the consumers provide the required/necessary arguments

packages/components/addon/components/hds/copy/button/index.js
```diff
@@ -0,0 +1,108 @@
...
+
+    return classes.join(' ');
+  }
+
+  @action
+  async copyCode() {
+    let textToCopy;
+    if (this.args.clipboardText) {
+      let clipboardTextContent = document
+        .querySelector(this.args.clipboardText)
```
didoo, line 72: if the consumer provides a wrong selector, this will fail with a rather cryptic message
screenshot_2917

we should show an error (to the consumer, not the end user) that the targeted element doesn't exist

packages/components/addon/components/hds/copy/button/index.js
```diff
@@ -0,0 +1,108 @@
...
+      console.log(`${textToCopy} is copied to the clipboard`);
+    } else {
+      this.isError = true;
+      // leaving this in while in development
+      console.log(
+        `something went wrong, nothing was copied to the keyboard. isError is ${this.isError}, isSuccess is ${this.isSuccess}.`
+      );
+    }
+    // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard
+    await navigator.clipboard.writeText(textToCopy);
```
didoo, line 85: I know we support only modern browsers, but at least we should fail gracefully if `navigator.clipboard` doesn't exists (otherwise will trigger a JS error)

packages/components/addon/components/hds/copy/button/index.js
```diff
@@ -0,0 +1,108 @@
...
+
+    if (navigator.clipboard.readText) {
+      const result = await navigator.clipboard.readText();
+
+      if (result === textToCopy) {
+        this.isSuccess = true;
+        this.icon = SUCCESS_ICON;
```
didoo, line 92: suggestion: instead of setting the icon in multiple places, we could have a getter function that returns the `icon` value depending on the tracked `isSuccess/isError` values)

packages/components/tests/dummy/app/templates/components/copy/button.hbs
didoo: I can't think of a simple way to do it... 🤔 (not even a complex one, tbh)

didoo (commented):
@MelSumner did a first high-level pass on the HBS/JS code of `Copy::Button` (ignoring for now the `Copy::Snippet` as you suggested in your comment below, and everything else too: tests, showcase, docs, CSS, etc)

packages/components/tests/dummy/app/templates/components/copy/button.hbs
MelSumner: I tried this a few different ways but was unable to make it work, so I just removed it for now.

packages/components/app/styles/components/copy/button.scss
MelSumner: ok, changed it.

packages/components/addon/components/hds/copy/button/index.js
MelSumner: wait why is there a readonly input that needs copying? what's the use case here? an input field is the wrong element for this especially if it's generated.

packages/components/tests/dummy/app/templates/components/copy/button.hbs
```diff
@@ -0,0 +1,97 @@
...
+            @size={{size}}
+            @icon={{this.icon}}
+            mock-state-value={{state}}
+            @targetToCopy="#targetToCopy"
+          />
+        </SG.Item>
+      {{/each}}
+    {{/each}}
+  </Shw::Grid>
+  {{!-- <Shw::Text::H2>Status</Shw::Text::H2>
```
MelSumner, line 64: @didoo @KristinLBradley if either of you have ideas for how we could make the status show up correctly, I'm open to ideas. It will currently show the correct status color but not the correct status icon.

packages/components/tests/dummy/app/templates/components/copy/button.hbs
```diff
@@ -0,0 +1,98 @@
...
+  <Shw::Text::H2>States</Shw::Text::H2>
+
+  <Shw::Grid @columns={{4}} as |SG|>
+    {{#each @model.SIZES as |size|}}
+      {{#each @model.STATES as |state|}}
+        <SG.Item @label="{{capitalize size}} / {{capitalize state}}">
+          <Hds::Copy::Button
+            @text="Copy"
+            @size={{size}}
+            @icon={{this.icon}}
```
KristinLBradley, line 57: I don't think you need to specify the icon here.

packages/components/tests/dummy/app/templates/components/copy/button.hbs
KristinLBradley: So probably you wouldn't want to do this since it would just be used for the showcase, but you could do this:

_components/addon/components/hds/copy/button/index.js_
```
// @tracked status = DEFAULT_STATUS;
@tracked status = this.args.status ?? DEFAULT_STATUS;
```

_components/tests/dummy/app/templates/components/copy/button.hbs_
```
<Hds::Copy::Button
    @text="Copy"
    @targetToCopy="#targetToCopy"
    @status="success"
/>
```

packages/components/tests/dummy/app/templates/components/copy/button.hbs
KristinLBradley: This was an idea I tried that worked but probably wouldn't want to change the code like this just to get Showcase examples working: https://github.com/hashicorp/design-system/pull/1488/files#r1261492130

didoo: Yeah, probably better not introduce arguments only for the sake of the showcase. 
I may have an idea, but I have to test it (essentially, at page load replace via JS the value of the symbol reference of the SVG icon in the `hds-copy-button--success` buttons).

packages/components/addon/components/hds/copy/button/index.hbs
```diff
@@ -0,0 +1,16 @@
...
+<Hds::Button
+  class="{{this.classNames}}"
+  @size={{this.size}}
+  @isFullWidth={{@isFullWidth}}
+  @text={{@text}}
+  @icon={{this.icon}}
+  @isIconOnly={{@isIconOnly}}
+  @color="secondary"
+  @iconPosition="trailing"
+  {{clipboard text=this.textToCopy target=this.targetToCopy onError=this.onError onSuccess=this.onSuccess}}
```
Dhaulagiri, line 14: [note] we'll need to pull ember-cli-clipboard in to `dependencies` if we're relying on it here

packages/components/app/styles/components/copy/snippet.scss
```diff
@@ -0,0 +1,108 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+//
+// COPY > SNIPPET
```
MelSumner, line 7: @andgen404 can you especially check the tokens to make sure I got them correct? Thank you!

packages/components/app/styles/components/copy/snippet.scss
```diff
@@ -0,0 +1,108 @@
...
+  .hds-copy-snippet__icon {
+    color: var(--token-color-foreground-success);
+  }
+}
+
+.hds-copy-snippet--error {
+  background-color: var(--token-color-surface-interactive);
+
+  .hds-copy-snippet__icon {
+    color: var(--token-color-foreground-critical-on-surface);
```
andgen404, line 101: Just to clarify, this will occur when there is an `onError` hook?

MelSumner: @andgen404 yes, it felt like the least we could do for now? I can take it out if you want though and just leave the class name there for them to use.

andgen404: Works for me 🙂 just curious on the interaction.

packages/components/app/styles/components/copy/button.scss
```diff
@@ -0,0 +1,30 @@
...
+// so the CSS is scoped to the component class (we're not changing any of the other class names)
+.hds-copy-button {
+  &.hds-button--color-secondary {
+    .hds-button__icon {
+      color: var(--token-color-foreground-action);
+    }
+
+    &.hds-copy-button--success {
+      .hds-button__icon {
+        color: var(--token-color-foreground-success);
```
jorytindall, line 20: should this be `--token-color-foreground-success-on-surface`?

cc @andgen404

packages/components/app/styles/components/copy/snippet.scss
```diff
@@ -0,0 +1,113 @@
...
+    background-color: var(--token-color-surface-interactive);
+    border-color: var(--token-color-border-strong);
+  }
+
+  &:active,
+  &.mock-active {
+    color: var(--token-color-foreground-primary);
+    background-color: var(--token-color-surface-interactive-active);
+    border-color: var(--token-color-border-strong);
+    border-radius: 0.3125rem;
```
jorytindall, line 64: Is this duplicative from what is defined in `.hds-copy-snippet`?

packages/components/app/styles/components/copy/snippet.scss
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    &:focus {
+      color: var(--token-color-foreground-action);
+    }
+  }
+}
+
+.hds-copy-snippet--success {
+  background-color: var(--token-color-surface-interactive);
```
jorytindall, line 90: @andgen404 you've used `Surface/Primary` in Figma. `Surface/Interactive` and `Surface/Primary map to the same value, should probably be `Surface/Interactive`

packages/components/app/styles/components/copy/snippet.scss
```diff
@@ -0,0 +1,113 @@
...
+
+  .hds-copy-snippet__icon {
+    color: var(--token-color-foreground-success);
```
jorytindall, line 93: Same question as the `CopyButton`, should this be `--token-color-foreground-success-on-surface`?

jorytindall (commented):
Let a couple of comments about alignment between Figma and code (both ways). @andgen404 can you have a look and LMK if you have any questions

packages/components/app/styles/components/copy/snippet.scss
MelSumner: yes, nice catch! That's what I get for trying Figma's Developer mode :)

packages/components/app/styles/components/copy/snippet.scss
```diff
@@ -0,0 +1,116 @@
...
+    background-color: var(--token-color-surface-interactive);
+    border-color: var(--token-color-border-strong);
+  }
+
+  &:active,
+  &.mock-active {
+    color: var(--token-color-foreground-primary);
+    background-color: var(--token-color-surface-interactive-active);
+    border-color: var(--token-color-border-strong);
+    border-radius: 0.3125rem;
```
MelSumner, line 65:
```suggestion
```

packages/components/addon/components/hds/copy/button/index.js
```diff
@@ -0,0 +1,96 @@
...
+export const DEFAULT_ICON = 'clipboard-copy';
+export const SUCCESS_ICON = 'clipboard-checked';
+export const ERROR_ICON = 'clipboard';
+export const DEFAULT_STATUS = 'idle';
+
+export default class HdsCopyButtonComponent extends Component {
+  @tracked status = DEFAULT_STATUS;
+  @tracked timer;
+
+  get icon() {
```
MelSumner, line 22: I feel like this code could be cleverer, but then maybe it wouldn't be so easy to read.

packages/components/app/styles/components/copy/button.scss
andgen404: Good catch, yeah I use `--token-color-foreground-success-on-surface` for the clipboard-checked icon

packages/components/app/styles/components/copy/snippet.scss
andgen404: @jorytindall I'll update the color in Figma to reflect this

packages/components/app/styles/components/copy/button.scss
MelSumner, line 20:
```suggestion
        color: var(--token-color-foreground-success-on-surface);
```

packages/components/app/styles/components/copy/snippet.scss
```diff
@@ -0,0 +1,115 @@
...
+      color: var(--token-color-foreground-action);
+    }
+  }
+}
+
+.hds-copy-snippet--success {
+  background-color: var(--token-color-surface-interactive);
+
+  .hds-copy-snippet__icon {
+    color: var(--token-color-foreground-success);
```
MelSumner, line 93:
```suggestion
    color: var(--token-color-foreground-success-on-surface);
```

alex-ju (commented):
I did a first pass on this and left a few comments.

I know things have been set in motion and it may be a late remark to have but, for the record, I don't understand the need for two components here. More precisely I struggle to see the different scenarios the `CopySnippet` tries to address. So I would challenge this design decision. Couldn't the `CopySnippet` be a `CopyButton` next to a snippet of text? This way the user can choose whether to copy the text/code manually or use the `CopyButton`. In the current implementation of `CopySnippet` it's much more difficult to do a manual copy as the element is interactive. Also, the two components look fairly similar (esp. on hover/active/focus) but they perform different actions – sometimes they copy the text within the interactive element, sometimes an external snippet which may lead to conduction (biased of my own interactions).

packages/components/tests/dummy/app/styles/app.scss
```diff
@@ -33,6 +33,9 @@
 @import "./showcase-pages/breadcrumb";
 @import "./showcase-pages/button";
 @import "./showcase-pages/card";
+@import "./showcase-pages/copy";
+@import "./showcase-pages/copy/button";
+@import "./showcase-pages/copy/snippet";
```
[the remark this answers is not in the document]
didoo: +1

packages/components/tests/dummy/app/templates/components/copy/button.hbs
```diff
@@ -0,0 +1,101 @@
...
+  <Shw::Grid @columns={{4}} as |SG|>
+    {{#each @model.SIZES as |size|}}
+      {{#each @model.STATES as |state|}}
+        <SG.Item @label="{{capitalize size}} / {{capitalize state}}">
+          <Hds::Copy::Button @text="Copy" @size={{size}} mock-state-value={{state}} @targetToCopy="#targetToCopy" />
+        </SG.Item>
+      {{/each}}
+    {{/each}}
+  </Shw::Grid>
+  <Shw::Text::H2>Status</Shw::Text::H2>
```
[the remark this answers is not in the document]
didoo: @alex-ju it's something I have to do, once the showcase it's finalized (I have to add some JS code that at `page render` swaps the icon reference, in the SVG (we couldn't think of another simple/smart way to do it)

didoo:
> Opinion: I know things have been set in motion and it may be a late remark to have but, for the record, I don't understand the need for two components here. More precisely I struggle to see the different scenarios the `CopySnippet` tries to address. So I would challenge this design decision. Couldn't the `CopySnippet` be a `CopyButton` next to a snippet of text? This way the user can choose whether to copy the text/code manually or use the `CopyButton`. In the current implementation of `CopySnippet` it's much more difficult to do a manual copy as the element is interactive. Also, the two components look fairly similar (esp. on hover/active/focus) but they perform different actions – sometimes they copy the text within the interactive element, sometimes an external snippet which may lead to conduction (biased of my own interactions).

@alex-ju if you look at the API of the two components, it would be hard to reconcile them under a single component, without creating combinations that we don't want to achieve (eg. a `CopyButton` with `@color="primary", just from the top of my head). Also, they do two different things:
- the `Copy::Button` is used to copy some text (any "kind" of text), with a weak (visual) connection between the text that is being copied and the location of that text in the UI (think of a "copy" on a code block); it's a bit generic per se, we don't know/control how the text that is meant to be copies is or how it looks like, so we need to support two way of providing the text: via an argument (`@textToCopy`) or a pointer to a DOM element (`@targetToCopy`)
- the `Copy::Snippet` is a specialized component (hence the minimum number of variants) with a very specific function: to copy a "snippet of code" for which we provide some sort of "preview" for the consumer, in a very specific/controlled way, position and style, in this case we always have a snippet, so we have a single argument which is the `@textToCopy` (that falls back to the snippet text if not provided)

The only way in which I see your proposal working, would be to change (also from the design perspective) how the "copy snippet" functionality works: I am thinking of an `Hds::Form::TextInput/Textarea` with an `Hds::Copy::Button` next  to it, but in that case we start to hit a few snags: the inputs should be `readonly`, in which case their look&feel is not the one we want; or we "emulate" an input creating a custom `[Something]::Snippet` which is just a code-like text with a border. But we end up anyway with two components anyway.

Happy to discuss further (maybe in the engineering sync?)

packages/components/addon/components/hds/copy/button/index.js
MelSumner: I observe that the alert code is doing more than just the status/icon mapping and maybe it's appropriate to be more complex there; I don't mind changing it here, but after thinking about the mapping, I wondered if it was overkill since it was only default/success/error. WDYT?

(also, added comments)

packages/components/addon/components/hds/copy/snippet/index.hbs
```diff
@@ -0,0 +1,13 @@
...
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<button
+  type="button"
+  class={{this.classNames}}
+  {{clipboard text=this.textToCopy onError=this.onError onSuccess=this.onSuccess}}
+  ...attributes
+>
+  {{@textToCopy}}
```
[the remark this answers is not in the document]
MelSumner: I made them the same with the idea that we'd be abstracting that common functionality later; I know we decided not to do that for now, but it's the reasoning behind the decision.

packages/components/tests/dummy/app/styles/showcase-pages/copy/button.scss
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+// COPY > BUTTON
+
+pre {
+  padding: 1rem;
+  color: white;
+  background-color: black;
+  border-radius: 5px;
+}
```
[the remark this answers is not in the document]
MelSumner: it adds some minimal styles to the code block on the component preview page.

packages/components/addon/components/hds/copy/button/index.js
alex-ju: Works for me!

packages/components/addon/components/hds/copy/snippet/index.hbs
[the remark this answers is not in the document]
didoo: the point is that you're potentially transforming the value received via argument `@textToCopy` so potentially `this.textToCopy` and `@textToCopy` could be different. so you have to yield `this.textToCopy`

packages/components/tests/dummy/app/styles/showcase-pages/copy/button.scss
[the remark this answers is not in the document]
didoo: but this is a global style! 

please, look at other examples on how it's done using the `body` class

packages/components/addon/components/hds/copy/button/index.js
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * @param textToCopy
+   * @type {string}
+   * @description The string to be copied.
+   */
+  get textToCopy() {
+    return this.args.encoded
+      ? decodeURI(this.args.textToCopy)
+      : this.args.textToCopy;
```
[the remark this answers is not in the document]
meirish: :wave: I don't know that we currently need to support this - I think a consumer could decode / encode the inputs if needed.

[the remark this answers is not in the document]
didoo: @meirish for context: the use case is if the consumer needs to show some markup, and the markup is inlined in the template

```
<Doc::CopyButton @textToCopy="%3CHds%3A%3AButton%20%40text%3D%22Button%22%20%7B%7Bon%20%22click%22%20this.action%7D%7D%20%2F%3E" @type="solid" @encoded={{true}} />
```

(this is how we implemented in the documentation website)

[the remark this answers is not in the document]
MelSumner: ok, removed it.

packages/components/tests/integration/components/hds/copy/button/index-test.js
```diff
@@ -0,0 +1,61 @@
...
+
+  test('it should render the default component variation with a CSS class that matches the component name', async function (assert) {
+    await render(
+      hbs`<Hds::Copy::Button id="test-copy-button" @text="copy" @clipboardText="#clipboardText" />`
+    );
+    assert.dom('#test-copy-button').hasClass('hds-copy-button');
+  });
+
+  test('it should only render an icon and also render an aria-label if isIconOnly is set to true', async function (assert) {
+    await render(
```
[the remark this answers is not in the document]
MelSumner: I did not do it that way because that's not the default use of the component. Typically for tests I use the defaults and change only the thing I'm testing.

packages/components/tests/integration/components/hds/copy/snippet/index-test.js
```diff
@@ -0,0 +1,36 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | hds/copy/snippet/index', function (hooks) {
```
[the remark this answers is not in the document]
MelSumner: Added.

packages/components/tests/integration/components/hds/copy/snippet/index-test.js
```diff
@@ -0,0 +1,36 @@
...
+  setupRenderingTest(hooks);
+
+  test('it should render the default component variation with a CSS class that matches the component name', async function (assert) {
+    await render(
+      hbs`<Hds::Copy::Snippet id="test-copy-snippet" @textToCopy="3423g-234525-h345346-f34rtf4" />`
+    );
+    assert.dom('#test-copy-snippet').hasClass('hds-copy-snippet');
+  });
+
+  test('it should render the secondary color if defined', async function (assert) {
```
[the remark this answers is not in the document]
MelSumner: I have a test for the correct default variation, see line 31.

packages/components/tests/integration/components/hds/copy/button/index-test.js
```diff
@@ -0,0 +1,61 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Component | hds/copy/button/index', function (hooks) {
```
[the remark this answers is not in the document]
MelSumner: Added.

Dhaulagiri:
Screenshot 2023-07-27 at 09 09 01

There seems to be some text overflow in the CopyItem now

MelSumner:
@Dhaulagiri ok fixed! 

CleanShot 2023-07-27 at 10 19 57

packages/components/tests/dummy/app/templates/components/copy/button.hbs
MelSumner: I exactly didn't ship that because it would only be for the showcase 👍

packages/components/app/styles/components/copy/button.scss
```diff
@@ -0,0 +1,31 @@
...
+.hds-copy-button {
+  cursor: pointer;
+
+  .hds-button__icon {
+    color: var(--token-color-foreground-action);
+  }
+
+  &.hds-copy-button--status-success {
+    .hds-button__icon {
+      color: var(--token-color-foreground-success-on-surface);
```
andgen404, line 21:
```suggestion
      color: var(--token-color-foreground-success);
```

packages/components/app/styles/components/copy/snippet.scss
```diff
@@ -0,0 +1,115 @@
...
+      color: var(--token-color-foreground-action);
+    }
+  }
+}
+
+.hds-copy-snippet--success {
+  background-color: var(--token-color-surface-interactive);
+
+  .hds-copy-snippet__icon {
+    color: var(--token-color-foreground-success-on-surface);
```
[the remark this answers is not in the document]
andgen404: I think that assumption is correct, but the `color: var(--token-color-foreground-success-on-surface)` should be `color: var(--token-color-foreground-success)`

packages/components/app/styles/components/copy/snippet.scss
```diff
@@ -0,0 +1,114 @@
...
+
+    &:focus {
+      color: var(--token-color-foreground-action);
+    }
+  }
+}
+
+.hds-copy-snippet--status-success {
+  .hds-copy-snippet__icon {
+    color: var(--token-color-foreground-success-on-surface);
```
andgen404, line 81:
```suggestion
    color: var(--token-color-foreground-success);
```

andgen404 (changes requested):
Just some `on-surface` changes to the `onSuccess` and `onError`

packages/components/tests/dummy/app/templates/components/copy/button.hbs
didoo: We discussed this and agreed is something I'll take care of, at the end of the process (when the component and the showcase is stable enough).

packages/components/tests/dummy/app/templates/components/copy/button.hbs
```diff
@@ -0,0 +1,101 @@
...
+    </SF.Item>
+  </Shw::Flex>
+  <Shw::Flex as |SF|>
+    <SF.Item>
+      <Shw::Label>Copy an input value</Shw::Label>
+      <Hds::Form::TextInput::Field @value="036140285924" @name="test-input" @id="test-input" as |F|>
+        <F.Label>Input Label</F.Label>
+      </Hds::Form::TextInput::Field>
+      <br />
+      <Hds::Copy::Button @text="Copy the readonly input" @targetToCopy="#test-input" />
```
Dhaulagiri, line 91: The input in this case is still not readonly (previous comment was lost or is in the history somewhere), is that important to the example here?

MelSumner: It doesn't matter but I can change the label.

MelSumner, line 91:
```suggestion
      <Hds::Copy::Button @text="Copy the input value" @targetToCopy="#test-input" />
```

packages/components/addon/components/hds/copy/button/index.hbs
```diff
@@ -0,0 +1,22 @@
...
+  @isFullWidth={{@isFullWidth}}
+  @text={{@text}}
+  @icon={{this.icon}}
+  @isIconOnly={{@isIconOnly}}
+  @color="secondary"
+  @iconPosition="trailing"
+  {{clipboard
+    text=this.textToCopy
+    target=this.targetToCopy
+    container=this.container
```
alex-ju, line 17: Good to have this covered here, now that we know it's useful when the Copy components are used in Modals/Flyouts

packages/components/addon/components/hds/copy/button/index.js
```diff
@@ -0,0 +1,118 @@
...
+    this.resetStatusDelayed();
+  }
+
+  @action
+  onError() {
+    this.status = 'error';
+    this.resetStatusDelayed();
+  }
+
+  resetStatusDelayed() {
```
alex-ju, line 111: Clean timeout implementation! ✨

packages/components/tests/dummy/app/styles/showcase-pages/copy/button.scss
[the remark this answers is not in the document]
alex-ju: blocker: agree with Cristiano, this should be at least scoped to `body.components-copy`

.changeset/two-pumpkins-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+Adds the CopyButton and CopySnippet components.
```
alex-ju, line 5:
```suggestion
Adds the `CopyButton` and `CopySnippet` components.
```

packages/components/addon/components/hds/copy/button/index.hbs
alex-ju, line 17: We could've probably used the arguments directly without defining getters for each one as we don't seem to be altering these values or returning any custom defaults. However it may be a good way to document these arguments.

```suggestion
    text=@textToCopy
    target=@targetToCopy
    container=@container
```

alex-ju (commented):
Overall looks good! Only one remaining blocker on my side, the global `pre` style in the dummy app.

packages/components/addon/components/hds/copy/button/index.hbs
KristinLBradley: +1 I would remove those getters to simplify the code if feasible

andgen404 (approved):
Design to Code LGTM 🎉

packages/components/addon/components/hds/copy/snippet/index.js
```diff
@@ -0,0 +1,127 @@
...
+    );
+
+    return color;
+  }
+
+  /**
+   * @param isFullWidth
+   * @type {boolean}
+   * @default false
+   * @description Indicates that the component should take up the full width of the parent container. The default is false.
```
KristinLBradley, line 70:
```suggestion
   * @description Indicates that the component should take up the full width of the parent container.
```

KristinLBradley: Super nit but the default is already documented above this line.

packages/components/addon/components/hds/copy/snippet/index.js
```diff
@@ -0,0 +1,127 @@
...
+   */
+  get isFullWidth() {
+    return this.args.isFullWidth ?? false;
+  }
+
+  /**
+   * @param isTruncated
+   * @type {boolean}
+   * @default false
+   * @description Indicates that the component should be truncated instead of wrapping text and using multiple lines. The default is false.
```
KristinLBradley, line 80:
```suggestion
   * @description Indicates that the component should be truncated instead of wrapping text and using multiple lines.
```

packages/components/app/styles/components/copy/snippet.scss
```diff
@@ -0,0 +1,114 @@
...
+
+.hds-copy-snippet {
+  @include hds-focus-ring-with-pseudo-element();
+
+  display: flex;
+  gap: 8px;
+  align-items: center;
+  justify-content: space-between;
+  width: 100%;
+  max-width: 100%;
```
KristinLBradley, line 19: Since you define the width as 100% already, you shouldn't need to add a max-width also.

packages/components/app/styles/components/copy/snippet.scss
```diff
@@ -0,0 +1,114 @@
...
+}
+
+.hds-copy-snippet__icon {
+  flex: none;
+}
+
+.hds-copy-snippet--width-full {
+  justify-content: center;
+  width: 100%;
+  max-width: 100%;
```
KristinLBradley, line 103: `.hds-copy-snippet` includes a width and max-width for `100%` already so this seems confusing to me. (Also as mentioned above you, you shouldn't need the max-width since the width is already set to `100%`.)

packages/components/tests/integration/components/hds/copy/button/index-test.js
KristinLBradley, line 1: Normally we seem to include a test for attribute spreading if the component supports it.

[See example starting on line 16](https://github.com/hashicorp/design-system/blob/main/packages/components/tests/integration/components/hds/reveal/index-test.js)

KristinLBradley (commented):
I added a few comments on some things to look at and consider.

packages/components/addon/components/hds/copy/button/index.hbs
MelSumner: @KristinLBradley @alex-ju this was really just so we could put this into our docs, otherwise the consumer would definitely have to leave our site just to figure out the API for the things they really need to use. I'd prefer to not do that, but willing to align with what the group wants. WDYT?

packages/components/tests/integration/components/hds/copy/button/index-test.js
MelSumner: I wonder why, though. I didn't add it because the fact that the tests work because we can add an attribute is, in itself, proving that `...attributes` is working as expected. WDYT?

packages/components/addon/components/hds/copy/button/index.js
```diff
@@ -0,0 +1,118 @@
...
+  get targetToCopy() {
+    return this.args.targetToCopy;
+  }
+
+  /**
+   * @param container
+   * @type {string}
+   * @description For cases where focus management is needed, set the focused element as the container value.
+   */
+  get container() {
```
Dhaulagiri, line 45: FWIW i don't see this documented in the docs PR

packages/components/addon/components/hds/copy/button/index.hbs
MelSumner: Ok after talking it through a bit more, am fine with the change. 👍

packages/components/addon/components/hds/copy/button/index.js
MelSumner: Good catch, we only just added support for this, I'll put it in the docs now. 👍

packages/components/tests/integration/components/hds/copy/button/index-test.js
KristinLBradley: To me it honestly doesn't seem super valuable but I've been requested to add tests for attribute spreading myself so it seemed to be our standard.

KristinLBradley (approved):
This is looking good to me overall now. Nice work!

PR #1467, opened 2023-06-30
Refactor references to submit support requests

Dhaulagiri:
### :pushpin: Summary

The purpose here is to streamline our channels for submitting support requests.

### :hammer_and_wrench: Detailed description

Places to test
* [Support](https://hds-website-git-br-support-workflow-hashicorp.vercel.app/about/support)
* [Contribution](https://hds-website-git-br-support-workflow-hashicorp.vercel.app/about/contribution)
* [Icons](https://hds-website-git-br-support-workflow-hashicorp.vercel.app/icons/usage-guidelines#resources)
* [Tabs](https://hds-website-git-br-support-workflow-hashicorp.vercel.app/components/tabs#overflow)
* [Patterns](https://hds-website-git-br-support-workflow-hashicorp.vercel.app/patterns)
* [Designers getting started](https://hds-website-git-br-support-workflow-hashicorp.vercel.app/getting-started/for-designers#detaching-components)
* Any component's a11y section - [Example](https://hds-website-git-br-support-workflow-hashicorp.vercel.app/components/alert?tab=accessibility)

### Open question(s)

- [x] Do we want _every_ instance to push people to the support page or are we ok with some instances routing directly to the intake form? (Answer: Err on going direct except in cases where external users may want to open an issue - i.e. a11y reports)

### :link: External links

Jira ticket: [HDS-2162](https://hashicorp.atlassian.net/browse/HDS-2162)

***

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-2162]: https://hashicorp.atlassian.net/browse/HDS-2162?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1493, opened 2023-07-13
Add /support redirect

Dhaulagiri:
### :pushpin: Summary

This PR adds a `/support` path that redirects to the actual support page which is nested under `/about/support`. The thinking here is that we could more easily link people to https://helios.hashicorp.design/support (or https://hds-website-git-br-support-redirect-hashicorp.vercel.app/support in the preview). 

I pulled this out of https://github.com/hashicorp/design-system/pull/1467 since it's related but ultimately a separate concern in my view and could be merged separately (or not if it doesn't seem valuable).

The changes uses vercel's [redirect logic](https://vercel.com/docs/concepts/projects/project-configuration#redirect-object-definition), and specifically the `permanent: false` option which gives this status code which seems appropriate for what we want to do here:

> 307 Temporary Redirect: Not cached by client, the method and body never changed. This type of redirect does not affect SEO and search engines will treat them as normal redirects.

Jira ticket: [HDS-2197](https://hashicorp.atlassian.net/browse/HDS-2197)

***

[HDS-2197]: https://hashicorp.atlassian.net/browse/HDS-2197?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/vercel.json
+++ b/website/vercel.json
@@ -20,5 +20,8 @@
     }
   ],
   "framework": "ember",
-  "rewrites": [{ "source": "/(.*)", "destination": "/_empty.html" }]
+  "rewrites": [{ "source": "/(.*)", "destination": "/_empty.html" }],
+  "redirects": [
+    { "source": "/support", "destination": "/about/support", "permanent": false }
+  ]
 }


didoo (approved):

heatherlarsen (approved):
:shipit:

PR #1504, opened 2023-07-17
add favicon to showcase

Dhaulagiri:
### :pushpin: Summary

Basically so I can see our logo in my bookmarks bar instead of the default.

If we don't want to do this for some reason I'm also ok with that...


alex-ju (approved):

didoo (approved):
☀️

Dhaulagiri:
Sadly this didn't work in production because the favicon is not getting fingerprinted for some reason...

PR #1506, opened 2023-07-17
Update the layout and styling of the "Component API" docs component

didoo:
### :pushpin: Summary

This PR implements the refreshed "Component API" component for the documentation.

In this PR I have:
- updated the `Doc::ComponentApi::Property` component
  - used simple `<div>` elements instead of `<dl>` elements (I don't think it's possible to maintain the same semantic structure now)
  - removed all the extra labels from the content
  - removed all the extra `<div>` containers (not needed anymore)
- updated the `Doc::ComponentApi::Property` styling
  - removed the grid layout, now normal stacking of elements works fine
  - followed the new design specs for `name/type/required/values`

👉 👉 👉 **Previews:**  
- https://hds-website-git-componen-api-refresh-hashicorp.vercel.app/components/alert?tab=code#component-api
- - https://hds-website-git-componen-api-refresh-hashicorp.vercel.app/components/table?tab=code#component-api

Alert-Helios-Design-System

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2195
Figma file: https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?type=design&node-id=3342%3A89190&mode=design&t=TAwEC01tainr46sv-1

***

### 👀 Reviewer's checklist:

### DIFF
--- a/website/app/styles/doc-components/component-api.scss
+++ b/website/app/styles/doc-components/component-api.scss
@@ -103,8 +103,11 @@
 
 .doc-component-api__property-description {
   .doc-component-api {
-    margin: 16px 0;
+    margin: 16px 0 16px 8px;
+    padding-right: 16px;
+    padding-left: 16px;
     border: none;
+    border-left: 2px solid var(--doc-color-feedback-information-300);
     border-radius: 0;
 
     &:last-child {
@@ -113,14 +116,14 @@
   }
 
   .doc-component-api__property {
-    padding-top: 16px;
-    padding-bottom: 16px;
-    padding-left: 12px;
-  }
+    padding-left: 0;
 
-  .doc-component-api__property-values,
-  .doc-component-api__property-description {
-    padding-left: 12px;
-    border-left: 2px solid var(--doc-color-feedback-information-300);
+    &:first-child {
+      padding-top: 12px;
+    }
+
+    &:last-child {
+      padding-bottom: 12px;
+    }
   }
 }
--- a/website/app/styles/doc-components/component-api.scss
+++ b/website/app/styles/doc-components/component-api.scss
@@ -68,18 +68,18 @@
 // second row - values
 
 .doc-component-api__property-values {
-  margin: 24px 0 0 0;
-  padding: 0;
+  @include doc-font-family("mono"); // custom font styling
+  margin: 12px 0 0 0;
+  padding: 12px 0 0 0;
+  font-size: 13px;
+  line-height: 1.714;
 }
 
 .doc-component-api__property-value {
-  @include doc-font-family("mono"); // custom font styling
   display: inline-block;
   margin-bottom: 8px;
   padding: 0 6px;
   color: var(--doc-color-gray-200);
-  font-size: 13px;
-  line-height: 1.714;
   background-color: var(--doc-color-gray-600);
   border: 1px solid transparent;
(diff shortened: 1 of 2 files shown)


heatherlarsen (commented):
It looks like there may be some style differences between this version and @cveigt's version in Figma. 

The ones I could see right off the bat were:

- Font size of the property values is a little larger (14px in code, 13px in Figma)
- Border radius of the container (5px in code, 3px in Figma)
- Vertical border of the nested properties isn't quite to spec (see image)

Screenshot 2023-07-19 at 8 09 36 PM

There may be viable reasons for this, but I thought I'd note them.

didoo:
> * Font size of the property values is a little larger (14px in code, 13px in Figma)
> * Border radius of the container (5px in code, 3px in Figma)
> * Vertical border of the nested properties isn't quite to spec (see image)

@cveigt @heatherlarsen 
- for points 1 and 2 I've fixed it now (we have a standard "code" style in the website and these two are custom; there was a mismatch of `line-height` values though, so I used `1.714` for all of them which is the "standard" one for the `code` style. 
- for point 3 instead it's technically correct as is now (the entire "property block" is indented, not just the property description, hence the border on the left is for all the block); making it as is designed is possible, but require much more CSS overriding that I would like to avoid unless strictly necessary

@cveigt please have a look if there are other discrepancies between the design in Figma and the implementation in code, and let me know. thanks

website/app/styles/doc-components/component-api.scss
```diff
@@ -5,118 +5,51 @@
...
 }
 
+
+// SINGLE PROPERTY "ITEM"
+
 .doc-component-api__property {
   position: relative;
-  display: grid;
-  grid-template-rows: repeat(6, auto);
-  grid-template-columns: repeat(3, 1fr);
```
alex-ju, line 28: I'm glad we got rid of this complex layout

alex-ju (approved):
Code looks good – agree `dl` is not suitable for the new design/structure.

KristinLBradley (approved):

cveigt:
@didoo it looks good overall. The only part is the nested property. Are saying that adding the border to `doc-component-api__property-description` instead of the whole block `doc-component-api__property` would require too much CSS overwriting? Could you explain the difference?

cveigt:
It looks much better with the border and separation as shown in the design. Since this is an internal component, I wouldn't mind the overrides, but if the consensus is that it's not acceptable, we can go with the alternative.

didoo:
@cveigt I've made the changes you requested. have a look if it's what you want

cveigt:
@heatherlarsen I paired with @didoo to try getting this closer to the specs, and this is the closer we got:

image

To get a 1:1 with the design, we'd have to refactor the spacing of the component/property block or keep overwriting styles.

In context, it still looks a bit weird in some cases because the border will show differently based on the combination of content (e.g., name + description, name + values + description, one-line description, three-line description, etc...)

image

At this point, I think the simpler solution looks better and easier to navigate visually vs the one shown above, but wanted to know your thoughts since this isn't what was approved.

Screen Shot 2023-07-21 at 1 39 00 PM

heatherlarsen:
@cveigt I'm good with going for the simpler approach you've outlined here.

PR #1507, opened 2023-07-17
Update copy-item to use copy/snippet

MelSumner:
### :pushpin: Summary

If merged, this would update the dropdown/copy-item to use the copy/snippet. 

I made this a separate PR because we are still on the fence about what to do, so we can still ship the rest of the copy component w/o this if we want to.

So it's not so much about reviewing the code at this point since it would get merged into the copy-component branch; it's more about answering our open question about shipping this change as well. 

Nothing changes for the consumer's invocation, but the styles are updated to match copy/snippet (see screenshot)

CleanShot 2023-07-17 at 16 42 12

### :link: External links

Figma file: [if it applies]

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/app/styles/components/dropdown.scss
+++ b/packages/components/app/styles/components/dropdown.scss
@@ -278,61 +278,17 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
 
 .hds-dropdown-list-item--variant-copy-item {
   width: 100%;
+  max-width: 100%;
   padding: 10px 16px 12px;
 
-  button {
-    display: flex;
-    justify-content: space-between;
+  .hds-copy-snippet {
     width: 100%;
-    padding: 12px 8px;
-    color: var(--token-color-foreground-primary);
-    background-color: transparent;
-    border: 1px solid var(--token-color-border-primary);
-    border-radius: 5px;
-
-    &:hover,
-    &.mock-hover {
-      background-color: var(--token-color-surface-interactive-hover);
-      cursor: pointer;
-    }
-
-    @include hds-focus-ring-basic();
-
-    &:focus,
-    &.mock-focus {
-      // TODO this focus is just way too complex
-      background-color: var(--token-color-surface-action);
-      border-color: var(--token-color-focus-action-internal);
-    }
-
-    &:active,
-    &.mock-active {
-      background-color: var(--token-color-surface-interactive-active);
-    }
-
-    &.is-success {
-      background-color: var(--token-color-surface-success);
-      border-color: var(--token-color-border-success);
-
-      .hds-dropdown-list-item__copy-item-icon {
-        color: var(--token-color-foreground-success);
-      }
-    }
+    max-width: 100%;
+    white-space: normal;
+    overflow-wrap: anywhere;
   }
 }
 
-.hds-dropdown-list-item__copy-item-text {
-  overflow: hidden;
-  white-space: nowrap;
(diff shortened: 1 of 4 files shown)


packages/components/app/styles/components/dropdown.scss
```diff
@@ -278,61 +278,17 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
 
 .hds-dropdown-list-item--variant-copy-item {
   width: 100%;
+  max-width: 100%;
```
didoo, line 281: I don't think you need this (I checked in the showcase, and commenting it out doesn't do anything)

packages/components/app/styles/components/dropdown.scss
```diff
@@ -278,61 +278,17 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
...
   padding: 10px 16px 12px;
 
-  button {
-    display: flex;
-    justify-content: space-between;
+  .hds-copy-snippet {
```
didoo, line 284: these stiles should not live here but in the CSS code for the `Copy::Snippet` component (so the other PR); plus I think you need only the `overflow-wrap` (TBD if `anywhere` or `break-word`)

packages/components/tests/dummy/app/templates/components/dropdown.hbs
```diff
@@ -548,27 +548,38 @@
     </SF.Item>
   </Shw::Flex>
 
-  <Shw::Text::H4>States</Shw::Text::H4>
+  <Shw::Text::H3>States</Shw::Text::H3>
 
   <Shw::Flex as |SF|>
-    <SF.Item>
+    <SF.Item @label="Base">
```
didoo, line 554: you can avoid this duplication using a wrapping `{{#let}}` declaration (see [an example here](https://github.com/hashicorp/design-system/blob/main/packages/components/tests/dummy/app/templates/components/accordion.hbs#L254)) and use the boolean to control if add the `@copyItemTitle` or not

didoo (commented):
added a few comments

packages/components/app/styles/components/copy/snippet.scss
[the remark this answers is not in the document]
Dhaulagiri: I think whatever these were went away when I rebased this again the branch from the main PR

heatherlarsen (approved):

packages/components/app/styles/components/dropdown.scss
MelSumner: I should have addressed this before I merged into the other branch, my apologies. I plan to address this in the feature branch. I think it might matter when it comes to long strings, but if it doesn't, I'll remove it.

packages/components/app/styles/components/dropdown.scss
MelSumner: I think they were not in that other component yet, which is why I might have felt the need to add them here. I will evaluate and address appropriately in the feature branch.

packages/components/tests/dummy/app/templates/components/dropdown.hbs
MelSumner: I probably didn't think about this since I was focused on copy/paste from other component pages, as our team has discussed many times before. I will integrate this improvement if there's time, or file a JIRA ticket to review this (if this component page needs it, there's a chance other preview pages do, too).

PR #1509, opened 2023-07-19

hashibot-hds:


didoo (approved):
I haven't checked all the keyword, just had a general glance, and everything seems OK

.changeset/late-suits-talk.md
```diff
@@ -0,0 +1,5 @@
+---
```
Dhaulagiri, line 1: Unless I'm missing something, I don't think this requires a changeset/release as the keywords are only used on the website.

cveigt: @Dhaulagiri, good call out. I didn't think of it. Just took it off

PR #1511, opened 2023-07-19
Document potential layout shift in `Modal` and `Flyout`

alex-ju:
### :pushpin: Summary

If a user's scroll bar is set to appear always or an external device (such as a mouse or trackpad) is connected, opening a Modal/Flyout will result in a slight layout shift on the x-axis. We've been aware of this issue when developing the Modal component but considering no clear solution (to be read as non-hacky or without causing side effects) was found and acknowledging this as a common issue across the industry the issue was deemed as acceptable. Based on previous inquiries on this topic and following a conversation with @emlanctot during office hours we decided to document it so that our consumers are aware of this behavior/limitation.

Default settings on MacOSX

modal-default

Scroll always visible/mouse connected on MacOSX (keep an eye on the right-side edge of the page)

modal-scroll-visible

### DIFF
--- a/website/docs/components/flyout/partials/code/how-to-use.md
+++ b/website/docs/components/flyout/partials/code/how-to-use.md
@@ -4,7 +4,7 @@ The `Hds::Flyout` component leverages the `<dialog>` element which is [currently
 
 ## Page scroll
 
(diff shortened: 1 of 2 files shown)


didoo (approved):
👍

PR #1518, opened 2023-07-20
`PageHeader` - Fix text overflow for non-breaking text

didoo:
### :pushpin: Summary

While discussing HDS components adoption with Bruce Ashtiani (TFC) he told me that there may be cases in which the title of a page can very long and doesn't break easily (could be a name of a resource). Doing a quick test in the showcase, I have realized that the `PageHeader` doesn't handle correctly this use case, so I made a quick fix.

In this PR I have:
- added `min-width: 0` + `max-width: 100%` + `overflow-wrap: break-word` where needed (there are multiple nested flexbox elements, so there were multiple places where we needed to force the container layout to not blow beyond the parent element width)

👉 👉 👉 **Preview** (see "Non-breaking Title/Subtitle/Description + badge"): https://hds-showcase-git-page-header-text-overflow-hashicorp.vercel.app/components/page-header

**Before:**
screenshot_2983

**After:**
screenshot_2984

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/page-header/title.hbs
+++ b/packages/components/addon/components/hds/page-header/title.hbs
@@ -2,4 +2,4 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<h1 class="hds-foreground-strong hds-typography-display-500 hds-page-header__title-text" ...attributes>{{yield}}</h1>
+<h1 class="hds-foreground-strong hds-typography-display-500 hds-page-header__title" ...attributes>{{yield}}</h1>
--- a/packages/components/addon/components/hds/page-header/title.hbs
+++ b/packages/components/addon/components/hds/page-header/title.hbs
@@ -2,4 +2,4 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<h1 class="hds-foreground-strong hds-typography-display-500 hds-page-header__title" ...attributes>{{yield}}</h1>
+<h1 class="hds-foreground-strong hds-typography-display-500 hds-page-header__title-text" ...attributes>{{yield}}</h1>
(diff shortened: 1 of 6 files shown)


packages/components/addon/components/hds/page-header/title.hbs
```diff
@@ -2,4 +2,4 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<h1 class="hds-foreground-strong hds-typography-display-500 hds-page-header__title" ...attributes>{{yield}}</h1>
\ No newline at end of file
+<h1 class="hds-foreground-strong hds-typography-display-500 hds-page-header__title-text" ...attributes>{{yield}}</h1>
```
alex-ju, line 5: I assume we're not worried about this class name change because we don't have users of this component _yet_

alex-ju (approved):
Thanks for putting this together! 🙌 

No strong opinion, but arguably `hds-page-header__title-text` could be kept as `hds-page-header__title` and maybe `hds-page-header__title-main` could be `hds-page-header__title-wrapper`, as we have `hds-form-radio-card__control` and `hds-form-radio-card__control-wrapper` or `hds-page-header__badges-wrapper` over badges.

didoo:
> No strong opinion, but arguably `hds-page-header__title-text` could be kept as `hds-page-header__title` and maybe `hds-page-header__title-main` could be `hds-page-header__title-wrapper`, as we have `hds-form-radio-card__control` and `hds-form-radio-card__control-wrapper` or `hds-page-header__badges-wrapper` over badges.

Actually it's a good suggestion! I've made the changes 👍

> I assume we're not worried about this class name change because we don't have users of this component yet

I've checked in GitHub (advanced search) and in Cloud UI code (IDE) and I haven't seen any occurrence of these class names. So we should be quite safe in renaming them.

alex-ju (approved):

PR #1524, opened 2023-07-26
`TextInput` Example size change for `Datetime`

andgen404:
### :pushpin: Summary

If merged, this PR will update the size of the datetime example to match the one in Figma

Removed @Width from examples 

### :link: External links

Jira ticket: [HDS-2247](https://hashicorp.atlassian.net/browse/HDS-2247)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

[HDS-2247]: https://hashicorp.atlassian.net/browse/HDS-2247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/docs/components/form/text-input/partials/guidelines/guidelines.md
+++ b/website/docs/components/form/text-input/partials/guidelines/guidelines.md
@@ -44,15 +44,15 @@ Text Input accepts [all native HTML types](https://developer.mozilla.org/en-US/d
 Date and time fields use the native browser functionality for the popovers. Some browsers do not display an icon or popover.
 !!!
 
-<Hds::Form::TextInput::Field @type="date" placeholder="mm/dd/yy" @width="150px" as |F|>
+<Hds::Form::TextInput::Field @type="date" placeholder="mm/dd/yy" @width="192px" as |F|>
   <F.Label>Date</F.Label>
 </Hds::Form::TextInput::Field>
 
-<Hds::Form::TextInput::Field @type="time" placeholder="--:-- --" @width="150px" as |F|>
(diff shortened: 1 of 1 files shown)


didoo:
These are the "native" widths of the inputs in a browser (which is probably what a consumer will use, and a user will see)
screenshot_3003

Not sure why we want to have them with the same width (also, not sure why in Figma they have different sizes than in the browser - by default, I mean)

/cc @hashicorp/hds-design

heatherlarsen:
@didoo I agree; these should not be the same size in the docs and should be reflective of the native elements. 

> not sure why in Figma they have different sizes than in the browser

I don't recall exactly why that is for the original two; probably just an oversight. Figma isn't very far off from the browser sizes, but I'll create a ticket to fix them.
- Date: 136px (browser), 140px (figma)
- Time: 109.2px (browser), 112px (figma)
- Datetime-local: 208px (browser), 192px (figma)


@andgen404 I think there's been a little confusion about what the task. Sorry for not having more clear direction for this. The intention was not the update all 3 of these to the same size but to fix the datetime-local example, which was being cut off. Now that I'm looking at the old code, though, they were incorrectly set to the same size to begin with. 🙁 My recommendation would be to remove the `@width` value from all 3 of them, which should put them back to their native size, but let me know if you run into trouble, and I can pair on it.

didoo (approved):
👍

heatherlarsen (approved):
🚀

PR #1530, opened 2023-07-27
Fix Embroider warnings caused by incorrect export of internal utility functions

alex-ju:
### :pushpin: Summary

Fix Embroider warnings caused by incorrect export of internal utility functions.

Initially, I was tempted to fix this by adding defaults to the utility functions, but then I realized these functions should not be part of the public API of the library (as they are designed for internal use only) so I decided to remove them from the API instead.

Remove `app/utils` exports as they cause the following embroider warnings in the consumer codebase:

```
WARNING in ../rewritten-packages/@hashicorp/design-system-components.4bc2410c/_app_/utils/hds-get-element-id.js 6:0-87
export 'default' (reexported as 'default') was not found in '@hashicorp/design-system-components/utils/hds-get-element-id' (possible exports: getElementId)
 @ ./assets/hcp.js 215:13-55

WARNING in ../rewritten-packages/@hashicorp/design-system-components.4bc2410c/_app_/utils/hds-set-aria-described-by.js 6:0-94
export 'default' (reexported as 'default') was not found in '@hashicorp/design-system-components/utils/hds-set-aria-described-by' (possible exports: setAriaDescribedBy)
 @ ./assets/hcp.js 218:13-62
```

### :link: External links

Jira ticket: [HDS-2275](https://hashicorp.atlassian.net/browse/HDS-2275)

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable

[HDS-2275]: https://hashicorp.atlassian.net/browse/HDS-2275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/serious-ravens-promise.md
+++ b/.changeset/serious-ravens-promise.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Fix Embroider warnings caused by incorrect export of internal utility functions
--- a/packages/components/app/utils/hds-get-element-id.js
+++ b/packages/components/app/utils/hds-get-element-id.js
@@ -1,6 +1,0 @@
-/**
- * Copyright (c) HashiCorp, Inc.
- * SPDX-License-Identifier: MPL-2.0
- */
-
-export { default } from '@hashicorp/design-system-components/utils/hds-get-element-id';
--- a/packages/components/app/utils/hds-set-aria-described-by.js
+++ b/packages/components/app/utils/hds-set-aria-described-by.js
@@ -1,6 +1,0 @@
-/**
- * Copyright (c) HashiCorp, Inc.
- * SPDX-License-Identifier: MPL-2.0
- */
-
-export { default } from '@hashicorp/design-system-components/utils/hds-set-aria-described-by';


didoo (commented):
Have it been tested in a consumer? (I am asking because my understanding is that these almost-empty duplicate files are needed by Ember because our components code is an addon, and so they are needed at compilation time.

MelSumner (changes requested):
I think these need to stay here. Any specific reason to remove them other than the warnings? 

Also what is the warning? Is it something we need to address?

alex-ju:
> Have it been tested in a consumer?

No, I have not – I tried yesterday but setting up `cloud-ui` to use embroider is not trivial. However, I am confident about the export/import mechanism we are altering in this PR. Alternatively, we can potentially ask the consumer to test the 'solution' (@aklkv do you think you would have time to review this change at some point?).

> (I am asking because my understanding is that these almost-empty duplicate files are needed by Ember because our components code is an addon, and so they are needed at compilation time.

> I think these need to stay here. Any specific reason to remove them other than the warnings?

@didoo @MelSumner my understanding is that these `app/*/*.js` files are required for consumers to be able to discover our components and helpers – basically, these exports constitute our add-on API. Our intention is to only use these functions internally and not make them part of the API so I consider removing the exports is the best way forward.

> Also what is the warning? Is it something we need to address?

The warnings complain that we're trying to expose these two utility functions in `app/utils/*.js`, but we're not exporting defaults `addon/utils/*.js`. I copied the warnings from the Jira card to the PR description.

aklkv:
@alex-ju Thank you for addressing this 🙏🏻 I have a branch with working embroider for cloud-ui I can test this change on it. Also I have confirm that everything that is in `app/` is a public API of the addon, if something is used only internally it should not have re export file. Also it is default behaviour of ember-cli to create this file when generating things like utils. we had to remove very large number of useless pre-exports from cloud-ui as well. @MelSumner you are absolutely correct in saying that it's just a warning and it doesn't break anything but why keep it if we don't need it and makes build output for consumer that much cleaner.

didoo (approved):

MelSumner (approved):
I still think we should fix the root cause instead of removing these files but I won't block on it.

PR #1535, opened 2023-07-28
Hds 2278 file upload component

KristinLBradley:
_**STATUS:** Very rough WIP_

### :pushpin: Summary
If merged, this PR will add the `FileInput` component to the design system.

**Done (still rough, will polish):**
- ~~Created new branch from Cristiano’s spike~~
- ~~focus applied the button (not the entire element/container)~~
- ~~Made button shadow elevation visible~~
- ~~Showcase page: Added states & preliminary rough examples~~ 

### :link: External links

Jira ticket: [HDS-2278](https://hashicorp.atlassian.net/browse/HDS-2278)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/app/styles/components/form/file-input.scss
```diff
@@ -0,0 +1,113 @@
...
+
+//
+// FORM > FILE-INPUT
+//
+
+@use "../../mixins/button" as *;
+
+.hds-form-file-input {
+  margin: -4px 0 -4px -4px; // offset shadow padding + default 1px spacing
+  padding: 3px 0 3px 3px; // add padding so button shadow is visible
```
alex-ju, line 14: this is because the user agent enforces an `overflow: hidden`, isn't it?

packages/components/app/styles/components/form/file-input.scss
```diff
@@ -0,0 +1,113 @@
...
+  font-family: var(--token-typography-font-stack-text);
+  border: 1px solid transparent;
+  border-radius: calc($hds-button-border-radius + 1px);
+  // font-size: 1.875rem; // we may want to use this to tweak the size of the text that contains the name of the file?
```
alex-ju, line 18: I think Safari has a tiny `font-size` by default. It would be great if we can set a certain size, but following the pattern in other components, using a typography class may be the preferred way of setting a `font-size` – hence you may actually need `get classNames() { }`

packages/components/tests/acceptance/percy-test.js
```diff
@@ -155,6 +155,10 @@ module('Acceptance | Percy test', function (hooks) {
     await visit('/utilities/dismiss-button');
     await percySnapshot('DismissButton');
 
+    // MOVE THIS BLOCK IN THE RIGHT POSITION
```
alex-ju, line 158: Don't forget to remove this line and group the snapshot with the other form components

packages/components/tests/dummy/app/router.js
```diff
@@ -42,6 +42,7 @@ Router.map(function () {
       this.route('textarea');
       this.route('toggle');
       this.route('radio-card');
+      this.route('file-input');
```
alex-ju, line 45: Would be good if we can sort these alphabetically – it was likely me who placed the `radio-card` at the end 🤦

packages/components/tests/dummy/app/templates/components/form/file-input.hbs
```diff
@@ -0,0 +1,80 @@
...
+    <Shw::Text::Body>
+      <code>Input[type=file]</code>
+      (styled via
+      <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/::file-selector-button">
+        <code>::file-selector-button</code>
+      </a>)
+    </Shw::Text::Body>
+    <Hds::Form::FileInput />
+
+    <Shw::Text::H2>Used next to other content</Shw::Text::H2>
```
alex-ju, line 27: I think we can group these examples with the first one under a 'Content' section – as you did for the [Reveal](https://hds-showcase-git-hds-2278-file-upload-component-hashicorp.vercel.app/components/reveal).

packages/components/tests/dummy/app/templates/index.hbs
```diff
@@ -136,6 +136,11 @@ SPDX-License-Identifier: MPL-2.0
           Form::Select
         </LinkTo>
       </li>
+      <li>
+        <LinkTo @route="components.form.file-input">
+          Form::FileInput
+        </LinkTo>
+      </li>
```
alex-ju, line 143: I would bump this further up to keep the alphabetical order.

alex-ju (commented):
Overall looks good! I recommend we use the base/field model as we do for other components (and thanks for bringing this up in our sync). In the showcase, I would add an example where a value is set (potentially with a lengthy value so we visualize the default truncation – I know it's not part of our component, but if we ever decide to alter it, it would be good to have this reference)

packages/components/addon/components/hds/form/file-input/index.hbs
```diff
@@ -0,0 +1,6 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<input type="file" class={{this.classNames}} ...attributes />
```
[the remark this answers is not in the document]
KristinLBradley: Right, thanks. I just updated it following the pattern of TextInput, etc.

packages/components/addon/components/hds/form/file-input/index.js
```diff
@@ -0,0 +1,22 @@
...
+
+import Component from '@glimmer/component';
+
+export default class HdsFormFileInputIndexComponent extends Component {
+  /**
+   * Get the class names to apply to the component.
+   * @method classNames
+   * @return {string} The "class" attribute to apply to the component.
+   */
+  get classNames() {
```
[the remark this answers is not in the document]
KristinLBradley: Right. I was just leaving this here for now until I knew more about the desired final visual design just in case.

packages/components/app/styles/components/form/file-input.scss
KristinLBradley: Yes

packages/components/app/styles/components/form/file-input.scss
KristinLBradley: I pushed updates using the hds classnames to replace these font-styles earlier.

Safari is problematic but I'll continue to experiment with improving styles for it when the more final visual designs are ready.

packages/components/tests/acceptance/percy-test.js
KristinLBradley: Yep, just noticed that earlier as I was going through refactoring.

packages/components/tests/dummy/app/templates/components/form/file-input.hbs
KristinLBradley: The Showcase is still very much a WIP as the final designs aren't ready yet.

KristinLBradley:
> 

Thanks! The file input doesn't have a `value` attribute like the text input and some other form inputs do for security reasons. We'd have to try to do something using JavaScript to fake it which I could try to do but not sure if it's really worth it?

packages/components/app/styles/components/form/file-input.scss
alex-ju: Yeah, I know, it's tricky. If there's no way around it, I think we can accept this visual difference.

alex-ju:
> > 
> 
> Thanks! The file input doesn't have a `value` attribute like the text input and some other form inputs do for security reasons. We'd have to try to do something using JavaScript to fake it which I could try to do but not sure if it's really worth it?

Yes, only via JavaScript – it may be an overhead. We can rely on manual testing for this state.

packages/components/tests/dummy/app/templates/components/form/file-input.hbs
KristinLBradley: I've cleaned up the Showcase more and added examples of base and field components.

Let me know if there are any other useful examples you think I should add.

packages/components/tests/integration/components/hds/form/file-input/base-test.js
```diff
@@ -0,0 +1,29 @@
...
+  setupRenderingTest(hooks);
+
+  test('it should render the component with a CSS class that matches the component name', async function (assert) {
+    await render(hbs`<Hds::Form::FileInput::Base id="test-form-file-input" />`);
+    assert.dom('#test-form-file-input').hasClass('hds-form-file-input');
+  });
+
+  // ATTRIBUTES
+
+  test('it should spread all the attributes passed to the component', async function (assert) {
```
KristinLBradley, line 21: I'll remove this...

andgen404 (commented):
Everything looks good to me 🎉 but I'm curious for the disabled version, the `upload` icon is still styled with `foreground/primary` but in the showcase the HDS::Button is styled with the disabled color below. Just wanted to make sure the `FileInput` disabled variant will also have the `--token-color-foreground-disabled` applied to the icon. 

Screenshot 2023-08-04 at 2 00 27 PM

packages/components/app/styles/components/form/file-input.scss
```diff
@@ -0,0 +1,114 @@
...
+// Button
+.hds-form-file-input::file-selector-button {
+  min-height: 2.25rem;
+  margin-right: 16px;
+  margin-inline-end: 16px;
+  padding: 0.5625rem 0.9375rem;
+  padding-left: 34px; // override to add space for the background image icon
+  color: var(--token-color-foreground-primary);
+  font: inherit;
+  line-height: 1rem;
```
didoo, line 37: I would remove this override (it comes from the copy&paste of the `Hds::Button` styling) and adjust the vertical padding accordingly (very unlikely this string will be ever multiline)

packages/components/app/styles/components/form/file-input.scss
```diff
@@ -0,0 +1,114 @@
...
+  text-decoration: none;
```
didoo, line 38: this is not needed, it will never be a link (this comes from the copy&paste of the `Hds::Button` styling)

packages/components/app/styles/components/form/file-input.scss
```diff
@@ -0,0 +1,114 @@
...
+
+// Button states
+.hds-form-file-input::file-selector-button:hover,
+.hds-form-file-input.mock-hover::file-selector-button {
+  color: var(--token-color-foreground-primary);
+  background-color: var(--token-color-surface-primary);
+  border-color: var(--token-color-border-strong);
+}
+
+.hds-form-file-input:focus-within::file-selector-button,
```
didoo, line 62: Nice! TIL 🙌

packages/components/app/styles/components/form/file-input.scss
```diff
@@ -0,0 +1,114 @@
...
+    &.mock-disabled,
+    &:disabled:focus,
+    &[disabled]:focus,
+    &.mock-disabled:focus,
+    &:disabled:hover,
+    &[disabled]:hover,
+    &.mock-disabled:hover {
+      @include hds-button-state-disabled();
+    }
+  }
```
didoo, line 113: I think there's something that needs to be cleaned up here (probably the `&::file-selector-button` is not needed).

didoo: Plus, the icon color should be also changed when the input is disabled; maybe you can use something like this: https://github.com/hashicorp/design-system/blob/main/packages/components/app/styles/components/form/text-input.scss#L98-L101

packages/components/tests/dummy/app/templates/components/form/file-input.hbs
```diff
@@ -0,0 +1,111 @@
...
+  <div class="{{if this.showHighlight 'shw-component-form-file-input-layout-highlight'}}">
+    <Shw::Flex {{style gap="2rem"}} as |SF|>
+      <SF.Item @label="FileInput">
+        <Shw::Text::Body>
+          <code>Input[type=file]</code>
+          (styled via
+          <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/::file-selector-button">
+            <code>::file-selector-button</code>
+          </a>)
+        </Shw::Text::Body>
```
didoo, line 27:
```suggestion
```
This was meant only for the spike.

packages/components/tests/dummy/app/templates/components/form/file-input.hbs
```diff
@@ -0,0 +1,111 @@
...
+        <Hds::Form::FileInput::Base />
+      </SF.Item>
+
+      <SF.Item @label="HdsButton for visual reference">
+        <Shw::Text::Body><code>Hds::Button[@color=secondary]</code></Shw::Text::Body>
+        <Hds::Button @icon="upload" @text="Choose file" @color="secondary" />
+      </SF.Item>
```
didoo, line 34:
```suggestion
```

This was meant only for the spike.

packages/components/app/styles/components/form/file-input.scss
didoo: And the `No file chosen` text color too

packages/components/tests/dummy/app/templates/components/form/file-input.hbs
```diff
@@ -0,0 +1,111 @@
...
+    <Shw::Flex {{style gap="2rem"}} as |SF|>
+      {{#each @model.STATES as |state|}}
+        <SF.Item @label={{state}}>
+          <Hds::Button @icon="upload" @text="Choose file" @color="secondary" mock-state-value={{state}} />
+        </SF.Item>
+      {{/each}}
+      <SF.Item @label="disabled">
+        <Hds::Button @icon="upload" @text="Choose file" @color="secondary" mock-state-value="disabled" disabled />
+      </SF.Item>
+    </Shw::Flex>
```
didoo, line 62:
```suggestion
```

This was extremely confusing for me, I was changing the properties on the `FileInput` CSS and expecting to see them reflected here, only to understand that this is actually an `Hds::Button`.

packages/components/tests/dummy/app/templates/components/form/file-input.hbs
```diff
@@ -0,0 +1,111 @@
...
+  <button id="dummy-toggle-highlight" type="button" {{on "click" this.toggleHighlight}}>
+    {{if this.showHighlight "Hide" "Show"}}
+    layout highlight
+  </button>
+
+  <Shw::Text::H2>“Base” control</Shw::Text::H2>
+
+  <div class="{{if this.showHighlight 'shw-component-form-file-input-layout-highlight'}}">
+    <Shw::Flex {{style gap="2rem"}} as |SF|>
+      <SF.Item @label="FileInput">
```
didoo, line 20:
```suggestion
      <SF.Item>
```

didoo (commented):
Added a bunch of comments.

KristinLBradley:
> Everything looks good to me 🎉 but I'm curious for the disabled version, the `upload` icon is still styled with `foreground/primary` but in the showcase the HDS::Button is styled with the disabled color below. Just wanted to make sure the `FileInput` disabled variant will also have the `--token-color-foreground-disabled` applied to the icon.
> 
> Screenshot 2023-08-04 at 2 00 27 PM

Thanks! For the disabled version, it uses a background image for the icon so it doesn't inherit the font color which I forgot about and overlooked originally. I've now fixed it by changing the background image to use the disabled color. (Build will take a couple minutes to see the update in case you re-check too soon.)

packages/components/app/styles/components/form/file-input.scss
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// https://developer.mozilla.org/en-US/docs/Web/CSS/::file-selector-button
+// https://caniuse.com/?search=file-selector-button
+
+// Button
+.hds-form-file-input::file-selector-button {
+  min-height: 36px;
+  margin-right: 16px;
+  margin-inline-end: 16px;
```
didoo, line 30: I see you have added both. Any specific reason? (if yes, then probably better to add a comment to explain why)

packages/components/app/styles/components/form/file-input.scss
```diff
@@ -0,0 +1,98 @@
...
+  padding: 7px 16px 7px 37px; // padding in code altered to match Figma visual design
+  color: var(--token-color-foreground-primary);
+  font: inherit;
+  background-color: var(--token-color-surface-faint);
+
```
didoo, line 35:
```suggestion
```

packages/components/app/styles/components/form/file-input.scss
```diff
@@ -0,0 +1,98 @@
...
+  // upload icon
+  background-image: url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' width='16' height='16' fill='none' viewBox='0 0 16 16'%3E%3Cg fill='%233b3d45'%3E%3Cpath d='M4.24 5.8a.75.75 0 001.06-.04l1.95-2.1v6.59a.75.75 0 001.5 0V3.66l1.95 2.1a.75.75 0 101.1-1.02l-3.25-3.5a.75.75 0 00-1.101.001L4.2 4.74a.75.75 0 00.04 1.06z'/%3E%3Cpath d='M1.75 9a.75.75 0 01.75.75v3c0 .414.336.75.75.75h9.5a.75.75 0 00.75-.75v-3a.75.75 0 011.5 0v3A2.25 2.25 0 0112.75 15h-9.5A2.25 2.25 0 011 12.75v-3A.75.75 0 011.75 9z'/%3E%3C/g%3E%3C/svg%3E");
+  background-repeat: no-repeat;
+  background-position: 15px 50%;
+  background-size: var(--token-form-text-input-background-image-size);
+
```
didoo, line 41:
```suggestion
```

packages/components/app/styles/components/form/file-input.scss
```diff
@@ -0,0 +1,98 @@
...
+  border: $hds-button-border-width solid transparent;
+  border-color: var(--token-color-border-strong);
```
didoo, line 43:
```suggestion
  border: $hds-button-border-width solid var(--token-color-border-strong);
```

didoo (approved):
Let a couple of suggestions, but overall is good to go live. Well done! 👍

packages/components/app/styles/components/form/file-input.scss
KristinLBradley: I'll test if it's doing anything then delete if not, thanks.

KristinLBradley: I'll test if it's doing anything then delete if not, thanks.

packages/components/app/styles/components/form/file-input.scss
KristinLBradley: I did as much as I could with Safari. It is respecting the font size but I haven't found a way to add spacing between the button and the status message text.

alex-ju (approved):
Well done! 🚀

andgen404 (approved):
LGTM 🚀

alex-ju (approved):

PR #1537, opened 2023-07-28
Turn `onClose` callback tests on

alex-ju:
### :pushpin: Summary

Turn `onClose` callback tests on

These two tests were disabled as they proved to be flaky in the past (possibly due to a race condition), but they now seem to behave well, so we're turning them back on.

### :link: External links

Jira ticket: [HDS-1493](https://hashicorp.atlassian.net/browse/HDS-1493)

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


meirish:
Would it be possible to add a test that has a component with an action that gets passed to the modal too? We've found that we often have to guard against `isDestroyed` / `isDestroying` in `onClose` Modal usage, and I'm wondering if that's something that could be detected in the Modal.

alex-ju:
> Would it be possible to add a test that has a component with an action that gets passed to the modal too? We've found that we often have to guard against `isDestroyed` / `isDestroying` in `onClose` Modal usage, and I'm wondering if that's something that could be detected in the Modal.

Sure I can look into that in a separate PR as it may require more than testing. I suspect an action that takes a bit of time would be the most relevant. If you have `cloud-ui` examples at hand it would be great to try and replicate, otherwise no worries, I will dig into the code and look for some myself.

MelSumner (approved):
:shipit:

KristinLBradley (approved):

alex-ju:
> @alex-ju I think we need to revert this change for at least the Flyout as it's flaking quite a bit ([example](https://github.com/hashicorp/design-system/actions/runs/5753907381/job/15598208072?pr=1546#step:5:500))

Stanley's change in #1549 may help, so I'll hold from reverting for a bit longer.

PR #1540, opened 2023-07-28
[SPIKE] Explore alternate pattern layout

jorytindall:
### :pushpin: Summary

This is a bit of an exploration around how to reorganize the content within the filter patterns documentation and how to break the content down a bit more so it's easier to digest.

This more or less mimics the foundations page. 

### 🔧 Details

In this PR I've:

- Reorganized the content so that Filter patterns lives in it's own section
- Added this content to the sidebar on the Patterns template page
- Mimicked the folder structure for other content

What still needs to be addressed/open questions:

Without changing the logic for how the table of contents and "sections" are generated, Filter patterns now lives in it's own folder outside of patterns more broadly. I don't think this is the correct path, but I don't understand enough about how the controller works to change how nested directories are captured (yet, at least).

Now that each "area" of the filter pattern documentation lives on it's own, additional work needs to occur to further break down this content into tabs

I added some quick updates in imagery and metadata to showcase each section, but this will likely need to be revisited.


didoo:
@jorytindall maybe worth a sync meeting (with who's interested/wants to participate)? would be easier to discuss pros/cons of the different options and brainstorm alternatives (it's mainly an IA/content organization problem, but with some technical limitations due to how routing works in Ember)

heatherlarsen:
+1 to what @didoo said. I think learning more about the technical limitations and working through this synchronously would be easier.

didoo:
@jorytindall what about closing this PR at this point? we can always reuse the code as starting point/inspiration, if and when we will need it.

jorytindall:
@didoo totally forgot about this 🤦‍♂️, yeah I think it can be closed!

PR #1543, opened 2023-08-02
add config and initializer to defer loading of icon sprite

Dhaulagiri:
### :pushpin: Summary

If merged this PR will offer a new way to "lazy" load the ember-flight-icon sprite _not_ in the index.html as it does currently.

[Updated docs](https://hds-website-git-br-efi-sprite-hashicorp.vercel.app/icons/usage-guidelines?tab=code#deferred-loading)

There is a more context in the linked RFC below, but effectively what this PR does is introduce a backwards compatible change that allows a consumer to opt into loading the sprite as part of their app bundle vs having it included in the index.html file. It was specifically implemented in a way that if we were to want to remove this feature it would not break any consuming code (those users would revert to the default).

A critical detail for why we don't make this new way the only option is that it doesn't work in fastboot mode which adversely impacts the Helios website.

### Testing
To verify the change you can inspect the showcase and website preview apps and see the difference:

* Showcase: sprite is loaded as part of a `chunk.js` file
* Website: sprite is loaded in index.html

**In both cases you should see `flight-sprite-container` in the DOM only once.**

#### Percy

~~For testing purposes I have included a commit that demonstrates this in action in our showcase app. Because this breaks our Percy tests for reasons out of scope for this PR, I will remove it before merging. _Ideally_ we could use this as a way to implicitly test that this behavior continues to work with future changes, but I was unable to think of a way around this that didn't involve introducing logic to solve this problem in the code we ship for the addon which felt undesirable to me.~~

I was able to solve this and fix the flakey test problem at the same time.

### :link: External links

* Jira ticket: [HDS-2159](https://hashicorp.atlassian.net/browse/HDS-2159)
* [DS-049](https://docs.google.com/document/d/1V9plDdEMuLxrJOOqNUVEIl63uyV2__pq785EZKn_VL8/edit)

***

### 👀 Reviewer's checklist:

[HDS-2159]: https://hashicorp.atlassian.net/browse/HDS-2159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/docs/icons/usage-guidelines/partials/code/how-to-use.md
+++ b/website/docs/icons/usage-guidelines/partials/code/how-to-use.md
@@ -15,6 +15,25 @@ yarn add @hashicorp/ember-flight-icons
 Because this addon exposes a `data-test-icon` helper, we recommend installing [`ember-test-selectors`](https://github.com/simplabs/ember-test-selectors) which strips out all `data-test-*` attributes for production builds.
 !!!
 
+
+#### Deferred loading
+
+By default, the SVG sprite will be injected into your application's `index.html` file. If you would like this to happen later as part of your app bundle you can set the `lazyEmbed` flag to `true` in the `emberFlightIcons` object in your app's `config/environment.js` file:
+
+```js
+module.exports = function(environment) {
+  const ENV = {
+    // your other config
+    ...
+    emberFlightIcons: {
+      lazyEmbed: true,
+    },
+  };
+}
+```
+
+For more information on why this may be helpful in certain scenarios, see [DS-049 - Improve Ember Flight Icons Loading Performance](https://go.hashi.co/rfc/ds-049).
+
 ### Using icons in React apps
 
 To use icons in a React application, install the `@hashicorp/flight-icons` package and import the icons as either inline SVGs or as a standalone React/SVG component. 
--- a/.changeset/pretty-jobs-destroy.md
+++ b/.changeset/pretty-jobs-destroy.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/ember-flight-icons": minor
+---
+
+Add opt in flag to allow consumers to move sprite loading out of index.html
--- a/packages/components/tests/dummy/config/environment.js
+++ b/packages/components/tests/dummy/config/environment.js
@@ -18,6 +18,10 @@ module.exports = function (environment) {
         // e.g. EMBER_NATIVE_DECORATOR_SUPPORT: true
       },
     },
+    // enabled here to aid in testing this feature within our component library
+    emberFlightIcons: {
+      lazyEmbed: true,
+    },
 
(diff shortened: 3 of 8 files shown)


website/docs/icons/usage-guidelines/partials/code/how-to-use.md
```diff
@@ -15,6 +15,25 @@ yarn add @hashicorp/ember-flight-icons
...
+
+#### Deferred loading
+
+By default, the SVG sprite will be injected into your application's `index.html` file. If you would like this to happen later as part of your app bundle you can set the `lazyEmbed` flag to `true`` in the `emberFlightIcons` object in your app's `config/environment.js` file:
+
+```js
+module.exports = function(environment) {
+  const ENV = {
+    // your other config
+
```
didoo, line 27:
```suggestion
  ...
```

website/docs/icons/usage-guidelines/partials/code/how-to-use.md
```diff
@@ -15,6 +15,25 @@ yarn add @hashicorp/ember-flight-icons
...
+    // your other config
+  
+    emberFlightIcons: {
+      lazyEmbed: true,
+    },
+  };
+}
+```
+
+For more information on why this may be helpful in certain scenarios, see [DS-049](https://go.hashi.co/rfc/ds-049).
```
didoo, line 35:
```suggestion
For more information on why this may be helpful in certain scenarios, see [DS-049 - Improve Ember Flight Icons Loading Performance](https://go.hashi.co/rfc/ds-049).
```

alex-ju (approved):
Smart to use the two apps to showcase different behaviours – seems to be working well! 👏

Dhaulagiri:
🤔  interesting that [this](https://github.com/hashicorp/design-system/actions/runs/5798846128/job/15717958745?pr=1543#step:5:826) modal test seems to fail semi-regularly in this PR

alex-ju:
> 🤔 interesting that [this](https://github.com/hashicorp/design-system/actions/runs/5798846128/job/15717958745?pr=1543#step:5:826) modal test seems to fail semi-regularly in this PR

yeah, I looked that up too, it's a CSS test selector.. 😕

Dhaulagiri:
@alex-ju it's a bit puzzling but I was able to come up with a solution in 13836a68ce73d51047cd6a659996c63c329664bb for the Percy problem that seems to have also fixed the test flakiness . Conceptually this _kinda_ makes sense that loading the sprite more directly in the testing container reduces risks for timing out trying to load it later but I'm not 100% convinced either 🤷

MelSumner (approved):
:shipit: 👍

PR #1549, opened 2023-08-04
use `@ember/test-waiters` to instruct Ember to wait for modal `close`

fivetanley:
### :pushpin: Summary

Makes tests less flakey in CI for consumers of the design system

This PR instructs Ember's test suite to wait for the `<dialog>`'s `close` event.

In one of our product's test suites, we were experiencing some flakiness with tests randomly failing when using modals in tests. This is due to Chrome/other browsers queing the `<dialog>`'s `close` event. From the [HTML Spec][1]:

> Queue an element task on the user interaction task source given the subject
> element to fire an event named close at subject.

This means that the `close` event could fire at the end of the current loop or at the beginning of the next turn of the event loop, or in other words, after our test code has run and waited for side effects. Because Ember is unaware of these events, Ember's test infrastructure isn't waiting for the `close` event to fire when using `await click(selector)`/etc from [@ember/test-helpers][2]. We can't depend on the browser to run these events in a particular order as other events may have been queued first.

In order to help consumers of the design system use this component successfully in their application without flakiness or manual polling logic, we can take advantage of Ember's [@ember/test-waiters library][3], which will [eventually become a standard part of Ember.js][4].

Before we close the `<dialog>` element, we register a `waiter` from `@ember/test-waiters` and then resolve it when the close event fires.

To reduce the impact on production builds, we use `ember-cli-babel`'s [General Purpose Env Flags][5] to strip out the instrumentation code during production builds.

[1]: https://html.spec.whatwg.org/multipage/interactive-elements.html#dom-dialog-close
[2]: https://github.com/emberjs/ember-test-helpers
[3]: https://github.com/emberjs/ember-test-waiters
[4]: https://github.com/emberjs/rfcs/blob/master/text/0581-new-test-waiters.md
[5]: https://github.com/babel/ember-cli-babel#general-purpose-env-flags

N/A

### :link: External links

* https://hashicorp.slack.com/archives/C03JP1TCGQM/p1691097709962489

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/modal/index.js
+++ b/packages/components/addon/components/hds/modal/index.js
@@ -8,6 +8,15 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { assert } from '@ember/debug';
 import { getElementId } from '@hashicorp/design-system-components/utils/hds-get-element-id';
+import { buildWaiter } from '@ember/test-waiters';
+import { DEBUG } from '@glimmer/env';
+
+let waiter;
+
+// Notice: this code will get stripped out in production builds (DEBUG evaluates to `true` in dev/test builds, but `false` in prod builds)
+if (DEBUG) {
+  waiter = buildWaiter('@hashicorp/design-system-components:modal');
+}
 
 export const DEFAULT_SIZE = 'medium';
 export const DEFAULT_COLOR = 'neutral';
@@ -156,7 +165,19 @@ export default class HdsModalIndexComponent extends Component {
   }
 
   @action
-  onDismiss() {
+  async onDismiss() {
+    // allow ember test helpers to be aware of when the `close` event fires
+    // when using `click` or other helpers from '@ember/test-helpers'
+    // Notice: this code will get stripped out in production builds (DEBUG evaluates to `true` in dev/test builds, but `false` in prod builds)
+    if (DEBUG && this.element.open) {
+      let token = waiter.beginAsync();
+      let listener = () => {
+        waiter.endAsync(token);
+        this.element.removeEventListener('close', listener);
+      };
+      this.element.addEventListener('close', listener);
+    }
+
     // Make modal dialog invisible using the native `close` method
     this.element.close();
 
--- a/.changeset/pretty-ways-rule.md
+++ b/.changeset/pretty-ways-rule.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`HDS::Modal` - reduce test flakiness around closing modals when using @ember/test-helpers
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -40,6 +40,7 @@
   },
   "dependencies": {
     "@ember/render-modifiers": "^2.0.5",
+    "@ember/test-waiters": "^3.0.2",
     "@hashicorp/design-system-tokens": "^1.7.0",
     "@hashicorp/ember-flight-icons": "^3.0.8",
     "dialog-polyfill": "^0.5.6",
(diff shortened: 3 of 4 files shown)


packages/components/addon/components/hds/modal/index.js
```diff
@@ -8,6 +8,14 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { assert } from '@ember/debug';
 import { getElementId } from '@hashicorp/design-system-components/utils/hds-get-element-id';
+import { buildWaiter } from '@ember/test-waiters';
+import { DEBUG } from '@glimmer/env';
+
+let waiter;
+
+if (DEBUG) {
```
fivetanley, line 16: With Ember-CLI's General Purpose ENV flags, this code will get stripped out in production builds

packages/components/addon/components/hds/modal/index.js
```diff
@@ -8,6 +8,14 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { assert } from '@ember/debug';
 import { getElementId } from '@hashicorp/design-system-components/utils/hds-get-element-id';
+import { buildWaiter } from '@ember/test-waiters';
+import { DEBUG } from '@glimmer/env';
```
fivetanley, line 12: With Ember-CLI's General Purpose ENV flags, this code will get stripped out in production builds

packages/components/addon/components/hds/modal/index.js
```diff
@@ -156,7 +164,18 @@ export default class HdsModalIndexComponent extends Component {
   }
 
   @action
-  onDismiss() {
+  async onDismiss() {
+    // allow ember test helpers to be aware of when the `close` event fires
+    // when using `click` or other helpers from '@ember/test-helpers'
+    if (DEBUG && this.element.open) {
```
fivetanley, line 170: With Ember-CLI's General Purpose ENV flags, `DEBUG` will be replaced with `false`, and this code will get stripped out in production builds by the minifier

fivetanley: Checking if the element is open is required, otherwise we'll set up an event listener that will never fire if the modal is already closed.

jarrodn-au (approved):
Thanks for: the detailed explanation, dealing with such an edge case 😅, reducing test flakiness. Excellent!

packages/components/addon/components/hds/modal/index.js
fivetanley: `DEBUG` evaluates to `true` in dev/test builds, but `false` in prod builds

packages/components/addon/components/hds/modal/index.js
didoo, line 16:
```suggestion
// Notice: this code will get stripped out in production builds (DEBUG evaluates to `true` in dev/test builds, but `false` in prod builds)
if (DEBUG) {
```

packages/components/addon/components/hds/modal/index.js
didoo, line 170:
```suggestion
    // Notice: this code will get stripped out in production builds (DEBUG evaluates to `true` in dev/test builds, but `false` in prod builds)
    if (DEBUG && this.element.open) {
```

didoo (approved):
Thanks for this PR ❤️

I have left a couple of suggestion about adding comments (it may not be obvious to other developers, myself included 😀)

alex-ju (approved):
Thank you for the detailed explanation and for taking the time to debug this! 🙌  Before merging, it would be good if you could add a changelog entry via [`yarn changeset`](https://github.com/hashicorp/design-system/blob/main/README.md?plain=1#L71) (could be as simple as: `Hds::Modal` - fix flaky tests for on `close` event, or however you see fit).

PR #1553, opened 2023-08-07
Update percy reference to permalink

Dhaulagiri:
The previous link was attached to a line that has moved over time, so this uses GitHub's permalink feature to anchor to the line we care about which is specifically about the testing container


alex-ju (approved):

didoo (approved):
👍

PR #1562, opened 2023-08-08
Update `Button` examples to not use copy functionality

jorytindall:
### :pushpin: Summary

In working on something else, I noticed the "How to use" examples for the `Button` showcase "copy" functionality and copying to the clipboard. Since we've released the copy components recently, it seems like these examples might guide users in the wrong direction and use the `Button` component with some custom logic to hook into the clipboard.

So I updated them to some real-ish world examples.

One additional change I made as part of the backing class was to change the `console.log` to a browser alert. Open to feedback whether this is annoying or not, but it seemed liked slightly more explicit and realistic example.

**Question:** Does it make sense to link to the Copy button somewhere in here?

Before
Screenshot 2023-08-08 at 4 26 33 PM

After
Screenshot 2023-08-08 at 4 27 02 PM

***

### 👀 Reviewer's checklist:

### DIFF
--- a/website/docs/components/button/index.js
+++ b/website/docs/components/button/index.js
@@ -11,4 +11,9 @@ export default class Index extends Component {
   copyToClipboard() {
     console.log('Clicked "Copy to clipboard" button!');
   }
+
+  @action
+  alertOnClick() {
+    alert('Hello from Helios!');
+  }
 }
--- a/website/docs/components/button/index.js
+++ b/website/docs/components/button/index.js
@@ -8,11 +8,6 @@ import { action } from '@ember/object';
 
 export default class Index extends Component {
   @action
-  copyToClipboard() {
-    console.log('Clicked "Copy to clipboard" button!');
-  }
-
-  @action
   alertOnClick() {
     alert('Hello from Helios!');
   }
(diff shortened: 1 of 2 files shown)


heatherlarsen (approved):
Good catch! The "Alert me" one doesn't feel as problematic as I thought it would because it's only on one of the buttons, and it's pretty clear from looking at the code what to expect from clicking on it. Also, do people actually click on these buttons?

jorytindall:
> do people actually click on these buttons

Who knows! We could probably attach a Fathom event to them, but that is likely _way_ overkill

website/docs/components/button/index.js
```diff
@@ -11,4 +11,9 @@ export default class Index extends Component {
   copyToClipboard() {
```
didoo, line 11: This action can be removed now.

website/docs/components/button/partials/code/how-to-use.md
```diff
@@ -132,5 +132,5 @@ If using an `@href` or `@route` and needing to disable the component, you’ll n
 !!!
 
 ```handlebars
-<Hds::Button @text="Copy to clipboard" disabled />
+<Hds::Button @text="Alert me" disabled />
```
didoo, line 135: probably I would go the whole hog and add an event handler also here to show that it doesn't fire on a disable button:
```suggestion
<Hds::Button @text="Alert me" disabled {{on "click" this.alertOnClick}} />
```

didoo (commented):
Left a couple of suggestions

MelSumner (approved):
:shipit:

PR #1573, opened 2023-08-14
Add clarity about typographic inheritance for `Link::Inline`

jorytindall:
### :pushpin: Summary

If merged, this PR will add clarity around what is inherited by the `Link::Inline` component. In this case, typographic and text styles, not color.

Additionally, this updates the image examples to _actually_ use the correct color in context.

### :link: External links

Jira ticket: [HDS-2383](https://hashicorp.atlassian.net/browse/HDS-2383)

***

[HDS-2383]: https://hashicorp.atlassian.net/browse/HDS-2383?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


website/docs/components/link/inline/partials/guidelines/guidelines.md
```diff
@@ -14,7 +14,7 @@
 
 The Inline Link is only published as a component in Ember, there is no corresponding component in Figma. To achieve the same results, apply **text** and **color** styles from the [Foundations library](https://www.figma.com/file/oQsMzMMnynfPWpMEt91OpH/HDS-Product---Foundations?type=design&node-id=2130%3A2&t=6XBReWOxMRTiW1Iv-1) to your text.
 
-The Ember component inherits the styles of the text around it. Depending on the surrounding text, the link style in Figma may need to be overridden to ensure that the link and body text are the same.
+The Ember component inherits the typographic styles of the text around it (font, weight, etc). Depending on the surrounding text, the link style in Figma may need to be overridden to ensure that the link and body text are the same.
```
KristinLBradley, line 17: This is helpful but since the consumer confusion this was in response to was around color it may be good to also specifically mention that it does _not_ inherit color.

KristinLBradley (commented):
This looks good but I would specifically mention that color is _not_ inherited as well.

website/docs/components/link/inline/partials/guidelines/guidelines.md
jorytindall: I had considered that and initially left it out since the section on color is directly adjacent. But, I guess since that was the case before and it was still missed or unclear it probably makes sense to include it. I've reworked the sentence a bit to communicate this, LMK what you think!

KristinLBradley (approved):

website/docs/components/link/inline/partials/guidelines/guidelines.md
KristinLBradley: Looks good, thanks for updating it!

website/docs/components/link/inline/partials/guidelines/guidelines.md
```diff
@@ -14,7 +14,9 @@
 
 The Inline Link is only published as a component in Ember, there is no corresponding component in Figma. To achieve the same results, apply **text** and **color** styles from the [Foundations library](https://www.figma.com/file/oQsMzMMnynfPWpMEt91OpH/HDS-Product---Foundations?type=design&node-id=2130%3A2&t=6XBReWOxMRTiW1Iv-1) to your text.
 
-The Ember component inherits the styles of the text around it. Depending on the surrounding text, the link style in Figma may need to be overridden to ensure that the link and body text are the same.
+The Ember component inherits the typographic styles of the text around it (font, weight, etc), while the [color](#color) is determined by the `@color` argument.
+
+Depending on the surrounding text, the link style in Figma may need to be overridden to ensure that the link and body text are the same.
```
heatherlarsen, line 19: > Depending on the surrounding text, the link style in Figma may need to be overridden to ensure that the link and body text are the same.

Can you clarify what you mean by this please? Do you mean the typographic style or the color?

jorytindall: What I'm trying to say here is: to match the style of text around it, the default Link (`Body / 200 / Link` for example) may need to be overwritten to match the text around it because it has a slightly heavier weight. Let me try to wordsmith that a little bit, do you have any thoughts?

heatherlarsen: I think there are two approaches to setting up the links within a paragraph in Figma: 
1. Select the text that needs to be a link and set the decoration to "Underline"
    - this method will already inherit the other typographic styles of the paragraph
2. Select the text that needs to be a link and change the font style to `Body/##/Link`
    - if the paragraph style doesn't use a `medium` weight, this will result in an incorrect weight shift and a misalignment between design and code
    - alternatively, they could update the weight manually
    - this method is more complex to do and document

It is known that a fix for this larger issue would be to offer Link styles for all weights, but this is not yet (if ever) something we plan to tackle. In the meantime, I recommend we advocate for the easier approach of updating the decoration to "Underline." So with that in mind, what about something like: 

> Since a component isn't available for this within Figma, you will need to manually adjust the style of the text. To do this, select the text needed for the link, and change the decoration to "Underline".

website/docs/components/link/inline/partials/guidelines/guidelines.md
```diff
@@ -14,7 +14,9 @@
 
 The Inline Link is only published as a component in Ember, there is no corresponding component in Figma. To achieve the same results, apply **text** and **color** styles from the [Foundations library](https://www.figma.com/file/oQsMzMMnynfPWpMEt91OpH/HDS-Product---Foundations?type=design&node-id=2130%3A2&t=6XBReWOxMRTiW1Iv-1) to your text.
 
-The Ember component inherits the styles of the text around it. Depending on the surrounding text, the link style in Figma may need to be overridden to ensure that the link and body text are the same.
+The Ember component inherits the typographic styles of the text around it (font, weight, etc), while the [color](#color) is determined by the `@color` argument.
```
heatherlarsen, line 17: I'm sorry I didn't catch this before, but it's a little odd that this is on the Guidelines page and listed under the heading "Usage in Figma." It feels like more appropriate content for the Code tab.

website/docs/components/link/inline/partials/guidelines/guidelines.md
```diff
@@ -14,7 +14,9 @@
...
+
+Since a component isn’t available for this within Figma, you will need to manually adjust the style of the text. To do this, select the text needed for the link, and change the decoration to "Underline".
```
heatherlarsen, line 19: Actually, this whole section is now a little odd because there are additional instructions above about achieving this in Figma. Maybe we should take a more holistic look at this section and provide more clarity around how to accomplish this within Figma and update the content to make it flow better. There's already guidance in the stickersheet...maybe you could reuse some of that content?

heatherlarsen (commented):
Sorry, just looked at this more holistically and think is needs a bit of a rethink 😬

website/docs/components/link/inline/partials/guidelines/guidelines.md
jorytindall: Looking through this content with a bit more of a fine-toothed comb, the "Overview" section also contains content that is more relevant to the code implementation. Let me think through this a bit to see what makes sense, I'll move some things around

website/docs/components/link/inline/partials/guidelines/guidelines.md
jorytindall: This was very closely mimicking the content in the stickersheet previously, let me dig in a bit more to see if I can make it flow a bit better.

heatherlarsen: Sorry! I thought there was more specific information about applying the styles, but you're right.

jorytindall:
@heatherlarsen I've reorganized the content a bit, LMK what you think:
- Moved the note about "what is generated" by the component out of the Overview and into the How to use section of the code docs.
    - I haven't pushed the change yet (it's still on my local branch), but if it makes sense to move this content for Inline Link, it also does for Standalone Link.
- Re-wrote the Usage in Figma guidelines to be more step-oriented. This seems like a better approach, though something to note: if a consumer changes the text decoration to `underline`, the text style will detach. Additionally, if a text style is applied, there isn't an option in the type settings to change the text decoration because it's part of the style. I can add a note about this behavior if we think it's necessary.

heatherlarsen (approved):
This is perfect. Much more clear. 👍🏻 Thank you!

PR #1577, opened 2023-08-16
Remove aria-hidden from optional span

DingoEatingFuzz:
### :pushpin: Summary

Removes `aria-hidden` from the optional span in `Form::Indicator` when `@isOptional` is true.

`aria-hidden` was added to the `@isRequired` badge because the input element itself is denoted as required and therefore having it on the badge too is repetitive for SRs and other AT. However, this is not true for the optional span. The aria-hidden attribute was likely added accidentally at the same time.

### :link: External links

Figma file: [if it applies]

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/curvy-rice-watch.md
+++ b/.changeset/curvy-rice-watch.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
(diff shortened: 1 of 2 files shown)


alex-ju (approved):
Good spot – thanks for fixing this!

PR #1581, opened 2023-08-18
`Masked Input` Copy button guidelines 

andgen404:
### :pushpin: Summary

If merged, this PR will add guidelines around the copy button in the masked input component

### :link: External links

Jira ticket: [HDS-2385](https://hashicorp.atlassian.net/browse/HDS-2385)

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable
- [x] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed
- [x] Confirm that A11y tests have been run locally for this component

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-2385]: https://hashicorp.atlassian.net/browse/HDS-2385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1587, opened 2023-08-23
`Hds::Form::MaskedInput` - Add `hasCopyButton` argument

alex-ju:
### :pushpin: Summary

Allow `Hds::Copy::Button` on `Hds::Form::MaskedInput::Base` and `Hds::Form::MaskedInput::Field`

To be able to manage the layout (placing the `Hds::Copy::Button` next to the `Hds::Form::MaskedInput::Base` with `8px` spacing) we needed to introduce an additional wrapper element (`hds-form-masked-input__wrapper`). I was not able to avoid this, but very much open to ideas.

### 🧑‍💻 Impact on consumers

There are no instances of `Hds::Form::MaskedInput` in our consumers' codebases.

### :link: External links

Jira ticket: [HDS-2320](https://hashicorp.atlassian.net/browse/HDS-2320)
Figma file: https://www.figma.com/file/j3b8rHoknu1WX3LLaRZCkj/Copyable-masked-input?type=design&node-id=39498-83754&mode=design&t=bMiZcWFNBp1jYOMz-4
[Related PR with guidelines](https://github.com/hashicorp/design-system/pull/1581)

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable

[HDS-2320]: https://hashicorp.atlassian.net/browse/HDS-2320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/addon/components/hds/form/masked-input/base.hbs
+++ b/packages/components/addon/components/hds/form/masked-input/base.hbs
@@ -2,33 +2,38 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<div class={{this.classNames}} {{style width=@width}}>
-  {{#if @isMultiline}}
-    <Hds::Form::Textarea::Base
-      class="hds-form-masked-input__control"
-      @value={{@value}}
-      @isInvalid={{@isInvalid}}
-      @height={{@height}}
-      id={{this.id}}
-      ...attributes
-    />
-  {{else}}
-    <Hds::Form::TextInput::Base
-      class="hds-form-masked-input__control"
-      @value={{@value}}
-      @isInvalid={{@isInvalid}}
-      id={{this.id}}
-      ...attributes
-    />
+<div class="hds-form-masked-input__wrapper">
+  <div class={{this.classNames}} {{style width=@width}}>
+    {{#if @isMultiline}}
+      <Hds::Form::Textarea::Base
+        class="hds-form-masked-input__control"
+        @value={{@value}}
+        @isInvalid={{@isInvalid}}
+        @height={{@height}}
+        id={{this.id}}
+        ...attributes
+      />
+    {{else}}
+      <Hds::Form::TextInput::Base
+        class="hds-form-masked-input__control"
+        @value={{@value}}
+        @isInvalid={{@isInvalid}}
+        id={{this.id}}
+        ...attributes
+      />
+    {{/if}}
+    <button
+      class="hds-form-masked-input__toggle-button"
+      type="button"
+      aria-label={{this.ariaLabel}}
+      aria-controls={{this.id}}
+      {{on "click" this.onClickToggle}}
+    >
+      <FlightIcon @name={{if this.isMasked "eye" "eye-off"}} @size="16" @isInlineBlock={{false}} />
+      <span class="sr-only" aria-live="polite">{{this.ariaMessageText}}</span>
+    </button>
+  </div>
+  {{#if @hasCopyButton}}
+    <Hds::Copy::Button @text="Copy masked content" @isIconOnly={{true}} @textToCopy="{{@value}}" />
   {{/if}}
-  <button
-    class="hds-form-masked-input__toggle-button"
-    type="button"
-    aria-label={{this.ariaLabel}}
-    aria-controls={{this.id}}
-    {{on "click" this.onClickToggle}}
-  >
-    <FlightIcon @name={{if this.isMasked "eye" "eye-off"}} @size="16" @isInlineBlock={{false}} />
-    <span class="sr-only" aria-live="polite">{{this.ariaMessageText}}</span>
-  </button>
 </div>
--- a/packages/components/addon/components/hds/form/masked-input/base.hbs
+++ b/packages/components/addon/components/hds/form/masked-input/base.hbs
@@ -34,7 +34,7 @@
   {{#if @hasCopyButton}}
     <Hds::Copy::Button
       class="hds-form-masked-input__copy-button"
-      @text="Copy masked content"
+      @text={{this.copyButtonText}}
       @isIconOnly={{true}}
       @textToCopy="{{@value}}"
     />
--- a/packages/components/addon/components/hds/form/masked-input/base.hbs
+++ b/packages/components/addon/components/hds/form/masked-input/base.hbs
@@ -2,38 +2,41 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<div class="hds-form-masked-input__wrapper">
-  <div class={{this.classNames}} {{style width=@width}}>
-    {{#if @isMultiline}}
-      <Hds::Form::Textarea::Base
-        class="hds-form-masked-input__control"
-        @value={{@value}}
-        @isInvalid={{@isInvalid}}
-        @height={{@height}}
-        id={{this.id}}
-        ...attributes
-      />
-    {{else}}
-      <Hds::Form::TextInput::Base
-        class="hds-form-masked-input__control"
-        @value={{@value}}
-        @isInvalid={{@isInvalid}}
-        id={{this.id}}
-        ...attributes
-      />
-    {{/if}}
-    <button
-      class="hds-form-masked-input__toggle-button"
-      type="button"
-      aria-label={{this.ariaLabel}}
-      aria-controls={{this.id}}
-      {{on "click" this.onClickToggle}}
-    >
-      <FlightIcon @name={{if this.isMasked "eye" "eye-off"}} @size="16" @isInlineBlock={{false}} />
-      <span class="sr-only" aria-live="polite">{{this.ariaMessageText}}</span>
-    </button>
-  </div>
+<div class={{this.classNames}} {{style width=@width}}>
+  {{#if @isMultiline}}
+    <Hds::Form::Textarea::Base
+      class="hds-form-masked-input__control"
+      @value={{@value}}
+      @isInvalid={{@isInvalid}}
+      @height={{@height}}
+      id={{this.id}}
+      ...attributes
+    />
+  {{else}}
+    <Hds::Form::TextInput::Base
+      class="hds-form-masked-input__control"
+      @value={{@value}}
+      @isInvalid={{@isInvalid}}
+      id={{this.id}}
+      ...attributes
+    />
+  {{/if}}
+  <button
+    class="hds-form-masked-input__toggle-button"
+    type="button"
+    aria-label={{this.ariaLabel}}
+    aria-controls={{this.id}}
+    {{on "click" this.onClickToggle}}
+  >
+    <FlightIcon @name={{if this.isMasked "eye" "eye-off"}} @size="16" @isInlineBlock={{false}} />
+    <span class="sr-only" aria-live="polite">{{this.ariaMessageText}}</span>
+  </button>
   {{#if @hasCopyButton}}
-    <Hds::Copy::Button @text="Copy masked content" @isIconOnly={{true}} @textToCopy="{{@value}}" />
+    <Hds::Copy::Button
+      class="hds-form-masked-input__copy-button"
+      @text="Copy masked content"
+      @isIconOnly={{true}}
+      @textToCopy="{{@value}}"
+    />
   {{/if}}
 </div>
--- a/packages/components/app/styles/components/form/masked-input.scss
+++ b/packages/components/app/styles/components/form/masked-input.scss
@@ -9,17 +9,15 @@
 
 $hds-form-masked-input-button-size: 24px;
 
-.hds-form-masked-input__wrapper {
-  display: flex;
-  gap: 8px;
-  align-items: flex-start;
-}
-
 .hds-form-masked-input {
   position: relative;
+  display: grid;
+  grid-column-gap: 8px;
+  grid-template-columns: auto min-content;
   width: 100%;
 
   .hds-form-masked-input__control {
+    grid-area: 1 / 1 / 2 / 2;
     padding-right: calc(var(--token-form-control-padding) + $hds-form-masked-input-button-size);
   }
 }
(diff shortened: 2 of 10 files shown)


andgen404 (approved):
Looks good to me, thank you Alex! 🎉

website/docs/components/form/masked-input/partials/code/component-api.md
```diff
@@ -20,6 +20,9 @@ The Masked Input component has two different variants with their own APIs:
   <C.Property @name="isInvalid" @type="boolean" @default="false">
     Applies an “invalid” appearance to the control but doesn’t modify its logical validity.
   </C.Property>
+  <C.Property @name="hasCopyButton" @type="boolean" @default="false">
+    If set to `true` it renders a [`Copy::Button`](/components/copy/button) next to the form control allowing the value of the input to be copied to clipboard.
```
KristinLBradley, line 24:
```suggestion
    If set to `true`, it renders a [`Copy::Button`](/components/copy/button) next to the form control allowing the value of the input to be copied to the clipboard.
```

website/docs/components/form/masked-input/partials/code/component-api.md
```diff
@@ -56,6 +59,9 @@ The Masked Input component has two different variants with their own APIs:
   <C.Property @name="isOptional" @type="boolean" @default="false">
     Appends an `Optional` indicator next to the label text when user input is optional.
   </C.Property>
+  <C.Property @name="hasCopyButton" @type="boolean" @default="false">
+    If set to `true` it renders a [`Copy::Button`](/components/copy/button) next to the form control allowing the value of the input to be copied to clipboard.
```
KristinLBradley, line 63:
```suggestion
    If set to `true`, it renders a [`Copy::Button`](/components/copy/button) next to the form control allowing the value of the input to be copied to the clipboard.
```

website/docs/components/form/masked-input/partials/code/how-to-use.md
```diff
@@ -77,6 +77,15 @@ Something to keep in mind when designing and implementing functionality that mak
 
 !!!
 
+#### Copy button
+
+To allow users to copy the input value to their clipboard set the `@hasCopyButton` argument to `true`.
```
KristinLBradley, line 82:
```suggestion
To allow users to copy the input value to their clipboard, set the `@hasCopyButton` argument to `true`.
```

packages/components/addon/components/hds/form/masked-input/base.hbs
```diff
@@ -2,33 +2,38 @@
...
-    />
-  {{else}}
-    <Hds::Form::TextInput::Base
-      class="hds-form-masked-input__control"
-      @value={{@value}}
-      @isInvalid={{@isInvalid}}
-      id={{this.id}}
-      ...attributes
-    />
+<div class="hds-form-masked-input__wrapper">
```
jorytindall, line 5: FWIW this wrapper is now referred to as `control` in Figma. This is probably a better naming convention in code, just noting the inconsistency here.

packages/components/addon/components/hds/form/masked-input/base.hbs
```diff
@@ -2,33 +2,38 @@
...
+      aria-label={{this.ariaLabel}}
+      aria-controls={{this.id}}
+      {{on "click" this.onClickToggle}}
+    >
+      <FlightIcon @name={{if this.isMasked "eye" "eye-off"}} @size="16" @isInlineBlock={{false}} />
+      <span class="sr-only" aria-live="polite">{{this.ariaMessageText}}</span>
+    </button>
+  </div>
+  {{#if @hasCopyButton}}
+    <Hds::Copy::Button @text="Copy masked content" @isIconOnly={{true}} @textToCopy="{{@value}}" />
```
jorytindall, line 37: Is there any reason why a consumer might want to access this `@text`? Understanding that it's fulfilling the `aria-label` since this is an icon-only variant. I think this is probably good, but if a screen reader user is listening to "Copy masked content" is that sufficient to communicate the actual value that's being copied? (e.g., "Copy Terraform cloud team token"?

I'm probably out of my depth here, but would it be useful to concatenate `@text="Copy {{ariaLabel}}"`? Obviously my Ember string interpolation knowledge is strong 😁

jorytindall (approved):
Left a couple of open-ended comments, but everything is looking _slick_ 🚀

packages/components/addon/components/hds/form/masked-input/base.hbs
alex-ju: We have [that wrapper too](https://github.com/hashicorp/design-system/blob/main/packages/components/addon/components/hds/form/field/index.hbs#L27-L29) in code but it sits outside Base, and part of the `Hds::Form::Field` hence we needed a new one 😞

packages/components/addon/components/hds/form/masked-input/base.hbs
alex-ju: Great observation here! Interpolating the field label (be that, for example, Datadog password) is a good idea – assuming the copy is not terrible (for example, Please type your Datadog password). 

However, there are a few challenges. The `ariaLabel` above is from the visibility toggle button and reads 'Show masked content'/'Hide masked content'. Being at the `::Base` level, we don't have access to the actual field label (only available in the `::Form` component). What we can do, perhaps, is allowing the user to customize this copy and make it more relevant to their usecase.

packages/components/addon/components/hds/form/masked-input/base.hbs
didoo: @alex-ju I want to rubberduck some ideas on this; I'll set up a meeting for tomorrow

packages/components/addon/components/hds/form/masked-input/base.hbs
didoo: > allowing the user to customize this copy and make it more relevant to their usecase

+1

packages/components/tests/dummy/app/templates/components/form/masked-input.hbs
```diff
@@ -106,6 +106,67 @@
...
+  <Shw::Text::H4>Multiline</Shw::Text::H4>
+
+  {{#let (array "base" "invalid" "readonly") as |variants|}}
+    {{#each variants as |variant|}}
+      <Shw::Grid @columns={{3}} as |SG|>
+        {{#each @model.STATES as |state|}}
+          <SG.Item
+            @label="{{capitalize variant}} / {{capitalize state}}"
+            mock-state-value={{state}}
+            mock-state-selector="input"
```
didoo, line 150: should this be `textarea`? (also above). I don't see the `focus` state rendered correctly in the showcase

packages/components/addon/components/hds/form/masked-input/base.hbs
```diff
@@ -2,33 +2,38 @@
...
-  {{else}}
-    <Hds::Form::TextInput::Base
-      class="hds-form-masked-input__control"
-      @value={{@value}}
-      @isInvalid={{@isInvalid}}
-      id={{this.id}}
-      ...attributes
-    />
+<div class="hds-form-masked-input__wrapper">
+  <div class={{this.classNames}} {{style width=@width}}>
```
didoo, line 6: One thing I noticed with this implementation is that now the definition of what `@width` applies to is ambiguous: if as a consumer I say `@width="200px"` am I including the `CopyButton` or not? Not obvious 🤔

packages/components/addon/components/hds/form/masked-input/base.hbs
jorytindall: Got it, yeah makes sense.

packages/components/addon/components/hds/form/masked-input/base.hbs
jorytindall: That might be helpful, but I understand leaving it since it adds an additional argument. Could be something that is optional in that we expose the argument but provide a fallback (e.g., the text you have in there now) in case nothing is passed to it.

alex-ju: yeah, that's what I'm thinking, to allow them to override – will update

packages/components/tests/dummy/app/templates/components/form/masked-input.hbs
alex-ju: indeed, pushed a commit to fix the two

packages/components/addon/components/hds/form/masked-input/base.hbs
alex-ju: Added a `copyButtonText` to allow a custom value here

packages/components/app/styles/components/form/masked-input.scss
```diff
@@ -44,3 +48,9 @@ $hds-form-masked-input-button-size: 24px;
   border-color: transparent;
   cursor: pointer;
 }
+
+.hds-form-masked-input__copy-button {
+  grid-area: 1 / 2 / 2 / 3;
+  align-self: flex-start;
+  margin-left: 8px;
```
alex-ju, line 55: Because the `VisibilityToggle` is always present we can't use the `grid-gap` property so we need to rely on `left-margin` on the `CopyButton` instead

Using `grid-gap: 8px`;
Screenshot 2023-08-25 at 13 31 22

Using `margin-left: 8px`;
Screenshot 2023-08-25 at 13 31 03

packages/components/app/styles/components/form/masked-input.scss
```diff
@@ -11,9 +11,12 @@ $hds-form-masked-input-button-size: 24px;
 
 .hds-form-masked-input {
   position: relative;
+  display: grid;
+  grid-template-columns: auto min-content;
   width: 100%;
 
   .hds-form-masked-input__control {
+    grid-area: 1 / 1 / 2 / 2;
```
didoo, line 19: what do you think using `grid-template-areas` in the parent element, and use the human-readable names in the `grid-area` declarations? it makes everything so much easier to read (and interpret), at least for me

website/docs/components/form/masked-input/partials/code/component-api.md
```diff
@@ -20,8 +20,14 @@ The Masked Input component has two different variants with their own APIs:
...
   </C.Property>
+  <C.Property @name="hasCopyButton" @type="boolean" @default="false">
+    If set to `true`, it renders a [`Copy::Button`](/components/copy/button) next to the form control allowing the value of the input to be copied to the clipboard.
+  </C.Property>
+  <C.Property @name="copyButtonText" @type="string" @default="Copy masked content">
+    Override this value to provide a meaninful `aria-label` for the [`Copy::Button`](/components/copy/button) component.
+  </C.Property>
   <C.Property @name="width" @type="string" @valueNote="any valid CSS width (px, rem, etc)">
-    By default, the input fills the parent container. If a `@width` parameter is provided, the control will have a fixed width.
+    By default, the input fills the parent container. If a `@width` parameter is provided, the control will have a fixed width. When `hasCopyButton` is `true` the width includes the associated copy button.
```
didoo, line 30: 👍

packages/components/app/styles/components/form/masked-input.scss
```diff
@@ -44,3 +48,9 @@ $hds-form-masked-input-button-size: 24px;
   border-color: transparent;
   cursor: pointer;
 }
+
+.hds-form-masked-input__copy-button {
+  grid-area: 1 / 2 / 2 / 3;
+  align-self: flex-start;
```
didoo, line 54: tricky one! I would have not thought about this 😅

packages/components/app/styles/components/form/masked-input.scss
```diff
@@ -36,6 +39,7 @@ $hds-form-masked-input-button-size: 24px;
   position: absolute;
   top: calc(var(--token-form-control-padding) - var(--token-form-control-border-width));
   right: calc(var(--token-form-control-padding) - var(--token-form-control-border-width));
+  grid-area: 1 / 1 / 2 / 2;
```
didoo, line 42: Wait, is this possible, even a thing!!?!! 🤯😱🙀

(grid-area + position absolute relative to the grid area)

didoo: when I suggested to use a grid, I was thinking of using margins for this:
```
grid-area: 1/1/2/2;
margin-top: calc(var(--token-form-control-padding) - var(--token-form-control-border-width));
margin-right: calc(var(--token-form-control-padding) - var(--token-form-control-border-width));
margin-left: auto;
```

didoo (commented):
Added a few (non-blocking) comments.

website/docs/components/form/masked-input/partials/code/component-api.md
KristinLBradley, line 30:
```suggestion
    By default, the input fills the parent container. If a `@width` parameter is provided, the control will have a fixed width. When `hasCopyButton` is `true`, the width includes the associated copy button.
```

packages/components/app/styles/components/form/masked-input.scss
alex-ju: Added a commit https://github.com/hashicorp/design-system/pull/1587/commits/aa496455523d19b7dc68fc355062b8a63e8eb7d3 to convert/simplify these to grid areas

packages/components/app/styles/components/form/masked-input.scss
```diff
@@ -11,9 +11,13 @@ $hds-form-masked-input-button-size: 24px;
 
 .hds-form-masked-input {
   position: relative;
+  display: grid;
+  grid-template-areas: "input copy-button";
+  grid-template-columns: auto min-content;
```
didoo, line 16: I am still unsure if this could (or should?) be `1fr auto` instead. But I don't have strong arguments for this gut feeling. 🤷‍♂️

didoo (approved):

packages/components/app/styles/components/form/masked-input.scss
alex-ju: in this case, the outcome will be the same – will move to `1fr auto` as it seems to be more consistent with what we have elsewhere in the repo

PR #1592, opened 2023-08-24

hashibot-hds:


packages/flight-icons/catalog.json
```diff
@@ -1920,37 +1956,37 @@
...
       "width": 16,
       "height": 16
     },
     {
-      "id": "855:197",
+      "id": "4425:101",
       "fileName": "twitter-color-24",
       "iconName": "twitter-color",
-      "description": "twitter",
+      "description": "",
```
Dhaulagiri, line 1980: any idea where these descriptions went?

andgen404: Not sure, would I be able to still re-input the descriptions here?

andgen404, line 1980:
```suggestion
      "description": "twitter",
```

packages/flight-icons/catalog.json
```diff
@@ -1920,37 +1956,37 @@
...
       "width": 24,
       "height": 24
     },
     {
-      "id": "855:199",
+      "id": "4425:103",
       "fileName": "twitter-color-16",
       "iconName": "twitter-color",
-      "description": "twitter",
+      "description": "",
```
andgen404, line 1989:
```suggestion
      "description": "twitter",
```

packages/flight-icons/catalog.json
```diff
@@ -263,6 +263,42 @@
       "width": 16,
       "height": 16
     },
+    {
+      "id": "4425:66",
+      "fileName": "aws-cdk-24",
+      "iconName": "aws-cdk",
+      "description": "",
```
andgen404, line 270:
```suggestion
      "description": "aws-cdk",
```

packages/flight-icons/catalog.json
```diff
@@ -263,6 +263,42 @@
...
+      "size": "24",
+      "width": 24,
+      "height": 24
+    },
+    {
+      "id": "4425:68",
+      "fileName": "aws-cdk-16",
+      "iconName": "aws-cdk",
+      "description": "",
```
andgen404, line 279:
```suggestion
      "description": "amazon, aws, amazon web services, cdk",
```

packages/flight-icons/catalog.json
```diff
@@ -263,6 +263,42 @@
...
+      "description": "amazon, aws, amazon web services, cdk",
+      "size": "16",
+      "width": 16,
+      "height": 16
+    },
+    {
+      "id": "4425:73",
+      "fileName": "aws-cdk-color-24",
+      "iconName": "aws-cdk-color",
+      "description": "",
```
andgen404, line 288:
```suggestion
      "description": "amazon, aws, amazon web services, cdk",
```

packages/flight-icons/catalog.json
```diff
@@ -263,6 +263,42 @@
...
+      "size": "24",
+      "width": 24,
+      "height": 24
+    },
+    {
+      "id": "4425:75",
+      "fileName": "aws-cdk-color-16",
+      "iconName": "aws-cdk-color",
+      "description": "",
```
andgen404, line 297:
```suggestion
      "description": "amazon, aws, amazon web services, cdk",
```

packages/flight-icons/catalog.json
```diff
@@ -1920,25 +1956,25 @@
       "height": 16
     },
     {
-      "id": "855:190",
+      "id": "4425:94",
       "fileName": "twitter-24",
       "iconName": "twitter",
-      "description": "twitter",
+      "description": "",
```
andgen404, line 1962:
```suggestion
      "description": "twitter",
```

packages/flight-icons/catalog.json
```diff
@@ -1920,25 +1956,25 @@
...
       "width": 24,
       "height": 24
     },
     {
-      "id": "855:192",
+      "id": "4425:96",
       "fileName": "twitter-16",
       "iconName": "twitter",
-      "description": "twitter",
+      "description": "",
```
andgen404, line 1971:
```suggestion
      "description": "twitter",
```

packages/flight-icons/catalog.json
Dhaulagiri: yes, we can't edit them here

Dhaulagiri: otherwise the change here will get overwritten then next time we do an export

packages/flight-icons/catalog.json
```diff
@@ -1920,7 +1956,7 @@
       "height": 16
     },
     {
-      "id": "855:190",
+      "id": "4425:94",
       "fileName": "twitter-24",
       "iconName": "twitter",
       "description": "twitter",
```
jorytindall, line 1962: I'm curious about is whether the description should also be updated for these icons. It would probably be useful to include "x" even though that might increase the noise when searching.

andgen404: Ah, yeah I see what you mean, I could adjust the naming to `twitter, x`.

jorytindall: Right, which may increase it's discoverability, but will also surface other icons (x-hexagon, x-square, etc)

andgen404: Hmm, I feel that leaving it as `twitter` shouldn't divert the searching experience too much for consumers since the brand identity change is still fresh with most users, at least from what i've seen when people discuss `twitters` brand/name change. So keeping as is should be okay, and in return decrease the extra `x` related searches. wdyt?

PR #1593, opened 2023-08-25
randomize flight export branch name

Dhaulagiri:
### :pushpin: Summary

Re-using the same branch name has created some (unnecessary) confusion, so this change uses the [branch-suffix](https://github.com/peter-evans/create-pull-request#alternative-strategy---always-create-a-new-pull-request-branch) option to randomize these moving forward.

Was `hds-icon-export`
Will be something like `hds-icon-export-6qj97jr`

### DIFF
--- a/.github/workflows/open-pull-request-for-icon-update.yml
+++ b/.github/workflows/open-pull-request-for-icon-update.yml
@@ -35,6 +35,7 @@ jobs:
         uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04 # v4.2.3
         with:
           branch: 'hds-icon-export'
+          branch-suffix: 'short-commit-hash'
           commit-message: 'sync & build of flight icons'
           title: 'Updated export of icons from Figma'
           body: ''


.github/workflows/open-pull-request-for-icon-update.yml
```diff
@@ -35,6 +35,7 @@ jobs:
         uses: peter-evans/create-pull-request@2b011faafdcbc9ceb11414d64d0573f37c774b04 # v4.2.3
         with:
           branch: 'hds-icon-export'
+          branch-suffix: 'random'
```
alex-ju, line 38: I remember using `short-commit-hash` in the past but not opinionated

alex-ju (approved):
Nice one! 🙌

.github/workflows/open-pull-request-for-icon-update.yml
Dhaulagiri: i would be fine with that too. it's almost a curse of too many options that all seem to do the same thing.

PR #1594, opened 2023-08-25
🚧 WIP demo of website a11y acceptance test spike

Dhaulagiri:
Finger painting on some ideas of how to DRY and more thoroughly test each page of the website for a11y

### DIFF
--- a/website/tests/acceptance/a11y-test.js
+++ b/website/tests/acceptance/a11y-test.js
@@ -1,0 +1,48 @@
+/* eslint-disable qunit/no-only */
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { module, test } from 'qunit';
+import {
+  visit,
+  setupContext,
+  setupApplicationContext,
+  teardownContext,
+} from '@ember/test-helpers';
+import { setupApplicationTest } from 'website/tests/helpers';
+import { a11yAudit } from 'ember-a11y-testing/test-support';
+import { getOwner } from '@ember/application';
+
+module.only('Acceptance | a11y tests', function (hooks) {
+  setupApplicationTest(hooks);
+
+  let list = [];
+
+  // the idea here is to boot the app and pull out the list of routes to visit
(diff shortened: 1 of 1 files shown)


website/tests/acceptance/a11y-test.js
```diff
@@ -0,0 +1,48 @@
...
+    await visit('/');
+    // debugger;
+    const owner = getOwner(this);
+    const app = owner.lookup('controller:application');
+    list = app.get('model.toc.flat').map((page) => page.pageURL);
+    await teardownContext(this);
+  });
+
+  // a hard coded list does work...
+  list = ['components/badge', 'components/button'];
```
Dhaulagiri, line 37: i don't think such a list exists anywhere, but if we _could_ generate it somehow we could read it in here

website/tests/acceptance/a11y-test.js
```diff
@@ -0,0 +1,48 @@
...
+  list.forEach(async function (f) {
```
didoo, line 38: not sure if it's a problem here, but in general I tend not to use `forEach` when I'm working with async code (I've learned the hard way it can be problematic and... unpredictable) but I prefer to use `for … of` when looping on an array.

website/tests/acceptance/a11y-test.js
didoo: can we read it from the JSON generated at build time? similar to what we do here: https://github.com/hashicorp/design-system/blob/main/website/lib/markdown/index.js#L65-L82

website/tests/acceptance/a11y-test.js
Dhaulagiri: indeed. there is a qunit `.each` mechanism we could try as well

website/tests/acceptance/a11y-test.js
Dhaulagiri: i'm not sure. i tried that but ran into issues doing node things in the context of how these tests are run.

Dhaulagiri: it could run into the same issue though in that the array needs to be defined _before_ the tests start trying to run

PR #1606, opened 2023-08-30
Remove `ember-named-blocks-polyfill`

Dhaulagiri:
### :pushpin: Summary

Remove `ember-named-blocks-polyfill` as all internal consumers of HDS are on Ember 3.25 or later now

### :link: External links

* [Docs](https://github.com/ember-polyfills/ember-named-blocks-polyfill/tree/main#ember-named-blocks-polyfill)

> This addon provides a polyfill for the [Yieldable Named Blocks](https://github.com/emberjs/rfcs/blob/master/text/0460-yieldable-named-blocks.md) feature. **On Ember.js versions with native support for the feature (3.25+), this addon is inert**.

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/.changeset/ninety-nails-report.md
+++ b/.changeset/ninety-nails-report.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Remove `ember-named-blocks-polyfill` as all consumers of HDS are on Ember 3.25 or later now. This can be installed locally if it is still needed.
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -54,7 +54,6 @@
     "ember-composable-helpers": "^4.5.0",
     "ember-focus-trap": "^1.0.2",
     "ember-keyboard": "^8.2.0",
-    "ember-named-blocks-polyfill": "^0.2.5",
     "ember-stargate": "^0.4.3",
     "ember-style-modifier": "^3.0.1",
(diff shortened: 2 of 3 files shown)


.changeset/ninety-nails-report.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
```
Dhaulagiri, line 2: @hashicorp/hds-engineering _technically_ this could be considered a breaking change, but we know that all of our consumers are on Ember versions greater than 3.24 so this shouldn't have a practical risk to it we release it as a patch (or minor). Any objections?

didoo: I would argue that our consumers:
- or they're [already importing it in their package.json files](https://github.com/search?q=%22ember-named-blocks-polyfill%22++user%3Ahashicorp+path%3A*.json&type=Code&ref=advsearch&l=&l=)
- or they don't need it
So probably safe to release it as a patch. unless we want to add it to the batch of things we want to release as "major" (which is also an option I would consider)

In both cases, we should clearly document/publicize the change.

Dhaulagiri: that they can add this themselves if needed is a good point

didoo (approved):

PR #1620, opened 2023-09-05
[SPIKE] Explore fix for `Copy::Snippet` issue with `Dropdown`

didoo:
🚨 **[DO NOT MERGE - SPIKE BRANCH]** 🚨

### :pushpin: Summary

This is a spike branch used to explore a fix for the issue of the `Copy::Snippet` closing the `Dropdown` menu on copy.

Once this works as we want, will be ported to a proper PR and reviewed appropriately.

For details see: 
- Jira ticket for the issue: https://hashicorp.atlassian.net/browse/HDS-2348
    - analysis of root cause of the issue: https://hashicorp.atlassian.net/browse/HDS-2348?focusedCommentId=314288
- Jira ticket for this fix: https://hashicorp.atlassian.net/browse/HDS-2467


MelSumner:
Overall, this seems to address the issue 👍 

I wonder if we should consider a couple of things:

1. Should we file an issue on the clipboard addon since it's widely used and we know we're not the only people to run into this issue? Maybe we can submit a PR to a fix and then it's available for everyone?
2. Should we release this as its own package?

didoo:
> Overall, this seems to address the issue 👍
👍

> 1. Should we file an issue on the clipboard addon since it's widely used and we know we're not the only people to run into this issue? Maybe we can submit a PR to a fix and then it's available for everyone?

@MelSumner there are already _some_ issues filed that at the root are related to this problem. Which can't be solved, as is now (otherwise I would have opened a PR myself) without completely rewriting the library.

> 2. Should we release this as its own package?

Maybe another time :) 

(in any case, our code is already open source, so if anyone want they can use / copy / port it.

alex-ju (commented):
The overall approach seems good 👍  I like that we steer towards using the web APIs

packages/components/addon/modifiers/hds-clipboard.js
```diff
@@ -0,0 +1,176 @@
...
+  if (textToCopy) {
+    try {
+      // notice: the "clipboard-write" permission is granted automatically to pages when they are in the active tab
+      // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write
+      await navigator.clipboard.writeText(textToCopy);
+      console.log('success', textToCopy);
+      return true;
+    } catch (error) {
+      // clipboard write failed
+      console.log(
```
KristinLBradley, line 96: Should the console.logs be removed?

KristinLBradley: Sorry just realized that this is a spike...

didoo:
Closing this now that the exploration is complete. Moved all the code to https://github.com/hashicorp/design-system/pull/1626 (squashing/cleaning up the commits)

PR #1626, opened 2023-09-07
Introduce a custom "copy to clipboard" library to replace `ember-cli-clipboard/clipboard.js` + Update `Copy::Button/Snippet` components

didoo:
### :pushpin: Summary

This PR is a porting (cleanup/squash) of the branch in #1620.
In that spike I have explored how to fix an issue with the `Copy::Snippet` component, used in the `Dropdown::LIstItem::CopyItem` that was causing the dropdown to close immediately after the user clicked on the copy button.

For details about the issue see [this ticket](https://hashicorp.atlassian.net/browse/HDS-2348), and for an analysis of the root cause of the issue see [this comment in the ticket](https://hashicorp.atlassian.net/browse/HDS-2348?focusedCommentId=314288).

in this PR I have:
- removed `ember-cli-clipboard` as dependency
- implemented `hds-clipboard` modifier
    - I've started from `ember-cli-clipboard` + `clipboard.js` as reference, but along the way I've completely refactored and rewrote the entire code and overall logic/architecture, using the web Clipboard API for copying to clipboard
- added “integration” + “unit” tests for the `hds-clipboard` modifier
- changed `Copy::Button/Snippet` components to use our internal `hds-clipboard` helper/modifier
   - in the process I've also added `onSuccess/onError` callbacks
   - notice: I've removed the `@container` argument from the `Button` template code, because it’s not needed anymore, the focus is not moved away by the copy action in the browser now
- updated/fixed integration tests for `Copy::Button/Snippet`
    - I've had to add back the `Sinon` package (as “dev” dependency) to stub `window.navigator.writeText`
- updated showcase for `Copy::Button` to cover more use cases (see link below)
- updated the website documentation for `Copy::Button/Snippet`

🙏 🙏 🙏 **Please reviewers:** 
- check thoroughly the code, possibly line by line
- check that the implementation of the `hds-clipboard` modifier is idiomatic to Ember and follows similar implementations (eg. in Cloud UI)
- question everything, challenge everything I've done
- double check that I've covered all the use cases (positive/negative outcomes, edge cases, etc)
- think if I've missed something, something is not covered

We need to make sure there are no regressions when we release this update

👉 👉 👉 **Previews**:
- [showcase for `Copy::Button`](https://hds-showcase-git-fix-copy-snippet-blur-in-dropdown-hashicorp.vercel.app/components/copy/button)
- [showcase for `Copy::Snippet`](https://hds-showcase-git-fix-copy-snippet-blur-in-dropdown-hashicorp.vercel.app/components/copy/snippet)
- [documentation page for `Copy::Button`](https://hds-website-git-fix-copy-snippet-blur-in-dropdown-hashicorp.vercel.app/components/copy/button?tab=code#component-api)
- [documentation page for `Copy::Snippet`](https://hds-website-git-fix-copy-snippet-blur-in-dropdown-hashicorp.vercel.app/components/copy/snippet?tab=code#component-api)
- documentation page for `Dropdown` - see [this example](https://hds-website-git-fix-copy-snippet-blur-in-dropdown-hashicorp.vercel.app/components/dropdown?tab=code#listitemgeneric) (open the menu and copy one of the values; then compare this with [the same page/example currently in production](https://helios.hashicorp.design/components/dropdown?tab=code#listitemgeneric))

### ⚠️ TODOs
- [x] Check everything works as expected in HCP/Cloud UI
    - Ticket: https://hashicorp.atlassian.net/browse/HDS-2484
    - Test PR: https://github.com/hashicorp/cloud-ui/pull/7480
- [x] Check everything works as expected in Terraform/Atlas
    - Ticket: https://hashicorp.atlassian.net/browse/HDS-2485
    - Test PR: https://github.com/hashicorp/atlas/pull/17025

### :link: External links

- Jira ticket for the issue: https://hashicorp.atlassian.net/browse/HDS-2348
    - analysis of root cause of the issue: https://hashicorp.atlassian.net/browse/HDS-2348?focusedCommentId=314288
- Jira ticket for this fix: https://hashicorp.atlassian.net/browse/HDS-2467

***

### 👀 Reviewer's checklist:

- [x] +1 Percy if applicable

### DIFF
--- a/.changeset/strong-bears-kick.md
+++ b/.changeset/strong-bears-kick.md
@@ -1,0 +1,14 @@
+---
+"@hashicorp/design-system-components": major
+---
+
+- removed `ember-cli-clipboard` as dependency and introduced a custom `hds-clipboard` modifier (using the web [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API))
+- `Copy::Button`
+  - replaced third-party `clipboard` modifier with `hds-clipboard`
+  - removed `@container` argument (not needed anymore)
+  - added `@onSuccess/onError` callbacks
+- `Copy::Snippet`
+  - replaced third-party `clipboard` modifier with `hds-clipboard`
+  - added `@onSuccess/onError` callbacks
+- `Dropdown::ListItem::CopyItem`
+  - the change to the underlying `Copy::Snippet` has fixed an issue with the focus being lost on copy (causing the dropdown to close on copy)
--- a/.changeset/strong-bears-kick.md
+++ b/.changeset/strong-bears-kick.md
@@ -1,11 +1,11 @@
 ---
-"@hashicorp/design-system-components": major
+"@hashicorp/design-system-components": minor
 ---
 
 - removed `ember-cli-clipboard` as dependency and introduced a custom `hds-clipboard` modifier (using the web [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API))
 - `Copy::Button`
   - replaced third-party `clipboard` modifier with `hds-clipboard`
-  - removed `@container` argument (not needed anymore)
+  - removed `@container` argument (not needed anymore, it was used in the third party library as a hack to account for focus trapping and focus shifting)
   - added `@onSuccess/onError` callbacks
 - `Copy::Snippet`
   - replaced third-party `clipboard` modifier with `hds-clipboard`
--- a/packages/components/addon/modifiers/hds-clipboard.js
+++ b/packages/components/addon/modifiers/hds-clipboard.js
@@ -39,7 +39,7 @@ export const getTargetElement = (target) => {
       );
       return;
     }
-  } else if (target instanceof Node && target.nodeType === 1) {
+  } else if (target instanceof Node && target.nodeType === Node.ELEMENT_NODE) {
     targetElement = target;
   } else {
     if (target instanceof NodeList) {
@@ -57,7 +57,10 @@ export const getTargetElement = (target) => {
 
 export const getTextToCopyFromTargetElement = (targetElement) => {
   let textToCopy;
-  if (targetElement instanceof Node && targetElement.nodeType === 1) {
+  if (
+    targetElement instanceof Node &&
+    targetElement.nodeType === Node.ELEMENT_NODE
+  ) {
     if (
       targetElement instanceof HTMLInputElement || // targetElement.nodeName === 'INPUT' ||
       targetElement instanceof HTMLTextAreaElement || // targetElement.nodeName === 'TEXTAREA' ||
@@ -150,8 +153,8 @@ export default class HdsClipboardModifier extends Modifier {
   }
 
   // defines a new `copyToClipboard` action on each click event
-  async #onClick(e) {
-    const trigger = e.delegateTarget || e.currentTarget;
+  async #onClick(event) {
+    const trigger = event.currentTarget;
     const success = await copyToClipboard(this.text, this.target);
 
     // fire the `onSuccess/onError` callbacks (if provided)
--- a/packages/components/addon/modifiers/hds-clipboard.js
+++ b/packages/components/addon/modifiers/hds-clipboard.js
@@ -6,7 +6,7 @@
 import Modifier from 'ember-modifier';
 import { assert } from '@ember/debug';
 
-async function copyToClipboard(text, target) {
+export const getTextToCopy = (text) => {
   let textToCopy;
 
   if (text) {
@@ -23,18 +23,26 @@ async function copyToClipboard(text, target) {
         `\`hds-clipboard\` modifier - \`text\` argument must be a string - provided: ${typeof text}`
       );
     }
-  } else if (target) {
-    let targetElement;
-    if (typeof target === 'string') {
-      targetElement = document.querySelector(target);
+  }
+  return textToCopy;
+};
 
-      assert(
+export const getTargetElement = (target) => {
+  let targetElement;
+  if (typeof target === 'string') {
+    targetElement = document.querySelector(target);
+
+    if (!targetElement) {
+      console.error(
         '`hds-clipboard` modifier - `target` selector provided does not point to an existing DOM node, check your selector string',
         targetElement
       );
-    } else if (target instanceof Node && target.nodeType === 1) {
-      targetElement = target;
-    } else if (target instanceof NodeList) {
+      return;
+    }
+  } else if (target instanceof Node && target.nodeType === 1) {
+    targetElement = target;
+  } else {
+    if (target instanceof NodeList) {
       assert(
         '`hds-clipboard` modifier - `target` argument must be a string or a DOM node - provided: a list of DOM nodes'
       );
@@ -43,7 +51,13 @@ async function copyToClipboard(text, target) {
         `\`hds-clipboard\` modifier - \`target\` argument must be a string or a DOM node - provided: ${typeof target}`
       );
     }
+  }
+  return targetElement;
+};
 
+export const getTextToCopyFromTargetElement = (targetElement) => {
+  let textToCopy;
+  if (targetElement instanceof Node && targetElement.nodeType === 1) {
     if (
       targetElement instanceof HTMLInputElement || // targetElement.nodeName === 'INPUT' ||
       targetElement instanceof HTMLTextAreaElement || // targetElement.nodeName === 'TEXTAREA' ||
@@ -63,12 +77,11 @@ async function copyToClipboard(text, target) {
       // textToCopy = selection.toString();
       // selection.removeAllRanges();
     }
-  } else {
-    assert(
-      '`hds-clipboard` modifier - either a `text` or a `target` argument is required'
-    );
   }
+  return textToCopy;
+};
 
+export const writeTextToClipboard = async (textToCopy) => {
   // finally copy the text to the clipboard using the Clipboard API
   // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API
   if (textToCopy) {
@@ -91,7 +104,24 @@ async function copyToClipboard(text, target) {
   } else {
     return false;
   }
-}
+};
+
+export const copyToClipboard = async (text, target) => {
+  let textToCopy;
+
+  if (text) {
+    textToCopy = getTextToCopy(text);
+  } else if (target) {
+    const targetElement = getTargetElement(target);
+    textToCopy = getTextToCopyFromTargetElement(targetElement);
+  } else {
+    assert(
+      '`hds-clipboard` modifier - either a `text` or a `target` argument is required'
+    );
+  }
+  const success = await writeTextToClipboard(textToCopy);
+  return success;
+};
 
 export default class HdsClipboardModifier extends Modifier {
   didSetup = false;
--- a/packages/components/addon/modifiers/hds-clipboard.js
+++ b/packages/components/addon/modifiers/hds-clipboard.js
@@ -98,11 +98,9 @@ export const writeTextToClipboard = async (textToCopy) => {
     } catch (error) {
       // clipboard write failed
       // this probably never happens (see comment above) or happens only for very old browsers that don't for which `navigator.clipboard` is undefined
-      warn(
-        'copy action failed, please check your browser‘s permissions',
-        textToCopy,
-        error
-      );
+      warn('copy action failed, please check your browser‘s permissions', {
+        id: 'hds-clipboard.write-text-to-clipboard.catch-error',
+      });
       return false;
     }
   } else {
--- a/packages/components/addon/modifiers/hds-clipboard.js
+++ b/packages/components/addon/modifiers/hds-clipboard.js
@@ -4,7 +4,7 @@
  */
 
 import { modifier } from 'ember-modifier';
-import { assert } from '@ember/debug';
+import { assert, warn } from '@ember/debug';
 
 export const getTextToCopy = (text) => {
   let textToCopy;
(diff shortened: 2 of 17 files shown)


packages/components/addon/modifiers/hds-clipboard.js
```diff
@@ -0,0 +1,177 @@
...
+
+export const getTextToCopyFromTargetElement = (targetElement) => {
+  let textToCopy;
+  if (targetElement instanceof Node && targetElement.nodeType === 1) {
+    if (
+      targetElement instanceof HTMLInputElement || // targetElement.nodeName === 'INPUT' ||
+      targetElement instanceof HTMLTextAreaElement || // targetElement.nodeName === 'TEXTAREA' ||
+      targetElement instanceof HTMLSelectElement // targetElement.nodeName === 'SELECT'
+    ) {
+      textToCopy = targetElement.value;
```
meirish, line 66: So I think there may be an issue generally with the modifier - that it only captures the value on init (render, really).

To see this in the showcases when the target is a `textarea` or `input`, change the text and then click the copy button and you'll get the original text from the render when you paste.

meirish: It could also be an issue with wiring up the reactivity (in the demo code) - I haven't gotten far enough to determine the cause.

didoo: @meirish I have double checked and it's working as expected for me 🤔

https://github.com/hashicorp/design-system/assets/686239/af964e79-2db4-4614-936f-3913c17273cb

were you checking on this page? https://hds-showcase-git-fix-copy-snippet-blur-in-dropdown-hashicorp.vercel.app/components/copy/button
or another one?

didoo: BTW, I would be surprised if it doesn't work: what the code does is to read the input/textarea/select `value` (as in DOM `[node].value`) whenever the user clicks the button, which means it's done dynamically, not at rendering.

packages/components/addon/modifiers/hds-clipboard.js
```diff
@@ -0,0 +1,177 @@
...
+    try {
+      // notice: the "clipboard-write" permission is granted automatically to pages when they are in the active tab
+      // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write
+      await navigator.clipboard.writeText(textToCopy);
+      // DEBUG uncomment this for easy debugging
+      // console.log('success', textToCopy);
+      return true;
+    } catch (error) {
+      // clipboard write failed
+      console.log(
```
meirish, line 97: I'd suggest using Ember's `debug` package here: either warn or debug would work, but the advantage of them is that in a production build, their implementations are empty functions so you don't get logs in a production app. Here's the `warn` docs: https://api.emberjs.com/ember/5.2/functions/@ember%2Fdebug/warn

packages/components/addon/modifiers/hds-clipboard.js
meirish: Ha, I just debugged a bit more and I think the examples on the masked inputs are passing `@text` arg, not `@target` so they are working as expected.

packages/components/addon/modifiers/hds-clipboard.js
```diff
@@ -0,0 +1,177 @@
...
+  } else {
+    assert(
+      '`hds-clipboard` modifier - either a `text` or a `target` argument is required'
+    );
+  }
+  const success = await writeTextToClipboard(textToCopy);
+  return success;
+};
+
+export default class HdsClipboardModifier extends Modifier {
```
meirish, line 126: I think this whole thing could be a function modifier, you'd create the handler as a local var inline to close over the args, but that would also let you have a reference to it for cleanup in a function you return.

packages/components/addon/modifiers/hds-clipboard.js
```diff
@@ -0,0 +1,177 @@
...
+      }
+    } else {
+      if (typeof this.onError === 'function') {
+        this.onError(args);
+      }
+    }
+  }
+
+  willDestroy() {
+    this.listener.destroy();
```
meirish, line 175: This won't do anything because `this.listener` will never be defined since  `[addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#return_value)` returns `undefined`. You'd need to call `removeEventListener` with the same handler reference to do the clean up.

packages/components/addon/modifiers/hds-clipboard.js
```diff
@@ -0,0 +1,177 @@
...
+  }
+}
```
meirish, line 177: Here's a quick, untested try at doing a functional modifier:

```suggestion
export default modifier((element, positional, named) => {
  assert(
    '`hds-clipboard` modifier - the modifier must be applied to an element',
    element
  );

  const { text, target, onSuccess, onError } = named;
  const onClick = async (e) => {
    const trigger = e.delegateTarget || e.currentTarget;
    const success = await copyToClipboard(text, target);

    // fire the `onSuccess/onError` callbacks (if provided)
    if (success) {
      if (typeof onSuccess === 'function') {
        onSuccess(args);
      }
    } else {
      if (typeof onError === 'function') {
        onError(args);
      }
    }
  }

  element.addEventListener('click', onClick)

  return () => {
    element.removeEventListener('click', onClick);
  }
});
```

If you wanted to keep it class-based, you could use `[registerDestructor](https://api.emberjs.com/ember/5.2/functions/@ember%2Fdestroyable/registerDestructor)` from the `@ember/destoryable package` to do cleanup.

packages/components/addon/modifiers/hds-clipboard.js
didoo: From discussion with @meirish (for future reference)
- MI: I don’t know why it was a class-based modifier to start
- CR: Because I used this as model: https://github.com/hashicorp/design-system/blob/main/packages/components/addon/modifiers/hds-tooltip.js
Which was based off this: https://github.com/hashicorp/cloud-ui/blob/main/addons/core/addon/modifiers/tooltip.js
- MI: https://github.com/ember-modifier/ember-modifier#function-based-modifiers has some documentation which can show the pros/cons of function-based vs. class-based. (class-based provides more control and allows you to inject a service - but you don't need either of those here imo

Thanks for the feedback/clarification Matthew! 🙏

website/docs/components/copy/snippet/partials/code/component-api.md
```diff
@@ -10,14 +10,17 @@ This component uses [ember-cli-clipboard](https://github.com/jkusa/ember-cli-cli
...
     Constrains text to one line and truncates it based on available width. Text will only be truncated if it does not fit within the available space.
     <br><br>
     There are accessibility concerns if using this feature. See the [Accessibility section](/components/copy/snippet?tab=accessibility) for more information.
   </C.Property>
+  <C.Property @name="onSuccess" @type="function">
+    Callback function invoked (if provided) when the "copy" action succeeds.
+  </C.Property>
+  <C.Property @name="onError" @type="function">
+    Callback function invoked (if provided) when the "copy" action fails.
+  </C.Property>
```
MelSumner, line 23: Previously we decided to provide these hooks but not document them since we didn't have design considerations for them. Just wanted to make sure that we have coordinated with design before shipping this as public API

yarn.lock
```diff
@@ -3534,6 +3534,51 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@sinonjs/commons@npm:^2.0.0":
```
MelSumner, line 3537: Question: can you help me understand why we had to add sinonjs?

MelSumner (commented):
I left a few questions/comments; I think Matt Irish's comments are worth looking into.

packages/components/addon/modifiers/hds-clipboard.js
```diff
@@ -0,0 +1,177 @@
...
+    targetElement = document.querySelector(target);
+
+    if (!targetElement) {
+      console.error(
+        '`hds-clipboard` modifier - `target` selector provided does not point to an existing DOM node, check your selector string',
+        targetElement
+      );
+      return;
+    }
+  } else if (target instanceof Node && target.nodeType === 1) {
```
alex-ju, line 42: for better readability we could write it as
```suggestion
  } else if (target instanceof Node && target.nodeType === Node.ELEMENT_NODE) {
```

packages/components/addon/modifiers/hds-clipboard.js
```diff
@@ -0,0 +1,147 @@
...
+    this.onError = onError;
+
+    if (!this.didSetup) {
+      this.#listenClick(element);
+      this.didSetup = true;
+    }
+  }
+
+  // add a "click" event listener to the element
+  #listenClick(element) {
```
alex-ju, line 148: I see that we're starting to use more and more the standard way for private functions; we should probably make a record to revisit some other components and extend this patterns across the library

packages/components/addon/modifiers/hds-clipboard.js
```diff
@@ -0,0 +1,147 @@
...
+    this.listener = element.addEventListener('click', (e) => this.#onClick(e));
+  }
+
+  // defines a new `copyToClipboard` action on each click event
+  async #onClick(e) {
+    const trigger = e.delegateTarget || e.currentTarget;
```
alex-ju, line 154: is `delegateTarget` still a thing? I've not heard about it since jQuery times, but I may be wrong

packages/components/addon/modifiers/hds-clipboard.js
```diff
@@ -0,0 +1,177 @@
...
+        `\`hds-clipboard\` modifier - \`target\` argument must be a string or a DOM node - provided: ${typeof target}`
+      );
+    }
+  }
+  return targetElement;
+};
+
+export const getTextToCopyFromTargetElement = (targetElement) => {
+  let textToCopy;
+  if (targetElement instanceof Node && targetElement.nodeType === 1) {
```
alex-ju, line 60:
```suggestion
  if (targetElement instanceof Node && targetElement.nodeType === Node.ELEMENT_NODE) {
```

website/docs/components/copy/snippet/partials/code/component-api.md
didoo: Technically, these hooks didn't exist (check the code before), they've been added because I see a potential need to use them for the consumers (for example if they need to test some functionality that involves these components). Also, without these callback functions, is impossible to test the components too.

@heatherlarsen do you see any downsides of exposing these two callback functions (there's no impact on the component in Figma, it's a code-only thing)

yarn.lock
didoo: Sinon is needed to mock/stub the `window.navigator.writeText` functionality, which is not available in the context of Ember testing (the "click" is synthetic, for this reason the browser throws an error when the clipboard is invoked).

See code in this set of tests: https://github.com/hashicorp/design-system/pull/1626/files#diff-8e3f45af3dda12cba8ca2e98080fe6fddf4023f8df76520a64d3c8da68bbde21

alex-ju (commented):
Works as expected – couldn't spot anything off. Left a few minor comments here and there.

.changeset/strong-bears-kick.md
```diff
@@ -0,0 +1,14 @@
...
+"@hashicorp/design-system-components": minor
+---
+
+- removed `ember-cli-clipboard` as dependency and introduced a custom `hds-clipboard` modifier (using the web [Clipboard API](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API))
+- `Copy::Button`
+  - replaced third-party `clipboard` modifier with `hds-clipboard`
+  - removed `@container` argument (not needed anymore, it was used in the third party library as a hack to account for focus trapping and focus shifting)
+  - added `@onSuccess/onError` callbacks
+- `Copy::Snippet`
+  - replaced third-party `clipboard` modifier with `hds-clipboard`
```
KristinLBradley, line 11: Would any consumers potentially use the modifier directly?

packages/components/tests/dummy/app/styles/showcase-pages/copy/button.scss
```diff
@@ -18,10 +18,78 @@ body.components-copy-button {
...
+
+  .shw-component-copy-button-flex-container {
+    display: flex;
+    gap: 8px;
+    align-items: center;
+  }
+
+  .shw-component-copy-button-demo-container {
+    .shw-text-h4 {
+      margin: 16px  0 8px;
```
KristinLBradley, line 63:
```suggestion
      margin: 16px 0 8px;
```

packages/components/tests/dummy/app/styles/showcase-pages/copy/button.scss
```diff
@@ -18,10 +18,78 @@ body.components-copy-button {
...
+    }
+
+    .shw-text-body {
+      margin: 8px  0;
```
KristinLBradley, line 67:
```suggestion
      margin: 8px 0;
```

didoo:
@meirish @alex-ju @MelSumner I've made the changes suggested by Matthew in I've made the suggested changes in this commit: https://github.com/hashicorp/design-system/pull/1626/commits/78e23b1971b31303c65ec39fd9b0a4633ce76e2b 
All the integration tests are passing, and everything works as expected in the showcase.
Have a look and let me know what you think (and approve, if the code is OK for you).

packages/components/addon/modifiers/hds-clipboard.js
didoo: OK, decided to change as you suggested. Thank you for raising the point.

packages/components/addon/modifiers/hds-clipboard.js
```diff
@@ -97,8 +97,9 @@ export const writeTextToClipboard = async (textToCopy) => {
...
     } catch (error) {
       // clipboard write failed
-      console.log(
-        '`hds-clipboard` modifier - an error occurred while writing to the clipboard - check your browser permissions',
+      // this probably never happens (see comment above) or happens only for very old browsers that don't for which `navigator.clipboard` is undefined
+      warn(
+        'copy action failed, please check your browser‘s permissions',
         textToCopy,
         error
       );
```
alex-ju, line 105: [`Ember.warn`s API](https://api.emberjs.com/ember/release/functions/@ember%2Fdebug/warn) is different than `console.log`'s – it seems to be at least requiring an object with an identifier as a third argument.

didoo: Ohhh, I see! Fixed, let's see if the tests are passing now...

alex-ju (approved):
🙌

MelSumner (approved):
👍

KristinLBradley (approved):

PR #1630, opened 2023-09-12
`Hds::SideNav` - add `@isCollapsable` and `@isMinimized` arguments

alex-ju:
### :pushpin: Summary

 - Add `@isCollapsable` to allow users to collapse/expand the `SideNav` (previously only available on narrow viewports)
 - Expose `@isMinimized` argument to allow applications to preserve/control the minimized state across sessions
 - Visual changes to the collapse/expand button and the width of the `SideNav` when minimized

TBA

TBA

### :link: External links

Jira ticket: [HDS-2388](https://hashicorp.atlassian.net/browse/HDS-2388)
Figma file: https://www.figma.com/file/I4gqjXgIvyrX1hBAodx9sC/Collapsible-SideNav-Exploration-%26-Milestones?type=design&node-id=43471-138048&mode=dev

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable


packages/components/app/styles/components/side-nav/main.scss
```diff
@@ -66,56 +72,98 @@
 
 .hds-side-nav__menu-toggle-button {
   position: absolute;
+  top: -2px;
+  left: -1px;
```
didoo, line 76: tip: abstract away this size in a variable (Sass or CSS) so it can be shared across different parts of this CSS implementation (they depend one on another)
screenshot_3080

didoo: (don't worry about the image, I am changing stuff to understand how you built the code to achieve the rounded tab)

packages/components/app/styles/components/side-nav/main.scss
```diff
@@ -66,56 +72,98 @@
...
-  transition-timing-function: var(--hds-app-sidenav-animation-easing);
-  transition-duration: var(--hds-app-sidenav-animation-duration);
-  transition-property: width, height, padding, border-radius, transform;
+  transition:
+    transform var(--hds-app-sidenav-animation-duration)
+    var(--hds-app-sidenav-animation-easing),
+    width var(--hds-app-sidenav-animation-duration)
+    var(--hds-app-sidenav-animation-easing);
+
+  &::before {
```
didoo, line 98: [MEMO] discuss if this could be something else (an SVG?)

packages/components/app/styles/components/side-nav/main.scss
```diff
@@ -66,56 +72,98 @@
 
 .hds-side-nav__menu-toggle-button {
   position: absolute;
+  top: -2px;
+  left: -1px;
   z-index: 1;
+  width: 24px;
+  height: 36px;
```
didoo, line 79: Notice: the size of the border is influencing the position/alignment of the icon (probably it should be more resilient)

screenshot_3081

didoo (commented):
@alex-ju did a first pass and stress-tested the implementation. for now I don't see any blocker, just general comments/suggestions (and a few questions I want to discuss with you, I'll set up a meeting for this)

didoo:
@alex-ju another thing to discuss: in the RFC I see mentioned that this feature is opt-in, but I don't see it in the code (but maybe I have missed it)

packages/components/addon/components/hds/side-nav/index.js
```diff
@@ -56,6 +60,10 @@ export default class HdsSideNavComponent extends Component {
     return this.isResponsive && !this.isDesktop && !this.isMinimized;
   }
 
+  get showToggleButton() {
+    return (this.isResponsive && !this.isDesktop) || this.isCollapsible;
```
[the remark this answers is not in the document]
didoo: For context: in the RFC I see mentioned that this feature is opt-in, but I don't see it in the code (but maybe I have missed it)

packages/components/addon/components/hds/side-nav/index.js
```diff
@@ -39,8 +40,11 @@ export default class HdsSideNavComponent extends Component {
   addEventListeners() {
     document.addEventListener('keydown', this.escapePress, true);
     this.desktopMQ.addEventListener('change', this.updateDesktopVariable, true);
-    // set initial state based on viewport
-    this.updateDesktopVariable({ matches: this.desktopMQ.matches });
+    // if not instantiated as minimized via arguments
+    if (!this.args.isMinimized) {
```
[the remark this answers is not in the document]
didoo: Discussed: there could be a use case when the application preserves the last state of the sidenav for the consumer from the previous session.

packages/components/app/styles/components/side-nav/main.scss
didoo: Discussed, this can be an option to explore if we want.

packages/components/addon/components/hds/side-nav/index.js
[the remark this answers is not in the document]
didoo: Discussed. My bad, I missed, it's `@isCollapsible` 🤷‍♂️

packages/components/app/styles/components/side-nav/main.scss
```diff
@@ -66,56 +72,98 @@
...
+    height: 10px;
+    color: var(--token-color-foreground-high-contrast);
+    border-top: 1px solid var(--token-color-palette-neutral-500);
+    border-left: 1px solid var(--token-color-palette-neutral-500);
+    border-top-left-radius: 5px;
+    box-shadow: 0 -5px 0 var(--token-side-nav-color-surface-primary);
+    content: "";
+  }
 
   &:hover {
```
jorytindall, line 127: Screenshot 2023-09-15 at 5 05 22 PM

Small note on this, the intention with the icon is to "follow" the right edge of the button; the button expands in width on hover, but the icon also maintains the same distance for the right edge of the button. I'm trying to be a bit agnostic in how I refer to the animation because I'm not 100% sure how you have all of this built 😅 , could be a transform, or could be an alignment/flex/whatever property that aligns the icon to the end of the button container.

TLDR, the intention here is for the button to always maintain a 4px distance from the right edge of the button.

jorytindall (commented):
Screenshot 2023-09-15 at 5 12 30 PM

This is looking slick! I'm still waiting for any additional feedback from stakeholders, which I'd hope to wrap by Monday next week. Otherwise, I dropped one initial comment, and noting here that it looks like the scroll bar (or some other element?) is floating a bit on small viewports?

packages/components/app/styles/components/side-nav/main.scss
alex-ju: Ah, good to know! I updated the code for it to behave as described above

jorytindall: Looks great! Thanks for digging into that

packages/components/app/styles/components/side-nav/main.scss
```diff
@@ -66,56 +72,100 @@
...
+    translateX(var(--hds-app-sidenav-width-expanded))
+    translateY(24px);
   cursor: pointer;
-  transition-timing-function: var(--hds-app-sidenav-animation-easing);
-  transition-duration: var(--hds-app-sidenav-animation-duration);
-  transition-property: width, height, padding, border-radius, transform;
+  transition:
+    transform var(--hds-app-sidenav-animation-duration)
+    var(--hds-app-sidenav-animation-easing),
+    width var(--hds-app-sidenav-animation-duration)
```
jorytindall, line 97: I know I'm starting to split hairs here, but I think the animation of the hover state could be tightened up a bit. Rather than the same duration value as the entire collapse animation, can this duration be 150ms? That seems like a happy medium given the size and impact.

alex-ju: Sure, let's give that a go and see how it feels

packages/components/addon/components/hds/side-nav/header/index.hbs
```diff
@@ -4,7 +4,7 @@
 }}
 
 <div class="hds-side-nav-header" ...attributes>
-  <div class="hds-side-nav-header__logo-container">
+  <div class="hds-side-nav-header__logo-container hds-side-nav-hide-when-minimized">
```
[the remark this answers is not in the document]
alex-ju: With the recent design updates, this change was reverted.

packages/components/addon/components/hds/side-nav/index.js
```diff
@@ -74,13 +82,13 @@ export default class HdsSideNavComponent extends Component {
     // add specific class names for the different possible states
     if (this.isDesktop) {
       classes.push('hds-side-nav--is-desktop');
-    } else {
+    } else if (this.isResponsive) {
```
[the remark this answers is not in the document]
alex-ju: I did a small update on this in abe7c900a to improve readability. I played a bit to see if I could simplify the existing arguments but the only options I ended up with required breaking changes, so I took the pragmatic approach of preserving the current arguments, at least for now. Happy to create a ticket for a future refactor.

packages/components/app/styles/components/side-nav/main.scss
alex-ju: I refactor the styling of these elements using CSS variables and tokens so we can easily change the border (`width`, `color`, and `radius`) in future. If you'd like to have a play, here are the associated variables:
```css
  --hds-app-sidenav-border-width
  --hds-app-sidenav-border-color
  --hds-app-sidenav-border-radius
```

alex-ju:
> This is looking slick! I'm still waiting for any additional feedback from stakeholders, which I'd hope to wrap by Monday next week. Otherwise, I dropped one initial comment, and noting here that it looks like the scroll bar (or some other element?) is floating a bit on small viewports?

@jorytindall, I've not been able to replicate this, and although I've not changed the layout of this component (I suspect if there's an issue with scrolling it's something that we'd see in the current implementation too) I'd be keen to look into this (even if it's part of a separate PR). Would you be able to provide me with the browser/os combination that's causing issues?

packages/components/app/styles/components/side-nav/main.scss
didoo: We both did a quick test, and the SVG option is not easy at all:
- it needs SVG SMIL animation (in theory, deprecated in favour of CSS animation)
    - it may not be hardware accelerated (depends on multiple factors, the documentation on the web is not very clear on this)
- it's hard to coordinate animations/transitions with `mousover/mouseleave` and get them in sync with other CSS-based animations
- it may be complex to maintain in the future, if someone work on this code without knowing the details

So we have decided to leave it as "plan B" in case "plan A" finds unexpected roadblocks.

website/docs/components/side-nav/partials/code/component-api.md
```diff
@@ -17,7 +17,15 @@ This is the full-fledged component (responsive and animated).
   <C.Property @name="isResponsive" @type="boolean" @default="true">
     Controls whether the SideNav is responsive to viewport changes. It can be programmatically turned off by passing `false`.
     <br>
-    <em>Notice: even if the @isResponsive parameter is set to false, some JavaScript is executed anyway in the background, and event listeners are attached to some DOM elements (even if this functionality is not used).</em>
+    <em>Notice: even if the `@isResponsive` parameter is set to false, some JavaScript is executed anyway in the background, and event listeners are attached to some DOM elements (even if this functionality is not used).</em>
+  </C.Property>
+  <C.Property @name="isCollapsible" @type="boolean" @default="false">
+    Controls whether the SideNav is collapsible on large viewports. When this argument and `isResponsive` are set to `true` a toggle button will permanently be rendered to collapse and expand the SideNav.
```
didoo, line 23:
```suggestion
    Controls whether the SideNav can be collapsed by the user on large viewports. When this argument and `isResponsive` are set to `true` a toggle button will permanently be rendered to collapse and expand the SideNav.
```

website/docs/components/side-nav/partials/code/component-api.md
```diff
@@ -17,7 +17,15 @@ This is the full-fledged component (responsive and animated).
...
+    <br>
+    <em>Notice: if `@isResponsive` is set to false this argument has no effects.</em>
+  </C.Property>
+  <C.Property @name="isMinimized" @type="boolean" @default="false">
+    Controls if the `SideNav` is rendered collapsed or expanded when initialized. This allows an application to preserve the collapsed/expanded state across sessions.
```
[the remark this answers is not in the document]
didoo: Maybe rewrite my suggestion, I couldn't come up with a clearer formulation of what I intended to say :)

packages/components/app/styles/components/side-nav/main.scss
```diff
@@ -160,6 +205,12 @@
   padding-right: var(--token-side-nav-wrapper-padding-horizontal);
   padding-bottom: 8px; // by design
   padding-left: var(--token-side-nav-wrapper-padding-horizontal);
+
+  .hds-side-nav--is-minimized & {
+    padding-top: var(--token-side-nav-wrapper-padding-vertical-minimized);
```
didoo, line 210: should also these paddding values have a smooth transition? 🤔

packages/components/addon/components/hds/side-nav/index.hbs
```diff
@@ -29,7 +29,7 @@
         {{on "click" this.toggleMinimizedStatus}}
         aria-label={{this.ariaLabel}}
       >
-        <FlightIcon @name={{if this.isMinimized "menu" "x"}} @size="24" @stretched={{true}} />
+        <FlightIcon @name={{if this.isMinimized "chevrons-right" "chevrons-left"}} />
```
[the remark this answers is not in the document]
didoo: (this would mean we could split also the CSS code for `hds-side-nav__menu-toggle-button` to its own file)

packages/components/app/styles/components/side-nav/main.scss
```diff
@@ -66,56 +72,100 @@
 
 .hds-side-nav__menu-toggle-button {
   position: absolute;
+  top: -2px;
+  left: calc(var(--hds-app-sidenav-border-width) * -1);
   z-index: 1;
+  display: flex;
+  box-sizing: content-box;
```
didoo, line 79: maybe add a short comment here, explaining why we're using `content-box` for this element

packages/components/app/styles/components/side-nav/main.scss
```diff
@@ -66,56 +72,100 @@
...
+  padding: calc(var(--token-side-nav-toggle-button-padding-vertical) - var(--hds-app-sidenav-border-width)) var(--token-side-nav-toggle-button-padding-horizontal);
   color: var(--token-color-foreground-high-contrast);
+  text-align: right;
```
didoo, line 82: what is this for?

packages/components/app/styles/components/side-nav/main.scss
```diff
@@ -66,56 +72,100 @@
...
   background: none;
-  border: 1px solid transparent;
+  background-color: var(--token-side-nav-color-surface-primary);
+  border: var(--hds-app-sidenav-border-width) solid var(--hds-app-sidenav-border-color);
+  border-left-color: transparent;
+  border-top-right-radius: var(--hds-app-sidenav-border-radius);
```
didoo, line 87: this variable is used only in this context, not across the entire sidenav. maybve better to rename it to `--hds-app-sidenav-toggle-button-border-radius`?

packages/components/app/styles/components/side-nav/main.scss
```diff
@@ -66,56 +72,100 @@
 
 .hds-side-nav__menu-toggle-button {
```
didoo, line 73: technically this is not anymore a "menu" toggle button, just a "toggle button". maybe rename this class to `hds-side-nav__toggle-button`  (or `hds-side-nav__collapse-toggle-button` if we want to be pedantic)

packages/components/app/styles/components/side-nav/main.scss
```diff
@@ -66,56 +72,100 @@
...
+    var(--hds-app-sidenav-animation-easing);
+
+  &::before {
+    position: absolute;
+    top: calc(var(--hds-app-sidenav-border-radius) * -2);
+    left: calc(var(--hds-app-sidenav-border-width) * -1);
+    box-sizing: border-box;
+    width: calc(var(--hds-app-sidenav-border-radius) * 2);
+    height: calc(var(--hds-app-sidenav-border-radius) * 2);
+    color: var(--token-color-foreground-high-contrast);
```
didoo, line 106: are you sure this is needed? 🤔

didoo (commented):
@alex-ju I haven't checked (yet) the CSS line by line, I intend to do it when it's 100% finalized.
In the meantime I've left a few comments that I hope can help in moving forward the PR to a "ready to merge" state.

packages/components/app/styles/components/side-nav/header.scss
```diff
@@ -60,6 +60,11 @@
   justify-content: center;
   width: var(--token-side-nav-header-home-link-logo-size);
   height: var(--token-side-nav-header-home-link-logo-size);
+
+  .hds-side-nav--is-minimized & {
```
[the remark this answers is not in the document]
alex-ju: Added the SideNav default transition to both the logo size and the SideNav padding.

packages/components/app/styles/components/side-nav/main.scss
alex-ju: Added the `SideNav` default transition values to both the logo size and the wrapper padding in c7f5bca5fadd6cd9e27bdfc44163d2357894f8b8

packages/components/app/styles/components/side-nav/main.scss
alex-ju: leftover, removed in 531c5b78cad742c6695961175c3f5afcd62334e3

packages/components/app/styles/components/side-nav/main.scss
alex-ju: leftover, removed it 531c5b78cad742c6695961175c3f5afcd62334e3

packages/components/app/styles/components/side-nav/header.scss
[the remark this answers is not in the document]
didoo: I _think_ there are some tunings to do, still (unless is intentional)

https://github.com/hashicorp/design-system/assets/686239/1fd5cb77-9864-47af-90c7-88dd43832e7a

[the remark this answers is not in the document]
alex-ju: Indeed, it was more than height/width that needed to be taken care of when animating this – updated to address all changes

packages/components/addon/components/hds/side-nav/index.js
```diff
@@ -11,7 +11,8 @@ import { registerDestructor } from '@ember/destroyable';
 
 export default class HdsSideNavComponent extends Component {
   @tracked isResponsive = this.args.isResponsive ?? true;
-  @tracked isMinimized = this.isResponsive; // we set it minimized by default so that if we switch viewport from desktop to mobile its already minimized
+  @tracked isMinimized = this.args.isMinimized ?? false;
```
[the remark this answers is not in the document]
alex-ju: `isMinimized` used to be an internal variable with meaning only on mobile screens (when `isDesktop` was `false` and `isResponsive was `true`). it was, indeed, set to `true` by default so that if the application starts on a mobile screen the SideNav will be minimized. Now that the collapsible feature may be available on any screen size, we adjusted the meaning of this variable to reflect the minimized state independent of the screen size and made it part of the public API so applications can programmatically set the default state (for state persistence purposes). The behavior on mobile screens is preserved because we do a media query check as part of the constructor of this component.

packages/components/addon/components/hds/side-nav/index.hbs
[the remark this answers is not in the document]
alex-ju: sure, I will look into refactoring this to extract the toggle button into its own component.

packages/components/addon/components/hds/side-nav/index.js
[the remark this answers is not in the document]
didoo: Makes sense. Can you maybe add a short comment to make sure this information is preserved (for future reference)? Thanks

jorytindall:
> > noting here that it looks like the scroll bar (or some other element?) is floating a bit on small viewports?
> 
> @jorytindall, I've not been able to replicate this, and although I've not changed the layout of this component (I suspect if there's an issue with scrolling it's something that we'd see in the current implementation too) I'd be keen to look into this  (even if it's part of a separate PR). Would you be able to provide me with the browser/os combination that's causing issues?

I just did some more testing and can't replicate it either 🤔 . Must have been an isolated issue or you already solved it!

jorytindall:
Keeping up to speed here, with the changes in the animation duration between the collapse button and the collapse/expand, it appears that the button is detached. Probably because it's animation finishes sooner than the expand animation?
Screenshot 2023-09-20 at 12 39 24 PM

alex-ju:
> Keeping up to speed here, with the changes in the animation duration between the collapse button and the collapse/expand, it appears that the button is detached. Probably because it's animation finishes sooner than the expand animation?

Yes, it seems to be more visible in WebKit browsers than V8 and Gecko. Although we only apply the short one (150ms) for changes in padding on the toggle button (for the hover effect), it still has implications in positioning the element, resulting in a slight delay (I've not managed to obtain the gap you pictured above, but there is a small gap when the durations are augmented, e.g. 1500ms). We may need to keep all animations at the duration.

jorytindall:
> We may need to keep all animations at the duration.

If that ends up being the case, the best route _may_ be to split the difference, e.g., use 200ms for both animation durations. Speeding up the collapse a little bit, and slowing down the collapse button interactive animation.

Additionally, for own my sanity, given a user `prefers-reduced-motion` is this accounted for by the collapse/expand occurring immediately with no animation?

alex-ju:
> > We may need to keep all animations at the duration.
> 
> If that ends up being the case, the best route _may_ be to split the difference, e.g., use 200ms for both animation durations. Speeding up the collapse a little bit, and slowing down the collapse button interactive animation.

That's a good suggestion, I find. Pushed a commit to get your feeling on it as well.

> Additionally, for own my sanity, given a user `prefers-reduced-motion` is this accounted for by the collapse/expand occurring immediately with no animation?

Yes, if a user prefers reduced motion the collapse/expand will essentially have no animation/motion. I also checked this as well, to make sure if behaves decently. If you'd like to see it in action I attached a screen-capture with how to enable it on osx.

Screenshot 2023-09-21 at 20 39 01

jorytindall:
> Pushed a commit to get your feeling on it as well.

Looks good to me! 🎉

packages/components/addon/components/hds/side-nav/index.hbs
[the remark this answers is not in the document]
alex-ju: Addressed in c9836dbbb15b5358f63a81161ea2ed094ef0f83a – it helped with setting up regression testing for the various states of the button as well.

packages/components/addon/components/hds/side-nav/index.js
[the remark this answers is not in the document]
alex-ju: Yes, added some clarity around arguments in ae32fba6a70761bb7ad220b29e1db897b208b98e

packages/components/app/styles/components/side-nav/vars.scss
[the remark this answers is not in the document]
alex-ju: Two of them are not in use anymore, so I got rid of those first. Then I moved the wrapper and toggle-button-related ones in 80aea6792a118ee6e6b28551ddb31cc453553e71. The remaining ones are partly used in JS and partly forwarding values or overridden in the reduced motion definition, so there are still a few left.

packages/components/app/styles/components/side-nav/main.scss
alex-ju: Renamed in 862e086318e37a9812cae807303e57546b928e0c

packages/components/app/styles/components/side-nav/main.scss
alex-ju: Renamed in 4571e124df5e4c0ccaa190349f55b30db4484f1a

alex-ju:
@didoo I tried to address the feedback from the last round, so it's good for another review whenever time allows

packages/components/app/styles/components/side-nav/toggle-button.scss
```diff
@@ -0,0 +1,108 @@
...
+  color: var(--token-color-foreground-high-contrast);
+  background: none;
+  background-color: var(--token-side-nav-color-surface-primary);
+  border: var(--token-side-nav-wrapper-border-width) solid var(--token-side-nav-wrapper-border-color);
+  border-left-color: transparent;
+  border-top-right-radius: var(--token-side-nav-toggle-button-border-radius);
+  border-bottom-right-radius: var(--token-side-nav-toggle-button-border-radius);
+  transform:
+    translateX(var(--hds-app-sidenav-width-expanded))
+    translateY(24px);
```
didoo, line 28: see comment above about the possibility of getting rid of the `translateY`

packages/components/app/styles/components/side-nav/toggle-button.scss
```diff
@@ -0,0 +1,108 @@
...
+    transform var(--hds-app-sidenav-animation-duration)
+    var(--hds-app-sidenav-animation-easing),
+    padding var(--hds-app-sidenav-animation-duration)
+    var(--hds-app-sidenav-animation-easing);
+
+  &::before {
+    position: absolute;
+    top: calc(var(--token-side-nav-toggle-button-border-radius) * -2);
+    left: calc(var(--token-side-nav-wrapper-border-width) * -1);
+    box-sizing: border-box;
```
didoo, line 40: a couple of considerations:
- even if I comment this out, nothing changes (which makes sense, since with this implementation the sizing is given by content + padding + border, not by an explicit width/height)
- if you try to change the border width variable, the border is applied internally, not externally, which is not (I suspect) what the designers want

what I suggest to do (not a blocker, just for consideration): 
- get rid of the `box-sizing: border-box;`
- use explicit width/height for the element (see other comment)
I _think_ this _may_ simplify a bit the overall code (but I may be wrong, so take this with a pinch of salt)

packages/components/app/styles/components/side-nav/toggle-button.scss
```diff
@@ -0,0 +1,108 @@
...
+
+    &::after {
+      box-shadow: 0 calc(var(--token-side-nav-toggle-button-border-radius) * -1) 0 var(--token-side-nav-color-surface-interactive-hover);
+    }
+  }
+
+  &:active,
+  &.mock-active {
+    background-color: var(--token-side-nav-color-surface-interactive-active);
+    border-color: transparent;
```
didoo, line 81: this is what happens on `active` state, probably not what we want:
screenshot_3108

didoo (commented):
Left a bunch of comments. Let me know if you want to discuss them in person.
I want to do a `final-last__really-final--v2` review 😀 before approving.

packages/components/app/styles/components/side-nav/toggle-button.scss
alex-ju: I can't seem to be able to replicate this 🤔 
This is what I see.
Screenshot 2023-09-27 at 17 14 02

I tried changing the border color and width but doesn't impact the button. I must be missing something.

alex-ju: ah, I see what you mean – I'll review these attributes as there seem to be some leftovers from different refactors.

.changeset/two-pumas-deny.md
```diff
@@ -0,0 +1,6 @@
+---
+"@hashicorp/design-system-components": minor
+"@hashicorp/design-system-tokens": minor
+---
+
+`SideNav` - add `@isCollapsible` and `@isMinimized` arguments
```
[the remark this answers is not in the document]
alex-ju: Added a brief description for each argument

packages/components/app/styles/components/side-nav/header.scss
```diff
@@ -60,6 +60,14 @@
   justify-content: center;
   width: var(--token-side-nav-header-home-link-logo-size);
   height: var(--token-side-nav-header-home-link-logo-size);
+  transition:
+    all var(--hds-app-sidenav-animation-duration)
```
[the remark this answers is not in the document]
alex-ju: Changed this to explicitly apply the transition to `width` and `height`

packages/components/addon/components/hds/side-nav/index.js
```diff
@@ -56,6 +66,10 @@ export default class HdsSideNavComponent extends Component {
     return this.isResponsive && !this.isDesktop && !this.isMinimized;
   }
 
+  get showToggleButton() {
+    return (this.isResponsive && !this.isDesktop) || this.isCollapsible;
```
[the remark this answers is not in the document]
alex-ju: The suggested alternative is not an equivalent, unfortunately, as it would show the toggle button even when `isCollapsible` is false.

It should read as: show the toggle button if the user is on mobile and `isResponsive` is set to `true' or, regardless of the viewport, if `isCollapsible` is `true`. This is the current logic we want to reflect, I'm not sure we can simplify it further 🤔

packages/components/app/styles/components/side-nav/toggle-button.scss
```diff
@@ -0,0 +1,108 @@
...
+ */
+
+//
+// SIDE-NAV > TOGGLE BUTTON
+//
+
+
+.hds-side-nav__toggle-button {
+  position: absolute;
+  top: -2px;
```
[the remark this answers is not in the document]
alex-ju: No performance reasons, remnant from when the button was moved on both axis.

packages/components/app/styles/components/side-nav/toggle-button.scss
```diff
@@ -0,0 +1,108 @@
...
+  left: calc(var(--token-side-nav-wrapper-border-width) * -1);
+  z-index: 1;
+  display: flex;
+  box-sizing: content-box; // keeps the border outside the element to consistently control the border width across SideNav elements
+  padding: calc(var(--token-side-nav-toggle-button-padding-vertical) - var(--token-side-nav-wrapper-border-width)) var(--token-side-nav-toggle-button-padding-horizontal);
```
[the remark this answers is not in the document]
alex-ju: Removed the padding declarations and tokens and changed the `flex-direction` to be able to keep the icon docked to the right.

packages/components/app/styles/components/side-nav/toggle-button.scss
alex-ju: Removed the box-sizing declaration and the padding and moved to width/height declarations + flex alignment.

alex-ju:
> Left a bunch of comments. Let me know if you want to discuss them in person. I want to do a `final-last__really-final--v2` review 😀 before approving.

@didoo I hope I addressed the comments from the last round of reviews. I also merged Jory's documentation in this PR so it can be reviewed together.

packages/components/tests/integration/components/hds/side-nav/index-test.js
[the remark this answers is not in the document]
alex-ju: This is a great workaround actually – I'll give it a go

didoo (approved):
✅ 🎉
pirate-pirate-ship

packages/components/tests/integration/components/hds/side-nav/index-test.js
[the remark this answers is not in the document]
alex-ju: added a series of tests for "mobile" in 5f3f84f1a0b3ea581797077e2412cdfba1fa211c by overriding the `--hds-app-desktop-breakpoint`. I used inline styling as I couldn't think of another way, but open to suggestions.

packages/components/addon/components/hds/side-nav/index.js
```diff
@@ -38,13 +39,22 @@ export default class HdsSideNavComponent extends Component {
 
   addEventListeners() {
     document.addEventListener('keydown', this.escapePress, true);
+    if (this.isCollapsible) {
+      document.addEventListener('keyup', this.altBracketLeftPress, true);
```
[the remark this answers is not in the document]
alex-ju: Removed now

packages/components/tests/dummy/app/router.js
```diff
@@ -80,4 +80,7 @@ Router.map(function () {
   this.route('overrides', function () {
     this.route('power-select');
   });
+  this.route('demo-app', function () {
```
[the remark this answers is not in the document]
alex-ju: I think there is a real value in having the `demo-app` around. I tried the second suggestion (not rendering the `shw-` elements in the `demo-app`) in 0298bd4f56cb4d339f0d38342917a595f010bbfe. It still shares the same `index.html` that contains the `<body class="showcase-app">` so one override is still needed, but we were able to remove the other ones. Happy to give the first suggestion a go as well, but more likely in a separate PR.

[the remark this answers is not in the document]
didoo: > I think there is a real value in having the demo-app around.

Ok, we agree :)

> I tried the second suggestion (not rendering the shw- elements in the demo-app) in https://github.com/hashicorp/design-system/commit/0298bd4f56cb4d339f0d38342917a595f010bbfe. It still shares the same index.html that contains the <body class="showcase-app"> so one override is still needed, but we were able to remove the other ones. Happy to give the first suggestion a go as well, but more likely in a separate PR.

I would prefer to see this (including your workaround in https://github.com/hashicorp/design-system/commit/0298bd4f56cb4d339f0d38342917a595f010bbfe) done is a separate PR for a few reasons:
- we can discuss during our engineering team sync the options and come up with a shared solution
- I suspect this demo app will play a part in the future, becoming a second "showcase", for dynamic interactions with the components, and more real-use-cases example, so having it done "properly" (whatever that means) will pay us back in the long term
- leaves this PR a bit cleaner, and we decouple it from finding a solution to the `demo-app` problem (that could take longer) so that the PR can be merged straightaway

What do you think?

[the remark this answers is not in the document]
alex-ju: Makes sense! Will cherry-pick them in a separate PR and remove them from here

didoo (approved):
Reviewed latest changes and re-approved

packages/components/tests/integration/components/hds/side-nav/index-test.js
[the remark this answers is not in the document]
didoo: It's a simple workaround. It does what it needs. Works for me 👍

jorytindall (approved):
For some reason the preview of the demo app isn't working for me (all I get is a blank page). The only thing I wanted to confirm was the animation/easing between the toggle button and the SideNav and whether that was either fixed or reverted to the same values to solve that separation issue in the animation rendering. Looking through the code I _think_ they are the same, but could be wrong.

Otherwise, everything looks solid to me 🎉

website/docs/components/side-nav/partials/code/component-api.md
jorytindall, line 28:
```suggestion
    Controls if the SideNav is rendered collapsed or expanded when initialized. This allows an application to preserve the collapsed/expanded state across sessions.
```

Removed backticks for consistency

packages/components/app/styles/components/side-nav/toggle-button.scss
```diff
@@ -0,0 +1,106 @@
...
+  left: calc(var(--token-side-nav-wrapper-border-width) * -1);
+  z-index: 1;
+  display: flex;
+  flex-direction: row-reverse;
+  align-items: center;
+  width: 26px;
+  height: 36px;
+  padding: 0 4px;
+  color: var(--token-color-foreground-high-contrast);
+  background: none;
```
KristinLBradley, line 23: Is this to override background styles being inherited from somewhere?

website/docs/components/side-nav/partials/code/component-api.md
```diff
@@ -17,7 +17,15 @@ This is the full-fledged component (responsive and animated).
   <C.Property @name="isResponsive" @type="boolean" @default="true">
     Controls whether the SideNav is responsive to viewport changes. It can be programmatically turned off by passing `false`.
     <br>
-    <em>Notice: even if the @isResponsive parameter is set to false, some JavaScript is executed anyway in the background, and event listeners are attached to some DOM elements (even if this functionality is not used).</em>
+    <em>Notice: even if the `@isResponsive` parameter is set to false, some JavaScript is executed anyway in the background, and event listeners are attached to some DOM elements (even if this functionality is not used).</em>
```
KristinLBradley, line 20:
```suggestion
    <em>Notice: even if the `@isResponsive` parameter is set to false, some JavaScript is still executed in the background, and event listeners are attached to some DOM elements (even if this functionality is not used).</em>
```

website/docs/components/side-nav/partials/code/component-api.md
KristinLBradley, line 23:
```suggestion
    Controls whether the SideNav is collapsible on large viewports. When this argument and `isResponsive` are set to `true`, a toggle button will permanently be rendered to collapse and expand the SideNav.
```

website/docs/components/side-nav/partials/code/component-api.md
```diff
@@ -17,7 +17,15 @@ This is the full-fledged component (responsive and animated).
...
+  </C.Property>
+  <C.Property @name="isCollapsible" @type="boolean" @default="false">
+    Controls whether the SideNav is collapsible on large viewports. When this argument and `isResponsive` are set to `true` a toggle button will permanently be rendered to collapse and expand the SideNav.
+    <br>
+    <em>Notice: if `@isResponsive` is set to false this argument has no effects.</em>
```
KristinLBradley, line 25:
```suggestion
    <em>Notice: if `@isResponsive` is set to false, this argument has no effect.</em>
```

KristinLBradley (commented):
Added a few suggested text/punctuation edits.

MelSumner:
@alex-ju any reason why the tab order is skipping the switchers/selects in the sidenav?

alex-ju:
> @alex-ju any reason why the tab order is skipping the switchers/selects in the sidenav?

yes, that's not an actual interactive element (it's a div styled to look like an organization select). this element is part of [the demo app work that Cristiano suggested we introduce in a separate PR](https://github.com/hashicorp/design-system/pull/1630#discussion_r1342589319) so we can address it there

packages/components/app/styles/components/side-nav/toggle-button.scss
alex-ju: a reset for the `<button>` element

PR #1624, opened 2023-09-07
Improve documentation for the `Dropdown::ListItem::CopyItem`

didoo:
### :pushpin: Summary

This PR intends to improve the documentation for the `Dropdown::ListItem::CopyItem`, in particular around the possibility of using `@isTruncated` as argument (forwarded to the `Copy::Snippet` component). This follows [a previous conversation](https://hashicorp.slack.com/archives/C02526PLZ9D/p1693307182523119?thread_ts=1693214133.191979&cid=C02526PLZ9D) with @jgwhite.

### :hammer_and_wrench: Detailed description

In this PR I have:
- updated the “Component API” documentation for the `Dropdown::ListItem::CopyItem` subcomponent
- added an example of `Dropdown::ListItem::CopyItem` (also with truncation) to the “How to use” section

 👉👉 👉 Preview:
- How to use: https://hds-website-git-update-dropdown-listitem-copyi-a8da7c-hashicorp.vercel.app/components/dropdown?tab=code#listitemcopyitem
- Component API: https://hds-website-git-update-dropdown-listitem-copyi-a8da7c-hashicorp.vercel.app/components/dropdown?tab=code#listitemcopyitem-1

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2449

***

### 👀 Reviewer's checklist:

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #1631, opened 2023-09-13
`Dropdown::ListItem::CopyItem` - Update defaults for `@color` and `@isTruncated`

didoo:
### :pushpin: Summary

While working on https://github.com/hashicorp/design-system/pull/1624 I noticed a couple of differences between the current vs the previous implementation (see screenshots). 
After a conversation with @heatherlarsen we agreed that the desired behaviour is:
- `color` variant should be `secondary`
- `truncation` should be on by default (but optionally can be turned off)

In this PR I have:
- changed defaults for `@color` (now `secondary`) and `@isTruncated` (now `true`)
- updated showcase for `Dropdown` component
- updated “How to use” and “Component API” documentation for `Dropdown::ListItem::CopyItem`

👉 👉 👉  **Previews**:
- showcase: https://hds-showcase-git-fix-dropdown-copy-item-defaults-hashicorp.vercel.app/components/dropdown
- documentation: https://hds-website-git-fix-dropdown-copy-item-defaults-hashicorp.vercel.app/components/dropdown?tab=code

**Before** (implementation before changes in #1488)
before

**After** (after changes in #1488)
after

**Current PR:**
screenshot_3075

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2490

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable

### DIFF
--- a/packages/components/addon/components/hds/dropdown/list-item/copy-item.hbs
+++ b/packages/components/addon/components/hds/dropdown/list-item/copy-item.hbs
@@ -8,5 +8,5 @@
       class="hds-dropdown-list-item__copy-item-title hds-typography-body-100 hds-font-weight-semibold"
     >{{@copyItemTitle}}</div>
   {{/if}}
-  <Hds::Copy::Snippet @textToCopy={{this.text}} @isTruncated={{@isTruncated}} />
+  <Hds::Copy::Snippet @color="secondary" @textToCopy={{this.text}} @isTruncated={{this.isTruncated}} />
 </li>
--- a/website/docs/components/dropdown/partials/code/how-to-use.md
+++ b/website/docs/components/dropdown/partials/code/how-to-use.md
@@ -226,9 +226,7 @@ Pass the argument `@isLoading={{true}}` to the item. This will show a “loading
 
 ### ListItem::CopyItem
 
-To enable users to copy a snippet of code (eg. URLs, secrets, code blocks, etc.).
-
-Using the `@isTruncated` argument it is possible to constrain the text to one-line and truncate it if it does not fit the available space. Care should be taken in choosing to use this feature as there are [accessibility concerns](/components/copy/snippet?tab=accessibility).
+To enable users to copy a snippet of code (eg. URLs, secrets, code blocks, etc.) use `ListItem::CopyItem`.
 
 ```handlebars
 <Hds::Dropdown as |dd|>
@@ -236,9 +234,13 @@ Using the `@isTruncated` argument it is possible to constrain the text to one-li
   <dd.Title @text="Integrate with Terraform Cloud" />
   <dd.Description @text="Create a new run task in Terraform using the URL and key below." />
   <dd.CopyItem @text="https://api.cloud.hashicorp.com" @copyItemTitle="Endpoint URL" />
-  <dd.CopyItem @isTruncated={{true}} @text="91ee1e8ef65b337f0e70d793f456c71d91ee1e8ef65b337f0e70d793f456c71d" @copyItemTitle="HMAC Key" />
+  <dd.CopyItem @text="91ee1e8ef65b337f0e70d793f456c71d91ee1e8ef65b337f0e70d793f456c71d" @copyItemTitle="HMAC Key" />
+  <dd.CopyItem @isTruncated={{false}} @text="91ee1e8ef65b337f0e70d793f456c71d91ee1e8ef65b337f0e70d793f456c71d" @copyItemTitle="HMAC Key (without truncation)" />
 </Hds::Dropdown>
 ```
+
+Using the `@isTruncated` argument is possible to disable the truncation constrain applied to the text (to be rendered on a single line, and truncate it if it does not fit the available space). Care should be taken in choosing how to use this feature as there are [accessibility concerns](/components/copy/snippet?tab=accessibility).
+
 ### ListItem::Checkmark
 
 For switching context (e.g., organization switchers, project switchers, etc.) use `ListItem::Checkmark`.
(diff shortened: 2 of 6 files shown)


packages/components/addon/components/hds/dropdown/list-item/copy-item.hbs
```diff
@@ -8,5 +8,5 @@
       class="hds-dropdown-list-item__copy-item-title hds-typography-body-100 hds-font-weight-semibold"
     >{{@copyItemTitle}}</div>
   {{/if}}
-  <Hds::Copy::Snippet @textToCopy={{this.text}} @isTruncated={{@isTruncated}} />
+  <Hds::Copy::Snippet @color="secondary" @textToCopy={{this.text}} @isTruncated={{this.isTruncated}} />
```
MelSumner, line 11: Just a note: one of the originally confusing things here was that design is using color in context, so our typical primary/secondary colors are not what they wanted to be used for this component. I'm pretty tired right now so it's entirely possible you have this correct, but since I found it confusing when originally implementing this component, I wanted to make sure to mention it.

MelSumner (changes requested):
Since truncation makes the content unavailable to keyboard only users, you'll need to also update the accessibility documentation for this component to indicate that it is no longer accessible by default. 

That probably raises the larger question of "are we sure about this change though" since we did agree (in our original RFC on the topic) to ship the accessible (conformant) component by default and then provide other conditional conformant variants if needed.

website/docs/components/dropdown/partials/code/how-to-use.md
```diff
@@ -226,19 +226,21 @@ Pass the argument `@isLoading={{true}}` to the item. This will show a “loading
...
   <dd.Title @text="Integrate with Terraform Cloud" />
   <dd.Description @text="Create a new run task in Terraform using the URL and key below." />
   <dd.CopyItem @text="https://api.cloud.hashicorp.com" @copyItemTitle="Endpoint URL" />
-  <dd.CopyItem @isTruncated={{true}} @text="91ee1e8ef65b337f0e70d793f456c71d91ee1e8ef65b337f0e70d793f456c71d" @copyItemTitle="HMAC Key" />
+  <dd.CopyItem @text="91ee1e8ef65b337f0e70d793f456c71d91ee1e8ef65b337f0e70d793f456c71d" @copyItemTitle="HMAC Key" />
+  <dd.CopyItem @isTruncated={{false}} @text="91ee1e8ef65b337f0e70d793f456c71d91ee1e8ef65b337f0e70d793f456c71d" @copyItemTitle="HMAC Key (without truncation)" />
 </Hds::Dropdown>
 ```
+
+Using the `@isTruncated` argument is possible to disable the truncation constrain applied to the text (to be rendered on a single line, and truncate it if it does not fit the available space). Care should be taken in choosing how to use this feature as there are [accessibility concerns](/components/copy/snippet?tab=accessibility).
```
KristinLBradley, line 242:
```suggestion
Using the `@isTruncated` argument, it is possible to disable the truncation constrain applied to the text (to render it on a single line and truncate it if it does not fit the available space). Care should be taken in choosing how to use this feature as there are [accessibility concerns](/components/copy/snippet?tab=accessibility).
```

heatherlarsen:
> Since truncation makes the content unavailable to keyboard only users, you'll need to also update the accessibility documentation for this component to indicate that it is no longer accessible by default.

@MelSumner When working on the previous version, **which was truncated by default**, I vaguely remember that we discussed truncation being okay because the content would be pasted in full, and thus, the user would have complete access to the content. Additionally, common content for this component often includes displaying a token value or other content that's not necessarily meant to be interpreted by the user but is meant to be copied in a quick interaction from a HashiCorp UI and pasted somewhere else, such as the user's terminal or AWS. Given the intended use case, I'm not sure the full content needs to always be displayed. In fact, for much longer snippets, displaying the full snippet could result in other important actions being pushed out of view if the dropdown is not properly configured to handle such content. 

I'm also curious why we didn't originally list this as "non-conformant" since this is being updated to more closely reflect the previous version of the CopyItem?

packages/components/addon/components/hds/dropdown/list-item/copy-item.hbs
heatherlarsen: @MelSumner I'm not following, can you please clarify what you mean? Is there a question or concern that you have about how this has been updated?

heatherlarsen (approved):
This looks good to me. Thanks for getting this resolved :shipit:

didoo:
After discussion with @alex-ju in order to unblock the merging in `release/3-0-0` we have decided to merge this for now, and leave the resolution of @MelSumner blocker to a follow-up discussion this afternoon, and in case open a new PR (pointing to `release/3-0-0`) if we decide to change back this behaviour.

didoo:
After discussion with @alex-ju in order to unblock the merging in `release/3-0-0` we have decided to merge this for now, and leave the resolution of @MelSumner blocker to a follow-up discussion this afternoon, and in case open a new PR (pointing to `release/3-0-0`) if we decide to change back this behaviour.

website/docs/components/dropdown/partials/code/how-to-use.md
KristinLBradley: We should make some changes to the above text even if not exactly as I suggested because there are grammar issues.

didoo: @KristinLBradley my bad, I missed to merge your suggestion before merging! 🤦‍♂️
I'll make the changes in the cleanup PR. Thanks for reminding me.

didoo:
Update: decision was taken to leave the implementation as is, but to document the implications at a11y level for the consumers. This will be done in #1648.

PR #1633, opened 2023-09-14
`Copy::Snippet` - Update `width` behaviour implementation

didoo:
### :pushpin: Summary

In a meeting dedicated to discuss the width/full-width behaviour of the `Code::Snippet` ([notes here](https://docs.google.com/document/d/1x6PXkwy4xtBvScZqeo8taj-n5FaAn5pVjvLfRk2Ck_M/edit))  we have decided:

- Width set to hug the contents (automatically max at 100%, since it’a a block element)
- `@isFullWidth` (default = `false`) - update to be a full-width (100%) variant 
    - (content should **not** be centered, but align to the left)
- `@isTruncated` (default = `false`) - truncates text (if true) otherwise the text wraps (if not set, or set to false)

In this PR I have:
- fixed the way in which “width/full-width” is applied to the `Code::Snippet`
- changed the way in which “truncation” is applied to its content
    - renamed (and moved) class name from `hds-copy-snippet__text--truncated`  to `hds-copy-snippet--is-truncated`
- updated integration tests for `Copy::Snippet`
- updated showcase for `Code::Snippet` to include “Containers” section
     - added option to the `Shw::Grid::Item` to force a `min-width` value (instead of `auto`)
     - updated “Containers” demo in showcase for `Button` elements to follow the `Copy::Snippet` one

👉 👉 👉 **Preview**: https://hds-showcase-git-update-copy-snippet-width-behaviour-hashicorp.vercel.app/components/copy/snippet

### :link: External links

Jira ticket: [TODO](https://hashicorp.atlassian.net/browse/HDS-2366)

***

### 👀 Component checklist

- [ ] A11y tests have been run locally

### DIFF
--- a/.changeset/strange-birds-juggle.md
+++ b/.changeset/strange-birds-juggle.md
@@ -1,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Copy::Snippet` - Fixed the way in which “width/full-width” is applied to the component + Internal update to the “truncation” implementation.
+- the component is not full-width by default (the width now fits the content); use `@isFullWidth={{true}}` to have a full-width layout
+- the internal class name `hds-copy-snippet__text--truncated` has been changed to `hds-copy-snippet--is-truncated` (and moved)
--- a/packages/components/addon/components/hds/copy/snippet/index.hbs
+++ b/packages/components/addon/components/hds/copy/snippet/index.hbs
@@ -8,11 +8,7 @@
   {{hds-clipboard text=@textToCopy onSuccess=this.onSuccess onError=this.onError}}
   ...attributes
 >
-  <Hds::Text::Code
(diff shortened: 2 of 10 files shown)


MelSumner (approved):
[issue] Since we're changing the component to be non-conformant by default (because of truncation), which is not aligned with our RFC'd approach in general, the documentation should also be updated to reflect this. We should also indicate that this is a very rare exception, a current known technical limitation, let the consumer know that this should be used sparingly, and encourage consumers to not truncate when possible because the keyboard-only user has no way of accessing truncated content. 

The rest of this seems good. 👍

heatherlarsen (approved):
Looks great! Thank you 👍🏻

didoo:
After discussion with @alex-ju in order to unblock the merging in `release/3-0-0` we have decided to merge this for now, and leave the resolution of @MelSumner blocker to a follow-up discussion this afternoon, and in case open a new PR (pointing to `release/3-0-0`) if we decide to change back this behaviour.

PR #1637, opened 2023-09-14
[WIP] Limit CI runs to where they are needed...

Dhaulagiri:
### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist

- [ ] A11y tests have been run locally


alex-ju (commented):
I think using the `path` flag makes sense and we can indeed test with the Flight icons first. As a downside, the author/reviewer would have to verify that the appropriate checks have passed, on the bright side tests would be much faster, especially for changes in the `website`.

Dhaulagiri:
> As a downside, the author/reviewer would have to verify that the appropriate checks have passed

That's a good point. Even a small amount of cognitive overhead is not really ideal here.

For now I've updated this PR to include only the change for ember-flight-icons and we can pretty easily experiment with that and see if we like this or not.

alex-ju (approved):

Dhaulagiri:
> If, however, a job within a workflow is skipped due to a conditional, it will report its status as "Success".

This sounds more appealing.

PR #1648, opened 2023-09-19
Update documentation for `3.0.0` release

didoo:
### :pushpin: Summary

This PR contains a few refinements to the documentation, to be merged in the `3.0.0` release and unblock it.

In this PR we have:
- fixed a [copy suggestion](https://github.com/hashicorp/design-system/pull/1631#discussion_r1327478705) that was missed in a previous PR

***

### 👀 Component checklist

### DIFF
--- a/website/docs/components/dropdown/partials/code/how-to-use.md
+++ b/website/docs/components/dropdown/partials/code/how-to-use.md
@@ -245,7 +245,7 @@ To enable users to copy a snippet of code (eg. URLs, secrets, code blocks, etc.)
 </Hds::Dropdown>
 ```
 
-Using the `@isTruncated` argument, it is possible to disable the truncation constrain applied to the text (to render it on a single line, and truncate it if it does not fit the available space). Care should be taken in choosing how to use this feature as there are [accessibility concerns](/components/copy/snippet?tab=accessibility).
+Using the `@isTruncated` argument, it is possible to disable the truncation applied to the text. Care should be taken in choosing how to use this feature as there are [accessibility concerns](/components/copy/snippet?tab=accessibility).
 
 ### ListItem::Checkmark
 
--- a/website/docs/components/dropdown/partials/code/how-to-use.md
+++ b/website/docs/components/dropdown/partials/code/how-to-use.md
@@ -245,7 +245,7 @@ To enable users to copy a snippet of code (eg. URLs, secrets, code blocks, etc.)
 </Hds::Dropdown>
 ```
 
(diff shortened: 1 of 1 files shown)


website/docs/components/dropdown/partials/code/how-to-use.md
```diff
@@ -245,7 +245,7 @@ To enable users to copy a snippet of code (eg. URLs, secrets, code blocks, etc.)
 </Hds::Dropdown>
 ```
 
-Using the `@isTruncated` argument is possible to disable the truncation constrain applied to the text (to be rendered on a single line, and truncate it if it does not fit the available space). Care should be taken in choosing how to use this feature as there are [accessibility concerns](/components/copy/snippet?tab=accessibility).
+Using the `@isTruncated` argument, it is possible to disable the truncation constrain applied to the text (to render it on a single line, and truncate it if it does not fit the available space). Care should be taken in choosing how to use this feature as there are [accessibility concerns](/components/copy/snippet?tab=accessibility).
```
KristinLBradley, line 248:
```suggestion
Using the `@isTruncated` argument, it is possible to disable the truncation constraint applied to the text (to render it on a single line, and truncate it if it does not fit the available space). Care should be taken in choosing how to use this feature as there are [accessibility concerns](/components/copy/snippet?tab=accessibility).
```

KristinLBradley: Thanks, I missed a word previously.

KristinLBradley: Another way to fix it would be to just delete the work "constrain(t)"

KristinLBradley: Other feedback is that it seems awkward to try to define the meaning of "truncation" in the middle of this text: "(to render it on a single line and truncate it if it does not fit the available space)" and the line related to accessibility is also confusing since we are enabling something by default which we imply is problematic. (I don't necessarily agree that the truncation is an accessibility problem here myself but the language around it is confusing.)

MelSumner: I agree with removing the definition in the middle of the sentence. Folks reading these docs know what truncation is. 

(also @KristinLBradley FWIW, the problematic part is that in DS-035 we indicate that we will ship WCAG conformant components as the default; however by turning truncation on by default, we are deviating from that. There is no way to make this an accessible experience for keyboard-only users, at least not in a way that is currently conformant with spec. This is definitely an exception for us, rather than the rule. I hope that helps clarify!)

didoo: @MelSumner @KristinLBradley I have updated the verbiage per suggestions above. Check if it works now.

MelSumner (approved):
:shipit:

MelSumner (approved):

KristinLBradley (approved):

PR #1649, opened 2023-09-19

hashibot-hds:
This PR adds several new icons:

- Google Drive
- Google Docs
- Google Sheets
- Google Slides
- Google Forms

As well as static versions of the animated icons:
- Loading static
- Running static

Questions:
- I know a bit of this is normal, but many icons are showing up as changed that weren't touched in Figma, is this cause for issue/alarm?
- I updated the test to account for two icons being returned for loading, just confirming this is the correct method?


packages/flight-icons/svg-original/running-static-24.svg
jorytindall, line 1: Hmmmm, there's a layout issue that I'll need to fix

didoo:
> I know a bit of this is normal, but many icons are showing up as changed that weren't touched in Figma, is this cause for issue/alarm?

@jorytindall it happens, at times. There is not much we can do, it can be Figma re-updating some points by a small tiny value for some reasons. What I suggest to do (which is the only thing you can practically do, tbh) is to compare the icon before/after in the GitHub "files" page and see if there are visible changes. Apart from that, we have to live with it.

> I updated the test to account for two icons being returned for loading, just confirming this is the correct method?
What I would do, if I were you, is to update one of the two tests to use an icon that has a unique result in response (eg. `cpu`) so the response is `1` item, and leave the other with `loading` so there are 2 items found. In this way we test both use cases :)

didoo (approved):
👍

heatherlarsen (approved):

Dhaulagiri:
@didoo this needed/needs a changeset for it to be released properly

didoo:
@Dhaulagiri don't know why, but I thought it would be created automatically.

@jorytindall can you create a new PR with the changeset and assign it to me for review? thanks

jorytindall:
@didoo Just to confirm, a changeset is need for the `flight-icons` package, but also for the `ember-flight-icons` which consumes `flight-icons` so it's version needs to be bumped, is that correct?

didoo:
TBH I don't know :)  @Dhaulagiri I don't know/remenber if by bumping only the `flight-icons` package, the `ember-flight-icons` is automatically bumped by changeset, or the developer need to bump both. The [documentation](https://github.com/hashicorp/design-system/blob/main/packages/flight-icons/CONTRIBUTING.md#updating-the-icons-in-flight-icons) is not clear on this.

Dhaulagiri:
We can update the documentation but one can also reference [previous exports](https://github.com/hashicorp/design-system/pull/1638/files) to see that these are a `minor` for `@hashicorp/flight-icons` and then changesets takes care of everything from there for `ember-flight-icons`.

didoo:
@Dhaulagiri https://github.com/hashicorp/design-system/pull/1656

jorytindall:
Added changeset here: #1657

didoo:
@jorytindall when opening a new PR don't forget to add the links to the Jira tickets that are related to the PR, so they're easy to find/reference.

PR #1623, opened 2023-09-07
`AppFooter` component (HDS-2480)

KristinLBradley:
### :pushpin: Summary
If merged, this PR will add the `AppFooter` component to HDS.

- **Showcase:** https://hds-showcase-git-hds-2480-application-footer-c-d73429-hashicorp.vercel.app/components/app-footer

### :link: External links

- Jira ticket: [HDS-2480](https://hashicorp.atlassian.net/browse/HDS-2480)
- 👉 AppFooter within Cloud-ui: https://hashicorp.atlassian.net/browse/HDS-2482
- Figma file: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/czK6k8hCT3vmJzG0eRfnaw/HDS-Product---Components?node-id=43239%3A378&mode=dev

***

### 👀 Reviewer's checklist:

- [x] Percy was checked for any visual regression
- [x] A11y tests have been run locally (`yarn test:a11y --filter="COMPONENT-NAME"`)
- [x] If documenting a new component, an acceptance test that includes the `a11yAudit` has been added
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-2480]: https://hashicorp.atlassian.net/browse/HDS-2480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1613, opened 2023-09-01
`Text` component replacements

didoo:
### :pushpin: Summary

This PR follows the introduction of the `Text` component in #1490, and intends to replace the usage of `.hds-typography` classes or `token-font` tokens, where a `Text` component would work.

As agreed during the last HDS Engineering Sync meeting, I've replaced only purely textual elements, leaving out any semantic element (eg. `button`, `label`, etc).

### :hammer_and_wrench: Detailed description

In this PR I have replaced the typographic styling in the following components, to use the `Text` component where possible:

- `Accordion`
- `ApplicationState`
- `Copy::Snippet`
- `Dropdown`
- `Form`
- `Flyout`
- `Modal`
- `PageHeader`
- `Pagination`
- `Stepper`
- `Tag`

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2354

***

### 👀 Reviewer's checklist:

- [ ] +1 Percy if applicable
- [ ] Confirm that PR has a changelog update via [Changesets](https://github.com/changesets/changesets) if needed
- [ ] Confirm that A11y tests have been run locally for this component

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #1652, opened 2023-09-21
`Text` component - Fix issue with whitespace

didoo:
### :pushpin: Summary

While reviewing #1623 and doing some tests with the `Hds::Link::Inline` in combination with the `Hds::Text` component, I noticed a classic effect of whitespace in Ember components: an extra underline in the links.

This PR intends to address this issue.

In this PR I have:
- updated the showcase for the `Link::Inline` to add use some cases where the text is rendered using the `<Hds::Text>` component (for all the styles: `Body`, `Display`, `Code`)
- then fixed the issue of whitespace in `Text` component by removing extra newlines in a few files

👉 👉 👉 **Preview**: https://hds-showcase-git-fix-whitespace-in-text-component-hashicorp.vercel.app/components/link/inline

Notice: I am considering this a patch, but if you have different opinions happy to increase the semver

**Before:**
modified

**After:**
screenshot_3099

***

### 👀 Component checklist

### DIFF
--- a/packages/components/addon/components/hds/text/index.hbs
+++ b/packages/components/addon/components/hds/text/index.hbs
@@ -2,7 +2,9 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-
-{{#let (element this.componentTag) as |Tag|}}
-  <Tag class={{this.classNames}} {{style color=this.customColor}} ...attributes>{{yield}}</Tag>
-{{/let}}
+{{! IMPORTANT: we removed any extra newlines before/after the `let` to reduce the issues with unexpected whitespaces (see https://github.com/hashicorp/design-system/pull/1652) }}
+{{#let (element this.componentTag) as |Tag|}}<Tag
+    class={{this.classNames}}
+    {{style color=this.customColor}}
+    ...attributes
+  >{{yield}}</Tag>{{/let}}
--- a/.changeset/wise-peas-deliver.md
+++ b/.changeset/wise-peas-deliver.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Text` - Fixed issue with whitespace adding extra underline when used in links (eg. inside a `Link::Inline`)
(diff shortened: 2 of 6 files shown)


alex-ju (approved):

Dhaulagiri:
Did this fail the visual diffs in Percy when we retrofitted the existing code to use the new text component?

alex-ju:
> Did this fail the visual diffs in Percy when we retrofitted the existing code to use the new text component?

Good question – I thought about this too. I've not seen [the VRT](https://percy.io/HashiCorp/hds-components/builds/30139864/changed/1668100686?browser=chrome&browser_ids=48%2C49&subcategories=approved&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280) failing in #1613

didoo:
> Did this fail the visual diffs in Percy when we retrofitted the existing code to use the new text component?

@alex-ju @Dhaulagiri no, because we didn't have any use case of the `Hds::Text` used inside an inline link (hence me adding new use cases to the showcase for the `Link::Inline` component)

packages/components/addon/components/hds/text/index.hbs
```diff
@@ -2,7 +2,9 @@
...
-{{#let (element this.componentTag) as |Tag|}}
-  <Tag class={{this.classNames}} {{style color=this.customColor}} ...attributes>{{yield}}</Tag>
-{{/let}}
\ No newline at end of file
+{{! IMPORTANT: we removed any extra newlines before/after the `let` to reduce the issues with unexpected whitespaces (see https://github.com/hashicorp/design-system/pull/1652) }}
+{{#let (element this.componentTag) as |Tag|}}<Tag
+    class={{this.classNames}}
+    {{style color=this.customColor}}
+    ...attributes
+  >{{yield}}</Tag>{{/let}}
```
KristinLBradley, line 10: Would it work to use the "~" character instead?

Example: https://github.com/hashicorp/design-system/blob/main/packages/components/addon/components/hds/interactive/index.hbs

didoo: No, it's a different thing (I tried myself what you suggested, but didn't work; unfortunately it's an exercise of trial and error for me, because I don't understand the logic of this; I suspect it's a matter of newlines, more than whitespaces)

KristinLBradley: Ok thanks for explaining

PR #1634, opened 2023-09-14
Version 3.0

alex-ju:
### :pushpin: Summary

This is a PR to group changes for the next major release, [version 3.0.0](https://github.com/hashicorp/design-system/milestone/1?closed=1).

PR #1653, opened 2023-09-21
Add codemods for Version 3.0

alex-ju:
### :pushpin: Summary

[As agreed during the engineering sync](https://docs.google.com/document/d/1twnk39hZ7p8oZYrld_FYsdiuUb6yn_weA46KnhF4nu4/edit#bookmark=id.1wyi4r3swd72), I've started to prepare the codemods for the breaking changes we're planning to release in [version 3.0](https://github.com/hashicorp/design-system/pull/1634).

- I added one codemod for the `Dropdown` changes. If we agree this is the right format, I will add the remaining two (for the `MaskedInput` and the `Form::RadioCard`).
- It seems to be the norm to publish these codemods as a separate package, so I placed them accordingly.

As always open for feedback on both the format and the location.

***

### 👀 Component checklist

- [x] A11y tests have been run locally


packages/codemods/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@hashicorp/design-system-codemods",
+  "description": "Helios Design System codemods",
+  "version": "0.1.0",
+  "scripts": {
+    "lint": "eslint --cache .",
+    "test": "codemod-cli test",
+    "test:coverage": "codemod-cli test --coverage",
+    "update-docs": "codemod-cli update-docs",
```
didoo, line 9: 🤔

packages/codemods/transforms/v3/dropdown/README.md
didoo, line 1: we should probably be more verbose and call the folder with the full semver value `3.0.0` (in case we have a 3.1.0 codemod, we never know)

packages/codemods/transforms/v3/dropdown/README.md
```diff
@@ -0,0 +1,26 @@
+# v3/dropdown
```
didoo, line 1:
```suggestion
# v3.0.0/dropdown
```

packages/codemods/transforms/v3/dropdown/README.md
```diff
@@ -0,0 +1,26 @@
...
+# or
+
+yarn global add @hashicorp/design-system-codemods
+@hashicorp/design-system-codemods v3/dropdown path/of/files/ or/some**/*glob.hbs
+```
+
+## Local Usage
+```
+node ./bin/cli.js v3/dropdown path/of/files/ or/some**/*glob.hbs
+```
```
didoo, line 18: I would do something like this to make it clearer what options they have (I can't write the exact markdown, because of the multiple backticks, but I hope it gives you the gist of it):

## Usage

### Using `npx`
```
npx @hashicorp/design-system-codemods v3.0.0/dropdown path/to/single/file.hbs
```
or
```
npx @hashicorp/design-system-codemods v3.0.0/dropdown path/to/some/glob/**/*.hbs
```

### Global installation
```
yarn global add @hashicorp/design-system-codemods
```
and then
```
@hashicorp/design-system-codemods v3.0.0/dropdown path/to/single/file.hbs
@hashicorp/design-system-codemods v3.0.0/dropdown path/to/some/glob/**/*.hbs
```

### Local installation
```
node ./bin/cli.js v3.0.0/dropdown path/to/single/file.hbs
node ./bin/cli.js v3.0.0/dropdown path/to/some/glob/**/*.hbs
```

didoo: (we may also want to limit the proposed approaches to only one or two)

packages/codemods/transforms/v3/dropdown/__testfixtures__/basic.input.hbs
didoo, line 1: I don't know how these test work, but should we also have a version where the dropdown is already using the `@listPosition="bottom-right"` (for example) and the test makes sure we don't touch it, the output remains the same?

packages/codemods/transforms/v3/dropdown/__testfixtures__/basic.input.hbs
```diff
@@ -0,0 +1,10 @@
+<Hds::Dropdown @listPosition="left" as |dd|>
```
didoo, line 1: I would probably use a much simpler code snippet, with a single child (unless we're worried it may modify the children too, in which case I would add one per kind of item)

packages/codemods/transforms/v3/dropdown/README.md
didoo: Or maybe here we mention only one of the methods (with the two file paths) and we link to the `packages/codemods/README.md` for details about other possible methods.

packages/codemods/README.md
```diff
@@ -0,0 +1,43 @@
+# codemods
+
+A collection of codemods for Helios components.
+
+## Usage
```
didoo, line 5: See comment below (sorry, I saw this after the comment; I don't approach the code review in linear way, at times)

packages/codemods/README.md
```diff
@@ -0,0 +1,43 @@
...
+## Local Usage
+```
+node ./bin/cli.js <TRANSFORM NAME> path/of/files/ or/some**/*glob.js
+```
+
+## Transforms
+
+<!--TRANSFORMS_START-->
+* [v3/dropdown](transforms/v3/dropdown/README.md)
+<!--TRANSFORMS_END-->
```
didoo, line 27: We should probably group them together (unless this is automated, which I suspect)
```suggestion
<!--TRANSFORMS_START-->
### v3.0.0
* [v3.0.0/dropdown](transforms/v3.0.0/dropdown/README.md)
<!--TRANSFORMS_END-->
```

packages/codemods/.github/workflows/ci.yml
```diff
@@ -0,0 +1,60 @@
...
+    - name: linting
+      run: yarn lint
+
+  test:
+    name: Tests
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node: ['10', '12', '14']
```
didoo, line 36: What are we testing here? 🤔

packages/codemods/transforms/.gitkeep
didoo, line 1: maybe this can be deleted?

didoo (commented):
This approach works for me (I have left only minor comments) so feel free to continue along this line. 

One general consideration: should we move some of the codemods from Cloud UI to here, so they can be used as reference/example? (only the ones that are generic, not specific to Cloud UI).

packages/codemods/transforms/v3/dropdown/index.js
```diff
@@ -0,0 +1,37 @@
...
+        if (node.tag === 'Hds::Dropdown') {
+          // filter out the `@listPosition` attribute
+          const outputAttrs = node.attributes.filter(a => a.name !== '@listPosition');
+
+          // look up for `@listPosition`
+          const attr = node.attributes.find(a => a.name === '@listPosition')
+
+          // if `@listPosition` is `left` or `right` update the argument value
+          if (attr.value.chars === "left" || attr.value.chars === "right") {            
+            outputAttrs.push(b.attr('@listPosition', b.text(`bottom-${attr.value.chars}`)));
```
didoo, line 18: I find this pattern ([used in Cloud UI](https://github.com/hashicorp/cloud-ui/blob/main/codemods/transforms/hds-migration/core/external-link/index.js#L84-L85)) a bit clearer/explicit:
```suggestion
            const updatedListPositionAttr = b.attr('@listPosition', b.text(`bottom-${attr.value.chars}`));
            outputAttrs.push(updatedListPositionAttr);
```

packages/codemods/transforms/v3/dropdown/__testfixtures__/basic.input.hbs
alex-ju: Good call, yes, I can add a test to make sure we don't accidentally touch code we don't want to touch.

packages/codemods/transforms/v3/dropdown/__testfixtures__/basic.input.hbs
alex-ju: yes, that works for me, just wanted to make sure we have node elements in the block content

packages/codemods/transforms/v3/dropdown/README.md
alex-ju: I worked under the assumption that a code mode is required only for API changes. If we made any change that would require a change in our consumer's codebase it means it's a breaking change. This is why I only see us adding code mods for major releases.

The reason I named it generically v3 (or it can be 3.x.x) is that these code mods can be used to move from 2.7 to 3.2 or even 2.7 to 4.2 (by consequently running the v3 mods and then the v4 mods). Happy to open this to a wider discussion.

packages/codemods/README.md
alex-ju: yes, it's unfortunately automated by running the `update-docs` command you touched on above.

packages/codemods/.github/workflows/ci.yml
alex-ju: This was the more or less auto-generated CI config. I would lean towards disabling these jobs and only running the tests locally when writing the code mods. I think the overhead is quite big and don't think it's worth running them in CI.

packages/codemods/transforms/.gitkeep
alex-ju: yes, indeed

alex-ju:
> One general consideration: should we move some of the codemods from Cloud UI to here, so they can be used as reference/example? (only the ones that are generic, not specific to Cloud UI).

it's probably worth porting those here when we find the time considering we have 2 products on v1.x; there are a few others (I'm aware of a few written by me for v2 where I was asked not to merge them in `cloud-ui` so I dropped the commits before bumping HDS to v2) that may be worth reviving as well.

packages/codemods/transforms/v3/dropdown/README.md
alex-ju: Updated with a simplified version, suggesting npx for consumers and local binary for maintainers. Agree there were to many options

packages/codemods/transforms/v3/radio-card/index.js
```diff
@@ -0,0 +1,63 @@
...
+                    attrs: outputAttrs,
+                    children: node.children,
+                    modifiers: node.modifiers,
+                    blockParams: node.blockParams,
+                  }
+                )
+              ];
+            }
+          }
+        } else {
```
didoo, line 34: does this apply to any other kind of node?

packages/codemods/transforms/v3/radio-card/index.js
```diff
@@ -0,0 +1,63 @@
+const CODEMOD_ANALYSIS = process.env.CODEMOD_ANALYSIS;
+
+module.exports = function ({ source /*, path*/ }, { parse, visit }) {
```
didoo, line 3: according to the changelog https://github.com/hashicorp/design-system/pull/1634/files#diff-fa1aa341f6153e2733520f2d8227380e02de23c96082cac195a30f4ddde9fbe7 also a `@maxWidth` should be applied. If this is the case, how is the codemod going to take care of it?

didoo (commented):
Added a few comments (apart from that, probably ready to be approved unless you have to add more codemods; on my side, I am going to push a codemod for the `SideNav` right now on this branch)

packages/codemods/transforms/v3/radio-card/index.js
alex-ju: 🤦 I forgot the node name – amended now

packages/codemods/transforms/v3/radio-card/index.js
didoo: Maybe something to add to the README for this codemod?

alex-ju: That's a good shout – added an extra note to the `v3/radio-card` instructions

didoo (approved):
👍

PR #1660, opened 2023-09-21
fix yml for ember-flight-icons ci

Dhaulagiri:
Bad indenting and a stray `||` i didn't mean to check in


Dhaulagiri:
it works!
Screenshot 2023-09-21 at 15 27 21

KristinLBradley (approved):

PR #1644, opened 2023-09-15
skip ember-flight-icons tests if no files changed

Dhaulagiri:
### :pushpin: Summary

Alternative to https://github.com/hashicorp/design-system/pull/1637 which is more complex for us but has the benefit of allowing the jobs to be skipped and show as "Success" (similar to if it had run and passed) vs "pending". Heavily inspired by how Consul set theirs up.

Starting small with ember-flight-icons as an experiment.

### :link: External links

Jira ticket: [HDS-2157](https://hashicorp.atlassian.net/browse/HDS-2157)

***

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-2157]: https://hashicorp.atlassian.net/browse/HDS-2157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1663, opened 2023-09-22
Conditional CI for components & website

Dhaulagiri:
### :pushpin: Summary

Follow up from https://github.com/hashicorp/design-system/pull/1644 to bring the same logic to other areas of our test suite

### :link: External links

### DIFF
--- a/.github/scripts/filter_changed_files_components.sh
+++ b/.github/scripts/filter_changed_files_components.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+trigger_ci=false
+
+# If we are on main we want to always run the tests
+if [[ $GITHUB_REF_NAME == "main" ]]; then
+	trigger_ci=true
+	echo "On main, running CI: $trigger_ci"
+	echo "trigger-ci=$trigger_ci" >>"$GITHUB_OUTPUT"
+	exit 0
+fi
+
+# Get the list of changed files
+files_to_check=$(git diff --name-only origin/main)
+
+# Define the directories to check
+included_directories=("packages/components" "packages/ember-flight-icons" "packages/flight-icons/catalog.json")
+
+# Loop through the changed files and find directories/files that trigger need for tests
+for file_to_check in $files_to_check; do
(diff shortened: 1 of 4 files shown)


.github/scripts/filter_changed_files_website.sh
```diff
@@ -0,0 +1,37 @@
...
+	echo "On main, running CI: $trigger_ci"
+	echo "trigger-ci=$trigger_ci" >>"$GITHUB_OUTPUT"
+	exit 0
+fi
+
+# Get the list of changed files
+files_to_check=$(git diff --name-only origin/main)
+
+# Define the directories to check
+included_directories=("website" "packages/flight-icons/catalog.json")
```
Dhaulagiri, line 17: This would run website CI if any website file changed or if the underlying icon sprite changes (tested on the icon docs page). Do we need to also run the tests if components themselves change?

.github/scripts/filter_changed_files_components.sh
```diff
@@ -0,0 +1,37 @@
...
+	echo "On main, running CI: $trigger_ci"
+	echo "trigger-ci=$trigger_ci" >>"$GITHUB_OUTPUT"
+	exit 0
+fi
+
+# Get the list of changed files
+files_to_check=$(git diff --name-only origin/main)
+
+# Define the directories to check
+included_directories=("packages/components" "packages/ember-flight-icons" "packages/flight-icons/catalog.json")
```
Dhaulagiri, line 17: Run tests (and by proxy Percy) if components or any icons change.

Dhaulagiri: If 👍  I can add this as a comment directly in the bash script too

Dhaulagiri:
In practice we could likely have a single bash script that is invoked with a list of directories/files to include since otherwise all 3 are exactly the same. Assuming this all works I can make that optimization as a next step.

MelSumner (approved):
+1 to trying this out and seeing how it works for our team.

PR #1667, opened 2023-09-25
clarify direct contributions

Dhaulagiri:
### :pushpin: Summary
This is based on some feedback I've received that this could be made more clear, so I took a pass at doing this by:

* Move it higher in the doc
* Make it more clear why we don't accept large direct contributions


MelSumner (approved):
:shipit:

heatherlarsen (approved):

PR #1670, opened 2023-09-27
`SideNav` component - Fix "extra" containers naming

didoo:
### :pushpin: Summary

While reviewing https://github.com/hashicorp/design-system/pull/1623 I realized that when I added the "extra" containers I have used lowercase names `extraBefore/After` but since all the other yielded elements have uppercase initials, is kinda inconsistent. I am not sure why I did it this way, but I've checked and currently [these containers are not used anywhere](https://github.com/search?q=%22.extraBefore%22+user%3Ahashicorp&type=code&ref=advsearch), apart from the HDS code itself, so it's pretty safe to change the naming to uppercase (it still remains a major breaking change anyway, so I am going to point the PR to the `release/3.0.0` branch).

In this PR I have:
- renamed `extraBefore/After` to `ExtraBefore/After` (uppercase `E`)

***

### 👀 Component checklist


.changeset/shy-bobcats-eat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": major
+---
+
+`SideNav` - renamed `extraBefore/After` generic containers to `ExtraBefore/After` (uppercase `E`)
```
alex-ju, line 5: Non-blocker: we tend to add an extra paragraph on what to do when migrating ('To migrate...') most of them (as it is this one) are pretty straightforward, but may be valuable to make these instructions explicit.

alex-ju (approved):

PR #1671, opened 2023-09-27
consolidate file change bash scripts

Dhaulagiri:
### :pushpin: Summary

This PR consolidates the 3 bash scripts which were effectively all the same into one that takes a list of strings file paths to check for changes.


.github/scripts/filter_changed_files.sh
```diff
@@ -2,24 +2,29 @@
 
 trigger_ci=false
 
-# If we are on main we want to always run the tests
+#If we are on main we want to always run the tests
```
Dhaulagiri, line 5:
```suggestion
# If we are on main we want to always run the tests
```

.github/scripts/filter_changed_files.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# Using `git merge-base` ensures that we're always comparing against the correct branch point. 
+#For example, given the commits:
+#
+# A---B---C---D---W---X---Y---Z # origin/main
+#             \---E---F         # feature/branch
+#
+# ... `git merge-base origin/main` would return commit `D`
+# `...HEAD` specifies from the common ancestor to the latest commit on the current branch (HEAD)..
+files_to_check=$(git diff --name-only "$(git merge-base origin/main HEAD~)"...HEAD)
```
Dhaulagiri, line 30: shamelessly lifted this from [consul's implementation](https://github.com/hashicorp/consul/blob/main/.github/scripts/filter_changed_files_go_test.sh#L16) as I'm 100% sure they know what they're doing compared to me here

alex-ju (approved):

PR #1672, opened 2023-09-27
Remove ci-components-nightly

alex-ju:
### :pushpin: Summary

Remove `ci-components-nightly.yml`

This was introduced in https://github.com/hashicorp/design-system/pull/1403 as an experiment to run `ember-a11y-testing` on `packages/components` as a nightly cron job. Since then we found alternative approaches to run automated a11y testing on our components, so there's no need to keep this GitHub Workflows job around.

Note 1: the previously stored artifacts expired and were automatically removed.
Note 2: look at the speed of those checks! 🏎️


MelSumner (changes requested):
@alex-ju I left this in there but with the automated part turned off, it's just run with a manual trigger now. 

Is it causing problems to leave it in? It's convenient to me to leave it be, but if it's causing issues for us that's another thing.

alex-ju:
> @alex-ju I left this in there but with the automated part turned off, it's just run with a manual trigger now.
> 
> Is it causing problems to leave it in? It's convenient to me to leave it be, but if it's causing issues for us that's another thing.

Not causing any issues – I thought of it as a leftover. Happy to keep it around if it may bring value in the future.

PR #1654, opened 2023-09-21
Control the selectedTabIndex from the consuming application

MiniHeyd:
### :pushpin: Summary

This PR will allow for the consuming application to manage the selected tab index state, in turn allowing for selected tab to be set after render by external controls.

### :hammer_and_wrench: Detailed description

Currently the tab index is controlled by the `didInsert` method and can be initially set to any of the tabs through use of the `@isSelected` tab arg. Subsequently the tab can only be changed by controls internal to the component, limiting more dynamic use cases like tracking selected tab as a query parameter and navigating with the browser back and forward buttons. This PR introduces the ability to pass in the selectedTabIndex as an arg on the parent component which will override the internal tracking of the selected tab index. This also moves the tab indicator logic to the `did-update` modifier, which is set up to fire when the selected index changes.

Inspiration was taken from https://github.com/concordnow/ember-aria-tabs

### :camera_flash: Screenshots

https://github.com/hashicorp/design-system/assets/2334991/30e31d4c-3d4e-411a-9b10-c7706d6de5d7

### :link: External links

Jira ticket: [HDS-XXX](https://hashicorp.atlassian.net/browse/HDS-XXX)
Figma file: [if it applies]

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [ ] A11y tests have been run locally
- [ ] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #1688, opened 2023-10-03
[WIP] `Tabs` - Fix issues for query-params-based status, nested tabs, variable tabs widths

didoo:
### :pushpin: Summary

TODO

This is a "reworking" of to understand if there was a more idiomatic (to HDS code) approach to solve the issue of tab selection and persistence using query params, initially proposed by @MiniHeyd in https://github.com/hashicorp/design-system/pull/1654.

In the process, I have also tried to tackle other two known tabs-related issues: the fact that when there are nested tabs, the indicator for the tabs inside hidden panels doesn't get initialized correctly, and the fact that when the content of the tab button (in particular the badge) changes, the indicator doesn't update its width accordingly.

The intent of this PR is to help a conversation with @MiniHeyd and see if we can have a synthesis of the two PRs in a single one.

TODO

### :link: External links

Jira ticket: TODO

***

### 👀 Component checklist


didoo:
@MiniHeyd as discussed today, the PR is ready for an initial review (I still have to update the documentation, but I'll do it in a separate follow-up PR tomorrow)

@hashicorp/hds-engineering it would be great if you could start to look at this PR, is pretty meaty and it would be better to get it merged in `main` before the `3.0` release so we don't block the TFC consumers in their `Tabs` adoption PRs (already opened).

packages/components/addon/components/hds/tabs/index.hbs
```diff
@@ -12,8 +17,8 @@
             "hds/tabs/tab"
             didInsertNode=this.didInsertTab
             willDestroyNode=this.willDestroyTab
+            didUpdateNode=this.didUpdateTab
```
alex-ju, line 20: maybe we can set `didUpdateNode` before the `willDestroyNode`

packages/components/addon/components/hds/tabs/index.js
```diff
@@ -7,45 +7,71 @@ import Component from '@glimmer/component';
...
+      return this._selectedTabIndex;
+    }
+  }
+
+  set selectedTabIndex(value) {
+    if (this.args.selectedTabIndex) {
+      // noop
+    } else {
+      this._selectedTabIndex = value;
+    }
```
alex-ju, line 33: do you think it reads easier if we swap this statement?

alex-ju (approved):
Left a few comments, non-blocking. Tested manually and looks/works great. Don't forget the changelog entry

packages/components/tests/integration/components/hds/tabs/index-test.js
```diff
@@ -3,232 +3,555 @@
...
+        <T.Tab data-test="tab-2">Two</T.Tab>
+        <T.Panel data-test="panel-1">Content 1</T.Panel>
+        <T.Panel data-test="panel-2">Content 2</T.Panel>
+      </Hds::Tabs>
+    `);
+    let tablistStyle = find('.hds-tabs__tablist').style;
+    assert.strictEqual(
+      tablistStyle['cssText'],
+      '--indicator-left-pos: 0px; --indicator-width: 81px;'
+    );
```
alex-ju, line 385: I'm thinking something like this could help us test an approximate match – we can abstract it into a small function
```suggestion
    let tablistStyle = find('.hds-tabs__tablist').style;
    let regex = /[a-z-]+ *:([^:;]+)px;/g;
    let indicatorLeftPos = regex.exec(tablistStyle['cssText'])[1];
    let indicatorWidth = regex.exec(tablistStyle['cssText'])[1];
    assert.true(Math.abs(0 - indicatorLeftPos) <= 1);
    assert.true(Math.abs(82 - indicatorWidth) <= 1);
```

MelSumner (approved):
I'm not sure what's causing the test failures, but after that's resolved, I have no objections to this PR after today's group review.

packages/components/addon/components/hds/tabs/index.hbs
didoo: my bad, fixed

packages/components/tests/dummy/app/styles/_typography.scss
```diff
@@ -170,6 +170,8 @@ $show-font-family-mono: ui-monospace, menlo, consolas, monospace;
   }
 }
 
+.shw-text-body code { font-size: 0.85em; } // more visually balanced
```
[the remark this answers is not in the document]
didoo: ah yes! 👍

packages/components/tests/integration/components/hds/tabs/index-test.js
didoo: Yeah I tried your suggestion as first thing, but in some cases the difference is bigger (eg. `109` vs `112`) so I had to go full-in with a custom function that does more complex checks depending on different conditions/use cases (including the case where we're comparing `0` with `0` so we can't use a percentage)

packages/components/addon/components/hds/tabs/index.js
didoo: I have changed the logic to make it more explicit and clear to understand (by introducing a `isControlled` tracked variable) - See commit https://github.com/hashicorp/design-system/pull/1688/commits/8fd10b9ac303b23fc534e3e295919df11c3a6e46

didoo:
@alex-ju this is ready for a second pass.

@MiniHeyd see my commit https://github.com/hashicorp/design-system/pull/1688/commits/8fd10b9ac303b23fc534e3e295919df11c3a6e46 (added after your review)

alex-ju (approved):
Re-reviewed this one – had a more in-depth look at the functional aspects and checked the updated tests – and it's good to be merged. You'll likely want to get the documentation in first (~~which may require a rebase on this judging by [the current diff](https://github.com/hashicorp/design-system/pull/1702/files)~~ no rebase required).

PR #1695, opened 2023-10-05
Acceptance test with a11y audit for side-nav component

MelSumner:
### :pushpin: Summary

If merged, this PR adds an acceptance test for the side-nav with an a11y audit. 

### :hammer_and_wrench: Detailed description

- ~~added `@color` support to the placeholder component to complement the existing `@background` argument. This provides support for developers to implement an accessible color combination for the placeholder.~~
- ~~Updated the sidenav preview page to include a foreground color that meets color-contrast thresholds with the background color.~~
- added a skip for the `listitem` rule to the test itself
- added a skip for color-contrast rule on the placeholder element

### :link: External links

Jira ticket: [HDS-2627](https://hashicorp.atlassian.net/browse/HDS-2627)
Figma file: [if it applies]

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [x] A11y tests have been run locally (`yarn test:a11y --filter="COMPONENT-NAME"`)

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-2627]: https://hashicorp.atlassian.net/browse/HDS-2627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1699, opened 2023-10-06
`Shw::Placeholder` - Fix color contrast

didoo:
### :pushpin: Summary

Context: https://github.com/hashicorp/design-system/pull/1695/files#r1347806114

In this PR I have:
- added default background color for `Shw::Placeholder`
- removed `@background` argument for `Shw::Placeholder` when the color was the same as the default
- replaced background colors in a couple of `Shw::Placeholder` instances

***

### 👀 Component checklist

### DIFF
--- a/packages/components/tests/dummy/app/styles/showcase-components/placeholder.scss
+++ b/packages/components/tests/dummy/app/styles/showcase-components/placeholder.scss
@@ -18,6 +18,7 @@
   line-height: 1.2;
   text-align: center;
   text-shadow: 0 0 5px #fff;
+  background-color: #eee;
 
   a,
   a > & {
--- a/packages/components/tests/dummy/app/templates/components/accordion.hbs
+++ b/packages/components/tests/dummy/app/templates/components/accordion.hbs
@@ -18,7 +18,7 @@
         <A.Item>
           <:toggle>Item one</:toggle>
           <:content>
-            <Shw::Placeholder @text="generic content" @height="40" @background="#eee" />
+            <Shw::Placeholder @text="generic content" @height="40" />
           </:content>
         </A.Item>
       </Hds::Accordion>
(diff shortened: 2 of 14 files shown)


packages/components/tests/dummy/app/styles/showcase-components/placeholder.scss
```diff
@@ -18,6 +18,7 @@
   line-height: 1.2;
   text-align: center;
   text-shadow: 0 0 5px #fff;
+  background-color: #eee;
```
KristinLBradley, line 21: Thanks for adding this. It makes the placeholder easier to use besides just fixing the contrast issue.

KristinLBradley (approved):

packages/components/tests/dummy/app/styles/showcase-components/placeholder.scss
didoo: I am not sure why I didn't think of it before 🤷‍♂️🤦‍♂️

MelSumner:
Do you think we should have a few different conformant placeholder background colors? 

Looking at these changes through the lens of the side-nav page, if the background color is changed then there's not enough of a difference between the top background color and lower background color, see the two areas indicated here: 
CleanShot 2023-10-06 at 12 31 43

didoo:
> Do you think we should have a few different conformant placeholder background colors?
> 
> Looking at these changes through the lens of the side-nav page, if the background color is changed then there's not enough of a difference between the top background color and lower background color, see the two areas indicated here: CleanShot 2023-10-06 at 12 31 43

Can you clarify what you suggest to do? I am not sure I understand

MelSumner (approved):

MelSumner:
@didoo let's merge this and then I'll rebase that other PR and we can figure it out there.

KristinLBradley:
> Do you think we should have a few different conformant placeholder background colors?
> 
> Looking at these changes through the lens of the side-nav page, if the background color is changed then there's not enough of a difference between the top background color and lower background color, see the two areas indicated here: CleanShot 2023-10-06 at 12 31 43

To me, it seems like that would overcomplicate usage of the Placeholder. I think we can rely on the a11y tests to catch other non-accessible color combinations when custom background colors are used.

PR #1700, opened 2023-10-09
`Pagination::Compact` - Expose the `SizeSelector`

didoo:
### :pushpin: Summary

In a first assessment I though the “problem” was the documentation, but after speaking with @andrew-hashicorp I have realized that the “problem” is an assumption that we made long time ago when we designed and built the Pagination component: that the Compact variant (cursor-based pagination) could not have not only the totalItems/currentPage but also the pageSize. In reality even with this type of pagination, it’s possible to define a pageSize (imagine it like the number of items retrieved by the backend per query). Demonstration is the fact that in the Consul UI Kit they [have implemented exactly this pattern](https://github.com/hashicorp/consul-ui-toolkit/blob/main/toolkit/src/components/cut/list/pagination.hbs#L5-L37) (as custom), as well as the fact that this is what is needed by @andrew-hashicorp.

So the goal of this ticket is to expose this sub-UI-component in the Compact variant too.

Notice: no need to do anything from the Figma/design side, because this combination is already achievable with the current Figma component (cc @heatherlarsen)

In this PR I have:
- added `Pagination::sizeSelector` to `Pagination::Compact` (hidden by default)
- updated “Pagination” showcase to include use case where `SizeSelector` is visible
- updated integration tests
- updated "How to use" and "Component API" documentation
👉 👉 👉 **Preview**:
- showcase: https://hds-showcase-git-pagination-compact-add-page-s-013ac9-hashicorp.vercel.app/components/pagination
- documentation: https://hds-website-git-pagination-compact-add-page-si-866250-hashicorp.vercel.app/components/pagination?tab=code

#### ⚠️ To discuss

In the current implementation, I have set the default of `@showSizeSelector` to `false` for the `Compact` variant, while for the `Numbered` variant is `true` by default. The reason for this differentiation is that, looking at the component's usage, seems that [in most of the cases it's used without the "size selector"](https://github.com/search?q=%3CHds%3A%3APagination%3A%3ACompact+user%3Ahashicorp&type=code&ref=advsearch). The alternative would be to set it to `true` for consistency, but in that case the change should be released as breaking (with the `3.0` release, likely) and a codemod should be added to handle the change in the consumers' codebases (adding a `@showSizeSelector={{false}}` in all the instances found).
Personally I see pros/cons in both alternatives, so I'm eager to hear the reviewer's opinion too.

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2638

***

### 👀 Component checklist

### DIFF
--- a/packages/components/addon/components/hds/pagination/compact/index.hbs
+++ b/packages/components/addon/components/hds/pagination/compact/index.hbs
@@ -27,4 +27,13 @@
       @disabled={{@isDisabledNext}}
     />
   </nav>
+
+  {{#if this.showSizeSelector}}
+    <Hds::Pagination::SizeSelector
+      @pageSizes={{this.pageSizes}}
+      @label={{@sizeSelectorLabel}}
+      @selectedSize={{this.currentPageSize}}
+      @onChange={{this.onPageSizeChange}}
+    />
+  {{/if}}
 </div>
--- a/website/docs/components/pagination/partials/code/component-api.md
+++ b/website/docs/components/pagination/partials/code/component-api.md
@@ -28,23 +28,26 @@ These Pagination sub-elements may be used directly if you need to cover a very s
 <C.Property @name="totalItems" @required="true" @type="number">
 Pass the total number of items to be paginated. If no value is defined an error will be thrown.
 </C.Property>
-<C.Property @name="currentPageSize" @type="number">
-Pass the maximum number of items to display on each page. If no value is defined, the first page size in `pageSizes` will be used. Default is `10` if custom `pageSizes` are not defined.
-</C.Property>
-<C.Property @name="pageSizes" @type="array" @values={{array "[10, 30, 50]" "array of integers" }} @default="[10, 30, 50]">
-Set the page sizes users can select from. Example: `@pageSizes=\{{array 5 20 30}}`.
-</C.Property>
 <C.Property @name="currentPage" @type="integer" @values={{array 1 "integer" }} @default={{1}}>
 Set a custom initial selected page.
 </C.Property>
-<C.Property @name="sizeSelectorLabel" @type="string" @default="Items per page">
-  Add a custom string for the label text overriding the default value.
+<C.Property @name="currentPageSize" @type="number">
+Pass the maximum number of items to display on each page. If no value is defined, the first page size in `pageSizes` will be used. Default is `10` if custom `pageSizes` are not defined.
+</C.Property>
+<C.Property @name="showLabels" @type="boolean" @default="false">
+Used to control the visibility of the "prev/next" text labels.
 </C.Property>
 <C.Property @name="isTruncated" @type="boolean" @default="true">
 Used to to limit the number of page numbers displayed (to save space, it will display an ellipsis for some numbers).
 </C.Property>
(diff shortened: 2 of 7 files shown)


didoo:
@WenInCode FYI 👆 (in relation to the [`Pagination` in the Consul UI Toolkit](https://github.com/hashicorp/consul-ui-toolkit/blob/main/toolkit/src/components/cut/list/pagination.hbs#L5-L37).

andrew-hashicorp:
This is just what I needed, thank you! 👨‍🍳's😘

andrew-hashicorp (approved):

website/docs/components/pagination/partials/code/component-api.md
```diff
@@ -69,18 +72,27 @@ This component supports use of [`...attributes`](https://guides.emberjs.com/rele
 <C.Property @name="ariaLabel" @type="string" @default="Pagination">
     Accepts a localized string.
 </C.Property>
+<C.Property @name="showLabels" @type="boolean" @default="true">
+Used to control the visibility of the "prev/next" text labels.
+</C.Property>
+<C.Property @name="isDisabledPrev/isDisabledNext" @type="boolean" @default="false">
+Used to disable the "prev" or "next" controls. Notice: when the control is disabled, it’s always rendered as an HTML `<button>` element.
+</C.Property>
+<C.Property @name="showSizeSelector" @type="boolean" @default="true">
```
alex-ju, line 81:
```suggestion
<C.Property @name="showSizeSelector" @type="boolean" @default="false">
```

alex-ju (approved):
Looks good! I left a comment in the API doc to reflect the `showSizeSelector` is `false` by default for the compact pagination.

Regarding the topic open for discussion, I think for the compact version it makes more sense to have the size selector hidden by default – by nature, the compact one should be minimal.

website/docs/components/pagination/partials/code/component-api.md
didoo: 🤦‍♂️

packages/components/addon/components/hds/pagination/compact/index.hbs
```diff
@@ -27,4 +27,13 @@
       @disabled={{@isDisabledNext}}
     />
   </nav>
+
+  {{#if this.showSizeSelector}}
+    <Hds::Pagination::SizeSelector
+      @pageSizes={{this.pageSizes}}
+      @label={{@sizeSelectorLabel}}
+      @selectedSize={{this.currentPageSize}}
```
andrew-hashicorp, line 35: I may be missing something, but there doesn't seem to be a way to set `this.currentPageSize` (i.e. this component doesn't have an `@currentPageSize` param.

didoo: @andrew-hashicorp you're right, I've missed it (for some reasons, I thought the sub-component was yielded). Let me investigate what needs to be done, I'll get back to you.

didoo: I've opened a new PR to "fix" this oversight: 
https://github.com/hashicorp/design-system/pull/1724

(but I need to discuss with you a detail before sending it to review, I'll set up some time to discuss it today)

PR #1705, opened 2023-10-10

hashibot-hds:


jorytindall:
Looks like none of the icons were actually changed, probably OK to close this?

Also note: a changeset wasn't created automatically for this. Is that an error in the pipeline?

PR #1704, opened 2023-10-10
Upgrade Ember to v5.3

alex-ju:
### :pushpin: Summary

- Upgrade Ember to the latest stable released version (5.3) using `ember-cli-update`
  - [x] `ember-flight-icons`
  - [x] `components`
  - [x] `website`
- Align our supported version of Node.js with Ember (this has already been addressed in #1708)

To be able to update to Ember 5.3 we needed to do a few things in conjunction:
 - ~~[Explicitly add dependency on `ember-get-config`](https://github.com/hashicorp/design-system/pull/1704/commits/caa5bdbc4863ffc468248b77f4ecc17188a5b534)~~ (addressed in a separate PR)
 - [Replace `babel-eslint` with `@babel/eslint-parser`](https://github.com/hashicorp/design-system/pull/1704/commits/57f3eb6ef8a7ecca52809068b3e1976b73049e8f) and configured babel to use proposed decorators
 - [Upgrade `ember-power-select`](https://github.com/hashicorp/design-system/pull/1704/commits/91b519af145af201cc678de3988b61e2d2a15f2b) as a dev dependency – we only use it for showcase purposes.
 - [Update `ember-cli-fastboot`](https://github.com/hashicorp/design-system/pull/1704/commits/fecfc11f3454001163ce146b35cb90c1d344c9b1) and [prember](https://github.com/hashicorp/design-system/pull/1704/commits/a51ef3d57c8d54022511b9a918e264446d9e37a9) in `website`
 - [Enforce typical resolutions for `ember-lts-3.28`](https://github.com/hashicorp/design-system/pull/1704/commits/4f702ccc1ad6ff4fe8923f63342a6146dcd0a86f)

**Note**: we decided to only update `devDependencies` in this PR to separate them out from `dependencies` that will require new releases for our packages. 

### :link: External links

Jira ticket: [HDS-2618](https://hashicorp.atlassian.net/browse/HDS-2618)

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [x] A11y tests have been run locally (`yarn test:a11y --filter="COMPONENT-NAME"`)
- [x] If documenting a new component, an acceptance test that includes the `a11yAudit` has been added
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[...]

PR #1708, opened 2023-10-12
Drop Node.js 14 support

alex-ju:
### :pushpin: Summary

Node.js 14 was end-of-life since April 30, 2023. As we prepare to [update to Ember 5.3](https://github.com/hashicorp/design-system/pull/1704) (which doesn't support 14 anymore) and dropping Node.js support is considered a breaking we take the opportunity of [the upcoming major release](https://github.com/hashicorp/design-system/pull/1634) to flag this change to our consumer

### :link: External links

Jira ticket: [HDS-2621](https://hashicorp.atlassian.net/browse/HDS-2621)

***

### 👀 Component checklist

[HDS-2621]: https://hashicorp.atlassian.net/browse/HDS-2621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/long-grapes-call.md
+++ b/.changeset/long-grapes-call.md
@@ -1,0 +1,6 @@
+---
+"@hashicorp/ember-flight-icons": major
+"@hashicorp/design-system-components": major
+---
+
+Drop support for Node 14
--- a/packages/components/README.md
+++ b/packages/components/README.md
@@ -11,7 +11,7 @@ Compatibility
 
 * Ember.js v3.28 or above
 * Ember CLI v3.28 or above
-* Node.js v14 or above
+* Node.js v16 or above
 
 
 Installation
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -140,7 +140,7 @@
     "webpack": "^5.84.1"
   },
   "engines": {
-    "node": "14.* || 16.* || >= 18"
+    "node": "16.* || >= 18"
   },
   "ember": {
     "edition": "octane"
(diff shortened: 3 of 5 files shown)


.changeset/long-grapes-call.md
```diff
@@ -0,0 +1,6 @@
+---
+"@hashicorp/ember-flight-icons": major
+"@hashicorp/design-system-components": major
```
didoo, line 3: Question: since in this PR we're just updating README files, in what technically this is a _major_ breaking change? 🤔

(just to understand)

alex-ju: When cherry-picking from #1704 I failed to port the `package.json` changes – have updated the PR now.

alex-ju: On the point of what constitutes the breaking change: it comes from us no longer guaranteeing that our packages work with Node 14. In practice this means our consumers may get a warning if they run their application with Node 14.

Dhaulagiri: > may get a warning if they run their application with Node 14.

FWIW I believe their apps will actually break in this case. The warning is for cases where you were using a _new_ version of Node.

Dhaulagiri (approved):

.changeset/long-grapes-call.md
alex-ju: Ah, interesting. I got warnings (in other applications) where I was told that I'm using a deprecated Node version but it went through with the deployment. In any case, that's why we flag it as breaking 😁

didoo (approved):
👍

PR #1709, opened 2023-10-12
`Tabs` - Handle `@isSelected` within `each` loop

didoo:
### :pushpin: Summary

This PR fixes one use case that was not covered before, where the `@isSelected` is dynamically set within a loop, where the condition of which tab is selected changes after the `Tabs` parent component has already rendered.

The problem was arising by the fact that there was a check in place, to avoid having multiple `Tab` elements with `@isSelected={{true}}` at the same time, but because of the nature of the Ember lifecycle and how this component is implemented under the hood, this is not possible to avoid: there are conditions (see screenshot below) where there are more `tab` in memory than the one actually rendered (the new tabs are _inserted_ but the old tabs are not yet _destroyed_).

screenshot_3145

In this PR I have:
- removed check on multiple "selected" tabs
- improved logic that handles the "selected" tab
    - now it handles also dynamic changes of the selected tab after the rendering of the tabs
- added a new example to the tabs showcase to cover also this use case

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2658

***

### 👀 Component checklist

### DIFF
--- a/.changeset/sharp-needles-kick.md
+++ b/.changeset/sharp-needles-kick.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Tabs` - Fixed issue with `@isSelected` dynamically changed within `#each` loops
--- a/packages/components/addon/components/hds/tabs/index.hbs
+++ b/packages/components/addon/components/hds/tabs/index.hbs
@@ -6,7 +6,9 @@
 <div
   class="hds-tabs"
   {{did-insert this.didInsert}}
-  {{did-update this.updateTabIndicator this.selectedTabIndex @isParentVisible}}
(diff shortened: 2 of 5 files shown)


alex-ju (approved):
I followed the logic here and it seems solid.

One thought that came to mind while reviewing the tabs work is that the first dynamic content example will always fail visual regression testing as it uses randomly generated content. We should probably initialize them with the same value and use the random function only on update. We can address this issue separately though.

didoo:
> One thought that came to mind while reviewing the tabs work is that the first dynamic content example will always fail visual regression testing as it uses randomly generated content. We should probably initialize them with the same value and use the random function only on update. We can address this issue separately though.

@alex-ju you're right!! ok, I'll open a separate PR, it should be quick

PR #1716, opened 2023-10-13
`Tabs` - replace `assert` with `warn` in `setIndicator`

didoo:
### :pushpin: Summary

Following this [report of a failing test in TFC](https://hashicorp.slack.com/archives/C03JP1TCGQM/p1697131130239269?thread_ts=1695236557.036929&cid=C03JP1TCGQM) after upgrading to `@hashicorp/design-system-components@2.14.1` (which contains the changes done in https://github.com/hashicorp/design-system/pull/1709) we have discussed the way in which we're throwing an assertion within the `setIndicator` function, and came the conclusion that we were being too strict: since we have decoupled the rendering of the tabs from the `setIndicator` execution (using `next() {}`) there may be cases in which the tabs are rendered and immediately after destroyed, leaving the `setIndicator` trying to apply to a set of tabs that don't exist anymore (this is the case in the tests, where the tabs were used in a loop of different conditions, each one rendering tabs and then destroying them at the end of the loop iteration). So we have decided that instead of an `assertion` (which blocks the JS execution) is better to just use a `warn` (that displays a message in the console, but without blocking the JS execution).

In this PR I have:
- replaced `assert` with `warn` in the `setIndicator function (and updated the warning message to handle two different possible cases)

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2662

***

### 👀 Component checklist

### DIFF
--- a/.changeset/olive-carpets-clean.md
+++ b/.changeset/olive-carpets-clean.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Tabs` - replace `assert` with `warn` in `setIndicator` function
--- a/packages/components/addon/components/hds/tabs/index.js
+++ b/packages/components/addon/components/hds/tabs/index.js
@@ -6,7 +6,7 @@
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
-import { assert } from '@ember/debug';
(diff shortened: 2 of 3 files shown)


alex-ju (approved):
Looks good. Initially, I left a comment about the test looking for the out-of-bound error, but I see that's fixed now by being removed.

PR #1687, opened 2023-10-03
`CodeBlock` component (HDS-2595 )

KristinLBradley:
### :pushpin: Summary
If merged, this PR will add the `CodeBlock` component to HDS.

**Showcase:** https://hds-showcase-git-hds-2595-codeblock-component-hashicorp.vercel.app/components/code-block

### :hammer_and_wrench: Detailed description

**Features**

`Title` and `Description` are available as contextual components to allow rich content.

Arguments:
* `@value`: string with the text/code content.
* `@isStandalone`: boolean (default = true)
* `@hasLineWrapping`: boolean (default = false)
* `@hasLineNumbers`: boolean (default = true)
* `@maxHeight`: string (any valid CSS unit, auto height by default)
  - if set, users need to vertically scroll to view overflow code lines
* `@hasCopyButton`: boolean (default = false)
  - if true, a button is displayed allowing users to copy code text
* ~~`@contentEditable`: boolean (default = false)~~
  - ~~if true, allows end-users to edit code text~~
* `@highlightLines`: string with a list or range of line numbers to highlight (Examples: "2, 4" | "6-10" )
* `@language`: undefined or a string with one of the pre-defined values 
  - we currently vendor a set of languages that provide highlight at runtime (bash, Go, HCL, JSON, log, shell-session, YAML)
  - we can extend these languages based on feedback from consumers

### :link: External links

- Jira ticket: [HDS-2595](https://hashicorp.atlassian.net/browse/HDS-2595)
- [Figma file](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/5gAU4mfvx8DvfXKb4QpjoB/HDS-Product---Components?type=design&node-id=44796-2787&mode=design&t=DfBNQj0hrXnYtGLX-0)
- [CRD](https://docs.google.com/document/d/1bFb6zo3tcMww9EBS-Wsd3qwRy1BQ8F4uXi-BCrb-ccw/edit#heading=h.gjdgxs)
- Original [RFC](https://docs.google.com/document/d/1RvtJrQ46eUe_su0f9VQhgJDuDDOcbAlxD7Sl2XIFwCs/edit)

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [x] A11y tests have been run locally (`yarn test:a11y --filter="COMPONENT-NAME"`)
- [x] If documenting a new component, an acceptance test that includes the `a11yAudit` has been added
[...]

PR #1721, opened 2023-10-13
Demo of auto prism language loading

Dhaulagiri:
### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist

### DIFF
--- a/packages/components/addon/components/hds/code-block/index.js
+++ b/packages/components/addon/components/hds/code-block/index.js
@@ -12,21 +12,19 @@ import Prism from 'prismjs';
 import { setup } from 'prismjs-glimmer';
 import { guidFor } from '@ember/object/internals';
 
+import { getOwnConfig, importSync } from '@embroider/macros';
+
+const config = getOwnConfig();
+
+// dynamically load languages defined in ember-cli-build
+for (const language of config.languages) {
+  importSync(`prismjs/components/prism-${language}.js`);
+}
+
 import 'prismjs/plugins/line-numbers/prism-line-numbers';
 import 'prismjs/plugins/line-highlight/prism-line-highlight';
 
-// Importing language individually because autoloader isn't currently working
-import 'prismjs/components/prism-go';
-import 'prismjs/components/prism-markdown';
-import 'prismjs/components/prism-shell-session';
-
-// languages_path isn't working so autoloader isn't working
-// https://prismjs.com/plugins/autoloader/
-import 'prismjs/plugins/autoloader/prism-autoloader';
-
 setup(Prism);
-// Path is supposed to normally not need to be specified but it's not working either way currently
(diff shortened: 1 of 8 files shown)


packages/components/index.js
```diff
@@ -12,8 +12,14 @@ module.exports = {
     return true;
   },
 
-  included: function (/* app */) {
+  included: function (app) {
     this._super.included.apply(this, arguments);
+    if (app.options && app.options['hds-code-block']) {
+      const options = app.options['hds-code-block'];
```
Dhaulagiri, line 18: [Example](https://github.com/froala/ember-froala-editor/blob/19fee2c9a2a0ec3a475a504773eb5ac88738fe6f/index.js#L123-L128) of how we could even validate that these languages exist

packages/components/index.js
```diff
@@ -12,8 +12,14 @@ module.exports = {
...
+      options.languages.forEach((language) => {
```
Dhaulagiri, line 19: [Example](https://github.com/froala/ember-froala-editor/blob/19fee2c9a2a0ec3a475a504773eb5ac88738fe6f/index.js#L123-L128) of how we could even validate that these languages exist

packages/components/index.js
```diff
@@ -25,5 +31,13 @@ module.exports = {
         'dialog-polyfill-css': 'dialog-polyfill/dist/dialog-polyfill.css',
       },
     },
+    '@embroider/macros': {
```
Dhaulagiri, line 34: [Docs](https://github.com/embroider-build/embroider/blob/main/packages/macros/README.md#setting-configuration-from-an-ember-addon)

packages/components/addon/components/hds/code-block/index.js
```diff
@@ -12,21 +12,19 @@ import Prism from 'prismjs';
 import { setup } from 'prismjs-glimmer';
 import { guidFor } from '@ember/object/internals';
 
-import 'prismjs/plugins/line-numbers/prism-line-numbers';
-import 'prismjs/plugins/line-highlight/prism-line-highlight';
+import { getOwnConfig, importSync } from '@embroider/macros';
```
Dhaulagiri, line 15: [Docs](https://github.com/embroider-build/embroider/blob/main/packages/macros/README.md#getownconfig-getconfig-and-getglobalconfig)

alex-ju (commented):
This looks really good! Thanks for taking the time to look into it. I'm happy with this approach, seems to be flexible enough. We can potentially provide a pre-defined list of languages (I believe doesn't need to be exhaustive or to match all available options in prism) and test in in our consumers' codebases. How would you like to move forward with this? Point this to #1687 and get it ready for a review, or shall I cherry-pick this work and continue in the component branch?

Dhaulagiri:
We can reference this but I'll close it for now as I think we'll go with another approach

PR #1728, opened 2023-10-17
button acceptance test with a11y audit

MelSumner:
### :pushpin: Summary

If merged this PR would add a button acceptance test for an a11y audit

- added acceptance test for button component
- removed instances where links were being rendered with the disabled attribute (invalid HTML, invalid WCAG)
- removed mock state for buttons with the `disabled` attribute

So I know how we talked about the fact that it's bad practice to disable an interactive element, but I don't think I realized that according to HTML spec, `disabled` is not a valid attribute for a `<a>` element: 
CleanShot 2023-10-17 at 09 16 15

Because of this, our links styled as buttons would be required to pass color contrast checks even if the links are not enabled. cc @hashicorp/hds-design 

### :link: External links

Jira ticket: [HDS-2555](https://hashicorp.atlassian.net/browse/HDS-2555)
Figma file: [if it applies]

***

### 👀 Component checklist

[HDS-2555]: https://hashicorp.atlassian.net/browse/HDS-2555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/tests/acceptance/components/hds/button-test.js
+++ b/packages/components/tests/acceptance/components/hds/button-test.js
@@ -1,0 +1,15 @@
+import { module, test } from 'qunit';
+import { visit } from '@ember/test-helpers';
+import { setupApplicationTest } from 'dummy/tests/helpers';
+import { a11yAudit } from 'ember-a11y-testing/test-support';
+
+module('Acceptance | Component | hds/button', function (hooks) {
+  setupApplicationTest(hooks);
+
+  test('Components/button passes a11y automated checks', async function (assert) {
+    await visit('/components/button');
+    await a11yAudit();
+
+    assert.ok(true, 'a11y automation audit passed');
+  });
+});
--- a/packages/components/tests/acceptance/components/hds/button-test.js
+++ b/packages/components/tests/acceptance/components/hds/button-test.js
@@ -7,8 +7,16 @@ module('Acceptance | Component | hds/button', function (hooks) {
   setupApplicationTest(hooks);
 
   test('Components/button passes a11y automated checks', async function (assert) {
+    // if making any changes to this component, enable the rule and run it to check that changes are still compliant
+    let axeOptions = {
+      rules: {
+        'color-contrast': {
+          enabled: false,
+        },
+      },
+    };
     await visit('/components/button');
-    await a11yAudit();
+    await a11yAudit(axeOptions);
 
     assert.ok(true, 'a11y automation audit passed');
   });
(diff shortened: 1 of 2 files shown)


packages/components/tests/dummy/app/templates/components/button.hbs
```diff
@@ -86,14 +86,42 @@
...
         {{#each @model.STATES as |state|}}
           <SG.Item @label="{{capitalize size}} / {{capitalize state}}">
-            <Hds::Button @icon="plus" @text="Lorem" @size={{size}} @color={{color}} mock-state-value={{state}} />
+            {{#if (eq state "disabled")}}
+              <Hds::Button
+                @icon="plus"
+                @text="Lorem"
+                @size={{size}}
+                @color={{color}}
+                mock-state-value={{state}}
```
didoo, line 95: I think you can remove the mocking of the state, since it's already applied as `disabled`
```suggestion
```

packages/components/tests/dummy/app/templates/components/button.hbs
```diff
@@ -86,14 +86,42 @@
...
         {{#each @model.STATES as |state|}}
           <SG.Item @label="Full width / {{capitalize state}}">
-            <Hds::Button @icon="plus" @text="Lorem" @color={{color}} @isFullWidth={{true}} mock-state-value={{state}} />
+            {{#if (eq state "disabled")}}
+              <Hds::Button
+                @icon="plus"
+                @text="Lorem"
+                @color={{color}}
+                @isFullWidth={{true}}
+                mock-state-value={{state}}
```
didoo, line 113: I think you can remove the mocking of the state, since it's already applied as `disabled`
```suggestion
```

packages/components/tests/acceptance/components/hds/button-test.js
```diff
@@ -0,0 +1,23 @@
...
+import { setupApplicationTest } from 'dummy/tests/helpers';
+import { a11yAudit } from 'ember-a11y-testing/test-support';
+
+module('Acceptance | Component | hds/button', function (hooks) {
+  setupApplicationTest(hooks);
+
+  test('Components/button passes a11y automated checks', async function (assert) {
+    // if making any changes to this component, enable the rule and run it to check that changes are still compliant
+    let axeOptions = {
+      rules: {
```
didoo, line 12: @MelSumner if we are disabling the color contrast tests for all the use cases, then no it doesn't make sense (I prefer to remove the failing use case at that point). I thought there was a way to disable a specific check only for an element (eg. via a DOM selector); hence the idea of an "exception" that I was suggesting.

didoo (commented):
@MelSumner I've left a couple of comments.

didoo (approved):
👍

PR #1729, opened 2023-10-17
Website: update language about using the `disabled` attribute on links.

MelSumner:
### :pushpin: Summary

If merged, this PR clarifies the color-contrast requirements on button components that render as links. 

### :link: External links

Jira ticket: [HDS-2675](https://hashicorp.atlassian.net/browse/HDS-2675)
Figma file: [if it applies]

***

### 👀 Component checklist

### DIFF
--- a/website/docs/components/button/partials/code/how-to-use.md
+++ b/website/docs/components/button/partials/code/how-to-use.md
@@ -132,7 +132,7 @@ To disable a Button, manually add the native `disabled` attribute:
 
 !!! Info
 
-If using an `@href` or `@route` and needing to disable the component, you’ll need to intercept the events since links can’t be disabled.
+Links cannot use the `disabled` attribute (per HTML specification); even if you were to intercept the event, they are still subject to color-contrast conformance requirements.
 !!!
 
 ```handlebars


website/docs/components/button/partials/code/how-to-use.md
```diff
@@ -132,7 +132,7 @@ To disable a Button, manually add the native `disabled` attribute:
 
 !!! Info
 
-If using an `@href` or `@route` and needing to disable the component, you’ll need to intercept the events since links can’t be disabled.
+Links cannot use the `disabled` attribute (per HTML specification); even if you were to intercept the event, they are still subject to color-contrast conformance requirements.
```
didoo, line 135: @heatherlarsen should we mention here that, also, there are specific guidance at product level (in Cloud UI for sure, but we may try to extend it to other products too) that we should try to avoid showing disabled elements to the users?

jorytindall (approved):
It's a bit verbose, but probably makes sense given the context in the "How to use" section

website/docs/components/button/partials/code/how-to-use.md
heatherlarsen: @didoo There is specific guidance at the product level for most of our components; what makes this scenario unique enough to document? To me, it feels more like an educational problem to solve, not a documentation problem. wdyt?

We also have our own guidance around the usage of disabled states in the [States](https://helios.hashicorp.design/components/button?tab=specifications#states) section.

didoo: OK

PR #1730, opened 2023-10-17
`Button/ButtonSet` - Documentation for "equal size" and "loading states"

didoo:
### :pushpin: Summary

This is a follow-up of a discussion we had within the team, to respond to different requests of guidance from our consumers around buttons with loading states and 

In this PR I have:
- updated the `Button/ButtonSet` documentation pages to include “equal width” and “loading states”
- added a few use cases to the `Button/ButtonSet` showcases for equal width and loading states

👉 👉 👉 **Previews:**
- website:
    - button: https://hds-website-git-equal-size-for-button-and-load-d1d389-hashicorp.vercel.app/components/button?tab=code#loading-state
    - button-set: https://hds-website-git-equal-size-for-button-and-load-d1d389-hashicorp.vercel.app/components/button-set?tab=code#equal-width-buttons
- showcase:
    - button: https://hds-showcase-git-equal-size-for-button-and-loa-c0f2f9-hashicorp.vercel.app/components/button
    - button-set: https://hds-showcase-git-equal-size-for-button-and-loa-c0f2f9-hashicorp.vercel.app/components/button-set

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2676

***

### 👀 Component checklist

### DIFF
--- a/website/docs/components/button-set/partials/code/how-to-use.md
+++ b/website/docs/components/button-set/partials/code/how-to-use.md
@@ -1,8 +1,42 @@
 ## How to use this component
 
+The basic invocation requires two or more buttons to be provided as children:
+
 ```handlebars
 <Hds::ButtonSet>
   <Hds::Button @text="Submit" type="submit" />
   <Hds::Button @text="Cancel" @color="secondary" @href="https://hashicorp.com" />
+</Hds::ButtonSet>
+```
+
+### Equal width buttons
+
+If you want to have buttons with equal width, apply a width (via inline style or via CSS class) to the `ButtonSet` container, and the argument `@isFullWidth={{true}}` to the `Button` elements:
+
+```handlebars
+<Hds::ButtonSet {{style width="15rem"}}>
+  <Hds::Button @text="Save" @isFullWidth={{true}} />
+  <Hds::Button @text="Cancel" @color="secondary" @href="https://hashicorp.com" @isFullWidth={{true}} />
+</Hds::ButtonSet>
+```
+
+#### With loading state
+
+This technique is useful if you need to show a loading state, to avoid the resizing and shifting of the buttons:
+
+```handlebars
+<Hds::ButtonSet {{style width="15rem"}}>
+  <Hds::Button
+    @icon={{if this.isLoading "loading"}}
+    @text={{if this.isLoading "Loading" "Save"}}
+    @isFullWidth={{true}}
+    {{on "click" this.toggleIsLoading}}
+  />
+  <Hds::Button
+    @text="Cancel"
+    @color="secondary"
+    @isFullWidth={{true}}
+    {{on "click" this.cancelLoading}}
+  />
 </Hds::ButtonSet>
 ```
--- a/website/docs/components/button-set/partials/code/how-to-use.md
+++ b/website/docs/components/button-set/partials/code/how-to-use.md
@@ -11,7 +11,7 @@ The basic invocation requires two or more buttons to be provided as children:
 
 ### Equal width buttons
 
-If you want to have buttons with equal width, apply a width (via inline style or via CSS class) to the `ButtonSet` container, and the argument `@isFullWidth={{true}}` to the `Button` elements:
+If you want to have buttons with equal width, apply a width (via inline style or CSS class) to the `ButtonSet` container and the argument `@isFullWidth={{true}}` to the `Button` components:
 
 ```handlebars
 <Hds::ButtonSet {{style width="15rem"}}>
--- a/website/docs/components/button/partials/code/how-to-use.md
+++ b/website/docs/components/button/partials/code/how-to-use.md
@@ -126,6 +126,25 @@ If the route is external to your current engine, you have to pass `@isRouteExter
 <Hds::Button @text="Back to homepage" @icon="arrow-left" @route="index" />
 ```
 
+### Loading state
+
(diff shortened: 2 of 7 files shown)


website/docs/components/button/partials/code/how-to-use.md
```diff
@@ -126,6 +126,25 @@ If the route is external to your current engine, you have to pass `@isRouteExter
 <Hds::Button @text="Back to homepage" @icon="arrow-left" @route="index" />
 ```
 
+### Loading state
+
+If the button needs to toggle between an "idle" and a "loading" state, we suggest to apply a width to it (via inline style or via CSS class) to prevent the button to resize on click (and potentially cause layout shifts):
+
+!!! Info
+
+While applying an explicit with to the button it's possible in general, we suggest to limit the application of this override **only** to this specific use case and let the button resize accordingly to its content.
```
heatherlarsen, line 135:
```suggestion
While applying an explicit width to the button is possible in general, we suggest limiting the application of this override **only** to this specific use case and letting the button resize accordingly to its content.
```

website/docs/components/button/partials/code/how-to-use.md
```diff
@@ -126,6 +126,25 @@ If the route is external to your current engine, you have to pass `@isRouteExter
 <Hds::Button @text="Back to homepage" @icon="arrow-left" @route="index" />
 ```
 
+### Loading state
+
+If the button needs to toggle between an "idle" and a "loading" state, we suggest to apply a width to it (via inline style or via CSS class) to prevent the button to resize on click (and potentially cause layout shifts):
```
heatherlarsen, line 131:
```suggestion
If the button needs to toggle between an "idle" and a "loading" state, we suggest applying a width to it (via inline style or CSS class) to prevent the button from resizing on click (and potentially causing layout shifts):
```

website/docs/components/button-set/partials/code/how-to-use.md
```diff
@@ -1,8 +1,42 @@
...
+```handlebars
+<Hds::ButtonSet {{style width="15rem"}}>
+  <Hds::Button @text="Save" @isFullWidth={{true}} />
+  <Hds::Button @text="Cancel" @color="secondary" @href="https://hashicorp.com" @isFullWidth={{true}} />
+</Hds::ButtonSet>
+```
+
+#### With loading state
+
+This technique is useful if you need to show a loading state, to avoid the resizing and shifting of the buttons:
```
heatherlarsen, line 25:
```suggestion
If you need to show a loading state, this technique is useful to avoid the resizing and shifting of the buttons:
```

website/docs/components/button-set/partials/code/how-to-use.md
```diff
@@ -1,8 +1,42 @@
...
 ```handlebars
 <Hds::ButtonSet>
   <Hds::Button @text="Submit" type="submit" />
   <Hds::Button @text="Cancel" @color="secondary" @href="https://hashicorp.com" />
 </Hds::ButtonSet>
 ```
+
+### Equal width buttons
+
+If you want to have buttons with equal width, apply a width (via inline style or via CSS class) to the `ButtonSet` container, and the argument `@isFullWidth={{true}}` to the `Button` elements:
```
heatherlarsen, line 14:
```suggestion
If you want to have buttons with equal width, apply a width (via inline style or CSS class) to the `ButtonSet` container and the argument `@isFullWidth={{true}}` to the `Button` elements:
```

heatherlarsen (commented):
I left a few grammar suggestions. 

I also noticed one thing in the showcase for the ButtonSet...the alignment in the Secondary Button in the second example is left aligned, while the Primary Button alignment is centered. They should both be centered. It's unrelated to this task, so it can be fixed separately, but I wanted to note it as a bug in case it's an easy fix.

website/docs/components/button-set/partials/code/how-to-use.md
didoo: With "this technique" I was referring to the technique (maybe I should use the word "approach") described in the previous example. In that case does your suggestion still apply?

didoo:
> I also noticed one thing in the showcase for the ButtonSet...the alignment in the Secondary Button in the second example is left aligned, while the Primary Button alignment is centered. They should both be centered. It's unrelated to this task, so it can be fixed separately, but I wanted to note it as a bug in case it's an easy fix.

@heatherlarsen yes, you're absolutely right. The reason for this is that we never applied (and we never noticed the "problem") an explicit text alignment to the text in the button, which means it uses the defaults of the parent elements:
- for a `<button> ` element, the text is centered by default
- for an `<a>` element, the text inherits the alignment by the parent (so left aligned by default)

When the button auto-resizes, and is on a single line, the difference is not noticeable. But when the text is multiline and you have a `<button>` next to an `<a>` like in the case you mentioned, then it's visible (not sure how we missed it before).

In any case, I've opened a [Jira ticket](https://hashicorp.atlassian.net/browse/HDS-2679) and created a PR for this: https://github.com/hashicorp/design-system/pull/1732

website/docs/components/button-set/partials/code/how-to-use.md
alex-ju: Would it be more accurate to say `Button` components instead of elements?

alex-ju (approved):
Looks good

website/docs/components/button-set/partials/code/how-to-use.md
didoo: Yes, it would.

PR #1736, opened 2023-10-20
`Pagination` - Remove handling of query parameters on page change for the "compact" variant

didoo:
### :pushpin: Summary

@andrew-hashicorp has raised an issue with the adoption of the fix done in #1724:

screenshot_3169

Upon further investigation, I think he’s right. We (I) ported the implementation from the “Numbered” variant, but actually also that one is flawed: the `onPageSizeChange` should only change the values when is not controlled; when is controlled it should delegate the responsibility to the consumers (they just need to update the value of the query parameter without needing to call a `router.transitionTo`, see `onPageSizeChange_demo4` on this PR).

Demonstration of this is the fact that once you remove the query params update on the `onPageSizeChange` action, you can also get rid of the `routeQueryParams` getter, which leads to removing entirely the need to import the `@service router` in the component (a code smell).

In this PR I have:
- removed route query handling on “page size” change for the “compact” pagination
    - **notice**: in theory we could do the same also for the “numbered” pagination (I tested it, and it works, all the integration/acceptance tests pass) but it would be a breaking change so we need to do it in a [different ticket in the future](https://hashicorp.atlassian.net/browse/HDS-2729)
- updated showcase for “compact” pagination to handle query update on page size change

Re. the changeset version number, technically is a breaking change, but since we released the previous change 12 hours ago I am going to consider this a fast-follow fix, and use a "patch" for it.

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2728

***

### 👀 Component checklist

### DIFF
--- a/.changeset/loud-readers-flash.md
+++ b/.changeset/loud-readers-flash.md
@@ -1,0 +1,7 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Pagination` - Removed handling of query parameters from `onPageSizeChange` function.
+
+_Notice: while technically this is a breaking change, we consider this a fast-follow fix for the previous release._
--- a/packages/components/addon/components/hds/pagination/compact/index.js
+++ b/packages/components/addon/components/hds/pagination/compact/index.js
@@ -7,15 +7,12 @@ import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { assert } from '@ember/debug';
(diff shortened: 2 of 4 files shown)


alex-ju (approved):
makes sense – we should probably track a similar change for the “numbered” pagination

didoo:
> makes sense – we should probably track a similar change for the “numbered” pagination

@alex-ju https://hashicorp.atlassian.net/browse/HDS-2729

PR #1691, opened 2023-10-04
`CodeBlock`: docs (deprecated branch)

KristinLBradley:
This branch is deprecated in favor of #1749

**Web docs preview:** https://hds-website-git-hds-2595-codeblock-docs-hashicorp.vercel.app/components/code-block

**Note to reviewers:** Due to the fact that we use Prism both when converting markdown to HTML to highlight the component code and within the CodeBlock component we currently encounter highlighting issues on the Code Block page.

### :pushpin: Summary

If merged, this PR will add the `CodeBlock` documentation to the component branch.

### :link: External links

- Jira ticket: [HDS-2595](https://hashicorp.atlassian.net/browse/HDS-2595)
- [Figma file](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/5gAU4mfvx8DvfXKb4QpjoB/HDS-Product---Components?type=design&node-id=44796-2786&mode=design&t=TUiY0Wlz97AKPyd5-0)

***

### 👀 Component checklist

- [ ] Percy was checked for any visual regression
- [ ] A11y tests have been run locally (`yarn test:a11y --filter="COMPONENT-NAME"`)
- [ ] If documenting a new component, an acceptance test that includes the `a11yAudit` has been added
- [ ] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-2595]: https://hashicorp.atlassian.net/browse/HDS-2595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1737, opened 2023-10-20
`Website` - Images optimization

didoo:
### :pushpin: Summary

While reviewing https://github.com/hashicorp/design-system/pull/1691/ I noticed that the illustration image for the component was not optimized (was able to reduce by 75% its side) which made me realize that we have probably stopped checking this some time ago.

In this PR I am have optimized all the existing assets using a Mac app ([ImageOptim](https://imageoptim.com/mac)) but maybe we should automate the process (@Dhaulagiri) if it's not yet happening in the ember cli build step (frankly, I don't remember, I have a vague recollection we discussed this, and it may be that we decided it was for later to implement).

In this PR I have 
- optimized all the images under the `public/assets` folder

Previews: 
- before: https://helios.hashicorp.design/
- after: https://hds-website-git-website-image-optimization-hashicorp.vercel.app/

screenshot_3182

***

### 👀 Component checklist


Dhaulagiri:
> I have a vague recollection we discussed this, and it may be that we decided it was for later to implement

We did capture this [in jira](https://hashicorp.atlassian.net/browse/HDS-1353) but never got around to do it

Dhaulagiri:
Can you have this PR be only the icon for the CodeBlock and make a separate one pointed at `main` for the rest (or make that a commit directly on that PR) so we can merge in the rest of the changes without tying them up with the CodeBlock work?

didoo:
This is not urgent, so as soon as the CodeBlock is merged we merge also this one

alex-ju (approved):
I don't think ember cli build has compression baked in out of the box – however, we could add an automated step in the future.

didoo:
> I don't think ember cli build has compression baked in out of the box 

@alex-ju unfortunately not.

> however, we could add an automated step in the future.

Tracked here: https://hashicorp.atlassian.net/browse/HDS-1353
(I did a quick spike the other day and... it's not a quick thing to do 😀)

heatherlarsen (approved):
Great catch! If we can't easily automate the process, is there something we should do after exporting the images from Figma? We can add it as a part of our workflow, if that's easier?

didoo:
> is there something we should do after exporting the images from Figma? We can add it as a part of our workflow, if that's easier?

@heatherlarsen I personally use https://imageoptim.com/mac (which is free). Then drag&drop the images I want to optimize over the window, and wait for the process to complete (it takes some time).
These are the settings I use: 
screenshot_3195

PR #1742, opened 2023-10-23
yarn 4

Dhaulagiri:
### :pushpin: Summary

https://yarnpkg.com/blog/release/4.0


wojtekmaj:
Not sure if you know, but if you use Corepack and `packageManager`, then you don't need `yarnPath` and `.yarn/releases` anymore at all! It's only an escape hatch for people with Corepack disabled & Yarn v1 (still) installed.

aklkv:
I know it might be a bigger lift but ember-cli is moving towards pnpm, it might a good option to try https://rfcs.emberjs.com/id/0907-pnpm-support/

Dhaulagiri:
@aklkv that's good to know but is likely something we would investigate in the future. at least for now yarn serves our needs well.

Dhaulagiri:
@wojtekmaj thanks for those tips. I think for now we'll leave that config in place and explore how/if to use Corepack as a future step.

alex-ju (approved):
This looks good! Always nervous when adopting an x.0.0 release 😄 . I would be really grateful if we merge this after all the other package updates in the pipeline #1704, #1755, #1756 as it affects every single dependency (new checksums) and will create a rebase hell.

Dhaulagiri:
> Always nervous when adopting an x.0.0 release 😄 . 

That's a fair point. There's not particular rush for this, so we _could_ always wait.

> I would be really grateful if we merge this after all the other package updates in the pipeline

💯 not a problem

alex-ju (commented):
This should be good to be merged now. You'll have to regenerate `yarn.lock` I'm afraid

PR #1750, opened 2023-10-25
Fix `pre-build` script invocation + Update release notes

didoo:
### :pushpin: Summary

Context: 
screenshot_3204

Upon investigation, we have found that `yarn` does not execute the `pre/post` scripts automatically:

screenshot_3203

See: https://github.com/yarnpkg/yarn/pull/800

From [the Yarn documentation](https://yarnpkg.com/advanced/lifecycle-scripts):

> In particular, we intentionally don’t support arbitrary pre and post hooks for user-defined scripts

In this PR I have:
- updated the node scripts
- re-run the scripts to update the “release notes” files

***

### 👀 Component checklist

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -15,8 +15,8 @@
     "test": "tests"
   },
   "scripts": {
-    "prebuild": "node lib/changelog/generate-changelog-markdown-files.mjs",
-    "build": "ember build --environment=production",
+    "generate-changelog-markdown-files": "node lib/changelog/generate-changelog-markdown-files.mjs",
+    "build": "yarn generate-changelog-markdown-files && ember build --environment=production",
     "lint": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"",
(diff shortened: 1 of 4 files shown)


website/package.json
```diff
@@ -15,8 +15,8 @@
     "test": "tests"
   },
   "scripts": {
-    "prebuild": "node lib/changelog/generate-changelog-markdown-files.mjs",
-    "build": "ember build --environment=production",
+    "generate-changelog-markdown-files": "node lib/changelog/generate-changelog-markdown-files.mjs",
+    "build": "yarn generate-changelog-markdown-files && ember build --environment=production",
```
Dhaulagiri, line 19: I'm curious the thinking of including this in the build as well: is this a workaround for the fact that `prebuild` doesn't actually do anything? If we are updating this as part of the `version` step I _think_ this would not actually ever trigger new changes. Am I thinking about that correctly?

didoo: it's just in case, until we find out if the `yarn version` works, then we can remove it

Dhaulagiri (approved):

PR #1755, opened 2023-10-27
Upgrade patch and minor `devDependencies`

alex-ju:
### :pushpin: Summary

Upgrade all `devDependencies` (patch and minors) across the monorepo.

These are upgrades for our tooling dependencies and do not require releases for our packages. Separating them per package is virtually impossible because of how `yarn workspaces` works, but I separated each dependency on its own commit to be able to test the impact of each bump and be to revert quickly if needed.

This branch is based on #1704 to avoid having too many conflicts (especially in `yarn.lock` 😱 ), with the intent to repoint it and merge it to `main` after #1704 gets in.

### :link: External links

Jira ticket: [HDS-2619](https://hashicorp.atlassian.net/browse/HDS-2619)

***


packages/components/addon/components/hds/button/index.ts
```diff
@@ -5,7 +5,7 @@
 
 import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
-import { HdsInteractiveSignature } from '../interactive';
+import { type HdsInteractiveSignature } from '../interactive';
```
Dhaulagiri, line 8: [question] this is technically a user facing change that would require a changeset, yes? If so, can we pull it out into a separate PR or pull it into one of the ones that already has user facing changes?

Dhaulagiri (approved):
LGTM other than my one comment

packages/components/addon/components/hds/button/index.ts
alex-ju: ah, this little one made its way in – good catch! it's enforced by the tsconfig update. will pull it out

PR #1759, opened 2023-10-31
Only run integration tests in `ember-try`

alex-ju:
### :pushpin: Summary

To skip running the a11y tests in `ember-try`

The default command is `ember test --reporter xunit`, we add a filter to only run integration tests (hence skip the acceptance ones) to save us some time. The acceptance tests only contain the a11y tests, which are already running in the main test suite (outside of `ember-try`)


MelSumner (approved):
:shipit:

PR #1760, opened 2023-10-31
Remove unnecessary trailing slash from canonical url

jorytindall:
### :pushpin: Summary

In looking for some of the metadata that is rendered along with various pieces of content for the search project, I found a small bug in the getter for the canonical URL in which an extra trailing slash is added to the root domain. This probably doesn't impact anything in reality, but I figured it was a quick fix?

Otherwise, disregard and 🗑️ this!

Before:
Screenshot 2023-10-31 at 9 50 48 AM

After:
Screenshot 2023-10-31 at 9 51 06 AM

***

### DIFF
--- a/website/app/services/head-data.js
+++ b/website/app/services/head-data.js
@@ -36,7 +36,7 @@ export default class CustomHeadDataService extends HeadDataService {
     const url =
       this.router.currentURL === '/'
         ? config['ember-meta'].url
-        : `${config['ember-meta'].url}/${this.router.currentURL}`;
+        : `${config['ember-meta'].url}${this.router.currentURL}`;
     return url;
   }
 }


Dhaulagiri:
this does serve a purpose, otherwise these links are missing that `/`, but it _would_ be nice to figure out how to avoid it being duplicated like you're seeing

Screenshot 2023-10-31 at 11 58 23

didoo:
> it _would_ be nice to figure out how to avoid it being duplicated

@Dhaulagiri this is what I see in the console when I do:
```
console.log(`config['ember-meta'].url = `, config['ember-meta'].url);
console.log('this.router.currentURL = ', this.router.currentURL);
```
in that function: 
screenshot_3236

I have no idea what is adding the extra `/` character. Hints where to look into?

Dhaulagiri:
Ultimately this is a problem that would be good to solve, however it would require more digging I think. I recall this being confusing when I originally implemented this, and what we have is the best I could come up with at the time.

jorytindall:
> this does serve a purpose

Interesting, I'm not seeing the same result when I inspect that branch. If it's a bigger issue though, I'll close this for now and open a ticket to address it in the future

Dhaulagiri:
@jorytindall which branch? the one for this PR (which is where I got the screenshot)?

jorytindall:
@Dhaulagiri correct, the preview from this branch. This is what I'm seeing 🤔 
Screenshot 2023-11-01 at 10 00 32 AM

jorytindall:
Closing this to solve in the future.

Dhaulagiri:
For context on the above back and forth, the difference was between "View Source" and "Inspect" views of the underlying markup

PR #1761, opened 2023-11-01
Upgrade major `dependencies`

alex-ju:
### :pushpin: Summary

Upgrade most (exceptions detailed below) `dependencies` (majors) across the monorepo.

These are upgrades that will be passed on to our consumers and require releases for our packages.

This branch is based on #1704 to avoid having too many conflicts, with the intent to repoint it and merge it to `main` after #1704 gets in.

Exceptions:
 - `ember-stargate` and `ember-truth-helpers` were updated to v2 add-ons and we're not yet ready to consume them and/or impose them to our consumers.

### :link: External links

Jira ticket: [HDS-2619](https://hashicorp.atlassian.net/browse/HDS-2619)

***

[HDS-2619]: https://hashicorp.atlassian.net/browse/HDS-2619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/popular-hornets-promise.md
+++ b/.changeset/popular-hornets-promise.md
@@ -1,0 +1,9 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Upgraded the following dependencies:
+ - `ember-cached-decorator-polyfill` from `0.1.4` to `1.0.2`
+ - `ember-cli-babel` from `7.26.11` to `8.2.0`
+ - `ember-cli-sass` from `10.0.1` to `11.0.1`
+ - `ember-composable-helpers` from `4.5.0` to `5.0.0`
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -48,7 +48,7 @@
     "ember-a11y-refocus": "^3.0.2",
     "ember-auto-import": "^2.6.3",
     "ember-cached-decorator-polyfill": "^1.0.2",
-    "ember-cli-babel": "^7.26.11",
+    "ember-cli-babel": "^8.2.0",
     "ember-cli-htmlbars": "^6.3.0",
     "ember-cli-sass": "^10.0.1",
     "ember-composable-helpers": "^4.5.0",
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -47,7 +47,7 @@
     "dialog-polyfill": "^0.5.6",
     "ember-a11y-refocus": "^3.0.2",
     "ember-auto-import": "^2.6.3",
-    "ember-cached-decorator-polyfill": "^0.1.4",
+    "ember-cached-decorator-polyfill": "^1.0.2",
     "ember-cli-babel": "^7.26.11",
     "ember-cli-htmlbars": "^6.3.0",
     "ember-cli-sass": "^10.0.1",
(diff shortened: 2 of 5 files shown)


packages/components/package.json
```diff
@@ -47,11 +47,11 @@
     "dialog-polyfill": "^0.5.6",
     "ember-a11y-refocus": "^3.0.2",
     "ember-auto-import": "^2.6.3",
-    "ember-cached-decorator-polyfill": "^0.1.4",
-    "ember-cli-babel": "^7.26.11",
+    "ember-cached-decorator-polyfill": "^1.0.2",
+    "ember-cli-babel": "^8.2.0",
```
Dhaulagiri, line 51: [question] do we understand what the impact will be for consumers like cloud-ui who are still on 7.x?

.changeset/popular-hornets-promise.md
```diff
@@ -0,0 +1,9 @@
+---
+"@hashicorp/design-system-components": patch
```
Dhaulagiri, line 2: [question] I think we've discussed this before so forgive my poor memory, but since these are major changes upstream would a `minor` be a better way to communicate these changes?

Dhaulagiri (approved):
Approving but left a couple of questions.

> were updated to v2 add-ons and we're not yet ready to consume them and/or impose them to our consumers.

Perhaps something to take offline but I'm curious to understand more why their migration to V2 addon is a blocker for us as that is counter to what I would expect.

.changeset/popular-hornets-promise.md
alex-ju: I know there is no standard in the industry around this, unfortunately. My thinking is that if dependency bump (be that a major) materializes in a feature for our consumers (let's say tippy allows a new option that cascades in our component) then it's a feature. In all other instances it's a patch. Changesets and Lerna operate similarly. For example when we bumped `@hashicorp/ember-flight-icons` to `3.0.0` this was forwarded by changesets into a patch for `@hashicorp/design-system-components`. I'm happy to review this as a team if we feel differently about it. I'm aware that we might enforce these majors to our consumers, and that's not ideal.

packages/components/package.json
alex-ju: we understand it [in theory](https://github.com/emberjs/ember-cli-babel/blob/master/CHANGELOG.md#boom-breaking-change) (it's made of a `brocolli-babel-transpiler` bump and a node version drop), but it's a good call to see how they resolve in our consumer's projects (will do just that). if it gets ugly we might need to postpone some of them

.changeset/popular-hornets-promise.md
Dhaulagiri: > Changesets and Lerna operate similarly

Good point and good context. I think your explanation makes sense to keep it as you have it here.

packages/components/package.json
alex-ju: Tested this branch in our consumers' codebases.

The likes of `cloud-ui` resolve them well, `atlas` (and any other products that use `dependency-lint`) requires a few bumps or, alternatively, the following manual resolution:
```
    "@ember/test-waiters": "3.0.2",
    "ember-cached-decorator-polyfill": "0.1.4",
    "ember-composable-helpers": "4.5.0",
    "ember-element-helper": "0.6.1",
```

PR #1762, opened 2023-11-01
Update focus details for `Modal`

alex-ju:
### :pushpin: Summary

Trying to clarify that the consumer is responsible for setting the focus on a relevant element.

### DIFF
--- a/website/docs/components/modal/partials/accessibility/accessibility.md
+++ b/website/docs/components/modal/partials/accessibility/accessibility.md
@@ -6,7 +6,7 @@ The Modal component is conformant when used as directed.
 
 ## Focus and focus order
 
-- When the Modal is triggered via the keyboard, focus is trapped within the Modal. 
+- When the modal is open, focus moves to the first interactive element within the modal and is trapped within the component.
 - Since a Modal is a complex pattern that can contain any combination of nested components and content, nested elements must adhere to their individual accessibility criteria.
 
 ### Focus order within a simple Modal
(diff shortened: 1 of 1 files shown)


website/docs/components/modal/partials/accessibility/accessibility.md
```diff
@@ -17,7 +19,7 @@ Given the Modal is triggered via a keyboard, the dismiss button is first in the
 
 ### Focus order within a complex Modal
 
-If the Modal body contains interactive content, the first element should receive focus first. This is true regardless of how the Modal is triggered; either via a mouse click or via the keyboard.
+If the Modal body contains interactive content, such as input fields, the first element should receive focus first regardless of how the Modal is triggered; either via a mouse click or via the keyboard. This behaviour should be implemented by [automatically setting the focus on a the first element](/components/modal?tab=code#form-within-a-modal-dialog).
```
didoo, line 22:
```suggestion
If the Modal body contains interactive content, such as input fields, the first element should receive focus first regardless of how the Modal is triggered; either via a mouse click or via the keyboard. This behaviour should be implemented by [setting the focus on the first element](/components/modal?tab=code#form-within-a-modal-dialog).
```

didoo: in case, replace `automatically` with `manually` (as a consumer, I have to "manually" do something)

didoo (approved):

website/docs/components/modal/partials/accessibility/accessibility.md
alex-ju: yeah, I can see how 'automatically' may be misinterpreted there. I was referring to the fact that the focus itself is automatically set by the browser (and not by a script) based on the attribute, but even if accurate it's not relevant.

website/docs/components/modal/partials/accessibility/accessibility.md
```diff
@@ -6,8 +6,10 @@ The Modal component is conformant when used as directed.
 
 ## Focus and focus order
 
-- When the Modal is triggered via the keyboard, focus is trapped within the Modal. 
+- When the Modal opens, focus moves to an element contained in the Modal.
```
KristinLBradley, line 9: Should this be an interactive element? Would we want to give any details on what type of element?

jorytindall: I agree, this reads a bit ambiguous to me. There's more clarity that you added below that is useful, I wonder if this bullet point and the next could be combined somehow:

"When the modal is open, focus moves to the first interactive element within the modal and is trapped within the component."

That's not really perfect either, but throwing out some options.

KristinLBradley (approved):

jorytindall (approved):

PR #1772, opened 2023-11-03
Upgrade `stylelint-config-standard-scss` to `11.1.0`

alex-ju:
### :pushpin: Summary

The standard SCSS via `stylelint` has changed quite a bit lately. Planning to bring these to the next engineering sync if we don't get a consensus through reviews.

#### Deprecated rules
 - [function-parentheses-space-inside](https://stylelint.io/user-guide/rules/function-parentheses-space-inside/)
 - [max-empty-lines](https://stylelint.io/user-guide/rules/max-empty-lines/)
 - [number-leading-zero](https://stylelint.io/user-guide/rules/number-leading-zero/)
 - [string-quotes](https://stylelint.io/user-guide/rules/string-quotes/)

These are [flagged as deprecated in `stylelint`](https://github.com/stylelint/stylelint/blob/main/docs/user-guide/rules.md#deprecated) and will be retired. We'd need to decide whether we'd like to just drop them (or some of them) or try to cover them via `prettier`.

#### New rules
 - [media-feature-range-notation](https://stylelint.io/user-guide/rules/media-feature-range-notation/)

Luckily the new rules only seem to be affecting `website`, so no changes to our consumers. We'll probably need to decide if we want to override this default or we'll stick with it.

### :link: External links

Jira ticket: [HDS-2461](https://hashicorp.atlassian.net/browse/HDS-2461)

https://hashicorp.atlassian.net/browse/HDS-2461

[HDS-2461]: https://hashicorp.atlassian.net/browse/HDS-2461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/.prettierrc.js
+++ b/packages/components/.prettierrc.js
@@ -10,5 +10,11 @@ module.exports = {
         printWidth: 120,
       },
     },
+    {
+      files: '*.scss',
+      options: {
+        singleQuote: false,
+      },
+    },
   ],
 };
--- a/packages/components/.stylelintrc.js
+++ b/packages/components/.stylelintrc.js
@@ -24,14 +24,6 @@ module.exports = {
     // see: https://stylelint.io/user-guide/rules/list/declaration-empty-line-before/
     'declaration-empty-line-before': null,
 
-    // we've decided to never have spaces inside the parentheses of functions (we all prefer this way, for readability)
-    // see: https://stylelint.io/user-guide/rules/list/function-parentheses-space-inside/
-    'function-parentheses-space-inside': 'never',
-
-    // we've agreed that in some cases is useful to add extra visual space between blocks to better separate/group content
-    // see: https://stylelint.io/user-guide/rules/list/max-empty-lines/
-    'max-empty-lines': 2,
-
     // we've agreed to not enforce a limit on the lines lenght (we embed SVGs so very long strings + plus modern browsers can esily soft-wrap code)
     // see: https://stylelint.io/user-guide/rules/list/max-line-length/
     'max-line-length': null,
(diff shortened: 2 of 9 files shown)


KristinLBradley:
I would agree with removing this one:
* [function-parentheses-space-inside](https://stylelint.io/user-guide/rules/function-parentheses-space-inside/)

Others:
* [max-empty-lines](https://stylelint.io/user-guide/rules/max-empty-lines/) - I prefer to limit empty lines to 2 max as we do currently.
* [number-leading-zero](https://stylelint.io/user-guide/rules/number-leading-zero/) - I think it helps readability to include a "0" before the fraction but it's not really a big deal to me so I'm fine either way. (I often am not too consistent on this one myself if not required to be.)
* [string-quotes](https://stylelint.io/user-guide/rules/string-quotes/) - my personal preference is to use single quotes within JS and double quotes within CSS & HTML. Sometimes there are reasons not to use single quotes in JS and in HTML containing JS but I can't think of a real reason to need an exception in CSS off the top of my head so I'd prefer either enforcing double-quotes in CSS or else leaving it open (but not changing to enforce single quotes).

I think it would be helpful to discuss these rules together though.

KristinLBradley (commented):
Left some comments. This looks good overall to me but I think we should discuss the CSS lint rule changes together first.

didoo:
Personally:
- function-parentheses-space-inside - no strong opinions
- max-empty-lines - I would prefer to enforce 2,  but I can live if not possible to enforce
- number-leading-zero - I can live with it
- string-quotes - I feel strongly about this :)

I agree with @KristinLBradley, better if we discuss them during the next engineering sync.

alex-ju:
@KristinLBradley, @didoo [as discussed](https://docs.google.com/document/d/1twnk39hZ7p8oZYrld_FYsdiuUb6yn_weA46KnhF4nu4/edit#bookmark=id.6mimimbf50xf) during the engineering sync:
 - I replaced the deprecated `string-quotes` stylelint rule with enforcing double quotes in scss files via prettier ([92ff8a2](https://github.com/hashicorp/design-system/pull/1772/commits/92ff8a2865ad898b2d593901e6c832ae29b9eb2b))
 - I couldn't find a decent way to cover the `max-empty-lines` rule with prettier, so I dropped it alongside with the other two ([c18d70c](https://github.com/hashicorp/design-system/pull/1772/commits/c18d70ca8771a62cebaa8deff8db49b967b15d4f))

These changes will result in no deprecations flagged by stylelint and also no dependency warnings flagged by yarn.

PR #1773, opened 2023-11-03
rename our version command

Dhaulagiri:
### :pushpin: Summary

rename our version command so our release action can work again. (it's [broken](https://github.com/hashicorp/design-system/actions/runs/6747109720/job/18342500797#step:5:21))

`yarn version` is a yarn command, so our attempts to call `yarn version` from the release action are triggering that command, not the one we have specific in our package.json. This wasn't a problem in yarn 3.x as the [`version` command had to be installed as a plugin](https://v3.yarnpkg.com/cli/version). In yarn 4 it is [installed by default](https://yarnpkg.com/cli/version).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -32,7 +32,7 @@ jobs:
         uses: changesets/action@f13b1baaa620fde937751f5d2c3572b9da32af23 # v1.4.5
         with:
           version: yarn version-packages
-          publish: yarn release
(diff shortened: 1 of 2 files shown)


alex-ju (approved):
should we update our release command too? to `release-packages`, for consistency?

Dhaulagiri:
Can do. If nothing else, the symmetry is nice.

PR #1779, opened 2023-11-06
Add clarity around adding changelog entries for Figma libraries

jorytindall:
### :pushpin: Summary

This PR adds a bit more clarity in the wiki around adding changelog entries for Figma libraries.

***

### DIFF
--- a/wiki/Website-Changelog.md
+++ b/wiki/Website-Changelog.md
@@ -128,7 +128,8 @@ Entries made for the Figma Components and Foundations UI Kits require a few addi
 
 1. Copy and paste the appropriate template into the relevant changelog file in the components package; `packages/components/CHANGELOG-FIGMA-{COMPONENTS or FOUNDATIONS}` depending on what library the change was made in.
 2. If the change in Figma has corresponding changes in code or requires an update to the documentation, the changelog entry can be added in the same PR as the rest of the changes. The release process will automatically add these changes to the changelog on the website.
-3. Otherwise, if the change only pertains to Figma and does not result in a documentation change, open a new PR to add a changelog entry. Since this change does not correspond with a code release, the script used to generate the Figma changelog must be triggered manually from the webiste directory by running `yarn generate-changelog-markdown-files`.
+3. Otherwise, if the change only pertains to Figma and does not result in a documentation change, open a new PR to add a changelog entry. 
+    - Since this change does not correspond with a code release, the script used to generate the Figma changelog must be triggered manually from the webiste directory by running `yarn generate-changelog-markdown-files`.
 4. Replace all elements in brackets with the specific component or style/token name.
 5. For consistency purposes it's recommended to publish the changes in Figma at the same time as the changelog entry is merged and added to the website.
 
--- a/wiki/Website-Changelog.md
+++ b/wiki/Website-Changelog.md
@@ -10,7 +10,6 @@
 - [Templates](#templates)
   - [Templates for npm packages](#templates-for-npm-packages)
   - [Templates for Figma changelog entries](#templates-for-figma-changelog-entries)
-- [Adding Figma changelog entries](#adding-figma-changelog-entries)
 
 ---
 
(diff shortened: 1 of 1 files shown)


KristinLBradley (approved):

wiki/Website-Changelog.md
```diff
@@ -124,7 +124,11 @@ _{...important details}_
 
 ### Templates for Figma changelog entries
 
-Entries made for the Figma Components and Foundations UI Kits require a few additional steps. Copy and paste this template and adjust as necessary after publishing changes in Figma. _Replace all elements in brackets._
+Entries made for the Figma Components and Foundations UI Kits require a few additional steps.
+
+1. Open a PR to capture the relevant updates to the changelog.
+2. After publishing changes in Figma, copy and paste the appropriate template into the relevant changelog file in the components package; either `packages/components/CHANGELOG-FIGMA-{COMPONENTS or FOUNDATIONS}` depending on what library the change was made in.
+3. Replace all elements in brackets with the specific component or style/token name.
```
andgen404, line 131:
```suggestion
1. Open a PR to capture the relevant updates to the changelog.
2. After publishing changes in Figma, copy and paste the appropriate template into the relevant changelog file in the components package, either `packages/components/CHANGELOG-FIGMA-{COMPONENTS or FOUNDATIONS}` depending on what library the change was made in.
3. Replace all elements in brackets with the specific component or style/token name.
```

andgen404: Sorry, it slipped my mind, but as a designer would I branch a PR from the Main?

jorytindall: **Re: where to branch from:** I think what should be considered is if this should be rolled in with a component release process. I'm not sure it matters 100% where/when this happens in the process, as long as it happens, especially because there are circumstances where there aren't changes to the components package or documentation.

So, to summarize, if there are component updates happening and it can easily be added to either the component PR or the docs PR, great, otherwise it can be added after the fact since there isn't a version dependency issue like there is in the Ember component.

**Re: the other comment:** I'm not sure if a comma is the right punctuation here, I _think_ that a semicolon makes more sense.

didoo: if we draw a parallel with the developers, we update the "log" via the `changeset`, which is added directly in the PR for the component creation/update, then we can infer that the same _should_ happen (in theory) on the design side; in other words, the update should happen in the PR directly.

_In reality_ things are a bit more complex: there could be cases in which there are changes to Figma without PRs (this would require a new PR only for the update of the Figma changelog); plus `changeset` automatically gets the updates in order by the time they're merged; while for the Figma changelog files we could have multiple PRs updating the same file, and this could lead to conflicts. 

So I am not sure what is the best option. Maybe something to discuss during the team sync?

MelSumner (approved):
:shipit:

wiki/Website-Changelog.md
```diff
@@ -127,9 +127,10 @@ _{...important details}_
...
-2. If the change in Figma has corresponding changes in code or requires an update to the documentation, the changelog entry can be added in the same PR as the rest of the changes. Otherwise, if the change only pertains to Figma and does not result in a documentation change, open a new PR to add a changelog entry.
-3. Replace all elements in brackets with the specific component or style/token name.
-4. For consistency purposes it's recommended to publish the changes in Figma at the same time as the changelog entry is merged and added to the website.
+2. If the change in Figma has corresponding changes in code or requires an update to the documentation, the changelog entry can be added in the same PR as the rest of the changes. The release process will automatically add these changes to the changelog on the website.
+3. Otherwise, if the change only pertains to Figma and does not result in a documentation change, open a new PR to add a changelog entry. Since this change does not correspond with a code release, the script used to generate the Figma changelog must be triggered manually from the webiste directory by running `yarn generate-changelog-markdown-files`.
```
didoo, line 131:
```suggestion
3. Otherwise, if the change only pertains to Figma and does not result in a documentation change, open a new PR to add a changelog entry. 
    - Since this change does not correspond with a code release, the script used to generate the Figma changelog must be triggered manually from the webiste directory by running `yarn generate-changelog-markdown-files`.
```

didoo (approved):
👍

PR #1789, opened 2023-11-08
Website Search - Search UI implementation

didoo:
### :pushpin: Summary

This is a follow-up of the exploration done in https://github.com/hashicorp/design-system/pull/1766 where we have cherry picked the relevant commits (also cleaned/squashed them) and then built the production-ready functionality. 

In this PR I have:
- see [commits here](https://github.com/hashicorp/design-system/pull/1789/commits)

⚠️ Notice: to run the application you need to add some API keys (`ALGOLIA_***`) to a `.env` file inside the `website` folder. Speak with @didoo or @Dhaulagiri to get these keys).

### 👷 Testing

Use the deployed version to test the search functionality (if you want, you can test it also on mobile viewport/environments):

https://hds-website-git-website-search-search-ui-imple-e748ea-hashicorp.vercel.app/

If you don't know what to search, here are some suggestions: https://hashicorp.atlassian.net/browse/HDS-2847

### :link: External links

Jira tickets:
- https://hashicorp.atlassian.net/browse/HDS-2082 (**Epic**)

***

### 👀 Component checklist


website/ember-cli-build.js
```diff
@@ -7,47 +7,54 @@
...
+      advanced: false,
     },
-  });
+  },
+  // https://cli.emberjs.com/release/advanced-use/asset-compilation/#fingerprintingandcdnurls
+  fingerprint: {
+    // override defaults to also include json files which our markdown is compiled to. without this images don't render properly.
+    replaceExtensions: ['html', 'css', 'js', 'json'],
+    // in the algolia index we want to store the fingerprinted illustrations' paths only in "production"
+    exclude: isVercelProduction ? [] : ['assets/illustrations/**'],
```
Dhaulagiri, line 32: follow up question from this morning, but is there any benefit to fingerprinting these in preview apps (but not production or dev) or would it be cleaner to disable fingerprinting across the board until/unless we figure out some way of making this work in algolia (which _may_ be possible)

Dhaulagiri: > may be possible

thinking out loud here that this could work if:
* Assets are fingerprinted in production
* The index is re-built based on those assets so that the fingerprinted URL in algolia matches what is in the deployment

For non-production environments we could disable fingerprinting so these assets work consistently across local dev and preview apps.

website/app/components/doc/page/header/index.hbs
```diff
@@ -14,27 +14,40 @@
...
-    {{#if @isSidebarVisibleOnSmallViewport}}
-      <FlightIcon @name="x" />
-      Close
-    {{else}}
-      <FlightIcon @name="menu" />
-      Menu
-    {{/if}}
-  </button>
+  <div class="doc-page-header__mobile-only-menu">
+    <button
```
didoo, line 35: This is a "fake" button that triggers the opening of the autocomplete modal (it's the only way to do it, unfortunately)

website/app/components/doc/page/header/search/index.hbs
```diff
@@ -0,0 +1,6 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<div class="doc-algolia-search__autocomplete-container" {{did-insert this.didInsertSearchContainer}}></div>
```
didoo, line 6: This is just a container in which Algolia Autocomplete injects the "trigger" for the search modal (DOM is built by the library, and has very limited possible customization).

We add all our logic via the `didInsertSearchContainer` function invoked once the element is inserted in the DOM.

website/app/components/doc/page/header/search/index.js
```diff
@@ -0,0 +1,207 @@
...
+
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+
+import { autocomplete } from '@algolia/autocomplete-js';
+
+import { getItemsFunction } from './parts/getItemsFunction';
+import { templatesHeaderFunction } from './parts/templatesHeaderFunction';
+// import { templatesFooterFunction } from './parts/templatesFooterFunction';
```
didoo, line 14: leaving this for when we will add a "show-all" link at the end of a list of results

website/app/components/doc/page/header/search/index.js
```diff
@@ -0,0 +1,207 @@
...
+          autocompleteInstance.setQuery('');
+        }
+      },
+
+      //
+      // GET SOURCES
+      // see: https://www.algolia.com/doc/ui-libraries/autocomplete/api-reference/autocomplete-js/autocomplete/#param-getsources
+      getSources({ query }) {
+        if (query.trim() !== '') {
+          return [
```
didoo, line 50: the `getSources` is used to return three different "queries" to the API (called in a single HTTP request) with different parameters so we can have different types of results (general documentation - aka "content" - and tokens and icons)

website/app/components/doc/page/header/search/index.js
```diff
@@ -0,0 +1,207 @@
...
+              },
+              templates: {
+                header: templatesHeaderFunction({ group: 'tokens' }),
+                // footer: templatesFooterFunction,
+                item: templatesItemFunction(),
+              },
+            },
+          ];
+        } else {
+          return [
```
didoo, line 115: this is shown when the query is empty. we can customize what we want to show, in case

website/app/components/doc/page/header/search/index.js
```diff
@@ -0,0 +1,207 @@
...
+      'doc-algolia-search-autocomplete-secondary-trigger'
+    );
+    alternativeTrigger.addEventListener('click', () => {
+      console.log('THIS SHOULD OPEN THE SEARCH');
+      autocompleteInstance.setIsOpen(true);
+    });
+
+    // add event listener to open/close the modal via `command-K` hot-key
+    document.addEventListener('keydown', (event) => {
+      if (event.metaKey && event.key === 'k') {
```
didoo, line 201: I'll finalize the command to use here later (https://hashicorp.atlassian.net/browse/HDS-2879)

website/ember-cli-build.js
```diff
@@ -7,43 +7,50 @@
 
 const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 
-module.exports = function (defaults) {
```
didoo, line 10: made this change to allow having conditional logic for the `fingerprint.exclude` setting.

website/app/styles/pages/application/header.scss
```diff
@@ -150,15 +196,4 @@
   &:hover {
     color: var(--doc-color-white);
   }
-
```
didoo, line 153: simplified the breakpoints logic, it was used only for the homepage/header, now everything has standard media queries.

website/app/styles/doc-components/algolia-search/index.scss
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// ALGOLIA AUTOCOMPLETE SEARCH
+
+// "autocomplete" target container - not used
+// .doc-algolia-search__autocomplete-container {}
+
+// root container of the "autocomplete" injected DOM elements (only the button, in "detached" mode)
+// .aa-Autocomplete {}
```
didoo, line 12: notice: to make it easier to debug and to reference the original styles, I've kept the original naming convention: https://github.com/algolia/autocomplete/blob/next/packages/autocomplete-theme-classic/src/theme.scss

(it's impossible to completely use a custom naming convention, so better to avoid mix & match of two naming conventions)

website/app/components/doc/token-preview/index.hbs
```diff
@@ -0,0 +1,22 @@
+{{!
```
didoo, line 1: I've abstracted away the "Token Preview" component under the assumption that I could re-use it in the search results too. In reality is not possible, but since it likely will come handy in the future, I've left it here. Let me know if you prefer me to move it to a separate PR.

website/app/components/doc/page/header/search/parts/htmlTemplatesItemPreview.js
```diff
@@ -0,0 +1,68 @@
...
+        tokenName: item['token-name'],
+        tokenValue: item['token-value'],
+      });
+    default:
+      if (item.previewImage) {
+        return htmlPreviewIllustration({ html, imgSrc: item.previewImage });
+      } else {
+        return htmlPreviewIllustration({
+          html,
+          // TODO! create custom placeholder image?
```
didoo, line 63: @jorytindall let me know if you want to create a generic placeholder or just use this image/illustration

website/app/components/doc/page/header/search/parts/htmlTemplatesItemBody.js
```diff
@@ -0,0 +1,91 @@
...
+
+const htmlItemContentTitle = ({ html, title }) => html`
+  <div class="aa-ItemContentTitle">${title}</div>
+`;
+
+const htmlItemContentDescription = ({ html, description }) => html`
+  <div class="aa-ItemContentDescription">${description}</div>
+`;
+
+const snippetInitialText = (text) => {
```
didoo, line 16: we use this trick to return only the first 5 words of a "snippet part" (the non highlighted one) so that the highlighted content is always visible in the description field of the results in the modal panel.

website/app/components/doc/page/header/search/parts/htmlTemplatesItemPreview.js
jorytindall: I'll tackle this, while I think the one you've been using is probably ok, its a bit small and could be simplified a bit to "branded"

KristinLBradley:
I tested searching using the suggested terms [in this ticket](https://hashicorp.atlassian.net/browse/HDS-2847).

When I search for the word "color", it brings up a list of typography tokens that don't seem related to color. Is that expected?

Screenshot 2023-11-22 at 11 30 32 AM

website/app/components/doc/page/header/index.hbs
KristinLBradley: Would you perhaps want to use similar markup for the mobile only nav items as for the non-mobile ones? IE, use a `nav` vs. a `div` to contain the actions and mark them up as a list. (Or perhaps move them inside the existing `nav` element and then hide and show separate the separate actions for mobile vs. non-mobile?)

website/app/components/doc/page/header/search/index.js
KristinLBradley: What will the "show-all" link do? Take users to a page listing the search results?

KristinLBradley:
Starting to see broken images for component related links:

image

website/app/components/doc/page/header/search/index.js
```diff
@@ -0,0 +1,204 @@
...
+import { autocomplete } from '@algolia/autocomplete-js';
+
+import { getItemsFunction } from './parts/getItemsFunction';
+import { templatesHeaderFunction } from './parts/templatesHeaderFunction';
+// import { templatesFooterFunction } from './parts/templatesFooterFunction';
+import { templatesItemFunction } from './parts/templatesItemFunction';
+import { htmlPanelFooter } from './parts/htmlPanelFooter';
+import { htmlPanelNoResults } from './parts/htmlPanelNoResults';
+
+export default class DocAlgoliaSearchComponent extends Component {
```
KristinLBradley, line 19: Would we want to include the name of the tool used in the component name like this ("Algolia")?

website/app/components/doc/page/header/search/index.js
```diff
@@ -0,0 +1,204 @@
...
+              sourceId: 'content',
+              getItems: getItemsFunction({
+                searchQuery: query,
+                searchParams: {
+                  hitsPerPage: 5,
+                  attributesToHighlight: ['pageTitle', 'content'],
+                  attributesToSnippet: ['content:20'],
+                  distinct: true,
+                  // attributeForDistinct: 'pageBaseURL', // this doesn't work, for some reason; it needs to be set at Algolia dashboard level under "Deduplication and Grouping" configuration
+                },
```
KristinLBradley, line 61: In the documentation I found it says `attributeForDistinct` can be used in the `setSettings` method so maybe it's not supposed to work here? 

https://www.algolia.com/doc/api-reference/api-parameters/attributeForDistinct/

website/app/components/doc/page/header/search/parts/htmlPanelFooter.js
```diff
@@ -0,0 +1,48 @@
...
+      <span>to select</span>
+    </div>
+    <div
+      class="aa-PanelFooter__keyboard-hint"
+      aria-label="Press Escape to close the search"
+    >
+      <kbd class="esc">Esc</kbd>
+      <span>to close</span>
+    </div>
+  </div>
```
KristinLBradley, line 32: Personally, I would mark this up as a list of hints (`ul`).

website/app/components/doc/page/header/search/parts/htmlPanelFooter.js
```diff
@@ -0,0 +1,48 @@
...
+      class="aa-PanelFooter__algolia-link"
+    >
+      <span class="aa-PanelFooter__algolia-text">Search by</span>
+      <img
+        class="aa-PanelFooter__algolia-logo"
+        src="/assets/logos/algolia.svg"
+        alt="Algolia logo"
+      />
+    </a>
+  </div>
```
KristinLBradley, line 47: And I would mark this up a a `p` vs. just a `div`.

website/app/components/doc/page/header/search/parts/htmlPanelFooter.js
```diff
@@ -0,0 +1,48 @@
...
+      href="https://www.algolia.com/"
+      rel="noopener noreferrer"
+      target="_blank"
+      class="aa-PanelFooter__algolia-link"
+    >
+      <span class="aa-PanelFooter__algolia-text">Search by</span>
+      <img
+        class="aa-PanelFooter__algolia-logo"
+        src="/assets/logos/algolia.svg"
+        alt="Algolia logo"
```
KristinLBradley, line 44: The full phrase that sighted users get from this is, "Search by algolia" so the alt text should say "Algolia" vs. telling users that it's a logo which makes the phrase be "Search by Algolia logo" which doesn't make as much sense.

website/app/components/doc/page/header/search/parts/htmlTemplatesItemBody.js
```diff
@@ -0,0 +1,91 @@
...
+  var splitted = text.split(' ');
+  var reduced = splitted.slice(-5);
```
KristinLBradley, line 18: Is there a reason to use `var` in here vs. `const`?

website/app/components/doc/page/header/search/parts/htmlTemplatesItemBody.js
KristinLBradley: Doesn't it return the last five words vs. first five?

jorytindall:
High level note about keyboard navigation: when the modal is opened, focus is automatically set on the input, however, if I tab through the entire list of results focus proceeds to the Algolia logo, then back to the main page. This seems like a pretty significant problem, is there any control over this? Attached a screen recording.

https://github.com/hashicorp/design-system/assets/2200899/ef03d58c-6fca-4bf4-b5bf-ff6417f237fc

On an additional note about keyboard navigation, I know the up/down arrow key navigation was removed, but I can't quite remember why. I tabbed through the results which was fine, but arrow keys seem a bit more intuitive.

website/app/components/doc/page/header/index.hbs
alex-ju: In relation to the automated a11y test failing on the `.aa-Autocomplete`. I'm wondering why don't we use this button (currently only for mobile) for the desktop version too. The markup is compliant and this way we avoid having to set the axe exception across our tests.

alex-ju (approved):
Looks good for a first iteration. I've not checked design parity (don't know where to look). Outstanding for me are the missing preview images (I know you're aware of this) and the fact that the focus is not trapped within the modal (as Jory mentioned above). However, there's a lot of code in here (1551 lines atm 🙀 ) so I'm approving it in case you want to get it in and iterate further in separate PRs

MelSumner:
Also, checking the page with the axe browser extension: 
CleanShot 2023-11-27 at 10 19 19

didoo:
> Here are some issues I've found while reviewing with a screen reader:
> 
> 1. Focus is not trapped inside of the search modal.

@MelSumner this is a known issue of the library we're using https://github.com/algolia/autocomplete/issues/588. According to their comment, it's in their plan to implement it, but I don't know if/when this will really happen (and I don't think is possible to add it ourselves).

> 2. there are no instructions read to the user so I assume that I use it the way I use a normal form. I type a search term and press enter, and I expect to be taken to the list of search results. However, in pressing ENTER/RETURN, the window closes entirely and I have no idea why.

It's how it's implemented in the Algolia autocomplete library. I've pushed a hacky fix for this: https://github.com/hashicorp/design-system/pull/1789/commits/5aebac28c7a84c11889508c9876417fb9a87bd45 not super elegant and it flashes for a few milliseconds, but it's the best we can do.

> 3. Check the names/labels that are put on the items for redundancy. if the item has a role of search, then it doesn't need "search" in the accessible name.

Can you be more specific? The only element with a `role="search"` is the `<form>` element, and I don't see elements with `Search` as accessible name:

screenshot_3297

> Also, checking the page with the axe browser extension: CleanShot 2023-11-27 at 10 19 19

Yes, that's the reason why I opened https://github.com/hashicorp/design-system/pull/1823. Unfortunately there is no way to fix it, the library doesn't provide overrides or hooks that we can use. In theory we could try to open an upstream PR, but I am not sure it would be easy (it's a very complex library, with a lot of other libraries depending on it).

website/app/styles/doc-components/algolia-search/index.scss
KristinLBradley: Would we perhaps want to scope the class names as a precaution still? (IE: Add a unique "hds-blah" class name to the outer container then nest other class name references inside so they don't leak.)

Just in the off chance this was ever reused elsewhere such as in an HDS component or even just to establish a convention which we stick to for this kind of thing (similarly to how I scoped the prism-dracula theme styles used by doc-code-block so they wouldn't leak and effect the HDS CodeBlock).

website/app/styles/doc-components/algolia-search/modal.scss
```diff
@@ -0,0 +1,250 @@
...
+  flex-direction: column;
+  width: 100%;
+  max-height: 100vh;
+  margin: 0;
+  overflow: hidden;
+  background: var(--doc-color-white);
+
+  @include breakpoint.with-fixed-sidebar () {
+    position: absolute;
+    top: 90px;
```
KristinLBradley, line 50: If this `top: 90px;` value is based partly on the height of the doc-page-header would it make sense to use a variable or else add a comment explaining?

website/app/styles/doc-components/algolia-search/modal.scss
```diff
@@ -0,0 +1,250 @@
...
+  align-items: center;
+  justify-content: space-between;
+  margin: 0;
+  padding: 16px;
+  background-color: #fff;
+  box-shadow: inset 0 1px 0 rgba(#807ea3, 30%);
+
+  &::after {
+    position: absolute;
+    top: -16px;
```
KristinLBradley, line 143: I prefer to use `bottom: 100%;` for positioning items like this. That way if the height were ever adjusted for some reason, you don't have to remember to change a value in 2 places.

website/app/styles/doc-components/algolia-search/modal.scss
```diff
@@ -0,0 +1,250 @@
...
+  kbd {
+    @include doc-font-family("mono");
+    display: inline-block;
+    min-width: 18px;
+    min-height: 18px;
+    padding: 1px 4px;
+    color: var(--doc-color-gray-200);
+    font-weight: 500;
+    font-size: 13px;
+    line-height: 14px;
```
KristinLBradley, line 181: If you change this to `line-height: 1;` and add: `display: flex; align-items: center;` then you don't need a fixed value like this for the line-height. (Also if the height of the button and/or text size is ever adjusted then it will still be vertically centered.)

website/app/styles/doc-components/algolia-search/modal.scss
```diff
@@ -0,0 +1,250 @@
...
+    // &.arrow-down,
+    // &.arrow-up {
+    //   // custom font to have better representation for the "↓/↑" keys
+    //   font-family: "Lucida Console", "Lucida Sans Typewriter", monospace;
+    // }
+
+    &.tab,
+    &.esc {
+      font-size: 11px;
+      letter-spacing: -0.6px;
```
KristinLBradley, line 201: It you use ems for letter-spacing values then it adjusts if the font-size is changed for any reason since it's relative to the font size instead of being a fixed value. (Try changing to: `letter-spacing: -.05em;`)

website/app/styles/doc-components/algolia-search/trigger.scss
```diff
@@ -0,0 +1,60 @@
...
+.aa-DetachedSearchButton {
+  @include doc-font-style-navigation();
+  display: flex;
+  gap: 4px;
+  align-items: center;
+  height: 36px;
+  margin: 0;
+  padding: 4px 12px;
+  color: var(--doc-color-gray-400);
+  line-height: 16px; // override
```
KristinLBradley, line 19: I didn't see any effect when I commented out this line-height value in the browser

KristinLBradley (commented):
Looking great overall! Went through and tested functionality & looked through code in detail. Left some comments & questions.

website/app/components/doc/page/header/index.hbs
didoo: @alex-ju in any case we need to have the `.aa-Autocomplete` injected, for the other "trigger" buttons to work. In that case, should the `.aa-Automplete` simply have `display: none` ?

website/app/components/doc/page/header/search/index.js
didoo: I don't know yet, likely will increase the limit from 5 results to infinite for that specific type of content. 
In any case, it will be done in https://hashicorp.atlassian.net/browse/HDS-2888

website/app/components/doc/page/header/search/parts/htmlTemplatesItemBody.js
didoo: the last 5 words of the _first_ snippet part (it's broken in multiple parts)

didoo:
> When I search for the word "color", it brings up a list of typography tokens that don't seem related to color. Is that expected?

@KristinLBradley It's because the tokens for the font families contains the text `Apple *Color* Emoji` in their values. Something I need to discuss with @jorytindall 

> Would it be possible to display a message saying "no exact matches for 'blah' " or similar followed by a list of similar word matches found?

I think it may be possible, but not out of the box (it would require some extra custom logic).

website/app/components/doc/page/header/search/index.js
didoo: In this case yes, it's very specific and it would be a super minor thing to change it, compared with the effort to use another tool. It's to avoid confusion with a very generic term like "search" (that in some cases is used to mean "filter", hence https://github.com/hashicorp/design-system/pull/1818/files)

website/app/components/doc/page/header/search/index.js
didoo: @KristinLBradley We're using the `getAlgoliaResults` function (docs [here](https://www.algolia.com/doc/ui-libraries/autocomplete/api-reference/autocomplete-js/getAlgoliaResults/)) which points to the [Search API Parameters](https://www.algolia.com/doc/api-reference/search-api-parameters/] which exposes the `[distinct](https://www.algolia.com/doc/api-reference/api-parameters/distinct/)` parameter, but strangely enough not the `attributeForDistinct` (strangely because they both appear together in the Algolia dashboard settings)
screenshot_3298

didoo:
> High level note about keyboard navigation: when the modal is opened, focus is automatically set on the input, however, if I tab through the entire list of results focus proceeds to the Algolia logo, then back to the main page. This seems like a pretty significant problem, is there any control over this? Attached a screen recording.

@jorytindall Unfortunately no. There is an open issue here https://github.com/algolia/autocomplete/issues/588 but is not being picked up apparently.

> On an additional note about keyboard navigation, I know the up/down arrow key navigation was removed, but I can't quite remember why. I tabbed through the results which was fine, but arrow keys seem a bit more intuitive.

It's not been removed. It works only when the focus is on the `input` element, but not when an element is focused. Same as above, I have no control on this (that's why I decided to hide the tip in the modal's footer).

website/app/styles/doc-components/algolia-search/modal.scss
didoo: It's kinda base, but not exactly. It's a number that "looks right" (visually, and numerically).

website/app/styles/doc-components/algolia-search/modal.scss
didoo: You're absolutely right. Fixed! 🙏

website/app/styles/doc-components/algolia-search/trigger.scss
didoo: Removed.

didoo:
@alex-ju @jorytindall @KristinLBradley @MelSumner I've addressed most of the comments/suggestions and for the ones that I couldn't address I've explained why (see comments above).

At this point I think the question to answer is: is it good enough to go live in production as "beta"? if the answer is "yes" then probably is ok to approve it. In any case we have a [list of follow-up tasks](https://hashicorp.atlassian.net/browse/HDS-2862) that will be worked on later anyway.

You can test the search functionality here: https://hds-website-git-website-search-search-ui-imple-e748ea-hashicorp.vercel.app/

/cc @Dhaulagiri

website/app/components/doc/page/header/index.hbs
```diff
@@ -14,27 +14,52 @@
       <Doc::Page::Header::NavItem @label="Foundations" @route="foundations" @currentTopRoute={{@currentTopRoute}} />
       <Doc::Page::Header::NavItem @label="Components" @route="components" @currentTopRoute={{@currentTopRoute}} />
       <Doc::Page::Header::NavItem @label="Patterns" @route="patterns" @currentTopRoute={{@currentTopRoute}} />
+      <li class="doc-page-header__nav-item-generic doc-page-header__nav-item--split">
+        <div class="sr-only" id="search-button">This button opens a dialog containing an input field. Entering text in the input will perform -automatically as you type- a search within the website.</div>
```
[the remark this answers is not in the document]
didoo: @alex-ju good point. done in https://github.com/hashicorp/design-system/pull/1848

MelSumner, line 18: Suggestion: 
```suggestion
        <div class="sr-only" id="search-button">This button opens a dialog containing an input field and some additional information that you may wish to explore. An automatic search will be performed as you type text into the search field, but the results may not be announced. Exploring the additional items in the modal will help you discover the search results.</div>
```

website/app/components/doc/page/header/index.hbs
```diff
@@ -14,27 +14,52 @@
...
+        <button
+          type="button"
+          class="doc-page-header__search-button"
+          data-doc-algolia-search-autocomplete-secondary-trigger
+          aria-describedby="search-button"
```
MelSumner, line 23: 👍

didoo:
> Not referencing the additions to the UI we discussed today

@jorytindall I've pushed the change we discussed (the "See all..." link at the bottom of the results lists for icons and tokens). Let me know what you think and if you would do some changes (font size or weight? color? add an icon? other?)

KristinLBradley (approved):

MelSumner (approved):
Let's ship it.

PR #1794, opened 2023-11-09
Clarify `isMinimized` docs

alex-ju:
### :pushpin: Summary

Expand `isMinimized` docs. Although we state that this argument controls if the SideNav is rendered collapsed or expanded **when initialized**, it might be tempting for the consumer to try and control its state after render. So we're expanding the docs explaining that this argument is altered by the component logic based on user interactions and it is not a suitable way of controlling the SideNav stated from outside after render.

### DIFF
--- a/website/docs/components/side-nav/partials/code/component-api.md
+++ b/website/docs/components/side-nav/partials/code/component-api.md
@@ -25,7 +25,7 @@ This is the full-fledged component (responsive and animated).
     <em>Notice: if `@isResponsive` is set to false, this argument has no effect.</em>
   </C.Property>
   <C.Property @name="isMinimized" @type="boolean" @default="false">
-    Controls if the SideNav is rendered collapsed or expanded when initialized. This allows an application to preserve the collapsed/expanded state across sessions.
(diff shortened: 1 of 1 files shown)


website/docs/components/side-nav/partials/code/component-api.md
```diff
@@ -25,7 +25,7 @@ This is the full-fledged component (responsive and animated).
     <em>Notice: if `@isResponsive` is set to false, this argument has no effect.</em>
   </C.Property>
   <C.Property @name="isMinimized" @type="boolean" @default="false">
-    Controls if the SideNav is rendered collapsed or expanded when initialized. This allows an application to preserve the collapsed/expanded state across sessions.
+    Controls if the SideNav is rendered collapsed or expanded when initialized. This allows an application to preserve the collapsed/expanded state across sessions. After the initial render this argument is altered based on user interactions (collapse/expand the SideNav or resize the window).
```
didoo, line 28:
```suggestion
    Controls if the SideNav is rendered collapsed or expanded when initialized. This allows an application to preserve the collapsed/expanded state across sessions. After the initial render this argument is altered based on user interactions (collapse/expand the SideNav or resize the window) and it is not a suitable way of controlling the SideNav state from outside after render (it's an internal state).
```

didoo (approved):
👍

PR #1799, opened 2023-11-10
Add Character Count documentation

jorytindall:
### :pushpin: Summary

This PR adds documentation for the character count related to form fields. A couple of important notes:

We've decided to colocate the guidelines with the rest of the `Form::Primitives` because it's treated the same way as other primitives in both code and Figma. The downside to this is that the contextual/sub-component is only available in text-based inputs (which we noted in the docs).

The alternative to this was to add a line item to the content section of the `TextInput`, `Textarea`, and `MaskedInput` docs to capture these guidelines, but the former solution seemed better. If necessary, we can link to the primitive docs from the respective content sections.

**To discuss:** whether the code doc updates happen on this branch or on #1802 and whether to point this branch to the component branch.

### :link: External links

Jira ticket: [HDS-2823](https://hashicorp.atlassian.net/browse/HDS-2823)
Figma file: [Proposed solution](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/VlL6XmEhttEXl1G2UcnP7E/HDS-Product---Components?type=design&node-id=49360%3A76493&mode=design&t=dh3DGKjBjWiBOynz-1)

***

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-2823]: https://hashicorp.atlassian.net/browse/HDS-2823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1802, opened 2023-11-13
Add `Form::CharacterCount`

alex-ju:
### :pushpin: Summary

Add `Form::CharacterCount` as a form primitive and expose it as a contextual component to `Form::Field`, and further to `Form::Field::MaskedInput`, `Form::Field::TexInput`, and `Form::Field::Texarea`.

 - Add `Form::CharacterCount` as a form primitive/base element:
   - it takes as an argument the maximum number of characters allowed (`maxLength`)
   - it takes as an argument the minimum number of characters allowed (`minLength`)
   - on input change, it calculates:
     - the number of characters entered in the associated input (exposed as `currentLength`)
     - the number of characters allowed (exposed as `remaining)`
     - the number of characters still required (exposed as `shortfall`)
   - it can render:
     - a predefined format message (default)
     - a custom format using the yielded block that can use any of the exposed values (`maxLength`, `currentLength`, `remaining`, `shortfall`)
 - Add `CharacterCount` as a contextual component to `Form::Field`, and further to `Form::Field::MaskedInput`, `Form::Field::TexInput`, and `Form::Field::Texarea`

Quick preview links:
 - [Form / Base elements](https://hds-showcase-git-alex-ju-character-count-hashicorp.vercel.app/components/form/base-elements)
 - [Form::Field::MaskedInput](https://hds-showcase-git-alex-ju-character-count-hashicorp.vercel.app/components/form/masked-input)
 - [Form::Field::TextInput](https://hds-showcase-git-alex-ju-character-count-hashicorp.vercel.app/components/form/text-input)
 - [Form::Field::Textarea](https://hds-showcase-git-alex-ju-character-count-hashicorp.vercel.app/components/form/textarea)

### :link: External links

Jira ticket: [HDS-2757](https://hashicorp.atlassian.net/browse/HDS-2757)
Figma file (original implementation): https://www.figma.com/file/VlL6XmEhttEXl1G2UcnP7E/Add-Character-Count?type=design&node-id=49360-76493&mode=dev
Figma file (updated message logic): https://www.figma.com/file/NLhvZo1948lFfga6ZKYORT/CharacterCount-Cleanup?type=design&node-id=49988-2428&mode=dev

***

### 👀 Component checklist

[HDS-2757]: https://hashicorp.atlassian.net/browse/HDS-2757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,102 @@
...
+
+  /**
+   * @param remainingLength
+   * @type {number}
+   * @default null
+   * @description The remaining number of characters.
+   */
+  get remainingLength() {
+    return this.maxLength - this.currentLength;
+  }
```
alex-ju, line 36: In theory, this can be calculated by the consumer, even in the template by using `ember-math-helpers`, for example `{{sub CC.maxLength CC.currentLegth}}`. However, this helper doesn't seem to be as popular within our consumers' codebase, so instead of forcing them to introduce a new dependency I'd rather do this calculation within the component.

jorytindall: Great idea! The remaining count is probably one of the more common use-cases here.

packages/components/app/styles/components/form/character-count.scss
jorytindall, line 1: Here's a question: the way that I had structure the Figma component was with three different text layers for these reasons:

- So that I could lock down the `/` so that it couldn't be changed (bit of a Figma nicety)
- So that I could expose text-replacements (a Figma nicety) for the individual layers `currentLength` and `maxLength`
- More importantly: To add a little bit of a gap (2px specifically) between each layer. Visually, when there isn't a gap the elements run together. The gap gives a bit of breathing room.

Screenshot 2023-11-13 at 11 13 15 AM

I know this would involve some complex wrapping and nesting of `spans` or similar elements so I won't die on this hill, but if there's an easy way to do it in code, great! I did some testing and simply adding a character space in the template string adds _too much_ space between each element.

Otherwise, I'll update the Figma component to match.

packages/components/addon/components/hds/form/character-count/index.hbs
```diff
@@ -0,0 +1,19 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<Hds::Text::Body
+  @tag="div"
```
jorytindall, line 6: [Question] Should this be a more semantic tag?

packages/components/tests/integration/components/hds/form/text-input/field-test.js
```diff
@@ -98,25 +98,29 @@ module('Integration | Component | hds/form/text-input/field', function (hooks) {
...
     assert.dom('.hds-form-field__helper-text').doesNotExist();
+    assert.dom('.hds-form-field__character-count').doesNotExist();
     assert.dom('.hds-form-field__error').doesNotExist();
   });
   test('it automatically provides all the ID relations between the elements', async function (assert) {
     await render(
       hbs`<Hds::Form::TextInput::Field @extraAriaDescribedBy="extra" as |F|>
           <F.Label>This is the label</F.Label>
           <F.HelperText>This is the helper text</F.HelperText>
+          <F.CharacterCount>20/40</F.CharacterCount>
```
didoo, line 123: here we're testing the "static" version of the character count (which is already tested in the "field" component). Should we test also (or only) the "dynamic" version, to make sure it works as expected, in relation to the `input` value/lenght? (or we test it only in the `field` component, as suggested above)

packages/components/tests/integration/components/hds/form/textarea/field-test.js
```diff
@@ -63,25 +63,29 @@ module('Integration | Component | hds/form/textarea/field', function (hooks) {
...
     assert.dom('.hds-form-field__helper-text').doesNotExist();
+    assert.dom('.hds-form-field__character-count').doesNotExist();
     assert.dom('.hds-form-field__error').doesNotExist();
   });
   test('it automatically provides all the ID relations between the elements', async function (assert) {
     await render(
       hbs`<Hds::Form::Textarea::Field @extraAriaDescribedBy="extra" as |F|>
           <F.Label>This is the label</F.Label>
           <F.HelperText>This is the helper text</F.HelperText>
+          <F.CharacterCount>20/40</F.CharacterCount>
```
didoo, line 88: same as above, should we test the "dynamic" variant too?

packages/components/tests/integration/components/hds/form/field/index-test.js
```diff
@@ -59,6 +61,7 @@ module('Integration | Component | hds/form/field/index', function (hooks) {
           <F.Label>This is the label</F.Label>
           <F.HelperText>This is the helper text</F.HelperText>
           <F.Control><pre id={{F.id}} aria-describedby={{F.ariaDescribedBy}}>This is a mock control</pre></F.Control>
+          <F.CharacterCount>20/40</F.CharacterCount>
```
didoo, line 64: should we include a test for the `dynamic` version? essentially, testing that these two lines do what we expect them to do: https://github.com/hashicorp/design-system/blob/6d1df7431e26bbc58d42e1473678cd19c5f17d76/packages/components/addon/components/hds/form/field/index.hbs#L34-L35

packages/components/tests/integration/components/hds/form/character-count/index-test.js
```diff
@@ -0,0 +1,63 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module(
+  'Integration | Component | hds/form/character-count/index',
```
didoo, line 12: at times we test also the `aria` attributes here; should we test that the `aria-live="polite"` is present in the component?

packages/components/app/styles/components/form/character-count.scss
```diff
@@ -0,0 +1,13 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+//
+// FORM > CHARACTER-COUNT
+//
+
+.hds-form-character-count {
+  display: block;
```
didoo, line 11: do we need this? 🤔
```suggestion
```

packages/components/app/styles/components/form/character-count.scss
didoo: My 2 cents: I think we can keep the implementation in Figma as is, but keep the (text-only) implementation in code as is, for sake of simplicity.

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
```diff
@@ -141,6 +141,36 @@
 
   <Shw::Divider />
 
+  <Shw::Text::H2>Character count</Shw::Text::H2>
+
+  <Shw::Flex @direction="column" as |SF|>
+    <SF.Item @label="With default content">
+      <input type="hidden" value="with default content" id="character-count-input-1" />
```
didoo, line 148: Nice trick! ;)

packages/components/addon/components/hds/form/text-input/field.hbs
```diff
@@ -12,7 +12,7 @@
 >
   {{! Notice: the order of the elements is not relevant here, because is controlled at "Hds::Form::Field" component level }}
   {{yield (hash Label=F.Label isRequired=F.isRequired isOptional=F.isOptional)}}
-  {{yield (hash HelperText=F.HelperText Error=F.Error)}}
+  {{yield (hash HelperText=F.HelperText Error=F.Error CharacterCount=F.CharacterCount)}}
```
didoo, line 15: So easy and neat! 🤩

packages/components/addon/components/hds/form/character-count/index.js
didoo: I agree with your approach, makes sense 👍

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
```diff
@@ -222,8 +222,12 @@
         <F.HelperText>This is the helper text</F.HelperText>
       </Hds::Form::TextInput::Field>
     </SG.Item>
-    {{! left empty on purpose }}
-    <SG.Item />
+    <SG.Item @label="Label + Character count">
```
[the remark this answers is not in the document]
didoo: I would also include an error message, so we can (visually) test the spacing between the counter and the following error element.

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,102 @@
...
+   * @default () => {}
+   */
+  get onInsert() {
+    let { onInsert } = this.args;
+
+    let inputControl = document.getElementById(this.args.controlId);
+    if (inputControl && inputControl.value) {
+      this.updateMessage(inputControl);
+      inputControl.addEventListener('input', this.onInput);
+    }
```
didoo, line 63: should we warn the consumer if they point to an element that doesn't exist? https://github.com/hashicorp/design-system/blob/main/packages/components/addon/modifiers/hds-clipboard.js#L35-L41

(counterpoint: do we expect them to use it as a standalone component? probably not, so maybe it's simply not useful in this case?)

(counter-counterpoint: they may want to use it in a generic "field" with custom input implementation?)

packages/components/addon/components/hds/form/character-count/index.hbs
alex-ju: Fair question – I was replicating what we have for the HelperText and Error, but it could be a `<p>` tag just as well.

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,102 @@
...
+   * @type {function}
+   * @default () => {}
+   */
+  get onInsert() {
+    let { onInsert } = this.args;
+
+    let inputControl = document.getElementById(this.args.controlId);
+    if (inputControl && inputControl.value) {
+      this.updateMessage(inputControl);
+      inputControl.addEventListener('input', this.onInput);
```
didoo, line 62: we never remove this event listener. can you think of a way of doing it?

packages/components/app/styles/components/form/character-count.scss
alex-ju: Thanks for highlighting this, Jory! I definitely missed it. I believe I can achieve this design (for the pre-defined format) without requiring additional nested elements. I can also use tabular numbers to avoid movement at every ke .stroke

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,102 @@
...
+
+    // notice: this is a guard used to prevent triggering an error when the component is used as standalone element
+    if (typeof onInsert === 'function') {
+      return onInsert;
+    } else {
+      return NOOP;
+    }
+  }
+
+  @action onInput(event) {
```
didoo, line 73:
```suggestion
  @action
  onInput(event) {
```

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,102 @@
...
+    let inputControl = event.target;
+    this.updateMessage(inputControl);
+  }
+
+  @action updateMessage(inputControl) {
```
didoo, line 78:
```suggestion
  @action
  updateMessage(inputControl) {
```

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,102 @@
...
+    if (inputControl) {
+      this.currentLength = inputControl.value.length;
+      this.message = `${this.currentLength}/${this.maxLength}`;
```
didoo, line 81: I don't think you need to have the `message` declared here, you can use directly `{{this.currentLength}}/{{this.maxLength}}` in the HBS template

in which case, the function can be called `updateCurrentLength`

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,102 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+export const ID_PREFIX = 'character-count-';
```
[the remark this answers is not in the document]
didoo: (also, as far as I know it's not used anywhere else (so non need to export it?) 🤔

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,102 @@
...
+    }
+  }
+
+  @action onInput(event) {
+    let inputControl = event.target;
+    this.updateMessage(inputControl);
+  }
+
+  @action updateMessage(inputControl) {
+    if (inputControl) {
```
didoo, line 79: this check is not necessary, it's already done in the `onInsert` function

or (even better?) move both the logical checks `if (inputControl && inputControl.value)` here, instead of the `onInsert` function 🤔

packages/components/addon/components/hds/form/character-count/index.hbs
```diff
@@ -0,0 +1,19 @@
...
+  @size="100"
+  class={{this.classNames}}
+  id={{this.id}}
+  {{did-insert this.onInsert}}
+  aria-live="polite"
+  ...attributes
+>
+  {{#if (has-block)}}
+    {{yield (hash maxLength=this.maxLength currentLength=this.currentLength remainingLength=this.remainingLength)}}
```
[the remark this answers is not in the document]
alex-ju: it's tempting, but for consistency that would mean the main argument we expose would change from `maxLength` to `max` which would require alignment with the recently merged Figma branch as well (and in Figma having just `max` may not be as clear, as I think it's a field-level property, not a CharacterCount one). @jorytindall @andgen404 what are your thoughts on this?

packages/components/addon/components/hds/form/character-count/index.js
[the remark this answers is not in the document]
alex-ju: yeah, we don't use them anymore – copied over from the HelperText and Error. I removed it here and will open a follow-up PR for the HelperText and Error.

packages/components/tests/dummy/app/templates/components/form/textarea.hbs
```diff
@@ -138,8 +138,12 @@
         <F.HelperText>This is the helper text</F.HelperText>
       </Hds::Form::Textarea::Field>
     </SG.Item>
-    {{! left empty on purpose }}
-    <SG.Item />
+    <SG.Item @label="Label + Character count">
```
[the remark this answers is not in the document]
alex-ju: reverted this change and added the character count example in a new section below this one 41ae938b1d8da1ada752c421c35beae526f0a8ea

packages/components/tests/integration/components/hds/form/text-input/field-test.js
alex-ju: indeed, using the 'dynamic' version makes more sense for the Field components – updated in 9127638

packages/components/tests/integration/components/hds/form/textarea/field-test.js
alex-ju: updated in 9127638

packages/components/tests/integration/components/hds/form/field/index-test.js
alex-ju: for the `Form::Field` all tests are using a mock-control (not a real `<input>`) so we can't have a dynamic version there. the two properties highlighted above (the `controlId` and appending descriptors) are tested in https://github.com/hashicorp/design-system/pull/1802/files#diff-7293ac7c93f148221709265bfb8098aefc04d3e239ba7f8c8fd88d6a54561490R81-R85

packages/components/tests/integration/components/hds/form/character-count/index-test.js
alex-ju: good call – added in 9450771ad1db25abbd4c094edf45374d41492c62

packages/components/app/styles/components/form/character-count.scss
alex-ju: not required anymore (if we're going forward with a `<div>` elements, as it currently is) – removed in a7ed41d

packages/components/addon/components/hds/form/character-count/index.js
alex-ju: lacking a valid `id` would only happen in a custom implementation where `@controlId` is provided by the consumer – and this is also the most likely scenario where a mistake can happen – so I added a similar warning, as suggested, in https://github.com/hashicorp/design-system/pull/1802/commits/b3cfdcbc184e3f97a6479ea8149ef3024acc0535#diff-369ec02c39a6239663ad5a4f7041563cc7c0c4492470465eddc134c12b0a6cd6R61-R64

packages/components/addon/components/hds/form/character-count/index.js
alex-ju: The removal of this event listener is handled by the garbage collector in most cases (except for when the CharacterCount is removed by the associated input remains on the page). Happy to add the logic and deal with this  https://github.com/hashicorp/design-system/pull/1802/commits/8523c3f27e700455d542dfeb976044d32f37441d

packages/components/addon/components/hds/form/character-count/index.js
alex-ju: addressed in b3cfdcbc184e3f97a6479ea8149ef3024acc0535 during the refactor. as the code was written initially we needed it in both places (in `onInsert` to check if we have an input before trying to attach the event listener and in `onInput`/`updateMessage` to check if a target existed.

packages/components/app/styles/components/form/character-count.scss
alex-ju: Updated to something that resembles the design – if you'd like to have another look Jory and see if that's acceptable from your point of view

alex-ju:
> General question/observation: I see we're considering only the `maxLength` and not the `minLength`. Any specific reason for this choice? How much complexity would it add to introduce also this parameter?

@didoo since the design changed, and we use the character count as guidance not as validation, we mainly count how many characters have been inserted into a field (out of a maximum, e.g. `current`/`max`). it didn't make sense to me to introduce a `minLength` with the counter (what would the default look like? `min`/`current`/`max` – not easy to get the grips with). if the consumer has a `minLength` they can use it in the custom message and/or in the validation error, but we wouldn't benefit from ingesting this value within the `CharacterCount` as there were no pertinent values we could determine, and return to the end user.

packages/components/addon/components/hds/form/character-count/index.hbs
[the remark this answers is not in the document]
didoo: OK, makes sense to keep it in sync with Figma. I'm resolving this.

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,115 @@
...
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+
+const ID_PREFIX = 'character-count-';
+const NOOP = () => {};
+
+export default class HdsFormCharacterCountIndexComponent extends Component {
+  // The current number of characters in the associated input
+  @tracked currentLength;
+  inputControl = document.getElementById(this.args.controlId);
```
didoo, line 16: do you see any case in which the input element is added to the DOM _after_ the counter is injected? 🤔

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,115 @@
...
+      return onInsert;
+    } else {
+      return NOOP;
+    }
+  }
+
+  @action
+  willDestroyNode() {
+    if (this.inputControl) {
+      this.inputControl.removeEventListener(
```
didoo, line 85: technical question: this gets invoked when the `counter` is removed, not when the `input` is, but the event is attached to the `input` so _in theory_ there could be cases in which the input is already removed from the DOM. in that case, would the event listener be removed automatically?

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,115 @@
...
+        'input',
+        this.updateCurrentLength,
+        true
+      );
+    }
+  }
+
+  @action
+  updateCurrentLength() {
```
didoo, line 94: do you remember we talked about having a callback function\* that the consumers can hook into? do you think it still makes sense to have it? otherwise how consumers would know when to trigger/show the error?

\* (with the function returning all the relevant values: `currentLenght`,`maxLength`, `remainingLength`)

didoo (commented):
added only minor comments. 

there are still two open questions/points from my point of view:

- having also a `min` value/argument (was discussed? what was the decision?)
- having a callback function that the consumers can hook into, to do their own validation/error triggering
    - this has led me to think that maybe we should have an extra use case added to the showcase, where we connect the error with the callback function, to have the full circle of functionality tested as we were consumers, to make sure we're not missing something (potentially, all the "counter" demos for `text-input/textarea` could be hooked to a validation, if that's not too complex, so we can test different scenarios)

didoo:
> @didoo since the design changed, and we use the character count as guidance not as validation, we mainly count how many characters have been inserted into a field 

> (out of a maximum, e.g. `current`/`max`). 

yes, because we chose to have the `max`

>  it didn't make sense to me to introduce a `minLength` with the counter (what would the default look like? `min`/`current`/`max` – not easy to get the grips with). 

it could be: 
- `${current} / ${max}` if only `max` is provided
- `${current} (min. ${min})` if only `min` is provided
- `${current} (min. ${min} / max ${max})` if both `min` and `max` are provided

Of course, this is a design decision (what to write in the default text message for the different cases, these are just examples).

> if the consumer has a `minLength` they can use it in the custom message and/or in the validation error, but we wouldn't benefit from ingesting this value within the `CharacterCount` as there were no pertinent values we could determine, and return to the end user.

This makes me question if we should have the `maxLength` at all, at this point, because the same `maxLength` argument is passed through directly (apart from being used for the `remainingLength`).

I don't want to block the component release, don't get me wrong: it's simply that there is something not quite working for me in this asymmetry (we're taking one use case - even if the maybe the most common - and elevating it to the "main" case, instead of making the component more agnostic/resilient.

packages/components/app/styles/components/form/character-count.scss
jorytindall: Looks good to me, thanks for addressing that!

packages/components/addon/components/hds/form/character-count/index.js
alex-ju: as requested, will add these in a new commit

packages/components/addon/components/hds/form/character-count/index.js
alex-ju: if the input gets removed first, the garbage collector will handle the event listener

packages/components/addon/components/hds/form/character-count/index.js
alex-ju: if used in a field component no. if used as a base element, hypothetically, yes, this can happen – not sure what we can do about that as we won't have an `id` to give the character count as a reference

didoo:
@alex-ju while talking with @jorytindall @andgen404 today, he mentioned that the `min/max` values should be numeric, and this made me think that we're not checking this or converting them to integer, and I was worried this may lead to possible bugs (`2 + "3" = "23"` in Javascript). But then I realized we do only one subtraction, to compute the remaining characters. So we should be safe. Can you think of other use cases where this could be a problem? Or should we convert them anyway, just in case (so maybe the consumer writes `@maxLength="100"` but then we convert it to an integer, so that when it's returned in the callback function it's ready to be used for them)? what do you think?

packages/components/addon/components/hds/form/character-count/index.hbs
```diff
@@ -0,0 +1,29 @@
...
+  ...attributes
+  aria-live="polite"
+>
+  {{#if (has-block)}}
+    {{yield
+      (hash
+        minLength=this.minLength
+        maxLength=this.maxLength
+        currentLength=this.currentLength
+        remainingLength=this.remainingLength
```
didoo, line 21: We should split this in two variables, one for the difference between `minLength` and `currentLength`, and one for the difference between `maxLength` and `currentLength` (names TBD, maybe `[min|max] + [Remaining/Balance/Left/Diff/Difference]`?)

packages/components/addon/components/hds/form/character-count/index.hbs
```diff
@@ -0,0 +1,29 @@
...
+      )
+    }}
+  {{else}}
+    {{this.currentLength}}
```
didoo, line 25: I would start to add the logic for the different use cases here, even if the copy is not finalized yet, so we can start to see them in action (see comment below about adding a "live/working" use case in the showcase)

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,136 @@
...
+    return parseInt(minLength);
+  }
+
+  /**
+   * @param remainingLength
+   * @type {number}
+   * @default null
+   * @description The remaining number of characters.
+   */
+  get remainingLength() {
```
didoo, line 46: see comment above about having two distinct arguments/values for the two possible differences (`min - curr` and `max - curr`)

packages/components/tests/dummy/app/templates/components/form/text-input.hbs
```diff
@@ -249,6 +259,44 @@
     </SG.Item>
   </Shw::Grid>
 
+  <Shw::Divider @level={{2}} />
+
+  <Shw::Text::H3>Character count</Shw::Text::H3>
```
didoo, line 264: What do you think of having an extra use case added to the showcase, where we connect the error with the callback function, to have the full cycle of functionality tested as we were consumers, to make sure we're not missing something (potentially, all the "counter" demos for text-input/textarea could be hooked to a validation, if that's not too complex, so we can test different scenarios)?

didoo (commented):
Left a few quick comment. By the look of it you are waiting for the final designs before making changes, so feel free to do everything in one last batch of changes.

alex-ju:
@didoo @jorytindall this PR should now reflect latest changes and is good for another review when you have time. if I get your thumbs up on this (in terms of API and functionality) I will rebase the documentation PR Andrew and Jory created (#1799) and add the tech docs.

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
```diff
@@ -141,6 +141,174 @@
 
   <Shw::Divider />
 
+  <Shw::Text::H2>Character count</Shw::Text::H2>
```
didoo, line 144: I had difficulties following the examples in this part of the showcase. If you don't mind, I am going to send you a patch (impossible to do a suggestion here) with how I would reorganize (and simplify) the code examples here, to see if works better for you too.

didoo (approved):
Left a few suggestions, but nothing blocking the release at this point.

I have to say, I'm super happy with the outcome: the labels for the different use case/conditions is simply perfect: it ticks all the requirements! Great job @andgen404 @jorytindall (an🙌d @alex-ju) 🙌

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,178 @@
...
+        'input',
+        this.updateCurrentLength,
+        true
+      );
+    }
+  }
+
+  @action
+  updateCurrentLength() {
+    this.currentLength = this.inputControl.value.length;
```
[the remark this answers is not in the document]
alex-ju: `updateCurrentLength` is only called if `this.inputControl` and `this.inputControl.value` exist, but sure, we can add these extra conditions

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
alex-ju: Applied the patch in ef3dcef

packages/components/addon/components/hds/form/character-count/index.js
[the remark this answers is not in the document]
didoo: No no you're right! Now I remember! All good then

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -0,0 +1,178 @@
...
+
+    if (typeof this.args.onInput === 'function') {
+      this.args.onInput(
```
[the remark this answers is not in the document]
alex-ju: yes, it feels awkward to skip so many arguments. I grouped them in an object in 959ad3b, let me know if that looks ok to you

jorytindall (approved):
This looks great! I did some brief testing on your branch, the flexibility is really nice here!

packages/components/addon/components/hds/form/character-count/index.js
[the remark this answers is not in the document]
didoo: yes, works better for me 👍

alex-ju:
@MelSumner when time allows, would you be able to give this feature a test and let us know if you find it conformant and/or usable with assistive tech?

didoo (approved):
Re-approving, just in case 👍 😀

PR #1803, opened 2023-11-13
Use dynamic import for the svg-sprite module

meirish:
### :pushpin: Summary
Change `lazyEmbed` feature to use dynamic `import()`.

👋 I was poking around a bundle analysis for `cloud-ui` and noticed I had a separate module from my webpack config, but also and embedded copy of the SVG sprite. After a little digging, I noticed the sprite was just using regular top-level`import` instead of the dynamic `import()` - so while it's not in the HTML, consumers would still be pulling the sprite in eagerly in their JS bundle.

This PR changes it so that the svg sprite will get bundled separately and imported when the initializer runs.

I tested this in the test app - adding this in the `config/environment.js`: 
```
emberFlightIcons: {
  lazyEmbed: true,
}
```
You can then see the difference in the browser dev tools Network panel when the application loads:
Before:
CleanShot 2023-11-13 at 14 38 30@2x

After:
CleanShot 2023-11-13 at 14 37 04@2x

### :link: External links
[Ember auto import docs on dynamic imports](https://github.com/embroider-build/ember-auto-import#dynamic-import)

Figma file: [if it applies]

***

### 👀 Component checklist


packages/ember-flight-icons/ember-cli-build.js
```diff
@@ -9,7 +9,9 @@ const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
 module.exports = function (defaults) {
   let app = new EmberAddon(defaults, {
-    // Add options here
+    babel: {
+      plugins: [require.resolve('ember-auto-import/babel-plugin')],
```
Dhaulagiri, line 13: [question] this is useful for local testing of this change within our addons (?), but is there something we need to ask consumer's to do on their side for this change to work properly?

meirish: I think this is for the dummy app, but the one in the `index.js` file makes it work for all consumers of the addon.

meirish: And now that I've said that I'm not sure - let me re-read the EAI docs.

meirish: https://github.com/embroider-build/ember-auto-import#installing-ember-auto-import-in-an-addon - yeah so I can remove this, it's just the `index.js` file change that's necessary.

meirish: Just pushed an update that removes this.

meirish:
I tested this in `cloud-ui` and it's working as expected.

The last bundle loaded here has the `load script` initiator indicating it's loaded from other JS (via dynamic import in this case):
CleanShot 2023-11-14 at 09 10 43

And here's the contents of that bundle:
CleanShot 2023-11-14 at 09 10 53

Dhaulagiri (approved):
Thanks!

PR #1805, opened 2023-11-16
remove trailing-history.js

Dhaulagiri:
### :pushpin: Summary

This was added in https://github.com/hashicorp/design-system/pull/690 as part of the field-guide exploration but AFAICT [is not being used](https://github.com/hashicorp/design-system/blob/eefa7dca3e19c1937c082b20f71808bb8cbe313f/website/config/environment.js#L13)


alex-ju (approved):
Indeed, doesn't seem to be used anywhere.

PR #1806, opened 2023-11-17
`Table` - Add example with `rowspan/colspan` to showcase

didoo:
### :pushpin: Summary

Context: https://hashicorp.slack.com/archives/C7KTUHNUS/p1700162017441919

In this PR I have:
- added to the showcase an example of an `Hds::Table` with `rowspan/colspan`attributes 

screenshot_3270

***

### 👀 Component checklist

### DIFF
--- a/packages/components/tests/dummy/app/templates/components/table.hbs
+++ b/packages/components/tests/dummy/app/templates/components/table.hbs
@@ -819,6 +819,57 @@
     </div>
   </div>
 
+  <Shw::Divider @level={{2}} />
+
+  <Shw::Text::H3>Table with <code>colspan/rowspan</code> attributes</Shw::Text::H3>
+
+  <Hds::Table>
+    <:head as |H|>
+      <H.Tr>
+        <H.Th>Lorem</H.Th>
+        <H.Th>Ipsum</H.Th>
+        <H.Th>Dolor</H.Th>
+        <H.Th>Sit amet</H.Th>
+      </H.Tr>
+    </:head>
+    <:body as |B|>
+      <B.Tr>
+        <B.Th rowspan="3">Scope Row with rowspan="3"</B.Th>
+        <B.Td>Cell Content</B.Td>
+        <B.Td>Cell Content</B.Td>
+        <B.Td>Cell Content</B.Td>
+      </B.Tr>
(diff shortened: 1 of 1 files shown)


alex-ju (approved):
Out of curiosity – do our designers have any thoughts on what cells spanning multiple rows should look like? (e.g. we could have vertical borders only for these cells to make it look less broken)

didoo:
> Out of curiosity – do our designers have any thoughts on what cells spanning multiple rows should look like? (e.g. we could have vertical borders only for these cells to make it look less broken)

@hashicorp/hds-design thoughts?

KristinLBradley:
The added example looks good to me but I would like to see input from design on the visual style first before merging.

didoo:
> Out of curiosity – do our designers have any thoughts on what cells spanning multiple rows should look like? (e.g. we could have vertical borders only for these cells to make it look less broken)

> The added example looks good to me but I would like to see input from @hashicorp/hds-design on the visual style first before merging.

@alex-ju @KristinLBradley here's @heatherlarsen comment: "In that case [it's only for the showcase], I think it’s fine to merge the PR. But I’ll create a ticket so we can come up with a better visual solution for this."

I think it's safe to merge at this point.

jorytindall:
Noted that the ticket has been created for addressing this in the future, but I think ideally there would be some visual differentiation between the rows and the spanning column. I'll defer to others, but if there's time it would be nice to include this in the design work for the the current table work/enhancements.

didoo:
> Noted that the ticket has been created for addressing this in the future, but I think ideally there would be some visual differentiation between the rows and the spanning column.

@hashicorp/hds-design I don't see how this is possible without introducing vertical lines for the cells. Should we introduce an option for consumers to have these vertical lines too, and provide guidance to use them only for specific use cases?

> I'll defer to others, but if there's time it would be nice to include this in the design work for the the current table work/enhancements.

Since we're working on the visual appearence of the table headers, and adding vertical lines impacts them visually, I would opt for a quick (quick!) exploration also of this option. As I remember, there were already explorations of this variant at the beginning, it should not be too complex (I _think_).

/cc @hashicorp/hds-engineering @Dhaulagiri

PR #1807, opened 2023-11-17
`SideNav` - Reduce the width of `SideNav::ToggleButton`

alex-ju:
### :pushpin: Summary

Reduce the width of `SideNav::ToggleButton` from `26px` to `24px`.

To prevent the button from overlapping other elements in pages where main content padding is narrow.

### :link: External links

As discussed with design manager and agreed with senior product designer

***

### 👀 Component checklist

### DIFF
--- a/packages/components/app/styles/components/side-nav/toggle-button.scss
+++ b/packages/components/app/styles/components/side-nav/toggle-button.scss
@@ -16,7 +16,7 @@
   display: flex;
   flex-direction: row-reverse;
   align-items: center;
-  width: 24px;
+  width: var(--hds-app-sidenav-toggle-button-width);
   height: 36px;
   padding: 0 4px;
   color: var(--token-color-foreground-high-contrast);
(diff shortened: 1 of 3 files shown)


packages/components/app/styles/components/side-nav/toggle-button.scss
```diff
@@ -16,7 +16,7 @@
   display: flex;
   flex-direction: row-reverse;
   align-items: center;
-  width: 26px;
+  width: 24px;
```
didoo, line 19: general question (not a blocker): since it's only a single entry point, should we convert it to a CSS variable so that we can keep it as is now (_**if**_ this is by design specs) and at the same time allow consumers to override it (via CSS variable) in "special" cases?

didoo (approved):
👍

packages/components/app/styles/components/side-nav/toggle-button.scss
alex-ju: exposed the button width as a CSS variable in c1666b187c7d8d47a5e0f7933730cfab51d7f92b – let me know what you think

didoo (approved):
👍

PR #1809, opened 2023-11-20
`Modal`, `Flyout` - Allow custom elements to be interposed with contextual components

alex-ju:
### :pushpin: Summary

Allow custom elements to be interposed with contextual components in `Modal` in `Flyout`

### :link: External links

Jira ticket: [HDS-2869](https://hashicorp.atlassian.net/browse/HDS-2869)

***

### 👀 Component checklist

[HDS-2869]: https://hashicorp.atlassian.net/browse/HDS-2869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/new-toes-brake.md
+++ b/.changeset/new-toes-brake.md
@@ -1,0 +1,6 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Modal` - Allowed custom elements to be interposed with contextual components
+`Flyout` - Allowed custom elements to be interposed with contextual components
--- a/packages/components/addon/components/hds/modal/index.hbs
+++ b/packages/components/addon/components/hds/modal/index.hbs
@@ -9,9 +9,13 @@
   {{did-insert this.didInsert}}
   {{focus-trap isActive=this.isOpen focusTrapOptions=(hash onDeactivate=this.onDismiss clickOutsideDeactivates=true)}}
 >
-  {{yield (hash Header=(component "hds/modal/header" id=this.id onDismiss=this.onDismiss))}}
-  {{yield (hash Body=(component "hds/modal/body"))}}
-  {{yield (hash Footer=(component "hds/modal/footer" onDismiss=this.onDismiss))}}
+  {{yield
+    (hash
+      Header=(component "hds/modal/header" id=this.id onDismiss=this.onDismiss)
+      Body=(component "hds/modal/body")
+      Footer=(component "hds/modal/footer" onDismiss=this.onDismiss)
+    )
+  }}
 </dialog>
 {{#if this.isOpen}}
   <div class="hds-modal__overlay"></div>
--- a/packages/components/tests/dummy/app/templates/components/modal.hbs
+++ b/packages/components/tests/dummy/app/templates/components/modal.hbs
@@ -244,11 +244,11 @@
   {{! template-lint-disable no-autofocus-attribute }}
   {{#if this.formModalActive}}
     <Hds::Modal id="form-modal" @onClose={{fn this.deactivateModal "formModalActive"}} as |M|>
-      <M.Header>
-        Why do you want to leave the beta?
-      </M.Header>
-      <M.Body>
-        <form name="leaving-beta-form">
+      <form name="leaving-beta-form">
+        <M.Header>
+          Why do you want to leave the beta?
+        </M.Header>
+        <M.Body>
           <Hds::Form::Select::Field autofocus @width="100%" as |F|>
             <F.Label>Select the primary reason</F.Label>
             <F.Options>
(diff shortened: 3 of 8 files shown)


.changeset/new-toes-brake.md
```diff
@@ -0,0 +1,6 @@
+---
+"@hashicorp/design-system-components": patch
```
didoo, line 2: Since this is technically a (potentially) breaking change(de facto, we're changing the API of the component, its intrinsic behaviour) should we raise it at least to `minor`?

packages/components/tests/integration/components/hds/flyout/index-test.js
```diff
@@ -72,6 +72,26 @@ module('Integration | Component | hds/flyout/index', function (hooks) {
     assert.dom('.hds-flyout__footer').hasText('Footer');
   });
 
+  test('it renders elements interposed with contextual components', async function (assert) {
```
didoo, line 75: nice!

didoo (commented):
Left a comment about semver.

Before approving, also the code snippets in the documentation files should be updated: 
- `website/docs/components/modal/partials/code/how-to-use.md`
- `website/docs/components/flyout/partials/code/how-to-use.md`

alex-ju:
> Left a comment about semver.
> 
> Before approving, also the code snippets in the documentation files should be updated:
> 
> * `website/docs/components/modal/partials/code/how-to-use.md`

I was under the impression that we want to accommodate such a use case (of interposing elements within the Modal) not to make it our recommendation. Will add a commit to update the form example.

> * `website/docs/components/flyout/partials/code/how-to-use.md`

Unsure what to update on this page. We don't have any examples with forms for the flyout as we currently recommend against it.

didoo:
> I was under the impression that we want to accommodate such a use case (of interposing elements within the Modal) not to make it our recommendation. Will add a commit to update the form example.

Mmm. I thought the reason we were doing it was because, when the modal contains a form, the buttons **should** (for accessibility reasons) be _inside_ the form. In which case, I think the demo/snippet should reflect this. 
If that is not the case, then why are we changing the working of the component? 

> Unsure what to update on this page. We don't have any examples with forms for the flyout as we currently recommend against it.

My bad 🤦‍♂️

KristinLBradley:
> ... We don't have any examples with forms for the flyout as we currently recommend against it.

Could we include an example just in the Showcase perhaps?

.changeset/new-toes-brake.md
KristinLBradley: It doesn't seem like it would break current usage. Can you think of any scenarios that could potentially cause issues that could be tested though?

KristinLBradley:
> > I was under the impression that we want to accommodate such a use case (of interposing elements within the Modal) not to make it our recommendation. Will add a commit to update the form example.
> 
> Mmm. I thought the reason we were doing it was because, when the modal contains a form, the buttons **should** (for accessibility reasons) be _inside_ the form. In which case, I think the demo/snippet should reflect this. If that is not the case, then why are we changing the working of the component?

To me, it makes sense to update this example so that the buttons are wrapped by the form element.

image

.changeset/new-toes-brake.md
alex-ju: The only breaking scenario I can think of is if someone invoked the subcomponents in a different order than the suggested one.

The example below would currently be corrected by the component and rendered in the pre-defined order (Header, Body, Footer).
```
  <Hds::Modal id="basic-modal" @onClose={{fn this.deactivateModal "basicModalActive"}} as |M|>
    <M.Body>
      <p class="hds-typography-body-300 hds-foreground-primary">Modal content</p>
    </M.Body>
    <M.Header>Modal title</M.Header>
    <M.Footer as |F|>
      <Hds::Button type="button" @text="Confirm" {{on "click" F.close}} />
    </M.Footer>
  </Hds::Modal>
```

After this change, the subcomponents would render in the order they were invoked in (Body, Header, Footer).

didoo:
@alex-ju @KristinLBradley let's wait for @MelSumner to be back and discuss with her before doing more extra work on this, it may not be necessary in the end (I was under the impression this was an a11y related issue, but by how @alex-ju explained the problem to me, seems is more of a documentation "problem")

packages/components/addon/components/hds/modal/index.hbs
```diff
@@ -9,9 +9,13 @@
   {{did-insert this.didInsert}}
   {{focus-trap isActive=this.isOpen focusTrapOptions=(hash onDeactivate=this.onDismiss clickOutsideDeactivates=true)}}
 >
-  {{yield (hash Header=(component "hds/modal/header" id=this.id onDismiss=this.onDismiss))}}
-  {{yield (hash Body=(component "hds/modal/body"))}}
-  {{yield (hash Footer=(component "hds/modal/footer" onDismiss=this.onDismiss))}}
+  {{yield
```
MelSumner, line 12: This is pretty consistent with how we do other components, yes?

packages/components/tests/dummy/app/templates/components/modal.hbs
```diff
@@ -244,11 +244,11 @@
   {{! template-lint-disable no-autofocus-attribute }}
   {{#if this.formModalActive}}
     <Hds::Modal id="form-modal" @onClose={{fn this.deactivateModal "formModalActive"}} as |M|>
-      <M.Header>
-        Why do you want to leave the beta?
-      </M.Header>
-      <M.Body>
-        <form name="leaving-beta-form">
+      <form name="leaving-beta-form">
```
MelSumner, line 247: This is how I would expect a form inside of a modal to work. 👍

.changeset/new-toes-brake.md
MelSumner: I think there's an incredibly low risk of these being invoked in the "wrong" order. Header > Body > Footer is overwhelmingly standard.

MelSumner (approved):
Yes this is a good change and improves alignment with how our other components work. 👍

packages/components/addon/components/hds/modal/index.hbs
didoo: No, that's the point in discussion. We do a single yield when the order doesn't matter, but in this case it matters. We would be introducing an "anti-pattern", making it difficult in future taking decisions when to use a single yield or multiple yields.

PR #1817, opened 2023-11-21
Website indexing script - Add custom `objectID` to records missing it

didoo:
### :pushpin: Summary

This _should_ "fix" the problem or duplicated records added when the script is run on some local machines (Brian/Alex) but not on others (Cristiano) or CI.

Important: it may have downsides, so we need to investigate better what is happening and why, but for now it could be an interim solution (still to be tested).

***

### 👀 Component checklist

### DIFF
--- a/website/docs-scraper/index-website-content.mjs
+++ b/website/docs-scraper/index-website-content.mjs
@@ -21,6 +21,7 @@ import dotenv from 'dotenv';
 import fs from 'fs-extra';
 import path from 'path';
 import chalk from 'chalk';
+import hash from 'object-hash';
 
 // import { pick } from 'lodash'; // not sure why this doesn't work in a `.mjs` module
 import _ from 'lodash';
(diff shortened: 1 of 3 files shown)


website/docs-scraper/index-website-content.mjs
```diff
@@ -324,6 +325,9 @@ async function indexWebsiteContent() {
     // here we construct the request to be sent to Algolia with the `batch/multiBatch` method
     // see: https://www.algolia.com/doc/api-reference/api-methods/batch/
     const algoliaRequests = algoliaRecords.map((record) => {
+      if (!record.objectID) {
+        record.objectID = `record__content-${record.type}-${hash(record)}`; // https://www.algolia.com/doc-beta/guides/sending-and-managing-data/send-and-update-your-data#unique-object-identifiers
```
Dhaulagiri, line 329: presumably using `hash` because there's nothing particularly unique on these records otherwise?

Dhaulagiri (approved):

PR #1818, opened 2023-11-22
Website search - Remove sidebar filter

didoo:
### :pushpin: Summary

Now that we have a proper search on the website, we can remove the filter in the sidebar.

In this PR I have: 

- removed the “filter” input from the sidebar

👉 👉 👉 Preview: https://hds-website-git-website-search-remove-sidebar-filter-hashicorp.vercel.app/

**Before:**

image

**After:** 

image

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2914

***

### 👀 Component checklist


heatherlarsen (approved):

alex-ju (commented):
I only reviewed the last commit (hope I got that right). Do we want to remove the `Doc::Form::Filter` component too?

didoo:
> I only reviewed the last commit (hope I got that right).

@alex-ju yeah, my bad: did't noticed it needed to be rebased. now it has only one commit (the one you've reviewed)

> Do we want to remove the `Doc::Form::Filter` component too?

I intended to keep it, but on a second thought is better to remove it (and re-name the `Doc::Form::Search` to `Doc::Form::Filter`, which is what technically is, since now we will have a proper search - I'll do it in a separate PR)

alex-ju (approved):

meirish:
:wave: Sorry I just stumbled across this, but do you think it might be useful to keep both filtering and search? I think a fast client-side filter is useful for when you know the name of something, but it may not be readily in front of you, and full-text search is better for searching over content instead of by name. Our API docs on https://developer.hashicorp.com seem to think of the content in this way as they provide both:

CleanShot 2023-11-27 at 16 37 03@2x

didoo:
> 👋 Sorry I just stumbled across this, but do you think it might be useful to keep both filtering and search? I think a fast client-side filter is useful for when you know the name of something, but it may not be readily in front of you, and full-text search is better for searching over content instead of by name. Our API docs on https://developer.hashicorp.com seem to think of the content in this way as they provide both:
> 
> CleanShot 2023-11-27 at 16 37 03@2x

No strong feelings about this. @heatherlarsen what do you think?

heatherlarsen:
@meirish @didoo Ah, that's a fair point. We could always keep it in place and gather feedback about it as folks start testing the new search.

PR #1819, opened 2023-11-22
`Table` - Add ability to visually hide a table column header

didoo:
### :pushpin: Summary

We have received some user requests to expose a way for them to be able to have tables without column headers. Since this is not possible for accessibility reasons, what we can do is to allow them to _visually hide_ the label in the table headers, but keeping the label in the DOM for screen readers.

In this PR I have:
- added `@isVisuallyHidden` argument to `Hds::Table::Th`
- updated `Hds::Table` to support `isVisuallyHidden` argument in the `@columns` object
- updated the showcase for `Table` component to include also examples container visually hidden table headers
- updated the “Component API” documentation
- added an example with visually hidden table headers to the “How to use” section
    - I've also updated second example to avoid complains from the compiler in case it’s rendered

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2355

***

### 👀 Component checklist

### DIFF
--- a/.changeset/tidy-rivers-sin.md
+++ b/.changeset/tidy-rivers-sin.md
@@ -3,4 +3,4 @@
 ---
 
 - `Hds::Table` - updated `@columns` object to support `isVisuallyHidden` argument
-- `Hds::Table::Th` - updated sub-component to support `isVisuallyHidden` argument
+- `Hds::Table::Th` - updated to support `isVisuallyHidden` argument
--- a/.changeset/tidy-rivers-sin.md
+++ b/.changeset/tidy-rivers-sin.md
@@ -1,0 +1,6 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+- `Hds::Table` - updated `@columns` object to support `isVisuallyHidden` argument
+- `Hds::Table::Th` - updated sub-component to support `isVisuallyHidden` argument
--- a/website/docs/components/table/partials/code/how-to-use.md
+++ b/website/docs/components/table/partials/code/how-to-use.md
@@ -547,6 +547,46 @@ In this case the table layout is still set to `auto` (default). If instead you w
 
 ### More examples
 
+#### Visually hidden table headers
+
+Labels within the table header are intended to provide contextual information about the column's content to the end user. There may be special cases in which that label is redundant from a visual perspective, because the kind of content can be inferred by looking at it (eg. a contextual dropdown). In this case it's possible to visually hide the label by passing `isVisuallyHidden=true` to the column (only for non-sortable tables).
+
+```handlebars
+<Hds::Table
+  @model={{this.model.myDemoData}}
+  @columns={{array
+    (hash key="artist" label="Artist" isSortable=true)
+    (hash key="album" label="Album" isSortable=true)
+    (hash key="year" label="Year" isSortable=true)
+    (hash key="other" label="Select an action from the menu" isVisuallyHidden=true width="60px")
+  }}
+>
+  <:body as |B|>
+    <B.Tr>
+      <B.Td>{{B.data.artist}}</B.Td>
+      <B.Td>{{B.data.album}}</B.Td>
+      <B.Td>{{B.data.year}}</B.Td>
+      <B.Td>
+          <Hds::Dropdown as |dd|>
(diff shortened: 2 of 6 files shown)


.changeset/tidy-rivers-sin.md
```diff
@@ -0,0 +1,6 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+- `Hds::Table` - updated `@columns` object to support `isVisuallyHidden` argument
+- `Hds::Table::Th` - updated sub-component to support `isVisuallyHidden` argument
```
alex-ju, line 6: This reads to me as _we've updated `Hds::Table::Th`s subcomponent to support `isVisuallyHidden` argument_, but it may be just me.

alex-ju (approved):

jorytindall (approved):
Quick turnaround on this, and great example to show where this might be useful!

.changeset/tidy-rivers-sin.md
MelSumner, line 6:
```suggestion
- `Hds::Table::Th` - updated to support `isVisuallyHidden` argument
```

KristinLBradley:
@didoo Are there any cases in which consumers may want to visually hide all th's?

This is a bad example but shows what it looks like if all th component are set to be visuallyHidden:
image

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -547,6 +547,46 @@ In this case the table layout is still set to `auto` (default). If instead you w
 
 ### More examples
 
+#### Visually hidden table headers
+
+Labels within the table header are intended to provide contextual information about the column's content to the end user. There may be special cases in which that label is redundant from a visual perspective, because the kind of content can be inferred by looking at it (eg. a contextual dropdown). In this case it's possible to visually hide the label by passing `isVisuallyHidden=true` to the column (only for non-sortable tables).
+
```
KristinLBradley, line 553: For clarity, I would call out which column has the header hidden (ie "the last column header" / etc) When looking at the example I was initially confused because I was expecting an example with all the table headers to be hidden (which may not be a valid use case but that was my initial expectation).

Also, if we don't want users to hide all table headers then I think it would be helpful to mention that they should not use isVisuallyHidden to do that.

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -555,10 +595,10 @@ Here’s a Table implementation that uses an array hash with localized strings f
...
   @model={{this.model.myDemoData}}
   @columns={{array
-      (hash key='artist' label=(t 'components.table.headers.artist') isSortable=true)
-      (hash key='album' label=(t 'components.table.headers.album') isSortable=true)
-      (hash key='year' label=(t 'components.table.headers.year') isSortable=true)
-      (hash key='other' label=(t 'global.titles.other'))
+      (hash key="artist" label=(t "components.table.headers.artist") isSortable=true)
+      (hash key="album" label=(t "components.table.headers.album") isSortable=true)
+      (hash key="year" label=(t "components.table.headers.year") isSortable=true)
+      (hash key="other" label=(t "global.titles.other"))
```
KristinLBradley, line 601: For me, since the array is JavaScript-like non-html code I would use single quotes.Is our standard preference to use double quotes in this context though?

KristinLBradley:
I know this has been merged already but I added just a few minor comments/suggestions & questions.

website/docs/components/table/partials/code/how-to-use.md
didoo: I think the chance that consumers will want to have a table with all the headers hidden would be very remote. It would be more likely that they would try to have a table completely without the table headers, which is something that already happened (and we provided guidance on why that was not OK). So I would not worry too much about this, but good thinking anyway 👍

website/docs/components/table/partials/code/how-to-use.md
didoo: > Is our standard preference to use double quotes in this context though?

@KristinLBradley yes (see all the other examples in the `how-to-use.md` files). The reason is that using single quotes is not valid in that context, and triggers an error:

screenshot_3296

PR #1821, opened 2023-11-22
allow for specific icons to be loaded in isolation

Dhaulagiri:
### :pushpin: Summary

Adds the ability for us (and consumers) to link to a specific icon in the icon library. This has a specific use case for our upcoming search feature.

https://hds-website-git-br-select-icon-hashicorp.vercel.app/

- [x] Add anchor link
- [x] Decide if we want to use new param or existing param
- [x] Add tests

### :link: External links

Jira ticket: [HDS-1708](https://hashicorp.atlassian.net/browse/HDS-1708)

***

[HDS-1708]: https://hashicorp.atlassian.net/browse/HDS-1708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/components/doc/icons-list/item.hbs
+++ b/website/app/components/doc/icons-list/item.hbs
@@ -18,5 +18,16 @@
     <Doc::MetaRow @label="Ember" @valueToCopy={{@meta.iconName}} @copyable={{true}} />
     <Doc::MetaRow @label="React" @valueToCopy={{@meta.name}} @copyable={{true}} />
     <Doc::MetaRow @label="Keywords" @valueToShow={{@meta.description}} />
+    {{#let (concat "icon:" @meta.iconName) as |searchQueryValue|}}
+      <LinkTo
+        class="doc-icons-list-grid-item__permalink"
+        @route="show"
+        @model="icons/library"
+        @query={{hash searchQuery=searchQueryValue selectedIconSize=@meta.size}}
+        aria-label="Permalink for the {{@meta.iconName}} icon"
+      >
+        <FlightIcon class="doc-icons-list-grid-item__permalink-icon" @name="link" />
+      </LinkTo>
+    {{/let}}
   </div>
 </li>
--- a/website/app/styles/doc-components/icons-list/index.scss
+++ b/website/app/styles/doc-components/icons-list/index.scss
@@ -78,6 +78,7 @@
 
   @include breakpoint.medium () {
     flex-direction: row;
+    align-items: flex-start;
   }
 }
 
@@ -85,14 +86,19 @@
   display: flex;
   flex: none;
   align-items: center;
-  justify-content: center;
+  justify-content: flex-start;
   width: 100%;
-  height: 120px;
+  height: 60px;
+  padding-left: 16px;
   border-bottom: 1px solid var(--doc-color-gray-500);
 
   @include breakpoint.medium () {
-    width: 120px;
+    align-items: flex-start;
+    justify-content: center;
+    width: 60px;
     height: 100%;
+    padding-top: 24px;
+    padding-left: 0;
     border-right: 1px solid var(--doc-color-gray-500);
     border-bottom: none;
   }
@@ -103,10 +109,39 @@
 }
 
 .doc-icons-list-grid-item__meta {
+  position: relative;
   width: 100%;
-  padding: 16px;
+  min-width: 0;
+  padding: 8px 16px;
 }
 
+.doc-icons-list-grid-item__permalink {
+  position: absolute;
+  top: 4px;
+  right: 4px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 24px;
+  height: 24px;
+  color: var(--doc-color-gray-300);
+
+  &:hover {
+    color: var(--doc-color-gray-100);
+  }
+
+  &:active {
+    color: var(--doc-color-feedback-information-100);
+  }
+}
+
+.doc-icons-list-grid-item__permalink-icon {
+  visibility: hidden;
+
+  .doc-icons-list-grid-item:hover & {
+    visibility: visible;
+  }
+}
 
 // not found
 
--- a/website/docs/icons/library/index.js
+++ b/website/docs/icons/library/index.js
@@ -34,11 +34,24 @@ export default class Index extends Component {
 
   get filteredIcons() {
     if (this.searchQuery) {
-      return this.allIcons.filter(
-        (i) =>
-          i.size === this.selectedIconSize &&
-          i.searchable.indexOf(this.searchQuery) !== -1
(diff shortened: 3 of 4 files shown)


website/docs/icons/library/index.js
```diff
@@ -34,11 +34,21 @@ export default class Index extends Component {
...
-      return this.allIcons.filter(
-        (i) =>
-          i.size === this.selectedIconSize &&
-          i.searchable.indexOf(this.searchQuery) !== -1
-      );
+      // check if the query is for an exact match (prefixed with `icon:`)
+      if (this.searchQuery.match(/^icon:/)) {
+        const iconValue = this.searchQuery.replace(/^icon:/, '');
+        const icon = this.allIcons.find((i) => {
+          // we want to match both `youtube` as well as `youtube-24`
```
Dhaulagiri, line 41: I'm curious about the 2nd part of this matching a string with the size in it. If `youtube-16` is passed in do we need to show that version of the icon?

website/app/components/doc/icons-list/item.hbs
```diff
@@ -18,5 +18,10 @@
     <Doc::MetaRow @label="Ember" @valueToCopy={{@meta.iconName}} @copyable={{true}} />
     <Doc::MetaRow @label="React" @valueToCopy={{@meta.name}} @copyable={{true}} />
     <Doc::MetaRow @label="Keywords" @valueToShow={{@meta.description}} />
+    <a
```
Dhaulagiri, line 21: there should be a way to ember-ify this so it doesn't cause the whole app to reload when this link is clicked. I'll see if i can find it.

Dhaulagiri: i got the link to be formed but clicking it doesn't refresh the page correctly 😭

website/app/components/doc/icons-list/grid.hbs
```diff
@@ -3,7 +3,7 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-<ul class="doc-icons-list-grid" role="list">
+<ul class="doc-icons-list-grid {{if @isExactSearch 'doc-icons-list-grid--is-exact-search'}}" role="list">
```
Dhaulagiri, line 6: i think we can drop these changes now?

website/app/components/doc/icons-list/item.hbs
Dhaulagiri: I spent a few more minutes looking at this and couldn't quite figure it out. I think the way it works now is ok though and we can always revisit later.

website/app/styles/doc-components/icons-list/index.scss
```diff
@@ -103,10 +109,36 @@
...
+  padding: 8px 16px;
 }
 
+.doc-icons-list-grid-item__permalink {
+  position: absolute;
+  top: 8px;
+  right: 8px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
```
KristinLBradley, line 124: The `align-items` & `justify-content` styles aren't having an effect. Add `display: flex;` to make them work.

website/app/styles/doc-components/icons-list/index.scss
```diff
@@ -103,10 +109,36 @@
...
   width: 100%;
-  padding: 16px;
+  min-width: 0;
+  padding: 8px 16px;
 }
 
+.doc-icons-list-grid-item__permalink {
+  position: absolute;
+  top: 8px;
+  right: 8px;
```
KristinLBradley, line 121: I would consider moving it farther into the corner to ad more space between the permalink and other content.

**Current 8px top & right:**
image
image

**Changed to 4px top & right:**
image
image

**Changed to 0 top & right:**
image
image

website/app/styles/doc-components/icons-list/index.scss
```diff
@@ -103,10 +109,36 @@
...
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 24px;
+  height: 24px;
+  color: var(--doc-color-gray-300);
+  visibility: hidden;
```
KristinLBradley, line 128: This is likely to cause accessibility issues. (Tagging @MelSumner)

See: https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#accessibility_concerns

website/app/styles/doc-components/icons-list/index.scss
didoo: @hashicorp/hds-design suggestions? alternative ideas?

website/app/styles/doc-components/icons-list/index.scss
didoo: Yeah, you're right, it was a sloppy implementation. I've pushed a change to fix it.

website/app/components/doc/icons-list/item.hbs
didoo: I did the same, and I don't understand why it doesn't work 🤔
Maybe something we can play with at the next HDS engineering sync?

Dhaulagiri: 👍  what I saw was that clicking the link updated the query params but it had no affect on the tracked property that was watching that param. It may be somewhat related to the unusual way these pages are wired up because everything shares the same route yet the "model" in this page's case is actually just reading directly from the flight-icons npm package so there is no model to "refresh" when the query params changes 🤷

website/app/styles/doc-components/icons-list/index.scss
heatherlarsen: I like @KristinLBradley's suggestion of `4px`; a happy medium!

Dhaulagiri: updated to `4px`

website/app/components/doc/icons-list/item.hbs
didoo: @Dhaulagiri I think we can go live with this version, and in a separate PR try to figure out how to overcome this issue.

didoo (approved):
Good to go (we'll figure out the re-loading in another PR)

website/app/styles/doc-components/icons-list/index.scss
KristinLBradley: Thanks! Yes I meant both the `top` and `right` positions

website/app/components/doc/icons-list/item.hbs
Dhaulagiri: Happy to back it out, but the diff we worked on was really close so I pushed up a working version here to get this in here. Have a look and let me know if you think it doesn't work.

PR #1823, opened 2023-11-22
Website - A11y testing / Share Axe config

didoo:
### :pushpin: Summary

In https://github.com/hashicorp/design-system/pull/1789 all the acceptance tests for the website pages  were failing because of an invalid/missing aria attribute on the injected (by Algolia Autocomplete) `.aa-Autocomplete` node:

screenshot_3289

After a discussion with @alex-ju we have come to the decision that the pragmatic solution is to introduce an `exclude` parameter for the website acceptance tests, so we can exclude that particular DOM node from the tests. To do that, we have to pass that extra argument to every `a11yAudit(...)` call in the tests, it's not possible to do it via the `setRunOptions()` method (apparently works only for the rules).

In this PR we have:
- created a shared a global Axe config
- used the shared config in all the `a11yAudit(...)` calls (only for the website)
- updated the blueprint for acceptance test for website pages

***

### 👀 Component checklist

### DIFF
--- a/website/tests/a11y-helper.js
+++ b/website/tests/a11y-helper.js
@@ -1,0 +1,16 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+export const globalAxeOptions = {
+  rules: {
+    'color-contrast': { enabled: false },
+    list: { enabled: false },
+  },
+  include: [['#ember-testing-container']],
+  exclude: [
+    // see: https://github.com/algolia/autocomplete/issues/963#issuecomment-1127507049
+    ['.aa-Autocomplete'],
+  ],
+};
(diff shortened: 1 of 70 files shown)


didoo:
screenshot_3291

@alex-ju all the tests are passing, so I am going to cherry pick only the relevant commits and exclude all the ones related to the website serch

alex-ju (approved):

website/tests/a11y-helper.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+export const globalAxeOptions = {
+  rules: {
+    'color-contrast': { enabled: false },
```
MelSumner, line 8: This should only be turned off in specific instances, not globally.

website/tests/a11y-helper.js
```diff
@@ -0,0 +1,16 @@
...
+    list: { enabled: false },
```
MelSumner, line 9: Same for this rule. No rule should be turned off globally but instead per instance that's needed.

website/tests/a11y-helper.js
didoo: @MelSumner this is how they were set before these changes:
https://github.com/hashicorp/design-system/pull/1823/files#diff-45093a48f9c59f17a4adbb1294bb86a6101e20105e5bd59fac1a5a36f293940b

maybe the declarations didn't get applied? 

In any case, I'll open a PR to remove these from the global declaration.

didoo: https://github.com/hashicorp/design-system/pull/1833

PR #1827, opened 2023-11-23
Website - Indexing script improvements

didoo:
### :pushpin: Summary

This PR is meant to fix an issue with discrepancies in the number of records logged in console at different times of the execution, and discrepancies between the record generated and the record actually available in the Algolia index after the script has completed its execution.

For context: believe me or not, but it took almost an entire day of debugging to find out all the issues and fix them.

In this PR I have:
- introduced a logical check to make sure the index is correctly cleared at the beginning of the process
  - if not, a `console.error` is triggered
- introduced logic to removed records with the same `objectID`
- improved overall messaging in console
- fixed the asynchronous array population

### :link: External links

Jira ticket:  no specific ticket, is part of the epic https://hashicorp.atlassian.net/browse/HDS-2862

***

### 👀 Component checklist

### DIFF
--- a/website/docs-scraper/index-website-content.mjs
+++ b/website/docs-scraper/index-website-content.mjs
@@ -206,7 +206,7 @@ async function indexWebsiteContent() {
 
     if (matches.length) {
       // extract from each "section" the tab name and the actual content
-      matches.forEach(async (match) => {
+      for (const match of matches) {
         const tabName = match[1];
         const tabContent = match[2];
         const currBaseRecord = _.merge({}, algoliaBaseRecord, {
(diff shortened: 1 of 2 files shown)


website/docs-scraper/index-website-content.mjs
```diff
@@ -186,11 +202,11 @@ async function indexWebsiteContent() {
     const matches = [...matchAllResults];
 
     // console.log('========================================');
-    console.log(chalk.white(`Processing file: ${fileRelativePath}`));
+    console.log(`Processing file: ${fileRelativePath}`);
 
     if (matches.length) {
       // extract from each "section" the tab name and the actual content
-      matches.forEach(async (match) => {
+      for (const match of matches) {
```
didoo, line 209: This was a silly error by me: essentially I was calling an async function within a `forEach` loop (🤦‍♂️) and this was causing the `markdownRecords` to be populated with different content even after the `forLoop` was executed and the effect was that the `markdownRecords.length` changed at different times in the execution of the code **_after_** the markdown was already processed.

alex-ju (approved):

PR #1831, opened 2023-11-27
Website - Remove leftover “dummy” class names

didoo:
### :pushpin: Summary

While working on the website indexing/search, I noticed there are some legacy `dummy-` class names left in the code (from the previous "scrappy" website).

This small PR cleans up these leftovers.

### 👀 Component checklist

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #1832, opened 2023-11-27
Remove draft code for `Avatar` and `EmptyState` components

didoo:
### :pushpin: Summary

Following @Dhaulagiri [comment](https://github.com/hashicorp/design-system/pull/1831#issuecomment-1827971971): should we remove `avatar` and `empty-state`? They were stubbed out over a year ago but we quickly changed plans and don't (currently) have any plans to finish them.

In this PR I have:
- removed code for  drafted `Avatar` component
- removed code for  drafted `EmptyState` component

Notice: I have not added a changelog entry because the components were never released.

***

### 👀 Component checklist


alex-ju (approved):
Technically this would mean a major release, but considering they've not been surfaced as components maybe it's fine to just remove them.

didoo:
> Technically this would mean a major release, but considering they've not been surfaced as components maybe it's fine to just remove them.

@Dhaulagiri what do you think? if we want, we can create a specific PR with the "major" changes like we did the last time, and little by little collect there all the breaking changes.

Dhaulagiri:
I see the point, but unless we have some evidence these are actually being used in code I don't think that is necessary. They were never documented or actually "released" in code (or figma for that matter).

PR #1833, opened 2023-11-27
Remove Axe global rules for `color-contrast` and `list`

didoo:
### :pushpin: Summary

Per @MelSumner request here: https://github.com/hashicorp/design-system/pull/1823/files#r1406379782

In this PR I have 
- removed Axe global rules for `color-contrast` and `list`

### DIFF
--- a/website/tests/a11y-helper.js
+++ b/website/tests/a11y-helper.js
@@ -4,10 +4,6 @@
  */
 
 export const globalAxeOptions = {
-  rules: {
-    'color-contrast': { enabled: false },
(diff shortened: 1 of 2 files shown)


alex-ju (approved):
So, it seems like `list` was disabled for the Dropdown component, and I suspect `color-contrast` might have been added for the SideNav (for the placeholder elements, you might remember the conversation), but now that's addressed locally so no need to port the rule anywhere.

MelSumner (approved):
Thank you for doing this!

PR #1757, opened 2023-10-30
Upgrade major `devDependencies`

alex-ju:
### :pushpin: Summary

Upgrade most (exceptions detailed below) `devDependencies` (majors) across the monorepo.

### :hammer_and_wrench: Detailed description

These are upgrades for our tooling dependencies and do not require releases for our packages.

This branch is based on #1704 to avoid having too many conflicts, with the intent to repoint it and merge it to `main` after #1704 gets in.

Exceptions:
 - `svgo`/`svgr`-related dependencies (we made [a conscious decision to use v 1.x](https://github.com/hashicorp/design-system/blob/main/packages/flight-icons/scripts/build-parts/optimizeAssetsSVG.ts#L9-L11))
 - `ember-template-lint`,  `eslint-config-prettier`, `eslint-plugin-prettier`, `prettier` (they all require reconfiguration and changes to package files, hence requiring a release, so will be addressed separately)
### :link: External links

Jira ticket: [HDS-2619](https://hashicorp.atlassian.net/browse/HDS-2619)

***

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-2619]: https://hashicorp.atlassian.net/browse/HDS-2619?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1838, opened 2023-11-28
Attempt to fix `yarn sync` on `flight-icons`

alex-ju:
### :pushpin: Summary

Not too familiar with the script, but if I got this right it should work as it used to.
Tested locally with a self-generated Figma token.

Follow-up on #1757


didoo (approved):
I tested it locally and _seems_ to work OK. Since currently is broken anyway, I think is safe to merge and give it a go.

PR #1853, opened 2023-12-04
`CodeBlock`: update to support dynamic content

KristinLBradley:
### :pushpin: Summary
If merged, this PR:
 - updates the `CodeBlock` component to enable it to render updated dynamic content
 - addresses an issue where line numbers are misaligned due to [an issue with font declaration](https://github.com/hashicorp/design-system/pull/1853/commits/cc377c8076e6fca1ebbd24435e956a2514518809)
 - suppresses [a build warning](https://hashicorp.atlassian.net/browse/HDS-2860) caused by parsing Ember comments in the component template.

Limitations:
 - dynamic content does not work with `@hasLineNumbers={{true}}`; this comes from a limitation with the [line-numbers Prism plugin](https://prismjs.com/plugins/line-numbers/) where it can only be instantiated/called once.
 - `@hasLineNumbers={{true}}` does not work as expected when used in a `Tabs` tab that is not visible at render time.

**Showcase:** https://hds-showcase-git-test-codeblock-with-dynamic-content-hashicorp.vercel.app/components/code-block

Misalignment issue in HCP. The more line there are the more you can see the issue.
<table>
<tr><td>

01-hcp-line-before

</td><td>

02-hcp-line-after

</td></tr>
</table>

### :link: External links

- Related Jira ticket: [HDS-2827](https://hashicorp.atlassian.net/browse/HDS-2827)

***

### 👀 Component checklist

- [x] If documenting a new component, an acceptance test that includes the `a11yAudit` has been added

[HDS-2827]: https://hashicorp.atlassian.net/browse/HDS-2827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/addon/components/hds/code-block/index.js
+++ b/packages/components/addon/components/hds/code-block/index.js
@@ -7,7 +7,7 @@ import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { assert } from '@ember/debug';
-import { next } from '@ember/runloop';
+import { next, schedule } from '@ember/runloop';
 import { htmlSafe } from '@ember/template';
 import { guidFor } from '@ember/object/internals';
 
@@ -121,16 +121,16 @@ export default class HdsCodeBlockIndexComponent extends Component {
           code,
           element,
         });
-      });
 
-      // Force prism-line-highlight plugin initialization
-      // Context: https://github.com/hashicorp/design-system/pull/1749#discussion_r1374288785
-      if (this.args.highlightLines) {
-        setTimeout(() => {
-          // we piggy-back on the plugin's `resize` event listener to trigger a new call of the `highlightLines` function: https://github.com/PrismJS/prism/blob/master/plugins/line-highlight/prism-line-highlight.js#L337
-          if (window) window.dispatchEvent(new Event('resize'));
-        }, 100);
-      }
+        // Force prism-line-highlight plugin initialization
+        // Context: https://github.com/hashicorp/design-system/pull/1749#discussion_r1374288785
+        if (this.args.highlightLines) {
+          schedule('afterRender', () => {
+            // we piggy-back on the plugin's `resize` event listener to trigger a new call of the `highlightLines` function: https://github.com/PrismJS/prism/blob/master/plugins/line-highlight/prism-line-highlight.js#L337
+            if (window) window.dispatchEvent(new Event('resize'));
+          });
+        }
+      });
     }
   }
 
--- a/packages/components/addon/components/hds/code-block/index.js
+++ b/packages/components/addon/components/hds/code-block/index.js
@@ -125,6 +125,7 @@ export default class HdsCodeBlockIndexComponent extends Component {
         // Force prism-line-highlight plugin initialization
         // Context: https://github.com/hashicorp/design-system/pull/1749#discussion_r1374288785
         if (this.args.highlightLines) {
+          // we need to delay re-evaluating the context for prism-line-highlight for as much as possible, and `afterRender` is the 'latest' we can use in the component lifecycle
           schedule('afterRender', () => {
             // we piggy-back on the plugin's `resize` event listener to trigger a new call of the `highlightLines` function: https://github.com/PrismJS/prism/blob/master/plugins/line-highlight/prism-line-highlight.js#L337
             if (window) window.dispatchEvent(new Event('resize'));
--- a/packages/components/addon/components/hds/code-block/index.js
+++ b/packages/components/addon/components/hds/code-block/index.js
@@ -107,18 +107,20 @@ export default class HdsCodeBlockIndexComponent extends Component {
     const grammar = Prism.languages[language];
 
     if (code) {
-      if (language && grammar) {
-        this.prismCode = htmlSafe(Prism.highlight(code, grammar, language));
-      } else {
-        this.prismCode = htmlSafe(Prism.util.encode(code));
-      }
+      setTimeout(() => {
+        if (language && grammar) {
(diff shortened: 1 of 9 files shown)


packages/components/addon/components/hds/code-block/index.js
```diff
@@ -107,18 +107,20 @@ export default class HdsCodeBlockIndexComponent extends Component {
...
+          this.prismCode = htmlSafe(Prism.util.encode(code));
+        }
+
+        // Force prism-line-numbers plugin initialization, required for Prism.highlight usage
+        // See https://github.com/PrismJS/prism/issues/1234
+        Prism.hooks.run('complete', {
+          code,
+          element,
+        });
+      }, 100);
```
KristinLBradley, line 123: I tried using `next` to replace this `setTimeout` and it still functions.

I tried using `schedule` instead and got errors but maybe I was using it wrong so I'll experiment a bit more.

didoo: OK, then let's wait for @alex-ju to come back from PTO and we can then decide how to take it from here.

KristinLBradley: It seems to work although I can do more testing.

alex-ju: if we could also test this change to make sure it addresses the actual need in `boundary` (where you identified the issue) that would be even better – let me know if I can be of help

KristinLBradley: @alex-ju I tried figuring out how to add an integration test but haven't been able to figure out a way to do it.

I can try to test it just in Boundary next though.

alex-ju: cool! I can look into the integration test bit

KristinLBradley: I tested the CodeBlock updates in Boundary and everything worked as expected. The content updates dynamically.

packages/components/tests/integration/components/hds/code-block/index-test.js
```diff
@@ -33,6 +38,23 @@ module('Integration | Component | hds/code-block/index', function (hooks) {
...
+  test('it renders the passed in dynamic content', async function (assert) {
+    this.set('value', "console.log('Hello world');");
+    await render(hbs`
+      <Hds::CodeBlock @value={{this.value}} id="test-code-block" />
+    `);
+    assert
+      .dom('#test-code-block pre code')
+      .hasText("console.log('Hello world');");
+    this.set('value', "console.log('Lorem ipsum');");
+    await settled();
```
alex-ju, line 52: I initially tried with `rerender`, but it's too 'quick' (in a sense that it doesn't wait for the whole promise to end) so I ended up using `settled`.

packages/components/tests/dummy/app/templates/components/code-block.hbs
```diff
@@ -77,6 +79,34 @@ console.log(`I am ${codeLang} code`);"
...
+  <Shw::Text::Body>
+    <label for="input">Change input value:</label>
+    <input id="input" type="text" value={{this.input}} {{on "input" this.updateInput}} />
+  </Shw::Text::Body>
+
+  <Hds::CodeBlock
+    @value={{this.codeValue}}
+    @language="javascript"
+    @hasCopyButton={{true}}
+    @hasLineNumbers={{false}}
```
didoo, line 101: I noticed that if you set this to "true" (and probably we should) when the lines gets updated, the line numbers don't.

packages/components/addon/components/hds/code-block/index.js
```diff
@@ -107,27 +108,29 @@ export default class HdsCodeBlockIndexComponent extends Component {
...
+        // See https://github.com/PrismJS/prism/issues/1234
+        Prism.hooks.run('complete', {
+          code,
+          element,
+        });
+
+        // Force prism-line-highlight plugin initialization
+        // Context: https://github.com/hashicorp/design-system/pull/1749#discussion_r1374288785
+        if (this.args.highlightLines) {
+          schedule('afterRender', () => {
```
didoo, line 128: do we still need to invoke the "afterRender" even if it's inside a `next()` callback?

didoo (commented):
Left a couple of comments.

packages/components/addon/components/hds/code-block/index.js
alex-ju: yes, unfortunately, we still need to delay the `prism-line-highlight` for as much as possible, and `afterRender` seems to be the 'latest' we can use.

didoo: OK (maybe add a comment in the code explaining this detail?)

packages/components/tests/dummy/app/templates/components/code-block.hbs
alex-ju: I looked into this during the first week of January and today again. Unfortunately, I can not see a way in which the `@hasLineNumbers` feature would work with dynamic content.

For this reason, I suggest the pragmatic approach of fixing the immediate issues (I updated the PR description to highlight what this PR covers and what doesn't) to address current instances in HCP and Boundary then leave the `hasLineNumbers` issue for future evaluation (it may require us to fork the addon or move away from Prism altogether). If you find this approach acceptable, I can create a ticket to track this use case and update the CodeBlock guidance to call out the current limitations.

packages/components/addon/components/hds/code-block/index.js
alex-ju: Added a comment explaining the need for `afterRender`

didoo (approved):
LGTM 👍

PR #1859, opened 2023-12-07
`Table` `Multi-select` component (HDS-2953)

KristinLBradley:
### :pushpin: Summary
If merged, this PR adds MultiSelect features and associated child components to the `Table`.

* **Showcase:** https://hds-showcase-git-hds-2953-table-multiselect-hashicorp.vercel.app/components/table

- TODO

### :link: External links

- Jira ticket: [HDS-2953](https://hashicorp.atlassian.net/browse/HDS-2953)
- Figma file: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/IzAfFGMUiLT14h8VAqBfor/HDS-Product---Components?type=design&node-id=16727%3A59704&mode=design&t=z27RMs2OCy5hYlxq-1

***

### 👀 Component checklist

[HDS-2953]: https://hashicorp.atlassian.net/browse/HDS-2953?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/tests/dummy/app/routes/components/table.js
```diff
@@ -4,6 +4,7 @@
  */
 
 import Route from '@ember/routing/route';
+import { DENSITIES } from '@hashicorp/design-system-components/components/hds/table';
```
KristinLBradley, line 7: Let me know if you don't like this and want me to revert it.

packages/components/addon/components/hds/table/index.js
```diff
@@ -8,7 +8,7 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { assert } from '@ember/debug';
 
-const DENSITIES = ['short', 'medium', 'tall'];
+export const DENSITIES = ['short', 'medium', 'tall'];
```
KristinLBradley, line 11: Let me know if you don't like this and want me to revert it.

packages/components/addon/components/hds/table/index.hbs
```diff
@@ -11,6 +11,9 @@
   <thead class="hds-table__thead">
     {{#if @columns}}
       <Hds::Table::Tr>
+        {{#if @isSelectable}}
+          <Hds::Table::ThSelectable />
+        {{/if}}
```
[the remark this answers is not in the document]
KristinLBradley: Right now this won't work because I'm rendering a `th` when it's in the thead and a `td` when it's in the tbody.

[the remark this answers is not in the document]
didoo: Two considerations:
- can we read it as a "code smell" that they should be the same HTML tag? @MelSumner any strong objection at using a `<th>` also in the `tbody` for the row selector?
- in any case, I suspect we may need to pass the information to the row if it's in the `thead` or `tbody`, to differentiate the behaviour of the checkbox (there may also be a workaround for this, it depends on how we implement the logic)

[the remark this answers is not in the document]
KristinLBradley: To me it makes sense to use a `th` for the cells in the `tbody` since they refer to selecting the entire row.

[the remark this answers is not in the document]
KristinLBradley: I went ahead and deleted the `td` version and replaced it with the `th`. This allowed me to simplify the way `@isSelectable` works. I'm having an issue with the `scope` attribute rendered on the `th`'s in the `tbody` though. They all have `scope="col"` even though I'm trying to pass "row" as the value.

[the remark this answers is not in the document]
KristinLBradley: I still can't figure out how to get the scope to work on the `ThSelectable`/`th`'s that render in the `tbody` (they need to have `scope="row"`).

packages/components/addon/components/hds/table/index.hbs
```diff
@@ -33,8 +34,9 @@
     {{else}}
       {{yield
         (hash
-          Tr=(component "hds/table/tr")
+          Tr=(component "hds/table/tr" isSelectable=@isSelectable)
           Th=(component "hds/table/th")
+          ThSelectable=(component "hds/table/th-selectable")
```
didoo, line 39: Now that the `ThSelectable` is controlled directly via the `Tr` element, we don't need (and want) to yield it anymore.
```suggestion
```

packages/components/addon/components/hds/table/index.hbs
```diff
@@ -54,8 +57,9 @@
       {{#each (sort-by this.getSortCriteria @model) key=this.identityKey as |record|}}
         {{yield
           (hash
-            Tr=(component "hds/table/tr")
+            Tr=(component "hds/table/tr" isSelectable=@isSelectable)
             Th=(component "hds/table/th" scope="row")
+            ThSelectable=(component "hds/table/th-selectable" scope="row")
```
didoo, line 62:
```suggestion
```

packages/components/addon/components/hds/table/index.hbs
```diff
@@ -65,9 +69,10 @@
     {{else}}
       {{yield
         (hash
-          Tr=(component "hds/table/tr")
+          Tr=(component "hds/table/tr" isSelectable=@isSelectable)
           Th=(component "hds/table/th" scope="row")
           Td=(component "hds/table/td" align=@align)
+          ThSelectable=(component "hds/table/th-selectable" scope="row")
```
didoo, line 75:
```suggestion
```

packages/components/addon/components/hds/table/index.js
didoo: No problem at all, it's a standard pattern in our repo.

packages/components/addon/components/hds/table/tr.hbs
```diff
@@ -3,5 +3,9 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 <tr class="hds-table__tr" ...attributes>
+  {{#if @isSelectable}}
+    <Hds::Table::ThSelectable @scope={{@scope}} />
```
didoo, line 7: In order to know which scope to use here, you have to have the information where the row is injected, if it's in the `<thead>` or in the `<tbody>`. What I would probably do, is to pass some argument to the row to provide this information, and add conditional logic accordingly to decide which scope value to use (`"row"` or `"col"`). This argument can be then used also to determine the function to invoke when the checkbox is clicked (this depends on the implementation of the logic associated to the checkbox elements).

didoo (commented):
Added a couple of comments/suggestions, I hope this unblocks you. In any case I am going to set up a meeting for tomorrow to make sure there are no other blockers and discuss the next steps to get this over the line.

packages/components/addon/components/hds/table/index.hbs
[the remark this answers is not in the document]
KristinLBradley: Thanks, I got the scope working now.

packages/components/addon/components/hds/table/index.hbs
KristinLBradley: Thanks I alrady figured out I need to remove these as I was working on it.

KristinLBradley:
Got very rough functionality working for clicking on the table header checkbox to select/deselect all rows. Probably not exactly how we'd want the final implementation to work though...

packages/components/addon/components/hds/table/th-selectable.js
```diff
@@ -0,0 +1,39 @@
...
+    if (typeof onChange === 'function') {
+      onChange();
+    }
+  }
+
+  @action
+  didInsert(element) {
+    const dataScope = element.getAttribute('data-scope');
+
+    if (dataScope === 'row') {
```
didoo, line 32: you don't need to pass this information down to the template (`.hbs`), store it as `data-` attribute, and then re-read it back: you already have this information here: try to console.log the value of `this.args.scope` in this function, and you will see what I mean

packages/components/addon/components/hds/table/tr.hbs
```diff
@@ -2,6 +2,16 @@
...
   SPDX-License-Identifier: MPL-2.0
 }}
+
 <tr class="hds-table__tr" ...attributes>
+  {{#if @isSelectable}}
+    <Hds::Table::ThSelectable
+      @isSelected={{@isSelected}}
+      @onChange={{@onChange}}
+      @scope={{@scope}}
+      @didInsert={{@didInsert}}
```
didoo, line 12:
```suggestion
      @scope={{@scope}}
      @isSelected={{@isSelected}}
      @didInsert={{@didInsert}}
      @onChange={{@onChange}}
```

KristinLBradley:
Pushed up implementation of basic callbacks when rows are selected and setting status of the header checkbox based on individual row selection.

I ended up adding some state in order to easily update the status of the checkboxes vs. trying to query the DOM which I was having problems with. This could be changed later if we don't want to do it this way.

packages/components/addon/components/hds/table/index.js
```diff
@@ -179,4 +182,57 @@ export default class HdsTableIndexComponent extends Component {
...
+    if (scope === 'row') {
+      this.checkboxRowItems.push(checkbox);
+    } else if (scope === 'col') {
+      this.checkboxHeaderItem = checkbox;
+    }
+  }
+
+  @action
+  onSelectionChange(id, checked) {
+    if (id === 'all') {
```
didoo, line 197: probably there was a bit of misunderstanding with [my writeup](https://docs.google.com/document/d/1JK8Wg3aHA6nWCf1XZA0DySwNIVnNdblKoRztss6VFiw/edit#heading=h.2zx3tm4vdiia): when I said "A single callback is invoked `onSelect` that returns which rows have been selected"  I meant on the consumer side (so they have a unified callback to handle); internally we can have two different functions associated with the two kind of select, one for the "select all" and one for the "select row". In this way we can avoid these `if`s with "all" a s a string passed as argument, and we can also differentiate the behaviours a bit more.

packages/components/addon/components/hds/table/index.js
```diff
@@ -179,4 +182,57 @@ export default class HdsTableIndexComponent extends Component {
       onSort(this.sortBy, this.sortOrder);
     }
   }
+
+  @action
+  addCheckbox(checkbox, scope) {
+    if (scope === 'row') {
+      this.checkboxRowItems.push(checkbox);
```
didoo, line 189: This works for the first rendering of the table, but what happens if rows are added/removed: then this array goes out of sync with what's in the DOM (same problem that happened with the `Tabs` and `didInsert`. 
One thing we discussed was to have the callback check the status of the rows **when** the user triggered the callback: in this way we always had the exact status in the DOM, not the one in memory that may not correspond to the one in the DOM.

(I've pushed a commit with a couple of examples for filtered and paginated data, you can play with it and you'll see what I mean)

packages/components/addon/components/hds/table/th-selectable.js
```diff
@@ -0,0 +1,37 @@
...
+   * @return {string}
+   */
+  checkboxId = 'checkbox-' + guidFor(this);
+
+  @action
+  onChange(event) {
+    let { onChange } = this.args;
+    if (typeof onChange === 'function') {
+      let { scope } = this.args;
+      if (scope === 'col') {
```
didoo, line 22: as mentioned above, we could differentiate the function provided to the `@onChange` argument and avoid the need for these conditionals

packages/components/addon/components/hds/table/th-selectable.js
```diff
@@ -0,0 +1,37 @@
...
+        onChange('all', event.target.checked);
+      } else if (scope === 'row') {
+        onChange(event.target.id, event.target.checked);
+      }
+    }
+  }
+
+  @action
+  didInsert(element) {
```
didoo, line 31: see comment above, about the limitations on relying on `didInsert` to track the rows

didoo (commented):
@KristinLBradley left a few comments and pointers (I've also pushed a couple of extra examples of tables with filters and pagination, that can help testing the current implementation of the selection logic)

packages/components/addon/components/hds/table/index.js
```diff
@@ -179,4 +182,57 @@ export default class HdsTableIndexComponent extends Component {
...
+    if (selectedRowCount === 0) {
+      this.checkboxHeaderItem.checked = false;
+      this.checkboxHeaderItem.indeterminate = false;
+    } else if (selectedRowCount === rowCount) {
+      this.checkboxHeaderItem.checked = true;
+      this.checkboxHeaderItem.indeterminate = false;
+    } else {
+      this.checkboxHeaderItem.checked = false;
+      this.checkboxHeaderItem.indeterminate = true;
+    }
```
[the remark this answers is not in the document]
KristinLBradley: Making change locally together with others

packages/components/addon/components/hds/table/index.js
KristinLBradley: Pushed up refactor

packages/components/addon/components/hds/table/index.js
KristinLBradley: Pushed up changes to now use the DOM vs tracking in memory.

The examples you added still don't work because when the DOM changes, the status of the header doesn't get updated. I need to see how to trigger the `setTableStatus` function next.

KristinLBradley: Do you have any suggestions on the best way to trigger it? (I'm still experimenting to see if I can figure it out.)

packages/components/addon/components/hds/table/index.js
```diff
@@ -179,4 +182,58 @@ export default class HdsTableIndexComponent extends Component {
...
+    checkboxes.forEach((checkbox) => {
+      checkbox.checked = checked;
+    });
+  }
+
+  setTableState() {
+    const tableHeader = document.querySelector(`#${this.tableId} thead`);
+    const tableHeaderCheckbox = tableHeader.querySelector(
+      '.hds-table__checkbox'
+    );
```
didoo, line 228: these two selectors can be merged

```suggestion
    const tableHeaderCheckbox = document.querySelector(
      #${this.tableId} thead .hds-table__checkbox`
    );
```

packages/components/addon/components/hds/table/index.js
```diff
@@ -179,4 +182,58 @@ export default class HdsTableIndexComponent extends Component {
...
+
+    let rowCount = this.allTbodyCheckboxes().length;
+    let selectedRowCount = this.allTbodyCheckboxes().filter(
+      (checkbox) => checkbox.checked === true
```
didoo, line 232: I _think_ this works as well
```suggestion
      (checkbox) => checkbox.checked
```

packages/components/addon/components/hds/table/index.js
```diff
@@ -179,4 +182,58 @@ export default class HdsTableIndexComponent extends Component {
...
+    return checkboxes;
+  }
+
+  changeAllRowsSelection(checkboxes, checked) {
+    checkboxes.forEach((checkbox) => {
+      checkbox.checked = checked;
+    });
+  }
+
+  setTableState() {
```
didoo, line 224: technically, as is now, this is not setting the "table" state but the "select all" state

packages/components/addon/components/hds/table/index.js
```diff
@@ -179,4 +182,58 @@ export default class HdsTableIndexComponent extends Component {
...
+    let checkboxes = [];
+
+    tableBodyRows.forEach((row) => {
+      const checkbox = row.querySelector('.hds-table__checkbox');
+      checkboxes.push(checkbox);
+    });
+    return checkboxes;
+  }
+
+  changeAllRowsSelection(checkboxes, checked) {
```
didoo, line 218: this is invoked only once, are you sure we need to abstract it as a separate function?

packages/components/addon/components/hds/table/index.js
```diff
@@ -179,4 +182,58 @@ export default class HdsTableIndexComponent extends Component {
...
+  allTbodyCheckboxes() {
+    const tableBody = document.querySelector(`#${this.tableId} tbody`);
+    const tableBodyRows = Array.from(tableBody.rows);
+    let checkboxes = [];
+
+    tableBodyRows.forEach((row) => {
+      const checkbox = row.querySelector('.hds-table__checkbox');
+      checkboxes.push(checkbox);
+    });
+    return checkboxes;
```
didoo, line 215:
```suggestion
    const tableBodyCheckboxes = document.querySelectorAll(
      `#${this.tableId} tbody .hds-table__checkbox`
    );
    return Array.from(tableBodyCheckboxes);
```

packages/components/addon/components/hds/table/index.hbs
```diff
@@ -54,7 +66,14 @@
       {{#each (sort-by this.getSortCriteria @model) key=this.identityKey as |record|}}
         {{yield
           (hash
-            Tr=(component "hds/table/tr")
+            Tr=(component
+              "hds/table/tr"
+              isSelectable=@isSelectable
+              didInsert=this.addCheckbox
```
didoo, line 72: I can't find the `addCheckbox` function, maybe these `didInsert` are not needed anymore?

didoo (commented):
Added a few comments and suggestions

packages/components/addon/components/hds/table/index.js
didoo: The only things I can think of are
- a dedicated function that the consumers can invoke to "refresh" the status of the "select all" whenever the rows are updated 
- re-introduce back the ember modifiers for `did-insert` + `will-destroy` and keep track ourselves of changes to the DOM structure (this may lead to resolve the need for some DOM traversing that we're doing in the code); but in theory these modifiers are deprecated in "modern" Ember, but frankly I don't see them go anywhere in the near future 

I have set up a sync meeting for Friday where we can brainstorm a bit about this a bit more.

/cc @alex-ju

KristinLBradley: Thanks.

I can't figure out how the first one would work to expose a function.

I've been trying the 2nd one to add back `did-insert` & `will-destroy` but so far there's an issue passing the callback function from the table to the checkbox I've been trying to figure out.

KristinLBradley: I pushed up updates to use `did-insert` & `will-destroy`. They are just calling the `setSelectAllState` function right now. It appears to fix the problems with the inline filter and pagination examples in terms of the header checkbox state but the state of the checkboxes in the `tbody` is still lost when the DOM updates.

didoo: > I pushed up updates to use `did-insert` & `will-destroy`. They are just calling the `setSelectAllState` function right now. It appears to fix the problems with the inline filter and pagination examples in terms of the header checkbox state 

Great, it means we're on the right track :)

> but the state of the checkboxes in the `tbody` is still lost when the DOM updates.

Which is expected, no? Seems the ambassadors agreed on the fact that preserving the status is a consumer's concern.

One thing I noticed: in this example the "select all" doesn't work anymore
modified

didoo: Some other pointers:
- https://github.com/hashicorp/cloud-ui/pull/7235 (shared by @meirish)
- https://www.material-react-table.com/docs/guides/row-selection (see in the devtools the console.log response on selection)
- https://react-table-library.com/?path=/docs/compact-table--select (see console in devtools)
- https://www.ag-grid.com/react-data-grid/row-selection/#selection-events
- https://tanstack.com/table/v8/docs/api/features/row-selection#onrowselectionchange
  - https://codesandbox.io/p/devbox/tanstack-table-example-row-selection-2llvty

packages/components/addon/components/hds/table/th-selectable.js
```diff
@@ -0,0 +1,58 @@
...
+      // TODO: Doesn't update in reaction to row checkboxes being selected
+      if (this._checked === true) {
+        return 'deselect all';
+      } else {
+        return 'select all';
+      }
+    } else {
+      return `Select ${this.args.selectionKey}`;
+    }
+  }
```
KristinLBradley, line 31: @MelSumner This is still a work in progress but does this make sense for the checkbox aria-labels?

KristinLBradley:
@alex-ju Pushed up a number of refactors relating to our discussion Thurs Dec. 21 including adding a selection key on the row that is passed back when row selection changes.

Getting some a11y errors currently that I haven't tracked down the cause of yet some of which look to be related to dropdowns.

alex-ju (commented):
Functionally, it seems to be doing what we want (from my tests using the showcase examples). Now, the next challenge (that may require some adjustments to the existing code and/or mental model) is providing relevant information through the callback functions.

P.S. couldn't see any issues when scanning the showcase page with axe – where are you seeing the a11y errors?

packages/components/addon/components/hds/table/index.hbs
```diff
@@ -65,7 +76,15 @@
...
         (hash
-          Tr=(component "hds/table/tr")
+          Tr=(component
+            "hds/table/tr"
+            isSelectable=@isSelectable
+            onSelectionChange=this.onSelectionRowChange
+            isSelected=this.isSelected
+            didInsert=this.didInsertRowCheckbox
+            willDestroy=this.willDestroyRowCheckbox
+            scope="row"
```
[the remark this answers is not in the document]
KristinLBradley: Sorry this was an accident. Will clean up

packages/components/addon/components/hds/table/index.hbs
```diff
@@ -54,7 +57,15 @@
...
           (hash
-            Tr=(component "hds/table/tr")
+            Tr=(component
+              "hds/table/tr"
+              isSelectable=@isSelectable
+              onSelectionChange=this.onSelectionRowChange
+              isSelected=this.isSelected
+              didInsert=this.didInsertRowCheckbox
+              willDestroy=this.willDestroyRowCheckbox
+              scope="row"
```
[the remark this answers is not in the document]
KristinLBradley: It's passed down. I could rename if it's confusing.

packages/components/addon/components/hds/table/index.hbs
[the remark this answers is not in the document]
KristinLBradley: It looks like the a11y errors resolved themselves. Maybe it was some temporary glitch from something.

packages/components/addon/components/hds/table/index.js
KristinLBradley: Btw, select all should now be working again. I had a mistake in one of the function names passed.

packages/components/addon/components/hds/table/index.hbs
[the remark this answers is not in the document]
KristinLBradley: I renamed it so it's hopefully less confusing now.

packages/components/addon/components/hds/table/index.hbs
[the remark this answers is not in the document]
KristinLBradley: Renamed argument for clarity.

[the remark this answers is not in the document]
alex-ju: ah, yes, you're right – it's an argument for a contextual component (using the helper) not an actual attribute

KristinLBradley:
@didoo @alex-ju Pushed up integration tests for the multi-select table. Please let me know if there's other things I should test.

jorytindall (commented):
One thing I noticed in the custom filter example was the state doesn't persist when the filter parameter is changed. I imagine this is something that will be a requirement for consumers to implement themselves, but just noting it in case it was unexpected. What I did:

- Select row #2 (even), then filter to show only even rows. My selection persists here.
- Switch to filter by only odd rows, then no rows show as selected (expected), but switching back to even rows clears my selection.

No problematic unless this was expected to persist.

didoo:
> One thing I noticed in the custom filter example was the state doesn't persist when the filter parameter is changed. I imagine this is something that will be a requirement for consumers to implement themselves, but just noting it in case it was unexpected. What I did:
> 
> * Select row 2 (even), then filter to show only even rows. My selection persists here.
> * Switch to filter by only odd rows, then no rows show as selected (expected), but switching back to even rows clears my selection.
> 
> Not a problem unless this was expected to persist.

@jorytindall yes, it's intended (at code level) in the sense that the state is not persisted on our side, it's a consumer concern. See thread here: https://hashicorp.slack.com/archives/C03A0N1QK8S/p1702939324037479

packages/components/tests/dummy/app/controllers/components/table.js
```diff
@@ -148,4 +156,49 @@ export default class ComponentsTableController extends Controller {
...
+  }
+
+  @action
+  onPaginatedSelectablePageChange(page) {
+    this.currentPaginatedSelectablePage = page;
+  }
+
+  @action
+  onPaginatedSelectablePageSizeChange(pageSize) {
+    // we agreed to reset the pagination to the first element (any alternative would result in an unpredictable UX)
```
andgen404, line 200: @KristinLBradley Does this mean, like the example in the showcase, that selecting the parent checkbox on page 1 will select all checkboxes, but on page 2, the parent checkbox will reset to unchecked, and so will the child ones?

didoo: @andgen404 This is the current behaviour, and could be the "safest" one. I'll give you an example. Let's say I am an admin and I on a page with a list of users. I see a table, with a pagination. There are 40 records (users), I see 10 of them (4 pages) so I see 10 rows in the table. Now, I click "select all" and all the rows are selected. If I click the button "delete" next to the table of users, what should I expect to happen? That I delete the 10 users that I see (and only them) or that I delete all the 40 users (even if I don't see them?) 
Let's imagine I intended to only delete the ones that I see on screen, how would I know that the application instead is deleting all of them (because "secretly" has selected also the rows on the other pages)?
Now, if this is the case, probably we want to avoid this dangerous case, and when "select all" is clicked, only the visible rows are selected, right? But then what happens if the user clicks "select all" on the first page (so the "select all" becomes checked, and all the rows are checked) and then clicks on the next page? Well, according to the "safe" logic above, the only option we have is to show the rows in the second table's "page" as unselected, as well as the "select all". Do you see any alternatives to this?

KristinLBradley: When I've used similar features features such as for managing email, it only selects the items visible on the page as Cristiano. Personally, I think that's the safest behavior and seems to be common.

andgen404: Ah, okay. Thank you, @didoo, for the example! I don't see any alternatives; I agree with this "Safe" approach as well.

didoo: > similar features features such as for managing email

Gmail inbox behaves exactly like this :)

MelSumner:
I think there should be an example of a sortable table with multi-select. WDYT?

KristinLBradley:
@didoo I refactored the onSelectionChange related functions to now pass selection keys corresponding to the selected rows to the callback function and added the 2nd functional example to animate selected users.

KristinLBradley:
@didoo Pushed up another refactor to add an isSelected argument so it's possible to persist table row selections in paginated data. (You can see this working in the pagination example in the Showcase.)

Users could also use this to write a function to select all rows in a paginated table even if not visible. (I can add an example of this if you think it would be useful.)

packages/components/tests/dummy/app/controllers/components/table.js
MelSumner: I think for "select all" it should mean "select all visible rows" (and possibly even give a row count - "select all 20 rows")

Then there's an upgrade to that, an additional option to select all rows, even non-visible ones ("select all 100 rows in this table")

packages/components/addon/components/hds/table/index.js
```diff
@@ -179,4 +184,106 @@ export default class HdsTableIndexComponent extends Component {
...
+    if (checkbox.checked === true) {
+      checkbox.ariaLabel = this.args.deselectRowAriaLabel
+        ? `${this.args.deselectRowAriaLabel} ${checkbox.id}`
+        : `Deselect ${checkbox.id}`;
+    } else {
+      checkbox.ariaLabel = this.args.selectRowAriaLabel
+        ? `${this.args.selectRowAriaLabel} ${checkbox.id}`
+        : `Select ${checkbox.id}`;
+    }
+  }
```
MelSumner, line 264: It's not meaningful enough to have the checkbox's generated ID as the aria-label value, so we're going to have to think about this one some more. 

Some potential options to consider:
- row count (select row #)
- row count out of total (select row # of ##) 
- use the row's TH (scope of row) (select row Nick Drake) 

Also ultimately, we want to allow them to set the aria-label themselves, which provides room to grow/provide translations as the value instead.

MelSumner (changes requested):

packages/components/addon/components/hds/table/index.js
KristinLBradley: Thanks, I did provide an option for translations with this though.

packages/components/addon/components/hds/table/index.js
KristinLBradley: I could also probably write a function to make the selections persist now also if you think it would be helpful.

KristinLBradley:
> One thing I noticed in the custom filter example was the state doesn't persist when the filter parameter is changed. I imagine this is something that will be a requirement for consumers to implement themselves, but just noting it in case it was unexpected. What I did:
> 
> * Select row 2 (even), then filter to show only even rows. My selection persists here.
> * Switch to filter by only odd rows, then no rows show as selected (expected), but switching back to even rows clears my selection.
> 
> Not a problem unless this was expected to persist.

This requires writing a custom function to make the selections persist. Since that example was added I made some refactors which now make this easier to do so I could write a function for the Showcase example to do that if you think it would be helpful.

packages/components/addon/components/hds/table/index.js
KristinLBradley: Cristiano is making some related updates in his test branch so I'll wait to go over those with him before making further accessibility-related updates in this branch. Thanks for reviewing.

KristinLBradley:
@didoo I could add another example with pagination which includes a button for selecting checkboxes on all pages (even if not visible) if you think that would be helpful. (@andgen404 thought it could be useful.)

packages/components/tests/dummy/app/controllers/components/table.js
didoo: As discussed, for this release we will settle on `Select all rows` / `Deselect all rows`.

packages/components/addon/components/hds/table/th-selectable.js
KristinLBradley: resolving as outdated

KristinLBradley:
> I think there should be an example of a sortable table with multi-select. WDYT?

@didoo was saying this was something we could consider later I believe...

packages/components/app/styles/components/table.scss
```diff
@@ -258,4 +267,23 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
       vertical-align: baseline;
     }
   }
+
+  // multi-select (isSelectable=true)
+
+  .hds-table--is-selectable & {
+    .hds-table__th--is-selectable {
+      width: 1px; // shrink-wrap content (will expand as needed)
+      padding-right: 8px;
```
KristinLBradley, line 276: @didoo Do you think I should create a sass variable for this padding value?

KristinLBradley:
@andgen404 @hashicorp/hds-design 

This is how the new table header border style looks with the checkbox column. Looks a bit strange to me but what do you think? Would you want to hide the in-between border in this case?

image

packages/components/addon/components/hds/table/index.js
```diff
@@ -179,4 +187,97 @@ export default class HdsTableIndexComponent extends Component {
       onSort(this.sortBy, this.sortOrder);
     }
   }
+
+  @action
+  onSelectionAllChange() {
+    // this.setSelectAllCheckboxAriaLabel(this.selectAllCheckbox);
+    this.selectableRows.forEach((row) => {
+      row.checkbox.checked = this.selectAllCheckbox.checked;
```
KristinLBradley, line 195: @didoo This isn't changing the `isSelected` tracked value of the row checkbox, so the ariaLabel for the rows don't change when you select or deselect all (using the header checkbox)

So far I haven't figured a fix for this (although I'll keep investigating). It looks like you were trying something previously that you commented out though?

didoo: It's a tricky one. I have to think about it. 🤔

packages/components/app/styles/components/table.scss
didoo: Let's review the table cell (`hds-table__th--is-selectable`) layout today, there is something strange for me... 🤔

packages/components/addon/components/hds/table/tr.js
```diff
@@ -0,0 +1,25 @@
...
+export default class HdsTableTrComponent extends Component {
+  /**
+   * @param selectionKey
+   * @type {string}
+   * @default undefined
+   */
+  get selectionKey() {
+    if (this.args.isSelectable && this.args.selectionScope === 'row') {
+      assert(
+        `@selectionKey must be defined when @isSelectable is true`,
```
didoo, line 18: @KristinLBradley we probably need to make the message more meaningful providing more context (eg. it's not clear that this refers to a table row).

modified

andgen404:
> @andgen404 @hashicorp/hds-design
> 
> This is how the new table header border style looks with the checkbox column. Looks a bit strange to me but what do you think? Would you want to hide the in-between border in this case?
> 
> image

Hmm, yeah, it does look off. I noticed there is no 16px padding on the right side of the `header column / select` are we not able to apply that padding on the right?

packages/components/addon/components/hds/table/tr.js
KristinLBradley: Pushed up clarified wording.

KristinLBradley:
> > @andgen404 @hashicorp/hds-design
> > This is how the new table header border style looks with the checkbox column. Looks a bit strange to me but what do you think? Would you want to hide the in-between border in this case?
> > image
> 
> Hmm, yeah, it does look off. I noticed there is no 16px padding on the right side of the `header column / select` are we not able to apply that padding on the right?

@andgen404 The Figma mocks only have 8px of padding on the right side of the checkboxes. Also the padding can't be different on the header cells in the thead vs. the tbody row cells.

packages/components/app/styles/components/table.scss
```diff
@@ -103,6 +103,12 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
   }
 }
 
+// multi-select (isSelectable=true)
+
+.hds-table__th--is-selectable {
+  width: 48px;
```
KristinLBradley, line 109: @didoo I changed the width to 48px to match the width of the checkbox. I set it to 1px before so that it would "shrink-wrap" the content without having to set an exact width (as the comment I had said.)

If I remove this width, then the table cell expands to be wider than the content. (I don't know what you were doing exactly when you looked at it.)

heatherlarsen:
> > > @andgen404 @hashicorp/hds-design
> > > This is how the new table header border style looks with the checkbox column. Looks a bit strange to me but what do you think? Would you want to hide the in-between border in this case?
> > > image
> > 
> > 
> > Hmm, yeah, it does look off. I noticed there is no 16px padding on the right side of the `header column / select` are we not able to apply that padding on the right?
> 
> @andgen404 The Figma mocks only have 8px of padding on the right side of the checkboxes. Also the padding can't be different on the header cells in the thead vs. the tbody row cells.
> 
> I just looked around in the Figma mocks some more and just noticed that the padding uses the default in some of the mocks now instead of the reduced padding that I thought was the original design. Which would you like?
> 
> @andgen404 I updated it to match your latest mocks now.

I'm having trouble following this thread. @KristinLBradley @andgen404 can one of you please confirm what the resolution was here? Thanks.

KristinLBradley:
> > > > @andgen404 @hashicorp/hds-design
> > > > This is how the new table header border style looks with the checkbox column. Looks a bit strange to me but what do you think? Would you want to hide the in-between border in this case?
> > > > image
> > > 
> > > 
> > > Hmm, yeah, it does look off. I noticed there is no 16px padding on the right side of the `header column / select` are we not able to apply that padding on the right?
> > 
> > 
> > @andgen404 The Figma mocks only have 8px of padding on the right side of the checkboxes. Also the padding can't be different on the header cells in the thead vs. the tbody row cells.
> > I just looked around in the Figma mocks some more and just noticed that the padding uses the default in some of the mocks now instead of the reduced padding that I thought was the original design. Which would you like?
> > @andgen404 I updated it to match your latest mocks now.
> 
> I'm having trouble following this thread. @KristinLBradley @andgen404 can one of you please confirm what the resolution was here? Thanks.

Sorry for the confusion, Andrew confirmed to me that the checkboxes should now use the default padding of 16px on both the right and left sides. I updated the CSS accordingly.

heatherlarsen:
Awesome, thanks for confirming @KristinLBradley.

packages/components/app/styles/components/table.scss
didoo: That's OK with 48px.

packages/components/package.json
```diff
@@ -94,6 +94,8 @@
     "ember-cli-string-helpers": "^6.1.0",
     "ember-cli-terser": "^4.0.2",
     "ember-concurrency": "^3.1.1",
+    "ember-copy": "^2.0.1",
```
didoo, line 97: @alex-ju once we will separate the showcase to its own standalone app, this medley of packages for the components vs the showcase should be fixed once for all

packages/components/app/styles/components/table.scss
KristinLBradley: Ok well I preferred 1px since then it doesn't matter what size the checkbox is since the cell would always tightly wrap it but if it causes confusion, 48px is ok.

packages/components/addon/components/hds/table/th-selectable.js
```diff
@@ -0,0 +1,75 @@
...
+  @action
+  didInsert(checkbox) {
+    let { didInsert } = this.args;
+    if (typeof didInsert === 'function') {
+      didInsert(checkbox, this.args.selectionKey);
+      // we need to use a custom event listener here because changing the `checked` value via JS
+      // (and this happens with the "select all") doesn't trigger the `change` event
+      // and consequently the `aria-label` won't be automatically updated (and so we have to force it)
+      checkbox.addEventListener(
+        'toggle',
```
alex-ju, line 39: as discussed, I suggest we prefix these events if we go down this path

alex-ju (commented):
Looks pretty solid to me. Left a few suggestions to overcome the failing a11y tests.

alex-ju (approved):

KristinLBradley:
@MelSumner Could you please re-review for accessibility? Do we still need to add some sort of unique name to the Table for referencing, etc? (I believe in the last discussions we didn't have a good answer for this.)

KristinLBradley:
I tried to rebase on top of the latest `main` and I didn't realize this component branch was still pointing to `[table-new-sorting-and-tooltip](https://github.com/hashicorp/design-system/tree/table-new-sorting-and-tooltip)` so it messed up the history.

KristinLBradley:
I [messed up the history of the component branch](https://github.com/hashicorp/design-system/pull/1859#issuecomment-1909151703). Need to figure out how to fix that to then fix this branch. :(

didoo:
> I tried to rebase on top of the latest `main` and I didn't realize this component branch was still pointing to `[table-new-sorting-and-tooltip](https://github.com/hashicorp/design-system/tree/table-new-sorting-and-tooltip)` so it messed up the history.

@KristinLBradley it happens, don't worry.

To fix it I have
- re-based `table-new-sorting-and-tooltip` off `main`
- re-based this branch off `table-new-sorting-and-tooltip`
- pushed (forced)

and everything seems OK. Can you double check?

packages/components/addon/components/hds/table/index.js
didoo: Closing this conversation now

packages/components/addon/components/hds/table/index.js
```diff
@@ -179,4 +184,106 @@ export default class HdsTableIndexComponent extends Component {
...
+      checkbox.ariaLabel = this.args.selectRowAriaLabel
+        ? `${this.args.selectRowAriaLabel} ${checkbox.id}`
+        : `Select ${checkbox.id}`;
+    }
+  }
+
+  setHeaderCheckboxAriaLabel(checkbox) {
+    checkbox.ariaLabel = checkbox.checked
+      ? this.args.deselectAllAriaLabel ?? 'Deselect all'
+      : this.args.selectAllAriaLabel ?? 'Select all';
```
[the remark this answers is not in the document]
didoo: @MelSumner we've updated the logic around the `aria-label`s for the checkbox elements, can you check it's OK for you?

packages/components/addon/components/hds/table/index.js
didoo: Closing this thread, now the `aria-label` values are more meaningful

packages/components/package.json
didoo: In the end I've removed the dependency and used a simple "clone" function.

MelSumner (approved):
I've reviewed the accessibility of these changes and see no issues with this approach at this time. :shipit:

packages/components/tests/dummy/app/controllers/components/table.js
```diff
@@ -148,4 +179,251 @@ export default class ComponentsTableController extends Controller {
...
+      });
+    } else {
+      selectableRowsStates.forEach((row) => {
+        const recordToUpdate = this.multiSelectModelData__demo1.find(
+          (modelRow) => modelRow.id === row.selectionKey
+        );
+        if (recordToUpdate) {
+          recordToUpdate.isSelected = row.isSelected;
+        }
+      });
```
alex-ju, line 244: I gave some thought to this, in conjunction with the conversation in Slack.

I think it works just fine as it is, but I added a suggestion if we're determined to improve computation.

In the current implementation we have a time complexity of O(n*m) in the worst case (n is the length of `selectableRowsStates`; m is the length of `multiSelectModelData__demo1`). That's because for each row in `selectableRowsStates`, we are searching for a matching row in `multiSelectModelData__demo1` using the `find` method, which can take up to m iteration.

If we create a map of `multiSelectModelData__demo1` by `id`, it will allow us to find the matching row in constant time, reducing the time complexity to O(n+m).

```suggestion
      const modelDataMap = this.multiSelectModelData__demo1.reduce(
        (map, modelRow) => {
          map[modelRow.id] = modelRow;
          return map;
        },
        {}
      );

      selectableRowsStates.forEach((row) => {
        const recordToUpdate = modelDataMap[row.selectionKey];
        if (recordToUpdate) {
          recordToUpdate.isSelected = row.isSelected;
        }
      });
```

Arguably, the proposed solution is not easier to read than what we currently have. And for a reasonable number of rows, it won't make a big difference.

alex-ju (approved):

KristinLBradley:
@andgen404 Could you please review the implemented component designs and leave feedback if you have any or else approve?

andgen404 (approved):
The design looks good to me 🎉

packages/components/tests/dummy/app/controllers/components/table.js
KristinLBradley: If it's more of a best practice to reduce time complexity in this way then it could be useful to update from that regard (to serve as a better example). @didoo what do you think?

jorytindall (approved):
:shipit:

packages/components/addon/components/hds/table/th-selectable.js
didoo: Opened PR https://github.com/hashicorp/design-system/pull/1912to address this, so we can discuss pros/cons of the proposed alternative approach.

packages/components/tests/dummy/app/controllers/components/table.js
didoo: Addressed here: https://github.com/hashicorp/design-system/pull/1859/commits/573baf9e3de7bd39352fb00b13c4182b84d25e40

packages/components/addon/components/hds/table/th-selectable.js
didoo: @alex-ju @KristinLBradley after further analysis (see comment https://github.com/hashicorp/design-system/pull/1912#issuecomment-1917085102) I have realized that the proposed solution was fundamentally flawed/incomplete, and have it working as expected would require a not-so-great change to the current code implementation.

I don't see many alternatives at this point:
- leave things as they are now, with the custom event
- have the `table/index.js` do DOM manipulation and update the `aria-label` values behind the scenes (this may become complex anyway, because the aria label suffix is declared at row level)
- imagine something even more complex (services, pubsub, etc)

I would lean towards leaving the code as it is (unless we decide to ask for help from other Ember devs in the company). 

Thoughts?

alex-ju:
Rebased this to get latest changes from `table-new-sorting-and-tooltip`. Will get it merged as soon as the tests pass.

PR #1860, opened 2023-12-07
`Table` - New sorting and tooltip

didoo:
### :pushpin: Summary

This PR implements the changes to the `Hds::Table` header for the new sorting styling and the introduction of tooltips.

In this PR I have:
- added `Table::ThButtonSort` and `Table::ThButtonTooltip` as standalone subcomponents
- replaced content of `Th/ThSort` with the newly created subcomponents and implemented the new layout according to the design specs
- multiple refactorings of the CSS code:
    - simplified CSS classes for Table by replacing `hds-table__th` with a modifier `hds-table__th--sort`
    - large refactoring and re-organization of the Table’s CSS to simplify code
    - renamed text alignment class names for consistency with `valign` one
- updated showcase with examples for table headers with tooltip and improved existing ones for sorting, ThSort, Th
- updated (and added missing) integration tests
- added codemod for `hbs` files for `ThSort` change of `onClickSort` callback

Notice: the documentation will be updated in https://github.com/hashicorp/design-system/pull/1871 for simplicity of reviewing

 👉👉 👉 **Preview**: https://hds-showcase-git-table-new-sorting-and-tooltip-hashicorp.vercel.app/components/table

### 🚧 TODOs

- [ ] ~~add codemod for `.gts` files~~ (see [thread](https://hashicorp.slack.com/archives/C04TTMDBPNV/p1705528939102509)) - to be discussed with @alex-ju 
    - decided it was not worth it, @aklkv mentioned he can do the required changes manually

### :link: External links

Jira tickets:
- https://hashicorp.atlassian.net/browse/HDS-2956
- https://hashicorp.atlassian.net/browse/HDS-2955

Figma file: 
- https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/KQH4JqkQeHxCYqa5RcOBTp/HDS-Product---Components?type=design&node-id=51135-63741&mode=design&t=TUWb1LyZhHe46FSZ-0

***

### 👀 Component checklist


packages/components/addon/components/hds/table/th-button-sort.hbs
```diff
@@ -0,0 +1,13 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<button
+  type="button"
+  class="hds-table__th-button hds-table__th-button--sort"
+  {{on "click" this.onClick}}
+  aria-label="sort"
```
didoo, line 9: @MelSumner is this OK? do you think we need to do something different or something else?

packages/components/addon/components/hds/table/th-button-tooltip.hbs
```diff
@@ -0,0 +1,14 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<button
+  type="button"
+  class="hds-table__th-button hds-table__th-button--tooltip"
+  {{hds-tooltip @text}}
+  aria-label="more information"
```
didoo, line 10: @MelSumner is this OK? do you think we need to do something different or something else?

packages/components/app/styles/components/table.scss
```diff
@@ -199,10 +190,68 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
 
-      // A TH WILL NEVER BE LAST CHILD, ONLY FIRST CHILD
+      // a <th> will never be last child, only first child
       td:last-child {
         border-bottom-right-radius: $hds-table-inner-border-radius;
       }
     }
   }
+
+  .hds-table__th,
```
jorytindall, line 200: Similarly to the cell that @andgen404 and I have been noticing, the way the padding and the spacing is handled is a bit different between Figma and code. I don't think that's a problem necessarily, but a small result is the border between columns as a pseudo element is 35px in height, in Figma it's 36 (I know, nit picky, we can decide whether this really matters).

packages/components/app/styles/components/table.scss
```diff
@@ -19,175 +19,166 @@ $hds-table-cell-padding-medium: 14px 16px 13px 16px; // the 1px difference is to
...
+  padding: 0;
+  color: var(--token-color-foreground-faint);
+  background-color: transparent;
+  border: 1px solid transparent;
+  border-radius: 3px;
 
-  .hds-table--density-medium & {
-    padding: $hds-table-cell-padding-medium;
+  &:hover,
+  &.mock-hover {
```
jorytindall, line 131: The hover state for the sort/tooltip button mimics the `secondary` button and has `Elevation / Low` that only shows on hover.

jorytindall (commented):
Left a couple of comments, the spacing question is tough, I remember it was hard to get Figma and code to match 100% when we built these initially, so it might be a non-issue.

The focus ring is also slightly different here (more accurate I think in this case), so I'll update my Figma component.

packages/components/app/styles/components/table.scss
didoo: This is because how borders are applied in an HTML table. Essentially we have one border all around the entity `<table>` (which you don't have) and then other borders are applied to the table cells (`<th>/<td>`) in different ways. This leads to the `<th>` cells in the `<thead>` to optically be `49px` instead of `48px` (they're `48px` in height, but there's an extra border on the top, the table border). 
In order to achieve an exact match with Figma, the CSS code would be overly complicated for little gain. So we decided (time ago, when the `Hds::Table` was implemented, to keep things simpler for sake or maintainability).

packages/components/app/styles/components/table.scss
didoo: Done in https://github.com/hashicorp/design-system/pull/1860/commits/8bcb86e36fe11b140d5ccbd8d801d6426758621e

packages/components/app/styles/components/table.scss
```diff
@@ -19,175 +19,170 @@ $hds-table-cell-padding-medium: 14px 16px 13px 16px; // the 1px difference is to
...
+      &::after {
+        position: absolute;
+        top: 6px;
+        right: 0;
+        bottom: 6px;
+        width: 1px;
+        background-color: $hds-table-border-color;
+        content: "";
+        pointer-events: none;
       }
```
KristinLBradley, line 69: This is fine, but this is how I prefer to handle this kind of "in-between" item in order to avoid needing an additional `:last-child` style to override it.

```
+ th::before {
    position: absolute;
    top: 6px;
    left: -1px; // positions it where the actual th border would be if "turned on"
    bottom: 6px;
    width: 1px;
    background-color: var(--token-color-border-primary);
    content: "";
    pointer-events: none;
}
```

(Final selector in browser would be something like: `.hds-table__thead .hds-table__tr th + th::before`)

packages/components/app/styles/components/table.scss
```diff
@@ -19,175 +19,170 @@ $hds-table-cell-padding-medium: 14px 16px 13px 16px; // the 1px difference is to
...
 
-      // LAST TH IN THE ROW, BORDER RADIUS
-      th:last-child {
-        border-top-right-radius: $hds-table-inner-border-radius;
+      &:last-child::after {
+        background-color: transparent;
```
KristinLBradley, line 72: See above for a suggestion to remove the need for this extra overriding style.

packages/components/addon/components/hds/table/td.js
```diff
@@ -42,7 +42,7 @@ export default class HdsTableTdComponent extends Component {
 
     // add a class based on the @align argument
     if (this.align) {
-      classes.push(`hds-table__td--text-${this.align}`);
+      classes.push(`hds-table__td--align-${this.align}`);
```
KristinLBradley, line 45: Nice. I think this naming is clearer as to the purpose

KristinLBradley (approved):
I went through the code and tested out the examples within the Showcase. Everything looks reasonable to me. I just added a suggestion for a way to possibly simplify styling for the new "inbetween" `th` borders.

packages/components/app/styles/components/table.scss
didoo: @KristinLBradley keep in mind that I've done a large refactoring of this Sass code (see https://github.com/hashicorp/design-system/pull/1860/commits/46526e38a90b00ff2d9f62b7f297012aaf1dce9f + https://github.com/hashicorp/design-system/pull/1860/commits/1ae7f227f7551a6d270d21dfa63327324c7a8c51) so I would not focus too much on the layout/styling details (until I've done the rebase of this branch on the other one) and more on the JavaScript for [select/unselect] [all/row]

jorytindall (approved):
This looks great! One note: now the example of the tooltip in the table in the tooltip showcase is obsolete/unnecessary. I tried removing it myself, but I think I may have broken a test for some reason. That'll teach me to step in and change things 😔

.changeset/modern-zebras-destroy.md
```diff
@@ -0,0 +1,28 @@
...
+- Refactored CSS code to simplify usage of `hds-table`-related class names
+
+`Table::ThSort`:
+
+- Added support for tooltip via the `@tooltip` argument
+- Updated visual treatment and location of the "sort" button
+- Updated DOM structure of the `<th>` content
+- Remove class `hds-table__th-sort--button-content`
+- Replaced class `hds-table__th-sort` with classes `hds-table__th` + `hds-table__th--sort`
+- Replaced class `hds-table__th-sort--text--[left|center|right]` with `hds-table__th--align-[left|center|right]`
```
alex-ju, line 18: Leaving a comment for clarification. We treated class name changes as major in [the past](https://github.com/hashicorp/design-system/blob/main/packages/components/CHANGELOG.md#200). Do we consider that past decision a one-off and moving forward we do not see class name changes as majors?

packages/components/addon/components/hds/table/th-button-sort.hbs
alex-ju, line 9: This will potentially result in multiple buttons with the same label, which is confusing for screen reader users. Ideally, we would incorporate the column name (the `<th>` text content) so that we have something like 'sort by Artist' and 'sort by Release Year'.

packages/components/addon/components/hds/table/th-button-sort.js
```diff
@@ -0,0 +1,20 @@
...
+
+export default class HdsTableThButtonSortComponent extends Component {
+  /**
+   * @param onClick
+   * @type {function}
+   * @default () => {}
+   */
+  get onClick() {
+    let { onClick } = this.args;
+    return onClick || NOOP;
```
alex-ju, line 18: In other places we check for this argument to be a function, we may want to consider using the same pattern for consistency.
```suggestion
    if (typeof onClick === 'function') {
      return onClick;
    } else {
      return NOOP;
    }
```

packages/components/app/styles/components/table.scss
```diff
@@ -19,175 +19,166 @@ $hds-table-cell-padding-medium: 14px 16px 13px 16px; // the 1px difference is to
...
-  }
+.hds-table__th-content {
+  display: flex;
+  gap: 8px;
+  align-items: center;
+  justify-content: flex-start;
 }
 
-// TABLE DENSITY (TBODY ROW HEIGHT)
+.hds-table__th-button {
```
alex-ju, line 114: not too familiar with the style of this button, but wanted to check if we considered using our button mixins

alex-ju (commented):
Looks good overall. Left a few comments/questions for your consideration.

packages/components/addon/components/hds/table/th-button-sort.js
didoo: good point. done in https://github.com/hashicorp/design-system/pull/1860/commits/4f3bb289ae280f0ac2cfbd9c95e2309cd5244322

packages/components/app/styles/components/table.scss
didoo: @alex-ju good thinking! I tried but there are some differences, so I would have to write a bunch of overrides, and at that point  I am not sure it would be worth it: what do you think?
screenshot_3366

alex-ju: I just compared the definitions in the mixin and the ones for this button and there's not as much of an overlap. we can always use other opportunities to do such refactoring with no impact on consumers

packages/components/addon/components/hds/table/th-button-sort.hbs
didoo: @MelSumner thoughts? I had a [thumbs up](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1702049406587669?thread_ts=1702034833.897939&cid=C025N5V4PFZ) on this 🤔

One consideration: the information of the column's name/label is available only when the consumers provide the data via `@columns`, while when they use directly the `Table::ThSort` the content is yielded so we don't have access to it.

.changeset/modern-zebras-destroy.md
didoo: I have mixed feelings about this. On one side yes, in the past we considered it major, but it's also true that these are not public APIs (plus, we have only one usage of these classes in the consumers codebases, in Atlas for a test selector).

@hashicorp/hds-engineering @Dhaulagiri thoughts?

----

For context, searching for the classnames usages here's what I've found:

**`hds-table__th-sort--button-content`** 
- used on the "web" repo (but not consumed, just replicating our component in React) https://github.com/search?q=%22hds-table__th-sort--button-content%22++user%3Ahashicorp+repo%3Ahashicorp%2Fweb&type=code&ref=advsearch

**`hds-table__th-sort`**
- used in Atlas for tests selectors: https://github.com/search?q=%22hds-table__th-sort%22++user%3Ahashicorp+repo%3Ahashicorp%2Fatlas&type=code&ref=advsearch
- used on the "web" repo (but not consumed, just replicating our component in React) - https://github.com/search?q=%22hds-table__th-sort%22++user%3Ahashicorp+repo%3Ahashicorp%2Fweb&type=code&ref=advsearch

**`hds-table__th--text-***`**
- used on the "web" repo (but not consumed, just replicating our component in React) - https://github.com/search?q=%22hds-table__th--text-%22++user%3Ahashicorp+repo%3Ahashicorp%2Fweb&type=code&ref=advsearch

**`hds-table__td--text-***`**
- used on the "web" repo (but not consumed, just replicating our component in React) - https://github.com/search?q=%22hds-table__td--text-%22++user%3Ahashicorp+repo%3Ahashicorp%2Fweb&type=code&ref=advsearch

packages/components/addon/components/hds/table/th-button-sort.hbs
alex-ju: the consumer can, in theory, to provide their own `aria-label`, so an alternative would be to address this through guidance

didoo: > the consumer can, in theory, provide their own aria-label

if I understand correctly, this would mean that the consumer would have to provide an extra `aria-label` when used without `@columns` but not when used with `@columns`. I think having two different APIs for the two ways of using the table is not great DX. Can we maybe think of an alternative? (we can always ask consumers to add an `aria-label` in both cases, but this would mean that all the existing instances of the `Hds::Table` with sorting would need to be refactored (so, this would be a breaking change to release).

MelSumner (approved):
I think this should be a major because there are some cases where the consumer may have needed to use some of the CSS classes for custom tables, and we have treated similar situations as majors in the past. 

I do not see any accessibility issues with this approach at this time, and do not object to this moving forward.

didoo:
> I think this should be a major because there are some cases where the consumer may have needed to use some of the CSS classes for custom tables, and we have treated similar situations as majors in the past.

@MelSumner even if I have checked and there are no such cases? see list here: https://github.com/hashicorp/design-system/pull/1860#discussion_r1430210661

.changeset/modern-zebras-destroy.md
KristinLBradley: Consumers shouldn't directly be using hds component classnames (although I know there's nothing currently preventing them). Ideally we would be scoping class names though. If we've searched and confirmed that the classnames are not used in a way that would cause issues, then I wouldn't consider class name changes as major.

didoo:
@MelSumner @alex-ju @KristinLBradley as [agreed](https://docs.google.com/document/d/1xXxMv70KCOEmAe_l8mfzrnbQ3EGrw3v6Qt107RPyWuA/edit), I've updated the current implementation of the "sort" and "tooltip" functionalities to support distinct `aria-label` values (and potentially custom values too). See commit here: https://github.com/hashicorp/design-system/pull/1860/commits/2b6a470004ba72d25a595606a25f859c9c70b71f

That said, I would like us to consider a simplification/automation (otherwise all the burden is on the consumers): to use `aria-labelledby` to compose the aria label string, by pointing to a hidden element that contains a special string (eg. "sort by" or "more information for") inside the button, and the value of the text inside the `<td>` element (we use a `<span class="hds-typography-body-200 hds-font-weight-semibold">` element to contain it). In this way the aria label would read `sort by [column name]` and it would be transparent (and automatic) for the consumer.

What do you think? If you want we can even discuss this in a sync meeting, if that's easier.

.changeset/modern-zebras-destroy.md
didoo: Closing this conversation, since we've decided to consider this a `major` version release.

packages/components/addon/components/hds/table/th.hbs
```diff
@@ -6,6 +6,15 @@
...
-    {{yield}}
+    {{#if @tooltip}}
+      <div class="hds-table__th-content">
+        {{#let (unique-id) as |labelId|}}
+          <span id={{labelId}} class="hds-typography-body-200 hds-font-weight-semibold">{{yield}}</span>
+          <Hds::Table::ThButtonTooltip @tooltip={{@tooltip}} @labelId={{labelId}} />
+        {{/let}}
+      </div>
+    {{else}}
+      <span class="hds-typography-body-200 hds-font-weight-semibold">{{yield}}</span>
```
didoo, line 17: If we want, we can remove this duplication by moving the `{{#if @tooltip}}` check just before the `<Hds::Table::ThButtonTooltip>` element (but in this case the table cell content will always have a  `<div class="hds-table__th-content">` wrapper. WDYT?

didoo:
@alex-ju @MelSumner @KristinLBradley I have updated the "sort" and "tooltip" implementation to include aria labelling as agreed, please re-review

packages/components/addon/components/hds/table/th.hbs
alex-ju: I don't mind having it as it is, the duplication is minimal and keeps the mark-up to a minimum.

packages/components/app/styles/components/table.scss
```diff
@@ -19,175 +19,167 @@ $hds-table-cell-padding-medium: 14px 16px 13px 16px; // the 1px difference is to
...
-    // BORDER RADIUS: TARGET FIRST AND LAST TH ELEMENTS IN THE ROW
-    &:first-of-type {
-      th:first-child {
-        border-top-left-radius: $hds-table-inner-border-radius;
-      }
 
-      // LAST TH IN THE ROW, BORDER RADIUS
-      th:last-child {
-        border-top-right-radius: $hds-table-inner-border-radius;
+    th {
```
alex-ju, line 49: Is there a reason for not using a class here (e.g `hds-table__td`)? I know this is not introduced in this PR, but we may want to consider relying on classes as much as possible.

alex-ju (approved):
Looks very solid overall! Initially left a comment about `unique-id`, but it seems like you've already addressed that.

Regarding the codemod for `.gts` – not yet convinced of the cross-product value of such codemod, but we can chat more about this separately.

packages/components/app/styles/components/table.scss
alex-ju: Trying to think if we're ever in a position where the class may not be applied 🤔

didoo: that's how it was implemented initially, but that's a good point. I've refactored it in https://github.com/hashicorp/design-system/pull/1860/commits/c8a56788e14895fb05f9680a0e67bd59d5a2778d

MelSumner:
CleanShot 2024-01-18 at 12 30 01@2x
This sort button on this example doesn't seem to be functional.

MelSumner:
I think this might be orthogonal to this PR, but numbers in tables should be right-aligned. 
CleanShot 2024-01-18 at 12 32 06@2x

MelSumner (approved):
There are a few improvements to make it perfect, but they are non-blocking.

didoo:
> CleanShot 2024-01-18 at 12 30 01@2x This sort button on this example doesn't seem to be functional.

@MelSumner I've added a note in the page to clarify that in that example the "sort" button doesn't work, is only for presentational reasons.

> I think this might be orthogonal to this PR, but numbers in tables should be right-aligned. CleanShot 2024-01-18 at 12 32 06@2x

Fixed in https://github.com/hashicorp/design-system/pull/1860/commits/89ce92d4aecefcd830c2d8b275ffa534c7076e44

> There are a few improvements to make it perfect, but they are non-blocking.

👍

KristinLBradley (approved):
Looks good to me!

packages/codemods/README.md
```diff
@@ -28,6 +28,11 @@ node ./packages/codemods/bin/cli.js v3/dropdown path/to/some/glob/**/*.hbs
 * [v3/side-nav](transforms/v3/side-nav/README.md)
 <!--TRANSFORMS_END-->
 
+### v4
+<!--TRANSFORMS_START-->
```
didoo, line 32: @alex-ju what happens to these automatic updates?

alex-ju: when I generate a new transform it will get appended to this list – giving a way for people to discover them through our readme docs. not sure this answers your question though

didoo: ok, so it's only on generation? in that case I should remove the start/end from the v3 block?

alex-ju: yes, only on generation; we can remove it from the v3 block. it may still work but not sure how it behaves if there are two blocks (it may pick the latest), but safer to remove it from the old block

PR #1869, opened 2023-12-14
`Tooltip` - Fix text alignment

didoo:
### :pushpin: Summary

While working on the implementation of tooltips in `Table` (https://github.com/hashicorp/design-system/pull/1860) I've realized that the text in the tooltip "bubble" was inheriting its alignment from the table cell:

image

We didn't notice previously because the example/test in the showcase has a too short text, which was not spanning on multiple lines, so the alignment was not visible.

In this PR I have:
- forced left alignment of text in tooltip “bubble”
- updated the alignment example in the showcase

👉 👉 👉 **Preview**: https://hds-showcase-git-tooltip-fix-text-alignment-hashicorp.vercel.app/components/tooltip

**Before:**
screenshot_3357

**After:**
screenshot_3358

***

### 👀 Component checklist

### DIFF
--- a/.changeset/five-buses-burn.md
+++ b/.changeset/five-buses-burn.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Tooltip` - Fixed issue with text alignment, which was inherited from the parent container (now it's always left aligned).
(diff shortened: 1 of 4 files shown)


KristinLBradley (approved):

heatherlarsen (approved):
:shipit:

PR #1871, opened 2023-12-14
`Table` - New sorting and tooltip documentation update

didoo:
### :pushpin: Summary

This is a follow-up of https://github.com/hashicorp/design-system/pull/1860 in which we update the website documentation

In this PR we have:
- updated “Component API” with `@tooltip` argument + renaming of `onClick` to `onClickSort` in the `ThSort` element
- updated “How to use” section by adding an example of tooltip implementation
- bulk updated images to reflect new component design
- updated guidelines, specs, and accessibility content around tooltips and their usage within a table header

 👉👉 👉 **Preview**: https://hds-website-git-table-new-sorting-and-tooltip-docs-hashicorp.vercel.app/components/table

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2973

***

### 👀 Component checklist


jorytindall:
I've updated some of this content around the guidelines to include notes about tooltips. I actually struggled to come up with some cut and dry scenarios when you would want to include a tooltip, if anybody has any ideas there I'm all ears.

Otherwise, I've also updated a bunch of imagery to match the new design. I'll add a follow-up commit to catch some of the images elsewhere, they're a bit spread.

website/docs/components/table/partials/accessibility/accessibility.md
```diff
@@ -7,7 +7,7 @@ When used as recommended, there should not be any WCAG conformance issues with t
 
 ## Focus in Tables
 
-- Focus will only move through sortable headers and will skip over non-sortable headers as they aren't interactive. 
+- Table headers and labels are not eligible to recieve focus, rather, focus will move through interactive elements (sort and tooltip buttons) contained within the header sequentially.
```
KristinLBradley, line 10:
```suggestion
- Table headers and labels are not eligible to receive focus, rather, focus will move through interactive elements (sort and tooltip buttons) contained within the header sequentially.
```

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -454,6 +454,29 @@ To create a column that has right-aligned content, set `@align` to `right` on bo
 </Hds::Table>
 ```
 
+### Tooltip
+
+Whenever possible [table headers](/table#headers) should be clear, concise, and straightforward. But there could be cases in which the label is not sufficient by itself and extra information is required. In this case it's possible to show a tooltip next to the label in the header:
```
heatherlarsen, line 459: Grammarly suggestions

```suggestion
[Table headers](/table#headers) should be clear, concise, and straightforward whenever possible. However, there could be cases where the label is insufficient by itself, and extra information is required. In this case, it's possible to show a tooltip next to the label in the header:
```

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -20,6 +20,24 @@
 - Columns that do contain a sortable data type are interactive and therefore have corresponding hover, active, and focus states.
 - A Table may only be sorted by a single value at a time.
 
+### Tooltips
+
+Labels within the Table column should be clear, concise, and straightforward. In the case that more context or details are necessary a [Tooltip](/components/tooltip) can be used in conjunction with the label, but should be used sparingly and as a last resort.
```
heatherlarsen, line 25: Grammarly suggestions 

```suggestion
Labels within the Table column should be clear, concise, and straightforward. In the case that more context or details are necessary, a [Tooltip](/components/tooltip) can be used in conjunction with the label but should be used sparingly and as a last resort.
```

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -20,6 +20,24 @@
...
+
+![Tooltips in a Header Column](/assets/components/table/table-tooltip-example.png)
+
+!!! Dont
+
+We recommend against using a tooltip in all or most columns in a table as this can add unnecessary visual clutter and increase the cognitive load on the user.
```
heatherlarsen, line 31: Is there an alternative we should suggest? Something to guide them to a different solution/help them avoid this scenario?

heatherlarsen (commented):
This is looking good! I left a few comments and noticed an interesting use case that I'm not sure has been discussed yet, so I wanted to call attention to it. I found this in the Code tab under the heading "More examples".
Screenshot 2023-12-18 at 4 54 52 PM

cc @jorytindall

website/docs/components/table/partials/guidelines/guidelines.md
jorytindall: This is the part that I'm struggling with a bit. A couple of things I've considered (a bit multi-faceted):
- Providing more context around/associated with the table. This could be descriptive text (maybe even in the PageHeader), or use the flyout as a vehicle for communicating more complex labels/concepts
- There's probably an enhancement that can be made in the flow prior to the user arriving at a table that might make this content more clear.
- Adding a sort of key or legend element associated with the table. This could also potentially exist within a flyout.

heatherlarsen: Hmm, yea, it seems there are some different alternative options depending on the use case. We can always address it later if we decide on more standardized guidance in the future.

KristinLBradley:
> This is looking good! I left a few comments and noticed an interesting use case that I'm not sure has been discussed yet, so I wanted to call attention to it. I found this in the Code tab under the heading "More examples". Screenshot 2023-12-18 at 4 54 52 PM
> 
> cc @jorytindall

Interesting case. Since the design was changed to keep the sorting icon button next to the table header text instead of aligning it to the right side of the table header cell, are the in-between borders really still needed?

jorytindall:
Regarding the scenario where there is a visually hidden label in the table header, I think this solution makes the most sense to solve the apparent white/empty space:

- The border was initially on the right side of the component, changing it to the left would potentially solve by then:
- Allowing us to conditionally render the border if `isVisuallyHidden` is passed to the component (am I correct in that assumption?)

I've made this change in Figma already, it was pretty small/quick, and didn't introduce any problematic changes. If this isn't the route we want to go I can change it back.

@KristinLBradley I do think it's valuable to have the visual differentiation between columns that the border provides. This change also received positive feedback from all of our stakeholders, so I think it makes sense to leave it in. We can assess feedback if it comes in specifically related to this visual element.

heatherlarsen (approved):
Documentation updates all look good to me 🚀

didoo:
@KristinLBradley (+ @alex-ju / @MelSumner in case) the PR is ready for re-review

website/docs/components/table/partials/code/component-api.md
```diff
@@ -59,6 +59,9 @@ The Table component itself is where most of the options will be applied. However
       <C.Property @name="sortingFunction" @type="function">
         Callback function to provide support for custom sorting logic. It should implement a typical bubble-sorting algorithm using two elements and comparing them. For more details, see the example of custom sorting in the How To Use section.
       </C.Property>
+      <C.Property @name="tooltip" @type="string">
+        Text string which will appear in the tooltip (see [`Tooltip`](/components/tooltip) for details). May contain basic HTML tags for formatting text such as `strong` and `em` tags. Not intended for multi-paragraph text or other more complex content. May not contain interactive content such as links or buttons. The `placement` and `offset` are automatically set and can't be overwritten.
```
KristinLBradley, line 63:
```suggestion
        Text string which will appear in the tooltip (see [`Tooltip`](/components/tooltip) for details). May contain basic HTML tags for formatting text such as `strong` and `em` tags. Not intended for multi-paragraph text or other more complex content. May not contain interactive content such as links or buttons. The `placement` and `offset` are automatically set and can’t be overwritten.
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -127,6 +130,9 @@ If the `Th` component is passed as the first cell of a table body row, `scope="r
   <C.Property @name="width" @type="string" @valueNote="Any valid CSS">
     If set, determines the column’s width.
   </C.Property>
+  <C.Property @name="tooltip" @type="string">
+    Text string which will appear in the tooltip (see [`Tooltip`](/components/tooltip) for details). May contain basic HTML tags for formatting text such as `strong` and `em` tags. Not intended for multi-paragraph text or other more complex content. May not contain interactive content such as links or buttons. The `placement` and `offset` are automatically set and can't be overwritten.
```
KristinLBradley, line 134:
```suggestion
    Text string which will appear in the tooltip (see [`Tooltip`](/components/tooltip) for details). May contain basic HTML tags for formatting text such as `strong` and `em` tags. Not intended for multi-paragraph text or other more complex content. May not contain interactive content such as links or buttons. The `placement` and `offset` are automatically set and can’t be overwritten.
```

website/docs/components/table/partials/code/component-api.md
```diff
@@ -152,7 +158,10 @@ This is the component that supports column sorting; use instead of `Hds::Table::
   <C.Property @name="width" @type="string" @valueNote="Any valid CSS">
     If set, determines the column’s width.
   </C.Property>
-  <C.Property @name="onClick" @type="function">
+  <C.Property @name="tooltip" @type="string">
+    Text string which will appear in the tooltip (see [`Tooltip`](/components/tooltip) for details). May contain basic HTML tags for formatting text such as `strong` and `em` tags. Not intended for multi-paragraph text or other more complex content. May not contain interactive content such as links or buttons. The `placement` and `offset` are automatically set and can't be overwritten.
```
KristinLBradley, line 162:
```suggestion
    Text string which will appear in the tooltip (see [`Tooltip`](/components/tooltip) for details). May contain basic HTML tags for formatting text such as `strong` and `em` tags. Not intended for multi-paragraph text or other more complex content. May not contain interactive content such as links or buttons. The `placement` and `offset` are automatically set and can’t be overwritten.
```

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -454,6 +454,29 @@ To create a column that has right-aligned content, set `@align` to `right` on bo
 </Hds::Table>
 ```
 
+### Tooltip
+
+[Table headers](/table#headers) should be clear, concise, and straightforward whenever possible. However, there could be cases where the label is insufficient by itself, and extra information is required. In this case, it's possible to show a tooltip next to the label in the header:
```
KristinLBradley, line 459:
```suggestion
[Table headers](/table#headers) should be clear, concise, and straightforward whenever possible. However, there could be cases where the label is insufficient by itself and extra information is required. In this case, it’s possible to show a tooltip next to the label in the header:
```

website/docs/components/table/partials/code/how-to-use.md
```diff
@@ -454,6 +454,29 @@ To create a column that has right-aligned content, set `@align` to `right` on bo
...
+
+```handlebars
+<Hds::Table
+  @model={{this.model.myDemoData}}
+  @columns={{array
+    (hash key="artist" label="Artist")
+    (hash key="album" label="Album" tooltip="Title of the album (in its first release)")
+    (hash key="vinyl-cost" label="Vinil Cost (USD)" isSortable=true tooltip="Cost of the vinyl (adjusted for inflation)" align="right")
```
KristinLBradley, line 467:
```suggestion
    (hash key="vinyl-cost" label="Vinyl Cost (USD)" isSortable=true tooltip="Cost of the vinyl (adjusted for inflation)" align="right")
```

KristinLBradley (approved):

PR #1872, opened 2023-12-14
Convert ember packages to v2 addon format

alex-ju:
### :pushpin: Summary

In this PR we are migrating the ember addons (`packages/components` and `packages/flight-icons`) to v2 format.

We follow [the official migration guidance](https://github.com/embroider-build/embroider/blob/main/docs/porting-addons-to-v2.md) adapted to the particularities of this monorepo.

The implementation was tested in `cloud-ui`, `consul-ui-toolkit`, and `atlas` with a sibling branch named `alex-ju/v2-addon-test-prepublish-3` containing the `dist` folders for each package (that will eventually be published to npm).

This means, at a high level, for `packages/components`:
 - separating the addon from dummy app
   - the addon is preserved in `packages/components` with the addon code itself located in `src`
   - the dummy app is renamed to `showcase` and moved outside `packages`, at the root of the monorepo
 - adding `@embroider/addon-shim` as a dependency
 - setting up the build pipeline (via Babel) for template colocation and decorator support
 - setting up rollup for compilation and module bundling
 - update `website` config to include sass paths to components' styles
 - update blueprints to match new format and structure
 - update vercel config (currently via config file, the commands will eventually be updated in vercel)

Similar steps were taken for `packages/ember-flight-icons` with a note that the test app (not in use) was completely removed as we test the `<FlightIcon>` component indirectly in `showcase`.

While the granular commits were preserved for better traceability, note there is a bit of back-and-forth, especially in configuration files that would be easier to review in their final state.

**Build-related notes:**
- `hds-showcase` – the build command would need to be updated to `yarn workspace @hashicorp/ember-flight-icons build && yarn workspace @hashicorp/design-system-components build && yarn workspace showcase build` (the serve command would need a similar update)
- `hds-website` – the build command would need to account for internal dependencies (e.g. components) being built; the updated command would be `yarn workspace @hashicorp/ember-flight-icons build && yarn workspace @hashicorp/design-system-components build && yarn workspace website build && yarn workspace website algolia:index`

**Testing-related notes:**
 - you can build the packages by running `yarn workspace @hashicorp/ember-flight-icons build` and `yarn workspace @hashicorp/design-system-components build` in the root directory (this must be done at least once before trying to run the showcase or website)
 - you can test the components on your machine via `showcase` using the standard commands (`yarn workspace showcase ember s` and/or `yarn workspace showcase ember t`)
 - you can run the `website` on your machine using the standard commands (`yarn workspace website ember s`)

### 🧑‍💻 Impact on consumers

To migrate update Sass configuration in `ember-cli-build.js` to include the paths for `ember-flight-icons` and `design-system-components`:

```js
sassOptions: {
  precision: 4,
  includePaths: [
    './node_modules/@hashicorp/design-system-tokens/dist/products/css',
    './node_modules/@hashicorp/ember-flight-icons/dist/styles'
    './node_modules/@hashicorp/design-system-components/dist/styles',
  ],
},
```

Alternatively, you can import the CSS by adding this configuration in `ember-cli-build.js`.

```js
app.import('node_modules/@hashicorp/design-system-components/dist/styles/@hashicorp/design-system-components.css');
```

### :link: External links

Jira ticket: [HDS-2870](https://hashicorp.atlassian.net/browse/HDS-2870), [HDS-2872](https://hashicorp.atlassian.net/browse/HDS-2872)

***

### 👀 Component checklist

- [x] If documenting a new component, an acceptance test that includes the `a11yAudit` has been added

[HDS-2875]: https://hashicorp.atlassian.net/browse/HDS-2875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HDS-2870]: https://hashicorp.atlassian.net/browse/HDS-2870?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/package.json
```diff
@@ -67,6 +67,7 @@
     "@types/ember-resolver": "^9.0.0",
     "@types/qunit": "^2.19.7",
     "@types/rsvp": "^4.0.6",
+    "ember-source": "~5.4.0",
```
aklkv, line 70: you would probably want to keep this in peerDeps

alex-ju:
@hashicorp/hds-engineering I would be keen to get your feedback on this PR to get your thoughts on the proposed conversion to v2. I am mainly looking to understand if there are any outstanding questions and/or alternative paths you would like us to explore before re-doing this work on the most recent version of `main`.

Conscious this stream of work is meant to be wrapped up by the end of this quarter (end of January), I plan to give an overview of this work during our sync on January 3rd and would appreciate your feedback by January 5th.

packages/components/package.json
```diff
@@ -16,46 +16,36 @@
...
+    "@embroider/addon-shim": "^1.8.7",
     "@hashicorp/design-system-tokens": "^1.9.0",
     "@hashicorp/ember-flight-icons": "^4.0.4",
     "dialog-polyfill": "^0.5.6",
     "ember-a11y-refocus": "^3.0.2",
-    "ember-auto-import": "^2.6.3",
     "ember-cached-decorator-polyfill": "^1.0.2",
-    "ember-cli-babel": "^8.2.0",
-    "ember-cli-htmlbars": "^6.3.0",
     "ember-cli-sass": "^11.0.1",
```
aklkv, line 43: I am pretty sure this addon has no effect in v2 format as we take care of sass compilation with rollup pipeline

alex-ju: Good call, yes it should be safe to remove it. Will amend

showcase/.npmignore
didoo, line 1: do we need a `.npmignore` file if the folder is not published to NPM? 🤔

showcase/package.json
```diff
@@ -0,0 +1,107 @@
...
+  "description": "Components showcase",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/hashicorp/design-system.git",
+    "directory": "showcase"
+  },
+  "license": "MPL-2.0",
+  "author": "HashiCorp Design Systems <design-systems@hashicorp.com>",
+  "directories": {
+    "doc": "doc",
```
didoo, line 14: Is this still needed? 🤔
```suggestion
```

website/ember-cli-build.js
```diff
@@ -14,6 +14,7 @@ const appConfig = {
   sassOptions: {
     precision: 4,
     includePaths: [
+      '../node_modules/@hashicorp/design-system-components/dist/styles',
       '../node_modules/@hashicorp/design-system-tokens/dist/products/css',
```
didoo, line 18:
```suggestion
      '../node_modules/@hashicorp/design-system-tokens/dist/products/css',
      '../node_modules/@hashicorp/design-system-components/dist/styles',
```

didoo (approved):
All good for me. Left only a couple of comments/suggestions.

showcase/.npmignore
alex-ju: no, we don't – good catch, will remove it.

showcase/package.json
alex-ju: nope, not needed anymore

alex-ju:
> One consideration, though: in the PR description you said _"This implementation was tested in cloud-ui"_: would it be worth it to test it also with other products (expecially those who have special setup or use older versions of Ember?)

Indeed, I'm currently in the process of testing with other products (such as TFC, Vault and if time allows others).

aklkv:
A couple of items might be good ideas to consider as part of this work:
* ship compiled sass, this will remove the need to process on the consumer side less work during build tile = faster build
* turn on embroider build on showcase up, this will prevent any build issues with consumers that use embroider + faster build for HDS

Side note, both of these items are implemented in PR  and requires very minimal configurations

alex-ju:
> A couple of items might be good ideas to consider as part of this work:
> 
> * ship compiled sass, this will remove the need to process on the consumer side less work during build tile = faster build
> * turn on embroider build on showcase up, this will prevent any build issues with consumers that use embroider + faster build for HDS
> 
> Side note, both of these items are implemented in PR and requires very minimal configurations

Thanks for the feedback! Shipping the compiled sass is something we plan on doing. However, we'll likely still ship sass as well, at least for a while to smooth the migration path and allow consumers to choose whichever method works best for them. Turning embroider build on in `showcase` (and eventually the documentation `website`) is also in the backlog (thanks for sharing the configuration in your spike PR) but we might do it in a follow-up PR. We also plan to turn on `embroider-safe` in our `ember-try` matrix.

aklkv:
Just to follow up on sass compilation point, my configuration ships both variants so that consumers can choose what is best for them. But if this is planned as follow up item I am cool with that

alex-ju:
> Just to follow up on sass compilation point, my configuration ships both variants so that consumers can choose what is best for them. But if this is planned as follow up item I am cool with that

Indeed, in your spike there are both versions exposed. That's what we're planning as well. I can try and make it part of this work – but will double-check with the team first.

packages/ember-flight-icons/package.json
```diff
@@ -16,75 +16,69 @@
...
-    "start": "ember serve",
-    "test": "npm-run-all lint test:*",
-    "test:ember": "ember test",
-    "test:ember-compatibility": "ember try:each"
+    "lint:js:fix": "eslint . --fix"
   },
   "dependencies": {
+    "@embroider/addon-shim": "^1.8.7",
     "@hashicorp/flight-icons": "^2.24.0",
-    "ember-auto-import": "^2.6.3",
```
Dhaulagiri, line 38: i'm _pretty_ sure this is/was necessary for import logic within the addon to work correctly? or has v2 format changed things?

alex-ju: the migration guidance [is suggesting we remove it](https://github.com/embroider-build/embroider/blob/main/docs/porting-addons-to-v2.md#part-4-convert-addon-to-v2) – happy to leave it there if we still need

alex-ju:
Rebased again after #1898, as the new version of yarn generates new hashes 😬
Screenshot 2024-01-22 at 21 11 20

packages/ember-flight-icons/package.json
Dhaulagiri: i will trust the guidance then 😄 . Just wanted to double check as testing this was pretty tricky if i recalled from past work on this.

showcase/package.json
```diff
@@ -0,0 +1,107 @@
...
+    "@percy/ember": "^4.2.0",
+    "broccoli-asset-rev": "^3.0.0",
+    "ember-a11y-testing": "^6.1.1",
+    "ember-auto-import": "^2.6.3",
+    "ember-body-class": "^3.0.0",
+    "ember-cli": "~5.3.0",
+    "ember-cli-babel": "^8.2.0",
+    "ember-cli-clean-css": "^3.0.0",
+    "ember-cli-dependency-checker": "^3.3.2",
+    "ember-cli-deprecation-workflow": "^2.1.0",
```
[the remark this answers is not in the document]
Dhaulagiri: it may or may not matter though given the deprecations we are currently guarding against are not very current/likely to ever show up

Dhaulagiri (commented):
This was amazing to read through and learn from 👏 . I really appreciate the organized approach to the work and setting up the review.

My review was fairly superficial given my limited experience with v2 addons, but things do seem to work as expected. I tried a few things
* Generating new components/docs
* Running the showcase & website
* Poking around the code in VSCode

The one caveat, and I believe this is expected behavior, is that changes to, say, a component don't show up in the showcase unless I re-build the component package manually. Do I have that right? I'd be curious how folks feel about this potential change to the workflow.

We already talked about separating out the cached decorator change, but I would lightly  suggest trying to do the same with other changes that cropped up like 67d42e0d7e6da585c619d4f88b988de1e92012e3 (within reason if it doesn't introduce to many rebase headaches).

packages/components/rollup.config.mjs
```diff
@@ -0,0 +1,78 @@
...
+
+  plugins: [
+    // These are the modules that users should be able to import from your
+    // addon. Anything not listed here may get optimized away.
+    addon.publicEntrypoints(['**/*.ts', '**/*.js', 'styles/@hashicorp/*.scss']),
+
+    // These are the modules that should get reexported into the traditional
+    // "app" tree. Things in here should also be in publicEntrypoints above, but
+    // not everything in publicEntrypoints necessarily needs to go here.
+    addon.appReexports(['**/*.ts', '**/*.js']),
```
meirish, line 24: I think you may want to be more specific here - this is essentially replacing having an `app` folder. Before it was only the contents of the  `components/`, `helpers/` and `modifiers/` folders, but this is also now re-exporting things like `utils` which is causing a warning when I'm trying it locally:

```
WARNING in ../../../../node_modules/@hashicorp/design-system-components/dist/_app_/utils/hds-get-element-id.js 1:0-87
export 'default' (reexported as 'default') was not found in '@hashicorp/design-system-components/utils/hds-get-element-id' (possible exports: getElementId)
```

.github/workflows/ci-components.yml
```diff
@@ -26,7 +26,6 @@ jobs:
       - name: Get changed files
         id: read-files
         run: ./.github/scripts/filter_changed_files.sh "packages/components" "packages/ember-flight-icons" "packages/flight-icons/catalog.json"
```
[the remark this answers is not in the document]
alex-ju: I thought that we wanted to only run this script for actual package-related changes, but changing the showcase is still relevant for visual regression testing.

showcase/package.json
```diff
@@ -0,0 +1,107 @@
...
+    "build": "ember build --environment=production",
+    "lint": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"",
+    "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
+    "lint:css": "stylelint \"app/styles/**/*.scss\"",
+    "lint:css:fix": "npm-run-all \"lint:css --fix\"",
+    "lint:hbs": "ember-template-lint .",
+    "lint:hbs:fix": "ember-template-lint . --fix",
+    "lint:js": "eslint . --cache",
+    "lint:js:fix": "eslint . --fix",
+    "start": "ember serve",
```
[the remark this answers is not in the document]
alex-ju: Good point! We can run rollup with a watcher on files, and the `start` command in each package does just that (`rollup --config --watch`).

So if we update the start command in `showcase` to be
```suggestion
    "start": "yarn workspace @hashicorp/ember-flight-icons start && yarn workspace @hashicorp/design-system-components start && ember serve",
```
we would be able to operate changes in `packages/*` and preview the results in `showcase` without having to rebuild

showcase/package.json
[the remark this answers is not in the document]
alex-ju: fair, I can check this up. we do however have it in `website` too, so _maybe_ leaving it in `showcase` too 🤔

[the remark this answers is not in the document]
Dhaulagiri: Thinking about this more  ast night I think what you have here is correct as the workflow should be adjacent to where the tests live, which in this case in the showcase.

showcase/package.json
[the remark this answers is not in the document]
alex-ju: hm.. my suggestion above is not possible in a single command, as the `--watch` threads will need to remain active blocking the execution of ember serve. but I will make sure to re-build every time the showcase is started

.github/workflows/ci-components.yml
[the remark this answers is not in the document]
alex-ju: Updated in 1e6b38e to include showcase

packages/components/rollup.config.mjs
alex-ju: updated in 875ac4a to only reexport the main 3 folders

PR #1874, opened 2023-12-18
Table Mutli-selection documentation

andgen404:
_**Note: This branch has been duplicated and renamed. New PR is here: https://github.com/hashicorp/design-system/pull/1881**_

---

### :pushpin: Summary

This PR will add documentation for the `Table` Multi-Select component if merged.

**Web preview:** https://hds-website-git-table-multi-select-documentation-hashicorp.vercel.app/components/table

### 🛠️ Detailed description

In this PR we have:
- Updated documentation around `Header column / select` and `Cell / select` 
- Anatomy section updated with `mutli-select` components 
- Images added and video interactions  

### :link: External links

Jira ticket: [HDS-2293](https://hashicorp.atlassian.net/browse/HDS-2293)
***

### 👀 Component checklist

[HDS-2293]: https://hashicorp.atlassian.net/browse/HDS-2293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


website/docs/components/form/primitives/partials/guidelines/guidelines.md
```diff
@@ -22,7 +22,7 @@ Use a character count to communicate the current length of the value in an input
 
 Depending on which property (or properties) are passed to the component, a different default message will be displayed by the component to communicate the relationship between the current length of the input value (`currentLength`) and the maximum length (`maxLength`), minimum length (`minLength`), or both.
 
-<video width="100%" controls loop>
+<video width="75%" controls loop>
```
jorytindall, line 25: Did you mean to change this in the character count docs?

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,33 @@ Row placement determines the visual styling based on where the row is placed rel
 - We recommend using `medium` cell density by default.
 - If content is complex or a smaller data set (e.g., a Table of basic user data), `tall` cell density allows for more breathing room surrounding the content.
 - If content is largely string/text-based, `short` allows for more content to be displayed within the page.
-- While denser content allows for more rows to be displayed within a single page, it also makes comprehension and scanning more difficult.
\ No newline at end of file
+- While denser content allows for more rows to be displayed within a single page, it also makes comprehension and scanning more difficult.
+
+## Multi-Select
+
+Multi-select enables users to select multiple rows from a dataset, significantly enhancing the efficiency of batch operations like delete, move, or update. This functionality is important for effective data management, especially in large datasets.
```
jorytindall, line 162: Maybe a complex answer here, but I think we need to differentiate between rows, tables, results, and datasets. These have different contextual meanings that relate to each other, for example:

- A data set consists of many results
- A table consists of many rows
- A data set can be rendered within a table, the table is the UI representation of the data set. but it can also be rendered in other ways.

So, in this scenario I would maybe approach from this angle:

- "Multi-select enables users to select multiple rows in a table…" or
- "Multi-select enables users to select multiple results from a data set when rendered as a table…"

Not perfect, but I think there's a benefit in the clarity of terminology here.

jorytindall: Also, global note: in the filtering docs I intentionally used "data set" rather than "dataset". We should strive for consistency here.

jorytindall: To summarize, maybe consider this:

> Multi-select enables users to select multiple rows from a table, allowing for batch actions like delete or update.

The point about effective data management seems a bit broad/ambiguous, is there a way to more accurately express the end-user value?

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,33 @@ Row placement determines the visual styling based on where the row is placed rel
...
+
+!!! Info 
+
+Avoid using indeterminate states in individual rows. It's more effective at the global selection checkbox to reduce confusion.
```
jorytindall, line 166: I think you can be more pointed here and essentially say that indeterminate state isn't supported in individual rows. Think back to what we discussed about parent/child relationships here. The rows (in the majority of scenarios) don't have the concept of children, therefore they can only be considered on/off, checked/un-checked.

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,33 @@ Row placement determines the visual styling based on where the row is placed rel
...
+<video width="100%" controls loop>
+  <source
+    src="/assets/components/table/multi-selection-interaction-01.mp4"
+    type="video/mp4"
+  />
+</video>
+
+### Suggested Application
+
+- For global selection and actions we recommend a master checkbox at the top of the table to select all dataset items in one click, simplifying bulk operations. The interface should allow operations like select all (rows) or deselct all (rows), and delete, or move to be applied to all selected rows for quick and efficient control across multiple pages.
```
jorytindall, line 178: Couple of points here:

- This can be much more pointed and direct as well, shouldn't be worded as a recommendation because of how this feature works in the Ember component. The property that determines selection is `isSelectable`, if it's set to `{{true}}` at the top level of the table, the relevant UI elements will be rendered by default.
- This bullet point combines information about the intended UI, functionality, and how to handle bulk actions, which is pretty confusing. I think this should be decoupled a bit and each item addressed individually.

jorytindall (commented):
Added a couple of first-round comments, great first pass! Let's sync on a couple of these topics to see if we can streamline the narrative a bit.

website/docs/components/form/primitives/partials/guidelines/guidelines.md
andgen404: Oh no, mistake on my end

andgen404, line 25:
```suggestion
<video width="100%" controls loop>
```

website/docs/components/table/partials/guidelines/guidelines.md
andgen404, line 162:
```suggestion
Multi-select enables users to select multiple rows from a table. Users benefit from multi-select by experiencing enhanced efficiency and reduced errors, improving overall data management. This feature lets users quickly perform bulk actions on multiple rows, significantly increasing productivity.

```

website/docs/components/table/partials/guidelines/guidelines.md
andgen404, line 178:
```suggestion
### Suggested Application

Header Column /  Select is the master checkbox that renders a checked state or indeterminate state when Cell / Selection is selected. 
Cell / Selection is the component that enables multi-selection. It is always in the leading position of the row and can be rendered with {{isStriped}} or Medium, Short, and Tall densities

```

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,33 @@ Row placement determines the visual styling based on where the row is placed rel
...
+!!!
+
+<video width="100%" controls loop>
+  <source
+    src="/assets/components/table/multi-selection-interaction-01.mp4"
+    type="video/mp4"
+  />
+</video>
+
+### Suggested Application
```
andgen404, line 176:
```suggestion
### Intended interaction
```

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,33 @@ Row placement determines the visual styling based on where the row is placed rel
...
+
+- For global selection and actions we recommend a master checkbox at the top of the table to select all dataset items in one click, simplifying bulk operations. The interface should allow operations like select all (rows) or deselct all (rows), and delete, or move to be applied to all selected rows for quick and efficient control across multiple pages.
+- It is beneficial to display a dynamically updating count of selected rows. This feature provides a continuous and straightforward overview of user selections, adjusting in real-time as selections are made or changed.
+- When pagination is used in a table, maintain user selections as consistent and manageable across different pages.
```
andgen404, line 180:
```suggestion
Selecting a single row highlights it for actions like edit or delete.
Selecting the top-level header selects all rows on the current page.
Selecting the top-level header automatically selects all visible rows on the current page, including any rows previously chosen individually.
The 'Select all' usually affects only the current page in paginated tables. While in comparison, a global selection will select all rows on all pages (i.e., select all 53 users)
```

website/docs/components/table/partials/guidelines/guidelines.md
andgen404, line 166:
```suggestion
While you’re able to have the indeterminate state in the Cell / Select component, in the ember component, it is not achievable. 
```

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,47 @@ Row placement determines the visual styling based on where the row is placed rel
 - We recommend using `medium` cell density by default.
 - If content is complex or a smaller data set (e.g., a Table of basic user data), `tall` cell density allows for more breathing room surrounding the content.
 - If content is largely string/text-based, `short` allows for more content to be displayed within the page.
-- While denser content allows for more rows to be displayed within a single page, it also makes comprehension and scanning more difficult.
\ No newline at end of file
+- While denser content allows for more rows to be displayed within a single page, it also makes comprehension and scanning more difficult.
+
+## Multi-Selection
+
+Multi-selection enables users to select multiple rows in a table, enhancing efficiency and minimizing errors. This feature facilitates bulk operations, boosting productivity and optimizing data management.
```
jorytindall, line 162:
```suggestion
Multi-selection enables users to select multiple rows in a table, facilitating bulk operations, actions, and editing.
```

I still think this point around productivity and data management is a bit irrelevant (at least for this type of documentation). Though curious what other folks thing.

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,47 @@ Row placement determines the visual styling based on where the row is placed rel
...
+
+To enable multi-selection in a table:
```
jorytindall, line 164: I would pivot this to be more like:

> Multi-selection consists of:

Then list each component out.

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,47 @@ Row placement determines the visual styling based on where the row is placed rel
...
+
+1. `Header Column / Selection` Component:
+Position it in the table's header row. This acts as a master control, enabling you to either select all rows simultaneously or deselect them.
```
jorytindall, line 167:
```suggestion
Used in the header row of the Table. This acts as a master control, enabling end-users to either select or deselect all rows simultaneously.
```

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,47 @@ Row placement determines the visual styling based on where the row is placed rel
...
+
+
+![Example of centered content within a table](/assets/components/table/multi-select-header.png)
+
+2. `Cell / Selection` Component:
+Integrate this into each table row. It allows for individual row selection and aligns with the Header Column component, reflecting the overall selection status of the table.
```
jorytindall, line 173:
```suggestion
Used in the start position each Table row. It allows for individual row selection and aligns with the Header Column component, reflecting the overall selection status of the table.
```

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,47 @@ Row placement determines the visual styling based on where the row is placed rel
...
+
+![Example of centered content within a table](/assets/components/table/multi-select-cells.png)
+
+!!! Info 
+
+In the Ember component, unlike the `cell/select` figma component, achieving an indeterminate state is not possible.
```
jorytindall, line 179: Maybe consider something like this:

> While technically possible within the Figma component, the Checkbox within selected rows should not reflect an indeterminate state.

You could also include something about the child/parent relationship between the Header selection and the Cell selection.

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,47 @@ Row placement determines the visual styling based on where the row is placed rel
...
+<video width="100%" controls loop>
+  <source
+    src="/assets/components/table/multi-selection-interaction-01.mp4"
+    type="video/mp4"
+  />
+</video>
+
+### Intended interaction
+
+- Selecting an individual row highlights it, indicating it is ready for actions such as editing or deletion.
```
jorytindall, line 191:
```suggestion
- Selecting an individual row indicates it is ready for actions such as editing or deletion.
```

Using "highlights" is confusing (and probably loaded) term here.

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,47 @@ Row placement determines the visual styling based on where the row is placed rel
...
+- Activating the top-level header checkbox selects every row on the current page.
```
jorytindall, line 192:
```suggestion
- Interacting with the top-level header checkbox selects every row on the current page. 
```

Could also say "Checking" as that refers to the actual property.

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,47 @@ Row placement determines the visual styling based on where the row is placed rel
...
+    src="/assets/components/table/multi-selection-interaction-01.mp4"
+    type="video/mp4"
+  />
+</video>
+
+### Intended interaction
+
+- Selecting an individual row highlights it, indicating it is ready for actions such as editing or deletion.
+- Activating the top-level header checkbox selects every row on the current page. 
+     - This action ensures that all visible rows on the page, including those previously selected individually, are chosen collectively.
```
jorytindall, line 193: This reads a bit weird. Maybe consider discussing when the `Header Column / Selection` will:

- render as non-checked
- render as indeterminate
- render as checked

And how the selected rows are impacted by each one of these states.

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,47 @@ Row placement determines the visual styling based on where the row is placed rel
...
+- The 'Select all' function impacts only the rows displayed on the current page in paginated tables. In contrast, a global 'Select all' option encompasses every row across all pages (for example, selecting all 53 users in a dataset).
```
jorytindall, line 194: I wouldn't use the term "Select all" here because 1) there isn't a label present that expresses this concept (even if we mean to be coupled with the Header Column / Selection), and 2) because this implies that we also provide a "Select all" function with a paginated table, which we do not.

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,47 @@ Row placement determines the visual styling based on where the row is placed rel
...
+!!!
+
+<video width="100%" controls loop>
+  <source
+    src="/assets/components/table/multi-selection-interaction-01.mp4"
+    type="video/mp4"
+  />
+</video>
+
+### Intended interaction
```
jorytindall, line 189: I think this section might need a bit more fleshing out. Maybe consider breaking it down into a couple of different sections, could look something like this:

- Intended interaction
    - Within a simple table (this is where you can discuss what will happen with the selection states if the table is a single page/doesn't use pagination).
    - Within a paginated table
    - (maybe something else here?)

jorytindall (commented):
The anatomy needs to be updated as well, for both the header and the cell.

website/docs/components/table/partials/guidelines/guidelines.md
KristinLBradley, line 179:
```suggestion
In the Ember component, unlike the `cell/select` Figma component, achieving an indeterminate state is not possible. 
```

KristinLBradley: @andgen404 I don't quite understand what you mean here. Are you talking about having an indeterminate state on the checkboxes in the table body rows (vs. in the table header)? Because that doesn't make sense to do anyway. I would be very confused by this as a consumer.

website/docs/components/table/partials/guidelines/guidelines.md
KristinLBradley: I think "checking" would be much clearer.

website/docs/components/table/partials/guidelines/guidelines.md
KristinLBradley: Agree that this wording is confusing. Just say what it actually does, ie, it selects all visible rows on the page.

website/docs/components/table/partials/guidelines/guidelines.md
KristinLBradley, line 194:
```suggestion
- The “Select all” function impacts only the rows displayed on the current page in paginated tables. In contrast, a global “Select all” option encompasses every row across all pages (for example, selecting all 53 users in a dataset).
```

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,54 @@ Row placement determines the visual styling based on where the row is placed rel
...
+
+- Global Actions:
+In addition to page-specific actions, paginated tables may benefit from global actions like a "Select all users" feature. This would apply to all rows across every page in the table, enabling actions on a larger scale (e.g., selecting all 53 users in a dataset or deselecting). 
+
+<video width="100%" controls loop>
+  <source
+    src="/assets/components/table/multi-selection-interaction-02.mp4"
+    type="video/mp4"
+  />
+</video>
```
andgen404, line 208:
```suggestion
<!--TODO: Add in for future use of pattern documentation #### Within a paginated table: 

- Global Actions:
In addition to page-specific actions, paginated tables may benefit from global actions like a "Select all users" feature. This would apply to all rows across every page in the table, enabling actions on a larger scale (e.g., selecting all 53 users in a dataset or deselecting). 

<video width="100%" controls loop>
  <source
    src="/assets/components/table/multi-selection-interaction-02.mp4"
    type="video/mp4"
  />
</video>-->
```

website/docs/components/table/partials/guidelines/guidelines.md
```diff
@@ -155,4 +155,54 @@ Row placement determines the visual styling based on where the row is placed rel
...
+
+## Multi-Selection
+
+Multi-selection enables users to select multiple rows in a table, facilitating bulk operations, actions, and editing.
+
+Multi-selection consist of: 
+
+1. `Header Column / Selection` Component:
+Used in the header row of the Table. This acts as a master control, enabling end-users to either select or deselect all rows simultaneously.
+
```
andgen404, line 168:
```suggestion
```

heatherlarsen:
Closing for @andgen404 in favor of the duplicate #1881

PR #1882, opened 2023-12-28
Add missing `@algolia/client-search` dependency

alex-ju:
### :pushpin: Summary

Requested by `@algolia/autocomplete-js` and `@algolia/autocomplete-preset-algolia`. Currently raising warnings via yarn.


didoo (approved):
Approved 👍

(question: how could it work before? it was imported by another dependency?)

alex-ju:
> (question: how could it work before? it was imported by another dependency?)

yes, it's currently available via `@algolia/client-analytics`. I only noticed because yarn was complaining, so I thought we'd better make it explicit.

PR #1896, opened 2024-01-19
Change `Form::CharacterCount` to use a `value` argument.

meirish:
### :pushpin: Summary
This PR changes the signature of `Form::CharacterCount` so that it takes a `value` argument and uses that to derive the `currentLength` instead of using an input event handler and reading the input's value from the DOM.

There was a report of some laggy input in one of the places we were using the CharacterCount component, so I went exploring. We later found other issues that caused the lag, so I don't think things in this PR are the sole cause, but it did surface some anti-patterns in regards to what the component does. After chatting with @alex-ju, I decided to open this PR. 

Event handlers to elements you don't render in the component is something that should be avoided if possible - there's not a good way to track the other element's removal from the DOM, which would be necessary to remove the handler to prevent memory leaks.

This change also removes any need to access the DOM, which is generally slow - if a component can be made to render directly from data, that is usually the faster execution path. 

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist


alex-ju (commented):
Thank you for raising this, Matthew! And sorry for the delayed review.
As discussed, I see the limitations of the existing implementation and understand how it's not idiomatic.

My current concerns are around what the new API would look like, for the `CharacterCount` component, and what the developer experience would be.

In the current implementation, a simple call would be
```
<input type="text" aria-label="currentLength = 0" id="my-control" />
<Hds::Form::CharacterCount @maxLength={{25}} @controlId="my-control" />
```

With the proposed API, this simple call would become
```
<input type="text" value={{this.value}} aria-label="currentLength = 0" id={{controlId}} {{on "input" this.update}}/>
<Hds::Form::CharacterCount @value={{this.value}} @maxLength={{25}} @controlId={{controlId}} />
```
where the consumer would need to store the `value` and create a mechanism for updating this `value` `onInput` (`update(event) { this.value = event.target.value; }`).
I find this to be a considerable ask for our consumers.

@hashicorp/hds-engineering what are your thoughts on this proposal?

P.S. The mentioned rounding error in tests don't seem to be an issue in CI, but we have a few visual regressions. To resolve these we'd need a few examples updated in `packages/components/tests/dummy/app/templates/components/form/base-elements.hbs` (and other showcase examples for TextInput, Textarea, MaskedInput).

meirish:
> My current concerns are around what the new API would look like, for the `CharacterCount` component, and what the developer experience would be.
> 
> In the current implementation, a simple call would be
> 
> ```
> <input type="text" aria-label="currentLength = 0" id="my-control" />
> <Hds::Form::CharacterCount @maxLength={{25}} @controlId="my-control" />
> ```
> 
> With the proposed API, this simple call would become
> 
> ```
> <input type="text" value={{this.value}} aria-label="currentLength = 0" id={{controlId}} {{on "input" this.update}}/>
> <Hds::Form::CharacterCount @value={{this.value}} @maxLength={{25}} @controlId={{controlId}} />
> ```
> 
> where the consumer would need to store the `value` and create a mechanism for updating this `value` `onInput` (`update(event) { this.value = event.target.value; }`). I find this to be a considerable ask for our consumers.
> 

This is true, but it is also how you use any input that is not the two-way bound `Input` that is left over from older versions of Ember. This is already how every use of any of the text inputs works in applications - there is a `{{on 'input' this.update}}` or `{{on 'change' this.update}}` modifier on the field to update the value that gets passed in as `@value`.

When we chatted on Slack, I also asked if this was ever used stand-alone, or only via the Field components. If usage is done only via Field, users won't need to wire up any on-change for this component specifically. The value passed to the text inputs gets passed to `CharacterCount`, and then whatever modifier consumers have wired up will update the value passed to Field which then will also get passed to `CharacterCount` - so data flows from the top of the component.

> P.S. The mentioned rounding error in tests don't seem to be an issue in CI, but we have a few visual regressions. To resolve these we'd need a few examples updated in `packages/components/tests/dummy/app/templates/components/form/base-elements.hbs` (and other showcase examples for TextInput, Textarea, MaskedInput).

I don't have access to Percy, and the tests passed for me locally (I'll re-run to double-check this) - I'm happy to make these updates if you can give me guidance on what needs changed.

meirish:
> Right, so the API would be the same for Field components (which is great). It means I'm missing something as the following call doesn't seem to be updating the character count ([this is the first character count example in the preview page](https://hds-showcase-git-b-character-count-value-hashicorp.vercel.app/components/form/text-input)). I thought we'd still need to update the value. 🤔
> 
> ```
> <Hds::Form::TextInput::Field @value={{this.value}} as |F|>
>   <F.Label>This is the label text</F.Label>
>   <F.CharacterCount @maxLength={{40}} />
> </Hds::Form::TextInput::Field>
> ```

I pushed a change to https://github.com/hashicorp/design-system/pull/1896/commits/9303ab23fd82dd7116d595f6524251b22d190c39 to have the showcase more accurately reflect real-world use of the text-input component. All of the other instances where there is no `on` modifier don't actually get their internal value updated (which isn't an issue for the showcase app, but would be an issue if you wanted to say send that value somewhere on a form submission).

alex-ju:
> > Right, so the API would be the same for Field components (which is great). It means I'm missing something as the following call doesn't seem to be updating the character count ([this is the first character count example in the preview page](https://hds-showcase-git-b-character-count-value-hashicorp.vercel.app/components/form/text-input)). I thought we'd still need to update the value. 🤔
> > ```
> > <Hds::Form::TextInput::Field @value={{this.value}} as |F|>
> >   <F.Label>This is the label text</F.Label>
> >   <F.CharacterCount @maxLength={{40}} />
> > </Hds::Form::TextInput::Field>
> > ```
> 
> I pushed a change to [9303ab2](https://github.com/hashicorp/design-system/commit/9303ab23fd82dd7116d595f6524251b22d190c39) to have the showcase more accurately reflect real-world use of the text-input component. All of the other instances where there is no `on` modifier don't actually get their internal value updated (which isn't an issue for the showcase app, but would be an issue if you wanted to say send that value somewhere on a form submission).

I see now. Thanks for exemplifying! Will bring this to the engineering sync tomorrow and get back to you.

MelSumner:
I support this change, not sure how we missed this. Thank you for your PR!

alex-ju:
updated the rest of the examples in the showcase and rebased to main for more accurate/relevant visual regression results (only one change caused by a label I updated, but will leave other reviewers to check and approve it)

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
```diff
@@ -150,12 +150,24 @@
...
-      <input type="text" aria-label="currentLength = 0" id="cc-cl-eq-0" />
-      <Hds::Form::CharacterCount @controlId="cc-cl-eq-0" />
+      <input
+        type="text"
+        aria-label="currentLength = 0"
+        id="cc-cl-eq-0"
+        value={{this.value1}}
+        {{on "input" (fn this.updateValue "value1")}}
+      />
+      <Hds::Form::CharacterCount @value={{this.value1}} @controlId="cc-cl-eq-0" />
```
didoo, line 160: Is the `@controlId` still needed? 🤔

packages/components/addon/components/hds/form/character-count/index.js
```diff
@@ -4,16 +4,16 @@
...
 
 import Component from '@glimmer/component';
-import { tracked } from '@glimmer/tracking';
-import { action } from '@ember/object';
 
 const ID_PREFIX = 'character-count-';
 const NOOP = () => {};
 
 export default class HdsFormCharacterCountIndexComponent extends Component {
   // The current number of characters in the associated input
```
didoo, line 12: This comment needs to be updated

packages/components/addon/components/hds/form/character-count/index.hbs
```diff
@@ -8,7 +8,6 @@
   class={{this.classNames}}
   id={{this.id}}
   {{did-insert this.onInsert}}
```
didoo, line 10: Is this still needed? 🤔

packages/components/addon/components/hds/form/character-count/index.js
didoo, line 12: This comment needs to be updated

packages/components/tests/dummy/app/controllers/components/form/textarea.js
```diff
@@ -8,12 +8,21 @@ import { action } from '@ember/object';
...
+  @tracked withErrorMessage =
+    'Ut enim ad minim veniam, quis nostrud exercitation ullamco';
 
   @action
   noop() {}
 
-  @action onFieldInput(args) {
-    this.fieldIsInvalid = args.currentLength > args.maxLength;
+  get fieldIsInvalid() {
+    return this.withErrorMessage.length > 50;
```
didoo, line 22: probably, to avoid duplication, the `50` value should be stored in a variable and accessed in the template code via `this.xyz`

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
didoo, line 160: Is the `@controlId` still needed? 🤔

didoo: If not (as I suspect) than also the function that generates the `id()` can be updated to generate a generic ID (like we do for other similar field sub-elements)

didoo (commented):
Left a few comments. Overall seems good to me. 
I'll do a second pass when everything is "done".

packages/components/tests/dummy/app/templates/components/form/base-elements.hbs
meirish: This and the `onInsert` are still needed to hook up the `aria-describedby` properly - I initially removed both of them but tests caught the breakage and I backed out the removals.

packages/components/addon/components/hds/form/character-count/index.hbs
meirish: Unfortunately, yes - it wires up `aria-describedby`.

alex-ju: Ah, yes, we need this function and the `controlId` as this is [the mechanism the field component uses to append descriptors](https://github.com/hashicorp/design-system/blob/main/packages/components/addon/components/hds/form/field/index.hbs#L22). We have the same implementation for the [helper-text](https://github.com/hashicorp/design-system/blob/main/packages/components/addon/components/hds/form/helper-text/index.js#L30-L39) and [error message](https://github.com/hashicorp/design-system/blob/main/packages/components/addon/components/hds/form/error/index.js#L30-L39).

packages/components/addon/components/hds/form/character-count/index.js
alex-ju: updated in 1267055

packages/components/tests/dummy/app/controllers/components/form/textarea.js
alex-ju: done in 2eb5879

packages/components/addon/components/hds/form/character-count/index.hbs
didoo: 👍

didoo (approved):
Re-review and everything works for me. 

Before merging though we need a changeset entry. @alex-ju can you take care of it? Thanks

PR #1899, opened 2024-01-22
`SideNav::List::Title`: wrap text of long strings (HDS-3047)

KristinLBradley:
### :pushpin: Summary
If merged, this PR restructures the SidNav ListTitle DOM and updates styles so long unbroken text strings will wrap. It also updates css for `SideNav::List::BackLink` so long text strings will wrap.

image
image

### :link: External links

Jira ticket: [HDS-3047](https://hashicorp.atlassian.net/browse/HDS-3047)

***

### 👀 Component checklist

- [ ] ~~A11y tests have been run locally (`yarn test:a11y --filter="COMPONENT-NAME"`)~~
- [ ] ~~If documenting a new component, an acceptance test that includes the `a11yAudit` has been added~~

### DIFF
--- a/packages/components/addon/components/hds/side-nav/list/title.hbs
+++ b/packages/components/addon/components/hds/side-nav/list/title.hbs
@@ -4,5 +4,7 @@
 }}
 
 <Hds::SideNav::List::Item>
-  <div class="hds-side-nav__list-title hds-typography-body-100 hds-font-weight-semibold" ...attributes>{{~yield~}}</div>
+  <div class="hds-side-nav__list-title" ...attributes>
+    <span class="hds-side-nav__list-title-content hds-typography-body-100 hds-font-weight-semibold">{{~yield~}}</span>
+  </div>
 </Hds::SideNav::List::Item>
--- a/packages/components/app/styles/components/side-nav/content.scss
+++ b/packages/components/app/styles/components/side-nav/content.scss
@@ -53,6 +53,12 @@
   }
 }
 
+.hds-side-nav__list-title-content {
+  display: block;
+  max-width: 100%;
+  overflow-wrap: break-word;
+}
+
 
 // LIST (root elements)
 
@@ -147,8 +153,10 @@
 // LIST ITEM > INNER ELEMENTS
 
 .hds-side-nav__list-item-text {
+  max-width: 100%;
   color: var(--token-side-nav-color-foreground-primary);
   text-align: left;
+  overflow-wrap: break-word;
 }
 
 .hds-side-nav__list-item-icon-leading {
--- a/packages/components/tests/dummy/app/templates/components/side-nav.hbs
+++ b/packages/components/tests/dummy/app/templates/components/side-nav.hbs
@@ -513,6 +513,14 @@
         </Hds::SideNav::List::Item>
       </ul>
     </SF.Item>
+    <SF.Item @label="With very long text and no spaces">
+      <ul class="shw-component-sim-side-nav-body">
+        <Hds::SideNav::List::Title>ThisIsLongTextThatShouldWrapToTwoLinesAndNotOverflow</Hds::SideNav::List::Title>
+        <Hds::SideNav::List::Item>
+          <Shw::Placeholder @height="108px" @text="following content" />
+        </Hds::SideNav::List::Item>
+      </ul>
+    </SF.Item>
   </Shw::Flex>
 
   <Shw::Divider @level={{2}} />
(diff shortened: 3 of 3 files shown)


heatherlarsen (approved):
Nice ✨

packages/components/app/styles/components/side-nav/content.scss
```diff
@@ -147,8 +153,10 @@
 // LIST ITEM > INNER ELEMENTS
 
 .hds-side-nav__list-item-text {
+  max-width: 100%;
```
didoo, line 156: We need to set a `min-width: 0` to force it to stay within its flex container
modified

```suggestion
  max-width: 100%;
  min-width: 0;
```

packages/components/addon/components/hds/side-nav/list/title.hbs
```diff
@@ -4,5 +4,7 @@
 }}
 
 <Hds::SideNav::List::Item>
-  <div class="hds-side-nav__list-title hds-typography-body-100 hds-font-weight-semibold" ...attributes>{{~yield~}}</div>
+  <div class="hds-side-nav__list-title" ...attributes>
+    <span class="hds-side-nav__list-title-content hds-typography-body-100 hds-font-weight-semibold">{{~yield~}}</span>
```
didoo, line 8: I don't think we need an extra `<span>` here: I did a quick test and if we remove the `display: flex` from the `hds-side-nav__list-title-content` (not needed, I'm not sure why it was added - probably by me - in the first place) and apply a `overflow-wrap: break-word;` to it, then it works as expected

packages/components/app/styles/components/side-nav/content.scss
```diff
@@ -53,6 +53,12 @@
   }
 }
 
+.hds-side-nav__list-title-content {
```
didoo, line 56: See comment above, this probably is not needed

didoo (commented):
@KristinLBradley left a couple of suggestions

packages/components/addon/components/hds/side-nav/list/title.hbs
KristinLBradley: Ok thanks, I wasn't sure why you added display: flex but thought maybe it was to make it work with some other kind of content. I can remove & clean up.

packages/components/app/styles/components/side-nav/content.scss
KristinLBradley: Adding locally, thanks

KristinLBradley:
Made refinements from feedback.

packages/components/tests/dummy/app/templates/components/side-nav.hbs
```diff
@@ -616,12 +624,16 @@
     </SG.Item>
     <SG.Item @label="Icon + text + yielded content">
       <Hds::SideNav::List::Link @icon="boundary" @text="Boundary">
-        <Shw::Placeholder @height="20px" @text="yielded content" />
+        {{! template-lint-disable no-inline-styles }}
+        <Shw::Placeholder style="flex-grow: 1; height: 20px; width: auto;" @text="yielded content" />
```
KristinLBradley, line 628: @didoo Would it be better to add a classname to these placeholders and move the inline styles there?

(I had to override the default `width: 100%` & add `flex-grow: 1` so the placeholder would be more flexible and not force "Boundary" text to wrap. Adding the inline stye overrode styles from the args so I had to use all inline styles vs. `height` & `width` args.)

didoo: I would prefer to introduce a `@flex` argument so that you can say `@flex="1 1 0" and this _should_ fix the problem). See how the `@background` is done: https://github.com/hashicorp/design-system/blob/main/website/app/components/doc/placeholder/index.js#L56-L58

didoo: BTW, to add inline styles (in the showcase) there is a special helper `{{style ...}}`

KristinLBradley: Added new flex argument & updated in Showcase.

.changeset/few-emus-walk.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`SideNav::List::Title`, `SideNav::List::BackLink`: fixed issue with long text strings without spaces not wrapping
```
didoo, line 5: Also the "List::Link" is impacted (it uses `hds-side-nav__list-item-text` inside)

```suggestion
`SideNav::List::Title`, `SideNav::List::BackLink`, `SideNav::List::Link`: fixed issue with long text strings without spaces not wrapping
```

didoo (approved):
Added a suggestion for the codemod. Apart from that, all good 👍

PR #1900, opened 2024-01-22
remove unused function in ember-flight-icons

Dhaulagiri:
### :pushpin: Summary

I couldn't find it referenced anywhere in the project. It looks like it was removed from the template way back in https://github.com/hashicorp/flight/pull/245 and never got removed.


didoo (approved):
👍

https://github.com/search?q=contextRootURL+user%3Ahashicorp&type=code&ref=advsearch

PR #1902, opened 2024-01-23
Update pull_request_template.md

Dhaulagiri:
### :pushpin: Summary

This was discussed during our sync on 10/24 and these two steps happen for us automatically now provided we follow the correct blueprints when creating components.


didoo (approved):
👍

PR #1908, opened 2024-01-26
Add documentation for using multiple `Alerts`

jorytindall:
### :pushpin: Summary

This PR adds documentation around when to use multiple Alerts; spoiler alert (😉), it shouldn't be frequent, but it is possible and this guidance hopes to outline some options and things to consider.

This feedback is useful:

- Do you agree with the overall guidance? e.g. We don't expect this to happen, but it's OK if it does occasionally.
- I've added some notes on "contextualizing" Alerts which overlaps a bit with Placement, would it be worth trying to combine these two sections or add more meaningful differentiation?
- Do the examples (both realistic, and agnostic) make sense? It's a bit tricky to find real-world examples in our products, but I can do a bit more digging.

### :link: External links

Jira ticket: [HDS-2296](https://hashicorp.atlassian.net/browse/HDS-2296)
Figma file: [Assets and some exploration/context](https://www.figma.com/file/42LK10XbP5IERhzzgMOiI2/Website-assets?type=design&node-id=2929%3A593&mode=design&t=AqiJFoQEEoWRuwQs-1)

***

[HDS-2296]: https://hashicorp.atlassian.net/browse/HDS-2296?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -176,4 +176,64 @@ Inline alerts can be added to a section or component or inline with content.
...
+!!! Dont
+
+Don’t display more than one of the same `color` Alert on the page. This adds unnecessary visual weight to page and may result in the user being confused about which Alert is more important or which might require attention first.
+
+![Example of multiple Alerts of the same color](/assets/components/alert/multiple-color-alerts-dont.png)
+!!!
+
+!!! Do
+
+Instead, try to combine relevant and overlapping content into a single Alert.
```
alex-ju, line 198: thought: I'm wondering if this statement/recommendation is fully valid only for related Alerts. If they are unrelated we could suggest separating them into distinct sections (be those paragraphs or bullets, whatever makes more sense in the context).

website/public/assets/components/alert/alerts-within-reveal-open.png
alex-ju, line 1: questions: is the `(2)` meant to communicate there are two warnings? if so, is it worth updating the example to reflect that (make it look like there are two warnings – either by having separate paragraphs or some other way)

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -176,4 +176,64 @@ Inline alerts can be added to a section or component or inline with content.
...
 
-![Placement of compact Alert](/assets/components/alert/alert-placement-compact.png =600x*)
\ No newline at end of file
+![Placement of compact Alert](/assets/components/alert/alert-placement-compact.png =600x*)
+
+## Multiple alerts
+
+In general, displaying multiple alerts on the same page is discouraged as it can negatively impact the page hierarchy and unnecessarily shift emphasis away from the main content on the page.
+
+However, depending on the logic within the application and the type content being expressed to the user, it may be necessary to display more than one Alert on a page. Consider these scenarios and their accompanying guidelines if this is the case.
```
alex-ju, line 185:
```suggestion
However, depending on the logic within the application and the type of content being expressed to the user, it may be necessary to display more than one Alert on a page. Consider these scenarios and their accompanying guidelines if this is the case.
```

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -176,4 +176,64 @@ Inline alerts can be added to a section or component or inline with content.
...
+
+### Alerts of the same color
+
+!!! Dont
+
+Don’t display more than one of the same `color` Alert on the page. This adds unnecessary visual weight to page and may result in the user being confused about which Alert is more important or which might require attention first.
```
alex-ju, line 191:
```suggestion
Don’t display more than one of the same `color` Alert on the page. This adds unnecessary visual weight to the page and may result in the user being confused about which Alert is more important or which might require attention first.
```

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -176,4 +176,64 @@ Inline alerts can be added to a section or component or inline with content.
...
+!!! Do
+
+Instead, use Alerts hierarchically according to their intended prominence and importance to the user.
+
+![An example of multiple types of Alerts on the same page](/assets/components/alert/multiple-alert-types-do.png)
+!!!
+
+### Contextualizing Alerts
+
+`Inline` and `compact` Alerts can have more meaning if they are [placed](#placement) or contextualized within the element that is responsible for the Alert. This can help when it's necessary to have more than one Alert on the page and is relevant for pages that aggregate content like dashbaords.
```
alex-ju, line 224:
```suggestion
`Inline` and `compact` Alerts can have more meaning if they are [placed](#placement) or contextualized within the element that is responsible for the Alert. This can help when it's necessary to have more than one Alert on the page and is relevant for pages that aggregate content like dashboards.
```

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -176,4 +176,64 @@ Inline alerts can be added to a section or component or inline with content.
...
+
+![Example of multiple Alerts contextualized to different parts of the UI](/assets/components/alert/combining-contextualized-alerts.png)
+
+### Alerts within a Reveal
+
+**As a last resort**, when multiple `inline` Alerts are necessary, display the most important Alert the user within the page, and obfuscate additional Alerts within a [Reveal](/components/reveal).
```
alex-ju, line 230: This read a bit odd, not sure my suggestions improves it though 😄 

```suggestion
**As a last resort**, when multiple `inline` Alerts are necessary, display the most important Alert to the user within the page, and obfuscate additional Alerts within a [Reveal](/components/reveal).
```

alex-ju (commented):
Looks good overall. Left a few comments for your consideration

website/public/assets/components/alert/alerts-within-reveal-open.png
alex-ju: nevermind, I missed the second, compact, alert

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -176,4 +176,64 @@ Inline alerts can be added to a section or component or inline with content.
...
+
+### Ordering of Alerts
+
+Order multiple Alerts by their importance and impact on the user, with `critical` communicating Alerts that must be resolved immediately, and descending in perceived impact to `neutral`.
+
+![Ordering alerts](/assets/components/alert/ordering-alerts.png)
+
+### Alerts of the same type
+
+Try to avoid multiple Alerts of the same type (`page`, `inline`, etc) on the same page.
```
didoo, line 213: [blocker?] Wait. The scope of this is too wide: I clearly see cases in which I have two (or even more) alerts in the same **page**  (our documentation has a lot of pages wtih more than one alert per page :) ).

didoo: Another common use case is a top-level/top-of-the-page alert ("global" I think we call it) and then potentially other "local" alerts in the page.

website/docs/components/alert/partials/guidelines/guidelines.md
didoo, line 230: I remember when we implemented the Alert in HDS, there were "stacked alerts" in Structure (imagine like stacked cards). Should we consider this as a (future) super-component to solve this pattern?

didoo (commented):
Added a bunch of comments. 

Re. the feedback you asked for:

> Do you agree with the overall guidance? e.g. We don't expect this to happen, but it's OK if it does occasionally.

Yes.

> I've added some notes on "contextualizing" Alerts which overlaps a bit with Placement, would it be worth trying to combine these two sections or add more meaningful differentiation?

For me it works as it is.

> Do the examples (both realistic, and agnostic) make sense? It's a bit tricky to find real-world examples in our products, but I can do a bit more digging.

Yes. They were clear to me (apart from the one with a non-immediately perceivable second alert)

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -176,4 +176,64 @@ Inline alerts can be added to a section or component or inline with content.
...
+Instead, try to combine relevant and overlapping content into a single Alert.
+
+![Example of multiple Alerts combined into one](/assets/components/alert/multiple-color-alerts-do.png)
+
+This may result in the length of the content being longer, but consolidating messages into a single instance of the Alert is a more concise communication pattern.
+!!!
+
+### Ordering of Alerts
+
+Order multiple Alerts by their importance and impact on the user, with `critical` communicating Alerts that must be resolved immediately, and descending in perceived impact to `neutral`.
```
KristinLBradley, line 207:
```suggestion
Order multiple Alerts by their importance and impact on the user, starting from `critical`, to communicate an Alert that must be resolved immediately, descending in order of perceived impact to `neutral`.
```

website/docs/components/alert/partials/guidelines/guidelines.md
KristinLBradley, line 230:
```suggestion
**As a last resort**, when multiple `inline` Alerts are necessary, display the most important Alert the user within the page and obfuscate additional Alerts within a [Reveal](/components/reveal).
```

KristinLBradley: **As a last resort**, when multiple `inline` Alerts are necessary, display the most important Alert to the user within the page **_while obfuscating_** additional Alerts within a [Reveal](/components/reveal).

KristinLBradley: highlighted and bolded slightly different phrasing

KristinLBradley (commented):
Added a few suggested edits

website/docs/components/alert/partials/guidelines/guidelines.md
jorytindall: Thats a nice cleanup, thank you both!

jorytindall: Yeah maybe, though it seems like it's pretty easy to build just using our components. I guess the tricky part is probably more around the logic for displaying them in this way, thinking:

- if alerts.length =< 1 then render the alert directly in the page
- if alerts.length > 1 then render alert[0] in the page and everything else in a reveal

Could get a little tricky, but might be helpful to have this as a HDS+ pattern 😉

website/docs/components/alert/partials/guidelines/guidelines.md
jorytindall: That's great, thank you @KristinLBradley !

website/docs/components/alert/partials/guidelines/guidelines.md
jorytindall: This makes me think of a couple of things:

1. This is a tough scenario to compare to our docs site, because while visually we do have multiple alerts on a page, they are actually acting more as banners. But I could see where we could get pushback on this (re: "you're using tabs for navigation, why can't we?")
2. I think I can maybe soften this language a bit and make it a bit broader. I can confidently say that there should not be more than a single page-level Alert on a page, but your point about `inline` is reasonable. I'll rework this a bit.

website/docs/components/alert/partials/guidelines/guidelines.md
jorytindall: This is a good point as well, as I'm reading through some of the feedback it seems necessary to be a bit more clear on this topic.

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -176,4 +176,64 @@ Inline alerts can be added to a section or component or inline with content.
...
+### Alerts of the same type
+
+Try to avoid multiple Alerts of the same type (`page`, `inline`, etc) on the same page.
+
+!!! Do
+
+Instead, use Alerts hierarchically according to their intended prominence and importance to the user.
+
+![An example of multiple types of Alerts on the same page](/assets/components/alert/multiple-alert-types-do.png)
+!!!
```
heatherlarsen, line 220: I know it wasn't the intention, but this example makes it seem like we're advocating for critical alerts to always be global. There may be instances, such as within a form, where using an inline alert for critical messages is more appropriate. Is there some way to make this more clear?

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -176,4 +176,64 @@ Inline alerts can be added to a section or component or inline with content.
...
+
+### Contextualizing Alerts
```
heatherlarsen, line 222: This section does feel a little duplicative of the placement section, but the content is good. There may be value in merging the two into one section.

website/docs/components/alert/partials/guidelines/guidelines.md
heatherlarsen: +1 to @KristinLBradley's suggestion. Could probably also drop the "to the user" part to make it a bit more concise.

heatherlarsen: [This](https://www.figma.com/file/B7yd8loYS7nA2TiQcJ8UtV/Structure-1.9-%F0%9F%9A%A8-FULLY-DEPRECATED-%F0%9F%9A%A8?type=design&node-id=12569-15900&mode=design&t=lSbTh6sVKZ6zZzgl-4) is the only thing I was able to find in Structure, which appears to be the same as what Jory is proposing. @didoo I imagine what you're suggesting is something like this:

collapsed-alerts

I'm curious about this direction because it gives the user more context on the severity of the alerts below which feels more informative and may help the user avoid missing necessary notifications.

website/public/assets/components/alert/ordering-alerts.png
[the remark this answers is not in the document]
jorytindall: Good point, I've reduced the prominence/size of these a bit

website/public/assets/components/alert/alerts-within-reveal-open.png
jorytindall: I changed this to an `inline` Alert, but I will add some content around mixing and matching

website/docs/components/alert/partials/guidelines/guidelines.md
jorytindall: I've replaced this with a `highlight` Alert to avoid this, but I'll also add some information to the contextual section that might add a bit more clarity

website/docs/components/alert/partials/guidelines/guidelines.md
didoo: @heatherlarsen unfortunately I can't get Storybook run on an old version of Structure to compare with that design, but I think it was something similar (can't remember what happened in terms of UI transition, when the user clicked on the "additional alerts").

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -29,86 +29,86 @@ Use compact alerts when a less prominent alert is needed.
...
+<A.Description>Lorem ipsum dolar sit amet.</A.Description>
+</Hds::Alert>
+<Hds::Alert @type="inline" @color="warning" as |A|>
+<A.Title>Warning alert title</A.Title>
+<A.Description>Lorem ipsum dolar sit amet.</A.Description>
+</Hds::Alert>
+<Hds::Alert @type="inline" @color="critical" as |A|>
+<A.Title>Critical alert title</A.Title>
+<A.Description>Lorem ipsum dolar sit amet.</A.Description>
+</Hds::Alert>
```
KristinLBradley, line 51: Curious why it wipes out the indenting?

jorytindall: Whoops, sometimes my Prettier config does this, I'll revert that!

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -56,59 +56,59 @@ Use color logically.
...
 
 Use the Alert for more intrusive message communication about errors or critical disruptions at an application, page, or section level where users need to take immediate action.
 
 <Hds::Alert @type="inline" @color="critical" as |A|>
-    <A.Title>You have exceeded 50 applies this month</A.Title>
-    <A.Description>You may only invoke applies that destroy managed resources. Upgrade now and access additional product features, unlimited applies, and increased concurrency.</A.Description>
-    <A.Button @text="Upgrade" @color="secondary" @onClick={{this.noop}} />
-    <A.Link::Standalone @color="secondary" @icon="arrow-right" @iconPosition="trailing" @text="View usage" @href="#" />
-  </Hds::Alert>
+  <A.Title>You have exceeded 50 applies this month</A.Title>
```
didoo, line 72: @jorytindall can you revert all these reformatting changes, to reduce the noise? thanks

website/docs/components/alert/partials/guidelines/guidelines.md
jorytindall: I did a bit of reorganizing here to fit this content together, though there's not really a super clean way of doing this. The upside is now relevant content is combined, but the downside is now the info on placement is a bit more buried. This might not be an issue given the naming conventions we use, but just noting it here.

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -158,22 +158,95 @@ We recommend keeping `critical` alerts non-dismissible as they are essential to
...
+
+![Ordering alerts of different types](/assets/components/alert/ordering-alerts-type.png)
+
+### Contextualizing Alerts
+
+`Inline` and `compact` Alerts can have more meaning if they are contextualized within the element that is responsible for the Alert. This can help when it’s necessary to have more than one Alert on the page and is relevant for pages that aggregate content like dashboards.
+
+![Example of multiple Alerts contextualized to different parts of the UI](/assets/components/alert/combining-contextualized-alerts.png)
+
+Contextualization is helpful when displaying a specific or localized message; e.g., within a form to highlight validation errors or problems with form submission.
```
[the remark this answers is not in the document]
jorytindall: I agree with you, and maybe it just requires a bit of rewording, but I have some content around inline alerts within forms in the [validation](https://helios.hashicorp.design/patterns/form-patterns?tab=validation#form-level-validation) section of the form patterns docs.

I could take out the specific reference to forms (make it more general)

website/docs/components/alert/partials/guidelines/guidelines.md
jorytindall: Is there an "easy" way to do that? I could use a little assistance there, it was likely included in a commit that had other changes as a part of it.

website/docs/components/alert/partials/guidelines/guidelines.md
heatherlarsen: Ah, I see. I had imagined that the content in "Contextualizing Alerts" would move into the Placement section (and the Placement section would remain where it was) as the placement content feels more important than multiple alerts. 

What if you dropped the heading "Contextualizing Alerts" and moved the content from that section directly into the "Placement" section? The two sections are trying to accomplish the same thing, but "placement" is a bit easier to understand as a concept. Let me know if you want to pair, happy to show you what I was thinking.

heatherlarsen (commented):
I took another pass-through and left mostly nit-picky comments and a few questions. For something we generally want teams to avoid doing, this is quite a bit of new content. Let's consider if there are ways to further reduce or reorganize the content.

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -158,22 +158,95 @@ We recommend keeping `critical` alerts non-dismissible as they are essential to
...
+
+While not recommended for `page` type Alerts, it may be necessary to have more than one `inline` or `compact` Alert on a page. If this is the case, ensure that the Alerts are [ordered](#ordering-of-alerts) by impact and [contextualized](#contextualizing-alerts) in areas where they provide the most relevance.
+
+### Ordering of Alerts
+
+Order multiple Alerts by their importance and impact on the user, starting from `critical`, to communicate an Alert that must be resolved immediately, and descending in order of perceived impact to `neutral`.
+
+![Ordering alerts](/assets/components/alert/ordering-alerts-color.png)
+
+Alert types have an implied hierarchy and should be ordered starting from `page` and descending in hierarchy to `compact`. This order may depend on how the Alerts are [contextualized](#contextualizing-alerts)
```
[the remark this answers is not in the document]
jorytindall: Hmmmm, I think I must have accidentally removed or combined something here that resulted in a sentence fragment. Let me revise/rework it a bit.

website/public/assets/components/alert/alerts-within-reveal-closed.png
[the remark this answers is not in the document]
jorytindall: Tightened up a bit 👍

website/public/assets/components/alert/alerts-within-reveal-open.png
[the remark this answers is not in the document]
jorytindall: Tightened up a bit 👍

website/docs/components/alert/partials/guidelines/guidelines.md
didoo: I've pushed a commit that undoes these changes (only these, don't worry 😀).

website/public/assets/components/alert/alerts-within-reveal-open.png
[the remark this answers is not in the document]
heatherlarsen: Shouldn't the alerts within the reveal be the same width as the persistent alert?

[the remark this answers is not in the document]
jorytindall: Yep, I noticed that in review today, thanks for the reminder

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -158,22 +158,95 @@ We recommend keeping `critical` alerts non-dismissible as they are essential to
...
+!!!
+
+### Alerts within a Reveal
+
+**As a last resort**, when multiple `inline` Alerts are necessary, display the most important Alert to the user within the page while obfuscating additional Alerts within a [Reveal](/components/reveal).
+
+!!! Warning
+
+This method should be **rare and infrequent**, and should only be used if the number of Alerts of the same type exceeds two. Instead, effort should be put into contextualizing Alerts that matter most to the user and in ways that make them easy to resolve.
+!!!
```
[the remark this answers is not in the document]
jorytindall: I don't think I would expect multiple different Alerts in the same reveal. I'll reword the previous point to say:

> **As a last resort**, when multiple Alerts of the same **type** are necessary, display the most important Alert.

But now that I'm thinking through it, I'm not sure if I would recommend this being used for multiple compact Alerts, because those should likely be used in a more specific context. Might be unlikely that that multiple compact Alerts would be displayed next to each other.

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -158,22 +158,95 @@ We recommend keeping `critical` alerts non-dismissible as they are essential to
...
+
+![Example of multiple Alerts combined into one](/assets/components/alert/multiple-color-alerts-do.png)
+
+![Multiple Alert messages combined in a bulleted list](/assets/components/alert/multiple-color-alerts-do-bulleted-list.png)
+
+### Alerts of the same type
+
+While not recommended for `page` type Alerts, it may be necessary to have more than one `inline` or `compact` Alert on a page. If this is the case, ensure that the Alerts are [ordered](#ordering-of-alerts) by impact and [contextualized](#contextualizing-alerts) in areas where they provide the most relevance.
+
+### Ordering of Alerts
```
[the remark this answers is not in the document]
jorytindall: I've moved it above the broader section for multiple alerts, but ordering really only comes into play if there _are_ multiple alerts on the page. So I'm a bit torn on that; I think the content around impact and hierarchy is more broad, though.

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -158,22 +158,95 @@ We recommend keeping `critical` alerts non-dismissible as they are essential to
...
+
+To avoid this, try combining relevant or overlapping content into a single Alert with a more concise narrative, or by translating the related messages into a list.
+
+![Example of multiple Alerts combined into one](/assets/components/alert/multiple-color-alerts-do.png)
+
+![Multiple Alert messages combined in a bulleted list](/assets/components/alert/multiple-color-alerts-do-bulleted-list.png)
+
+### Alerts of the same type
+
+While not recommended for `page` type Alerts, it may be necessary to have more than one `inline` or `compact` Alert on a page. If this is the case, ensure that the Alerts are [ordered](#ordering-of-alerts) by impact and [contextualized](#contextualizing-alerts) in areas where they provide the most relevance.
```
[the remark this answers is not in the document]
jorytindall: Ok I've rolled off of this term and have switched everything to "placed" or "placement"

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -158,22 +216,37 @@ We recommend keeping `critical` alerts non-dismissible as they are essential to
...
-Compact alerts can be added to a section or component or inline with content.
+![Multiple Alert messages combined in a bulleted list](/assets/components/alert/multiple-color-alerts-do-bulleted-list.png)
+
+### Alerts of the same type
+
+While not recommended for `page` type Alerts, it may be necessary to have more than one `inline` or `compact` Alert on a page. If this is the case, ensure that the Alerts are [ordered](#ordering-of-alerts) by impact and [placed](#placing-alerts) in areas where they provide the most relevance.
+
+### Alerts within a Reveal
+
+**As a last resort**, when multiple Alerts of the same type are necessary, display the Alerts that require immediate attention or have the most impact to the user within the page while obfuscating additional Alerts within a [Reveal](/components/reveal). We recommend against hiding `critical` and `warning` Alerts in this manner as they generally must be addressed quickly to avoid a negative outcome.
```
heatherlarsen, line 243: I'm still unsure about this as a pattern recommendation. What do you think about adding a note and asking teams to include this in future product research and share the findings with us? Or some other way to help elicit more feedback around this pattern.

jorytindall: I'm not opposed to removing it, I can add a note in a banner or something, but this it does seem helpful to have something like this as a "catch-all" or just-in-case.

If we're unsure about it (and think that it's unlikely to  happen frequently, if at all), then it might just be better to remove it entirely so that the idea doesn't exist out in the wild. What do you think?

heatherlarsen: I think it's fine to keep it (this was ultimately the goal of the original ticket, after all), but I do think it's worth noting that we'd like real user feedback on this pattern from teams that do end up needing to use it.

jorytindall: Added a bit of a note here requesting that consumers perform usability testing and share their findings with the team.

heatherlarsen (approved):
Looks great! Thanks for all the iterations.

jorytindall:
I imagine the showcase failing to deploy is due to the rearchitecting? I opened this branch prior to that being merged. Do I need to rebase from the main branch? cc. @didoo @alex-ju

didoo:
> I imagine the showcase failing to deploy is due to the rearchitecting? I opened this branch prior to that being merged. Do I need to rebase from the main branch? cc. @didoo @alex-ju

@alex-ju @jorytindall I have
- rebased the branch on main and fixed a few conflicts
- removed Jory's last commit: it was essentially committing the `showcase/dist` folder, not sure why this happened

Please check again that all the expected content is there.

website/docs/components/alert/partials/guidelines/guidelines.md
```diff
@@ -26,6 +26,12 @@ Use inline alerts when the message is contextual or specific to the section it
 
 Use compact alerts when a less prominent alert is needed.
 
+### Ordering Alerts by type
+
+Alert types have an implied hierarchy and should be ordered starting from `page` and descending in hierarchy to `compact`.
+
+![Ordering alerts of different types](/assets/components/alert/ordering-alerts-type.png)
```
didoo, line 33: not to be picky, but maybe this image could be something like this to convey even more the hierarchy?

screenshot_3499

didoo (approved):
Left only one non-blocking small suggestion. Apart from that, good to go for me 👍

website/docs/components/alert/partials/guidelines/guidelines.md
jorytindall: Good suggestion, I've updated this image to add a bit more implied hierarchy.

PR #1909, opened 2024-01-26
`SideNav` - Disable keyboard navigation/focus of hidden elements

alex-ju:
### :pushpin: Summary

Currently, interactive elements within the SideNav are focusable even when the SideNav is minimized/collapsed. This makes them visible and interactive to screen-reader users even though they are not visible.

The usual and easiest way to fix this issue is to set `display: none` on such elements. However, because of the desired fade effect, we can't use this method and we rely on changing the `visibility` of these containers. To fix this issue while preserving the fade effect, we make inert the containers potentially holding such interactive elements.

To facilitate the review of this feature I resurrected [the demo app](https://hds-showcase-git-alex-ju-local-demo-app-hashicorp.vercel.app/demo-app) previously used for other SideNav improvements.

### :link: External links

Jira ticket: [HDS-2936](https://hashicorp.atlassian.net/browse/HDS-2936)

***

### 👀 Component checklist

[HDS-2936]: https://hashicorp.atlassian.net/browse/HDS-2936?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/strong-balloons-repeat.md
+++ b/.changeset/strong-balloons-repeat.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`SideNav` - fixed issue with navigation elements remaining interactive when minimized
--- a/packages/components/src/components/hds/side-nav/index.js
+++ b/packages/components/src/components/hds/side-nav/index.js
@@ -111,11 +111,26 @@ export default class HdsSideNavComponent extends Component {
   toggleMinimizedStatus() {
     this.isMinimized = !this.isMinimized;
 
+    this.containersToHide.forEach((element) => {
+      if (this.isMinimized) {
+        element.setAttribute('inert', '');
+      } else {
+        element.removeAttribute('inert');
+      }
+    });
+
     let { onToggleMinimizedStatus } = this.args;
 
     if (typeof onToggleMinimizedStatus === 'function') {
       onToggleMinimizedStatus(this.isMinimized);
     }
+  }
+
+  @action
+  didInsert(element) {
+    this.containersToHide = element.querySelectorAll(
+      '.hds-side-nav-hide-when-minimized'
+    );
   }
 
   @action
--- a/showcase/tests/integration/components/hds/side-nav/index-test.js
+++ b/showcase/tests/integration/components/hds/side-nav/index-test.js
@@ -217,6 +217,8 @@ module('Integration | Component | hds/side-nav/index', function (hooks) {
     assert
       .dom('#test-side-nav-footer')
       .doesNotHaveAttribute('data-test-minimized');
+    assert.dom('.hds-side-nav__wrapper-body').doesNotHaveAttribute('inert');
+    assert.dom('.hds-side-nav__wrapper-footer').doesNotHaveAttribute('inert');
 
     await click('.hds-side-nav__toggle-button');
 
@@ -230,6 +232,8 @@ module('Integration | Component | hds/side-nav/index', function (hooks) {
     assert.dom('#test-side-nav-header').hasAttribute('data-test-minimized');
     assert.dom('#test-side-nav-body').hasAttribute('data-test-minimized');
     assert.dom('#test-side-nav-footer').hasAttribute('data-test-minimized');
+    assert.dom('.hds-side-nav__wrapper-body').hasAttribute('inert');
+    assert.dom('.hds-side-nav__wrapper-footer').hasAttribute('inert');
   });
 
   // CALLBACKS
--- a/showcase/tests/integration/components/hds/side-nav/index-test.js
+++ b/showcase/tests/integration/components/hds/side-nav/index-test.js
@@ -195,6 +195,7 @@ module('Integration | Component | hds/side-nav/index', function (hooks) {
         </:header>
         <:body as |B|>
           <span id="test-side-nav-body" data-test-minimized={{B.isMinimized}} />
+          <span class="hds-side-nav-hide-when-minimized" />
         </:body>
         <:footer as |F|>
           <span id="test-side-nav-footer" data-test-minimized={{F.isMinimized}} />
@@ -217,6 +218,9 @@ module('Integration | Component | hds/side-nav/index', function (hooks) {
     assert
       .dom('#test-side-nav-footer')
       .doesNotHaveAttribute('data-test-minimized');
+    assert
+      .dom('.hds-side-nav-hide-when-minimized')
+      .doesNotHaveAttribute('inert');
 
     await click('.hds-side-nav__toggle-button');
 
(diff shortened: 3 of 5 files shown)


packages/components/addon/components/hds/side-nav/base.hbs
```diff
@@ -9,10 +9,10 @@
     <div class="hds-side-nav__wrapper-header">
       {{yield to="header"}}
     </div>
-    <div class="hds-side-nav__wrapper-body">
+    <div class="hds-side-nav__wrapper-body" inert={{@isMinimized}}>
```
didoo, line 12: Nice and neat! 🙌

packages/components/addon/components/hds/side-nav/base.hbs
```diff
@@ -9,10 +9,10 @@
...
       {{yield to="body"}}
     </div>
-    <div class="hds-side-nav__wrapper-footer">
+    <div class="hds-side-nav__wrapper-footer" inert={{@isMinimized}}>
```
didoo, line 15: Let's imagine that consumers want to add something that persists in the UI even when the sidenav is collapsed (do you remember we have a lot of "extra" content slots in the sidenav, you can see them in the [demo app](https://hds-showcase-git-alex-ju-local-demo-app-hashicorp.vercel.app/demo-app) too). 
From [this quick test](https://codepen.io/didoo/pen/NWJwvNb) they would not be able to assign an `inert="false"` to their custom content to bypass the parent's `inert="true"`.

Thoughts of how we could bypass this limitation? 🤔

didoo (commented):
Left a praise and... a question (sorry). 😀

packages/components/addon/components/hds/side-nav/base.hbs
MelSumner: it's not something to be bypassed, it's a constraint that consumers have to understand and accept as well.

MelSumner (approved):
This is a tidy approach, nice thinking!

packages/components/addon/components/hds/side-nav/base.hbs
didoo: @MelSumner the point is that they may have _extra_ content that is correct semantically and functionally, is fully accessible, and for [reasons] they want to leave it available even when the sidenav is collapsed (think of an extra button, for example to quickly open a CLI). if we make the entire `body` and `footer` inert, this would become inert too. my question was about bypassing the limitation in our solution. In theory if it was a CSS property we could tell them to add a special class to it, like we already do with the `hds-side-nav-hide-when-minimized` class ([reference](https://helios.hashicorp.design/components/side-nav?tab=code#content-transitions-between-states)) but since this is an HTML attribute I can't think of other ways to do it.

alex-ju: The only element that can escape inertness is the modal dialog.

Alternatively to this PR, we can offer guidance and ask the consumer to manage the inert state. Of course, it comes with tradeoffs (if they mismanage it, it becomes unusable or non-conformant).

For portal content

```hbs
<Hds::SideNav>
  <:body as |Body|>
    <Hds::SideNav::Portal::Target inert={{Body.isMinimized}} @targetName="sidenav-portal" />
  </:body>
</Hds::SideNav>
```

For list content

```hbs
<Hds::SideNav>
  <:body as |Body|>
    <Hds::SideNav::List aria-label="Services" inert={{Body.isMinimized}} as |SNL|>
      <SNL.Title>Services</SNL.Title>
      <SNL.Link @text="Boundary" @icon="boundary" @href="#" />
      <SNL.Link @text="Consul" @icon="consul" @href="#" />
    </Hds::SideNav::List>
  </:body>
</Hds::SideNav>
```

didoo: What if apply `inert` via JS, like `[DOM Element].setAttribute('inert', true)` (and we can point to specific selectors of things we know we want to make inert, and when, in the `index.js` code?

alex-ju: The problem with us setting inert only on certain elements is that the consumer may not be aware of it and will have no way to opt-in with other elements to become inert/non-inert based on the collapsed/expanded state of the nav. We only need one element (that is not visible when the SideNav is collapsed) to miss the inert attribute and the implementation is non-compliant.

alex-ju: I'll give it some more thought, but I don't think the approach in this PR allows for exceptions

alex-ju: Updated this PR (and the demo app) to make elements with `.hds-side-nav-hide-when-minimized` inert when the SideNav is collapsed. This way we can have parity between hidden elements and the inert attribute, regardless of whether the class is set within the component, or by the consumer.

.changeset/strong-balloons-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`SideNav` - fixed navigation elements remaining interactive when minimized
```
didoo, line 5:
```suggestion
`SideNav` - fixed issue with navigation elements remaining interactive when minimized
```

packages/components/src/components/hds/side-nav/index.js
```diff
@@ -8,6 +8,7 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { assert } from '@ember/debug';
 import { registerDestructor } from '@ember/destroyable';
+import { schedule } from '@ember/runloop';
```
didoo, line 11: I think this is not needed anymore

packages/components/src/components/hds/side-nav/index.js
```diff
@@ -111,13 +112,28 @@ export default class HdsSideNavComponent extends Component {
...
     let { onToggleMinimizedStatus } = this.args;
 
     if (typeof onToggleMinimizedStatus === 'function') {
       onToggleMinimizedStatus(this.isMinimized);
     }
   }
 
+  @action
+  didInsert(element) {
+    this.containersToHide = element.querySelectorAll(
```
didoo, line 132: 🙌 🙌 🙌 🙌

showcase/tests/integration/components/hds/side-nav/index-test.js
```diff
@@ -217,6 +218,9 @@ module('Integration | Component | hds/side-nav/index', function (hooks) {
     assert
       .dom('#test-side-nav-footer')
       .doesNotHaveAttribute('data-test-minimized');
+    assert
```
didoo, line 221: should we also test that another element (`<span>`) without class `hds-side-nav-hide-when-minimized` doesn't have the `inert` attribute (in both conditions)? just in case

didoo (commented):
left a couple of comments, but I am happy with the solution

let me know when the comments are resolved that I'll approve

packages/components/addon/components/hds/side-nav/base.hbs
didoo: Yes!!!! I was sure you would have found the "idiomatic" solution in the end! Great job 👏 👏 👏

showcase/tests/integration/components/hds/side-nav/index-test.js
alex-ju: added in 4e561f0

didoo (approved):
Nice job! 👏

PR #1910, opened 2024-01-26
`Dropdown` - Fixed dropdown list missing an accessible name when Checkmark items were passed in

alex-ju:
### :pushpin: Summary

Elements with `listbox` role require an accessible name (be that via `aria-label`, `aria-labelledby` or `title`).

~~One approach (illustrated in this PR) would be to allow consumers to set an `aria-label` for those instances of `Hds::Dropdown` that contain `ListItem::Checkmark`. This would need to be accompanied by guidance.~~

An alternative approach would be to use the `Dropdown::Toggle::Button` as a label (using `aria-labelledby`), but currently we don't have a direct way to access that element or its properties.

Opened this up to discuss approaches and tradeoffs.

Update: based on feedback, we decided to take the alternative approach, using `aria-labelledby`.

### :link: External links

Jira ticket: [HDS-3067](https://hashicorp.atlassian.net/browse/HDS-3067)

***

### 👀 Component checklist

[HDS-3067]: https://hashicorp.atlassian.net/browse/HDS-3067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/long-meals-appear.md
+++ b/.changeset/long-meals-appear.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Dropdown` - Fixed dropdown list missing an accessible name when Checkmark items were passed in
--- a/packages/components/src/components/hds/dropdown/index.js
+++ b/packages/components/src/components/hds/dropdown/index.js
@@ -36,12 +36,16 @@ export default class HdsDropdownIndexComponent extends Component {
   }
 
   /**
-   * @param listAriaLabel
-   * @type {string}
-   * @description Sets a label for when  checkmark list items are used in the dropdown list
+   * Gets the id of the associated toggle button
+   * @method labelledby
+   * @return {string} The `id` attribute to be provided to `aria-labelledby`
    */
-  get listAriaLabel() {
-    return this.args.listAriaLabel || '';
+  get labelledby() {
+    let dropdownButton = this.listElement
+      .closest('.hds-dropdown')
+      .querySelector('.hds-dropdown-toggle-button');
+    let id = dropdownButton?.getAttribute('id');
+    return id;
   }
 
   /**
@@ -81,10 +85,11 @@ export default class HdsDropdownIndexComponent extends Component {
 
   @action
   didInsertList(element) {
+    this.listElement = element;
     let checkmarkItems = element.querySelectorAll(`[role="option"]`);
     if (checkmarkItems.length) {
       element.setAttribute('role', 'listbox');
-      element.setAttribute('aria-label', this.listAriaLabel);
+      element.setAttribute('aria-labelledby', this.labelledby);
     }
   }
 }
--- a/packages/components/src/components/hds/dropdown/index.js
+++ b/packages/components/src/components/hds/dropdown/index.js
@@ -72,9 +72,9 @@ export default class HdsDropdownIndexComponent extends Component {
 
   @action
   didInsertList(element) {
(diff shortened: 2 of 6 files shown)


DingoEatingFuzz:
Two cents: Since listbox is inferred/abstracted, it would be really nice if the name for it was also inferred/abstracted (though I know that's easier said than done).

Short of that, what about a more generic argument like "name" or "label"? `@listAriaLabel` is pretty in the weeds and leaks implementation details.

alex-ju:
> Two cents: Since listbox is inferred/abstracted, it would be really nice if the name for it was also inferred/abstracted (though I know that's easier said than done).
> 
> Short of that, what about a more generic argument like "name" or "label"? `@listAriaLabel` is pretty in the weeds and leaks implementation details.

Yeah, I hear you, but then it would be inconsistent with all the other similar arguments (such as `tagAriaLabel`, `visibilityToggleAriaLabel` and [other `ariaLabel`s](https://github.com/search?q=repo%3Ahashicorp%2Fdesign-system+%40ariaLabel+language%3AHandlebars+path%3A%2F%5Epackages%5C%2Fcomponents%5C%2F%2F&type=code)).

DingoEatingFuzz:
Yeah, consistency is good and you know the Helios code base more than I do 👍 I imagine since this is a dropdown component and all, "list"--although situational--isn't too surprising.

packages/components/addon/components/hds/dropdown/index.js
```diff
@@ -35,6 +35,15 @@ export default class HdsDropdownIndexComponent extends Component {
     return listPosition;
   }
 
+  /**
+   * @param listAriaLabel
+   * @type {string}
+   * @description Sets a label for when  checkmark list items are used in the dropdown list
+   */
+  get listAriaLabel() {
+    return this.args.listAriaLabel || '';
```
didoo, line 44: A comment and a consideration:
- should we have a meaningful default
- if there is a meaningful default we can use (eg. "List of options" or "A list containing a set of options") can we just use it directly without asking the consumer to provide an extra argument (they may not know, and we don't enforce it, so it's very likely it will be missed)

didoo (commented):
Added a comment

packages/components/addon/components/hds/dropdown/index.js
alex-ju: @MelSumner when you have a moment to spare I would like to get your feedback on this change and potentially for a default label

MelSumner: I would say to try to use the trigger button label. Give it an ID, and then put `aria-labelledby` on the `<ul>` with the `role="listbox"`, where the value of `aria-labelledby` is the ID of the `<button>` text used for the trigger. 

WDYT?

alex-ju: That sounds appropriate – I'll work towards that

didoo: @alex-ju and I played a little bit with the code and we found that `closest()` works (to be tested more, just to be sure) and so the solution proposed is OK

MelSumner: I don't have any objection to this approach. 👍

alex-ju: @didoo when time allows, I think this is good for another review

PR #1912, opened 2024-01-29
Table Multiselect - Remove custom `toggle` event and use ad-hoc callback to force `aria-label` updates

didoo:
### :pushpin: Summary

While reviewing https://github.com/hashicorp/design-system/pull/1859/ one conversation was around the use of a custom `toggle` event to force the update of the aria labels for the checkboxes (see comment https://github.com/hashicorp/design-system/pull/1859/files#r1463725754)

This PR proposes an alternative approach, using a callback function, to allow a discussion in this specific change (easier to review and comment) 

**Showcase:** https://hds-showcase-git-table-multiselect-remove-cust-809fa9-hashicorp.vercel.app/components/table

***

### 👀 Component checklist


KristinLBradley:
In testing this, it's now not updating the aria-label of the rows when they are selected using the header checkbox to select/deselect all. (The header checkbox aria-label seems to update properly though when you manually select or deselect all rows individually.)

KristinLBradley (commented):
I tried to figure out a solution for the row aria-labels not updating but haven't figured out anything so far. See other comment...

didoo:
> In testing this, it's now not updating the aria-label of the rows when they are selected using the header checkbox to select/deselect all. (The header checkbox aria-label seems to update properly though when you manually select or deselect all rows individually.)

At least the good thing is that now the integration tests are catching these errors. I'll have a look today.

didoo:
OK, after some investigation I've found the problem, and it's that the solution I implemented is conceptually wrong. In the sense that when I declare the function in `table/th-selectable.js`:
```
  @action
  forceUpdateAriaLabel(checkbox) {
    this.isSelected = checkbox.checked;
  }
```
the `this` refers to the `this` of the checkbox this function is associated with and when we do the binding in the `didInsert`:
```
  @action
  didInsert(checkbox) {
    let { didInsert } = this.args;
    if (typeof didInsert === 'function') {
      didInsert(
        checkbox,
        this.args.selectionKey,
        this.forceUpdateAriaLabel.bind(this)
      );
    }
  }
```
we still refer to the same `this`.

But then in the `table/index.js` we store the function `forceUpdateAriaLabel` as `forceUpdateAriaLabelCallback` using the "select all" checkbox, so when we call:
```
this.forceUpdateAriaLabelCallback(row.checkbox);
```
the `row.checkbox` refers to the row's checkbox, but the `this` refers to the "select all" checkbox, so the row's checkbox never gets forcibly updated, only the select all (no matter what checkbox we toggle)

In order to get it working, I would have to store all the `forceUpdateAriaLabel` for each checkbox, and invoke it (by searching for the corresponding match checkbox-callback) and this would be an even worse solution than the current one, so I am going to close this PR and continue the conversation in the original thread.

/cc @KristinLBradley @alex-ju

didoo:
@alex-ju @KristinLBradley an alternative could be to simply ignore which specific checkbox has been changed, and force the refresh of all the checkboxes' aria labels. what do you think?

alex-ju:
> @alex-ju @KristinLBradley an alternative could be to simply ignore which specific checkbox has been changed, and force the refresh of all the checkboxes' aria labels. what do you think?

If the `toggle` event does the job and only updates the necessary label, I'd be more inclined to leave it as is – but thanks for exploring other options here.

PR #1916, opened 2024-01-31
`PageHeader` - Force Dropdown positioning to left when used in Actions

alex-ju:
### :pushpin: Summary

Force Dropdown positioning to left when used in `PageHeader::Actions`

A potential solution to force Dropdown list positioning to the left when used in `PageHeader::Actions`. I don't _like_ this override as it breaks the encapsulation principle but paired with the existing guidance where we expect to only have Buttons and Dropdowns as PageHeader actions it may be worth considering.

### :link: External links

[Slack thread](https://hashicorp.slack.com/archives/C7KTUHNUS/p1706299619555619?thread_ts=1705943470.509759&cid=C7KTUHNUS)

***

### 👀 Component checklist

### DIFF
--- a/packages/components/app/styles/components/page-header.scss
+++ b/packages/components/app/styles/components/page-header.scss
@@ -44,6 +44,15 @@
     justify-content: space-between;
     min-width: 0; // this is important or it will blow beyond the parent flexbox width with long non-breaking names
   }
+
+  // Dropdowns with @listPosition="bottom-right" need an override to prevent them from spilling outside the page
+  @container (max-width: 767px) {
+    .hds-page-header__actions {
+      .hds-dropdown__content {
+        left: 0;
+      }  
+    }
+  }
 }
 
 .hds-page-header__content {


packages/components/app/styles/components/page-header.scss
```diff
@@ -44,6 +44,15 @@
     justify-content: space-between;
     min-width: 0; // this is important or it will blow beyond the parent flexbox width with long non-breaking names
   }
+
+  // Dropdowns with @listPosition="bottom-right" need an override to prevent them from spilling outside the page
```
didoo, line 48:
```suggestion
  // Dropdowns with @listPosition="bottom-right" need an override to prevent them from spilling outside the page
  // Likely we will be be able to remove this override once we introduce collision detection for the dropdown - see: https://hashicorp.atlassian.net/browse/HDS-1995
```

didoo (commented):
I don't like this too, but maybe it's an OK (as in "hold your nose") workaround while we wait to get the collision detection done?

alex-ju:
> I don't like this too, but maybe it's an OK (as in "hold your nose") workaround while we wait to get the collision detection done?

Yes, that's my thinking too. I can leave a note to remove the hack when a `@listPosition="auto"` (or similar) becomes available.

alex-ju:
Consumers have a temporary fix for now so we'll wait to address this properly after introducing collision detection

PR #1918, opened 2024-02-01
`DialogPrimitive` shared components (HDS-3081)

KristinLBradley:
---
The release of these changes is being put off to a future date. [See related doc for more context](https://docs.google.com/document/d/1yixV5kfGzrao8LMT0YHRrUkdN24nxf1II0-lC6pd9Vk/edit) around this decision.
---

### :pushpin: Summary
If merged, this PR separates out shared components between `Flyout` & `Modal` as new "DialogPrimitive" components and uses them within the `Modal` & `Flyout` components.

- [DialogPrimitive components web docs](https://hds-website-git-hds-3081-dialog-primitive-components-hashicorp.vercel.app/utilities/dialog-primitive)
- [DialogPrimitive components showcase](https://hds-showcase-git-hds-3081-dialog-primitive-components-hashicorp.vercel.app/utilities/dialog-primitive)

Compare to:
- [Flyout showcase](https://hds-showcase-git-hds-3081-dialog-primitive-components-hashicorp.vercel.app/components/flyout)
- [Modal showcase](https://hds-showcase-git-hds-3081-dialog-primitive-components-hashicorp.vercel.app/components/modal)

~~**See also documentation branch:** https://github.com/hashicorp/design-system/pull/1945~~ _- merged in_

**New `DialogPrimitive` components:**

- `Wrapper` (layout container)
- `Header`
- `Description` (used only in Flyout)
- `Body`
- `Footer`

### :link: External links

Jira ticket: [HDS-3081](https://hashicorp.atlassian.net/browse/HDS-3081)

***

### 👀 Component checklist

[HDS-3081]: https://hashicorp.atlassian.net/browse/HDS-3081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KristinLBradley:
There are a couple of failing tests I still need to fix. Could use feedback on things implemented so far.

packages/components/app/styles/components/dialog-primitive.scss
```diff
@@ -0,0 +1,117 @@
...
+
+// FOOTER
+
+.hds-dialog-primitive__footer {
+  flex: none;
+  padding: 16px 24px;
+  background: var(--token-color-surface-faint);
+  border-top: 1px solid var(--token-color-border-primary);
+
+  // Tertiary buttons must always be placed/aligned at the end of the row
```
didoo, line 101: This rule doesn't apply to the `Modal` component, so this declaration should be moved back in the CSS for the `Flyout` (@jorytindall keep me honest on this)

packages/components/app/styles/components/flyout.scss
```diff
@@ -50,58 +50,17 @@
...
   border-top: 1px solid var(--token-color-border-primary);
 }
 
 .hds-flyout__footer {
-  flex: none;
-  padding: 16px 24px;
-  background: var(--token-color-surface-faint);
   border-top: 1px solid var(--token-color-border-primary);
-
-  // Tertiary buttons must always be placed/aligned at the end of the row
```
didoo, line 97: See comment above about this rule, which is specific to the `Flyout` only (so this rule should stay here)

packages/components/app/styles/components/modal.scss
```diff
@@ -120,7 +85,7 @@
     border-bottom: 1px solid var(--token-color-border-warning);
   }
 
-  .hds-modal__tagline {
+  .hds-dialog-primitive__tagline {
```
didoo, line 88: A reminder to discuss later if it's better at that point also change the sibling declaration to override the `hds-dialog-primitive__header` instead of the `hds-modal__header`.

packages/components/tests/dummy/app/router.js
```diff
@@ -30,6 +30,7 @@ Router.map(function () {
     this.route('button-set');
     this.route('card');
     this.route('code-block');
+    this.route('dialog-primitive');
```
didoo, line 33: TBD if we want to have it under the "Utilities" categories as for the other "primitives" (personally, I don't dislike at all the `DialogPrimitive` name).

packages/components/tests/integration/components/hds/flyout/index-test.js
```diff
@@ -80,10 +80,10 @@ module('Integration | Component | hds/flyout/index', function (hooks) {
             <F.Header>Title</F.Header>
           </Hds::Flyout>`
     );
-    assert.dom('.hds-flyout__title').exists();
-    assert.dom('.hds-flyout__title').hasText('Title');
-    assert.dom('.hds-flyout__icon').doesNotExist();
-    assert.dom('.hds-flyout__tagline').doesNotExist();
+    assert.dom('.hds-dialog-primitive__title').exists();
```
didoo, line 83: I am not convinced of this change. I'll think about it more and review the integration tests when the PR is ready for final review

didoo (commented):
@KristinLBradley I did a first pass, and... great job! 🙌 🙌 🙌 We're definitely in the right direction and seems all the assumptions we made were correct, so... I don't see that much else to do (once @hashicorp/hds-design decides on the small visual details/differences) than polish and finalize the code, and then document the new component (TBD in how much detail)

packages/components/addon/components/hds/dialog-primitive/header/index.hbs
```diff
@@ -0,0 +1,36 @@
...
+
+  <Hds::Text::Display class="hds-dialog-primitive__title" @tag="div" @size="300" @weight="semibold" id={{@id}}>
+    {{#if @tagline}}
+      <Hds::Text::Body class="hds-dialog-primitive__tagline" @tag="div" @size="100" @weight="regular">
+        {{@tagline}}
+      </Hds::Text::Body>
+    {{/if}}
+
+    {{yield}}
+  </Hds::Text::Display>
```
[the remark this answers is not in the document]
KristinLBradley: Thanks, I'll make this change locally.

packages/components/addon/components/hds/dialog-primitive/description.hbs
```diff
@@ -0,0 +1,8 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<Hds::Text::Body class="hds-dialog-primitive__description" @tag="div" @size="200" @color="primary" ...attributes>
+  {{yield}}
+</Hds::Text::Body>
```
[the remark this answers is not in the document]
KristinLBradley: There had been mention that we may want to make the `Description` available to the `Modal` component as well instead of only using it in the `Flyout` as it is currently so that could effect the styling.

@majedelass any thoughts on this yet?

packages/components/addon/components/hds/dialog-primitive/header/index.hbs
```diff
@@ -0,0 +1,36 @@
...
+* FLYOUT TAGGLINE:
+<Hds::Text::Body class="hds-flyout__tagline" @tag="div" @size="100" @weight="regular" @color="faint">
+  {{@tagline}}
+</Hds::Text::Body>
+
+* MODAL TAGGLINE:
+<Hds::Text::Body class="hds-modal__tagline" @tag="div" @size="100" @weight="regular">
+  {{@tagline}}
+</Hds::Text::Body>
+--}}
```
KristinLBradley, line 36: This is just here as a temporary reference while I work on this. It will be deleted in the end.

KristinLBradley, line 36: This is just here as a temporary reference while I work on this. It will be deleted in the end.

packages/components/app/styles/components/dialog-primitive.scss
```diff
@@ -0,0 +1,117 @@
...
+
+/*
+HEADER DIFFERENCES:
+.hds-modal__header {
+  border-top-left-radius: inherit;
+  border-top-right-radius: inherit;
+}
+
+.hds-flyout__header {}
+*/
```
KristinLBradley, line 46: This is just here as a temporary reference while I work on this. It will be deleted in the end.

packages/components/app/styles/components/flyout.scss
KristinLBradley: The exact same styles were originally in the modal.scss too though. See below.

```
  // Tertiary buttons must always be placed/aligned at the end of the row
  .hds-button-set {
    .hds-button--color-tertiary {
      margin-left: auto;
    }
  }
```

packages/components/app/styles/components/modal.scss
```diff
@@ -40,55 +40,18 @@
...
-  border-top: 1px solid var(--token-color-border-primary);
   border-bottom-right-radius: inherit;
   border-bottom-left-radius: inherit;
-
-  // Tertiary buttons must always be placed/aligned at the end of the row
-  .hds-button-set {
-    .hds-button--color-tertiary {
-      margin-left: auto;
-    }
-  }
```
KristinLBradley, line 89: These are the same as the styles that were in the Flyout styles

```
  // Tertiary buttons must always be placed/aligned at the end of the row
  .hds-button-set {
    .hds-button--color-tertiary {
      margin-left: auto;
    }
  }
```

packages/components/app/styles/components/dialog-primitive.scss
KristinLBradley: The same styles were in both so I moved them to these shared styles. Let me know if that was a mistake in the old code or something though.

The new component (or whatever) which will make use of the Footer will likely want to follow this layout I would think btw.

packages/components/tests/dummy/app/router.js
KristinLBradley: Maybe make a new "Primitives" category?

packages/components/app/styles/components/dialog-primitive.scss
didoo: My bad 🤦‍♂️
screenshot_3457

packages/components/tests/dummy/app/router.js
didoo: Yeah, maybe.

KristinLBradley:
@cristiano

So currently I've created 4 new "Dialog Primitive" components:
* `DialogPrimitive:Header`
* `DialogPrimitive:Description`
* `DialogPrimitive:Body`
* `DialogPrimitive:Footer`

I updated the existing `Modal` and `Flyout` components to use these new primitive components.

What makes sense to me to do next is to create a new type of component to satisfy this consumer request: https://docs.google.com/document/d/1qJ9YKVwfyVz5pi-ooIv3cC-AZp1GANJL7rIOJN9IIBA/edit

I would call this new component `Split Window Dialog`. It would be basically like the `Flyout` but would be docked within a right-side slot added to the App Frame instead of overlaying other content.

This is what makes the most sense to me and what I think consumers would find most useful.

However, from what I think you (Cristiano) were saying, you don't think we should create a new `Split Window Dialog` component which is like the `Modal` and `Flyout` components? 

What would we do instead though? Create a shared `Layout` primitive component to contain the other primitives? IE: `DialogPrimitive:Layout` or something else?

KristinLBradley:
Note: I still need to double check the rendering change to the Modal from the last refactor...

didoo:
For context, re. the conversation started by @KristinLBradley above, we're continuing the exploration directly in code and we'll report here once we have come to a conclusion/decision.

packages/components/app/styles/components/modal.scss
KristinLBradley: I can change it if you think that would be better

packages/components/app/styles/components/dialog-primitive.scss
```diff
@@ -0,0 +1,119 @@
...
+// DIALOG PRIMITIVE
+//
+
+
+// MAIN CONTAINER (<dialog>)
+
+// :where reduces the specificity of the selector
+// & can be used to set “loose” defaults which can be easily overridden
+dialog:where(.hds-dialog-primitive) {
+  all: unset; // removes default dialog & inherited styles. Must be the first style to be applied
```
KristinLBradley, line 16: Using `all: unset` causes inherited styles such as border-box styling to be wiped out and also wipes out some default `dialog` styles the `Modal` & `Flyout` were relying upon so may still want to consider changing to use a `div` to control the layout vs. using `dialog`.

Let me know what you think @didoo

website/docs/components/flyout/partials/guidelines/guidelines.md
```diff
@@ -114,11 +114,13 @@ The Flyout header features several properties to better communicate the purpose
 
 **With title icon**
 
-<Hds::Flyout::Header @icon="info" @onDismiss={{this.noop}}>Title</Hds::Flyout::Header>
+<!-- TODO: Temporary, replace with static image of Flyout Header -->
```
KristinLBradley, line 117: @majedelass as part of refactors to the `Modal` & `Flyout` child components we'll have to replace these examples with images.

We'll need images for:
* Modal Header (with variations)
* Modal Footer (with variations)
* Flyout Header (with variations)
* Flyout Description (The Header together with the Description)
* Flyout Footer (with variations)

Basically we will need images to replace examples of the above child components on these Website pages:
* https://helios.hashicorp.design/components/modal (Guidelines tab)
* https://helios.hashicorp.design/components/flyout (Guidelines tab)

showcase/tests/integration/components/hds/flyout/index-test.js
```diff
@@ -148,7 +150,7 @@ module('Integration | Component | hds/flyout/index', function (hooks) {
     await render(
       hbs`<Hds::Flyout id="test-flyout" as |F|><F.Header>Title</F.Header></Hds::Flyout>`
     );
-    assert.dom('button.hds-flyout__dismiss').isFocused();
+    assert.dom('button.hds-dialog-primitive__dismiss').isFocused();
```
didoo, line 153: This was failing because the `@extraClass` was not passed down. Once that is added, you can undo these changes in this test.

showcase/app/styles/showcase-pages/dialog-primitive.scss
```diff
@@ -0,0 +1,25 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// CODE-BLOCK
```
didoo, line 6:
```suggestion
// DIALOG PRIMITIVE
```

showcase/app/styles/showcase-pages/dialog-primitive.scss
```diff
@@ -0,0 +1,25 @@
...
+      outline: 1px dotted #d3d3d3;
+    }
+  }
+
+  .shw-component-dialog-primitive-base-container {
+    width: 400px;
+    height: 500px;
+    outline: 1px dotted #d3d3d3;
+  }
+
```
didoo, line 24:
```suggestion
```

showcase/tests/integration/components/hds/dialog-primitive/body-test.js
```diff
@@ -0,0 +1,41 @@
...
+
+  test('it renders the passed in content', async function (assert) {
+    await render(
+      hbs`
+        <Hds::DialogPrimitive::Body id="test-body">
+            Body
+        </Hds::DialogPrimitive::Body>
+      `
+    );
+    assert.dom('.hds-dialog-primitive__body').exists().hasText('Body');
```
didoo, line 39: we already know it exists, we tested it above (it's implicit)

```suggestion
    assert.dom('.hds-dialog-primitive__body').hasText('Body');
```

showcase/tests/integration/components/hds/dialog-primitive/description-test.js
```diff
@@ -0,0 +1,49 @@
...
+      await render(
+        hbs`
+        <Hds::DialogPrimitive::Description>
+          Description
+        </Hds::DialogPrimitive::Description>
+      `
+      );
+      assert
+        .dom('.hds-dialog-primitive__description')
+        .exists()
```
didoo, line 45:
```suggestion
```

showcase/tests/integration/components/hds/dialog-primitive/footer-test.js
```diff
@@ -0,0 +1,45 @@
...
+
+    test('it renders the passed in content', async function (assert) {
+      await render(
+        hbs`
+        <Hds::DialogPrimitive::Footer>
+          <Hds::Button type="submit" @text="Primary" />
+        </Hds::DialogPrimitive::Footer>
+      `
+      );
+      assert.dom('.hds-dialog-primitive__footer').exists();
```
didoo, line 41:
```suggestion
```

showcase/tests/integration/components/hds/dialog-primitive/header-test.js
```diff
@@ -0,0 +1,74 @@
...
+        <Hds::DialogPrimitive::Header>
+          Title
+        </Hds::DialogPrimitive::Header>
+      `
+      );
+      assert.dom('.hds-dialog-primitive__title').exists();
+      assert.dom('.hds-dialog-primitive__title').hasText('Title');
+      assert.dom('.hds-dialog-primitive__icon').doesNotExist();
+      assert.dom('.hds-dialog-primitive__tagline').doesNotExist();
+      assert.dom('.hds-flyout__description').doesNotExist();
```
didoo, line 43: leftover
```suggestion
```

showcase/tests/integration/components/hds/dialog-primitive/header-test.js
```diff
@@ -0,0 +1,74 @@
...
+      `
+      );
+      assert.dom('.hds-dialog-primitive__title').exists();
+      assert.dom('.hds-dialog-primitive__title').hasText('Tagline Title');
+      assert.dom('.hds-dialog-primitive__icon.flight-icon-info').exists();
+      assert.dom('.hds-dialog-primitive__tagline').exists();
+      assert.dom('.hds-dialog-primitive__tagline').hasText('Tagline');
+    });
+
+    // DISMISS
```
didoo, line 61: we should also add a test to make sure that the `onDismiss` callback is invoked

didoo (commented):
@KristinLBradley I did a first pass, and we're in a **very** good position 🙌

I've reviewed 
- Template/JavaScript files for components (DialogPrimitive, Flyout, Modal)
- Integration tests
- Showcase

Tomorrow I'll review the CSS changes (and the `unset: all`)

showcase/tests/integration/components/hds/flyout/index-test.js
KristinLBradley: _**TODO:** (for Kristin) revert change to test_

showcase/tests/integration/components/hds/dialog-primitive/header-test.js
KristinLBradley: @KristinLBradley **TODO**

packages/components/src/styles/components/dialog-primitive.scss
```diff
@@ -0,0 +1,121 @@
...
+
+.hds-dialog-primitive__wrapper-body {
+  flex: 1 1 auto;
+  // this is necessary, otherwise if there are words longer than the width of the available space for the list "item" content
+  overflow-x: hidden;
+  // we want the content to vertically scroll if needed
+  overflow-y: auto;
+  overscroll-behavior: contain;
+  // do we want to move the "border" logic here?
+  // border-top: 1px solid var(--token-color-border-primary);
```
didoo, line 49: at this point, since we've moved here also the background, I think is better to move also the borders

packages/components/src/styles/components/dialog-primitive.scss
```diff
@@ -0,0 +1,121 @@
...
+}
+
+.hds-dialog-primitive__wrapper-footer {
+  flex: none;
+  // do we want to move the "border" logic here?
+  // border-top: 1px solid var(--token-color-border-primary);
```
didoo, line 55: same here

packages/components/src/styles/components/dialog-primitive.scss
```diff
@@ -0,0 +1,121 @@
...
+
+
+// HEADER
+
+.hds-dialog-primitive__header {
+  display: flex;
+  gap: 16px;
+  align-items: flex-start;
+  padding: 16px 24px;
+  color: var(--token-color-foreground-strong); // default color
```
didoo, line 68:
```suggestion
  color: var(--token-color-foreground-strong); // default color (applied to both the title and the icon)
```

packages/components/src/styles/components/dialog-primitive.scss
```diff
@@ -0,0 +1,121 @@
...
+.hds-dialog-primitive__description {
+  padding: 0 24px 16px;
+}
+
+
+// BODY
+
+.hds-dialog-primitive__body {
+  padding: 24px;
+  border-top: 1px solid var(--token-color-border-primary);
```
didoo, line 103: see comment above about the body wrapper

packages/components/src/styles/components/dialog-primitive.scss
```diff
@@ -0,0 +1,121 @@
...
+}
+
+
+// FOOTER
+
+.hds-dialog-primitive__footer {
+  padding: 16px 24px;
+  background: var(--token-color-surface-faint);
+  border-top: 1px solid var(--token-color-border-primary);
```
didoo, line 112: see comment above about the footer wrapper

packages/components/src/styles/components/dialog-primitive.scss
```diff
@@ -0,0 +1,121 @@
...
+  padding: 24px;
+  border-top: 1px solid var(--token-color-border-primary);
+}
+
+
+// FOOTER
+
+.hds-dialog-primitive__footer {
+  padding: 16px 24px;
+  background: var(--token-color-surface-faint);
```
didoo, line 111: if we move the background style at wrapper level for the header, we should do the same for the footer

packages/components/src/styles/components/flyout.scss
```diff
@@ -17,8 +17,10 @@
 }
 
 .hds-flyout {
-  z-index: 49; // modal component is 50; we want to make sure the flyout always stays below the modal, regardless the DOM order
+  all: revert; // this must be first
```
didoo, line 20: in reality I don't think we need to revert everything. the only thing we need to apply are the `top/right/bottom: 0` styles, no?

packages/components/src/styles/components/flyout.scss
```diff
@@ -27,6 +29,10 @@
   border: none;
   box-shadow: 0 2px 3px 0 #3b3d4540, 0 12px 24px 0 #3b3d4559; // TODO: convert to token
 
+  &:not([open]) {
+    display: none;
+  }
+
   &[open] {
     position: fixed;
```
didoo, line 37: I think we can move the `position: fixed` at parent (`.hds-flyout`) level

packages/components/src/styles/components/flyout.scss
```diff
@@ -27,6 +29,10 @@
   border: none;
   box-shadow: 0 2px 3px 0 #3b3d4540, 0 12px 24px 0 #3b3d4559; // TODO: convert to token
 
+  &:not([open]) {
```
didoo, line 32: I would expect this to come _after_ `&[open]`

packages/components/src/styles/components/modal.scss
```diff
@@ -8,13 +8,18 @@
 //
 
 .hds-modal {
+  all: revert; // this must be first
   z-index: 50;
-  flex-direction: column;
+  box-sizing: inherit;
```
didoo, line 13: I don't think you need this 🤔

packages/components/src/styles/components/modal.scss
```diff
@@ -8,13 +8,18 @@
 //
 
 .hds-modal {
+  all: revert; // this must be first
```
didoo, line 11: same comment as for the `Flyout`, maybe is better to just set the fixed position (even if I tried and it's not so obvious, I'll get back to this in a following comment)

packages/components/src/styles/components/modal.scss
```diff
@@ -101,26 +58,19 @@
...
-  }
-
-  .hds-modal__tagline {
-    color: var(--token-color-foreground-faint);
-  }
-}
+// Note ".hds-modal--color-neutral" uses default color and background-color styles from the dialog-primitive
 
 .hds-modal--color-warning {
   .hds-modal__header {
```
didoo, line 64: If in the future we decide to yield also the `Description` in the Modal, this will happen:
screenshot_3479

This tells me we should apply the background and border to the wrappers "containers", not the header/body/footer "content"

packages/components/src/styles/components/modal.scss
```diff
@@ -101,26 +58,19 @@
...
-    border-bottom: 1px solid var(--token-color-border-primary);
-  }
-
-  .hds-modal__tagline {
-    color: var(--token-color-foreground-faint);
-  }
-}
+// Note ".hds-modal--color-neutral" uses default color and background-color styles from the dialog-primitive
 
 .hds-modal--color-warning {
```
didoo, line 63: General idea: what if we use dedicated (local) CSS variables, instead of overriding colors using CSS selectors? In this way we don't enter the weeds of using `dialog-primitive` selectors inside a `modal` CSS file. See for example how we're doing for the Dropdown component https://github.com/hashicorp/design-system/blob/main/packages/components/src/styles/components/dropdown.scss#L423-L454 (and we should do also in other parts too, eg. the Alert component; but that's another story for the future :) )

didoo (commented):
Left a bunch of comments on the CSS files. Nothing major, though.

alex-ju (commented):
Did a first pass on this PR. I believe abstracting the subcomponents is great.
I see the idea behind `all` `unset/revert`, but I'm afraid it could make the `<dialog>` functionality brittle. Have we considered having an argument on the `Hds::DialogPrimitive` that dictates the type of the root element? For example `<Hds::DialogPrimitive>` could generate a `<div>` while `<Hds::DialogPrimitive @modal={{true}}>` generates a `<dialog>` (the naming is just a suggestion and it could be improved, I'm sure)

showcase/tests/integration/components/hds/flyout/index-test.js
KristinLBradley: I'll have to refactor the `@extraClass` to use as a variable for the "component" name (such as "dialog-primitive", "flyout", or "modal")...

Should probably rename it then too since then the name "extraClass" won't be clear. Maybe "componentName"?

KristinLBradley: Actually maybe it can be simpler. Still need to work through it...

packages/components/src/styles/components/flyout.scss
didoo: after discussion with @alex-ju this morning and with @KristinLBradley this evening, we think there is a possible path forward to continue using the `<dialog>` HTML tag even if it comes with "baggage" (the user-agent CSS, the JS methods, the `::backdrop` pseudo-element, that we don't use btw)
- make sense to use the `all: unset;` declaration in the dialog primitive
- instead of using the `all: revert` in the Modal and Flyout, we can add the "missing styles" for the fixed layout, that is the `inset: 0`, the `width/height/max-height/max-width` where necessary (depends if it's a Flyout or a Modal), etc.

Once these changes are done, we can have a final look and see if we all agree to continue using the `<dialog>` for this component.

showcase/tests/integration/components/hds/flyout/index-test.js
KristinLBradley: I think maybe we could move the @extraClass to the container index.hds and then pass it down to the child components so for example if you set the extraClass to "hds-flyout" it could get appended to child component classes.

Or am I thinking about this wrong? @didoo

didoo: Yes, `extraClass` is a way to bypass a limitation in how Ember does the `yield` in templates, and pass an extra class name to a (yielded) component. It's not a way to provide classnames prefixes or similar.

No looking at the suggestion above, I may have added the comment in the wrong place: this is a button, so there's no way to pass a custom extraClass to it. Sorry for the conclusion.

packages/components/app/styles/components/dialog-primitive.scss
didoo: conversation continued here: https://github.com/hashicorp/design-system/pull/1918/files#r1482873567

packages/components/src/styles/components/flyout.scss
KristinLBradley: Cleaning this up as part of related refactors, thanks

packages/components/src/styles/components/modal.scss
KristinLBradley: Yeah looks like you are right. I was having issues with boxsizing still so added that but it could have been that the style change hadn't compiled correctly or something when I last checked. (I keep having to rebuild and restart Showcase and Web to see changes and/or fix errors that occur.)

packages/components/src/styles/components/modal.scss
KristinLBradley: Updated

showcase/app/templates/index.hbs
```diff
@@ -96,6 +96,11 @@ SPDX-License-Identifier: MPL-2.0
           Copy::Snippet
         </LinkTo>
       </li>
+      <li>
+        <LinkTo @route="components.dialog-primitive">
+          DialogPrimitive
+        </LinkTo>
+      </li>
```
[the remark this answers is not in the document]
KristinLBradley: It may be renamed.

showcase/tests/integration/components/hds/dialog-primitive/header-test.js
KristinLBradley: I focused on CSS updates to start will come back to tests tomorrow.

packages/components/src/styles/components/modal.scss
KristinLBradley: I will look at this more closely, thanks

showcase/app/templates/index.hbs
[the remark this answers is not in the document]
alex-ju: fair

extraclass:
You sent me a dozen emails and I have tried to unsubscribe with no luck.
Please stop.

I have no idea what you are talking about!  Please stop!

packages/components/src/styles/components/flyout.scss
```diff
@@ -17,18 +17,20 @@
...
   flex-direction: column;
+  box-sizing: inherit;
   height: 100vh;
   max-height: 100vh;
   margin: 0;
   padding: 0;
   background: var(--token-color-surface-primary);
   border: none;
   box-shadow: 0 2px 3px 0 #3b3d4540, 0 12px 24px 0 #3b3d4559; // TODO: convert to token
+  inset: 0;
```
didoo, line 31: nitpick: I would move it under the `z-index` declarations so it's near the "position/layout" block (they're related)

packages/components/src/styles/components/modal.scss
```diff
@@ -8,16 +8,18 @@
...
   z-index: 50;
-  flex-direction: column;
+  height: fit-content;
+  margin: auto;
   padding: 0;
   background: var(--token-color-surface-primary);
   border: none;
   border-radius: 8px;
   box-shadow: var(--token-surface-overlay-box-shadow);
+  inset: 0;
```
didoo, line 20: nitpick: I would move it under the `z-index` declarations so it's near the "position/layout" block (they're related)

packages/components/src/styles/components/dialog-primitive.scss
```diff
@@ -0,0 +1,121 @@
...
+// MAIN CONTAINER (<dialog>)
+
+// Since the `:where()` selector has a specificity of `0` (see https://developer.mozilla.org/en-US/docs/Web/CSS/:where)
+// the styles declared below are applied to the `.hds-dialog-primitive` element, but are overwritten by any other style applied via classname
+:where(.hds-dialog-primitive) {
+  all: unset; // removes default dialog & inherited styles. Must be the first style to be applied
+  // unsetting all the styles causes it to not inherit the box-sizing from the parent
+  box-sizing: border-box;
+
+  &:not([open]) {
```
didoo, line 20: Why the `&[open]` is declared in the CSS for `Flyout` and `Modal` instead of here? 🤔

packages/components/src/styles/components/modal.scss
```diff
@@ -8,16 +8,18 @@
...
+  position: fixed;
   z-index: 50;
-  flex-direction: column;
+  height: fit-content;
+  margin: auto;
   padding: 0;
   background: var(--token-color-surface-primary);
   border: none;
   border-radius: 8px;
   box-shadow: var(--token-surface-overlay-box-shadow);
```
didoo, line 19: reminder: let's abstract also the `Dialog::Overlay` component, it's shared in both the `Modal` and the `Flyout`

didoo (commented):
Left a few comments, mainly around the CSS

website/docs/components/flyout/partials/guidelines/guidelines.md
didoo: @KristinLBradley can you sync with @majedelass to explain them what needs to be done here? thanks

didoo:
For future context: we decided to re-organize our "dialog-related" components following this new structure:
- Dialog
   - Dialog (the primitive)
   - Modal
   - Flyout
But after doing all the refactoring, we have realized that unfortunately the CSS class names (and component names) of the `Modal` and `Flyout` were heavily used across the consumers codebases (see [here](https://github.com/search?q=hds-modal+user%3Ahashicorp+-repo%3Ahashicorp%2Fdesign-system+-repo%3Ahashicorp%2Fdesign-system-metrics&type=code&ref=advsearch) and [here](https://github.com/search?q=hds-flyout+user%3Ahashicorp+-repo%3Ahashicorp%2Fdesign-system+-repo%3Ahashicorp%2Fdesign-system-metrics&type=code&ref=advsearch)) 

See also Slack thread [here](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1707477336887229?thread_ts=1707336051.967529&cid=C025N5V4PFZ) and [here](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1707520082676669?thread_ts=1707336051.967529&cid=C025N5V4PFZ).

alex-ju (commented):
Looking good overall! I spotted a few visual differences that may have been discussed and agreed with @majedelass, but wanted to make sure we're aware of them.

The Flyout header is now `var(--token-color-surface-faint)` where previously was inheriting `var(--token-color-surface-primary)` from the root element (left is before, right is after).

flyout-regression

Icons and Title in Modal header do not inherit the color anymore (as the Tagline does).

modal-regression

From an API perspective, in other instances across the library the `@extraClass` argument is named `@contextualClass` – should we try and align them or do you think they differ in scope?

P.S. The long content modal example is a great addition! 👏

didoo:
> The Flyout header is now `var(--token-color-surface-faint)` where previously was inheriting `var(--token-color-surface-primary)` from the root element (left is before, right is after).

Yes, this is intentional.

> Icons and Title in Modal header do not inherit the color anymore (as the Tagline does).

No, this was a regression. Fixed.

> From an API perspective, in other instances across the library the `@extraClass` argument is named `@contextualClass` – should we try and align them or do you think they differ in scope?

Not sure why I thought it was called `@extraClass`. In any case, fixed.

> P.S. The long content modal example is a great addition! 👏

Yeah, we missed that use case (and it bit me)

alex-ju (approved):

packages/components/src/styles/components/dialog-primitive.scss
```diff
@@ -0,0 +1,136 @@
...
+  flex-grow: 1;
+}
+
+.hds-dialog-primitive__tagline {
+  margin-bottom: 4px;
+  color: var(--token-color-foreground-faint); // default color
+}
+
+.hds-dialog-primitive__dismiss {
+  align-self: center;
```
majedelass, line 90: We should top align the dismiss button (instead of center). Upon further investigation, it looks like the current figma implementation of modal and flyout had different behaviors (center for modal, top aligned for flyout). I'm aligning them to the top in this branch: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/Yo8J7OzRaIBA4ytBPPDEqh/HDS-Product---Components?type=design&node-id=53258%3A135&mode=design&t=gGHcTGqSRTFmxDVT-1

packages/components/src/styles/components/dialog-primitive.scss
```diff
@@ -0,0 +1,136 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+//
+// DIALOG PRIMITIVE
+//
+
+$hds-dialog-primitive-horizontal-padding: 24px;
```
majedelass, line 10: This is a nit pick (and also still trying to understand naming conventions here), but since we use 24px for both horizontal and vertical (like the body padding) shouldn't we use a naming convention that reflects maybe something more agnostic (like maybe md and lg instead of horizontal/vertical-padding)?

majedelass (changes requested):
First time, please let me know if there are other ways of doing a review (or more efficient/better).

I appreciate this amazing work 🙌

packages/components/src/styles/components/dialog-primitive.scss
KristinLBradley: I'm with you but I don't think it's going to happen at least not anytime soon...

packages/components/src/styles/components/dialog-primitive.scss
KristinLBradley: I pushed an update

packages/components/src/styles/components/dialog-primitive.scss
didoo: @majedelass there's been a long ongoing discussion about spacing tokens. The answer is still no at the moment: no tokenization of spacing values. I'll send you in DM all the links for context.

packages/components/src/styles/components/dialog-primitive.scss
didoo: @KristinLBradley I don't see the commit 🤔

didoo: @majedelass @KristinLBradley this will cause a visual mis-alignment in the modal, when there's no tagline (which is the most common use case)

screenshot_3501

didoo (approved):
At this point everything seems OK

packages/components/src/styles/components/dialog-primitive.scss
majedelass: Yeah that's fine, that currently exists in the figma design too: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/Yo8J7OzRaIBA4ytBPPDEqh/HDS-Product---Components?type=design&node-id=22928%3A56204&mode=design&t=RIpe2DJSgIZS5TKY-1

majedelass: Also current day flyout behaves this way: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components?type=design&node-id=24759%3A104119&mode=design&t=WbToqy24HvmSKjqw-1

didoo: > current day flyout behaves this way

Not in code, from what I see:
screenshot_3503

@jorytindall do you remember if it was an intentional decision?

majedelass: Yeah both modal and flyout behaved differently _in figma_ and wanted to align them moving forward. I did discuss with the design team that it's beneficial for the dismiss button to align to the top for its placement to be predictable since it's an actional UI piece.

alex-ju: we had quite a bit of feedback from consumers on what they considered being 'misaligned' items so I would be surprised if they won't call this out

didoo: +1 on @alex-ju's comment

KristinLBradley: I pushed it to the wrong branch :( 
https://github.com/hashicorp/design-system/pull/1945/commits

didoo: @KristinLBradley [all designers agree on aligning the icon to the top](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1707847200682889). Can you port the change in this branch, so I can approve?

KristinLBradley: I moved the commit to this branch now.

majedelass:
I added images and created a new sub folder within website/public/assets for dialog-primitives for shared assets. Wanted to confirm that everyone was ok with that?

majedelass (approved):
Everything looks good to go from my side. Thanks for the great work, support and guidance y'all :D

didoo:
> Is that a style we are trying to standardize for documentation? If it is, then maybe we should go down that path.

@majedelass maybe sync with @jorytindall on this (he did **a lot** of these illustrations!)

didoo:
For context: I have moved a commit that @majedelass made to replace the header/body/footer component instances with static images ([context](https://github.com/hashicorp/design-system/pull/1918/files#r1481892451)) to the documentation branch (https://github.com/hashicorp/design-system/pull/1945/commits/f100a1a657e0abc3400dad3801aaddc9589296c2) to avoid any confusion of where the documentation changes should go (this was a special case, so having this specific fix in this branch was OK too).

website/docs/components/flyout/partials/guidelines/guidelines.md
didoo: Resolving this (done in https://github.com/hashicorp/design-system/pull/1945/commits/f100a1a657e0abc3400dad3801aaddc9589296c2)

didoo (approved):
Good to go 👍

jorytindall (approved):
Nothing major, but one question for my own sanity, should some of the text elements be using more semantic elements rather than divs? Does it matter? I only ask because there are couple of different methods:

- The `Description` uses a div with the typography class helpers
- The `Title` and `Tagline` use the Text component with a div tag defined.

If that isn't a problem or has already been considered, please ignore this, just something that stood out to me.

website/docs/utilities/dialog-primitive/partials/guidelines/overview.md
```diff
@@ -0,0 +1,7 @@
+The DialogPrimitive contains the building blocks used to contruct dialog-based components and consists of:
```
heatherlarsen, line 1:
```suggestion
The DialogPrimitive contains the building blocks used to construct dialog-based components and consists of:
```

heatherlarsen (approved):
I found one typo, otherwise the documentation looks good!

KristinLBradley:
> Nothing major, but one question for my own sanity, should some of the text elements be using more semantic elements rather than divs? Does it matter? I only ask because there are couple of different methods:
> 
> * The `Description` uses a div with the typography class helpers
> * The `Title` and `Tagline` use the Text component with a div tag defined.
> 
> If that isn't a problem or has already been considered, please ignore this, just something that stood out to me.

I prefer using semantic elements wherever possible/practical but for these components I was just taking what had already been implemented for the `Modal` & `Flyout` and moved it to shared components so I didn't change any of the DOM really.

didoo:
> > Nothing major, but one question for my own sanity, should some of the text elements be using more semantic elements rather than divs? Does it matter? I only ask because there are couple of different methods:
> > 
> > * The `Description` uses a div with the typography class helpers
> > * The `Title` and `Tagline` use the Text component with a div tag defined.
> > 
> > If that isn't a problem or has already been considered, please ignore this, just something that stood out to me.
> 
> @jorytindall I prefer using semantic elements wherever possible/practical but for these components I was just taking what had already been implemented for the `Modal` & `Flyout` and moved it to shared components so I didn't change any of the DOM really.

@jorytindall @KristinLBradley unless we use a `<p>` element for all of them (title, caption and description), we can't use headings, because which heading level is the "right" one, to respect the order in the page, since we don't know (being primitives) where they will be injected/consumed, what is their context?

KristinLBradley:
> > > Nothing major, but one question for my own sanity, should some of the text elements be using more semantic elements rather than divs? Does it matter? I only ask because there are couple of different methods:
> > > 
> > > * The `Description` uses a div with the typography class helpers
> > > * The `Title` and `Tagline` use the Text component with a div tag defined.
> > > 
> > > If that isn't a problem or has already been considered, please ignore this, just something that stood out to me.
> > 
> > 
> > @jorytindall I prefer using semantic elements wherever possible/practical but for these components I was just taking what had already been implemented for the `Modal` & `Flyout` and moved it to shared components so I didn't change any of the DOM really.
> 
> @jorytindall @KristinLBradley unless we use a `<p>` element for all of them (title, tagline and description), we can't use headings, because which heading level is the "right" one, to respect the order in the page, since we don't know (being primitives) where they will be injected/consumed, what is their context? or were you thinking of other HTML semantic tags?

Yes, there would be an issue using heading tags due to hierarchy. I do prefer to use `p` tags to mark up pieces of text vs. divs but I don't think it would add much here so don't think there's a strong enough reason to change the markup from what it was before.

didoo:
Now that we have #2213 we can close this one

PR #1920, opened 2024-02-01
Improved resilience of SVG sprite loading script

alex-ju:
### :pushpin: Summary

When testing the `lazyEmbed` solution in `website` (as part of the v2 addon testing) we discovered that the script assumes `config` and `document` are always present, which is not always the case (in various environments, such as tests and server-side rendering they might not). For example, when building `website` with prember this script fails execution without the few safeguards added in this commit.

This commit is ported from https://github.com/hashicorp/design-system/commit/67d42e0d7e6da585c619d4f88b988de1e92012e3.

***

### 👀 Component checklist

### DIFF
--- a/.changeset/silly-ducks-decide.md
+++ b/.changeset/silly-ducks-decide.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/ember-flight-icons": patch
+---
+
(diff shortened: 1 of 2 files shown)


Dhaulagiri:
👍  thanks for fixing this

PR #1924, opened 2024-02-01
Delete .github/workflows/update-search-index.yml

Dhaulagiri:
I added this but, we don't use this. Instead, this happens as part of our Vercel builds now.


didoo (approved):
👍

PR #1925, opened 2024-02-01
trigger ci when github actions change

Dhaulagiri:
### :pushpin: Summary

I realized that with the forthcoming automated tsccr updates we'll need to ensure that tests are run in those PRs to validate the actions still work. This PR adds each ci workflow to the files checked to see if each respective set of tests should run or not.

### DIFF
--- a/.github/workflows/ci-components.yml
+++ b/.github/workflows/ci-components.yml
@@ -25,7 +25,7 @@ jobs:
           fetch-depth: 0  
       - name: Get changed files
         id: read-files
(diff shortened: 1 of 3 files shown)


MelSumner (approved):
:shipit: 

(TBH I feel like this is something GitHub should do automatically? IDK.)

Dhaulagiri:
> I feel like this is something GitHub should do automatically?

_nod_ Ideally this would be a lot easier

PR #1931, opened 2024-02-06
run tsccr update

Dhaulagiri:
### :pushpin: Summary

I set this up to happen [automatically](https://github.com/hashicorp/security-tsccr/pull/924) monthly, however we missed the 1st of the month boat so it won't run automatically for a few more weeks. This runs it [manually](https://github.com/hashicorp/security-tsccr/blob/main/tsccr-helper/help/cmd.gha-update.md) to confirm that the configuration I used in the automated PR is what we want and that the [ci changes I made](https://github.com/hashicorp/design-system/pull/1925) also work, so we might as well ship it now to avoid the deprecation warnings we're seeing.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -9,7 +9,7 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/labeler@ac9175f8a1f3625fd0d4fb234536d26811351594 # v4.3.0
+    - uses: actions/labeler@8558fd74291d67161a8a78ce36a881fa63b766a9 # v5.0.0
       with:
         # Token setup in hashibot-hds' account
         repo-token: "${{ secrets.PAT_TOKEN }}"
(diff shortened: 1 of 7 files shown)


alex-ju (approved):

.github/workflows/labeler.yml
```diff
@@ -9,7 +9,7 @@ jobs:
       pull-requests: write
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/labeler@ac9175f8a1f3625fd0d4fb234536d26811351594 # v4.3.0
+    - uses: actions/labeler@8558fd74291d67161a8a78ce36a881fa63b766a9 # v5.0.0
```
alex-ju, line 12: @Dhaulagiri this requires changes in our setup [it seems](https://github.com/hashicorp/design-system/actions/runs/7813047503/job/21311381094)

Dhaulagiri: 🤕 oops. I'll take a peek and get it fixed

PR #1932, opened 2024-02-07
Fix labeler action

Dhaulagiri:
### :pushpin: Summary

Updates our labeler setup to use the updated format that comes with v5 which we updated to in #1931 . We were not alone in being surprised by this change https://github.com/actions/labeler/issues/712

Note: I've included a commit for testing in this PR that I will drop after things work correctly

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,12 +1,24 @@
 docs-website:
-  - website/docs/**/*
+  - changed-files:
+    - any-glob-to-any-file:
+      - website/docs/**/*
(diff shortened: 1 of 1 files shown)


Dhaulagiri:
It looks like this did work [here](https://github.com/hashicorp/design-system/actions/runs/7816493698/job/21322143814?pr=1932). Once this is approved I'll clean up the commits.

alex-ju (approved):
In alignment with the docs. Thanks for the quick investigation!

PR #1936, opened 2024-02-07
Add a Figma changelog entry for the 4.0 release

jorytindall:
:pushpin: Summary

Added a couple of changelog entries, including:

- Changes and additions in the `Table` components, including breaking changes
- Changes in the `Alert` and `Toast` components, including breaking changes
- Retroactively adding an entry for the `IconTile` with the addition of the Vault Radar logo.

_Is this everything, or is there something I missed?_

I put a placeholder in the date for now (and marked this as a draft) until we actually settle on a 4.0 release date.

_Also:_ Appears that there were several empty spaces that were removed at the end of lines and carriage returns added by Prettier. Doesn't appear to be anything problematic, but I can revert these if necessary.

***

### DIFF
--- a/packages/components/CHANGELOG-FIGMA-COMPONENTS.md
+++ b/packages/components/CHANGELOG-FIGMA-COMPONENTS.md
@@ -1,5 +1,25 @@
 # [HDS Product - Components](https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components?type=design&node-id=6790-10926&mode=design&t=Ps0aMGZ6F3z7bAJ4-0)
 
+## Date TBD
+
+`Alert` and `Toast` - Updated the `StandaloneLink` to the secondary variant from the primary variant to align with our documented usage recommendations.
+
+`Table` - multiple enhancements to the `Table` components including:
+
+- Added support for multi-selection with `Header Column / Selection` and `Cell / Selection`.
+- Added support for a `Tooltip` in the `Header Column` component.
+- Updated the visual language of the sort button in `Header Column` by decoupling the sort functionality into a nested `Sort Button`
+
+### Breaking changes
+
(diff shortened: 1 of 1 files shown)


packages/components/CHANGELOG-FIGMA-COMPONENTS.md
```diff
@@ -1,5 +1,25 @@
...
+
+- Added support for multi-selection with `Header Column / Selection` and `Cell / Selection`.
+- Added support for a `Tooltip` in the `Header Column` component.
+- Updated the visual language of the sort button in `Header Column` by decoupling the sort functionality into a nested `Sort Button`
+
+### Breaking changes
+
+`Table` - Adding support for a `Tooltip` and updates to the `Sort Button` result in a breaking change in sortable variants.
+
+`Alert` and `Toast` - Updating the `StandaloneLink` to the secondary variant will revert any changes you've made to the text and leading icon properties.
```
heatherlarsen, line 17: [suggestion] These aren't the most informative extra details about the breaking change. In addition to stating why and how this is a breaking change, we should try to include guidance/tips for how to help them make updates and reduce the "pain" this may cause them. 

For example, for the text that will be reverted due to the Alert change, one suggestion could be that before they've accepted the changes in Figma, they add a comment near every alert in their active or delivered (eng still referencing them) files with the text and icon name so they don't have to remember what every instance was when they have to go back and update them.

heatherlarsen (commented):
Added a couple of suggestions, one blocking.

packages/components/CHANGELOG-FIGMA-COMPONENTS.md
jorytindall: Added a bit of clarification and and some steps to avoid/ease the pain of making these updates.

heatherlarsen (approved):

PR #1939, opened 2024-02-08
Fix dependencies watch

alex-ju:
### :pushpin: Summary

Fix dependencies watch by adjusting the rollup config.

Currently when building a package with a watcher (via `yarn start`) and starting `showcase` or `website` (via `ember s`) we often see build fails (TemplateCompiler, SassCompiler, or BroccoliMergeTrees) complaining certain updated files don't exist anymore. With this change we're being more specific about what changes should trigger a rebuild, avoiding such errors.

#### How to test

```bash
cd packages/components
yarn start
```
then, in a separate window/tab
```
cd showcase
ember s
```

change any file in `packages/components/src` or `packages/ember-flight-icons/src` and see the result in `showcase`

### DIFF
--- a/packages/components/rollup.config.mjs
+++ b/packages/components/rollup.config.mjs
@@ -78,4 +78,7 @@ export default {
     'ember-modifier',
     'prismjs',
   ],
+  watch: {
+    include: 'src/**',
+  },
 };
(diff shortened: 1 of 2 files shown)


didoo (approved):

alex-ju:
This doesn't seem to be working consistently 😞 will have to investigate further

PR #1941, opened 2024-02-12
Fix links after migration to v2 format

jgwhite:
### :pushpin: Summary

The GitHub link for the CodeBlock component was using the old `addon/` path rather than the new `src/` path.

### DIFF
--- a/website/docs/components/code-block/index.md
+++ b/website/docs/components/code-block/index.md
@@ -4,7 +4,7 @@ description: A container for displaying formatted chunks of code with syntax hig
 caption: A container for formatting chunks of code with syntax highlighting.
 links:
   figma: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components?type=design&node-id=44796%3A2786&mode=design&t=pZ5PoxE7Q7U0kDaa-1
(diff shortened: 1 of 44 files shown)


website/docs/components/code-block/index.md
```diff
@@ -28,4 +28,4 @@ navigation:
 
 <section data-tab="Accessibility">
   @include "partials/accessibility/accessibility.md"
-</section>
\ No newline at end of file
+</section>
```
jgwhite, line 31: Whoops, I made this change it GitHub’s editor, so I guess it either added or deleted a `\n`. Not sure which!

alex-ju:
Hey @jgwhite! Thanks for spotting this! We seem to have this in a few other places – would you mind if I push an extra commit to address them all at once?

alex-ju (approved):

website/docs/components/code-block/index.md
alex-ju: Yeah, we're not so much in sync with GitHub's default new line rules (as we expect then not to exist). Fixed it now

jgwhite:
> Hey @jgwhite! Thanks for spotting this! We seem to have this in a few other places – would you mind if I push an extra commit to address them all at once?

Be my guest!

PR #1946, opened 2024-02-13
Fix build errors when watching packages

alex-ju:
### :pushpin: Summary

This PR changes the Rollup config to skip the 'deep clean' done by `addon.clean()` when running Rollup build with watch for development purposes. This is to overcome a [known issue](https://discord.com/channels/480462759797063690/1066065904384876626/1066103413638705172) where the build fails when a `js` file is deleted (even for a fraction of a second, as it happens with Rollup). 

This is an alternative to #1939 which improved things, but some builds still failed (more commonly when js files were changed).

#### How to test

```bash
cd packages/components
yarn start
```

then, in a separate window/tab

```bash
cd showcase
ember s
```

changing any file in `packages/components/src` should trigger a Rollup rebuild and a `showcase` rebuild, without throwing any errors.

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -19,7 +19,7 @@
   "scripts": {
     "build": "rollup --config",
     "prepublishOnly": "rollup --config",
-    "start": "rollup --config --watch",
+    "start": "rollup --config --watch --environment development",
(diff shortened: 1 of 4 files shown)


didoo (approved):
For me, we can merge and try it.

PR #1954, opened 2024-02-15
Showcase - Add automatic anchor links generation for headings

didoo:
### :pushpin: Summary

Long overdue functionality for the web pages in the showcase.

In this PR I have:
- added logic and styling to automatically inject anchor links in page headings (h2-h6)

(most of the logic and styling has been ported and adapted from a similar implementation on the Helios website)

screenshot_3521

***

### 👀 Component checklist


showcase/app/styles/_globals.scss
```diff
@@ -168,6 +168,61 @@ body {
...
+  :hover > & {
+    &::after {
+      opacity: 1;
+    }
+  }
+
+  &:focus-visible {
+    &::after {
+      background-color: aliceblue;
+      border: 2px dashed blue;
```
KristinLBradley, line 217: Are these colors temporary for testing or would you want to replace them with more "official" colors?

KristinLBradley (commented):
This will be very handy I think. Thanks for doing this!

didoo:
@MelSumner any idea why [these tests](https://percy.io/HashiCorp/hds-components/builds/32681859?utm_campaign=HashiCorp&utm_content=hds-components&utm_source=github_status_private) are failing now? I haven't touched any component-related code in this PR 
screenshot_3523

showcase/app/components/shw/text/index.js
```diff
@@ -136,4 +137,33 @@ export default class ShwTextIndexComponent extends Component {
...
+      }
+    }
+    // update the heading
+    element.id = uniqueId;
+    element.classList.add('shw-page-heading-scroll-margin-top');
+    // inject an anchor element
+    const anchor = document.createElement('a');
+    anchor.href = `#${uniqueId}`;
+    anchor.className = 'shw-page-heading-link';
+    anchor.setAttribute('aria-labelledby', uniqueId);
```
MelSumner, line 166: I don't think you need this. The link is already getting an accessible name from the content of the link. By the time it renders to the page it has an accessible name.

MelSumner:
> @MelSumner any idea why [these tests](https://percy.io/HashiCorp/hds-components/builds/32681859?utm_campaign=HashiCorp&utm_content=hds-components&utm_source=github_status_private) are failing now? I haven't touched any component-related code in this PR 

@didoo it's possible that they weren't run before? I would check these components to see if the test failures are indeed correct and we missed something. If they are false positives, then maybe it's just a race condition and we should try running the tests again.

showcase/app/components/shw/text/index.js
didoo: @MelSumner can you double check (please 🙏) in the browser? 
https://hds-showcase-git-showcase-add-anchor-links-to-775174-hashicorp.vercel.app/
we do something very similar in the website, but in that case we use `aria-label` (because these are empty links, they don't wrap the text content).

MelSumner:
> > If they are false positives but running the tests again still results in failures, I'd need to dig deeper to figure out what's going on.
> 
> @MelSumner yes, I've re-run the tests multiple times, and they always fail. I'll try again, just in case, but if you see they fail again then please have a look at what it may be causing them.

@didoo so in the case of the app footer I can see where some of the failures are correct, what about the table? Are the checkboxes 24x24?

didoo:
> @didoo so in the case of the app footer I can see where some of the failures are correct, what about the table? Are the checkboxes 24x24?

No, they're 16x16.

MelSumner:
@didoo I think _technically_ they are still okay because of the padding around the element. It's likely we'll need to add some rule exemptions (with notes) and let our consumers know as well.

didoo:
> @didoo It's likely we'll need to add some rule exemptions (with notes)

@MelSumner where should I add these rule exemptions?

MelSumner:
@didoo for now we've been using inline disable/enable. That would be enough for me to ship this PR; I need to run tests on the component package itself because if we're having this failure it should be addressed at the component level.

didoo:
> @didoo for now we've been using inline disable/enable. That would be enough for me to ship this PR; I need to run tests on the component package itself because if we're having this failure it should be addressed at the component level.

@MelSumner since this PR is not touching anything related to these components, can you open a different PR to make sure these tests don't fail also in `main`, so I can rebase on that PR once merge, and keep this PR specific to the initial scope? Thanks

MelSumner:
@didoo I'm seeing a lot of failed tests, I'm going to have to work through the root of the failures and probably file some issues. I'll update our team in Slack.

didoo:
For future reference, here's the slack threads related to the investigation about the mysteriously failing tests (https://github.com/hashicorp/design-system/pull/1954#issuecomment-1952202174):
- https://hashicorp.slack.com/archives/C025N5V4PFZ/p1708630150902979
- https://hashicorp.slack.com/archives/C025N5V4PFZ/p1708634378389389

MelSumner (approved):
It's still super weird that there were failures so I'm going to troubleshoot more to figure it out. 👍

PR #1955, opened 2024-02-16
update ambassador references

Dhaulagiri:


heatherlarsen (approved):
Good catch! Thank you

PR #1958, opened 2024-02-22
Align component names in `website`

alex-ju:
### :pushpin: Summary

When it comes to letter casing we seem to be using capitals for each word (also referred to as 'Start Case') when we refer to our components in general (inherently comprising the Figma version and the Ember version).

We use `PascalCase` (and mark it as code, using a fixed-width style) when we refer to Ember components and subcomponents.

In this PR we align AppFooter and CodeBlock to the existing naming convention and split them apart, resulting in App Footer and Code Block. We also correct two instances where the button set does not have capitals.

In terms of impact, for the URLs in `website` we convert both 'Start Case' and 'PascalCase' to 'kebab-case', so operating these changes will not impact the URLs for these components.

<table>
<tr><td>

component-name-before

</td><td>

component-name-after

</td></tr>
</table>

***

### DIFF
--- a/website/docs/components/app-footer/index.md
+++ b/website/docs/components/app-footer/index.md
@@ -1,5 +1,5 @@
 ---
-title: AppFooter
+title: App Footer
 description: A footer that appears on every screen to display supplementary information and links.
 caption: Displays supplementary information and links for the application.
 links:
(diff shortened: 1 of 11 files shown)


MelSumner (approved):

heatherlarsen (approved):
Nice catch. Thanks for fixing these!

PR #1963, opened 2024-02-28
[bugfix] adds aria-label to the copy snippet button

MelSumner:
### :pushpin: Summary

If merged, this PR resolves the issue where there wasn't an appropriate `aria-label` on the copy snippet button. 
It also adds a test to ensure that it's there in the future.

### :link: External links

Jira ticket: [HDS-3146](https://hashicorp.atlassian.net/browse/HDS-3146)

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/components/hds/copy/snippet/index.hbs
+++ b/packages/components/src/components/hds/copy/snippet/index.hbs
@@ -6,6 +6,7 @@
   type="button"
   class={{this.classNames}}
   {{hds-clipboard text=@textToCopy onSuccess=this.onSuccess onError=this.onError}}
+  aria-label={{concat 'copy ' @textToCopy}}
   ...attributes
 >
   <Hds::Text::Code class="hds-copy-snippet__text" @tag="span" @size="100">
--- a/showcase/tests/integration/components/hds/copy/snippet/index-test.js
+++ b/showcase/tests/integration/components/hds/copy/snippet/index-test.js
@@ -39,6 +39,13 @@ module('Integration | Component | hds/copy/snippet/index', function (hooks) {
     assert.dom('#test-copy-snippet').hasClass('hds-copy-snippet');
   });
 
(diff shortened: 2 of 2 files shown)


MelSumner:
This is still in draft mode because I'm not sure how to do the changelog since switching to a v2 addon, so I need to check with our team for that. But the proposed fix itself can be reviewed.

packages/components/src/components/hds/copy/snippet/index.hbs
```diff
@@ -6,6 +6,7 @@
   type="button"
   class={{this.classNames}}
   {{hds-clipboard text=@textToCopy onSuccess=this.onSuccess onError=this.onError}}
+  aria-label={{concat 'copy ' @textToCopy}}
```
didoo, line 9: a comment in the [initial request](https://hashicorp.slack.com/archives/C7KTUHNUS/p1709145109837559) was:

>  There is no announcement that the text was copied to clipboard so I wonder if someone who isn't sighted might have the perspective that nothing happens when they click on the button?

should we consider also this?

didoo (commented):
left a comment

alex-ju:
> This is still in draft mode because I'm not sure how to do the changelog since switching to a v2 addon, so I need to check with our team for that. But the proposed fix itself can be reviewed.

no changes for the changelog process. running `yarn changeset` from the monorepo root, then selecting the relevant package (in this case `@hashicorp/design-system-components`)

packages/components/src/components/hds/copy/snippet/index.hbs
MelSumner: I think we should discuss it because it is a lot trickier to do. I would make a separate issue to track the conversation and potential implementations. But I don't think it should block this more immediate bugfix. WDYT?

didoo: works for me 👍

didoo (approved):
LGTM

PR #1964, opened 2024-02-29
`HDS PowerSelect`: Spike (HDS-3143)

KristinLBradley:
### :pushpin: Summary

_**Spike:** Not intended for merge_

Test build of HDS SuperSelect (PowerSelect) built on top of ember-power-select.

- **Showcase:** https://hds-showcase-git-hds-3143-power-select-compone-bc31d0-hashicorp.vercel.app/components/super-select

### :link: External links

- Jira ticket: [HDS-3143](https://hashicorp.atlassian.net/browse/HDS-3143)

***

### 👀 Component checklist

[HDS-3143]: https://hashicorp.atlassian.net/browse/HDS-3143?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/components/hds/super-select/index.hbs
```diff
@@ -0,0 +1,30 @@
...
+  }}
+  <PowerSelect
+    @ariaDescribedBy={{@ariaDescribedBy}}
+    @ariaInvalid={{@ariaInvalid}}
+    @ariaLabel={{@ariaLabel}}
+    @ariaLabelledBy={{@ariaLabelledBy}}
+    @disabled={{@disabled}}
+    @onChange={{@onChange}}
+    @options={{@options}}
+    @renderInPlace={{@renderInPlace}}
```
alex-ju, line 21: we know that some of these arguments should have a different default to make the compliant, for example, `renderInPlace` should be `true`. It's probably worth exploring changing such defaults in this spike.

packages/components/src/components/hds/super-select/index.hbs
```diff
@@ -0,0 +1,30 @@
...
+    @searchEnabled={{@searchEnabled}}
+    @selected={{@selected}}
+    @verticalPosition={{@verticalPosition}}
+    ...attributes
+    as |option term|
+  >
+    {{yield option term}}
```
alex-ju, line 28: is `term` meant to be used for a matcher function? if so, it would be good to have an example

alex-ju (commented):
thanks for drafting this so quickly, I left a few initial comments for your consideration

packages/components/src/components/hds/super-select/index.hbs
KristinLBradley: I'll just remove that for now. It was from an example I was looking at.

packages/components/src/styles/@hashicorp/design-system-power-select-overrides.scss
```diff
@@ -167,6 +167,7 @@
 
   .ember-power-select-options {
     padding: 6px 0;
+    list-style: none;
```
KristinLBradley, line 170: There was a stray bullet displaying.

packages/components/src/components/hds/super-select/index.hbs
```diff
@@ -0,0 +1,51 @@
...
+      @ariaLabel={{@ariaLabel}}
+      @ariaLabelledBy={{@ariaLabelledBy}}
+      @disabled={{@disabled}}
+      @onChange={{@onChange}}
+      @options={{@options}}
+      @renderInPlace={{true}}
+      @searchEnabled={{@searchEnabled}}
+      @selected={{@selected}}
+      @verticalPosition={{@verticalPosition}}
+      ...attributes
```
KristinLBradley, line 27: It seems to make more sense to spread attributes on these components vs. the wrapper div (Maybe I could also just get rid of the wrapper div too?)

showcase/tests/integration/components/hds/super-select/index-test.js
```diff
@@ -0,0 +1,29 @@
...
+module('Integration | Component | hds/super-select/index', function (hooks) {
+  setupRenderingTest(hooks);
+
+  // DISABLEtest('it should render the component with a CSS class that matches the component name', async function (assert) {
+  //   this.set('NOOP', () => {});
+  //   await render(
+  //     hbs`<Hds::SuperSelect @onChange={{this.NOOP}} id="test-super-select" />`
+  //   );
+  //   assert.dom('#test-super-select').hasClass('hds-super-select');
+  // });
```
KristinLBradley, line 20: Commented this out for now because it doesn't work with where I am currently spreading attributes...

packages/components/src/styles/components/super-select.scss
```diff
@@ -0,0 +1,303 @@
...
+  }
+}
+
+// VARIANTS
+// When renderInPlace=true, setting matchTriggerWidth=false has no effect unless we override the width of the dropdown content
+.hds-super-select--match-trigger-width-false {
+  .ember-power-select-dropdown.ember-basic-dropdown-content--in-place {
+      width: auto;
+  }
+}
```
KristinLBradley, line 303: Result:
image

packages/components/src/styles/@hashicorp/design-system-power-select-overrides.scss
alex-ju: indeed, this is something I've seen happening since we upgraded to `ember-power-select` version 7 – we should cherry-pick this and open a PR to fix it. clearly a disadvantage of providing the styles without being in control of the version

packages/components/src/components/hds/super-select/index.hbs
alex-ju: it would probably make sense to aply the attributes here and aim to remove the wrapper. in theory, we could develop a PowerSelect theme, instead of using overrides

KristinLBradley: I tried removing the wrapper but couldn't get it to apply a custom class name to the underlying components directly so far.

alex-ju:
@KristinLBradley now that we have a separate PR for the implementation I would close this one

PR #1966, opened 2024-02-29
Update Helios logo in main readme file

alex-ju:
### :pushpin: Summary

Suggesting a potential change for the header of our main readme file – using the one we have in our emailing template.

This is subjective and ultimately comes down to personal preference, but I thought it's worth sharing this minor proposal with the team. Share your thoughts or 👍 / 👎

#### Preview

[Current](https://github.com/hashicorp/design-system/blob/main/README.md) / [Proposed](https://github.com/hashicorp/design-system/blob/9bcb68039a40a2b9a54363cc6d6bce4ac4eb72bf/README.md) 

____

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,6 @@
+<img width="315" alt="helios-logo" src="https://github.com/hashicorp/design-system/assets/788096/da644ef0-f6eb-4070-8f10-b595be815a0d" role="presentation" alt="">
 
-<p>
-  <img width="315" alt="helios-logo" src="https://github.com/hashicorp/design-system/assets/788096/da644ef0-f6eb-4070-8f10-b595be815a0d" role="presentation" alt="">
-</p>
-
-<p>The Helios Design System provides the building blocks to design and implement consistent, accessible, and delightful product experiences across HashiCorp.</p>
+The Helios Design System provides the building blocks to design and implement consistent, accessible, and delightful product experiences across HashiCorp.
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -1,8 +1,9 @@
-<p align="center">
-  <img src="https://user-images.githubusercontent.com/1672302/213473448-ac8cd0f7-cde2-49ef-b8ef-4067677569f2.jpg" role="presentation" alt="">
+
+<p>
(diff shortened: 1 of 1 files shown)


KristinLBradley (approved):
Looks good to me

jorytindall:
No overly opinionated on this, but in a dark theme (or dark background) this looks a little strange to me. There's probably not a way to conditionally show something based on theme in this case and I don't want to overthink it, just noting it here though.
Screenshot 2024-02-29 at 10 36 26 AM

alex-ju:
> No overly opinionated on this, but in a dark theme (or dark background) this looks a little strange to me. There's probably not a way to conditionally show something based on theme in this case and I don't want to overthink it, just noting it here though.

I had the same thought just a few minutes ago – this would be valuable especially as the dark theme is now the default one in GitHub..

jorytindall:
@alex-ju I did some quick poking around and Supabase actually does this in their readme https://github.com/supabase/supabase

```
<p align="center">
  <img src="https://user-images.githubusercontent.com/8291514/213727234-cda046d6-28c6-491a-b284-b86c5cede25d.png#gh-light-mode-only">
  <img src="https://user-images.githubusercontent.com/8291514/213727225-56186826-bee8-43b5-9b15-86e839d89393.png#gh-dark-mode-only">
</p>
```
Screenshot 2024-02-29 at 10 47 14 AM
Screenshot 2024-02-29 at 10 47 22 AM

alex-ju:
> @alex-ju I did some quick poking around and Supabase actually does this in their readme 

the fragment method is [going to be deprecated](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#specifying-the-theme-an-image-is-shown-to), unfortunately, but I just tested the `prefers-color-scheme` solution and it works well

didoo:
Apparently is an official thing: https://github.blog/changelog/2021-11-24-specify-theme-context-for-images-in-markdown/

An alternative is to use an inline SVG and use `var(--color-fg-default)` for the `fill` attribute of the text part of the "logo" (but we have to test and see if it works)

alex-ju:
Thanks all for entertaining this idea! I will put this to rest for now and try to focus on more relevant matters.

PR #1980, opened 2024-03-07
Typescript - `Hds::DismissButton`

WenInCode:
### :pushpin: Summary

Kept this one extra small for the first TS change. Converts `Hds::DismissButton` to TypeScript.

- Converts `Hds::DismissButton` to TypeScript
- Updates template-registry for glint

#### Questions
- I ran a `git mv` to rename the component file, hoping to maintain git history. Is there a better way to be doing this?

### :link: External links

Jira ticket: [HDS-2721](https://hashicorp.atlassian.net/browse/HDS-2721)

***

### 👀 Component checklist

[HDS-2721]: https://hashicorp.atlassian.net/browse/HDS-2721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


natmegs (approved):

alex-ju (approved):
Looking good! I might have mentioned this before, but we'd need an additional alias in the registry (for all components), after that it's good to be merged.

> I ran a git mv to rename the component file, hoping to maintain git history. Is there a better way to be doing this?

it's a good idea, but unfortunately it wouldn't make any difference in the latest versions of git as the heuristics that decide whether it's the same file or a different one are not taking this into consideration (it's mainly about the similarity of the two – and in this case, because the initial code is just a few lines, the similarity is likely low).

PR #1906, opened 2024-01-24
Add image error handling to `Dropdown::Toggle::Icon`

cbfx:
### :pushpin: Summary

Add error handling to the `Dropdown::Toggle::Icon` component that handles setting `@imageSrc`.
If the image fails to load it falls back to showing the default icon.

### :camera_flash: Screenshots

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td>

Screenshot 2024-03-05 at 09 32 46

</td><td>

Screenshot 2024-03-05 at 09 31 52

</td></tr>
</table>

### :link: External links

Jira ticket: [HDS-3167](https://hashicorp.atlassian.net/browse/HDS-3167)

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#templates-for-npm-packages))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-3167]: https://hashicorp.atlassian.net/browse/HDS-3167?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1981, opened 2024-03-08
Skip flake in Dropdown toggle icon test

alex-ju:
### :pushpin: Summary

A test I introduced in #1906 is still flaky in CI despite being tested locally. I suggest we skip it until I find a more resilient way of testing the functionality.

***

### DIFF
--- a/showcase/tests/integration/components/hds/dropdown/toggle/icon-test.js
+++ b/showcase/tests/integration/components/hds/dropdown/toggle/icon-test.js
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import { module, test } from 'qunit';
+import { module, skip, test } from 'qunit';
 import { setupRenderingTest } from 'ember-qunit';
 import {
   render,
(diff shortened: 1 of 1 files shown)


KristinLBradley (approved):

meirish:
@alex-ju - @DingoEatingFuzz used base64-encoded data urls in the HCP test that does similar - it's nice because you get around needing a network request and you can test the success and error paths: https://github.com/hashicorp/cloud-ui/pull/8720/files#diff-cbe6e4d72dcbbd709afe2afe426f321cfd01bf4e4fa0c6ea2d2e66ca04a2ab66R6-R10

alex-ju:
thanks for the tip @meirish! I [tried the suggested approach](https://github.com/hashicorp/design-system/commit/1e97b8e768d15c3d0b75c7bc46688cb98e047d59) (and it's nice that we can skip the additional HTTP requests), but unfortunately it's still a flake for us. I will have to look a bit more in-depth to understand what causes the fails.

PR #1985, opened 2024-03-11
Add glint to showcase

natmegs:
### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist


alex-ju (commented):
I initially thought you could get away without having to do this work, but I agree with having a testing ground for the components work. We should not need to add ember types via DefinitelyTyped, as Alexey mentioned, but let me know if you run into any issues with using `ember-source/types` and I can look into it as well.

This branch can also be pointed to `main`, if you'd like.

showcase/app/templates/components/text.hbs
```diff
@@ -99,11 +107,12 @@
 
   <Shw::Text::H4 @tag="h3">Pre-defined colors</Shw::Text::H4>
   <Shw::Flex @direction="row" @wrap={{true}} {{style gap="2rem"}} as |SF|>
+    {{! @glint-expect-error }}
```
WenInCode, line 110: For all these, could you just convert the route to `ts`?  I guess that's not the goal but it seems like an easy one to convert.

WenInCode (approved):

showcase/app/templates/components/text.hbs
natmegs: We talked about this offline and have decided to move the changes in this file to a separate PR

showcase/package.json
```diff
@@ -43,14 +45,37 @@
...
+    "@types/ember__polyfills": "^4.0.6",
+    "@types/ember__routing": "^4.0.22",
+    "@types/ember__runloop": "^4.0.10",
+    "@types/ember__service": "^4.0.9",
+    "@types/ember__string": "3.0.15",
+    "@types/ember__template": "^4.0.7",
+    "@types/ember__test": "^4.0.6",
+    "@types/ember__utils": "^4.0.7",
+    "@types/qunit": "^2.19.10",
+    "@types/rsvp": "^4.0.9",
```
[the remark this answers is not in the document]
natmegs: Updated, thank you!

alex-ju (approved):
Thanks for addressing the feedback!

PR #1991, opened 2024-03-13
TypeScript: Convert badge to Typescript

chris-hut:
### :pushpin: Summary
We love TypeScript! This PR converts the famous 🦡 component over to Typescript!

With glint being [enabled across the design-system](https://github.com/hashicorp/design-system/pull/1976) - this PR converts a component over to Typescriptland!

### :link: External links
- Jira ticket: [HDS-2721](https://hashicorp.atlassian.net/browse/HDS-2721)
- 
***

### 👀 Component checklist

### References
image

### DIFF
--- a/.changeset/real-pets-work.md
+++ b/.changeset/real-pets-work.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Badge` - Converted component to typescript
--- a/packages/components/src/components/hds/badge/index.ts
+++ b/packages/components/src/components/hds/badge/index.ts
@@ -4,39 +4,44 @@
  */
 
 import Component from '@glimmer/component';
-import { assert } from '@ember/debug';
+import {assert} from '@ember/debug';
+import {type Color, ColorValues, type Size, SizeValues, type Type, TypeValues} from "./types.ts";
 
-export const DEFAULT_SIZE = 'medium';
-export const DEFAULT_TYPE = 'filled';
-export const DEFAULT_COLOR = 'neutral';
-export const SIZES = ['small', 'medium', 'large'];
-export const TYPES = ['filled', 'inverted', 'outlined'];
-export const COLORS = [
-  'neutral',
-  'neutral-dark-mode',
-  'highlight',
-  'success',
-  'warning',
-  'critical',
-];
+export interface HdsBadgeSignature {
+  Args: {
+    size: Size;
+    type: Type;
+    color: Color;
+    text: string;
+    icon: string | null;
+    isIconOnly: boolean | null;
+  },
+  Element: HTMLDivElement;
+}
+export const AVAILABLE_SIZES: string[] = Object.values(SizeValues);
+export const AVAILABLE_TYPES: string[] = Object.values(TypeValues);
+export const AVAILABLE_COLORS: string[] = Object.values(ColorValues);
+export const DEFAULT_SIZE = SizeValues.MEDIUM;
+export const DEFAULT_TYPE = TypeValues.FILLED;
+export const DEFAULT_COLOR = ColorValues.NEUTRAL;
 
-export default class HdsBadgeIndexComponent extends Component {
+export default class HdsBadgeIndexComponent extends Component<HdsBadgeSignature> {
   /**
    * Sets the size for the component
    * Accepted values: small, medium, large
    *
    * @param size
-   * @type {string}
+   * @type {Size}
    * @default 'medium'
    */
-  get size() {
+  get size(): Size {
     let { size = DEFAULT_SIZE } = this.args;
 
     assert(
-      `@size for "Hds::Badge" must be one of the following: ${SIZES.join(
+      `@size for "Hds::Badge" must be one of the following: ${AVAILABLE_SIZES.join(
         ', '
       )}; received: ${size}`,
-      SIZES.includes(size)
+        AVAILABLE_SIZES.includes(size)
     );
 
     return size;
@@ -47,17 +52,17 @@ export default class HdsBadgeIndexComponent extends Component {
    * Accepted values: filled, inverted, outlined
    *
    * @param type
-   * @type {string}
+   * @type {Type}
    * @default 'filled'
    */
-  get type() {
+  get type(): Type {
     let { type = DEFAULT_TYPE } = this.args;
 
     assert(
-      `@type for "Hds::Badge" must be one of the following: ${TYPES.join(
+      `@type for "Hds::Badge" must be one of the following: ${AVAILABLE_TYPES.join(
         ', '
       )}; received: ${type}`,
-      TYPES.includes(type)
+        AVAILABLE_TYPES.includes(type)
     );
 
     return type;
@@ -68,17 +73,17 @@ export default class HdsBadgeIndexComponent extends Component {
    * Accepted values: neutral, neutral-dark-mode, highlight, success, warning, critical
    *
    * @param color
-   * @type {string}
+   * @type {Color}
    * @default 'neutral'
    */
-  get color() {
+  get color(): Color {
     let { color = DEFAULT_COLOR } = this.args;
 
     assert(
-      `@color for "Hds::Badge" must be one of the following: ${COLORS.join(
+      `@color for "Hds::Badge" must be one of the following: ${AVAILABLE_COLORS.join(
         ', '
       )}; received: ${color}`,
-      COLORS.includes(color)
+        AVAILABLE_COLORS.includes(color)
     );
 
     return color;
@@ -89,7 +94,7 @@ export default class HdsBadgeIndexComponent extends Component {
    * @type {string}
    * @description The text of the badge. If `isIconOnly` is set to `true`, the text will be visually hidden but still available to assistive technology. If no text value is defined, an error will be thrown.
    */
-  get text() {
+  get text(): string {
     let { text } = this.args;
 
     assert(
@@ -107,7 +112,7 @@ export default class HdsBadgeIndexComponent extends Component {
    * @type {string|null}
    * @default null
    */
-  get icon() {
+  get icon(): string | null {
     return this.args.icon ?? null;
   }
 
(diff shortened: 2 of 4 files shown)


natmegs (commented):
Awesome 🙌 🙌 could we remove the `glint-nocheck` in the template too?

packages/components/src/template-registry.ts
```diff
@@ -29,4 +30,10 @@ export default interface HdsComponentsRegistry {
   // Helpers
   'hds-link-to-models': typeof HdsLinkToModelsHelper;
   'hds-link-to-query': typeof HdsLinkToQueryHelper;
+  HdsBadgeComponent: typeof HdsBadgeIndexComponent;
+
```
valeriia-ruban, line 34:
```suggestion
  // Badge
  'Hds::Badge': typeof HdsBadgeIndexComponent;
  'hds/badge': typeof HdsBadgeIndexComponent;
  'HdsBadge': typeof HdsBadgeIndexComponent;
  HdsBadgeComponent: typeof HdsBadgeIndexComponent;
```

valeriia-ruban (commented):
Looks great! 
I've just added very minor thing to take a look at.

packages/components/src/template-registry.ts
valeriia-ruban: just guessing, we might consider some order, like components, then helpers 🤔

didoo: If it can help, in terms of mental model and organization, we tend to follow the same order on the website as well as on the showcase: foundations → components → layouts → overrides → utilities (alphabetical within each group)

.changeset/real-pets-work.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Badge` - Converted component to typescript
```
didoo, line 5: uber-nitpick :)

```suggestion
`Badge` - Converted component to TypeScript
```

packages/components/src/components/hds/badge/index.ts
```diff
@@ -4,39 +4,44 @@
...
-  'neutral-dark-mode',
-  'highlight',
-  'success',
-  'warning',
-  'critical',
-];
-
-export default class HdsBadgeIndexComponent extends Component {
+import {assert} from '@ember/debug';
+import {type Color, ColorValues, type Size, SizeValues, type Type, TypeValues} from "./types.ts";
```
didoo, line 8: In other PRs (eg. [here](https://github.com/hashicorp/design-system/pull/1984/files#diff-b321dcb3b8e11ce30d48a349345e73fe4a3b35eef62e2ea1d20547cf3dc5ecec)) I see this formatted as below. 

Plus, in the text the types are prefixed with `Hds[ComponentName]` while here they're not.

@alex-ju do you think it would be possible to settle on one common formatting (including order) and naming convention?  /cc @hashicorp/hds-engineering

```suggestion
import {
  ColorValues,
  SizeValues,
  TypeValues,
  type Color,
  type Size,
  type Type,
} from "./types.ts";
```

packages/components/src/components/hds/badge/types.ts
```diff
@@ -0,0 +1,24 @@
+export type Size = `${SizeValues}`;
```
didoo, line 1: in other PRs (eg. [here](https://github.com/hashicorp/design-system/pull/1984/files#diff-6d495fe2de524a98c786418a10c8399941e00e4048187a6b595529d1fd1eb7e2)) the order of the declarations is first the enum and then the type. Plus the keys here all all UPPERCASE, in the other are PascalCase.

@alex-ju same as above, do you think it would be possible to have a standard "order" (and naming convention) to follow? /cc @hashicorp/hds-engineering

didoo (commented):
Left a few comments, mainly for internal discussion.

packages/components/src/components/hds/badge/index.ts
```diff
@@ -4,39 +4,44 @@
...
+
+export interface HdsBadgeSignature {
+  Args: {
+    size: Size;
+    type: Type;
+    color: Color;
+    text: string;
+    icon: string | null;
+    isIconOnly: boolean | null;
+  },
```
alex-ju, line 18: `@text` is the only required argument, the other ones are optional. 
`@isIconOnly` is false by default, so it won't be null at any point

```suggestion
  Args: {
    size?: Size;
    type?: Type;
    color?: Color;
    text: string;
    icon?: string | null;
    isIconOnly?: boolean;
  },
```

packages/components/src/components/hds/badge/index.ts
alex-ju, line 8: In the same spirit with Cristiano's comment above, we'd prefer these written as below (cascading changes needed) as [we try to align these across components](https://docs.google.com/document/d/17T9v7pYUNrm2TtcZE2le_fbX30Jop-iG3um7b0trRlg/edit#bookmark=id.kw9kdqlfg87r).
```suggestion
import {HdsBadgeColorValues, HdsBadgeSizeValues, HdsBadgeTypeValues} from "./types.ts";
import type { HdsBadgeColors, HdsBadgeSizes, HdsBadgeTypes } from "./types.ts";
```

packages/components/src/template-registry.ts
```diff
@@ -29,4 +30,10 @@ export default interface HdsComponentsRegistry {
...
+  // Badge
+  'Hds::Badge': typeof HdsBadgeIndexComponent;
+  'hds/badge': typeof HdsBadgeIndexComponent;
+  'HdsBadge': typeof HdsBadgeIndexComponent;
 }
```
alex-ju, line 39: +1 to the above comments, with the addition that the last entry should be `HdsBadge`. they should be placed alphabetically, but component and before the helpers, indeed

```suggestion
  // Badge
  'Hds::Badge': typeof HdsBadgeIndexComponent;
  'hds/badge': typeof HdsBadgeIndexComponent;
  HdsBadge: typeof HdsBadgeIndexComponent;

    // Helpers
  'hds-link-to-models': typeof HdsLinkToModelsHelper;
  'hds-link-to-query': typeof HdsLinkToQueryHelper;
}
```

WenInCode:
@alex-ju / @didoo -- Chris is out on medical leave for a few weeks so I updated this branch. Could one of you give this another look when you have some time? Appreciate it!

packages/components/src/components/hds/badge/index.ts
```diff
@@ -129,8 +133,8 @@ export default class HdsBadgeIndexComponent extends Component {
    * @method Badge#classNames
    * @return {string} The "class" attribute to apply to the component.
    */
-  get classNames() {
-    let classes = ['hds-badge'];
+  get classNames(): string {
```
didoo, line 136: In other components (eg. the `Text` component) this is not defined as `string`, which one of the two approaches is "correct"?

https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/text/index.ts#L128

packages/components/src/components/hds/badge/index.ts
```diff
@@ -144,3 +148,11 @@ export default class HdsBadgeIndexComponent extends Component {
     return classes.join(' ');
   }
 }
+
+declare module '@glint/environment-ember-loose/registry' {
```
didoo, line 152: In other components this is not added to the file, which one of the two approaches is "correct"?

https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/card/container.ts

packages/components/src/components/hds/badge/index.ts
```diff
@@ -107,7 +111,7 @@ export default class HdsBadgeIndexComponent extends Component {
    * @type {string|null}
    * @default null
    */
-  get icon() {
+  get icon(): string | null {
```
didoo, line 114: in other components these are declared inside the `types.ts` files

didoo (commented):
I have left a bunch of comments, but since I see differences between the conversion of this component in comparison with the previous ones, is better to wait for @alex-ju to come back from PTO (he has more context and knowledge)

didoo:
> I have left a bunch of comments, but since I see differences between the conversion of this component in comparison with the previous ones, is better to wait for @alex-ju to come back from PTO (he has more context and knowledge)

@WenInCode see my comment above

packages/components/src/components/hds/badge/index.ts
WenInCode: ah my apologies. I definitely missed this when going over the work that was already in here. You're correct that it can be removed.

packages/components/src/components/hds/badge/index.ts
WenInCode: Great point. I don't believe you _need_ to explicitly add the return type as it's inferred by typescript in this example.

Screenshot 2024-03-26 at 9 48 51 AM

packages/components/src/components/hds/badge/index.ts
WenInCode: Yea I mentally skipped over these. My bad. I've tested it out and typescript implicitly sets the return type. There is no harm in adding an explicit return type here in that it will only restrict accidentally changing the return type. 

Screenshot 2024-03-26 at 9 51 41 AM

We don't have a guideline around this right now, but I'm going to move it to the way it's been done in the other components so far for consistency.  Happy to return to it after a discussion with Alex once he's back from PTO.

packages/components/src/components/hds/badge/index.ts
WenInCode: Typescript seems to infer the return type really well when it's a basic type.  It's not as great when your return type is something like the enum values types that we are using. 

Screenshot 2024-03-26 at 10 00 14 AM

vs.

Screenshot 2024-03-26 at 10 00 00 AM

aklkv: my two cents here is to your point I think primitive time is easily inferred by TS on its own so it would only make sense to add return types in cases where we return complex types, otherwise it might add more noise then benefit

alex-ju: agree – if they are primitive types we shouldn't need to add them – for more elaborate types (such as `HdsBadgeColors`) there are benefits, and we usually place them in `type.js` (which is how they're setup for this component as well).

this being said, @WenInCode if you have time could you please remove the returned primitive types from the getter in this component (there a few of them)? if time is tight let me know and I can add a commit for this

alex-ju (commented):
thanks for getting this to the finish line @WenInCode! there are a few changes before we can get this merged, but let me know if you don't have the time to address them and I can take it over

alex-ju (approved):
Rebased and addressed the remaining feedback

PR #1999, opened 2024-03-15
Check Sass compiles to CSS before publish

alex-ju:
### :pushpin: Summary

In this setup, if `dist/styles/@hashicorp/design-system-components.css` doesn't exist, the script will log an error message and exit with a status code of 1, stopping the npm publish process. If the file does exist, the script will complete successfully and the npm publish process continues.

Open to suggestions about where to place the script in our current directory structure. Or if you'd prefer I can rewrite it in bash.

### :link: External links

Jira ticket: [HDS-3208](https://hashicorp.atlassian.net/browse/HDS-3208)

***

### 👀 Component checklist

[HDS-3208]: https://hashicorp.atlassian.net/browse/HDS-3208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


didoo:
@alex-ju just an idea: could this be a Rollup plugin instead (that exits the process if the file is missing?)

alex-ju:
> @alex-ju just an idea: could this be a Rollup plugin instead (that exits the process if the file is missing?)

I've not written one before, but yes, that could be an option too

alex-ju:
> @alex-ju just an idea: could this be a Rollup plugin instead (that exits the process if the file is missing?)

I gave the rollup addon path a try, but unfortunately all the evens we can hook into happen before write and we want to check for the file to exists after write.

packages/components/package.json
```diff
@@ -19,7 +19,7 @@
   "author": "HashiCorp Design Systems <design-systems@hashicorp.com>",
   "scripts": {
     "build": "rollup --config",
-    "prepublishOnly": "rollup --config",
+    "prepublishOnly": "rollup --config && node scripts/check-compiled-css.mjs",
```
didoo, line 22: could we use a command directly in the CLI? something like this?
```
"prepublishOnly": "rollup --config && test -f /path/to/file && [do something here, not sure what]
```

alex-ju: if a bash one-liner is preferred, we can go with that – not overly opinionated

packages/components/package.json
```diff
@@ -19,7 +19,7 @@
   "author": "HashiCorp Design Systems <design-systems@hashicorp.com>",
   "scripts": {
     "build": "rollup --config",
-    "prepublishOnly": "rollup --config",
+    "prepublishOnly": "rollup --config && test -f 'dist/styles/@hashicorp/design-system-components.css' || echo 'The pre-compiled CSS file was not found' && exit 1",
```
didoo, line 22: 🙌

didoo (approved):
OK for me (but I haven't tested it personally, not even sure how we could tbh).

MelSumner (approved):

PR #1993, opened 2024-03-13
New addon blueprint

aklkv:
### :pushpin: Summary

This mainly exists for demonstration

### :link: External links

Jira ticket: [HDS-XXX](https://hashicorp.atlassian.net/browse/HDS-XXX)
Figma file: [if it applies]

***

### 👀 Component checklist

- [ ] Percy was checked for any visual regression
- [ ] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#templates-for-npm-packages))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #2004, opened 2024-03-15
Publish types from `packages/components`

natmegs:
### :pushpin: Summary

The main issue this PR attempts to fix is the lack of type declarations being produced for the components package.

The impact of no type declarations is consumers get no type information, and do not get type checking when using components from this package.

Two steps to resolve this:
* Align our configs with the latest ember blueprint where possible
* Adjust configs until types and usage outside of package are happy

In `packages/components`:
* `yarn`
* `yarn build`
Build should succeed and you should see a declarations folder (adjacent to dist) with type declarations
Screenshot 2024-03-19 at 3 16 33 PM

In `showcase` (important: after doing the above):
* `yarn`
* Restart the glint server in your code editor
* Go to types/global.d.ts
* You should get types for the HdsComponentRegistry 
Screenshot 2024-03-19 at 3 17 31 PM
Screenshot 2024-03-19 at 3 17 48 PM

### 👀 Component checklist


packages/components/src/components/hds/button/index.ts
```diff
@@ -5,7 +5,7 @@
 
 import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
-import { type HdsInteractiveSignature } from '../interactive';
+import { type HdsInteractiveSignature } from '../interactive/index.ts';
```
aklkv, line 8:
```suggestion
import type { HdsInteractiveSignature } from '../interactive/index.ts';
```

packages/components/tsconfig.json
```diff
@@ -13,16 +13,11 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "noPropertyAccessFromIndexSignature": false,
+    "declarationDir": "declarations",
```
alex-ju, line 16: I believe these would need to be placed inside `dist` for an app to be able to access them
```suggestion
    "declarationDir": "./dist/declarations",
```

alex-ju: nevermind, `.npmignore` seems not to care about it, so it should end up being published

packages/components/package.json
```diff
@@ -14,7 +14,6 @@
     "url": "https://github.com/hashicorp/design-system.git",
     "directory": "packages/components"
   },
-  "types": "dist/index.d.ts",
   "license": "MPL-2.0",
   "author": "HashiCorp Design Systems <design-systems@hashicorp.com>",
   "scripts": {
```
aklkv, line 19: need to update the script to match the blueprint in order to build types and probably remove npm-run-all

aklkv:
also blueprint is missing this dep which I had to install manually https://github.com/hashicorp/design-system/pull/1993/commits/681ac018d0e9d00d15ca69de217fe301d97d163e

packages/components/package.json
natmegs: The build scripts in this PR do match the blueprints afaict? We could change the lint script to use concurrently instead of npm-run-all in a separate PR if needed as it's a bit separate from the intent of this PR

natmegs:
> blueprint is missing this dep

What is that used for/how do we know we need it? Just curious as checks seem to be ✅ without

aklkv:
> > blueprint is missing this dep
> 
> What is that used for/how do we know we need it? Just curious as checks seem to be ✅ without

https://github.com/hashicorp/design-system/pull/2004/files#diff-949397ffa49ee9acf2f8cf58a09becfe5c27e2bc3673eaf251bdbbf4a07d1538R23

natmegs:
@aklkv thanks for the example usage, I'm still not certain on how to know that we need that? If it's necessary, it makes me wonder if we're missing something in our checks that would signal that to us

PR #2011, opened 2024-03-18
bugfix(components): fix flaky tests by restoring waiters usage from @ember/test-waiters

fivetanley:
### :pushpin: Summary

Makes `@ember/test-waiters` run again in the test environment.

Since @hashicorp/design-system-components became a V2 Ember Addon, it no longer gets some of the default babel plugins when compiling its output to the `dist` folder that gets included in the npm package. One of those packages is [babel-plugin-debug-macros][1], which compiles the following code in development/testing environments from:

```javascript
import { DEBUG } from '@glimmer/env';

if (DEBUG) {
  doThing();
}
```

to the following compiled version:

```javascript
if (true) {
  doThing();
}
```

in production environments it gets compiled to:

```javascript
if (false) {
  doThing();
}
```

which then gets stripped out.

Since this babel plugin is no longer included and V2 addons don't seem to use the same build pipeline as V1 addons, what happens today is that the waiters we added for Flyout and Modal never get set up since `DEBUG` from `@glimmer/env` is always false in the runtime version.

It turns out that [@ember/test-waiters][2] does not have a huge performance impact in production code, since the version of the library that gets included in production builds is essentially a collection of no-ops. From the `@ember/test-waiters` README:

> When in production mode, we make this instance inert and essentially no cost to
> invoke. Since test waiters are intended to be called from application or addon
> code, but are only required to be active when in tests, this process of making
> the instance inert is important. Even though code is still invoked, this has a
> negligible impact on performance.

We can remove our usage of `DEBUG` since it is un-necessary and doesn't save that much in terms of code size or runtime performance.

[1]: https://github.com/ember-cli/babel-plugin-debug-macros
[2]: https://www.npmjs.com/package/@ember/test-waiters

### :link: External links

- [JIRA](https://hashicorp.atlassian.net/browse/TF-14041)
- [Related Slack Convo](https://hashicorp.slack.com/archives/C7KTUHNUS/p1705001922424289)

***

### 👀 Component checklist

### DIFF
--- a/.changeset/ten-falcons-run.md
+++ b/.changeset/ten-falcons-run.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-Fix flaky tests by making sure @ember/test-waiters run in all environments
+`Modal`, `Flyout` - Fixed flaky tests by running `@ember/test-waiters` in all environments
--- a/.changeset/ten-falcons-run.md
+++ b/.changeset/ten-falcons-run.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Fix flaky tests by making sure @ember/test-waiters run in all environments
--- a/packages/components/src/components/hds/flyout/index.js
+++ b/packages/components/src/components/hds/flyout/index.js
@@ -9,14 +9,8 @@ import { action } from '@ember/object';
 import { assert } from '@ember/debug';
 import { getElementId } from '../../../utils/hds-get-element-id';
 import { buildWaiter } from '@ember/test-waiters';
-import { DEBUG } from '@glimmer/env';
 
-let waiter;
-
-// Notice: this code will get stripped out in production builds (DEBUG evaluates to `true` in dev/test builds, but `false` in prod builds)
-if (DEBUG) {
-  waiter = buildWaiter('@hashicorp/design-system-components:flyout');
-}
+let waiter = buildWaiter('@hashicorp/design-system-components:flyout');
 
 export const DEFAULT_SIZE = 'medium';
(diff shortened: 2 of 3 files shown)


jarrodn-au:
Thanks for the good analysis and write-up Stanley, I really appreciate your annotated spelunking through Ember internals like this.

.changeset/ten-falcons-run.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Fix flaky tests by making sure @ember/test-waiters run in all environments
```
alex-ju, line 5: `Modal`, `Flyout` - Fixed flaky tests by running `@ember/test-waiters` in all environments

alex-ju (approved):
Thanks for taking the time to investigate this @fivetanley! While the explanation and the change make sense to me, I'd like @meirish (who I believe introduced us to using `@ember/test-waiters` in the first place) to have a look as well before moving on with this change.

fivetanley:
@alex-ju it was I who introduced test waiters but happy for @meirish to weigh in as well

alex-ju:
> @alex-ju it was I who introduced test waiters but happy for @meirish to weigh in as well

ha! sorry about that – I mixed up contributions here 😄 
nevermind then – happy for us to get it merged. if Matthew has thoughts on this we can address them the release

meirish (approved):
Yep, everything that @fivetanley said ! 🎉 

The only thing I'll add (that I don't think is necessary to add here) is that embroider now has its own macros package for doing things like what the DEBUG babel transform used to do. Of particular interest are the `isTesting` and `isDevelopingApp` macros: https://github.com/embroider-build/embroider/tree/main/packages/macros#istesting-isdevelopingapp

PR #2015, opened 2024-03-20
Bug Hunting: A11y Audits on tab content

MelSumner:
Issue: https://hashicorp.atlassian.net/browse/HDS-3176 

The accessibility checker is not catching some valid issues in tab content. This branch is an attempt to troubleshoot the issue.

I purposefully added an element that should cause a failure:
CleanShot 2024-03-20 at 09 58 21

Then I added a few extra tests to the accordion acceptance test for the website repo.

Tests pass locally: 
CleanShot 2024-03-20 at 10 04 17
 
 
 I've also tried introducing the failure on a page without tabbed content: 
 
CleanShot 2024-03-20 at 10 44 32

Tests still seem to pass:
CleanShot 2024-03-20 at 10 45 31

Need dig more into this so I can determine why the failures are not triggering.


MelSumner:
Update: I am able to get a non-color-contrast related failure to be caught by the test suite. So narrowing things down.

MelSumner:
It seems like the accessibility tests for the website were not configured in the same way as in the showcase, which is why it hasn't been catching failures. This is going to take some refactoring work and some remediation to fix things that didn't get caught properly. I'm going to update the JIRA ticket with some related issues that outline the work that needs to be done.

PR #2020, opened 2024-03-25
`RichTooltip/Popover` component - `hds-anchored-position` modifier [01]

didoo:
### :pushpin: Summary

This PR is part of a larger set of PRs: see https://github.com/hashicorp/design-system/pull/2016

It's mainly a porting (with cleanup/refactoring) of #1947 

This PR introduces a modifier built around the [Floating UI](https://floating-ui.com) library, to provide:
- **anchoring** (where an element position is “anchored” to a target element) 
- **collision detection** (when the position of the element in relation to the anchor is changed dynamically to avoid collision with the browser window/viewport).

For details about why we opted for [Floating UI](https://floating-ui.com) instead of [Popper.js](https://popper.js.org/) refer to this RFC document: https://docs.google.com/document/d/1ey8LVQn0sA0bPbKoaz93ZoYVO1cLcbXUeg1o-tf6LIY/edit

In this PR I have:
- introduced a custom `hds-anchored-position` modifier whose responsibility is to:
  - identify the different DOM elements (anchor element, floating element, arrow element)
  - process the `options` provided to the modifier and create a list of Floating UI middleware functions to invoke
  - define an async function that invokes the Floating UI `computePosition` main function
  - process the returned values/state of Floating UI and apply the computed coordinates to the floating and arrow elements
  - invoke the Floating UI `autoUpdate` function
  - assign the "cleanup" function to the modifier's teardown
- added a unit and integration tests for the modifier
    - the intent is to avoid testing the Floating UI functionality, but to make sure the `options` are correctly read and processed by the underlying code logic

_Notice: while initially the modifier was called `hds-float-popover` and contained references to the "toggle" and "popover" elements, on a second thought we decided to rename it `hds-anchored-position` (inspired by [this](https://github.com/primer/behaviors/blob/775efef80607e44eb40b7287e4715b09708e6e60/src/anchored-position.ts)) and generalize code to be more agnostic and remove all the references to "trigger" and “popover” (the modifier itself doesn't know if these concepts, which are related to the "popover" entity)_

### :link: External links

***

### 👀 Component checklist


packages/components/src/modifiers/hds-float-popover.js
```diff
@@ -0,0 +1,155 @@
...
+  } = popoverOptions;
+
+  // we build dynamically the list of middleware functions to invoke, depending on the options provided
+
+  const popoverMiddleware = [];
+
+  // https://floating-ui.com/docs/offset
+  popoverMiddleware.push(offset(popoverOffsetOptions));
+
+  // TODO? split the option in two to support collision detection only in one direction like for the power-select?
```
KristinLBradley, line 42: For the PowerSelect, by default, the dropdown list matches the "trigger" width. In that case, the ShiftOptions probably don't make sense to use.

However, if matchTriggerWidth is set to "false" (meaning that the dropdown list could be wider than the trigger), then it could make sense to shift the position of the dropdown to keep it in view. I only found one example of this option in use in HashiCorp projects when I searched previously, however.

So it could make sense to split these options into two...

showcase/tests/integration/modifiers/hds-float-popover-test.js
```diff
@@ -0,0 +1,235 @@
...
+module('Integration | Modifier | hds-float-popover', function (hooks) {
+  setupRenderingTest(hooks);
+
+  hooks.beforeEach(function () {
+    this.styleElement = document.createElement('style');
+
+    // see: https://codepen.io/didoo/pen/VwNpOJb
+    this.styleElement.textContent = `
+      body {
+        padding: 0px;
```
KristinLBradley, line 127: Just to mention. Browsers add a margin to the HTML body element by default. They don't add padding.

KristinLBradley (commented):
Added a few comments/questions

packages/components/src/modifiers/hds-float-popover.js
```diff
@@ -0,0 +1,155 @@
...
+    );
+
+    let { x, y, placement, strategy, middlewareData } = state;
+
+    Object.assign(popoverElement.style, {
+      position: strategy,
+      top: `${y}px`,
+      left: `${x}px`,
+      // TODO? commenting this for now, will need to make this conditional to some argument (and understand how this relates to the `@height` argument)
+      // maxHeight: `${middlewareData.size.availableHeight - 10}px`,
```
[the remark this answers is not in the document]
didoo: yes

packages/components/src/modifiers/hds-float-popover.js
didoo: The `enableCollisionDetection` now supports a boolean (`true`) as well as a string (`flip` or `shift`) to enable distinct behaviours.

showcase/tests/integration/modifiers/hds-float-popover-test.js
didoo: You're right 🤦‍♂️

alex-ju (approved):
Looking great!! Can't come up with any suggestions for improvement

packages/components/src/modifiers/hds-anchored-position.js
```diff
@@ -0,0 +1,223 @@
...
+];
+
+// we use this function to process all the options provided to the modifier in a single place,
+// in relation to the Floating UI APIs, and keep the modifier code more clean/simple
+export const getFloatingUIOptions = (options) => {
+  let {
+    placement = DEFAULT_PLACEMENT,
+    strategy = 'absolute', // we don't need to use `fixed` if we use the Popover API for the "floating" element (it puts the element in the `top-layer`)
+    offsetOptions,
+    flipOptions = { padding: 8 },
```
KristinLBradley, line 50: Would it make sense to create a variable for the padding values?

packages/components/src/modifiers/hds-anchored-position.js
```diff
@@ -0,0 +1,223 @@
...
+  // https://floating-ui.com/docs/shift
+  // https://floating-ui.com/docs/autoPlacement
+  if (
+    enableCollisionDetection === true ||
+    enableCollisionDetection === 'flip'
+  ) {
+    middleware.push(flip(flipOptions));
+  }
+  if (
+    (enableCollisionDetection === true) |
```
KristinLBradley, line 75: This shouldn't be `||` vs. `|` here as in the above if statement?

didoo: I am not sure how I missed this! Great catch!

packages/components/src/modifiers/hds-anchored-position.js
didoo: Yeah, makes sense. Done.

didoo:
@alex-ju just a heads up that in 12a758310d6945f57fa91124c5d82f1962ad97b9 I've slightly changed the API of the modifier, flattening the `named` arguments (for simplicity, and because Ember complained in some cases when I was overwriting some properties)

KristinLBradley (approved):

alex-ju (approved):
Reviewed the code and it looks/works well!
It makes sense to have the arrow-related settings grouped, but flattening it makes the in-line modifier call much simpler as it does require an extra hash within the call (e.g. `{{hds-anchored-position '#anchor' arrowSelector='#arrow' }}` instead of `{{hds-anchored-position '#anchor' arrowOptions=(hash selector='#arrow') }}`)

didoo:
Decided to merge this directly to `main` to speed up the adoption of the modifier in other components, if needed.

PR #2023, opened 2024-03-26
HDS-2717: `Hds::Toast` - Converts to Typescript

WenInCode:
### :pushpin: Summary

Converts `Hds::Toast` to Typescript

- Converts `Hds::Toast` to Typescript
  - I initially tried the recommended approach of using `templateOnlyComponent` to type the `Toast` component as it had no backing class. However, I was getting Typescript errors when trying to pass it the type signature stating that `templateOnlyComponent` does not take type arguments. It seems like that only became available in later version of Ember. I thought it would be fine given this is an embroider addon but idk, typescript was not happy! Willing to revisit this if anyone has any ideas.
- Updates the `Hds::Alert` `onDismiss` argument type to take the `MouseEvent` and then any number of args with `any` value. I originally had this set to `() => void`, which works great if you never want to handle arguments on the action handler, but it falls on it's face if you do. I thought it would be better to switch to always handle an event, and then any number of arguments of any value, as you can pass in whatever you would like with a `(fn)` helper. I used `any` here because you can't type the handler functions additional args if you set it to `unknown`

Screenshot 2024-03-26 at 2 59 49 AM

Screenshot 2024-03-26 at 3 05 11 AM
Screenshot 2024-03-26 at 3 05 19 AM

### :link: External links

Jira ticket: [HDS-2717](https://hashicorp.atlassian.net/browse/HDS-2717)

***

### 👀 Component checklist

[HDS-2717]: https://hashicorp.atlassian.net/browse/HDS-2717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/toast/index.ts
+++ b/packages/components/src/components/hds/toast/index.ts
@@ -1,0 +1,4 @@
+import Component from '@glimmer/component';
+import type { HdsToastSignature } from './types.ts';
+
+export default class HdsToastComponent extends Component<HdsToastSignature> {}
--- a/packages/components/src/components/hds/toast/index.ts
+++ b/packages/components/src/components/hds/toast/index.ts
@@ -1,4 +1,6 @@
-import Component from '@glimmer/component';
+import TemplateOnlyComponent from '@ember/component/template-only';
 import type { HdsToastSignature } from './types.ts';
 
-export default class HdsToastComponent extends Component<HdsToastSignature> {}
+const HdsToastComponent = TemplateOnlyComponent<HdsToastSignature>();
+
+export default HdsToastComponent;
--- a/packages/components/src/components/hds/toast/types.ts
+++ b/packages/components/src/components/hds/toast/types.ts
@@ -1,0 +1,5 @@
+import type { HdsAlertSignature } from '../alert/types.ts';
+
+export interface HdsToastSignature extends Omit<HdsAlertSignature, 'Args'> {
+  Args: Omit<HdsAlertSignature['Args'], 'type'>;
+}
--- a/.changeset/soft-buttons-burn.md
+++ b/.changeset/soft-buttons-burn.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Toast` - Converted component to TypeScript
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -256,6 +256,7 @@
       "./components/hds/text/index.js": "./dist/_app_/components/hds/text/index.js",
       "./components/hds/text/types.js": "./dist/_app_/components/hds/text/types.js",
       "./components/hds/toast/index.js": "./dist/_app_/components/hds/toast/index.js",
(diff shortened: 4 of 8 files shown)


Dhaulagiri:
> It seems like that only became available in later version of Ember

Like if the components package was on a later version of Ember you would not have this problem? Or something else?

aklkv:
@WenInCode I am curious ti dig into this issue, did [this](https://typed-ember.gitbook.io/glint/environments/ember/template-only-components) fail?

aklkv:
Also side note, I found [this](https://github.com/ijlee2/ember-codemod-args-to-signature) codemod very useful in adding signatures it takes care of these conversions. you can supply the exact component path to be converted in order to avoid converting other components

packages/components/src/components/hds/toast/index.ts
```diff
@@ -0,0 +1,4 @@
+import Component from '@glimmer/component';
+import type { HdsToastSignature } from './types.ts';
+
+export default class HdsToastComponent extends Component<HdsToastSignature> {}
```
alex-ju, line 4: Having a backing class is a good fallback, but considering we expect to have many other components as templates only, I'd try to get it right now so we can replicate this in the future.

You probably tried something like this and it complained with something along the lines of 'Expected 0 type arguments, but got 1.'. This, as you mentioned in the PR description, means it picks up `@types/ember__component` with a 3.x version (which seems to be required by a 3rd degree dependency in `website`, so unrelated to this package).

```suggestion
import templateOnlyComponent from '@ember/component/template-only';
import type { HdsToastSignature } from './types.ts';

const HdsToastComponent = templateOnlyComponent<HdsToastSignature>();

export default HdsToastComponent;
```

Instead of `@types/ember__component/template-only.d.ts`, it should be using `ember-source/types/stable/@ember/component/template-only.d.ts` (available in our ember version), however that's not happening and it's puzzling 🤔

WenInCode: Ah that's exactly what it was. Appreciate the commit. I tried it out and it LGTM

packages/components/src/components/hds/toast/types.ts
```diff
@@ -0,0 +1,5 @@
+import type { HdsAlertSignature } from '../alert/types.ts';
+
+export interface HdsToastSignature extends Omit<HdsAlertSignature, 'Args'> {
+  Args: Omit<HdsAlertSignature['Args'], 'type'>;
+}
```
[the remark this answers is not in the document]
WenInCode: Yes, I initially had it implementing the Alert interface and realized afterwards that `inline` is hardcoded in `Toast`. 👍

WenInCode:
@alex-ju thanks for helping with the resolution issue. The changes LGTM I think this is ready for another set of 👀.

alex-ju (approved):

PR #2028, opened 2024-03-27
Relocate types

WenInCode:
### :pushpin: Summary

Relocates the types to a folder outside the components/ folder. 

It seems like rollup adds all the ts/js files within the components/ folder into the ember-addon/app-js bit in the package.json and expects there to a `default` export. I think this is because it expects all those files to be components as is the ember convention? [This results in noisey errors for our consumers.](https://hashicorp.slack.com/archives/C06NMSYDZPA/p1711509633996269) 

I think there are 2 options for now. We can move our types to a `types/` folder outside the `components/` folder or we can put out types alongside their components within the files that define the components. 

Aside: would it be beneficial to export all the types from a single types index file to make it easier for consumers to import? Not required for this.

Before, the build has files trying to export `default` from the types which don't necessarily have `default` exports.
Screenshot 2024-03-27 at 9 32 36 AM

After, no file for the type here.
Screenshot 2024-03-27 at 9 59 55 AM

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist


WenInCode:
Closing this in favour of #2030

PR #2030, opened 2024-03-27
Stop reexporting type files

WenInCode:
### :pushpin: Summary

Updates the rollup glob pattern to exlude `types.js` files within the `components` folder.

It seems like rollup adds all the ts/js files within the components/ folder into the ember-addon/app-js bit in the package.json and expects there to a `default` export. I think this is because it expects all those files to be components as is the ember convention? [This results in noisey errors for our consumers.](https://hashicorp.slack.com/archives/C06NMSYDZPA/p1711509633996269) 

I thought we could just add the types pattern under the `{ exclude: string[] }` options in the `appReexports` but it did not work as I would have expected, or perhaps my glob was not right. So  I ended up updating the `patterns` globs to exclude types. 

**Before:**
Screenshot 2024-03-27 at 9 32 36 AM

**After:**
Screenshot 2024-03-27 at 9 59 55 AM

### :link: External links

[appReexports](https://github.com/embroider-build/embroider/blob/11f2c708829cebd1d1770a942fe18b67e1a901aa/packages/addon-dev/src/rollup.ts#L21-L23)

***

### 👀 Component checklist


packages/components/rollup.config.mjs
```diff
@@ -27,7 +27,7 @@ const plugins = [
   // "app" tree. Things in here should also be in publicEntrypoints above, but
   // not everything in publicEntrypoints necessarily needs to go here.
   addon.appReexports([
-    'components/**/*.js',
+    'components/**/!(*types).js',
```
WenInCode, line 30: if we are following this pattern for type location going forward, should I also update the helper and modifiers pattern below?

didoo: Currently the hds modifiers and helpers live under the same folder(s). which folder structure are you envisioning?

Types next to the files:
```
packages/components/src/helpers/hds-link-to-models.ts
packages/components/src/helpers/hds-link-to-models.types.ts
packages/components/src/modifiers/hds-clipboard.ts
packages/components/src/modifiers/hds-clipboard.types.ts
```

Creation of parent folders (is it possible? does it work in Ember?):
```
packages/components/src/helpers/hds-link-to-models/index.ts
packages/components/src/helpers/hds-link-to-models/types.ts
packages/components/src/modifiers/hds-clipboard/index.ts
packages/components/src/modifiers/hds-clipboard/types.ts
```

WenInCode: That is a very good question. I'm not sure what the answer is atm, and so maybe it's best not to address this now.

I'll try out the above options just to see what it looks like/what works. It would be _nice_ if it could be consistent across all components/modifiers/helpers/etc.

natmegs: I'm not super familiar with app reexports - if the types are removed does this mean that a consumer in a v1 addon can't import things from type files (like enums etc), or is that still possible since they are included in the public entrypoints above?

WenInCode: Ya I believe that is handled by the `publicEntrypoints`. CUT currently [re-exports the types ](https://github.com/hashicorp/consul-ui-toolkit/blob/main/toolkit/package.json#L111)(whoops!) but we also have an [index file that imports and exports all the types](https://github.com/hashicorp/consul-ui-toolkit/blob/main/toolkit/src/types/index.ts). That file is not part of the `appReexports` but it is [accessible from cloud-ui.](https://github.com/hashicorp/cloud-ui/blob/9285425b1ade4c27ab1a20fa5ee4e33addddf160/engines/monitoring/addon/components/page/platform-logs/resource/index.ts#L19)

aklkv: `appReexports` is a compatability layer tat only needed for v1 apps with hbs/js, with current setup it would still export types

.changeset/slow-scissors-flow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Exclude `types.js` files from appReexports
```
alex-ju, line 5:
```suggestion
Fixed default export warnings by preventing `types.js` files from being reexported
```

packages/components/rollup.config.mjs
alex-ju: We currently don't have type files for helpers and modifiers, so I wouldn't worry about it just now.

alex-ju (approved):
Left a suggestion for the changelog entry, otherwise it's good to get in

alex-ju:
> @alex-ju thanks for the suggestion and review

Feel free to merge whenever time allows

PR #2033, opened 2024-03-29
`IconTile` - Convert to TypeScript

Dhaulagiri:
### :pushpin: Summary

Concerts the `IconTile` component to TS

### 👀 Component checklist

- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets)

### DIFF
--- a/packages/components/src/components/hds/icon-tile/index.ts
+++ b/packages/components/src/components/hds/icon-tile/index.ts
@@ -6,16 +6,19 @@
 import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
 import {
-  HdsIconTileSizeValues,
-  HdsIconTileColorValues,
+  HdsIconTileColorNeutral,
   HdsIconTileProductValues,
+  HdsIconTileSizeValues,
 } from './types.ts';
 import type { HdsIconTileSignature } from './types.ts';
 
 export const DEFAULT_SIZE = 'medium';
 export const DEFAULT_COLOR = 'neutral';
 export const SIZES: string[] = Object.values(HdsIconTileSizeValues);
-export const COLORS: string[] = Object.values(HdsIconTileColorValues);
+export const COLORS: string[] = Object.values({
+  ...HdsIconTileColorNeutral,
+  ...HdsIconTileProductValues,
+});
 export const PRODUCTS: string[] = Object.values(HdsIconTileProductValues);
 
 export default class HdsIconTileIndexComponent extends Component<HdsIconTileSignature> {
--- a/packages/components/src/components/hds/icon-tile/index.ts
+++ b/packages/components/src/components/hds/icon-tile/index.ts
@@ -5,26 +5,20 @@
 
 import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
+import {
+  HdsIconTileSizeValues,
+  HdsIconTileColorValues,
+  HdsIconTileProductValues,
+} from './types.ts';
+import type { HdsIconTileSignature } from './types.ts';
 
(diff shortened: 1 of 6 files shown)


Dhaulagiri:
I saw this linting failure locally but couldn't quite sort out why it was happening 🤔 

Screenshot 2024-03-29 at 15 51 23

packages/components/src/components/hds/icon-tile/index.ts
```diff
@@ -87,7 +81,7 @@ export default class HdsIconTileIndexComponent extends Component {
     } else if (this.args.icon) {
       return this.args.icon;
     } else {
-      return null;
+      return undefined;
```
Dhaulagiri, line 84: this fixed [this](https://github.com/hashicorp/design-system/pull/2033#issuecomment-2027725005) linting error

alex-ju: `undefined` seems to be the correct value here. in the [FlightIcon signature](https://github.com/hashicorp/design-system/blob/main/packages/components/unpublished-development-types/%40hashicorp/ember-flight-icons/flight-icon.d.ts#L8) we define the name argument as `string`, if it exists (otherwise it's `undefined`). also in the component API documentation for the Icon Tile component we state that it must be a string (no mention of allowing null).

packages/components/src/components/hds/icon-tile/types.ts
```diff
@@ -0,0 +1,52 @@
...
+  HCP = 'hcp',
+  Nomad = 'nomad',
+  Packer = 'packer',
+  Terraform = 'terraform',
+  Vagrant = 'vagrant',
+  Vault = 'vault',
+  'Vault Secrets' = 'vault-secrets',
+  'Vault Radar' = 'vault-radar',
+  Waypoint = 'waypoint',
+}
```
alex-ju, line 39: We could use a union type to avoid duplication here
```
enum HdsIconTileProductValues {
  ...
}

enum HdsIconTileNeutral {
  Neutral = 'neutral',
}

type HdsIconTileColorValues = HdsIconTileProductValues | HdsIconTileNeutral; 
```

alex-ju (approved):
Left a suggestion otherwise is good to get merged

packages/components/src/components/hds/icon-tile/types.ts
Dhaulagiri: Nice! That did the trick.

packages/components/src/components/hds/icon-tile/types.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export type HdsIconTileProducts = `${HdsIconTileProductValues}`;
+
+export enum HdsIconTileColorNeutral {
+  Neutral = 'neutral',
+}
+
+type HdsIconTileColorValues =
```
alex-ju, line 29:
```suggestion
export type HdsIconTileColorValues =
```

Dhaulagiri: actually, that isn't quite it...

alex-ju: indeed, the [initial suggestion](https://github.com/hashicorp/design-system/pull/2033#discussion_r1551414299) addressed the type, but not the array of colours we use for validation and in our showcase app. it's rather tricky to properly extend enums, so I'd use the same approach we used for types and amend 'neutral' to product values. pushed a commit for illustration

Dhaulagiri: Thanks! TIL!

PR #2041, opened 2024-04-08
HDS-3124 Select doc update

majedelass:
### :pushpin: Summary

Update the text in the overview section to explain that the select component is a native HTML element.

Following consumer requests to update the list within the select component, we needed to explain that any non-native HTML implementations are not possible within this component.

### :link: External links

Jira ticket: [HDS-3124](https://hashicorp.atlassian.net/browse/HDS-3124)

***

### 👀 Component checklist

[HDS-3124]: https://hashicorp.atlassian.net/browse/HDS-3124?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/docs/components/form/select/partials/guidelines/guidelines.md
+++ b/website/docs/components/form/select/partials/guidelines/guidelines.md
@@ -2,7 +2,7 @@
 
 **This component uses a native HTML element**
 
-It's not possible to create custom variations of this component as it uses a native HTML element. Find more information [on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select).
+It's not possible to create technically custom variations of this component as it uses a native HTML element. Find more information [on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select).
 
 !!!
 
--- a/website/docs/components/form/select/partials/guidelines/guidelines.md
+++ b/website/docs/components/form/select/partials/guidelines/guidelines.md
@@ -2,7 +2,7 @@
 
 **This component uses a native HTML element**
 
-It's not possible to customize options within this component as it uses a native HTML element, and uses browser defaults to render it. Find more information [on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select).
+It's not possible to customize options within this component as it is a native HTML element, and uses browser defaults to render it. Find more information [on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select).
 
 !!!
 
--- a/website/docs/components/form/select/partials/guidelines/guidelines.md
+++ b/website/docs/components/form/select/partials/guidelines/guidelines.md
@@ -2,7 +2,7 @@
 
 **This component uses a native HTML element**
 
-It's not possible to create technically custom variations of this component as it uses a native HTML element. Find more information [on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select).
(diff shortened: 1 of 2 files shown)


website/docs/components/form/select/partials/guidelines/guidelines.md
```diff
@@ -1,3 +1,11 @@
+!!! Info
+
+**This component is a native HTML element**
+
+Creating custom non-native variations of this component is not possible as we rely on the brower to render it as-is. More information can be found [on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select).
```
heatherlarsen, line 5: Great addition! I found the original content to be easier to parse and more accurate (since the visual stylings aren't rendered "as-is"). Also updated the second sentence to use active voice instead of passive voice. 

```suggestion
**This component uses a native HTML element**

It's not possible to create custom variations of this component as it uses a native HTML element. Find more information [on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select).
```

website/docs/components/form/select/partials/guidelines/guidelines.md
```diff
@@ -1,3 +1,11 @@
+!!! Info
+
+**This component uses a native HTML element**
+
+It's not possible to create custom variations of this component as it uses a native HTML element. Find more information [on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select).
```
didoo, line 5: Should we be more specific, and explain that it's not possible to create custom variations of the `<option>` elements?

majedelass: Isn't this also applicable to the `<select>` element too, where we can't apply non-native HTML customizations to it?

majedelass: Initially when updating this my intent was to make it seem that any "custom variation" (which to your point Jory may be not accurate enough) within the select component (meant to be technical variation) was not possible and intentionally kept it vague so that it encompasses more than just the `<option>`.

But to your point, maybe we need to say "any technical custom variation" isn't allowed? I Want to include more than the option list so that this is "future" proof and that no one comes and says we didn't explicitly say the `<select>` element wasn't mentioned.

majedelass: Ok wording was updated again.

didoo (approved):
👍

website/docs/components/form/select/partials/guidelines/guidelines.md
```diff
@@ -1,3 +1,11 @@
+!!! Info
+
+**This component uses a native HTML element**
+
+It's not possible to customize options within this component as it is a native HTML element, and uses browser defaults to render it. Find more information [on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select).
```
jorytindall, line 5: I'm not sure if the comma is needed here

```suggestion
It's not possible to customize options within this component as it is a native HTML element and uses browser defaults to render it. Find more information [on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select).
```

jorytindall (approved):
One small note, I'm not sure if I'm correct but it stood out to me. :shipit:

PR #2017, opened 2024-03-21
`RichTooltip/Popover` component - `hds-register-event` modifier [02]

didoo:
### :pushpin: Summary

This PR is part of a larger set of PRs: see https://github.com/hashicorp/design-system/pull/2016

It's mainly a porting (with cleanup/refactoring) of #1947 

The reason for creating a custom register for DOM event listeners is because the native `{{on}}` Ember modifier can't be imported directly in the backing class (to apply it conditionally). See https://github.com/emberjs/ember.js/issues/19869#issuecomment-1909118910 / https://github.com/emberjs/ember.js/pull/20629. Thanks @fivetanley for putting me on the right track ;) 

### :hammer_and_wrench: Detailed description

In this PR I have:
- introduced a custom `hds-register-event` modifier to assign/remove event listeners to DOM elements
- added a simple integration test for the modifier
  - _I am totally open to ideas of what other tests I could add_

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-3211 (Main task)

***

### 👀 Component checklist

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#templates-for-npm-packages))

PR #2045, opened 2024-04-10
`RichTooltip/Popover` component - `PopoverPrimitive` component implementation [03]

didoo:
### :pushpin: Summary

This PR is part of a larger set of PRs: see https://github.com/hashicorp/design-system/pull/2069

It started from a porting (with cleanup/refactoring) of #1947 but then evolved to its own line of work with a second "spike" to iron out all the details, bugs, special cases, etc in https://github.com/hashicorp/design-system/pull/2044 and then I've carried on iterating on it (most of the changes have been squashed to have a more meaningful history of commits).

This PR introduces a `PopoverPrimitive` built  upon #2020  and #2017 to provide:
- **anchoring** (where an element position is “anchored” to a target element) 
- **collision detection** (when the position of the element in relation to the anchor is changed dynamically to avoid collision with the browser window/viewport).
- **popover** (where some content is laid out over the rest of the page, and its visibility is turned on/off via an associated toggle element)

The primitive will be used in #2048 to implement the `RichTooltip` component. It's been built quite generic and with **a lot** of options to configure and adapt it to most of the use cases we could think of. For this reason the API is pretty complex, and so I've tried to come up with the best naming convention possible (but open to suggestions).

In this PR I have:
- added the `PopoverPrimitive` component
- added `@oddbird/popover-polyfill` as dependency (used to polyfill the Popover API for Firefox)
- added the showcase page for the `PopoverPrimitive`
- added integration tests
- added visual regression test
    - _unfortunately for reasons that I don't understand (it's not a JS error) the popovers are not rendered in the Percy screenshot, maybe because it uses the top layer?_

### 👓 Preview

- showcase: https://hds-showcase-git-03-popover-primitive-implementation-hashicorp.vercel.app/utilities/popover-primitive
- documentation (from #2046): https://hds-website-git-03b-popover-primitive-documentation-hashicorp.vercel.app/utilities/popover-primitive

### :link: External links

Figma file: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/ZiYg2uadc2fCNQ5tMzQqi8/HDS-Product---Components?type=design&node-id=55408-54938&mode=design&t=qDUIxW9WBWLLM1HG-0

***

### 👀 Component checklist

- [x] Tested in multiple browsers
    - Chrome 123
    - Safari 17.3
    - Edge 123
    - Firefox 124 (no support for Popover API)
    - Firefox 125 (Popover API supported)
- [x] axe devtools + lighthouse accessibility passed
    - screenshot_3668
    - screenshot_3669


packages/components/src/components/hds/popover-primitive/index.hbs
```diff
@@ -0,0 +1,44 @@
+{{! @glint-nocheck: not typesafe yet }}
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<div class={{this.classNames}} ...attributes {{this.setupPrimitiveContainer}}>
+  {{#if this.toggleContainsInteractive}}
+    <div
+      class="hds-popover-primitive__toggle"
+      aria-controls={{this.popoverId}}
```
MelSumner, line 10: https://www.w3.org/TR/wai-aria-1.2/#aria-controls

This `div` won't ever be able to control the popover content because it's not a control (interactive). 

Is there a way for you to forward these attributes to the yielded interactive element?

packages/components/src/components/hds/popover-primitive/index.hbs
```diff
@@ -0,0 +1,44 @@
...
+    >
+      {{yield (hash popoverElement=this.popoverElement isOpen=this.isOpen) to="toggle"}}
+    </div>
+  {{else}}
+    <button
+      type="button"
+      class="hds-popover-primitive__toggle"
+      aria-controls={{this.popoverId}}
+      aria-describedby={{this.popoverId}}
+      aria-details={{this.popoverId}}
```
MelSumner, line 23: Question: why do you have both `aria-describedby` and `aria-details` with the same values used here?

MelSumner (changes requested):

packages/components/src/components/hds/popover-primitive/index.hbs
didoo: > https://www.w3.org/TR/wai-aria-1.2/#aria-controls
> This `div` won't ever be able to control the popover content because it's not a control (interactive).

Ok, so I have to remove it?

> Is there a way for you to forward these attributes to the yielded interactive element? Perhaps a helper?

As mentioned in other conversations, no it's not possible.

packages/components/src/components/hds/popover-primitive/index.hbs
didoo: I _think_ I have found a code example of a tooltip using it (but I can't find it anymore, apparently). Reading this https://www.scottohara.me/blog/2018/11/08/aria-details.html my understanding was that `aria-details` provides an extra layer of information. If that's not the case, I can remove one of the two (which one, in that case?).

packages/components/src/components/hds/popover-primitive/index.hbs
MelSumner: When I look at the flyout's `.hbs` file, I see that we're forwarding the `id` attribute inside of the yield's hash. I think we could try a similar approach here, WDYT?

packages/components/src/components/hds/popover-primitive/index.hbs
MelSumner: Ahh okay. We can safely remove `aria-details`, then.

didoo: done dc10306062651c106f8c22d71e79280a2ae98081

packages/components/src/components/hds/popover-primitive/index.hbs
MelSumner: well, we already know the value will be `{{this.popoverId}}` for the relevant `aria-` attributes, yes? They wouldn't have to do anything, I wouldn't think? 


But we do have a concise list of what could be the trigger, from the API: 

> "Invoking attributes" are only supported on buttons (see [https://open-ui.org/components/popover.research.explainer](https://open-ui.org/components/popover.research.explainer/#:~:text=Invoking%20attributes%20only%20supported%20on%20buttons)). From what I can tell, this is what that means:
`<button type=[button|submit|reset]>` (if the button would not otherwise submit a form)
`<input type=[button|submit|reset|image]>` (if it would not otherwise submit a form)
`<input type=[text|email|password|search|tel|url]>` (activated by down arrow only)

didoo: The toggle works in two ways:
- it can be activated via “soft” events (hover/focus) using the native JS methods [`showPopover`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/showPopover), [`hidePopover`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidePopover), [`togglePopover`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/togglePopover)
    - https://github.com/hashicorp/design-system/blob/03-popover-primitive-implementation/packages/components/src/components/hds/popover-primitive/index.js#L77-L83
    - https://github.com/hashicorp/design-system/blob/03-popover-primitive-implementation/packages/components/src/components/hds/popover-primitive/index.js#L242-L255
- it can be activated via”click” events using the native [`popovertarget`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#popovertarget) attribute (we assign it via JS to the `<button class="hds-rich-tooltip__toggle">`)
   - https://github.com/hashicorp/design-system/blob/03-popover-primitive-implementation/packages/components/src/components/hds/popover-primitive/index.js#L99-L110

The first way ("soft" events) is the default one (and preferred one) in the `RichTooltip` component; for that case, any element can be used as content of the "toggle" container (interactive as well as non-interactive elements).

packages/components/src/components/hds/popover-primitive/index.js
```diff
@@ -0,0 +1,334 @@
...
+    // TODO! discuss with Alex if/why we need this check here
+    // due to inconsistent implementation of relatedTarget across browsers we use the activeElement as a fallback
+    // if the related target is not part of the disclosed content we close the disclosed container
+    if (
+      !this.containerElement.contains(
+        event.relatedTarget || document.activeElement
+      )
+    ) {
+      this.hidePopover();
+    }
```
alex-ju, line 332: This logic was set in place in the `MenuPrimitive` to ensure the revealed content disappears only when the focus is moved away from the component (trigger or content). If you remove this check you will not be able to tab into the popover content; the focus will move to the next element on the page.

### With this check
PopoverPrimitive-focusout-logic


### Without this check
PopoverPrimitive-focusout--no-logic

alex-ju (approved):
Looking good overall! 👏  Left a few comments here and there, but no blockers.

> unfortunately for reasons that I don't understand (it's not a JS error) the popovers are not rendered in the Percy screenshot, maybe because it uses the top layer?
We have other components using the top layer (Modal, Flyout) and this wasn't a problem. I can dig a bit into this when time allows, but I wouldn't stall the PR for this.

showcase/app/styles/showcase-components/autoscrollable.scss
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+.shw-autoscrollable__container {
+  position: relative;
+  display: grid;
+  align-items: center;
+  justify-items: center;
+  width: 100%;
```
KristinLBradley, line 11: Nit: `width: 100%` doesn't appear to be needed

packages/components/src/components/hds/popover-primitive/index.js
```diff
@@ -0,0 +1,245 @@
...
+    { eager: true }
+  );
+
+  setupPrimitiveToggle = modifier(
+    (element) => {
+      this.toggleElement = element;
+
+      assert(
+        `The toggle element of "Hds::PopoverPrimitive" must be a <button>; element received: <${element.tagName.toLowerCase()}>`,
+        element instanceof HTMLButtonElement
```
KristinLBradley, line 82: Does it have to be a button tag? `<button type="button>Example HTML button</button>` or does it also work with `<input type="button" value="Example Input button" />`?

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/popoverTargetAction

didoo: Technically yes, it supports also the `<input type="button" />` but I would avoid even suggesting the idea that they should use it.

showcase/app/styles/showcase-components/autoscrollable.scss
```diff
@@ -0,0 +1,17 @@
...
+  aspect-ratio: 1;
+  overflow-x: auto;
+  overflow-y: auto;
```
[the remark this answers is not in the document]
didoo: Yes, you're also right!

packages/components/src/components/hds/popover-primitive/index.js
```diff
@@ -0,0 +1,245 @@
...
+        );
+      });
+    },
+    { eager: true }
+  );
+
+  @action
+  showPopover() {
+    this.popoverElement.showPopover();
+  }
```
alex-ju, line 156: I suggest we do some error handling on the `showPopover`, `hidePopover`, `togglePopover` web API methods (a try/catch) – in case the 'bubble' element doesn't exist (e.g. gets removed for some reasons) it will likely throw a hard error.

packages/components/src/components/hds/popover-primitive/index.js
```diff
@@ -0,0 +1,245 @@
...
+      if (this.timer) {
+        clearTimeout(this.timer);
+      }
+      this.showPopover();
+    }
+  }
+
+  @action
+  onMouseLeave() {
+    this.timer = setTimeout(this.hidePopover.bind(this), 500);
```
alex-ju, line 230: Not too opinionated about this, but it can also be writted as `this.timer = setTimeout(() => this.hidePopover(), 500);` depending on whether folks in our team read an arrow function easier than a bind method (I found this format to be more popular in our library)

packages/components/src/components/hds/popover-primitive/index.js
```diff
@@ -0,0 +1,245 @@
...
+} from '../../../modifiers/hds-anchored-position';
+
+export default class HdsPopoverPrimitiveComponent extends Component {
+  @tracked isOpen = this.args.isOpen ?? false;
+  @tracked isClosing = false;
+  // this will enable "soft" events for the toggle ("hover" and "focus")
+  @tracked enableSoftEvents = this.args.enableSoftEvents ?? false;
+  // this will enable "click" events for the toggle
+  @tracked enableClickEvents = this.args.enableClickEvents ?? false;
+
```
alex-ju, line 42: I believe challenged these properties need to be tracked in the previous review – any thoughts on this?

alex-ju (commented):
Generally looks good. I left a couple of suggestions/questions (some have been also raised in the initial review)

packages/components/src/components/hds/popover-primitive/index.js
didoo: I _think_ I made the change but then it may have been lost in some rebase. You're absolutely right, I'll remove the tracking.

packages/components/src/components/hds/popover-primitive/index.js
didoo: @alex-ju would you expect something like this:
```
@action
showPopover() {
  try {
    this.popoverElement.showPopover();
  } catch (error) {
    warn(
      'The invocation of `showPopover` for the popover element caused an unexpected error.',
      {
        id: 'hds-popover.show-popover-action.invocation-failed',
        error: error,
      }
    );
  }
}
```
or like this?
```
@action
showPopover() {
  if (this.popoverElement) {
    this.popoverElement.showPopover();
  }
}
```
or something else?

didoo: OK, done. See my last commit.

alex-ju (approved):

MelSumner (approved):
As discussed in Slack, I think we should ship it and watch to see how it's used by consumers.

PR #2048, opened 2024-04-11
`RichTooltip/Popover` component - `RichTooltip` component implementation [04] 

didoo:
### :pushpin: Summary

This PR is part of a larger set of PRs: see https://github.com/hashicorp/design-system/pull/2069

The code in this PR is the result of all the explorations done in #1947 and #2044 and then the iterations done within #2045. Most of the work has been squashed into single commits, to have a more meaningful history.

This PR introduces a new `RichTooltip` built  upon the `PopoverPrimitive` built in #2045.

In this PR I have:
- added the `RichTooltip` component
    - it's based off the `PopoverPrimitive` customizing it via `@...` options and CSS overrides.
- added the showcase page for the `RichTooltip`
    - very similar to the `PopoverPrimitive` but with subtle differences here and there
- added integration tests
- added visual regression test
    - _unfortunately for reasons that I don't understand (it's not a JS error) the popovers are not rendered in the Percy screenshot, maybe because it uses the top layer?_

### 👓 Preview

- showcase: https://hds-showcase-git-04-richtooltip-component-impl-6f74a1-hashicorp.vercel.app/components/rich-tooltip
- documentation: https://hds-website-git-04b-richtooltip-component-docu-0eb425-hashicorp.vercel.app/components/rich-tooltip

### :link: External links

Figma file: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/ZiYg2uadc2fCNQ5tMzQqi8/HDS-Product---Components?type=design&node-id=55408-54938&mode=design&t=qDUIxW9WBWLLM1HG-0

***

### 👀 Component checklist

- [x] Tested in multiple browsers
    - Chrome 123
    - Safari 17.3
    - Edge 123
    - Firefox 124 (no support for Popover API)
    - Firefox 125 (Popover API supported)
- [x] axe devtools + lighthouse accessibility passed
    -  screenshot_3670
    - screenshot_3671


jorytindall (commented):
I think visually speaking everything is consistent between design and code here, I have a few questions:

When a custom width is passed to the "bubble", the absolute position of the arrow changes, like in this example instead of being 12px from edge is 42.88. Is that intentional or even something we can control?
Screenshot 2024-04-22 at 2 16 35 PM

I'm trying to remember where we landed on changing out the icon, is this correct: exposed in the InfoText component in Ember so that it can be overridden if necessary, but not in the Figma component to enforce a bit of friction (consumers can still technically swap it out, but we don't expose a property)

I'll do a bit more VQA just to double check, but I think this is looking great!

didoo:
> When a custom width is passed to the "bubble", the absolute position of the arrow changes, like in this example instead of being 12px from edge is 42.88. Is that intentional or even something we can control? Screenshot 2024-04-22 at 2 16 35 PM

@jorytindall it's not just the width per se, it's a (complex) combination of factors (width/height of the toggle, width/height of the popover, their position in relation to the viewport and the placement algorithm declared by the user, the "padding" of the popover, the "padding" of the arrow) that determines the position of the arrow, and it's controlled by the Floating UI library. In essence it tries to center the arrow with the "toggle", but at the same time preserve _some_ of the user options. At times seems to not follow the consumer's options, but it's not the case, it's that some of these factors are impacting the position of the arrow.

If you think this is something that could be confusing for the developers, I can add a note about it in the documentation.

> I'm trying to remember where we landed on changing out the icon, is this correct: exposed in the InfoText component in Ember so that it can be overridden if necessary, but not in the Figma component to enforce a bit of friction (consumers can still technically swap it out, but we don't expose a property)

Yes, that's what's happening. In code at the moment there is no "default", consumers have to explicitly use the `@icon` argument to add an icon, and they can use any icon they want. In Figma instead we decided to give some friction to the process by not exposing a prop for it (but if they know how this works, they can always replace the instance inspecting the "tree" of the component.

> I'll do a bit more VQA just to double check, but I think this is looking great!

👍

showcase/app/templates/components/rich-tooltip.hbs
```diff
@@ -0,0 +1,1223 @@
...
+      </Shw::Grid>
+    {{/each}}
+  {{/let}}
+
+  <Shw::Divider />
+
+  <Shw::Text::H2>Demo</Shw::Text::H2>
+
+  <Shw::Flex as |SF|>
+    <SF.Item @label="Inline with other text" {{style width="250px"}}>
```
MelSumner, line 730: Question: I thought that if it's inline with other text, it should not have an icon? @jorytindall can you confirm?

didoo: @MelSumner this is the showcase page. The design guidance may be different, but we have to make sure the component is resilient in case consumers use it in other ways (that may not be suggested, but still accessible/valid).

jorytindall: I've added a recommendation for this 👍

alex-ju (approved):
I played quite extensively with it and couldn't spot anything off, looks pretty solid. I left a few thoughts, but no blockers.

showcase/app/templates/components/rich-tooltip.hbs
```diff
@@ -0,0 +1,1095 @@
...
+      </SG.Item>
+    {{/each}}
+  </Shw::Grid>
+
+  <Shw::Divider @level={{2}} />
+
+  <Shw::Text::H3>Generic</Shw::Text::H3>
+
+  <Shw::Text::Body>
+    In this case the "toggle" doesn't have special styles of its own other than for the "focus" state.
```
MelSumner, line 583:
```suggestion
    Not for use in production, but potentially for future use cases. Note that In this case the "toggle" doesn't have special styles of its own other than for the "focus" state.
```

Suggesting some additional wording that makes it clear that we don't intend consumers to use this and it's for future use cases that might come up and need extra styling.

MelSumner:
I think we still have a few a11y issues: 

CleanShot 2024-04-30 at 14 34 53

When inline with text, the interactive text must be discernible as interactive and have the correct color contrast ratio to the surrounding text.

didoo:
> I think we still have a few a11y issues:
> 
> CleanShot 2024-04-30 at 14 34 53
> 
> When inline with text, the interactive text must be discernible as interactive and have the correct color contrast ratio to the surrounding text.

These are examples only to test the resilience of the component. This is a generic container, and we don't know what the consumer will put into it (hence the clear disclaimers we added in the documentation, about this "generic" option). It may be as well that consumers add their own styling applied to the text provided/yielded, making it accessible. We don't have a way to know it, it's their responsibility. Here the showcase is testing that the layout (inline) works as expected also with a generic content.

showcase/app/templates/components/rich-tooltip.hbs
didoo: added

alex-ju (commented):
I've not found oportunities for improvement in the component code, but there are a few examples/usecases with issues (aside from the know Safari issue that I assume is under now investigation).

didoo:
> the known Safari issue that I assume is now under investigation

Yes, started this morning but then I stopped to focus on the SmartSelect review.

> There's an excess space in this example that I couldn't pin; I'm concerned it comes from the component

@alex-ju It's a mix of component code and component instance. I want to test more tomorrow morning, to avoid overdoing "whitespace killing".

> If we have only an icon in the Toggle and we set `@isInline={{true}}` there is an obvious misalignment between the icon and the text. 

Nice catch! 😮 Didn't think of this use case (inline icon-only). I've added the use case to the showcase, and found an OK fix (`vertical-align: middle`).

> we may need to set a different `display` for when we have an icon only instance

Unfortunately the toggle is a `<button>` element, so there is not much choice there (it can't be `display: inline`).

jorytindall (approved):

jorytindall:
No feedback aside from what I added previously which I think we've resolved, but this is a pretty technical component for me to review 🫣. Tested in the showcase and everything seems to be working as expected! :shipit:

MelSumner (approved):
So I think we've reviewed this a lot and we should just ship it and see how consumers implement it. 

We might need to tweak things here and there but we can do that in follow-up PRs.

didoo:
> We might need to tweak things here and there but we can do that in follow-up PRs.

@MelSumner can you expand on this? if possible I would avoid having to do work after the release if we have an opportunity to do it before the component is adopted across the consumers' codebases (which increases considerably the "cost of change").

packages/components/src/components/hds/rich-tooltip/index.hbs
```diff
@@ -0,0 +1,36 @@
...
+  @popoverWidth={{@popoverWidth}}
+  @popoverHeight={{@popoverHeight}}
+  @anchoredPositionOptions={{this.anchoredPositionOptions}}
+  ...attributes
+>
+  <:toggle as |t|>
+    {{yield
+      (hash
+        ToggleInfoText=(component "hds/rich-tooltip/toggle/info-text" isInline=@isInline)
+        ToggleGeneric=(component "hds/yield")
```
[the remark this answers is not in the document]
didoo: Outdated, resolving the comment.

packages/components/src/components/hds/rich-tooltip/index.js
```diff
@@ -0,0 +1,51 @@
...
+import { assert } from '@ember/debug';
+
+import { PLACEMENTS as PRIMITIVE_PLACEMENTS } from '../popover-primitive/index';
+
+export const DEFAULT_PLACEMENT = 'bottom';
+export const PLACEMENTS = [...PRIMITIVE_PLACEMENTS];
+
+export default class HdsRichTooltipIndexComponent extends Component {
+  /**
+   * @param placement
```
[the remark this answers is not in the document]
didoo: Outdated, resolving the comment.

packages/components/src/components/hds/rich-tooltip/toggle/info-text.hbs
```diff
@@ -0,0 +1,13 @@
+{{! @glint-nocheck: not typesafe yet }}
```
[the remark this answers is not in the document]
didoo: Outdated, resolving the comment.

alex-ju (approved):

MelSumner:
> @MelSumner can you expand on this? if possible I would avoid having to do work after the release if we have an opportunity to do it before the component is adopted across the consumers' codebases (which increases considerably the "cost of change").

There's nothing to expand on. We've done more than due diligence here, and we need to see how consumers will use it, and then if there are any adjustments to be made, we can make them later.

PR #2050, opened 2024-04-12
`Tooltip` - Fix max-width

didoo:
### :pushpin: Summary

While working on the `RichTooltip` I noticed that there is a difference between the (max) width of the `Tooltip` in Figma, and the `max-width` of the `Hds::Tooltip` in code. 

The difference is because the max width is applied to the content of the tooltip, and so we have to take into account also the padding of the "bubble" that acts as container.

In this PR I have:
- updated `max-width` property for the tooltip content to take into account the padding
- updated showcase to keep tooltips content on two lines

screenshot_3656

screenshot_3657

### :link: External links

Figma file: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/HDS-Product---Components?type=design&node-id=32036-51885&mode=design

***

### 👀 Component checklist

### DIFF
--- a/.changeset/chilly-avocados-serve.md
+++ b/.changeset/chilly-avocados-serve.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Tooltip` - Fixed max-width applied to the "bubble" (it was `304px`, now is `280px` per design specs)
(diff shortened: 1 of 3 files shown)


KristinLBradley (approved):

jorytindall (approved):
Good catch!

PR #2051, opened 2024-04-12
add image compression github action

Dhaulagiri:
### :pushpin: Summary

Introduces a GitHub action using [calibreapp/image-actions](https://github.com/calibreapp/image-actions) that can be invoked to open a PR with optimized image assets.

For the moment this is more intended to be used to experiment with the right setup, so the action will only happen if it is invoked directly. Future options for this could include either running it on a schedule or having it run against PRs that change files in a specific directory (i.e. `/website`). The latter may prove most useful since otherwise there is no way to scope this action to a specific folder hence by default it'll try to compress images across the entire monorepo which may be undesirable depending on which file types it picks up.

### :link: External links

Jira ticket: [HDS-1353](https://hashicorp.atlassian.net/browse/HDS-1353)

[HDS-1353]: https://hashicorp.atlassian.net/browse/HDS-1353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci-compress-images.yml
+++ b/.github/workflows/ci-compress-images.yml
@@ -1,0 +1,29 @@
+name: Compress Images
+on:
+  workflow_dispatch:
+  # schedule:
+  #   - cron: '00 23 * * 0'
+
+env:
+  NODE_VERSION: '20.x'
+
+jobs:
+  build:
+    name: Compress Images & Open PR
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: Compress Images
+        id: calibre
+        uses: calibreapp/image-actions@790ef463cba7b26ed3b1c80c1449deb72d4c0bcc # main
+        with:
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          compressOnly: true
+      - name: Open a PR if images can be compressed
+        if: steps.calibre.outputs.markdown != ''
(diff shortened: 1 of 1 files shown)


didoo:
> The latter may prove most useful since otherwise there is no way to scope this action to a specific folder

My understanding is that this is possible using the `paths` parameter, eg.

```
paths: 
    - '/website/public/assets/**.jpg'
    - '/website/public/assets/**.jpeg'
    - '/website/public/assets/**.png'
```

Dhaulagiri:
> My understanding is that this is possible using the paths parameter, eg.

Correct. That would only be possible if we have the action run against every PR that is opened vs something scheduled once a week.

.github/workflows/ci-compress-images.yml
```diff
@@ -0,0 +1,29 @@
+name: Compress Images
+on:
+  workflow_dispatch:
+  # schedule:
+  #   - cron: '00 23 * * 0'
```
alex-ju, line 5: noting here that (if I read it correctly) this means every Sunday at 11 pm

alex-ju (approved):
Looks good! Happy to give it a try manually and if we're happy with the result, we can turn on the cron.

.github/workflows/ci-compress-images.yml
Dhaulagiri: I believe that is correct, yes

PR #2065, opened 2024-04-25
Website + Showcase / Rebranding followup

didoo:
### :pushpin: Summary

Following the recent rebranding of HashiCorp, we have to update the assets and fonts used in the documentation website and showcase.

In this PR I have:
- website - replaced SVG paths in top header logo
- website - replaced SVG paths in bottom footer logo
- website - replaced “gilmer” font with “hashicorp-sans”
- showcase - replaced SVG paths in top header logo
- ~~showcase - replaced “gilmer” font with “hashicorp-sans”~~
    - to avoid all the visual regression tests to fail, I've left "gilmer" which works better for the showcase

### 👓 Preview:

- website: https://hds-website-git-rebranding-followup-hashicorp.vercel.app/
- showcase: https://hds-showcase-git-rebranding-followup-hashicorp.vercel.app/

image

### :link: External links

Figma file: https://www.figma.com/file/42LK10XbP5IERhzzgMOiI2/Website-assets?type=design&node-id=1-504&mode=design&t=oBx3Vj4bvQceSwy8-0

***

### 👀 Component checklist


website/app/styles/pages/home.scss
```diff
@@ -120,7 +120,7 @@ body.application.index {
   }
 
   p {
-    @include doc-font-family("gilmer");
+    @include doc-font-family("hashicorp-sans");
```
heatherlarsen, line 123: In looking at the hashicorp.com pages and from what I recall of the rebrand slides, the new font is for headings and logos, not body copy. Can this be updated to use the `sans` font-family instead? It probably should've been this way from the beginning and was just missed.

website/app/styles/typography/mixins.scss
```diff
@@ -170,7 +170,7 @@
 }
 
 @mixin doc-font-style-label {
-  @include doc-font-family("gilmer");
+  @include doc-font-family("hashicorp-sans");
```
heatherlarsen, line 173: Because this font was meant for Display text, it's a little hard to read at this size (it can be seen on the [icon library](https://hds-website-git-rebranding-followup-hashicorp.vercel.app/icons/library) page). Can this be updated to use the `sans` font-family instead?

KristinLBradley (approved):

heatherlarsen (changes requested):
I requested a few small changes that align more with the brand strategy around using this font. Otherwise, this looks great! Thanks for the quick turnaround. Per the PMs planning the rollout, we shouldn't release this until May 1.

website/app/styles/typography/mixins.scss
didoo: done

MelSumner (approved):
[question, non-blocker] I think it would probably be okay to change the showcase type and let Percy be updated; is there something I'm missing here?

didoo:
> [question, non-blocker] I think it would probably be okay to change the showcase type and let Percy be updated; is there something I'm missing here?

As mentioned above, the reason why I reverted the changes that I already made (I replaced "gilmer" in the showcase too) was that when looking at the pages, the new font would be hard to distinguish from the font of the HDS components, and the visual scan of the page was terrible. Gilmer is a good balance, in the sense that it's easy to distinguish but works also well visually.

heatherlarsen (approved):
Looks great! Thanks for those edits 🙂

PR #2064, opened 2024-04-25
Compressed Image Assets

hashibot-hds:
Images automagically compressed by [Calibre](https://calibreapp.com)'s [image-actions](https://github.com/marketplace/actions/image-actions) ✨

Compression reduced images by <strong>23.8%</strong>, saving <strong>1.40 MB</strong>.

| Filename | Before | After | Improvement |
| -------- | ------ | ----- | ----------- |
| <code>website/public/assets/components/accordion/content-custom.png</code> | 59.03 KB | 44.23 KB | -25.1% |
| <code>website/public/assets/components/alert/ordering-alerts-type.png</code> | 72.89 KB | 54.68 KB | -25.0% |
| <code>website/public/assets/components/app-footer/app-footer-additional-links.png</code> | 11.03 KB | 6.88 KB | -37.6% |
| <code>website/public/assets/components/app-footer/app-footer-anatomy.png</code> | 46.81 KB | 33.10 KB | -29.3% |
| <code>website/public/assets/components/app-footer/app-footer-dark.png</code> | 14.08 KB | 9.83 KB | -30.2% |
| <code>website/public/assets/components/app-footer/app-footer-extra-content-do.png</code> | 18.44 KB | 11.99 KB | -35.0% |
| <code>website/public/assets/components/app-footer/app-footer-extra-content.png</code> | 16.21 KB | 10.64 KB | -34.3% |
| <code>website/public/assets/components/app-footer/app-footer-items.png</code> | 9.97 KB | 6.32 KB | -36.6% |
| <code>website/public/assets/components/app-footer/app-footer-large.png</code> | 15.71 KB | 9.92 KB | -36.9% |
| <code>website/public/assets/components/app-footer/app-footer-legal-links.png</code> | 12.40 KB | 7.82 KB | -37.0% |
| <code>website/public/assets/components/app-footer/app-footer-order-priority.png</code> | 33.19 KB | 26.85 KB | -19.1% |
| <code>website/public/assets/components/app-footer/app-footer-small.png</code> | 21.28 KB | 13.45 KB | -36.8% |
| <code>website/public/assets/components/badge/badge-inverted-dont.png</code> | 34.07 KB | 24.49 KB | -28.1% |
| <code>website/public/assets/components/code-block/code-block-focus-header.png</code> | 28.75 KB | 22.97 KB | -20.1% |
| <code>website/public/assets/components/code-block/code-block-focus-no-header.png</code> | 18.31 KB | 15.08 KB | -17.7% |
[...]

PR #2066, opened 2024-04-26
run image compression action weekly

Dhaulagiri:
### :pushpin: Summary

https://github.com/hashicorp/design-system/pull/2064 seems to have worked well, so this change will automatically run the action every Sunday evening and open a PR only if there are changes to be merged. The action can also still be triggered manually as needed.


MelSumner (approved):
Let's do it!

PR #2072, opened 2024-05-01
Adjust guidance for forms in Modal

alex-ju:
### :pushpin: Summary

Follow up on a (2 months old now 🙈) discussion on Slack around a recommended pattern for forms in Modal dialogs. The guidance could potentially be applied to the Flyout component, but I restrained from doing so as we don't recommend using forms in Flyouts.

### :link: External links

Jira ticket: [HDS-3144](https://hashicorp.atlassian.net/browse/HDS-3144)

***

[HDS-3144]: https://hashicorp.atlassian.net/browse/HDS-3144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/docs/components/modal/partials/code/how-to-use.md
+++ b/website/docs/components/modal/partials/code/how-to-use.md
@@ -26,7 +26,11 @@ When a Modal is opened with the keyboard, the focus is automatically set to the 
 />
 
 {{#if this.basicModalActive}}
-  <Hds::Modal id="basic-modal" @onClose={{fn this.deactivateModal "basicModalActive"}} as |M|>
+  <Hds::Modal
+    id="basic-modal"
+    @onClose={{fn this.deactivateModal "basicModalActive"}}
+    as |M|
+  >
     <M.Header>
       Modal title
     </M.Header>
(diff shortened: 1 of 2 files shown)


website/docs/components/modal/partials/code/how-to-use.md
```diff
@@ -63,7 +69,10 @@ When the Modal dialog contains information that might be lost on close, use a co
       Why do you want to leave the beta?
     </M.Header>
     <M.Body>
-      <form name="leaving-beta-form">
+      <form
+        id="leaving-beta-form"
+        {{on "submit" (fn this.deactivateModal "formModalActive")}}
```
didoo, line 74: to avoid confusion, could it make sense to create another function in the backing class, called `submitForm` and use it here? (even if technically what it does, for the purpose of the demo, is still deactivate the modal)

website/docs/components/modal/partials/code/how-to-use.md
```diff
@@ -42,7 +46,9 @@ When a Modal is opened with the keyboard, the focus is automatically set to the
 
 ### Form within a Modal dialog
 
-If a Modal dialog contains interactive elements, such as a form, the initial focus should be set on the first input, which is the first focusable element within the form. This can be achieved by setting the `autofocus` property on the first form element.
+If a Modal dialog presents a form, the initial focus should be set on the first input, as the first focusable element in the form. This can be achieved by setting the `autofocus` property on the first form input.
+
+We recommend placing the `<form>` element in the `Hds::Modal::Body` subcomponent and associating it to the submit button using the `form` attribute, as shown below.
```
didoo, line 51: Is this a recommendation or a "requirement" to work correctly? If the latter, maybe we could use a more direct language, something like this (feel free to edit/adapt)?

```suggestion
The `<form>` element should be placed in the `Hds::Modal::Body` subcomponent. We also recommend to associate it to the submit button using the `form` attribute, as shown below.
```

didoo (approved):
Left a couple of comments, but not blockers.

Nice improvement! 👏

website/docs/components/modal/partials/code/how-to-use.md
alex-ju: sure, will do so

website/docs/components/modal/partials/code/how-to-use.md
alex-ju: sounds clearer – will update

PR #2073, opened 2024-05-02
Add examples of `Tag` components with long text to the showcase

didoo:
### :pushpin: Summary

While reviewing the `SuperSelect` component I noticed that we don't have examples of the `Tag` instances with long text in the showcase.

In this PR I have:
- added `Tag` examples with long text
- added dividers for better visual scanning

_As you can see, the background of the button overlaps the border of the `Tag`. Unfortunately there's not much we can do, because the width of the button is fixed so it can't be half of a circle._

👉 👉 👉 Preview: https://hds-showcase-git-tag-showcase-long-text-hashicorp.vercel.app/components/tag

screenshot_3714
screenshot_3713

***

### 👀 Component checklist

### DIFF
--- a/showcase/app/templates/components/tag.hbs
+++ b/showcase/app/templates/components/tag.hbs
@@ -42,6 +42,8 @@
     </SF.Item>
   </Shw::Flex>
 
+  <Shw::Divider @level={{2}} />
+
   <Shw::Text::H2>States</Shw::Text::H2>
 
   <Shw::Grid @columns={{4}} {{style width="max-content"}} as |SG|>
@@ -51,6 +53,8 @@
       </SG.Item>
     {{/each}}
   </Shw::Grid>
+
+  <Shw::Divider @level={{2}} />
 
   <Shw::Text::H2>Colors</Shw::Text::H2>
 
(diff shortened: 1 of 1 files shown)


MelSumner:
> As you can see, the background of the button overlaps the border of the Tag. Unfortunately there's not much we can do, because the width of the button is fixed so it can't be half of a circle.

Can't we clip the background?

didoo:
> > As you can see, the background of the button overlaps the border of the Tag. Unfortunately there's not much we can do, because the width of the button is fixed so it can't be half of a circle.
> 
> Can't we clip the background?

How would you do it? 🤔

MelSumner:
> How would you do it? 🤔

If you add `background-clip: content-box` to the dismiss, you get this:

CleanShot 2024-05-08 at 13 22 50

didoo:
> > How would you do it? 🤔
> 
> If you add `background-clip: content-box` to the dismiss, you get this:

@MelSumner in that case, the hover/active states would be impacted too:
screenshot_3734

The reason is because `background-clip: content-box` means that the background color is applied to the box _excluding_ the padding. The issue in this case is that the `hds-tag__dismiss` element overlaps the border of the parent `hds-tag` element. It's more clear using a tinted background:
screenshot_3735

With the `background-clip: content-box` this happens:
screenshot_3736

In _theory_ the solution would be to apply an `overflow: hidden` to the parent, but then the focus will be truncated (because of how we implement the focus for our components, using pseudo-elements):
modified

PR #2080, opened 2024-05-07
Add scoll to top feature to website

Dhaulagiri:
### :pushpin: Summary

This PR introduces a feature to the website that affords users a button that can be clicked to return to the top of the current page. 

Implementation choices:
* Only appears after user has scrolled some (configurable) distance down the page
* Appears on any page using the `<Doc::Page::Content>` component, which is pretty much every page except the homepage

scroll

### :link: External links

Jira ticket: [HDS-3298](https://hashicorp.atlassian.net/browse/HDS-3298)
[Figma file](https://www.figma.com/file/Ky0qWjvHZR3je1lCBlzNB1/HDS-website?type=design&node-id=1305-116325&mode=design&t=CgwVTcyQ84atmUzD-0)

***

[HDS-3298]: https://hashicorp.atlassian.net/browse/HDS-3298?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/components/doc/scroll-to-top/index.hbs
+++ b/website/app/components/doc/scroll-to-top/index.hbs
@@ -1,5 +1,11 @@
 {{#if this.isButtonVisible}}
-  <button aria-label="Scroll back to top" class="doc-scroll-to-top" type="button" tabindex="0" {{on "click" this.scrollToTop}}>
+  <button
+    aria-label="Scroll back to top"
+    class="doc-scroll-to-top"
+    type="button"
+    tabindex="0"
+    {{on "click" this.scrollToTop}}
+  >
     <FlightIcon @name="arrow-up" @color="white" @size="24" />
   </button>
 {{/if}}
--- a/website/app/components/doc/scroll-to-top/index.hbs
+++ b/website/app/components/doc/scroll-to-top/index.hbs
@@ -1,5 +1,5 @@
 {{#if this.isButtonVisible}}
-  <button class="doc-scroll-to-top" type="button" {{on "click" this.scrollToTop}}>
+  <button class="doc-scroll-to-top" type="button" tabindex="0" {{on "click" this.scrollToTop}}>
     <FlightIcon @name="arrow-up" @color="white" @size="24" />
   </button>
 {{/if}}
--- a/website/app/components/doc/scroll-to-top/index.hbs
+++ b/website/app/components/doc/scroll-to-top/index.hbs
@@ -1,5 +1,5 @@
 {{#if this.isButtonVisible}}
-  <button class="doc-scroll-to-top" type="button" tabindex="0" {{on "click" this.scrollToTop}}>
+  <button aria-label="Back to Top" class="doc-scroll-to-top" type="button" tabindex="0" {{on "click" this.scrollToTop}}>
     <FlightIcon @name="arrow-up" @color="white" @size="24" />
   </button>
 {{/if}}
--- a/website/app/components/doc/scroll-to-top/index.hbs
+++ b/website/app/components/doc/scroll-to-top/index.hbs
@@ -1,5 +1,5 @@
 {{#if this.isButtonVisible}}
-  <button aria-label="Back to Top" class="doc-scroll-to-top" type="button" tabindex="0" {{on "click" this.scrollToTop}}>
+  <button aria-label="Scroll back to top" class="doc-scroll-to-top" type="button" tabindex="0" {{on "click" this.scrollToTop}}>
     <FlightIcon @name="arrow-up" @color="white" @size="24" />
   </button>
 {{/if}}
--- a/website/app/components/doc/scroll-to-top/index.hbs
+++ b/website/app/components/doc/scroll-to-top/index.hbs
@@ -1,11 +1,9 @@
-{{#if this.isButtonVisible}}
-  <button
-    aria-label="Scroll back to top"
-    class="doc-scroll-to-top"
-    type="button"
-    tabindex="0"
-    {{on "click" this.scrollToTop}}
-  >
-    <FlightIcon @name="arrow-up" @color="white" @size="24" @isInlineBlock={{false}} />
-  </button>
-{{/if}}
+<button
+  aria-label="Scroll back to top"
+  class="doc-scroll-to-top {{if this.isButtonVisible 'doc-scroll-to-top--is-visible'}}"
+  type="button"
+  tabindex="0"
+  {{on "click" this.scrollToTop}}
+>
+  <FlightIcon @name="arrow-up" @color="white" @size="24" @isInlineBlock={{false}} />
+</button>
--- a/website/app/components/doc/scroll-to-top/index.hbs
+++ b/website/app/components/doc/scroll-to-top/index.hbs
@@ -6,6 +6,6 @@
     tabindex="0"
     {{on "click" this.scrollToTop}}
   >
-    <FlightIcon @name="arrow-up" @color="white" @size="24" />
+    <FlightIcon @name="arrow-up" @color="white" @size="24" @isInlineBlock={{false}} />
   </button>
 {{/if}}
--- a/website/app/components/doc/scroll-to-top/index.hbs
+++ b/website/app/components/doc/scroll-to-top/index.hbs
@@ -1,3 +1,5 @@
-<button class="doc-scroll-to-top" type="button" {{on "click" this.scrollToTop}}>
-  <FlightIcon @name="arrow-up" @color="white" @size="24" />
-</button>
+{{#if this.isButtonVisible}}
+  <button class="doc-scroll-to-top" type="button" {{on "click" this.scrollToTop}}>
+    <FlightIcon @name="arrow-up" @color="white" @size="24" />
+  </button>
+{{/if}}
--- a/website/app/components/doc/scroll-to-top/index.hbs
+++ b/website/app/components/doc/scroll-to-top/index.hbs
@@ -2,7 +2,7 @@
   aria-label="Scroll back to top"
   class="doc-scroll-to-top {{if this.isButtonVisible 'doc-scroll-to-top--is-visible'}}"
   type="button"
-  tabindex="0"
+  tabindex={{if this.isButtonVisible 0 -1}}
   {{on "click" this.scrollToTop}}
 >
   <FlightIcon @name="arrow-up" @color="white" @size="24" @isInlineBlock={{false}} />
--- a/website/app/components/doc/scroll-to-top/index.hbs
+++ b/website/app/components/doc/scroll-to-top/index.hbs
@@ -1,0 +1,3 @@
+<button class="doc-scroll-to-top" type="button" {{on "click" this.scrollToTop}}>
+  <FlightIcon @name="arrow-up" @color="white" @size="24" />
+</button>
--- a/website/app/components/doc/scroll-to-top/index.js
+++ b/website/app/components/doc/scroll-to-top/index.js
@@ -14,45 +14,39 @@ export default class DocScrollToTopComponent extends Component {
 
   constructor() {
     super(...arguments);
-    this.addScrollListener();
+    if (!this.fastboot.isFastBoot) {
+      this.addScrollListener();
+    }
   }
 
   willDestroy() {
     super.willDestroy(...arguments);
-    this.removeScrollListener();
+    if (!this.fastboot.isFastBoot) {
+      this.removeScrollListener();
+    }
   }
 
   @action
   scrollToTop() {
-    if (!this.fastboot.isFastBoot) {
-      window.scrollTo({
-        top: 0,
-        behavior: 'smooth',
-      });
-    }
+    window.scrollTo({ top: 0, behavior: 'smooth' });
   }
 
   @action
   checkScroll() {
-    if (window.scrollY > 200) {
-      this.isButtonVisible = true;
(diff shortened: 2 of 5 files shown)


alex-ju:
@heatherlarsen the CSS linter seems unhappy, but you can fix it by running `yarn lint:css:fix` in `website`

website/app/components/doc/scroll-to-top/index.hbs
```diff
@@ -0,0 +1,11 @@
+{{#if this.isButtonVisible}}
+  <button
+    aria-label="Scroll back to top"
+    class="doc-scroll-to-top"
```
didoo, line 4: we could consider adding a class here, instead of doing the `{{#if this.isButtonVisible}}` above, which shows/hides the button without transition:
```suggestion
    class="doc-scroll-to-top {{if this.isButtonVisible 'doc-scroll-to-top--is-visible'}}"
```
in this was we can set the default opacity to `0`, and transition it to `1` in the `.doc-scroll-to-top--is-visible' CSS selector. A bit more whimsy, what do you think @heatherlarsen ?

website/app/styles/doc-components/scroll-to-top.scss
```diff
@@ -0,0 +1,29 @@
+@use "../typography/mixins" as *;
+
```
didoo, line 2: not needed anymore
```suggestion
```

website/app/styles/doc-components/scroll-to-top.scss
```diff
@@ -0,0 +1,29 @@
...
+.doc-scroll-to-top {
+  position: fixed;
+  right: 40px;
+  bottom: 40px;
+  z-index: 99;
+  width: 44px;
+  height: 44px;
+  padding: 8px 8px 4px 8px;
```
didoo, line 10:
```suggestion
  padding: 8px;
```

website/app/styles/doc-components/scroll-to-top.scss
```diff
@@ -0,0 +1,29 @@
...
+  background-color: rgba(52, 53, 54, 80%);
+  border: 2px solid transparent;
+  border-radius: 50%;
+  box-shadow: 0 8px 12px 0 rgba(37, 38, 45, 8%);
+  cursor: pointer;
+  transition: bottom 150ms linear;
```
didoo, line 16: position transitions are [inefficient](https://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/), better to use transforms:
```suggestion
  transition: transform 150ms linear;
```

website/app/styles/doc-components/scroll-to-top.scss
```diff
@@ -0,0 +1,29 @@
...
+
+  &:hover {
+    bottom: 43px;
+    background-color: rgba(29, 30, 31, 80%);
+    box-shadow: 0 16px 28px 0 rgba(37, 38, 45, 12%);
+  }
```
didoo, line 22: see comment above about performance
```suggestion
  &:hover {
    background-color: rgba(29, 30, 31, 80%);
    box-shadow: 0 16px 28px 0 rgba(37, 38, 45, 12%);
    transform: translate3d(0, -3px, 0);
  }
```

website/app/styles/doc-components/scroll-to-top.scss
```diff
@@ -0,0 +1,29 @@
...
+
+  &:focus {
+    border-color: white;
+    outline: none;
+    box-shadow: 0 0 0 3px var(--doc-focus-ring-color);
```
didoo, line 27: we tend not to use HTML names for colors, and in the Figma designs the shadow is still visible in the focused state (correct me if I'm wrong @heatherlarsen)
```suggestion
    border-color: #fff;
    outline: 3px solid var(--doc-focus-ring-color);
```

website/app/styles/doc-components/scroll-to-top.scss
```diff
@@ -0,0 +1,29 @@
...
+  right: 40px;
+  bottom: 40px;
+  z-index: 99;
+  width: 44px;
+  height: 44px;
+  padding: 8px 8px 4px 8px;
+  background-color: rgba(52, 53, 54, 80%);
+  border: 2px solid transparent;
+  border-radius: 50%;
+  box-shadow: 0 8px 12px 0 rgba(37, 38, 45, 8%);
```
didoo, line 14: @heatherlarsen quick question, to double check: in the Figma designs I see the opacity is applied to the whole button (so it includes the icon and the focus ring) while here is only applied to the background color; I imagine this was a choice done in the implementation phase, is it correct? in that case is it worth it to update the Figma file to match the new designs? (if not, @Dhaulagiri you have to apply the opacity to the whole button, not just the background)

website/app/styles/doc-components/scroll-to-top.scss
didoo: @heatherlarsen is it ok a linear transition?

website/app/components/doc/scroll-to-top/index.js
```diff
@@ -0,0 +1,58 @@
...
+import { inject as service } from '@ember/service';
+import { tracked } from '@glimmer/tracking';
+
+export default class DocScrollToTopComponent extends Component {
+  @service fastboot;
+  @tracked isButtonVisible = false;
+
+  constructor() {
+    super(...arguments);
+    this.addScrollListener();
```
didoo, line 17: consider adding the check here, to keep the `addScrollListener` more agnostic
```suggestion
    if (!this.fastboot.isFastBoot) {
      this.addScrollListener();
    }
```

website/app/components/doc/scroll-to-top/index.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  willDestroy() {
+    super.willDestroy(...arguments);
+    this.removeScrollListener();
```
didoo, line 22: consider adding the check here, to keep the `removeScrollListener` more agnostic
```suggestion
    if (!this.fastboot.isFastBoot) {
      this.removeScrollListener();
    }
```

alex-ju (approved):
Looking good!

website/app/components/doc/scroll-to-top/index.js
```diff
@@ -0,0 +1,58 @@
...
+    } else {
+      this.isButtonVisible = false;
+    }
+  }
+
+  @action
+  addScrollListener() {
+    if (!this.fastboot.isFastBoot) {
+      this.scrollFunction = () => this.checkScroll();
+      window.addEventListener('scroll', this.scrollFunction);
```
didoo, line 48: are you sure you need to define a `scrollFunction`? 🤔
```suggestion
      window.addEventListener('scroll', this.checkScroll);
```

website/app/components/doc/scroll-to-top/index.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+  }
+
+  @action
+  removeScrollListener() {
+    if (!this.fastboot.isFastBoot) {
+      window.removeEventListener('scroll', this.scrollFunction);
```
didoo, line 55: see comment above
```suggestion
      window.removeEventListener('scroll', this.checkScroll);
```

website/app/components/doc/scroll-to-top/index.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @action
+  scrollToTop() {
+    if (!this.fastboot.isFastBoot) {
+      window.scrollTo({
+        top: 0,
+        behavior: 'smooth',
+      });
+    }
```
didoo, line 32: you don't need to check for fastboot here, it's an action invoked by the user
```suggestion
    window.scrollTo({ top: 0, behavior: 'smooth' });
```

website/app/components/doc/scroll-to-top/index.js
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  @action
+  checkScroll() {
+    if (window.scrollY > 200) {
+      this.isButtonVisible = true;
+    } else {
+      this.isButtonVisible = false;
+    }
```
didoo, line 41: consider this for brevity
```suggestion
    this.isButtonVisible = window.scrollY > 200;
```

didoo (commented):
Added a few comments and suggestions.

website/app/components/doc/scroll-to-top/index.js
Dhaulagiri: It doesn't look like it was necessary 👍

website/app/components/doc/scroll-to-top/index.hbs
heatherlarsen: I think that's a great idea. Let's do it.

website/app/styles/doc-components/scroll-to-top.scss
heatherlarsen: Thanks for these suggestions @didoo! It's been a long time since I've done CSS animations, and I wasn't trying to dig too deeply into it. Do you have another suggestion we should try? 

cc @jorytindall @majedelass any preferences/ideas for how to improve this animation?

website/app/styles/doc-components/scroll-to-top.scss
heatherlarsen: Ah, yeah, there were some tweaks to the styling during the implementation phase. It's probably a good idea to clean this up and have the opacity on the whole element instead of just the background color. Good catch!

website/app/components/doc/scroll-to-top/index.hbs
Dhaulagiri: Added in 000b87a24196a323423000626892b3ae0f8fac99

website/app/styles/doc-components/scroll-to-top.scss
Dhaulagiri: Is there something wrong with the current transition? It's not clear from the comments here.

heatherlarsen: I don't think there's anything "wrong" with it. I assume @didoo is proposing we do something more polished since "linear" is the default.

alex-ju (dismissed):
The approach to always have the button in the DOM with opacity 0 has usability flaws when the button overlaps another interactive element, such as a link on the page (see the example below where it partly overlaps with a doc card image). When the user clicks the area highlighted in red nothing will happen – as you can imagine the problem becomes more acute on narrow screens.

Screenshot 2024-05-20 at 10 15 04

didoo:
> The approach to always have the button in the DOM with opacity 0 has usability flaws when the button overlaps another interactive element, such as a link on the page (see the example below where it partly overlaps with a doc card image). When the user clicks the area highlighted in red nothing will happen – as you can imagine the problem becomes more acute on narrow screens.

You're right, I didn't think of this case. Do you reckon that adding a `pointer-events: none;` by default, and removing this style on `doc-scroll-to-top--is-visible` would be enough?

alex-ju:
> > The approach to always have the button in the DOM with opacity 0 has usability flaws when the button overlaps another interactive element, such as a link on the page (see the example below where it partly overlaps with a doc card image). When the user clicks the area highlighted in red nothing will happen – as you can imagine the problem becomes more acute on narrow screens.
> 
> You're right, I didn't think of this case. Do you reckon that adding a `pointer-events: none;` by default, and removing this style on `doc-scroll-to-top--is-visible` would be enough?

I've not tested `pointer-events`, it may allow the click to go through the button and reach the underneath target, but the issue I see with the button being in the DOM when not visible (in contrast with the original implementation) is that when navigating using keyboard and reaching that element the focus is lost (as in, it's on the button, but not visible, hence visually lost), violating criteria WCAG 2.4.7.

website/app/styles/doc-components/scroll-to-top.scss
jorytindall: I did some _very_ scrappy testing with a few different easing methods and it's pretty hard to tell the difference. The length of the page does slightly effect the easing so in the case of a very long page (e.g., table), it's more apparent. But it does seem to smooth out a small amount using `ease-in-out` instead of `linear`.

didoo:
> I've not tested `pointer-events`, it may allow the click to go through the button and reach the underneath target, but the issue I see with the button being in the DOM when not visible (in contrast with the original implementation) is that when navigating using keyboard and reaching that element the focus is lost (as in, it's on the button, but not visible, hence visually lost), violating WCAG criterion 2.4.7.

@alex-ju see my last commit: does this address all the issues?

website/app/styles/doc-components/scroll-to-top.scss
heatherlarsen: Thanks @jorytindall! This is more polished. @didoo, if this is a conformant version, then I say we go with Jory's version.

didoo: yes, already changed to `ease-in-out`.

website/app/styles/doc-components/scroll-to-top.scss
didoo: Fixed in https://github.com/hashicorp/design-system/pull/2080/commits/2c1abaaee2129d7c0ed60ded1bf67c6c3a0148e7

didoo (approved):
LGTM 👍

PR #2082, opened 2024-05-08
Fathom Event Tracking

Dhaulagiri:
### :pushpin: Summary

This PR creates an event tracking service to send events to Fathom and wires up a few proposed events as a starting point

Under the hood this is built using a Service which can be injected wherever we want to track events. I tried to keep the naming relatively agnostic as, in theory, Fathom could be replaced at some point.

Screenshot 2024-05-17 at 14 32 51

***

### DIFF
--- a/website/app/components/doc/link-with-icon/index.hbs
+++ b/website/app/components/doc/link-with-icon/index.hbs
@@ -4,13 +4,26 @@
 }}
 
 {{#if @route}}
-  <LinkTo class={{this.classNames}} @route={{@route}} @models={{this.models}} ...attributes>
+  <LinkTo
+    class={{this.classNames}}
+    @route={{@route}}
+    @models={{this.models}}
+    ...attributes
+    {{on "click" (fn this.eventTracking.trackEvent @eventName)}}
+  >
     {{@label}}
     <FlightIcon class="doc-link-with-icon__icon" @name={{@icon}} />
   </LinkTo>
 {{/if}}
 {{#if @href}}
-  <a class={{this.classNames}} href={{@href}} target="_blank" rel="noopener noreferrer" ...attributes>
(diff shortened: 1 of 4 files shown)


website/app/components/doc/link-with-icon/index.hbs
```diff
@@ -4,13 +4,26 @@
 }}
 
 {{#if @route}}
-  <LinkTo class={{this.classNames}} @route={{@route}} @models={{this.models}} ...attributes>
+  <LinkTo
+    class={{this.classNames}}
+    @route={{@route}}
+    @models={{this.models}}
+    ...attributes
+    {{on "click" (fn this.eventTracking.trackEvent @eventName)}}
```
didoo, line 12: @Dhaulagiri while discussing modifiers with @aklkv he mentioned that this could be a [custom modifier](https://github.com/ember-modifier/ember-modifier) (it would make the code cleaner). He can probably expand on this more, but something to consider.

Dhaulagiri: ooh, that sounds interesting. Admittedly I built this based on my very stale ember knowledge, so if there is a more idiomatic, current way to do this I'm open to refactoring this.

didoo (approved):
👍

Dhaulagiri:
I'm going to merge what I have here for now. I'll make a note about the modifier suggestion and add it to a follow-up epic I'm building for this work.

website/app/components/doc/link-with-icon/index.hbs
Dhaulagiri: [Opened a task to investigate this suggestion](https://hashicorp.atlassian.net/browse/HDS-3351)

PR #2085, opened 2024-05-09
`Tabs` - Fix issue with z-index of the active tab "indicator"

didoo:

In this PR I have:
- added `isolation: isolate` to the tabs list container

**Before:**
before

**After:**
after

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-3310

***

### 👀 Component checklist


packages/components/src/styles/components/tabs.scss
```diff
@@ -34,6 +34,7 @@ $hds-tabs-sizes: ( "medium", "large" );
   margin: 0;
   padding: 0;
   overflow-x: auto;
+  isolation: isolate;
```
KristinLBradley, line 37: This seems like a reasonable fix to me.

Did you also check the z-index of the .hds-tabs__tab-indicator though?

Code snippet:
```
.hds-tabs__tab-indicator {
  position: absolute;
  right: 0;
  bottom: 0;
  // notice: this custom prop is set dynamically via JavaScript
  left: var(--indicator-left-pos, 0);
  z-index: 10;
```

KristinLBradley (approved):
Left one comment to consider but looks good!

packages/components/src/styles/components/tabs.scss
didoo: @KristinLBradley what do you mean? that we could lower it (eg. to `1`)?

KristinLBradley: I just wondered if z-index: 10 might be too high? Anyway I think the fix you made is a good solution.

didoo: Yeah, no no it's a very good point. I am testing with `z-index: 1` and if there are no downsides (I don't think) I'll commit the change.

didoo: OK, made the change. Thanks again, good spot!

PR #2092, opened 2024-05-14
Added small comment to file with link mentioning Container Queries and z-index

didoo:
### :pushpin: Summary

Micro PR to add a comment with a link to [a resource](https://www.oddbird.net/2023/07/05/contain-root/#:~:text=Layout%20containment%20creates%20a%20new%20(z%2Dindex)%20stacking%20context%2C%20as%20well%20as%20containing%20floated%20and%20fixed%2Dposition%20elements.) that mentions the fact that _"Layout containment creates a new (z-index) stacking context"_ 

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/styles/components/page-header.scss
+++ b/packages/components/src/styles/components/page-header.scss
@@ -9,6 +9,7 @@
 
 .hds-page-header {
   // `container-type: inline-size` creates a new stacking context; to control the position of Page Header content on the z-axis
+  // see: https://www.oddbird.net/2023/07/05/contain-root/#:~:text=Layout%20containment%20creates%20a%20new%20(z%2Dindex)%20stacking%20context%2C%20as%20well%20as%20containing%20floated%20and%20fixed%2Dposition%20elements.
   // we set this element to `relative` to allow consumers to define the `z-index` value that works in their context
   position: relative;
   display: flex;


packages/components/src/styles/components/page-header.scss
```diff
@@ -9,6 +9,7 @@
 
 .hds-page-header {
   // `container-type: inline-size` creates a new stacking context; to control the position of Page Header content on the z-axis
+  // see: https://www.oddbird.net/2023/07/05/contain-root/#:~:text=Layout%20containment%20creates%20a%20new%20(z%2Dindex)%20stacking%20context%2C%20as%20well%20as%20containing%20floated%20and%20fixed%2Dposition%20elements.
```
alex-ju, line 12: I would avoid backing a blog post for this. I'd be happy to add details, but from more neutral sources (such as mdn)

didoo: I would be happy too, but - at least as far as I know - on MDN there are no mentions of the fact that containment creates a new stacking context which impacts the z-index computation. This is the first - I think - case I've found in which is clearly stated, that's why I jumped in to create a PR so if we need a reference we know where to find it.

didoo: OK, I've found this: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context#:~:text=Element%20with%20a%20container%2Dtype%20value%20size%20or%20inline%2Dsize%20set%2C%20intended%20for%20container%20queries but it's not as clear as the other, but happy to replace it (or add both)

alex-ju: agree, not very clear – if you find it helpful let's keep the original one then

alex-ju (approved):

PR #2087, opened 2024-05-10
Add TypeScript migration docs

alex-ju:
### :pushpin: Summary

As discussed during the engineering sync,  I'm uploading the notes taken during the kick-off session into our wiki.

### :link: External links

Jira ticket: [HDS-3316](https://hashicorp.atlassian.net/browse/HDS-3316)

***

### 👀 Component checklist

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-3316]: https://hashicorp.atlassian.net/browse/HDS-3316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2097, opened 2024-05-15
`Reveal` - Convert to TypeScript

didoo:
### :pushpin: Summary

Converts the `Reveal` component to TS

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2708

***

### 👀 Component checklist


packages/components/src/components/hds/reveal/index.ts
```diff
@@ -0,0 +1,53 @@
...
+    text: string,
+    textWhenOpen: string | undefined
+  ): string => {
+    if (isOpen && textWhenOpen !== undefined) {
+      return textWhenOpen;
+    } else {
+      if (text !== undefined) {
+        return text;
+      } else {
+        assert('@text for "Hds::Reveal" must have a valid value');
```
didoo, line 49: @alex-ju when do you think we should start to remove these kind of assertions, used as guards against the required arguments, for example, or arguments that should be one of a list? is this covered by TypeScript now (in fact TS is not complaining that I'm not returning a value at line 49/50) or there may be consumers that still benefit from these assertions?

alex-ju: yes, I was just reviewing the code – `text` can't be undefined (as per signature; not optional) and it should throw a type error in this case, making the assert unnecessary for TypeScript users, but it would still benefit non-TypeScript consumers, so I believe it's too early to remove them.

packages/components/src/components/hds/reveal/index.hbs
```diff
@@ -7,7 +7,7 @@
   <:toggle as |t|>
     <Hds::Reveal::Toggle::Button
       aria-controls={{this.contentId}}
-      @text={{if (and t.isOpen @textWhenOpen) @textWhenOpen this.text}}
+      @text={{this.getText t.isOpen @text @textWhenOpen}}
```
alex-ju, line 10: thought: could we simplify `this.getText` to get only the required argument `t.isOpen`? `@text` and `@textWhenOpen` can be retrieved via the `args` object in the backing class, right?

didoo: yeah, good point! 👍

packages/components/src/components/hds/reveal/index.ts
didoo: @alex-ju can you add this specific point (keep the existing assertions in place, because this will benefit non-TypeScript consumers) to [the documentation](https://github.com/hashicorp/design-system/pull/2087)? thanks

packages/components/src/components/hds/reveal/index.hbs
didoo: Done

alex-ju (approved):

PR #2099, opened 2024-05-15
`ButtonSet`:  Convert to TypeScript (HDS-2686)

KristinLBradley:
### :pushpin: Summary

If merged, this PR converts the ButtonSet component to TypeScript.

### :link: External links

- Jira ticket: [HDS-2686](https://hashicorp.atlassian.net/browse/HDS-2686)

***

### 👀 Component checklist

[HDS-2686]: https://hashicorp.atlassian.net/browse/HDS-2686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/button-set/index.ts
+++ b/packages/components/src/components/hds/button-set/index.ts
@@ -3,8 +3,11 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import templateOnlyComponent from '@ember/component/template-only';
+import Component from '@glimmer/component';
 
-const HdsButtonSetComponent = templateOnlyComponent();
+interface HdsButtonSetSignature {
+  Blocks: { default: [] };
+  Element: HTMLDivElement;
+}
 
-export default HdsButtonSetComponent;
+export default class HdsButtonSetComponent extends Component<HdsButtonSetSignature> {}
--- a/packages/components/src/components/hds/button-set/index.ts
+++ b/packages/components/src/components/hds/button-set/index.ts
@@ -1,0 +1,10 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import templateOnlyComponent from '@ember/component/template-only';
+
+const HdsButtonSetComponent = templateOnlyComponent();
+
+export default HdsButtonSetComponent;
--- a/packages/components/src/components/hds/button-set/index.ts
+++ b/packages/components/src/components/hds/button-set/index.ts
@@ -3,11 +3,11 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import Component from '@glimmer/component';
-
+import TemplateOnlyComponent from '@ember/component/template-only';
 interface HdsButtonSetSignature {
   Blocks: { default: [] };
   Element: HTMLDivElement;
 }
+const HdsButtonSetComponent = TemplateOnlyComponent<HdsButtonSetSignature>();
 
-export default class HdsButtonSetComponent extends Component<HdsButtonSetSignature> {}
+export default HdsButtonSetComponent;
--- a/.changeset/strange-hairs-judge.md
+++ b/.changeset/strange-hairs-judge.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`ButtonSet` - Converted component to TypeScript
--- a/packages/components/src/components/hds/button-set/index.hbs
+++ b/packages/components/src/components/hds/button-set/index.hbs
@@ -1,4 +1,3 @@
-{{! @glint-nocheck: not typesafe yet }}
 {{!
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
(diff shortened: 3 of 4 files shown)


didoo (approved):
👍

packages/components/src/components/hds/button-set/index.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import templateOnlyComponent from '@ember/component/template-only';
+
+const HdsButtonSetComponent = templateOnlyComponent();
```
natmegs, line 8: (feel free to take or leave) If we wanted to express to a consumer that this component takes yielded content, we could add something like
```
interface HdsButtonSetSignature {
  Blocks: { default: [] }
}
const HdsButtonSetComponent = templateOnlyComponent<HdsButtonSetSignature>()
```

KristinLBradley: @alex-ju Should I add this?

didoo: @natmegs I have the same doubt as @KristinLBradley because if we add a signature TS complains that `templateOnlyComponent` doesn't accept type arguments (`Expected 0 type arguments, but got 1`)
modified
so I am not sure the usefulness of a `template-only` declaration: in which _real_ use case would be used?

natmegs: Oh good to know! I was just going off the docs [here](https://typed-ember.gitbook.io/glint/environments/ember/template-only-components). If the types don't allow passing the component signature then IMO it's not super useful.

An alternative would be:
```
interface HdsButtonSetSignature {
  Blocks: { default: [] }
}
export default class HdsButtonSetComponent extends Component<HdsButtonSetSignature> {}
```

didoo: Yes, that's what I started to do too 👍

didoo: @aklkv do you know if the documentation linked by @natmegs is wrong? should we open an issue somewhere?

alex-ju: this is a known issue – if the repo contains a dependency on [@types/ember__component@v3](https://www.npmjs.com/package/@types/ember__component/v/3.16.6?activeTab=code) (and unfortunately we do, at the workspace level, via a third party dependency – can't remember which one, but I can look again). if you look under the FAQ here, [the second question](https://github.com/ijlee2/ember-codemod-args-to-signature/?tab=readme-ov-file#frequently-asked-questions) you'll see they suggest the same options you considered. In the existing code we took the approach @natmegs and @didoo mentioned.

I tried [resolving it](https://github.com/hashicorp/design-system/pull/2023/commits/9abb66b892b431f22fe932e0f391fd87dc1c77e0) a while back, but it didn't stick for some reason.

didoo: Should we document it in https://github.com/hashicorp/design-system/pull/2087 then?

didoo: @alex-ju also, would it work if we forced the version of `@types/ember__component` in the `resolutions` field of `package.json`?

alex-ju: Applied a fix to this – until we manage to get rid of `@types/ember__component@v3` from our dependencies.  @KristinLBradley after #2113 gets merged you can follow the docs/codemod/Natalie's suggestion on this

KristinLBradley: Thanks for the feedback everyone. I added a Signature following @natmegs' example above.

packages/components/src/components/hds/button-set/index.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+import Component from '@glimmer/component';
+
+interface HdsButtonSetSignature {
+  Blocks: { default: [] };
+  Element: HTMLDivElement;
+}
+
+export default class HdsButtonSetComponent extends Component<HdsButtonSetSignature> {}
```
KristinLBradley, line 13: Just to confirm, did I follow the correct patterns above? @alex-ju

alex-ju, line 13:
```suggestion
import TemplateOnlyComponent from '@ember/component/template-only';

interface HdsButtonSetSignature {
  Blocks: { default: [] };
  Element: HTMLDivElement;
}

const HdsButtonSetComponent = TemplateOnlyComponent<HdsButtonSetSignature>();

export default HdsButtonSetComponent;
```

alex-ju: no, not really. we would want to use `TemplateOnlyComponent` in this case, as we don't need a backing class.
here's [an example for reference](https://github.com/hashicorp/design-system/blob/f48aeeee8c25a986e7b8529160153c60f6521cc8/packages/components/src/components/hds/accordion/index.ts), I added a suggestion in a separate comment as well.

KristinLBradley: Thanks. Updated to fix

packages/components/src/components/hds/button-set/index.ts
KristinLBradley: Updated

PR #2100, opened 2024-05-15
`CopyButton`: convert to TypeScript (HDS-2688)

KristinLBradley:
### :pushpin: Summary

If merged, this PR converts the `CopyButton` to TypeScript.

The CopyButton depends upon:
* https://github.com/hashicorp/design-system/pull/2120

### :link: External links

* Jira ticket: [HDS-2688](https://hashicorp.atlassian.net/browse/HDS-2688)

***

### 👀 Component checklist

[HDS-2688]: https://hashicorp.atlassian.net/browse/HDS-2688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -15,9 +15,27 @@ export const SUCCESS_ICON = 'clipboard-checked';
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onError?: (...args: any[]) => void;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onSuccess?: (...args: any[]) => void;
+    size?: 'small' | 'medium'; // TODO: create new type?
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    targetToCopy?: string | ((...args: any[]) => void); // TODO: create new type?
+    text: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    textToCopy?: string | ((...args: any[]) => void); // TODO: create new type?
```
KristinLBradley, line 31: For the above TODOs should I be adding a `types.ts` file and define new types for them?

packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -15,9 +15,27 @@ export const SUCCESS_ICON = 'clipboard-checked';
...
+  };
+  Element: HTMLElement;
+}
+
+export default class HdsCopyButtonComponent extends Component<HdsCopyButtonSignature> {
   @tracked status = DEFAULT_STATUS;
-  @tracked timer;
+  @tracked timer = 0; // TODO: what value shoulf this be initialized to?
```
KristinLBradley, line 38:
```suggestion
  @tracked timer = 0; // TODO: what value should this be initialized to?
```

packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -15,9 +15,27 @@ export const SUCCESS_ICON = 'clipboard-checked';
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    textToCopy?: string | ((...args: any[]) => void); // TODO: create new type?
+  };
+  Element: HTMLElement;
+}
+
+export default class HdsCopyButtonComponent extends Component<HdsCopyButtonSignature> {
   @tracked status = DEFAULT_STATUS;
-  @tracked timer;
+  @tracked timer = 0; // TODO: what value should this be initialized to?
```
KristinLBradley, line 38: Is this ok? (Setting beginning value to `0`)

packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -70,23 +88,25 @@ export default class HdsCopyButtonComponent extends Component {
   }
 
   @action
-  onSuccess(args) {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  onSuccess(args: any) {
```
KristinLBradley, line 92: Should I be defining the args here and below using a specific type instead of `any`?

packages/components/src/components/hds/copy/button/index.ts
KristinLBradley: Also should I be using different types than `any`?

alex-ju: For consistency, yes I would create an enum for size and, indeed you can define a type that can be shared by targetToCopy and textToCopy (as I understand they both share the same potential structure)

packages/components/src/components/hds/copy/button/index.ts
alex-ju: We know `timer` is going to get assigned a value by the `setTimeout` function, so we can say that the type expected for this argument is whatever type `setTimeout` returns or `undefined` (for until it actually gets a value assigned to it)
```suggestion
  @tracked timer: ReturnType<typeof setTimeout> | undefined;
```

packages/components/src/components/hds/copy/button/index.ts
alex-ju: we should probably keep it generic as it can have any arguments I assume

alex-ju (commented):
Gave this a quick review – as you figured already, there is a bit of complexity on this one as it requires the `hds-clipboard` modifier to be converted as well.

KristinLBradley:
> Gave this a quick review – as you figured already, there is a bit of complexity on this one as it requires the `hds-clipboard` modifier to be converted as well.

@alex-ju Should I try to convert `hds-clipboard` as part of this component PR or should that be done separately?

alex-ju:
> > Gave this a quick review – as you figured already, there is a bit of complexity on this one as it requires the `hds-clipboard` modifier to be converted as well.
> 
> @alex-ju Should I try to convert `hds-clipboard` as part of this component PR or should that be done separately?

up to you – considering it's only in use by this component and `CopySnippet` we can consider doing it here too

packages/components/src/components/hds/copy/button/index.ts
KristinLBradley: Pushed up

KristinLBradley:
> > > Gave this a quick review – as you figured already, there is a bit of complexity on this one as it requires the `hds-clipboard` modifier to be converted as well.
> > 
> > 
> > @alex-ju Should I try to convert `hds-clipboard` as part of this component PR or should that be done separately?
> 
> up to you – considering it's only in use by this component and `CopySnippet` we can do it here too

@alex-ju I wen ahead and converted `hds-clipboard` as part of this.

packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -15,9 +16,25 @@ export const SUCCESS_ICON = 'clipboard-checked';
...
+    isFullWidth?: boolean;
+    isIconOnly?: boolean;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onError?: (...args: any[]) => void;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onSuccess?: (...args: any[]) => void;
+    size?: HdsCopyButtonSizes;
+    targetToCopy?: HdsCopyButtonItemToCopy;
+    text: string;
+    textToCopy?: HdsCopyButtonItemToCopy;
```
didoo, line 30: the type for this should come from the `hds-clipboard` signature (the property is forwarded to it)

packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -15,9 +16,25 @@ export const SUCCESS_ICON = 'clipboard-checked';
...
+  Args: {
+    isFullWidth?: boolean;
+    isIconOnly?: boolean;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onError?: (...args: any[]) => void;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onSuccess?: (...args: any[]) => void;
+    size?: HdsCopyButtonSizes;
+    targetToCopy?: HdsCopyButtonItemToCopy;
+    text: string;
```
didoo, line 29: the type for this should come from the `Button` signature (the property is forwarded to it)

packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -15,9 +16,25 @@ export const SUCCESS_ICON = 'clipboard-checked';
 export const ERROR_ICON = 'clipboard-x';
 export const DEFAULT_STATUS = 'idle';
 
-export default class HdsCopyButtonComponent extends Component {
+export interface HdsCopyButtonSignature {
+  Args: {
```
didoo, line 20: Probably you could `Pick` a few properties from `ButtonSignature` (`text`, `isFullWidth`, `isIconOnly`) and remove them from here.

packages/components/src/components/hds/copy/button/types.ts
```diff
@@ -0,0 +1,13 @@
...
+ */
+
+export enum HdsCopyButtonSizeValues {
+  Small = 'small',
+  Medium = 'medium',
+}
+export type HdsCopyButtonSizes = `${HdsCopyButtonSizeValues}`;
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type HdsCopyButtonItemToCopy = string | ((...args: any[]) => void);
```
didoo, line 13: As mentioned above, I don't think you need this, the type(s) should come from the signature of the `hds-clipboard` modifier

packages/components/src/modifiers/hds-clipboard.ts
didoo, line 1: No strong opinions, but consider having a separate PR for the modifier only.

packages/components/src/modifiers/hds-clipboard.ts
```diff
@@ -129,15 +136,35 @@ export const copyToClipboard = async (text, target) => {
 // because it's quite simple in its logic, and doesn't require injecting services
 // see: https://github.com/ember-modifier/ember-modifier#function-based-modifiers
 
-export default modifier((element, positional, named) => {
+export default modifier((element, _positional, named) => {
```
didoo, line 139: When porting the popover-related modifiers (https://github.com/hashicorp/design-system/pull/2103 and https://github.com/hashicorp/design-system/pull/2104) I've tried to follow a signature pattern that I've found on other codebases (eg. https://github.com/cardstack/cardstack/blob/dc2e0f9a42837feb073e3a3033b1a9188f50b4f7/packages/boxel/addon/modifiers/autoscroll.ts#L8-L14 or https://github.com/ember-bootstrap/ember-bootstrap/blob/6e9f791fd065a35e680cf26217682b7779ec19aa/ember-bootstrap/addon/modifiers/bs-conditional-attribute.ts#L3-L8 or https://github.com/adopted-ember-addons/ember-popper-modifier/blob/main/addon/-base-popper-modifier.ts#L76)

Maybe we can use a similar pattern here: see for example this https://github.com/hashicorp/design-system/blob/popover-primitive-ts-conversion/packages/components/src/modifiers/hds-register-event.ts

didoo: I've asked in the channel to confirm this: https://hashicorp.slack.com/archives/C06NMSYDZPA/p1715956057393179

packages/components/src/modifiers/hds-clipboard.ts
```diff
@@ -129,15 +136,35 @@ export const copyToClipboard = async (text, target) => {
...
   assert(
     '`hds-clipboard` modifier - the modifier must be applied to an element',
     element
   );
 
-  const { text, target, onSuccess, onError } = named;
+  const {
```
didoo, line 145: For readability, better to move these types to a standalone type declaration (in this way, it can also be imported by the consumers of the modifier)

packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -15,9 +16,25 @@ export const SUCCESS_ICON = 'clipboard-checked';
...
+    isFullWidth?: boolean;
+    isIconOnly?: boolean;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onError?: (...args: any[]) => void;
```
didoo, line 24: the type for this should come from the hds-clipboard signature (the property is forwarded to it)

packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -15,9 +16,25 @@ export const SUCCESS_ICON = 'clipboard-checked';
...
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onSuccess?: (...args: any[]) => void;
```
didoo, line 26: the type for this should come from the hds-clipboard signature (the property is forwarded to it)

packages/components/src/components/hds/copy/button/index.ts
didoo: Same for the properties from `hds-clipboard`

packages/components/src/template-registry.ts
```diff
@@ -23,9 +24,12 @@ import type HdsTagComponent from './components/hds/tag';
 import type HdsToastComponent from './components/hds/toast';
 import type HdsTextCodeComponent from './components/hds/text/code';
 import type HdsYieldComponent from './components/hds/yield';
+
 import type HdsLinkToModelsHelper from './helpers/hds-link-to-models';
 import type HdsLinkToQueryHelper from './helpers/hds-link-to-query';
 
+import type HdsClipboardModifier from './modifiers/hds-clipboard';
```
didoo, line 31: see: https://hashicorp.slack.com/archives/C06NMSYDZPA/p1715876987605319?thread_ts=1715872354.358549&cid=C06NMSYDZPA and Alexey's answer
```suggestion
import type HdsClipboardModifier from './modifiers/hds-clipboard.ts';
```

packages/components/src/modifiers/hds-clipboard.ts
```diff
@@ -129,15 +136,35 @@ export const copyToClipboard = async (text, target) => {
...
+    target: Node;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onSuccess: (...args: any[]) => void;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onError: (...args: any[]) => void;
+  };
 
-  const onClick = async (event) => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const onClick = async (event: { currentTarget: any }) => {
```
didoo, line 167: The reason why you see an error is because the `element` in the modifier (line 139) is not typed to be an `HTMLElement`
```suggestion
  const onClick = async (event: MouseEvent) => {
```

packages/components/src/modifiers/hds-clipboard.ts
```diff
@@ -6,8 +6,8 @@
 import { modifier } from 'ember-modifier';
 import { assert, warn } from '@ember/debug';
 
-export const getTextToCopy = (text) => {
-  let textToCopy;
+export const getTextToCopy = (text: { toString: () => string }) => {
```
didoo, line 9: Considered the logic below, in this function, I _think_ we should do something different here. 
@alex-ju what do you think?

packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -15,9 +20,25 @@ export const SUCCESS_ICON = 'clipboard-checked';
...
+    size?: HdsCopyButtonSizes;
+    targetToCopy?: HdsCopyButtonTargetToCopy;
+    text: string;
+    textToCopy?: HdsCopyButtonTextToCopy;
+  };
+  Element: HTMLElement;
```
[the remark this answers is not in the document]
alex-ju: That's correct – if we use another component as the main element it's safer to point to its signature

packages/components/src/modifiers/hds-clipboard.ts
KristinLBradley: I'm sorry, I should have done that from the beginning. If I try to do it now I'm worried it will create more problems because I'll lose review comments.

packages/components/src/components/hds/copy/button/index.ts
KristinLBradley: I'm having trouble figuring out how to import the properties from the `hds-clipboard`

didoo: Let's sync later

packages/components/src/modifiers/hds-clipboard.ts
didoo: I can add back my comments in the new PR, don't worry.

KristinLBradley: clipboard modifier changes have been split off into a separate branch: https://github.com/hashicorp/design-system/pull/2120

.changeset/popular-eels-melt.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`CopyButton` - Converted component to TypeScript
+
+`hds-clipboard` - Converted modifier to TypeScript
```
KristinLBradley, line 7:
```suggestion
```

.changeset/popular-eels-melt.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`CopyButton` - Converted component to TypeScript
+
```
KristinLBradley, line 6:
```suggestion
```

packages/components/src/components/hds/copy/button/types.ts
KristinLBradley: Removed

packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -7,17 +7,32 @@ import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
+import { HdsCopyButtonSizeValues } from './types.ts';
+import type { HdsCopyButtonSizes } from './types.ts';
+import type { HdsButtonSignature } from '../../button/';
+import type { ModifierSignature } from '../../../../modifiers/hds-clipboard.ts';
```
didoo, line 13: If you rename the signature to `HdsClipboardModifierSignature` in the other PR, remember to update also this import.

packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -7,17 +7,32 @@ import Component from '@glimmer/component';
...
 export const DEFAULT_ICON = 'clipboard-copy';
 export const SUCCESS_ICON = 'clipboard-checked';
 export const ERROR_ICON = 'clipboard-x';
 export const DEFAULT_STATUS = 'idle';
 
-export default class HdsCopyButtonComponent extends Component {
+export interface HdsCopyButtonSignature {
+  Args: HdsButtonSignature['Args'] & {
+    size?: HdsCopyButtonSizes;
+    textToCopy?: ModifierSignature['Args']['Named']['text'];
```
didoo, line 25: @alex-ju I suggested Kristin to do this. Do you agree with this approach? Other/better ideas?

packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -70,23 +85,25 @@ export default class HdsCopyButtonComponent extends Component {
   }
 
   @action
-  onSuccess(args) {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  onSuccess(args: any) {
```
didoo, line 89: @alex-ju would this be correct?
```suggestion
  onSuccess(args: ModifierSignature['Args']['Named']['onSuccess']) {
```

packages/components/src/components/hds/copy/button/index.ts
```diff
@@ -70,23 +85,25 @@ export default class HdsCopyButtonComponent extends Component {
...
 
     if (typeof onSuccess === 'function') {
       onSuccess(args);
     }
   }
 
   @action
-  onError(args) {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  onError(args: any) {
```
didoo, line 102: @alex-ju would this be correct?
```suggestion
  onError(args: ModifierSignature['Args']['Named']['onError']) {
```

didoo (commented):
@KristinLBradley left a few suggestions, but before approving I would wait for @alex-ju feedback on some of the suggested changes (I may be wrong)

packages/components/src/components/hds/copy/button/index.ts
alex-ju: considering the two (the component and the modifier) share a similar structure for the signature, yes I think this makes sense. alternatively, we could create a shared interface and update both `HdsCopyButtonSignature ` and `HdsClipboardModifierSignature` to use it, but to me they don't seem like siblings as the CopyButton component implements the modifier, so this format works well

packages/components/src/components/hds/copy/button/index.ts
alex-ju: yes, that would work too

packages/components/src/components/hds/copy/button/index.ts
alex-ju: yes, same as above

packages/components/src/template-registry.ts
```diff
@@ -151,6 +152,11 @@ export default interface HdsComponentsRegistry {
   'hds/card': typeof HdsCardContainerComponent;
   HdsCard: typeof HdsCardContainerComponent;
 
+  // Copy Button
+  'Hds::Copy::Button': typeof HdsCopyButtonComponent;
+  'hds/copy/button': typeof HdsCopyButtonComponent;
+  HdsCopyButton: typeof HdsCopyButtonComponent;
```
didoo, line 158: See: https://github.com/hashicorp/design-system/pull/2127
```suggestion
```

didoo (approved):
Left a suggestion, but apart from that all good 👍

KristinLBradley:
@alex-ju Does this one look ok to merge now?

PR #2101, opened 2024-05-16
`Accordion` - Convert to TypeScript

alex-ju:
### :pushpin: Summary

Converting Accordion (and subcomponents) to TypeScript

### :link: External links

Jira ticket: [HDS-2680](https://hashicorp.atlassian.net/browse/HDS-2680)

***

### 👀 Component checklist

[HDS-2680]: https://hashicorp.atlassian.net/browse/HDS-2680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/components/hds/accordion/index.ts
```diff
@@ -0,0 +1,16 @@
+import Component from '@glimmer/component';
+import type { ComponentLike } from '@glint/template';
+import type { HdsAccordionItemIndexSignature } from './item/index.ts';
+
+interface HdsAccordionIndexSignature {
+  Blocks: {
+    default: [
+      {
+        Item?: ComponentLike<HdsAccordionItemIndexSignature>;
```
didoo, line 9: this makes me thing that we should provide some guidance around the `Blocks` in the signature, in the shared document https://github.com/hashicorp/design-system/pull/2087; until now I haven't paid much attention to this entity, because I thought the `npx` codemod would take care of this, but now I see here is much more structured than the ones I have added in my PRs. Do you think is it worth a quick 101 on how to do `Blocks` signatures?

packages/components/src/template-registry.ts
```diff
@@ -3,6 +3,9 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
+import type HdsAccordionIndexComponent from './components/hds/accordion';
```
didoo, line 6: See comment here https://github.com/hashicorp/design-system/pull/2087/files#r1602930931

This tells me is better to decide sooner than later ;)

didoo (commented):
Left a couple of comments. Pre-approving in case you want to merge, but I think we should decide some standards around where/when use `Index` in the naming.

didoo (approved):
Didn't check the `Approve` toggle 🤦‍♂️

packages/components/src/components/hds/accordion/index.ts
alex-ju: It depends very much on whether there is structured content (e.g. subcomponents) within the blocks

packages/components/src/template-registry.ts
alex-ju: Indeed – let's align on this. As I mentioned on the other thread, we may need to align on class names as well

packages/components/src/template-registry.ts
```diff
@@ -27,6 +30,19 @@ import type HdsLinkToModelsHelper from './helpers/hds-link-to-models';
...
 
 export default interface HdsComponentsRegistry {
+  // Accordion
+  'Hds::Accordion': typeof HdsAccordionIndexComponent;
+  'hds/accordion': typeof HdsAccordionIndexComponent;
+  HdsAccordion: typeof HdsAccordionIndexComponent;
+
+  'Hds::Accordion::Item': typeof HdsAccordionItemButtonComponent;
+  'hds/accordion/item': typeof HdsAccordionItemButtonComponent;
+  HdsAccordionItem: typeof HdsAccordionItemIndexComponent;
```
didoo, line 40: @alex-ju the second side of the declarations should be `ItemIndex`, no?
```suggestion
  'Hds::Accordion::Item': typeof HdsAccordionItemIndexComponent;
  'hds/accordion/item': typeof HdsAccordionItemIndexComponent;
  HdsAccordionItem: typeof HdsAccordionItemIndexComponent;
```

didoo: (btw, this `template-registry` process is **so** error prone! I have to double check everything twice, because is so much easy to make a typo and not notice it)

alex-ju: good catch! yeah, ideally we would have this automated

alex-ju: Addressed in [3775a5a](https://github.com/hashicorp/design-system/pull/2101/commits/3775a5adc609b74d069e056da20ede9f42fafbe8)

packages/components/src/template-registry.ts
alex-ju: Addressed in [3775a5a](https://github.com/hashicorp/design-system/pull/2101/commits/3775a5adc609b74d069e056da20ede9f42fafbe8) as discussed (removed `Index`)

packages/components/src/components/hds/accordion/index.ts
didoo: We've discussed today, so I am closing this

alex-ju:
Rebased after #2113 to use TemplateOnlyComponent

PR #2102, opened 2024-05-16
`Sidenav` - Convert to TypeScript

didoo:
### :pushpin: Summary

Converts the `SideNav` component and subcomponents to TypeScript.

Notice: because of failing tests for `ember-lts-3.28` it's not possible at the moment to upgrade `ember-stargate` package to `0.5.0`, so we had to add loose typings for `Portal/PortalTarget`.

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2711

***

### 👀 Component checklist


alex-ju:
@didoo I had another look at this PR to understand the remaining bits. `SideNav::Base` is a very generic component, so I don't see a way around it other than deferring type checking to runtime (so, defining the named blocks as `any`). Let me know if you find this acceptable so I can do an overall review.

packages/components/src/components/hds/side-nav/base.ts
```diff
@@ -0,0 +1,19 @@
+import TemplateOnlyComponent from '@ember/component/template-only';
+
+export interface HdsSideNavBaseSignature {
+  Blocks: {
+    root?: [];
```
didoo, line 5: @alex-ju looking at the `base.hbs` file, there's no difference between `root` and `header/body/footer`. Should this one too be `any`?

packages/components/src/components/hds/side-nav/index.ts
```diff
@@ -9,20 +9,64 @@ import { action } from '@ember/object';
...
+    a11yRefocusNavigationText?: string;
+    a11yRefocusRouteChangeValidator?: string;
+    ariaLabel?: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onToggleMinimizedStatus?: (arg: boolean) => void;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onDesktopViewportChange?: (arg: boolean) => void;
+  };
+  Blocks: {
+    root?: [];
```
didoo, line 31: @alex-ju I just noticed that we don't expose a `:root` block, we just consume it from the "base"
```suggestion
```

didoo:
> @didoo I had another look at this PR to understand the remaining bits and pushed two commits. `SideNav::Base` is a very generic component, so I don't see a way around it other than deferring type checking to runtime (so, defining the named blocks as `any`). If you find this acceptable I can do an overall review.

@alex-ju I've added a few comments, to continue the conversation. question: what do you think if we ask (again) if someone has different/better ideas in the #[#proj-hds-ts-migration](https://hashicorp.enterprise.slack.com/archives/C06NMSYDZPA) and/or [#tech-modern-ember](https://hashicorp.enterprise.slack.com/archives/C05P1FMLP6U) channels?

> update: ~seems to be failing on ember-lts-3.28, maybe a rebase will help 🤞~ looking at the error it may have to do with the `ember-stargate` bump, although it was meant to work with 3.28

what do you think we should do? @meirish is back from PTO next week, should we wait? or try to debug and in case open an issue in the `ember-stargate` repo?

(I've also rebased, just in case)

packages/components/src/components/hds/side-nav/base.ts
alex-ju: in `base.hbs` yes, there's no difference – I forgot about it because when `SideNav::Base` is used in `index.hbs` it didn't have a hash (so there was no need to make it generic). We can make it `any`, for consistency.

packages/components/src/components/hds/side-nav/index.ts
```diff
@@ -9,20 +9,64 @@ import { action } from '@ember/object';
...
+        Body?: [];
+        isMinimized?: boolean;
+      }
+    ];
+    footer?: [
+      {
+        Footer?: [];
+        isMinimized?: boolean;
+      }
+    ];
```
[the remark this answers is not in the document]
alex-ju: yes, we can use the `HdsSideNavBaseSignature` signature

alex-ju:
> > @didoo I had another look at this PR to understand the remaining bits and pushed two commits. `SideNav::Base` is a very generic component, so I don't see a way around it other than deferring type checking to runtime (so, defining the named blocks as `any`). If you find this acceptable I can do an overall review.
> 
> @alex-ju I've added a few comments, to continue the conversation. question: what do you think if we ask (again) if someone has different/better ideas in the #[#proj-hds-ts-migration](https://hashicorp.enterprise.slack.com/archives/C06NMSYDZPA) and/or [#tech-modern-ember](https://hashicorp.enterprise.slack.com/archives/C05P1FMLP6U) channels?

I'm totally up to ask around – I'm sure there are ways to make the Types stricter, but wanted to have a fallback to make sure we're not blocked.

didoo:
> wanted to have a fallback to make sure we're not blocked.

@alex-ju see my last two commits, let's see if they unblock the tests, and then we decide how to proceed

packages/components/src/components/hds/side-nav/index.ts
```diff
@@ -9,20 +9,63 @@ import { action } from '@ember/object';
...
+    footer?: [
+      {
+        Footer?: HdsSideNavBaseSignature['Blocks']['footer'];
+        isMinimized?: boolean;
+      }
+    ];
+  };
+  Element: HdsSideNavBaseSignature['Element'];
```
MelSumner, line 50: question: is there documentation somewhere that demonstrates how/why/when to do this kind of element definition specifically? I'm asking because I thought that this was constrained to the existing/valid HTML Element list, so I'm curious to learn more.

didoo: @alex-ju can you keep me honest on the above? 👆

MelSumner (approved):
I don't think I see anything amiss. I agree we should figure out the stargate thing but we could also save that for after we're able to upgrade, maybe.

packages/components/src/components/hds/side-nav/index.ts
alex-ju: what Cristiano said is accurate; to help cascading/propagating types we oftentimes use the signature of the referenced component

alex-ju (approved):

PR #2103, opened 2024-05-16
Modifier `hds-register-event` - Convert to TypeScript

didoo:
### :pushpin: Summary

Converts the `hds-register-event` modifier to TypeScript

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-3337

***

### 👀 Component checklist


packages/components/src/template-registry.ts
```diff
@@ -130,7 +134,17 @@ export default interface HdsComponentsRegistry {
...
   'hds-link-to-models': typeof HdsLinkToModelsHelper;
+
+  // hds-link-to-query
   'hds-link-to-query': typeof HdsLinkToQueryHelper;
+
+  // ----- MODIFIERS ----------------------------------------------------
+
+  // hds-register-event
+  'hds-register-event': typeof HdsRegisterEventModifier;
+  registerEvent: typeof HdsRegisterEventModifier;
```
didoo, line 149: @alex-ju is this the "right"/expected format for the modifiers? /cc @aklkv @natmegs

aklkv: kebab-case should do just fine, this is how ember resolves them and this is how helpers/modifiers should be used in order for ember to resolve it non-stricked env

didoo: So, when they're invoked in a backing class like [here](https://github.com/hashicorp/design-system/blob/popover-primitive-ts-conversion/packages/components/src/components/hds/popover-primitive/index.ts#L14-L15) it's not necessary to have a counterpart in the `template-registry.ts` file. Am I understanding correctly?

alex-ju: yes, that's correct, kebab case should suffice.

since we're using a default export when imported the name can be whatever we want

```
import whateverMakesSenseInTheContext from '../../../modifiers/hds-register-event.ts';
```

alex-ju (approved):
Looking good! I believe this is our first modifier being converted 🙌

PR #2104, opened 2024-05-16
Modifier `hds-anchored-position` - Convert to TypeScript

didoo:
### :pushpin: Summary

Converts the `hds-anchored-position` modifier to TypeScript

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-3338

***

### 👀 Component checklist


packages/components/src/modifiers/hds-anchored-position.ts
```diff
@@ -0,0 +1,274 @@
...
+  OffsetOptions,
+  FlipOptions,
+  ShiftOptions,
+  AutoPlacementOptions,
+  ArrowOptions,
+  Middleware,
+} from '@floating-ui/dom';
+
+export const DEFAULT_PLACEMENT = 'bottom';
+export const PLACEMENTS: Placement[] = [
```
didoo, line 44: @alex-ju can you think of a way to get this list directly from the `@floating-ui` package(s)? I tried but I couldn't see how, but maybe you (or @aklkv) have better ideas?

alex-ju: if the `Placement` type (imported on L23) was an enum, we could've gotten these values using `Object.values(Placement)`, but it seems to be a union type (of another type and some string literals) so I don't see any way to get the values.

packages/components/src/modifiers/hds-anchored-position.ts
```diff
@@ -0,0 +1,274 @@
...
+      arrowElement = named.arrowElement;
+    } else if (named.arrowSelector) {
+      assert(
+        '`hds-anchored-position` modifier - the `selector` provided for the "arrow" element must be a string',
+        typeof named.arrowSelector === 'string'
+      );
+
+      const selectedArrowElement = document.querySelector(named.arrowSelector);
+      if (selectedArrowElement) {
+        arrowElement = selectedArrowElement as HTMLElement;
```
alex-ju, line 208: We can avoid the type assertion here by stating the expected type at declaration
```suggestion
      const selectedArrowElement: HTMLElement | null = document.querySelector(
        named.arrowSelector
      );
      if (selectedArrowElement) {
        arrowElement = selectedArrowElement;
```

alex-ju: or we can use a type guard in the conditional statement, however makes more sense to you
```js
if (selectedArrowElement instanceof HTMLElement) {
```

alex-ju (approved):
Looks good! Left only one suggestion.

Is this branch meant to be pointed to `main`? I don't see the register-event modifier depending on it

packages/components/src/modifiers/hds-anchored-position.ts
didoo: I like the second one, thank you for the suggestion!

didoo:
> Is this branch meant to be pointed to `main`? I don't see the register-event modifier depending on it

@alex-ju there's only a tiny dependency, but also it helps me in remembering in which order to the rebase (there are four PRs related)

didoo:
@alex-ju in the context of https://github.com/hashicorp/design-system/pull/2205 I've added this commit, can you check it's OK? thanks https://github.com/hashicorp/design-system/pull/2104/commits/eb122da536531a1a5cb9853990c8f55ce43b6cd4

PR #2105, opened 2024-05-16
`PopoverPrimitive` - Convert to TypeScript

didoo:
### :pushpin: Summary

Converts `PopoverPrimitive` to TypeScript. 

Depends on #2103 and #2104.

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-3339

### 🙏 Credits

Thanks @aklkv for all the help getting this PR over the line, it was not an easy one (context: https://github.com/ember-modifier/ember-modifier/issues/774#issuecomment-2145471993 and https://github.com/aklkv/modifier-like-repro and https://github.com/ember-modifier/ember-modifier/pull/855 and https://discord.com/channels/480462759797063690/483601670685720591/1256103260943552512)

***

### 👀 Component checklist


packages/components/src/components/hds/popover-primitive/index.ts
```diff
@@ -32,16 +32,64 @@ export {
   DEFAULT_PLACEMENT,
 } from '../../../modifiers/hds-anchored-position.ts';
 
-export default class HdsPopoverPrimitiveComponent extends Component {
+import type { FloatingUIOptions } from '../../../modifiers/hds-anchored-position.ts';
+import type { ModifierLike } from '@glint/template';
+
+export interface HdsPopoverPrimitiveSignature {
+  Args: {
+    isOpen: boolean;
```
aklkv, line 40: we probably want to mark some of this as optional

aklkv (approved):
Left a small comment otherwise looks good 👍🏻

packages/components/src/components/hds/popover-primitive/index.ts
didoo: you're right, fixed in https://github.com/hashicorp/design-system/pull/2105/commits/931642fc2ee9ec2ea17fd1087dd0541d90ef3472

didoo:
For context and future reference: after @aklkv has pushed the patch for the `ember-modifier` signature, @NullVoxPopuli commented [in this GitHub issue](https://github.com/ember-modifier/ember-modifier/issues/774#issuecomment-2198126063) and [in this Discord thread](https://discord.com/channels/480462759797063690/491905849405472769/1256578444981043240). 

So @aklkv force pushed a commit to undo the changes he made, and remove the `{  eager: true }` from our code. To preserve the history, I preferred to revert the changes and have distinct commits for each operation (the final code remains the same)

packages/components/src/components/hds/popover-primitive/index.ts
```diff
@@ -57,36 +99,42 @@ export default class HdsPopoverPrimitiveComponent extends Component {
...
-  setupPrimitiveContainer = modifier(
-    (element) => {
+  setupPrimitiveContainer = modifier<SetupPrimitiveContainerModifier>(
+    (element: HTMLElement) => {
       this.containerElement = element;
 
       // we register the "soft" events
       if (this.enableSoftEvents) {
+        // @ts-expect-error: known issue with type of invocation
         registerEvent(this.containerElement, ['mouseenter', this.onMouseEnter]);
```
alex-ju, line 109: Last time I checked out this PR these were the only errors returned by TypeScript linter. I take it we couldn't find a way around it, right?

alex-ju (approved):
I believe the main conversion issue of this PR still stands, but I'm ok with ignoring them for now to move things forward.

packages/components/src/components/hds/popover-primitive/index.ts
didoo: Yes, this is something I wanted to ask @aklkv too. With the patch you created, this error went away, but removing the patch this is still something TS complains about. Can you expand on why this is?

aklkv: The patch only tried addressing missing options for `eager` and as we learned it's no longer needed and we don't need this patch or need the `eager` option. When it comes to the comment above I have asked the question on ember Discord https://discord.com/channels/480462759797063690/1256103260943552512/1256103588430614559
so this is legal use case but not totally covered by community nor it has piping in place. So I would keep an eye on this and it might get better TS support as I mentioned we might need something like `invokeHelper`

PR #2106, opened 2024-05-16
`RichTooltip` - Convert to TypeScript

didoo:
### :pushpin: Summary

Converts `RichTooltip` to TypeScript. 

Depends on #2103, #2104, and #2105.

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-3340

***

### 👀 Component checklist


packages/components/src/components/hds/popover-primitive/index.ts
```diff
@@ -261,15 +294,17 @@ export default class HdsPopoverPrimitiveComponent extends Component {
...
-      !this.containerElement.contains(
-        event.relatedTarget || document.activeElement
-      )
-    ) {
-      this.hidePopover();
+    if (this.containerElement) {
+      // TODO!
+      const relatedTarget: EventTarget | null = event.relatedTarget;
+      const activeElement: Node | null = document.activeElement;
+      const xxx = relatedTarget || activeElement;
```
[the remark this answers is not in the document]
didoo: This PR is still in draft, waiting for #2105 to be "resolved" (changes in that PR would impact this PR too)

packages/components/src/components/hds/rich-tooltip/bubble.ts
```diff
@@ -6,12 +6,32 @@
...
 
-export default class HdsRichTooltipBubbleComponent extends Component {
+export interface HdsRichTooltipBubbleSignature {
+  Args: {
+    placement?: FloatingUIOptions['placement'];
+    offset?: FloatingUIOptions['offsetOptions'];
+    enableCollisionDetection?: FloatingUIOptions['enableCollisionDetection'];
+    width?: string;
+    height?: string;
+    isOpen?: boolean;
```
zamoore, line 23: Looks like the docs are missing some arguments

packages/components/src/components/hds/rich-tooltip/toggle.ts
```diff
@@ -6,11 +6,42 @@
...
+  HdsRichTooltipToggleIconPositionValues
+);
+export const DEFAULT_ICONPOSITION =
+  HdsRichTooltipToggleIconPositionValues.Trailing;
+export const SIZES: string[] = Object.values(HdsRichTooltipToggleSizeValues);
+
+export interface HdsRichTooltipToggleSignature {
+  Args: {
+    text?: string;
+    icon?: FlightIconSignature['Args']['name'];
```
zamoore, line 30: I hadn't considered coupling the types like this, but I think it makes sense.

packages/components/src/components/hds/rich-tooltip/bubble.ts
didoo: Oh, strange! Let me have a look 👀

didoo: Ah yes, now I see: in some cases, when we don't want consumers to use certain argument for the yielded sub-components (because are [provided directly by the parent component](https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/rich-tooltip/index.hbs#L25-L28)) we omit them from the API documentation

packages/components/src/components/hds/rich-tooltip/toggle.ts
didoo: Yes, for me that's the power of TS in a design system, the fact that you don't have to worry about changing types across components, when they are connected in this way you can be sure that a change gets automatically "propagated" to the consumers without having to remember to do it manually

packages/components/src/components/hds/rich-tooltip/bubble.ts
zamoore: Ah, ok. That makes sense. Thanks for clarifying.

packages/components/src/components/hds/rich-tooltip/toggle.ts
zamoore: Makes sense!

zamoore (approved):

aklkv (approved):

alex-ju (approved):
Generally the code looks good. I've not tested it in the browser thinking I will give it a final functional test in the `popover-primitive-ts-conversion` branch

packages/components/src/components/hds/rich-tooltip/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import Component from '@glimmer/component';
+import { getElementId } from '../../../utils/hds-get-element-id.js';
+import type { WithBoundArgs } from '@glint/template';
+import type { HdsPopoverPrimitiveSignature } from '../popover-primitive';
+import HdsRichTooltipToggle from './toggle.ts';
+import HdsRichTooltipBubble from './bubble.ts';
+
+interface HdsRichTooltipSignature {
+  Args: Omit<HdsPopoverPrimitiveSignature['Args'], 'enableSoftEvents'>;
```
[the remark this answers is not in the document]
didoo: thanks

packages/components/src/components/hds/rich-tooltip/bubble.ts
```diff
@@ -6,12 +6,32 @@
...
+    popoverId: string;
+    arrowId: string;
+    setupPrimitivePopover: HdsPopoverPrimitiveSignature['Blocks']['default'][0]['setupPrimitivePopover'];
```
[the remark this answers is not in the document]
didoo: No, I totally agree with you, I just wasn't sure which one of the two approaches was better.
In terms of traceability, the fact that there is a reference to the `SetupPrimitivePopoverModifier` I think is enough in this case. I'll make the changes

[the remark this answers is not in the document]
didoo: See: https://github.com/hashicorp/design-system/pull/2106/commits/029d40332055c5c931df002f51260f8793698b63

didoo:
@alex-ju in the context of https://github.com/hashicorp/design-system/pull/2205 I've added this commit, can you check it's OK? thanks https://github.com/hashicorp/design-system/pull/2106/commits/01ac7fab76bd18216fedc175ecd991bc0f98c9f7

PR #2108, opened 2024-05-16
Standardize type declarations for existing components

didoo:
### :pushpin: Summary

Per discussion and agreement today, I've reviewed components already converted to TypeScript to follow as much as possible the patterns we have agreed upon.

In this PR I have:

`Hds::Interactive` - Aligned type declaration to agreed patterns

- Updated class name (removed `Index` from the name)

`Hds::Button` - Aligned type declaration to agreed patterns

- Updated class name (removed `Index` from the name)
- Updated “Component API” documentation
- Sorted signature arguments to match documentation

`Hds::Alert` - Aligned type declaration to agreed patterns

- Updated class name (removed `Index` from the name)

`Hds::Badge` - Aligned type declaration to agreed patterns

- Updated class name (removed `Index` from the name)
- Updated “Component API” documentation to be consistent with `Button` and `BadgeCount`
- Sorted signature arguments to match documentation

`Hds::BadgeCount` - Aligned type declaration to agreed patterns

- Updated class and signature names (removed `Index` from the names)
- Updated “Component API” documentation to match signature and be consistent with `Button` and `Badge`

`Hds::Card::Container` - Aligned type declaration to agreed patterns

- Sorted signature arguments to match documentation

`Hds:DisclosurePrimitive` - Updated documentation

- Added missing `isOpen` argument to the "Component API" section

`Hds::DismissButton` - Aligned type declaration to agreed patterns

- Updated class name (removed `Index` from the name)

`Hds::IconTile` - Aligned type declaration to agreed patterns

- Updated class name (removed `Index` from the name)
- Updated “Component API” documentation
- Sorted signature arguments to match documentation

`Hds::Link::Inline` - Aligned type declaration to agreed patterns

- Remove extra args (they are already provided by the `Interactive` signature)

`Hds::Link::Standalone` - Aligned type declaration to agreed patterns

- Updated “Component API” documentation to be consistent with `Link::Inline` and `Badge/BadgeCount`
- Sorted signature arguments to match documentation
- Remove extra args (they are already provided by the `Interactive` signature)

`Hds::Tag` - Aligned type declaration to agreed patterns

- Updated class and signature names (removed `Index` from the names)
- Updated “Component API” documentation to match signature and be consistent with `Button`, `Badge`, etc.

`Hds::Text` - Aligned type declaration to agreed patterns for text components

- Updated class and signature names (removed `Index` from the names)
- Sorted signature arguments to match documentation

***

### 👀 Component checklist

### DIFF
--- a/.changeset/lemon-masks-carry.md
+++ b/.changeset/lemon-masks-carry.md
@@ -1,0 +1,6 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Alert`, `Badge`, `BadgeCount`, `Button`, `Card::Container`, `DisclosurePrimitive`, `DismissButton`, `IconTile`, `Interactive`, `Link::Inline`, `Link::Standalone`, `Tag`, `Text` - Standardized class names and signatures
+
--- a/packages/components/src/components/hds/alert/index.ts
+++ b/packages/components/src/components/hds/alert/index.ts
@@ -60,7 +60,7 @@ export interface HdsAlertSignature {
   Element: HTMLDivElement;
 }
 
-export default class HdsAlertIndexComponent extends Component<HdsAlertSignature> {
+export default class HdsAlertComponent extends Component<HdsAlertSignature> {
   @tracked role = 'alert';
   @tracked ariaLabelledBy?: string;
 
--- a/packages/components/src/components/hds/alert/index.ts
+++ b/packages/components/src/components/hds/alert/index.ts
@@ -12,7 +12,7 @@ import { tracked } from '@glimmer/tracking';
 import { HdsAlertColorValues, HdsAlertTypeValues } from './types.ts';
 
 import type { ComponentLike, WithBoundArgs } from '@glint/template';
-import type HdsButtonIndexComponent from '../button';
+import type HdsButtonComponent from '../button';
 import type HdsLinkStandaloneComponent from '../link/standalone';
 import type { HdsYieldSignature } from '../yield';
 import type { HdsAlertColors, HdsAlertTypes } from './types.ts';
@@ -53,7 +53,7 @@ export interface HdsAlertSignature {
           typeof HdsLinkStandaloneComponent,
           'size'
         >;
-        Button?: WithBoundArgs<typeof HdsButtonIndexComponent, 'size'>;
+        Button?: WithBoundArgs<typeof HdsButtonComponent, 'size'>;
       }
     ];
   };
--- a/packages/components/src/components/hds/badge-count/index.ts
+++ b/packages/components/src/components/hds/badge-count/index.ts
@@ -24,17 +24,17 @@ export const DEFAULT_SIZE = HdsBadgeCountSizeValues.Medium;
 export const DEFAULT_TYPE = HdsBadgeCountTypeValues.Filled;
 export const DEFAULT_COLOR = HdsBadgeCountColorValues.Neutral;
 
-interface HdsBadgeCountIndexSignature {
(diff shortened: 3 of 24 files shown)


.changeset/lemon-masks-carry.md
```diff
@@ -0,0 +1,66 @@
+---
+"@hashicorp/design-system-components": minor
```
alex-ju, line 2: I see these as fixes, not features. In my view a changelog entry should be a summary, not a log, and should be focused on changes that might affect our consumers (should not include website changes or internal renamings or sorting that don't affect the API)

What about: `Alert`, `Button`, `BadgeCount`, ... ,`Tag`, `Text` – standardize class names and signatures

```suggestion
"@hashicorp/design-system-components": patch
```

alex-ju (commented):
I see that in this PR we change both the argument order in signatures (from alphabetical to match the order we currently have in the website) and the order of the arguments in the website, which makes the order consistent but arbitrary. If we are to use this pattern going forward would you be able to document/articulate the rules for sorting arguments so we can have it as a reference?

.changeset/lemon-masks-carry.md
didoo: OK, I've updated the changeset 👍

didoo:
> I see that in this PR we change both the argument order in signatures (from alphabetical to match the order we currently have in the website) and the order of the arguments in the website, which makes the order consistent but arbitrary. If we are to use this pattern going forward would you be able to document/articulate the rules for sorting arguments so we can have it as a reference?

@alex-ju as you say, it's somehow arbitrary. What I've tried to do is to use other similar components as reference, and follow the same (or at least a similar) order. For example, I've noticed that `Button`, `Badge/BadgeCount`, `IconTile`, `LinkInline/Standalone`, `Tag` they all share some common props, like `type`, `size`, `color`, `text`, `textSomething`, `icon`, `iconSomething` so I tried to keep this order; most of the components have the `...attributes` at the very bottom; for the links I have tried to keep the props coming from the `Interactive` all in the same block and order. I can write this down somewhere, where do you suggest to do it?

alex-ju (approved):

alex-ju:
> > I see that in this PR we change both the argument order in signatures (from alphabetical to match the order we currently have in the website) and the order of the arguments in the website, which makes the order consistent but arbitrary. If we are to use this pattern going forward would you be able to document/articulate the rules for sorting arguments so we can have it as a reference?
> 
> @alex-ju as you say, it's somehow arbitrary. What I've tried to do is to use other similar components as reference, and follow the same (or at least a similar) order. For example, I've noticed that `Button`, `Badge/BadgeCount`, `IconTile`, `LinkInline/Standalone`, `Tag` they all share some common props, like `type`, `size`, `color`, `text`, `textSomething`, `icon`, `iconSomething` so I tried to keep this order; most of the components have the `...attributes` at the very bottom; for the links I have tried to keep the props coming from the `Interactive` all in the same block and order. I can write this down somewhere, where do you suggest to do it?

not something for this PR, more of a thing for our backlog to revisit

didoo:
> not something for this PR, more of a thing for our backlog to revisit

would be OK to add something to the https://github.com/hashicorp/design-system/pull/2087 (I can add a suggestion) for the time being, then when the migration is complete we can review all the components-related documentation (there will be new engineering onboardings to do, it will be a work worth to do in advance)

didoo:
> yes, that sounds good!

https://github.com/hashicorp/design-system/pull/2087/files#r1604785808

PR #2112, opened 2024-05-17
Add back babel plugins to transform private methods

alex-ju:
### :pushpin: Summary

After the refactor in #2004 the HDS code fails in `atlas` with the following error
Screenshot 2024-05-17 at 11 43 28

Restoring [the babel plugins we initially used](https://github.com/hashicorp/design-system/pull/2004/files#diff-949397ffa49ee9acf2f8cf58a09becfe5c27e2bc3673eaf251bdbbf4a07d1538L6-L8) seems to do the trick – so I suggest we do this for now, to unblock the next release.

***

### 👀 Component checklist

### DIFF
--- a/packages/components/babel.config.json
+++ b/packages/components/babel.config.json
@@ -23,6 +23,9 @@
           "import": "decorator-transforms/runtime"
         }
       }
-    ]
+    ],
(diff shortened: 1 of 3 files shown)


didoo (approved):
Didn't see associated reviewers, so approved just in case.

PR #2126, opened 2024-05-22
feat(ember-flight-icons): convert to typescript & publish types

aklkv:
### :pushpin: Summary

The main goal of this change is to make sure types are created and published for the Ember Flight Icon component

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-3315

***

### 👀 Component checklist


packages/ember-flight-icons/src/components.ts
```diff
@@ -0,0 +1,3 @@
+import FlightIcon from './components/flight-icon.gts';
+
+export { FlightIcon };
```
aklkv, line 3: this gives consumers the ability to import components like so:

```js
import { FlightIcon } from '@hashicorp/ember-flight-icons/components'
```

packages/components/src/components/hds/app-footer/link.ts
```diff
@@ -8,11 +8,12 @@ import Component from '@glimmer/component';
 import type { HdsInteractiveSignature } from '../interactive/';
 import type { HdsLinkColors, HdsLinkIconPositions } from '../link/types.ts';
 import type { HdsLinkInlineSignature } from '../link/inline.ts';
+import type { IconName } from '@hashicorp/flight-icons/svg/index';
```
didoo, line 11: Question: should we instead import the signature from the `EmberFlightIcon` component, and then use its `FlightIconSignature['Args']['name']?

.changeset/long-rats-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/ember-flight-icons": patch
+---
+
+convert to `.gts` and publish types
```
didoo, line 5: This needs to be updated

packages/components/package.json
```diff
@@ -20,17 +20,20 @@
...
+    "lint:hbs": "ember-template-lint . --no-error-on-unmatched-pattern",
+    "lint:hbs:fix": "ember-template-lint . --fix --no-error-on-unmatched-pattern",
     "lint:js": "eslint . --cache",
-    "lint:js:fix": "eslint . --fix"
+    "lint:js:fix": "eslint . --fix",
+    "lint:types": "glint",
+    "prepublishOnly": "yarn workspace @hashicorp/ember-flight-icons build && yarn build && test -f 'dist/styles/@hashicorp/design-system-components.css' || (echo 'The pre-compiled CSS file was not found' && exit 1)",
+    "start": "rollup --config --watch --environment development",
+    "start:js": "rollup --config --watch --no-watch.clearScreen",
+    "start:types": "glint --declaration --watch",
```
didoo, line 35: In the other `package.json` files we have only the `start` command. Are the other two needed? What are they used for? Doesn't the `start` command do the same (only a bit more)?
```suggestion
    "start": "rollup --config --watch --environment development",
```

packages/components/unpublished-development-types/@hashicorp/ember-flight-icons/flight-icon.d.ts
didoo, line 1: Question: I see that this file has been deleted, but could it have been an option to just adde the `IconName` signature here, without converting the `EmberFlightIcon` component to TS? In that case, can you follow this minimum path with the PR? As mentioned by @alex-ju, we're planning to create a brand new component `Hds::Icon` within the `components` package (relatively soon) so it would be better to have the simplest PR possible here, much easier to review and approve.

didoo (commented):
@aklkv left a few comments

packages/components/package.json
aklkv: I guess it just a convenience command that got just introduced in the latest blueprint, happy drop this changes is there is no value

aklkv: [components](https://github.com/hashicorp/design-system/blob/a3c3f3da5abea4fecdce9d0d190562aa9241664c/packages/components/package.json#L34) packages has similar commands but they are bypassed for some reason, happy to drop this change if you see no value

packages/ember-flight-icons/babel.config.json
```diff
@@ -1,9 +1,28 @@
...
+      {
+        "allExtensions": true,
+        "onlyRemoveTypeImports": true,
+        "allowDeclareFields": true
+      }
+    ],
     "@embroider/addon-dev/template-colocation-plugin",
-    ["@babel/plugin-transform-typescript", { "allowDeclareFields": true }],
-    ["@babel/plugin-proposal-decorators", { "legacy": true }],
-    "@babel/plugin-transform-class-properties"
```
alex-ju, line 7: unfortunately, we can't remove these – they are still needed for older codebases (such as `atlas`) to work. I ended up re-adding them in `components`.

packages/ember-flight-icons/package.json
```diff
@@ -17,64 +17,96 @@
...
+    "prepack": "concurrently 'yarn:build:*'",
+    "start": "concurrently 'yarn:start:*'",
+    "start:js": "rollup --config --watch --no-watch.clearScreen",
+    "start:types": "glint --declaration --watch",
+    "test": "echo 'A v2 addon does not have tests, run tests in test-app'"
   },
   "dependencies": {
     "@embroider/addon-shim": "^1.8.7",
     "@hashicorp/flight-icons": "^3.2.0",
+    "decorator-transforms": "^1.0.1",
```
alex-ju, line 39: I believe we use a more recent version of `decorator-transforms` in `components`; we can try to align them

alex-ju (commented):
[CI doesn't seem to be happy with the new setup](https://github.com/hashicorp/design-system/actions/runs/9350118677/job/25732918986?pr=2126) (although it appears to be building fine locally) – would you be keen to investigate it?

alex-ju:
> @alex-ju I definitely fix that, I might be a little slower since I am traveling for the next couple of weeks

no rush on our side – safe travels! ✈️

packages/ember-flight-icons/src/components/flight-icon.hbs
```diff
@@ -1,7 +1,3 @@
-{{!
```
didoo, line 1: I think this header needs to be preserved

packages/ember-flight-icons/src/components/flight-icon.ts
```diff
@@ -6,17 +6,49 @@
...
-export default class FlightIconComponent extends Component {
-  constructor() {
-    super(...arguments);
+import type { IconName } from '@hashicorp/flight-icons/svg';
+
+export interface FlightIconSignature {
+  Args: {
+    /**
+     * The color of the icon. Defaults to `currentColor`.
+     */
```
didoo, line 17: Currently we don't have inline comments to describe the arguments in the other signatures. Do you mind removing them? We may decide later to do it systematically across all the codebase, but for now I prefer to have consistency in how we declare the signatures, to avoid having a jungle of formats. Thank you

packages/ember-flight-icons/src/components/flight-icon.ts
```diff
@@ -6,17 +6,49 @@
...
+    color?: string;
+    /**
+     * The name of the icon to display.
+     */
+    name?: IconName;
```
didoo, line 22: Should this be required? 🤔

packages/ember-flight-icons/src/components/flight-icon.ts
```diff
@@ -6,17 +6,49 @@
...
+export default class FlightIcon extends Component<FlightIconSignature> {
+  constructor(owner: unknown, args: FlightIconSignature['Args']) {
+    super(owner, args);
+
     if (!this.args.name) {
       assert('Please provide to <FlightIcon> a value for @name');
     } else if (!iconNames.includes(this.args.name)) {
       assert(
-        `The icon @name "${this.args.name}" provided to <FlightIcon> is not correct. Please verify it exists on https://helios.hashicorp.design/icons/library`
+        `The icon @name "${this.args.name}" provided to <FlightIcon> is not correct. Please verify it exists on https://helios.hashicorp.design/icons/library`,
```
didoo: why this extra comma at the end of the line?
modified

packages/ember-flight-icons/src/components/flight-icon.ts
```diff
@@ -6,17 +6,49 @@
...
-
 import { iconNames } from '@hashicorp/flight-icons/svg';
 
-export default class FlightIconComponent extends Component {
-  constructor() {
-    super(...arguments);
+import type { IconName } from '@hashicorp/flight-icons/svg';
+
+export interface FlightIconSignature {
+  Args: {
```
didoo, line 14: We [tend to follow](https://github.com/hashicorp/design-system/blob/main/wiki/TypeScript-Migration.md#typescript-migration) the same order as in the documentation, so it's easy to review and compare:
https://helios.hashicorp.design/icons/usage-guidelines?tab=code#component-api

packages/ember-flight-icons/package.json
```diff
@@ -17,64 +17,96 @@
...
   "author": "HashiCorp Design Systems <design-systems@hashicorp.com>",
   "scripts": {
-    "build": "rollup --config",
-    "prepublishOnly": "rollup --config",
-    "start": "rollup --config --watch --environment development",
-    "lint": "npm-run-all --aggregate-output --continue-on-error --parallel \"lint:!(fix)\"",
-    "lint:fix": "npm-run-all --aggregate-output --continue-on-error --parallel lint:*:fix",
-    "lint:hbs": "ember-template-lint .",
-    "lint:hbs:fix": "ember-template-lint . --fix",
+    "build": "concurrently 'yarn:build:*'",
```
didoo: Should these changes go in https://github.com/hashicorp/design-system/pull/2142 ?

packages/ember-flight-icons/package.json
```diff
@@ -17,64 +17,96 @@
...
     "version": 2,
+    "type": "addon",
+    "main": "addon-main.cjs",
     "app-js": {
+      "./components.js": "./dist/_app_/components.js",
       "./components/flight-icon.js": "./dist/_app_/components/flight-icon.js",
-      "./instance-initializers/load-sprite.js": "./dist/_app_/instance-initializers/load-sprite.js"
+      "./instance-initializers/load-sprite.js": "./dist/_app_/instance-initializers/load-sprite.js",
+      "./template-registry.js": "./dist/_app_/template-registry.js",
+      "./flight-icon-CZreFL2Q.js": "./dist/_app_/flight-icon-CZreFL2Q.js"
```
didoo, line 86: what is this? 🤔

didoo (commented):
Left a few comments (after rebasing on https://github.com/hashicorp/design-system/pull/2142/)

packages/components/package.json
didoo: @aklkv we discussed that these changes would be removed or moved to https://github.com/hashicorp/design-system/pull/2142. Did I misunderstand?

packages/components/unpublished-development-types/@hashicorp/ember-flight-icons/flight-icon.d.ts
didoo: Discussed in person with @aklkv and clarified that just changing this file would not emit the types (in fact, the file lives under the `unpublished-development-types` folder)

packages/ember-flight-icons/src/components/flight-icon.ts
aklkv: in the newer prettier version dangling comma is a default now

packages/ember-flight-icons/package.json
aklkv: missconfugured rollup

packages/ember-flight-icons/src/components/flight-icon.ts
didoo: left a comment about it. I would prefer if updates like this are done systematically across the monorepo, to avoid that different parts behave in different ways.

packages/ember-flight-icons/src/components/flight-icon.ts
aklkv: I have tried that but, several consuming components are not happy about this change, I could potentially look into fixing the consuming side but since `Hds::Icon` is coming pretty soon I would say we have an assert segment that would catch potential errors, this should be enough till the new `Hds::Icon` is shipped

didoo: OK, makes sense. 👍

didoo (approved):
I _think_ is OK, but let's wait for @alex-ju to approve too before merging

.changeset/long-rats-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/ember-flight-icons": minor
+---
+
+`FlightIcon` component - Convert to typescript & publish types
```
alex-ju, line 5: nit (we have some very specific guidance about [consistent terminology](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#consistent-terminology) for the changelog)
```suggestion
`FlightIcon` - converted component to TypeScript and published types
```

alex-ju (approved):
Looking great! Thanks for addressing all the feedback

PR #2127, opened 2024-05-22
Template Registry - Remove redundant entries

didoo:
### :pushpin: Summary

While discussing with @aklkv he told explained me that there are redundant entries in the `template-registry.ts` file, in particular all the ones that have this syntax: `HdsXyz: typeof HdsXyzComponent`. The reason is that the `template-registry.ts` is used only to determine types/signatures of components used in the **_templates_** (HBS) and this means there's no use for such pattern (`HdsXyz`). In the `gjs/gts` files the components are imported so their signature are explicit.

In this PR I have:
- removed redundant entries in the template registry

***

### 👀 Component checklist

### DIFF
--- a/.changeset/lucky-stingrays-call.md
+++ b/.changeset/lucky-stingrays-call.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Template Registry - Removed redundant entries
--- a/packages/components/src/template-registry.ts
+++ b/packages/components/src/template-registry.ts
@@ -51,185 +51,144 @@ export default interface HdsComponentsRegistry {
   // Accordion
   'Hds::Accordion': typeof HdsAccordionComponent;
   'hds/accordion': typeof HdsAccordionComponent;
-  HdsAccordion: typeof HdsAccordionComponent;
 
   'Hds::Accordion::Item': typeof HdsAccordionItemComponent;
   'hds/accordion/item': typeof HdsAccordionItemComponent;
-  HdsAccordionItem: typeof HdsAccordionItemComponent;
 
   'Hds::Accordion::Item::Button': typeof HdsAccordionItemButtonComponent;
   'hds/accordion/item/button': typeof HdsAccordionItemButtonComponent;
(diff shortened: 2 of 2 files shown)


alex-ju (commented):
We introduced these signatures [at HCPs request](https://hashicorp.slack.com/archives/C7KTUHNUS/p1699295276662279) 😕

didoo:
> We introduced these signatures [at HCP's request](https://hashicorp.slack.com/archives/C7KTUHNUS/p1699295276662279) 😕

@aklkv I've mentioned you in the [slack conversation](https://hashicorp.slack.com/archives/C7KTUHNUS/p1716368123018029?thread_ts=1699295276.662279&cid=C7KTUHNUS) but for future reference, can you also answer here: are these PascalCase entry actually needed or not?

aklkv:
> > We introduced these signatures [at HCP's request](https://hashicorp.slack.com/archives/C7KTUHNUS/p1699295276662279) 😕
> 
> @aklkv I've mentioned you in the [slack conversation](https://hashicorp.slack.com/archives/C7KTUHNUS/p1716368123018029?thread_ts=1699295276.662279&cid=C7KTUHNUS) but for future reference, can you also answer here: are these PascalCase entry actually needed or not?

Certainly, I think I might have prompted this request 6 months ago with out having proper knowledge of how things work, I am hoping today I better equipped, but correct me if I am wrong.
first of template-registry  file is only needed for consumer that are looking to add glint  but still want to use .hbs  file, with that in mind there are only two ways we can call components from template:

```js
'Hds::Alert': typeof HdsAlertIndexComponent;
'hds/alert': typeof HdsAlertIndexComponent;
```

to summarize template-registry needs following formats:

```js
 export default interface HdsComponentsRegistry { 
  // Components
  'My::Component': typeof MyComponent,
  'my/component': typeof MyComponent,

  // Helpers
  'my-helper': typeof MyHelper,

  // Modifiers
  'my-modifier': typeof MyModifier,
}
```

didoo:
@aklkv thank you so much for the clarification!

alex-ju (approved):

PR #2136, opened 2024-05-31
`Tabs` documentation - Update to the output of the `logClickedTab` function to demo the arguments provided

didoo:
### :pushpin: Summary

Context: https://hashicorp.slack.com/archives/C7KTUHNUS/p1717015038457629

The intent is to make it more explicit in the demo that the callback receives two arguments, `event` and `index`.
### 👀 Component checklist


alex-ju (approved):
Makes sense – easier to point consumers to an example

PR #2142, opened 2024-06-03
feat(ember-flight-icons): update blueprint

aklkv:
### :pushpin: Summary

Preparatory PR for https://github.com/hashicorp/design-system/pull/2170

(context: https://hashicorp.slack.com/archives/C06NMSYDZPA/p1717440951790419?thread_ts=1717099148.439879&cid=C06NMSYDZPA)

***

### 👀 Component checklist


packages/ember-flight-icons/src/components/flight-icon.js
```diff
@@ -16,7 +16,7 @@ export default class FlightIconComponent extends Component {
       assert('Please provide to <FlightIcon> a value for @name');
     } else if (!iconNames.includes(this.args.name)) {
       assert(
-        `The icon @name "${this.args.name}" provided to <FlightIcon> is not correct. Please verify it exists on https://helios.hashicorp.design/icons/library`
+        `The icon @name "${this.args.name}" provided to <FlightIcon> is not correct. Please verify it exists on https://helios.hashicorp.design/icons/library`,
```
didoo, line 19: I've left a comment in the other PR about this change: https://github.com/hashicorp/design-system/pull/2126#discussion_r1625652230

didoo (commented):
I've had a quick look, but my knowledge about these settings is _very_ limited, so I'll defer to @alex-ju to do a in depth review.

packages/ember-flight-icons/package.json
```diff
@@ -17,64 +17,94 @@
...
     "eslint-plugin-ember": "^11.11.1",
-    "eslint-plugin-node": "^11.1.0",
-    "eslint-plugin-prettier": "^4.2.1",
+    "eslint-plugin-import": "^2.29.1",
+    "eslint-plugin-n": "^16.4.0",
+    "eslint-plugin-prettier": "^5.0.1",
     "eslint-plugin-qunit": "^8.0.1",
-    "npm-run-all": "^4.1.5",
-    "prettier": "^2.8.8",
+    "prettier": "^3.1.1",
```
didoo, line 68: @aklkv all the other `package.json` files are pointing to `^2.8.8`, can we keep them consistent to avoid having different behaviours in different parts of the monorepo? it may be confusing.

```suggestion
    "prettier": "^2.8.8",
```

packages/ember-flight-icons/.prettierrc.cjs
```diff
@@ -1,19 +1,17 @@
-'use strict';
+"use strict";
 
 module.exports = {
-  singleQuote: true,
```
didoo, line 4: @aklkv just to double check: now that we've reverted prettier to `2.8.8` is this change still OK?

Dhaulagiri:
With an eye toward some consistency across the monorepo I excised most of the blueprint-related changes here and got those in https://github.com/hashicorp/design-system/pull/2166. We should be able to rebase this one to get it down to just the TS-related changes and more cleanly review and release that.

didoo (approved):
I _think_ is OK, but please wait an approval from @alex-ju or @Dhaulagiri before merging

packages/ember-flight-icons/package.json
```diff
@@ -17,65 +17,97 @@
...
-    "main": "addon-main.cjs",
-    "type": "addon",
     "version": 2,
+    "type": "addon",
+    "main": "addon-main.cjs",
     "app-js": {
       "./components/flight-icon.js": "./dist/_app_/components/flight-icon.js",
-      "./instance-initializers/load-sprite.js": "./dist/_app_/instance-initializers/load-sprite.js"
+      "./instance-initializers/load-sprite.js": "./dist/_app_/instance-initializers/load-sprite.js",
+      "./template-registry.js": "./dist/_app_/template-registry.js"
```
alex-ju, line 87: now that `template-registry.js` is [a public entrypoint](https://github.com/hashicorp/design-system/pull/2142/files#diff-d6d1229bd915324e84ed0d1ead12d2665b51fb992054efe46249e5988546907bR19) I believe we don't need the alias here anymore.

update: running `yarn build` seems to be removing this line, so we should probably update as such

alex-ju (approved):
Looking good overall! It took a bit more than expected as I wanted to make sure I test with `atlas` as well

PR #2147, opened 2024-06-07
`Accordion` - expand/collapse enhancements

alex-ju:
### :pushpin: Summary

In this PR we propose a set of enhancements to our Accordion component to accommodate the following use cases without introducing any breaking changes:

**Allow consumers to implement an "expand all"/"collapse all" functionality** ([request reference](https://hashicorp.slack.com/archives/C7KTUHNUS/p1691525412883139))
   - we considered embedding the button inside the component but decided that would be too restrictive for our consumers
   - the flexible implementation also allows consumers to control the state (`open`/`close`) of each accordion item from outside the component

Example of usage
  ```hbs
<button type="button" {{on "click" this.toggleState}}>
    {{if (eq this.state "open") "Collapse all" "Expand all"}}
</button>

<Hds::Accordion @forceState={{this.state}} as |A|>
    {{! accordion items }}
</Hds::Accordion>
```

**Expose a function to close an accordion item from within its content** ([request reference](https://hashicorp.slack.com/archives/C7KTUHNUS/p1697824296621389?thread_ts=1691525412.883139&cid=C7KTUHNUS))

Example of usage
```hbs
<Hds::Accordion::Item>
  <:toggle>Item one</:toggle>
  <:content as |c|>
     <button type="button" {{on "click" c.close}}>Close</button>
  </:content>
</Hds::Accordion::Item>
```

 - Updated the `DisclosurePrimitive` to allow external control of the `isOpen` state. We do this by introducing two private properties (`_isOpen`, representing the internal state and `_isControlled` set to `true` every time the `@isOpen` argument is changed externally). Tried to follow the [concepts previously explored](https://hashicorp.atlassian.net/jira/polaris/projects/HDSP/ideas/view/3955253?selectedIssue=HDSP-23&focusedCommentId=435444).
 - Updated the `Accordion::Item` and `Accordion` adding a `@forceState` argument (`open`/`close`) to control the state from outside the component after the initial render.
- Updated the `Accordion::Item` component to forward the `close` action from `DisclosurePrimitive`.

### :link: External links

Jira ticket: [HDS-3364](https://hashicorp.atlassian.net/browse/HDS-3364)

***

### 👀 Component checklist

- [x] Make the type adjustments in the relevant `types.ts` files
- [x] Extend test suites to cover the new functionality
- [x] Test/validate the solution with consumers

[HDS-3364]: https://hashicorp.atlassian.net/browse/HDS-3364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/accordion/index.hbs
+++ b/packages/components/src/components/hds/accordion/index.hbs
@@ -4,5 +4,5 @@
 }}
 
 <div class={{this.classNames}} ...attributes>
-  {{yield (hash Item=(component "hds/accordion/item" size=this.size type=this.type))}}
+  {{yield (hash Item=(component "hds/accordion/item" size=this.size type=this.type forceState=@forceState))}}
 </div>
--- a/packages/components/src/components/hds/accordion/index.ts
+++ b/packages/components/src/components/hds/accordion/index.ts
@@ -10,13 +10,13 @@ import { SIZES, DEFAULT_SIZE, TYPES, DEFAULT_TYPE } from './item/index.ts';
 
 import type { ComponentLike } from '@glint/template';
 import type { HdsAccordionItemSignature } from './item/index.ts';
-import type { HdsAccordionSizes, HdsAccordionTypes } from './types.ts';
+import type { HdsAccordionForceStates, HdsAccordionSizes, HdsAccordionTypes } from './types.ts';
 
 interface HdsAccordionSignature {
   Args: {
     size?: HdsAccordionSizes;
     type?: HdsAccordionTypes;
-    forceState?: 'open' | 'close';
+    forceState?: HdsAccordionForceStates;
   };
   Blocks: {
     default: [
--- a/packages/components/src/components/hds/accordion/index.ts
+++ b/packages/components/src/components/hds/accordion/index.ts
@@ -16,6 +16,7 @@ interface HdsAccordionSignature {
   Args: {
     size?: HdsAccordionSizes;
     type?: HdsAccordionTypes;
+    forceState?: 'open' | 'close';
   };
   Blocks: {
     default: [
--- a/packages/components/src/components/hds/accordion/index.ts
+++ b/packages/components/src/components/hds/accordion/index.ts
@@ -40,7 +40,7 @@ export default class HdsAccordionComponent extends Component<HdsAccordionSignatu
    * @type {HdsAccordionSizes}
    * @default 'medium'
    */
-  get size() {
+  get size(): HdsAccordionSizes {
     const { size = DEFAULT_SIZE } = this.args;
 
     assert(
@@ -60,7 +60,7 @@ export default class HdsAccordionComponent extends Component<HdsAccordionSignatu
    * @type {HdsAccordionTypes}
    * @default 'card'
    */
-  get type() {
+  get type(): HdsAccordionTypes {
     const { type = DEFAULT_TYPE } = this.args;
 
     assert(
@@ -78,7 +78,7 @@ export default class HdsAccordionComponent extends Component<HdsAccordionSignatu
    * @method classNames
    * @return {string} The "class" attribute to apply to the component.
    */
-  get classNames() {
+  get classNames(): string {
     const classes = ['hds-accordion'];
 
     // add a class based on the @size argument
--- a/packages/components/src/components/hds/accordion/index.ts
+++ b/packages/components/src/components/hds/accordion/index.ts
@@ -10,7 +10,11 @@ import { SIZES, DEFAULT_SIZE, TYPES, DEFAULT_TYPE } from './item/index.ts';
 
 import type { ComponentLike } from '@glint/template';
 import type { HdsAccordionItemSignature } from './item/index.ts';
-import type { HdsAccordionForceStates, HdsAccordionSizes, HdsAccordionTypes } from './types.ts';
+import type {
+  HdsAccordionForceStates,
+  HdsAccordionSizes,
+  HdsAccordionTypes,
+} from './types.ts';
 
 interface HdsAccordionSignature {
   Args: {
--- a/packages/components/src/components/hds/accordion/item/index.hbs
+++ b/packages/components/src/components/hds/accordion/item/index.hbs
@@ -3,7 +3,7 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-<Hds::DisclosurePrimitive class={{this.classNames}} @isOpen={{@isOpen}} ...attributes>
+<Hds::DisclosurePrimitive class={{this.classNames}} @isOpen={{(or @isOpen (eq @forceState "open"))}} ...attributes>
   <:toggle as |t|>
     <div class="hds-accordion-item__toggle">
       <Hds::Accordion::Item::Button
--- a/packages/components/src/components/hds/accordion/item/index.hbs
+++ b/packages/components/src/components/hds/accordion/item/index.hbs
@@ -27,7 +27,7 @@
     </div>
   </:toggle>
 
-  <:content>
+  <:content as |c|>
     <Hds::Text::Body
       class="hds-accordion-item__content"
       @tag="div"
@@ -36,7 +36,7 @@
       @color="primary"
       id={{this.contentId}}
     >
-      {{yield to="content"}}
+      {{yield (hash close=c.close) to="content"}}
     </Hds::Text::Body>
   </:content>
 </Hds::DisclosurePrimitive>
--- a/packages/components/src/components/hds/accordion/item/index.hbs
+++ b/packages/components/src/components/hds/accordion/item/index.hbs
@@ -3,7 +3,12 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-<Hds::DisclosurePrimitive class={{this.classNames}} @isOpen={{(or @isOpen (eq @forceState "open"))}} @onClickToggle={{@onClickToggle}} ...attributes>
+<Hds::DisclosurePrimitive
+  class={{this.classNames}}
+  @isOpen={{(or @isOpen (eq @forceState "open"))}}
+  @onClickToggle={{@onClickToggle}}
+  ...attributes
+>
   <:toggle as |t|>
     <div class="hds-accordion-item__toggle">
       <Hds::Accordion::Item::Button
--- a/packages/components/src/components/hds/accordion/item/index.hbs
+++ b/packages/components/src/components/hds/accordion/item/index.hbs
@@ -3,7 +3,7 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-<Hds::DisclosurePrimitive class={{this.classNames}} @isOpen={{(or @isOpen (eq @forceState "open"))}} ...attributes>
+<Hds::DisclosurePrimitive class={{this.classNames}} @isOpen={{(or @isOpen (eq @forceState "open"))}} @onClickToggle={{@onClickToggle}} ...attributes>
   <:toggle as |t|>
     <div class="hds-accordion-item__toggle">
       <Hds::Accordion::Item::Button
--- a/packages/components/src/components/hds/accordion/item/index.ts
+++ b/packages/components/src/components/hds/accordion/item/index.ts
@@ -31,6 +31,8 @@ export interface HdsAccordionItemSignature {
     size?: HdsAccordionSizes;
     type?: HdsAccordionTypes;
     forceState?: HdsAccordionForceStates;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    onClickToggle?: (event: MouseEvent, ...args: any[]) => void;
   };
   Blocks: {
     toggle?: [];
--- a/packages/components/src/components/hds/accordion/item/index.ts
+++ b/packages/components/src/components/hds/accordion/item/index.ts
@@ -81,7 +81,7 @@ export default class HdsAccordionItemComponent extends Component<HdsAccordionIte
    * @type {HdsTextSizes}
    * @default 'medium'
    */
-  get toggleTextSize() {
+  get toggleTextSize(): number {
     const size = this.args.size ?? DEFAULT_SIZE;
     return TEXT_SIZE_MAP[size];
   }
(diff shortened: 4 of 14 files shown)


packages/components/src/components/hds/accordion/item/index.ts
```diff
@@ -11,10 +11,16 @@ export interface HdsAccordionItemSignature {
...
     isOpen?: boolean;
+    forceState?: 'open' | 'close';
   };
   Blocks: {
-    content?: [];
     toggle?: [];
+    content: [
+      {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        close: (...args: any[]) => void;
```
zamoore, line 21: Could this be `unknown` instead of `any`?

packages/components/src/components/hds/disclosure-primitive/index.ts
```diff
@@ -33,11 +33,38 @@ export interface HdsDisclosurePrimitiveSignature {
...
+  }
 
   @action
   onClickToggle() {
     this.isOpen = !this.isOpen;
+    this._isControlled = false;
+  }
+
+  @action
+  onStateChange() {
```
zamoore, line 63: It looks like you're passing `@isOpen` from the template as well.

packages/components/src/components/hds/disclosure-primitive/index.ts
```diff
@@ -33,11 +33,38 @@ export interface HdsDisclosurePrimitiveSignature {
...
 
 export default class HdsDisclosurePrimitiveComponent extends Component<HdsDisclosurePrimitiveSignature> {
-  @tracked isOpen = this.args.isOpen ?? false;
+  @tracked _isOpen = false;
+  @tracked _isControlled = this.args.isOpen != undefined;
+
+  get isOpen() {
+    if (this._isControlled) {
+      // if the state is controlled from outside, the argument overrides the internal state
+      return this.args.isOpen !== undefined ? this.args.isOpen : this._isOpen;
```
zamoore, line 42: I think this could be shortened to:

`return this.args.isOpen ?? this._isOpen`

zamoore (commented):
Given the requirements, this approach makes sense to me.

packages/components/src/components/hds/accordion/item/index.ts
KristinLBradley: Oh interesting (`unknown` vs. `any`). Was looking at this: https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type

Seems like a better choice if it works for us.

packages/components/src/components/hds/disclosure-primitive/index.ts
```diff
@@ -33,11 +33,38 @@ export interface HdsDisclosurePrimitiveSignature {
 }
 
 export default class HdsDisclosurePrimitiveComponent extends Component<HdsDisclosurePrimitiveSignature> {
-  @tracked isOpen = this.args.isOpen ?? false;
+  @tracked _isOpen = false;
+  @tracked _isControlled = this.args.isOpen != undefined;
```
didoo, line 37: when I implemented the other controlled/uncontrolled components I remember it was not possible to 100% determine the internal and external states at the same time, using this "trick": at times they would go out of sync (and I know that in the `Tabs` is implemented this way.

see this video for example
https://github.com/hashicorp/design-system/assets/686239/e46f6a31-c2ef-4c65-9abc-84547c26145c

if you look at this code for example: https://github.com/prysmex/ember-eui/blob/master/packages/core/src/components/eui-accordion.gts you will see how they have two distinct variables, `isInitiallyOpen` and `isOpen`, I suspect exactly for this reason.

happy to chat more and see if I can help with my (fainted) memory of how I implemented the code time ago (or maybe @zamoore has better knowledge/ideas)

didoo (commented):
@alex-ju added a comment that may reveal to be a blocker, have a looko

packages/components/src/components/hds/disclosure-primitive/index.ts
alex-ju: looking at the referenced code, conceptually `@initialIsOpen` is our `@isOpen` (avoided changing it as it would be a breaking change) and their `_opened` is our `_isOpen`. thanks for sharing that video, I seem to have missed a usacase – I'll look a bit more to understand how they get out of sync.

packages/components/src/components/hds/accordion/index.ts
```diff
@@ -8,6 +8,9 @@ import type { ComponentLike } from '@glint/template';
 import type { HdsAccordionItemSignature } from './item/index.ts';
 
 interface HdsAccordionSignature {
+  Args: {
+    forceState?: 'open' | 'close';
```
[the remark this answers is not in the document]
alex-ju: I will eventually pull this type in `types.ts` and reference it from there

packages/components/src/components/hds/accordion/item/index.ts
alex-ju: `unknown` might be a better fit although, for consistency with other functions, I would be tempted to leave it as `any` and review them holistically.

packages/components/src/components/hds/disclosure-primitive/index.ts
alex-ju: Yes, we can simplify this by using nullish coalescing. Updated as suggested.

packages/components/src/components/hds/accordion/index.ts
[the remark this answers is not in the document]
alex-ju: Extracted this to `type.ts` in [50d2ddd](https://github.com/hashicorp/design-system/pull/2147/commits/50d2ddd67c0b4fa1e3c538f28832632e284c9dde)

packages/components/src/components/hds/accordion/index.hbs
```diff
@@ -4,5 +4,5 @@
 }}
 
 <div class="hds-accordion" ...attributes>
-  {{yield (hash Item=(component "hds/accordion/item"))}}
+  {{yield (hash Item=(component "hds/accordion/item" forceState=@forceState))}}
```
didoo, line 7: playing advocate's devil: should this new argument called something like `forceIsOpen` (or `forceOpen`) and be a boolean? 🤔

packages/components/src/components/hds/accordion/item/index.hbs
```diff
@@ -35,7 +35,7 @@
       @color="primary"
       id={{this.contentId}}
     >
-      {{yield to="content"}}
+      {{yield (hash close=c.close) to="content"}}
```
didoo, line 38: just a consideration: the asymmetry here, that we have a `close` and not an `open` as well, does it mean something? are we missing something? 🤔

packages/components/src/components/hds/accordion/index.hbs
didoo: another consideration: I was imagining that the forced state would be at the item level, not at the parent accordion level (the `isOpen` is at item level). why is it necessary at this level? one can achieve the same "collapse all/expand all" by passing the `@forceState` argument to all the items, plus (in that case) it's possible to control a single item independently from the others (while with the current implementation it's not possible)

alex-ju: my understanding is that our consumers want a way to manage the states of all accordion items (not individual items), hence the approach of setting it at the root and cascading it to items.

packages/components/src/components/hds/accordion/item/index.hbs
alex-ju: if the content is `open` there's no need to provide a way to open the accordion; conversely, if the content is closed there is no element to perform an `open` function

didoo: ah yes, this is yielded to the `::content` 🤦‍♂️

packages/components/src/components/hds/accordion/index.hbs
alex-ju: updated with the suggested adjustments, adding `onClickToggle` to `DisclosurePrimitive` and forwarding it to `Accordion::Item`

packages/components/src/components/hds/disclosure-primitive/index.ts
```diff
@@ -33,11 +35,44 @@ export interface HdsDisclosurePrimitiveSignature {
...
+      this.args.onClickToggle &&
+      typeof this.args.onClickToggle === 'function'
+    ) {
+      this.args.onClickToggle(this.isOpen);
+    }
+  }
+
+  @action
+  onStateChange() {
+    if (this.args.isOpen !== undefined) {
```
didoo, line 72: if I understand correctly, we don't use `this._isControlled` here because this function is invoked when the value of `@isOpen` changes, which triggers the `did-update` which is run without re-rendering the component, so line 39 is not re-evaluated

if that is the case, what do you think of adding a comment to explain this "detail"? 😀 (it took me a while to get it)

if not, then maybe we can use `this._isControlled` as condition? would make sense (in which case TS complains, but we can cast the assignment via `Boolean(this.args.isOpen)` like we do at line 49

showcase/app/templates/components/accordion.hbs
```diff
@@ -244,6 +244,61 @@
...
+          <:toggle>Item Three</:toggle>
+          <:content>
+            <Shw::Placeholder @text="generic content" @height="40" />
+          </:content>
+        </A.Item>
+      </Hds::Accordion>
+    </SG.Item>
+  </Shw::Grid>
+
   <Shw::Text::H3>Edge cases</Shw::Text::H3>
```
didoo, line 302:
```suggestion
  <Shw::Divider @level={{2}} />
  
  <Shw::Text::H3>Edge cases</Shw::Text::H3>
```

didoo (commented):
Left a few more comments

emailnitram:
I tested this PR via the deployed showcase link https://hds-showcase-git-alex-ju-accordion-expand-all-hashicorp.vercel.app/components/accordion#externally-controlled and https://hds-showcase-git-alex-ju-accordion-expand-all-hashicorp.vercel.app/utilities/disclosure-primitive and the functionality and API looks good 👍

packages/components/src/components/hds/accordion/index.hbs
didoo: OK, I understand.

showcase/tests/integration/components/hds/accordion/index-test.js
```diff
@@ -143,4 +143,86 @@ module('Integration | Component | hds/accordion/index', function (hooks) {
...
+    this.set(
+      'onClickToggle',
+      () => (state = state === 'open' ? (state = 'close') : (state = 'open'))
+    );
+    await render(hbs`
+      <Hds::Accordion::Item @forceState={{this.state}} @onClickToggle={{this.onClickToggle}}>
+        <:toggle>Item one</:toggle>
+        <:content as |c|>
+          <button type="button" {{on "click" c.close}}>Close</button>
+        </:content>
```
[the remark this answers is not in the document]
alex-ju: updated as suggested

showcase/tests/integration/components/hds/accordion/index-test.js
```diff
@@ -143,4 +143,86 @@ module('Integration | Component | hds/accordion/index', function (hooks) {
...
+          </A.Item>
+          <A.Item @isOpen={{false}}>
+            <:toggle>Item one</:toggle>
+            <:content>Content two</:content>
+          </A.Item>
+        </Hds::Accordion>
+      `
+    );
+    // first item open at rendering
+    assert.dom('.hds-accordion-item__content').exists({ count: 1 });
```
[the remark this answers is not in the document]
alex-ju: updated to check it's the content of the first item

packages/components/src/components/hds/disclosure-primitive/index.ts
```diff
@@ -33,11 +35,44 @@ export interface HdsDisclosurePrimitiveSignature {
...
+  @tracked _isOpen = false;
+  @tracked _isControlled = this.args.isOpen != undefined;
+
+  get isOpen() {
+    if (this._isControlled) {
+      // if the state is controlled from outside, the argument overrides the internal state
+      return this.args.isOpen ?? this._isOpen;
+    } else {
+      // if the state is changes internally, the internal state overrides the argument
+      return this._isOpen !== undefined
```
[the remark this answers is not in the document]
alex-ju: now that we have a default this conditional check is not needed anymore; removed it

showcase/app/templates/components/accordion.hbs
alex-ju: updated as suggested

didoo (approved):
I am going to approve just in case, because for me it's OK. Up to you to decide when to merge, given that you're waiting for feedback from other engineers outside the team.

packages/components/src/components/hds/disclosure-primitive/index.ts
```diff
@@ -33,11 +35,42 @@ export interface HdsDisclosurePrimitiveSignature {
 }
 
 export default class HdsDisclosurePrimitiveComponent extends Component<HdsDisclosurePrimitiveSignature> {
-  @tracked isOpen = this.args.isOpen ?? false;
+  @tracked _isOpen = false;
+  @tracked _isControlled = this.args.isOpen !== undefined;
```
didoo, line 39: I asked copilot to provide me an example of a component in Ember that was both controlled and uncontrolled and this is what it came up with:
screenshot_3915
this made me question if we should determine if it's controlled not by the argument `isOpen` but by the fact that consumers would provide an `@onClickToggle` argument to handle/control the state change.
which btw is not very dissimilar from what we're doing here https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/pagination/compact/index.js#L43-L52 

just an extra data input, in case can help in some way

alex-ju:
Added changelog entry and rebased to `main`

packages/components/src/components/hds/disclosure-primitive/index.ts
jgwhite: I’d say the implementation you’ve landed on is somewhere between controlled and uncontrolled. I’d call it a “latch” — if the caller provides an initial `isOpen` state but not the `onChange` callback, then the component manages its state internally until the caller sends in a new `isOpen` state. Appropriate for this use case, but a little more complex then straight “controlled” (where the caller must also provide `onChange` if they want to control the state).

PR #2148, opened 2024-06-07
Change start command in docs to include build step

zamoore:
### :pushpin: Summary

If merged, this PR will change the instructions for running the Showcase application in the CONTRIBUTING document in the components package.

The command changes from `ember start` to `yarn run start`. Unless a contributor has already built the dependent packages, `ember start` won't work. Alternatively, `yarn run start` will work the first time because it first runs `yarn build:packages`.


didoo (approved):
👍  (🎉)

alex-ju (approved):

PR #2151, opened 2024-06-07
`doc-track-event` modifier

zamoore:
### :pushpin: Summary

Adds the `doc-track-event` modifier which can be used to track an event in fathom when clicking an element.

TODO

### DIFF
--- a/website/app/components/doc/link-with-icon/index.hbs
+++ b/website/app/components/doc/link-with-icon/index.hbs
@@ -9,7 +9,7 @@
     @route={{@route}}
     @models={{this.models}}
     ...attributes
-    {{on "click" (fn this.eventTracking.trackEvent @eventName)}}
+    {{doc-track-event @eventName}}
   >
     {{@label}}
     <FlightIcon class="doc-link-with-icon__icon" @name={{@icon}} />
@@ -21,7 +21,7 @@
     href={{@href}}
     target="_blank"
     rel="noopener noreferrer"
-    {{on "click" (fn this.eventTracking.trackEvent @eventName)}}
+    {{doc-track-event @eventName}}
     ...attributes
   >
     {{@label}}
--- a/website/app/modifiers/doc-track-event.js
+++ b/website/app/modifiers/doc-track-event.js
@@ -1,0 +1,34 @@
+import Modifier from 'ember-modifier';
+import { registerDestructor } from '@ember/destroyable';
+import { inject as service } from '@ember/service';
+
+function handleClick(eventName) {
+  window.fathom?.trackEvent(eventName);
+}
+
+function cleanup(instance) {
+  instance.eventName = null;
+  instance.element.removeEventListener('click', handleClick);
+}
+
+export default class DocTrackEvent extends Modifier {
+  @service fastboot;
+
+  element = null;
+  eventName = null;
+
+  modify(element, [eventName]) {
+    if (!this.fastboot.isFastBoot && window.fathom && eventName) {
+      this.addEventListener(element, eventName);
+
+      registerDestructor(this, cleanup);
+    }
+  }
+
+  addEventListener(element, eventName) {
+    this.element = element;
+    this.eventName = eventName;
+
+    element.addEventListener('click', handleClick(eventName));
+  }
+}
--- a/website/package.json
+++ b/website/package.json
@@ -122,6 +122,7 @@
     "remark-rehype": "^11.0.0",
     "resolve": "^1.22.8",
     "showdown": "^2.1.0",
+    "sinon": "^18.0.0",
     "stylelint": "^16.3.1",
     "stylelint-config-rational-order": "^0.1.2",
     "stylelint-config-standard-scss": "^13.1.0",
--- a/website/tests/integration/modifiers/doc-track-event-test.js
+++ b/website/tests/integration/modifiers/doc-track-event-test.js
@@ -1,0 +1,32 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+import { click } from '@ember/test-helpers';
+import sinon from 'sinon';
+
+module('Integration | Modifier | doc-track-event', function (hooks) {
+  setupRenderingTest(hooks);
+
+  let originalFathom;
+  let trackEventSpy;
+
+  hooks.beforeEach(function () {
+    originalFathom = window.fathom;
+    trackEventSpy = sinon.spy();
+    window.fathom = {
+      trackEvent: trackEventSpy,
+    };
+  });
+
+  hooks.afterEach(function () {
+    window.fathom = originalFathom;
+  });
+
+  test('it adds click event listener and calls handleClick', async function (assert) {
+    await render(hbs`<div {{doc-track-event 'testEvent'}}></div>`);
+    await click('div');
+
+    assert.ok(trackEventSpy.calledOnceWith('testEvent'));
+  });
+});
--- a/yarn.lock
+++ b/yarn.lock
@@ -4825,6 +4825,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@sinonjs/commons@npm:^3.0.1":
+  version: 3.0.1
+  resolution: "@sinonjs/commons@npm:3.0.1"
+  dependencies:
+    type-detect: "npm:4.0.8"
+  checksum: a0af217ba7044426c78df52c23cedede6daf377586f3ac58857c565769358ab1f44ebf95ba04bbe38814fba6e316ca6f02870a009328294fc2c555d0f85a7117
+  languageName: node
+  linkType: hard
+
 "@sinonjs/fake-timers@npm:^10.0.2":
   version: 10.3.0
   resolution: "@sinonjs/fake-timers@npm:10.3.0"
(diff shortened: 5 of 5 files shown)


website/app/modifiers/doc-track-event-on.js
```diff
@@ -0,0 +1,54 @@
+/**
```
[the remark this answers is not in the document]
Dhaulagiri: There is an automation that runs weekly that will add this if you forget to do it yourself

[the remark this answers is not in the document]
zamoore: Thank you!

website/app/components/doc/link-with-icon/index.hbs
```diff
@@ -9,7 +9,7 @@
     @route={{@route}}
     @models={{this.models}}
     ...attributes
-    {{on "click" (fn this.eventTracking.trackEvent @eventName)}}
+    {{doc-track-event-on "click" @eventName}}
```
didoo, line 12: general consideration, regarding the API of the modifier:
- since 99% of the cases will be on `click` what if we make this the default type of event trigger?
- in that case, it means we shouldn't use positional arguments, but named arguments

website/app/modifiers/doc-track-event-on.js
```diff
@@ -0,0 +1,54 @@
...
+    instance.triggerEvent,
+    handleTriggerEvent
+  );
+
+  instance.element = null;
+  instance.eventName = null;
+  instance.triggerEvent = null;
+}
+
+export default class DocTrackEventOn extends Modifier {
```
didoo, line 26: My understanding is that we can't use a [function-based modifier](https://github.com/ember-modifier/ember-modifier?tab=readme-ov-file#function-based-modifiers) here because we're using the `@service`, is that right?

website/app/modifiers/doc-track-event-on.js
```diff
@@ -0,0 +1,54 @@
...
+  @service fastboot;
+
+  element = null;
+  eventName = null;
+  triggerEvent = null;
+
+  modify(element, [triggerEvent, eventName]) {
```
didoo, line 33: currently we tend to use these names for the modifiers' arguments: `element`, `positional`, `named` and destructure them inside the function, see for example here:
https://github.com/hashicorp/design-system/blob/main/packages/components/src/modifiers/hds-tooltip.js#L72

website/app/modifiers/doc-track-event-on.js
```diff
@@ -0,0 +1,54 @@
...
+      eventName != null &&
+      triggerEvent != null
+    ) {
+      this.addEventListener(element, triggerEvent, eventName);
+
+      registerDestructor(this, cleanup);
+    }
+  }
+
+  addEventListener(element, triggerEvent, eventName) {
```
didoo, line 47: what is the benefit of having a dedicated function, that gets executed only once?

website/app/modifiers/doc-track-event-on.js
```diff
@@ -0,0 +1,54 @@
...
+  eventName = null;
+  triggerEvent = null;
+
+  modify(element, [triggerEvent, eventName]) {
+    if (
+      // Only attempt to do something if we are in the right environment
+      !this.fastboot.isFastBoot &&
+      window.fathom != null &&
+      eventName != null &&
+      triggerEvent != null
```
didoo, line 39: if we use `click` as default for `triggerEvent` we can remove this extra check

website/app/modifiers/doc-track-event-on.js
```diff
@@ -0,0 +1,54 @@
...
+  element = null;
+  eventName = null;
+  triggerEvent = null;
+
+  modify(element, [triggerEvent, eventName]) {
+    if (
+      // Only attempt to do something if we are in the right environment
+      !this.fastboot.isFastBoot &&
+      window.fathom != null &&
+      eventName != null &&
```
didoo, line 38: if the `eventName` is null, the modifier fails silently. should we instead add a check and fail with some form of error, so the developer will know they're missing an argument?

website/app/components/doc/link-with-icon/index.hbs
didoo: also, but this is a personal preference, I would keep the modifier name a bit simpler (`doc-track-event`)

didoo (commented):
Left a few comments

didoo:
@zamoore one thing I forgot to mention in the review: would it be possible to have this modifier implemented directly in TypeScript?

website/app/modifiers/doc-track-event-on.js
zamoore: Yes, exactly

website/app/components/doc/link-with-icon/index.hbs
zamoore: That's what I started with, but I added the `on` when I changed it to take different event triggers. I was meh about it though and don't have a strong opinion. I'll change it back.

website/app/modifiers/doc-track-event-on.js
zamoore: Sure, that makes sense.

website/app/modifiers/doc-track-event-on.js
zamoore: None, just that I used [this](https://github.com/ember-modifier/ember-modifier?tab=readme-ov-file#class-based-modifiers) as my jumping off point when writing this/refreshing my memory on modifier syntax.

zamoore:
> @zamoore one thing I forgot to mention in the review: would it be possible to have this modifier implemented directly in TypeScript?

It doesn't look like TS is setup for the website app, so that's probably out of scope for this issue.

didoo:
> > @zamoore one thing I forgot to mention in the review: would it be possible to have this modifier implemented directly in TypeScript?
> 
> It doesn't look like TS is setup for the website app, so that's probably out of scope for this issue.

@zamoore 🤦‍♂️ my bad

MelSumner (approved):

website/app/modifiers/doc-track-event.js
```diff
@@ -0,0 +1,60 @@
...
+
+    const hasValidEventName = typeof eventName === 'string';
+
+    assert(
+      `@eventName for "doc-track-event" must be a string; received: ${eventName}`,
+      hasValidEventName
+    );
+
+    this.element = element;
+    this.eventName = eventName;
```
didoo, line 53: Is it me, or these two are not used anywhere? 🤔
```suggestion
```

website/app/modifiers/doc-track-event.js
```diff
@@ -0,0 +1,60 @@
...
+    this.triggerEvent = triggerEvent;
+
+    element.addEventListener(this.triggerEvent, handleTriggerEvent(eventName));
```
didoo, line 56: I think you can avoid defining an extra variable, no?
```suggestion
    element.addEventListener(triggerEvent, handleTriggerEvent(eventName));
```

That said, there is a more burning problem here: `handleTriggerEvent(eventName)` is an invocation of a function, not a pointer to it, so this line actually executes the `window.fathom?.trackEvent(eventName)` which sends a call to the analytics service, even without the user interacting with the element (if you add a `console.log` to the `handleTriggerEvent` function you will see it).

So this line should be something like this:

```suggestion
    element.addEventListener(triggerEvent, handleTriggerEvent);
```

Now, in order to have make the `eventName` argument available to the `handleTriggerEvent` is to define the function _inside_  the "modifier" call, see for example how it's done here: https://github.com/hashicorp/design-system/blob/main/packages/components/src/modifiers/hds-clipboard.ts#L160-L179

Let me know if you want to pair on this, ok?

website/app/modifiers/doc-track-event.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    const { triggerEvent = 'click', eventName } = named;
+
+    const hasValidEventName = typeof eventName === 'string';
+
+    assert(
+      `@eventName for "doc-track-event" must be a string; received: ${eventName}`,
+      hasValidEventName
+    );
```
didoo, line 50: I don't think you need to define an extra variable, if it's used only once, no?
```suggestion
    assert(
      `@eventName for "doc-track-event" must be a string; received: ${eventName}`,
      typeof eventName === 'string'
    );
```

website/app/modifiers/doc-track-event.js
```diff
@@ -0,0 +1,60 @@
...
+
+function cleanup(instance) {
+  instance.element.removeEventListener(
+    instance.triggerEvent,
+    handleTriggerEvent
+  );
+
+  instance.element = null;
+  instance.eventName = null;
+  instance.triggerEvent = null;
```
didoo, line 24: See comments below, if you don't define them below you can also remove them from here
```suggestion
```

website/app/modifiers/doc-track-event.js
```diff
@@ -0,0 +1,60 @@
...
+      hasValidEventName
+    );
+
+    this.element = element;
+    this.eventName = eventName;
+    this.triggerEvent = triggerEvent;
+
+    element.addEventListener(this.triggerEvent, handleTriggerEvent(eventName));
+
+    registerDestructor(this, cleanup);
```
didoo, line 58: If the extra variables are removed, the `cleanup` function becomes just a call to `removeEventListener`. Does this mean that we could use a simple arrow function here (without declaring an extra `cleanup` function)?

```suggestion
    registerDestructor(this, () => {
      element.removeEventListener(this.triggerEvent, handleTriggerEvent);
    });
```

didoo (changes requested):
I've left a few comments, but one of them is a blocker.
As I said below, let me know if you prefer to pair on this.

website/app/modifiers/doc-track-event.js
zamoore: You're correct! These are no longer being used. Good call.

zamoore: `this.element` is accessed as `instance.element` on lines 17 & 22 in the cleanup function

`instance.element.removeEventListener`
`instance.element = null`

You're correct on `eventName` though, that can go.

PR #2152, opened 2024-06-10
HDS 3377 & HDS 3379 - Additional Fathom event tracking

zamoore:
### :pushpin: Summary

We want to add the following events

- `Navigation - Scroll to Top` event whenever a user clicks the "Scroll to top" button.
- `Footer - Accessibility`, `Footer - Careers`, & `Footer - Feedback` when corresponding footer links are clicked
- When clicking a link to a related component, track the event as `Related - FROM - TO`


website/app/templates/show.hbs
```diff
@@ -14,7 +14,7 @@
     <DynamicTemplate @templateString={{this.renderedContent}} @componentId={{this.model.id}} />
 
     {{#if this.relatedComponents}}
-      <Doc::RelatedComponents @cards={{this.relatedComponents}} />
+      <Doc::RelatedComponents @cards={{this.relatedComponents}} @title={{this.title}} />
```
didoo, line 17: Nitpick: I think `@component` or `@componentTitle` would work better as argument name because
- it's what we're technically passing
- will be more explicit what we're concatenating at line 20 of the `website/app/components/doc/related-components/index.hbs` file

website/app/components/doc/related-components/index.hbs
```diff
@@ -6,4 +6,18 @@
...
+  {{#each @cards as |card|}}
+    <Doc::Cards::Card
+      @image={{card.image}}
+      @title={{card.title}}
+      @caption={{card.caption}}
+      @route={{card.route}}
+      @model={{card.model}}
+      @href={{card.href}}
+      @layout="horizontal"
+      {{doc-track-event-on "click" "Related - {{@title}} - {{card.title}}"}}
```
didoo, line 20: Question for @Dhaulagiri @heatherlarsen: what are we trying to measure here?
- if people click on the "related" components?
- what of the "related components" are clicked, in which page?

If it was me, I would measure only the first, because the second one would be **very** fragmented and probably we would not see any meaningful signal

didoo (approved):
Left a few comments, but nothing blocking per se so I am preemptively approving 👍

website/app/components/doc/related-components/index.hbs
Dhaulagiri: That could be. I think the idea was that we could see which related components are being clicked from which parent component to see which combinations were being used the most.

website/app/modifiers/doc-track-event.js
```diff
@@ -0,0 +1,60 @@
...
+    const hasValidEventName = typeof eventName === 'string';
+
+    assert(
+      `@eventName for "doc-track-event" must be a string; received: ${eventName}`,
+      hasValidEventName
+    );
+
+    this.element = element;
+    this.eventName = eventName;
+    this.triggerEvent = on;
```
didoo: what do you think of this?
```suggestion
    this.triggerEvent = triggerEvent;
```

(where `triggerEvent` is the variable destructured at line 43)

website/tests/integration/modifiers/doc-track-event-test.js
```diff
@@ -0,0 +1,41 @@
...
+  });
+
+  test('it adds click event listener by default and tracks an event', async function (assert) {
+    await render(hbs`<div {{doc-track-event eventName='testEvent'}}></div>`);
+    await click('div');
+
+    assert.ok(trackEventSpy.calledOnceWith('testEvent'));
+  });
+
+  test('it adds a focus event listener and tracks an event', async function (assert) {
```
didoo: 👏

didoo (approved):
Left a minor comment (about naming). Apart from that, all good for me 👍

website/app/components/doc/related-components/index.hbs
```diff
@@ -6,4 +6,18 @@
...
+  {{#each @cards as |card|}}
+    <Doc::Cards::Card
+      @image={{card.image}}
+      @title={{card.title}}
+      @caption={{card.caption}}
+      @route={{card.route}}
+      @model={{card.model}}
+      @href={{card.href}}
+      @layout="horizontal"
+      {{doc-track-event eventName="Related - {{@componentTitle}} - {{card.title}}"}}
```
Dhaulagiri, line 20: @zamoore this isn't actually working. I think you need to `concat` these values into this string or something similar

Screenshot 2024-06-13 at 09 42 39

zamoore: Addressed [here](https://github.com/hashicorp/design-system/pull/2165)

PR #2155, opened 2024-06-11
converted the application-state component to ts

zamoore:
### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/components/hds/application-state/body.ts
+++ b/packages/components/src/components/hds/application-state/body.ts
@@ -1,0 +1,18 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+
+export interface HdsApplicationStateBodySignature {
+  Args: {
+    text?: string;
+  };
+  Blocks: {
+    default?: [];
+  };
+  Element: HTMLDivElement;
+}
+
+export default class HdsApplicationStateBodyComponent extends Component<HdsApplicationStateBodySignature> {}
--- a/packages/components/src/components/hds/application-state/header.ts
+++ b/packages/components/src/components/hds/application-state/header.ts
@@ -1,0 +1,17 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+
+export interface HdsApplicationStateHeaderSignature {
+  Args: {
+    title: string;
+    errorCode?: string;
+    icon?: string;
+  };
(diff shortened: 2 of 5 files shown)


alex-ju (commented):
Generally looks good! There are a couple of templates within `application-state` where we need to remove the `@glint-nocheck` from the top of the file

packages/components/src/components/hds/application-state/body.ts
```diff
@@ -0,0 +1,25 @@
...
+}
+
+export default class HdsApplicationStateBodyComponent extends Component<HdsApplicationStateBodySignature> {}
+
+declare module '@glint/environment-ember-loose/registry' {
+  export default interface Registry {
+    'Hds::ApplicationState::Body': typeof HdsApplicationStateBodyComponent;
+    'hds/application-state/body': typeof HdsApplicationStateBodyComponent;
+  }
+}
```
alex-ju, line 25: Thanks for addressing the feedback! We have a shared registry (`template-registry.ts`) where we export the aliases. If you've not seen this yet, we have [a short migration guide](https://github.com/hashicorp/design-system/blob/main/wiki/TypeScript-Migration.md) to make sure we keep things consistent. We'd also need a re-export in `packages/components/src/components.ts` (see step 7).

packages/components/src/components/hds/application-state/header.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import TemplateOnlyComponent from '@ember/component/template-only';
+
+export interface HdsApplicationStateHeaderSignature {
+  Args: {
+    title?: string;
```
MelSumner, line 10: **Question:** Are you sure this is optional? Could you check that? I think this _might_ provide the ACCNAME in which case, it shouldn't be optional.

alex-ju: Fair question – I made that comment based on the current documentation. Happy for us to revisit this.

MelSumner: If it's according to the docs, then it's fine; but we should probably re-visit this component and look at the pattern/guidance and see if everything really should be optional.

alex-ju: I created [a task](https://hashicorp.atlassian.net/browse/HDS-3384) for us to look into it.

alex-ju (approved):
Great work!

packages/components/src/components/hds/application-state/header.ts
zamoore: Thanks for the discussion all, and for creating that task @alex-ju

PR #2159, opened 2024-06-12
remove unnecessary assertion

MelSumner:
### :pushpin: Summary

If merged, this PR will remove an assertion that should never need to fire, because the underlying property has a default value.

While looking at this component's code to add a new feature, I noticed an assertion that should not be necessary. In the [Additional Options](https://github.com/ember-a11y/ember-a11y-refocus?tab=readme-ov-file#additional-options) section of the addon's README, it indicates that if no custom value is provided for `skipTo`, then it defaults to `#main`. 

- no changelog because this is an internal change
- removed the assertion (conditional statement)
- removed the assert import for the component
- removed the related test

### DIFF
--- a/packages/components/src/components/hds/side-nav/index.ts
+++ b/packages/components/src/components/hds/side-nav/index.ts
@@ -6,7 +6,6 @@
 import Component from '@glimmer/component';
 import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
-import { assert } from '@ember/debug';
 import { registerDestructor } from '@ember/destroyable';
 
 import type { HdsSideNavBaseSignature } from './base';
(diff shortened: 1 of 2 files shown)


didoo:
@MelSumner this implementation was intentional, to "force" the consumers to declare the target explicitly, even if there is a default `#main` in the addon, because the ID of the DOM node is very likely specific to every app (and `#main` is totally arbitrary).

MelSumner:
@didoo I do understand the intent; however I question if an assertion can be forced for something "undefined" if it's...never undefined. Am I missing a trick in the implementation?

didoo:
> it's...never undefined.
@MelSumner it's never undefined _in the addon_, here we're intercepting its value (if undefined) _before_ it's passed down to the addon, and we don't want that in the `Hds::SideNav` that argument to be undefined. Makes sense?

MelSumner:
@didoo ok, I'll close this PR. 👍

PR #2161, opened 2024-06-13
Enterprise Nav: `AppHeader` component (HDS-3373)

KristinLBradley:
_STATUS: Prepping for release_

### :pushpin: Summary
If merged, this PR adds the `AppHeader` component to HDS.

* MERGED IN: ~~**Related PR (web docs):** https://github.com/hashicorp/design-system/pull/2239~~

**Showcase:**
- AppHeader: https://hds-showcase-git-hds-3373-app-header-component-hashicorp.vercel.app/components/app-header
- AppHeader in AppFrame: https://hds-showcase-git-hds-3373-app-header-component-hashicorp.vercel.app/layouts/app-frame#content
- SideNav w/o :header & :footer: https://hds-showcase-git-hds-3373-app-header-component-hashicorp.vercel.app/components/side-nav#only-body-content-used-with-appheader

Changes include:
* New `AppHeader` component, including:
  - `:logo` Named block: Used to contain HomeLink/Logo
  - `:globalActions` named block: Used mainly to contain OrgPicker
  - `:utilityActions` named block: Used for Search, HelpMenu, GeoPicker, UserMenu, etc.
* Modifications to `SideNav` to coordinate with `AppHeader`
  - Adjusted padding & spacing when :header & :footer blocks are left empty
  - Adjusted position of Toggle button tab when :header block is empty
  - Added `@hasHeader` arg to adjust height and position of SideNav to account for fixed position AppHeader
* Style addition to AppFrame to reserve space for for fixed position `AppHeader` preventing it from overlapping main content
* Added new related examples to `SideNav` & `AppFrame` Showcases.  
(In addition to new `AppHeader` Showcase)

👉 Tested AppHeader & modified SideNav within HCP/Cloud-ui & Terraform/Atlas: https://hashicorp.atlassian.net/browse/HDS-3376

**Desktop view:**
image

**Mobile/Tablet view:**
(_Menu closed_)
image

(_Menu open_)
image

**Used within HCP/Cloud-ui:**
image

**Used within Terraform/Atlas:**
image

**Edge case test with extra long content**
_Layout shifts to mobile view after this point:_
image

### :link: External links

- Jira ticket: [HDS-3373](https://hashicorp.atlassian.net/browse/HDS-3373)
- Figma file: https://www.figma.com/design/noyY6dUMDYjmySpHcMjhkN/branch/5RLNe1engBID42LhAm75NN/HDS-Product---Components?node-id=58766-15927&t=JLqVjKsbPaRrpAqt-0

***

### 👀 Component checklist

[HDS-3373]: https://hashicorp.atlassian.net/browse/HDS-3373?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/components/hds/app-header/home-link.hbs
```diff
@@ -0,0 +1,20 @@
...
+  @replace={{@replace}}
+  @route={{@route}}
+  @isRouteExternal={{@isRouteExternal}}
+  @href={{@href}}
+  @isHrefExternal={{@isHrefExternal}}
+  ...attributes
+  aria-label={{this.ariaLabel}}
+>
+  <FlightIcon @name={{@icon}} @color={{@color}} @stretched={{true}} />
+</Hds::Interactive>
```
KristinLBradley, line 20: This is the same as the SideNav::HomeLink component...

packages/components/src/components/hds/app-header/index.hbs
```diff
@@ -0,0 +1,16 @@
...
+  <div class="hds-app-header__sub-section">
+    {{yield to="globalBefore"}} {{! Intended for: ProductMenu }}
+    {{yield to="logo"}}
+    {{yield to="globalAfter"}} {{! Intended for: OrgPicker }}
+  </div>
+
+  <div class="hds-app-header__sub-section">
+    {{yield to="utility"}} {{! Intended for: Search, HelpMenu, GeoPicker, UserMenu }}
+  </div>
+</div>
```
KristinLBradley, line 16: This is the current structure I defined. Subject to change. (Would appreciate feedback on this structure.)

Need to test within product apps next...

packages/components/src/components/hds/app-header/index.ts
```diff
@@ -0,0 +1,43 @@
...
+  Element: HTMLDivElement;
+}
+// More info on types and signatures: https://github.com/hashicorp/design-system/blob/main/wiki/TypeScript-Migration.md
+
+export default class HdsAppHeaderComponent extends Component<HdsAppHeaderSignature> {
+  // UNCOMMENT THIS IF YOU NEED A CONSTRUCTOR
+  // constructor() {
+  //   super(...arguments);
+  //   // ADD YOUR ASSERTIONS HERE
+  // }
```
KristinLBradley, line 28: Will clean all this up later.

packages/components/src/components/hds/app-header/index.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export interface HdsAppHeaderSignature {
+  Blocks: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    globalBefore?: any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    logo: any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    globalAfter?: any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
KristinLBradley, line 16: Used "any" type for now for these. Should these have more specific types?

(The logo named block is intended to only accept the HomeLink component so I may update that next.)

packages/components/src/styles/components/app-header.scss
```diff
@@ -0,0 +1,71 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+//
+// APP-HEADER
+//
+
+@use "../mixins/focus-ring" as *;
+
+// TODO: Copy & pasted, clean up
```
KristinLBradley, line 12: Still need to refine styles. These aren't yet following latest designs. Will update.

packages/components/src/styles/components/side-nav/main.scss
```diff
@@ -111,6 +111,12 @@
     padding-right: var(--token-side-nav-wrapper-padding-horizontal-minimized);
     padding-left: var(--token-side-nav-wrapper-padding-horizontal-minimized);
   }
+
+  // Remove padding when it contains no child elements
+  // Note: styles will also apply if it contains only text nodes
+  &:not(:has(*)) {
```
KristinLBradley, line 117: This is the solution I came up with so far to remove excess space when the SideNav header and footer aren't used (when it's paired with the `AppHeader`). Curious what others will think... 😅

didoo: If you remove the whitespace using `~` in the Handlebars template and use `:empty` in CSS I _think_ it should work

KristinLBradley: I kept trying that and there was still whitespace so I couldn't get it to work. I'll try again though.

didoo: @KristinLBradley let me know if you want to pair on this (did you remove the whitespace in both the `index.hbs` and the `base.hbs` files?)

KristinLBradley: Thanks, I looked at it again and realized I had missed adding "squishies" to the base.hbs component when I experimented previously. Once I added them there I was able to get it to work. :)

showcase/app/templates/components/app-header.hbs
```diff
@@ -0,0 +1,175 @@
...
+          <Hds::Dropdown class="hds-side-nav__dropdown" as |dd|>
+            <dd.ToggleButton @text="Choose an organization" @icon="org" />
+            <dd.Checkmark>
+              organizationName
+            </dd.Checkmark>
+          </Hds::Dropdown>
+        </:globalAfter>
+
+        <:utility>
+          <Hds::Button class="hds-side-nav__icon-button" @icon="search" @isIconOnly={{true}} @text="Search" />
```
KristinLBradley, line 99: Icon only buttons **icon size** vs. Dropdown icon size needs to be more closely examined and perhaps reworked. (Examined & explored this and discussed with @jorytindall.)

I also am now realizing that the "one-off" `SideNav::Header::IconButton` was likely a mistake that we should also rethink. http://localhost:4201/components/side-nav#sidenav-header-iconbutton

We should look to come up with a better strategy for implementing the **dark "theme"** for the Button and Dropdown as well. (@jorytindall as a reminder)

showcase/app/templates/components/app-header.hbs
```diff
@@ -0,0 +1,175 @@
...
+          </Hds::Dropdown>
+        </:utility>
+      </Hds::AppHeader>
+    </SG.Item>
+
+    <SG.Item @label="With max recommended future content">
+      <Hds::AppHeader>
+        <:globalBefore>
+          {{! TODO: Rethink / Refactor buttons (Hds::SideNav::Header::IconButton) for use in AppHeader (& SideNav ) }}
+          <Hds::Button class="hds-side-nav__icon-button" @icon="grid-alt" @isIconOnly={{true}} @text="Product menu" />
```
KristinLBradley, line 127: Note: Reusing class name from side-nav currently to start but we need to rethink what approach is best for both the "dark theme" and icon sizing.

KristinLBradley:
@jorytindall and I discovered a possible design issue with the sizing of icons in the icon only `Button` vs. in the `Dropdown` which we may want to explore solutions for:
image

On the right side in the "utility" items area, notice that the search icon in the button is noticeably smaller than the icons in the dropdowns next to it.

For the SideNav, we had ended up creating a `Hds::SideNav::Header::IconButton` child component which both increased the size of the icon and also applied dark theme styling. This was probably not the best solution however so I think we should rethink this as part of the enterprise nav work.

**2 issues to consider:**
* Visual inconsistency of icon sizing in `Button` vs. `Dropdown`
  - We could consider reducing the icon size in the Dropdown to match the Button.
* Cohesive dark theme/design variant solution for `Button` & `Dropdown` 
  - Currently we apply an extra classname for SideNav Dropdowns while we created a child component for Buttons. These 2 approaches should either be replaced or consolidated.
  - We could possibly add "dark" variant theme options to Button & Dropdown as a solution.

packages/components/src/styles/components/app-header.scss
```diff
@@ -0,0 +1,76 @@
...
+  border: 1px solid transparent;
+  border-radius: var(--token-side-nav-body-list-item-border-radius);
+  cursor: pointer;
+
+  @if ($add-visible-border) {
+    border-color: var(--token-color-palette-neutral-500);
+  }
+
+  &:focus,
+  &.mock-focus {
```
jorytindall, line 25: I think this is getting overridden somewhere with the default styles
Screenshot 2024-06-14 at 1 36 37 PM

jorytindall (commented):
Left one initial note (I know this is a draft) on the focus/active state of the icon-only buttons being overridden. This is looking really solid so far!

One thing to consider is the width of the various "pickers" (context switcher/geo picker):
- Since they aren't within the `SideNav` anymore, what should the width of these elements be?
- Should they adapt to the content within (and change when something longer/shorter is selected)?
- Should we set a min/max width? That might result in a long label getting cut off, or requiring truncation 👎

I'll think about this a bit, but just highlighting as another topic of discussion. Great job getting this throw together so quickly!

packages/components/src/styles/components/app-header.scss
KristinLBradley: Thanks, I haven't worked on "real" styles for the dropdown and button content yet (was waiting for visual designs to be more nailed down before over-focusing on styling details so I mainly focused on the AppHeader and the slots where things get dropped in to start) so haven't looked at focus styles or anything like that. Those are part of what I want to focus on next which is part of why I brought up the questions about icon sizing and how best to handle the dark "theme".

showcase/app/templates/components/app-header.hbs
```diff
@@ -0,0 +1,175 @@
...
+
+  <Shw::Grid @columns={{1}} {{style gap="2rem"}} as |SG|>
+    <SG.Item @label="With minimum recommended phase 1 content">
+      <Hds::AppHeader>
+        <:logo>
+          <Hds::AppHeader::HomeLink @icon="hashicorp" @ariaLabel="HashiCorp home menu" @href="#" />
+        </:logo>
+
+        <:globalAfter>
+          <Hds::Dropdown class="hds-side-nav__dropdown" as |dd|>
```
KristinLBradley, line 52: **Note:** classnames are used here _temporarily_ for initial rough visual styling purposes until visual designs are more nailed down and also until we decide upon the best way to handle dark theming.

packages/components/src/components/hds/app-header/index.ts
KristinLBradley: I tried changing the logo block to only accept the HomeLink component.

Example: `Logo: ComponentLike<HdsAppHeaderHomeLinkSignature>;`

Which was fine but then I couldn't get it to work within the hbs template. I think because it won't work together with using a named block.

showcase/app/templates/components/app-header.hbs
didoo: Not sure what you need as feedback here, so I'll go with my instinct:

The size of the icon in the IconButton appears to be the same as the icons on the dropdown": https://hds-showcase.vercel.app/components/side-nav#using-yielded-content
screenshot_3857

If instead of the `<Hds::Button>` I use the `<Hds::SideNav::Header::IconButton>` (that should be probably cloned into a `<Hds::AppHeader::IconButton>` component) this is what I see:
screenshot_3858
screenshot_3859

didoo (commented):
I've left a couple of comments. Not sure what else provide feedback on, at this stage, in case let me know and I can look in more detail.

showcase/app/templates/components/app-header.hbs
KristinLBradley: The `Hds::SideNav::Header::IconButton` forces a non-standard icon size in addition to adding dark theme styling. It was created to follow custom styling Terraform had been using for their search button. In my mind, needing a special component to change the button icon size highlights a possible design issue which should be solved in the base `Button` and `Dropdown` components. @jorytindall and I can look at this though and decide what to do.

This is the `AppHeader` using standard icon sizes:
image

packages/components/src/components/hds/app-header/index.hbs
```diff
@@ -0,0 +1,16 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<div class={{this.classNames}} ...attributes>
+  <div class="hds-app-header__sub-section">
+    {{yield to="globalBefore"}} {{! Intended for: ProductMenu }}
```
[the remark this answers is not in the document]
KristinLBradley: My original intention was to enforce usage of the HomeLink/Logo and a recommended order for the content. In thinking it over though, you are right that with the current 3 slots, the order isn't actually enforced. I also talked it over with @jorytindall who brought up that there's less likely potential for misuse of this type of component than with more general use components (since consumers would add it just in one or a few places and then mostly leave it alone).

I'll go ahead and refactor it into one general slot for the global items.

packages/components/src/styles/components/app-header.scss
KristinLBradley: Just overhauled all the interactive children styles (Buttons & Dropdowns). Added examples to Showcase. Now ready for visual design reviews. @jorytindall

KristinLBradley: Still need to make updates to Dropdown icons, etc. next BTW.

packages/components/src/styles/components/app-header.scss
```diff
@@ -0,0 +1,51 @@
...
+  box-shadow: 
+    0 4px 4px -2px hsla(0deg, 0%, 0%, 20%),
+    0 2px 3px 0 hsla(0deg, 0%, 0%, 10%);
+
+  // Apply dark theme to child interactive components
+  .hds-button,
+  .hds-dropdown-toggle-button,
+  .hds-dropdown-toggle-icon {
+    @include hds-interactive-dark-theme();
+  }
```
KristinLBradley, line 35: @didoo Let me know if you think it's a bad idea to apply styles to the buttons & dropdowns by nesting this way...

packages/components/src/styles/components/side-nav/header.scss
```diff
@@ -7,40 +7,7 @@
...
-  &.mock-active {
-    color: var(--token-side-nav-color-foreground-strong); // to avoid overrides by specificity (eg. `a:hover`)
-    background: var(--token-side-nav-color-surface-interactive-active);
-
-    @if ($add-visible-border) {
-      border-color: var(--token-color-palette-neutral-400);
-    }
-  }
-}
+@use "../../mixins/interactive-dark-theme" as *;
```
KristinLBradley, line 10: By replacing the original mixin with this new shared one, the SideNav color tokens are no longer being used.

@didoo Please let me know what you think of this. (Is it a bad idea to use a shared mixin?)

packages/components/src/styles/components/app-header.scss
didoo: no, actually is quite neat 👍

packages/components/src/styles/components/side-nav/header.scss
didoo: > the SideNav color tokens are no longer being used.

This is a tricky one! 
In theory we could deprecate and then remove them, and B) use the raw "palette" colors in the mixin like you're doing, or B) introduce a new set of tokens, more generic, and deprecate and then remove them. 
In reality these tokens are used in some our consumers codebases: https://github.com/search?q=%22--token-side-nav-color%22++user%3Ahashicorp+-repo%3Ahashicorp%2Fdesign-system&type=code&ref=advsearch
So the question becomes: what is the _correct™_ and most clean way of doing it (preserving the shared mixin, which I think is a good idea)? 🤔

KristinLBradley: Thanks, I'll have to think more...

KristinLBradley: Alright so what I ended up doing for now was that I updated the mixin to include an `$is-within-side-nav` option with conditionals that will use the sidenav tokens instead of the general tokens. Maybe a bit messy, but what do you think?

didoo: I'll have a look tomorrow

packages/components/src/styles/mixins/_interactive-dark-theme.scss
```diff
@@ -0,0 +1,95 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+ @use "../mixins/focus-ring" as *;
+
+ // TODO: Deprecate SideNav specific dark theme tokens in favor of general dark theme tokens
+
+// Used to apply dark theme to interactive elements such as Button & Dropdown
+@mixin hds-interactive-dark-theme($add-visible-border: true, $is-within-side-nav: false) {
+  @if $is-within-side-nav {
```
didoo, line 12: It's not _that_ bad, at least it's hidden all in one place

packages/components/src/styles/mixins/_interactive-dark-theme.scss
```diff
@@ -0,0 +1,95 @@
...
+    // sidenav specific dark theme tokens
+    color: var(--token-side-nav-color-foreground-strong);
+    background-color: transparent; // test
+    border-radius: var(--token-side-nav-body-list-item-border-radius);
+  } @else {
+    // general dark theme tokens
+    color: var(--token-color-palette-neutral-0);
+    background-color: var(--token-color-palette-neutral-700);
+    border-radius: var(--token-form-control-border-radius);
```
didoo, line 21: technically it's not the `form-control-border-radius` but something like `button-border-radius`; unfortunately we don't have such token, only a variable `$hds-button-border-radius` in the `mixins/_button.scss`, maybe you can import the mixin and use that variable instead? (I've never tried, but probably should work)

KristinLBradley: Ok, will see. (I picked that one because it seemed closest.)

packages/components/src/components/hds/app-header/index.hbs
```diff
@@ -0,0 +1,34 @@
...
+<div class={{this.classNames}} ...attributes>
+  {{!-- GlobalItems --}}
+  <div class="hds-app-header__sub-section">
+    {{! Intended for: ProductMenu, HomeLink/Logo }}
+    {{yield to="globalItemsBefore"}}
+    {{! Intended for: OrgPicker }}
+    <div class="hds-app-header__global-items-mobile-view">
+      {{yield to="globalItemsAfter"}}
+    </div>
+  </div>
```
KristinLBradley, line 15: Refactored to divide `:globalItems` into `:globalItemsBefore` & `:globalItemsAfter`. This is to allow the globalItemsAfter content to be rendered within the responsive dropdown menu which is rendered in mobile/tablet view.

(Not sure if this is the best naming or best way to implement this.)

packages/components/src/components/hds/app-header/index.hbs
```diff
@@ -0,0 +1,34 @@
...
+    {{!-- Button appears in mobile/tablet view (hidden in desktop view) --}}
+    <Hds::AppHeader::MenuButton 
+      @onClickToggle={{this.onClickToggle}} 
+      @isOpen={{this.isOpen}} 
+      @menuContentId={{this.menuContentId}} 
+    />
+
+    {{! Intended for: Search, HelpMenu, GeoPicker, UserMenu }}
+    <div class="hds-app-header__menu" id={{this.menuContentId}}>
+      <div class="hds-app-header__global-items-desktop-view">
```
KristinLBradley, line 28: Not sure if these are the best class names:
"hds-app-header__global-items-mobile-view"
& "hds-app-header__global-items-mobile-view" (used above within globalItems)

jorytindall: Is this essentially handling the "conditional" rendering of the elements based on the viewport?

packages/components/src/styles/components/app-header.scss
```diff
@@ -0,0 +1,119 @@
...
+    // Display content as dropdown menu in mobile/tablet view
+    .hds-app-header__menu {
+      position: absolute;
+      top: calc(100% + 1px); // We add 1px to account for the bottom border
+      right: 0;
+      left: 0;
+      z-index: 5; // TODO: determine correct z-index to use
+      display: flex;
+      flex-wrap: wrap;
+      // height: calc(100vh - 52px); // it should take the full height of the viewport minus the header height
```
jorytindall, line 49: Is this commented out specifically for the showcase?

packages/components/src/styles/components/app-header.scss
```diff
@@ -0,0 +1,119 @@
...
+    // Hide :globalAfter content in mobile/tablet view
+    .hds-app-header__global-items-mobile-view {
+      display: none;
+    }
+
+    // Since the same component is used for both mobile/tablet and desktop views, we need to force the width to 100% in mobile view
+    .hds-app-header__global-items-desktop-view {
+      &,
+      .hds-dropdown,
+      .hds-dropdown-toggle-button {
```
jorytindall, line 68: Should the list of items in the Dropdown also be set to width 100%? I mocked [this up quickly](https://www.figma.com/design/noyY6dUMDYjmySpHcMjhkN/branch/5RLNe1engBID42LhAm75NN/HDS-Components?m=auto&node-id=60962-22032&t=ptWRQcJDDMh0jOFM-1) for reference

packages/components/src/styles/components/app-header.scss
```diff
@@ -0,0 +1,119 @@
...
+@use "../mixins/interactive-dark-theme" as *;
+
+// ----------------
+// Q: Create tokens?
+$app-header-home-link-padding: 4px;
+
+// Breakpoints WIP: START ----------------------------
+
+// Break points
+$tablet-min-width: 768px;
```
jorytindall, line 19: Is this breakpoint being used? Or is this a leftover from some of the earlier design explorations.

packages/components/src/styles/components/app-header.scss
```diff
@@ -0,0 +1,119 @@
...
+
+.hds-app-header {
+  position: relative;
+  display: flex;
+  gap: 12px;
+  align-items: center;
+  justify-content: space-between;
+  min-height: 60px; 
+  padding: 12px 16px;
+  color: #FFF;
```
jorytindall, line 94: Should this be:

```suggestion
  color: var(--token-color-foreground-high-contrast);
```

packages/components/src/styles/components/app-header.scss
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+}
+
+// Breakpoints WIP: END ------------------------------
+
+.hds-app-header {
+  position: relative;
+  display: flex;
+  gap: 12px;
```
jorytindall, line 89: Just so I'm following here, setting this at the root level of the component cascades down to the child elements to enforce a consistent gap in _all_ nested child elements?

packages/components/src/styles/components/app-header.scss
```diff
@@ -0,0 +1,119 @@
...
+  padding: 12px 16px;
+  color: #FFF;
+  background: var(--token-color-palette-neutral-700);
+  border-bottom: 1px solid var(--token-color-palette-neutral-500);
+
+  // Apply dark theme to child interactive components
+  .hds-button,
+  .hds-dropdown-toggle-button,
+  .hds-dropdown-toggle-icon {
+    @include hds-interactive-dark-theme();
```
jorytindall, line 102: 🙌

packages/components/src/styles/components/app-header.scss
```diff
@@ -0,0 +1,119 @@
...
+
+// Breakpoints WIP: END ------------------------------
+
+.hds-app-header {
+  position: relative;
+  display: flex;
+  gap: 12px;
+  align-items: center;
+  justify-content: space-between;
+  min-height: 60px;
```
jorytindall, line 92: I see the `min-height` here, and technically the resulting outcome in Figma is 60px of height, but the showcase is 61px, is there a reason for this or is it an issue?

packages/components/src/styles/components/app-header.scss
jorytindall: Also asking because in the `AppFrame` showcase I guess I would expect to see a more realistic example where the mobile menu occupies the entire height.

jorytindall (commented):
This is looking GREAT, love to see the work, I mostly just left questions and few things to figure out. I can provide a more thorough review and a ✅ when the rest of the details have been settled.

packages/components/src/styles/components/app-header.scss
KristinLBradley: Yes, just for now. I plan to add it back and test within HCP & Terraform. I'll add another style to override it within the Showcase.

packages/components/src/components/hds/app-header/index.hbs
KristinLBradley: Updated now but don't know if the new names are much better...

packages/components/src/styles/components/app-header.scss
KristinLBradley: I'm worried about trying to override/hack too many styles but I can see...

(Actually don't know if the way I'm forcing the width now is the best way to handle this...)

packages/components/src/styles/components/app-header.scss
KristinLBradley: Left over. Cleaned everything up in the latest push.

packages/components/src/styles/components/app-header.scss
KristinLBradley: Perhaps. Let me see. I need to get back to looking at theming and design tokens. I took a break from focusing on those to get the responsive behavior working...

packages/components/src/styles/components/app-header.scss
KristinLBradley: It does wherever I'm telling it to inherit the style (`gap: inherit;`)

packages/components/src/styles/components/app-header.scss
KristinLBradley: I probably forgot to account for the border as I updated the height just recently. I will fix.

KristinLBradley: I think I need to adjust padding values too. Will look at it more closely, thanks

packages/components/src/styles/components/app-header.scss
jorytindall: To potentially match the theming, maybe it should be `token-color-foreground-primary`? At least that's somewhat of the equivalent in the light theme

KristinLBradley:
> Left one initial note (I know this is a draft) on the focus/active state of the icon-only buttons being overridden. This is looking really solid so far!
> 
> One thing to consider is the width of the various "pickers" (context switcher/geo picker):
> 
> * Since they aren't within the `SideNav` anymore, what should the width of these elements be?
> * Should they adapt to the content within (and change when something longer/shorter is selected)?
> * Should we set a min/max width? That might result in a long label getting cut off, or requiring truncation 👎
> 
> I'll think about this a bit, but just highlighting as another topic of discussion. Great job getting this throw together so quickly!

**Update:** I've since added examples to the showcase of longer names in the OrgSwitcher. Both HCP & Terraform limit the length of names to 40 characters max.

packages/components/src/styles/components/app-header.scss
KristinLBradley: Uncommented height and added additional related layout styles with overrides for Showcase examples.

packages/components/src/styles/components/app-header.scss
KristinLBradley: Thanks, I will adopt your suggestion for now. I'm currently leaning towards avoiding creating new color tokens for phase 1 but will see...

packages/components/src/styles/components/side-nav/header.scss
KristinLBradley: Update: I'm leaning towards keeping things simple for phase one and avoiding introducing new color tokens while designs are still in flux.

packages/components/src/styles/components/app-header.scss
KristinLBradley: Applied locally

packages/components/src/styles/components/app-header.scss
KristinLBradley: This should be correct now. I adjusted and compared screenshots directly to Figma components

packages/components/src/styles/mixins/_interactive-dark-theme.scss
```diff
@@ -0,0 +1,97 @@
...
+  &:focus,
+  &:focus-visible,
+  &.mock-focus {
+    @include hds-focus-ring-with-pseudo-element($top: -1px, $right: -1px, $bottom: -1px, $left: -1px);
+
+    color: var(--token-color-palette-neutral-0);
+    background-color: var(--token-color-palette-neutral-700);
+
+    @if ($add-visible-border) {
+      border-color: var(--token-color-palette-neutral-500); // TODO: confirm if this is the correct token
```
KristinLBradley, line 41: @jorytindall Is this the correct border color for the focused Buttons/Dropdowns?

packages/components/src/styles/components/side-nav/header.scss
```diff
@@ -90,19 +57,24 @@
...
-  .hds-dropdown-toggle-button {
-    @include hds-side-nav-icon-button($add-visible-border: true);
+.hds-side-nav-header,
+.hds-side-nav__wrapper-body,
+.hds-side-nav__wrapper-footer {
+  // Apply dark theme to child interactive components
+  .hds-button,
+  .hds-dropdown-toggle-button,
+  .hds-dropdown-toggle-icon {
+    @include hds-interactive-dark-theme();
```
KristinLBradley, line 69: Previously the interactive-dark-theme colors were applied to SideNav Dropdowns using a classname, ".hds-side-nav__dropdown". This change removes the need for a classname to instead rely upon inheritance as I did for Buttons & Dropdowns in the AppHeader.

I did a search for the classname within consumer codebases and it was not being overridden anywhere: https://github.com/search?q=org%3Ahashicorp+hds-side-nav__dropdown&type=code

However, if you think it would be better to instead deprecate the class name and remove in a separate PR I can do so.

didoo: @KristinLBradley what I can see from the search results is that it's not being _overwritten_, but it's being _used_ to apply a dark mode to some dropdowns: https://github.com/search?q=org%3Ahashicorp+hds-side-nav__dropdown+-repo%3Ahashicorp%2Fdesign-system&type=code

From what I understand
- [here](https://github.com/hashicorp/consul/blob/main/ui/packages/consul-ui/app/components/hashicorp-consul/index.hbs#L88) it's used in the SideNav header, so you should add `hds-side-nav__wrapper-header` to the selector above
- [here](https://github.com/hashicorp/consul/blob/main/ui/packages/consul-acls/app/components/consul/token/selector/index.hbs#L60) + [here](https://github.com/hashicorp/consul/blob/main/ui/packages/consul-ui/app/components/hashicorp-consul/index.hbs#L108) it's being used also in the SideNav header
- other instances too seem to also in the context of the SideNav
So _in theory_ it should be ok to just remove the class and rely on inheritance. One consideration is that, since the `hds-side-nav__wrapper-[header/body/footer]` are practically the only containers for the `SideNav` (apart from the `root/yield`) then maybe you can simplify things and just use `hds-side-nav` as parent class for the inheritance.

packages/components/src/styles/mixins/_interactive-dark-theme.scss
jorytindall: I've better mapped these out [here](https://www.figma.com/design/noyY6dUMDYjmySpHcMjhkN/branch/5RLNe1engBID42LhAm75NN/HDS-Components?m=auto&node-id=59659-45808&t=CICqCkO7FCafQBOb-1), but the intention is that it should match the re-themed dropdown.

packages/components/src/styles/components/side-nav/header.scss
KristinLBradley: Thanks, I'll simplify the selectors and also add back the ".hds-side-nav__dropdown" classname for now.

KristinLBradley: Resolving for now, thanks for your help

packages/components/src/styles/mixins/_interactive-dark-theme.scss
KristinLBradley: Pushed some updates to better match mocks

packages/components/src/styles/mixins/_interactive-dark-theme.scss
KristinLBradley: Thanks, to me I think this is ok for phase 1 & could be refined/replaced as needed in upcoming phases.

packages/components/src/components/hds/accordion/index.ts
```diff
@@ -26,7 +26,7 @@ interface HdsAccordionSignature {
     default: [
       {
         Item?: ComponentLike<HdsAccordionItemSignature>;
-      },
+      }
```
KristinLBradley, line 29: I merged in the latest code from main and got all these linting changes from the new rule for including/excluding commas. Need to figure out how to fix...

packages/components/src/styles/components/side-nav/main.scss
```diff
@@ -7,7 +7,20 @@
 // SIDE-NAV
 //
 
+:root {
+  // TODO: Move to sidenav tokens
+  --hds-app-sidenav-headerless-width-minimized: 16px;
+
+  // TODO: Create app-header token:
+  --hds-app-header-height: 59px;
```
KristinLBradley, line 15: @didoo Should I create design tokens for these? I assume yes but want to confirm. (Could make them sass variables instead?)

packages/components/src/components/hds/side-nav/index.ts
```diff
@@ -27,6 +27,7 @@ interface HdsSideNavSignature {
     onToggleMinimizedStatus?: (arg: boolean) => void;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     onDesktopViewportChange?: (arg: boolean) => void;
+    hasHeader?: boolean;
```
KristinLBradley, line 30: Does this arg name make sense? This is meant to set whether the `SideNav is being accompanied by the `AppHeader`.

Would `hasAppHeader` or another name be clearer?

packages/components/src/styles/components/app-frame.scss
```diff
@@ -7,6 +7,11 @@
 // APP-FRAME
 //
 
+// TODO: Create app-header tokens:
+:root {
+  --token-app-header-height: 60px;
```
KristinLBradley, line 12: I wanted to confirm what pattern should be followed for token naming before adding tokens.

packages/components/src/styles/components/app-header.scss
```diff
@@ -0,0 +1,117 @@
...
+//
+
+@use "../mixins/interactive-dark-theme" as *;
+
+// ----------------
+
+// TODO: Create app-header tokens:
+:root {
+  --token-app-header-height: 60px;
+  --token-app-header-home-link-padding: 4px;
```
KristinLBradley, line 17: I will add these tokens to JSON but wanted to confirm if naming pattern I used is correct?

packages/components/src/components/hds/side-nav/index.ts
jorytindall: Personally I think that `hasHeader` is more clear because it's specific to the component the argument is being invoked on.

packages/components/src/styles/components/app-header.scss
```diff
@@ -0,0 +1,117 @@
...
+}
+
+.hds-app-header {
+  position: fixed;
+  top: 0;
+  right: 0;
+  left: 0;
+  z-index: 20; // needs to stay above the main content, TODO: determine correct z-index to use
```
KristinLBradley, line 25: Is this z-index ok? I used the same as the SideNav currently uses.

packages/components/src/styles/components/app-header.scss
KristinLBradley: Jory and I discussed this a bit more and we are leaving the width as is for now (at least for phase 1).

packages/components/src/styles/components/side-nav/main.scss
```diff
@@ -7,7 +7,26 @@
...
+  --token-hds-app-header-height: 60px;
+  --token-hds-app-header-home-link-padding: 4px;
+}
+
+:root {
+  // TODO: Create sidenav token
+  --hds-app-sidenav-headerless-width-minimized: 16px;
+
+  // TODO: Create app-header token:
+  --hds-app-header-height: 59px;
```
KristinLBradley, line 21: I will add these tokens to JSON but wanted to confirm if naming pattern I used is correct?

KristinLBradley:
Could use input on the following as well as overall in depth review.

* Is the naming patterns I followed for new component-level design tokens correct? (wanted to confirm before adding them to JSON)
* Any more examples that would be helpful to add to the Showcase?
* Do we need to add a solution for long org names in the OrgSwitcher? @jorytindall   
(Currently extreme edge cases will force the AppHeader to begin to overflow its container.)
See [example in Showcase](https://hds-showcase-git-hds-3373-app-header-component-hashicorp.vercel.app/components/app-header#with-long-content)
image

packages/components/src/styles/components/app-header.scss
```diff
@@ -0,0 +1,117 @@
...
+      height: calc(100vh - var(--token-app-header-height));
+      padding: 16px;
+      overflow: auto; // allow users to scroll within the menu if the content is too long
+      background: var(--token-color-palette-neutral-700);
+    }
+
+    // Force nested dropdown content to align left in mobile/tablet view
+    .hds-dropdown__content {
+      left: 0;
+    }
```
KristinLBradley, line 63: I can either delete the above styles forcing dropdown list position (and await collision detection being added to `Dropdown`) or else add more to better control the list position within all responsive views (which could always be deleted later when no longer needed.)

packages/components/src/components/hds/side-nav/index.ts
KristinLBradley: The `hasHeader` arg is used to alter the height and top position of the SideNav so it's visually aligned together with the fixed AppHeader. The name of the arg could possibly be mistaken to be related to the SideNav :header block is what I'm worried about. (When the SideNav is paired with the AppHeader, it's true that its :header block is meant to be left empty but the `hasHeader` arg is unrelated to that.)

```
.hds-side-nav--has-header {
  top: var(--token-hds-app-header-height);
  height: calc(100vh - var(--token-hds-app-header-height));
  min-height: calc(100vh - var(--token-hds-app-header-height));
}
```

packages/components/src/components/hds/app-header/index.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  removeEventListeners() {
+    this.desktopMQ.removeEventListener(
+      'change',
+      this.updateDesktopVariable,
+      true
+    );
+  }
+
+  /**
```
zamoore, line 62: Can remove

packages/components/src/components/hds/app-header/index.ts
```diff
@@ -0,0 +1,100 @@
...
+   * Generates a unique ID for the Menu Content
+   *
+   * @param menuContentId
+   */
+  menuContentId = 'menu-content-' + guidFor(this);
+
+  /**
```
zamoore, line 69: Can remove

packages/components/src/components/hds/app-header/index.ts
```diff
@@ -0,0 +1,100 @@
...
+   * Get the class names to apply to the component.
+   * @method classNames
+   * @return {string} The "class" attribute to apply to the component.
+   */
+  get classNames() {
```
zamoore, line 74: Can you add a return type to the getters and methods?

zamoore (commented):
Looking really good. Added some stuff around the return types/jsdoc conversation we had last week. I'll finish reviewing more thoroughly tomorrow morning.

jorytindall (commented):
I think this is looking good, the technical side of this is arguably a bit above me, but I did a fair amount of testing and one thing I came across was:

When the viewport is smaller than the breakpoint (1088px) and the SideNav is toggled open, the overlay displays over the main page, but the home link and menu button can still be interacted with (they have interactive states), but they don't function. I imagine it's because focus is trapped in the SideNav (I can't tab out of it), but I can still click or press on these elements.

Is this something we should worry about? Or not a big deal? I think it's unlikely that users will try to interact with these elements when the SideNav is open, but there's currently no visual indication that they don't function. Something I can explore a bit if we want, but wanted to highlight in case it was an issue.

packages/components/src/styles/components/side-nav/main.scss
didoo: While here is `59px` while above is `60px`? 🤔

didoo (commented):
@KristinLBradley I did a first, shallow check, and left a few comments. I need more time to get my head around the implementation (quite complex, with a lot of customizations). I'll continue tomorrow morning with a fresh head.

packages/components/src/components/hds/app-header/index.hbs
KristinLBradley: It's difficult to know what to call the blocks since we don't restrict what's contained in them.

The `:globalItemsBefore` contains only the HomeLink/Logo right now but in phase 2 it could potentially contain the Product menu before the logo (although there was talk that it would maybe be dropped.)

The `:globalItemsAfter` contains only the Org switcher right now but it's not restricted to it. It doesn't seem like it would be used for anything else from the design explorations but I don't know for sure.

Here's a screen shot of the potential content for the "global items" (from an older design):
image

KristinLBradley: I could rename the first block (:globalItemsBefore) to `:logo` (but then what do I do if phase 2 includes the Product menu, add another named block?)

Then I could call the 2nd block (:globalItemsAfter) just `:globalItems`.

@jorytindall do you have any ideas on naming?

packages/components/src/components/hds/side-nav/index.ts
KristinLBradley: @didoo what do you think?

packages/components/src/components/hds/app-header/index.hbs
jorytindall: Hmmmmm, thinking about this a little bit, within the `global` block there are almost primary and secondary elements within that hierarchy. Maybe for now it should simply be `:logo`, I'm not sure about the horizon of the `AppSwitcher`, but it somewhat seems like that idea has been abandoned. Additionally, that design was also somewhat speculative, I'm not sure if we would go the same direction if we needed to implement something like it in the future.

Then the next block could be `:globalItems`, that makes sense to me.

packages/components/src/styles/components/side-nav/main.scss
KristinLBradley: Oops, I will fix, thanks

packages/components/src/styles/components/app-frame.scss
```diff
@@ -24,6 +29,10 @@
 .hds-app-frame__header {
   z-index: 7;
   grid-area: header;
+
+  &:has(.hds-app-header) {
+    min-height: var(--token-app-header-height); // reserve space for fixed header
```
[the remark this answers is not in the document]
KristinLBradley: The AppHeader has a fixed position so this wrapper element collapses. I added the height to this wrapper so that content on the page wouldn't be overlapped by the fixed position AppHeader.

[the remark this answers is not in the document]
KristinLBradley: **With the min-height:** 
_(all main content including breadcrumbs can be seen)_
image

**With min-height commented out:**
_(main content is overlapped by the AppHeader so breadcrumbs are hidden)_
image

packages/components/src/components/hds/app-header/index.hbs
KristinLBradley: Thanks for weighing in. Ok, I think I will go ahead with renaming those blocks to `:logo` & `:globalItems` as that seems to make the most sense.

KristinLBradley:
> I think this is looking good, the technical side of this is arguably a bit above me, but I did a fair amount of testing and one thing I came across was:
> 
> When the viewport is smaller than the breakpoint (1088px) and the SideNav is toggled open, the overlay displays over the main page, but the home link and menu button can still be interacted with (they have interactive states), but they don't function. I imagine it's because focus is trapped in the SideNav (I can't tab out of it), but I can still click or press on these elements.
> 
> Is this something we should worry about? Or not a big deal? I think it's unlikely that users will try to interact with these elements when the SideNav is open, but there's currently no visual indication that they don't function. Something I can explore a bit if we want, but wanted to highlight in case it was an issue.

@jorytindall The appHeader and SideNav don't know about each other so it seems it would be difficult to change the behavior of the AppHeader menu button dependent upon the state of the SideNav.

@didoo Do you have any ideas around this?

packages/components/src/styles/components/side-nav/main.scss
```diff
@@ -7,7 +7,23 @@
...
+
+:root {
+  // TODO: Create sidenav token
+  --hds-app-sidenav-headerless-width-minimized: 16px;
+}
+
 .hds-side-nav {
+  // set narrower collapsed width when the header is empty
+  &:has(.hds-side-nav__wrapper-header:empty) {
+    --hds-app-sidenav-width-minimized: var(--hds-app-sidenav-headerless-width-minimized);
```
KristinLBradley, line 24: @didoo Should I create a new JSON token for `--hds-app-sidenav-headerless-width-minimized`?

Or should I just set the value directly here?
```
--hds-app-sidenav-width-minimized: 16px;
```

KristinLBradley:
@MelSumner Could you please review the AppHeader implementation for accessibility?

packages/components/src/components/hds/app-header/index.ts
```diff
@@ -0,0 +1,100 @@
...
+import { tracked } from '@glimmer/tracking';
+import { registerDestructor } from '@ember/destroyable';
+
+export interface HdsAppHeaderSignature {
+  Blocks: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    logo?: any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    globalActions?: any;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
shleewhite, line 18: Question: sorry, not a11y related - but why are these typed as any?

KristinLBradley: Because we aren't restricting what content can go in them.

packages/components/src/styles/components/app-frame.scss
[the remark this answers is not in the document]
KristinLBradley: Replaced with adding `position: sticky` to AppFrame Header

packages/components/src/components/hds/app-header/index.hbs
```diff
@@ -0,0 +1,34 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<div class={{this.classNames}} ...attributes>
```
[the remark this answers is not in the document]
KristinLBradley: The AppHeader component is used within the AppFrame Frame.Header which is a `header` element.

https://helios.hashicorp.design/layouts/app-frame

packages/components/src/components/hds/app-header/index.ts
KristinLBradley: One is typed as an array the other isn't. Do you have a suggestion about what it should be?

Anyway I'm not a Typescript expert. We only started using TypeScript recently on the team and I'd only been using it for a short time before joining...

packages/components/src/components/hds/app-header/menu-button.ts
```diff
@@ -0,0 +1,34 @@
...
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+
+import type { HdsButtonSignature } from '../button/';
+
+export interface HdsAppHeaderMenuButtonSignature {
+  Args: {
+    isOpen?: boolean;
+    menuContentId: string;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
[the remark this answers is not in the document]
KristinLBradley: Thanks I think that is left over. Will remove.

packages/components/src/components/hds/app-header/index.ts
shleewhite: Hahah I think this is where me being new to ember is confusing me.. so feel free to ignore! 

I just noticed in some components where it seems like we allow any content to be passed, it used the array syntax (example: https://github.com/hashicorp/design-system/blob/e4cf01addb3c650a2e1871e575db2d7bf27e7b3e/packages/components/src/components/hds/page-header/title.ts#L12)

shleewhite (approved):

packages/components/src/components/hds/app-header/index.ts
KristinLBradley: I mean, an empty array could very well be better to use.

Searching our code now, I'm seeing several examples of that so I'll see if I can change it and then remove the eslint disable comment which would be good.

KristinLBradley: That ended up working, thanks! (Sorry I just didn't understand the context of your original question since I'm new to TypeScript.)

packages/components/src/styles/components/side-nav/main.scss
KristinLBradley: I decided to set the value directly for phase 1 at least. (Can reconsider in phase 2)

packages/components/src/components/hds/app-header/index.ts
shleewhite: You're good! I should have added more context.

packages/components/src/components/hds/app-header/menu-button.hbs
```diff
@@ -0,0 +1,15 @@
...
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<Hds::Button
+  class="hds-app-header__menu-button"
+  @text="Menu"
+  @icon={{this.icon}}
+  @iconPosition="trailing"
+  {{on "click" this.onClick}}
+  aria-controls={{@menuContentId}}
```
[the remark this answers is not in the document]
KristinLBradley: I'll look at this more, thanks

packages/components/src/components/hds/side-nav/index.ts
KristinLBradley: I'm going to change the name of this parameter to `@withAppHeader`. I think that will make it's purpose & usage clearer without being too long.

packages/components/src/components/hds/app-header/menu-button.hbs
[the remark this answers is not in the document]
KristinLBradley: I was looking at our `Dropdown` component to compare how it works and it doesn't use aria-controls, it just has aria-expanded=true/false dependent on whether the associated menu is being shown. So maybe I don't need to use aria-controls at all?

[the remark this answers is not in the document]
KristinLBradley: @MelSumner how do you think we should handle this?

packages/components/src/components/hds/app-header/index.ts
```diff
@@ -0,0 +1,102 @@
...
+export default class HdsAppHeaderComponent extends Component<HdsAppHeaderSignature> {
+  @tracked isOpen = false;
+  @tracked isDesktop = true;
+  desktopMQ: MediaQueryList;
+
+  desktopMQVal =
+    this.args.breakpoint ||
+    getComputedStyle(document.documentElement).getPropertyValue(
+      '--hds-app-desktop-breakpoint'
+    );
```
KristinLBradley, line 33: Consumers can set a width value for `@breakpoint` to optionally override the preset value if they wish. Let me know if this makes sense to do this way... @didoo

packages/tokens/src/products/shared/app-header.json
```diff
@@ -0,0 +1,18 @@
+{
+  "app-header": {
+    "height": {
+      "value": "60",
+      "type": "size"
+    },
+    "home-link": {
+      "padding": {
```
didoo, line 8: Consider an alternative approach: using two variables, one for the overall `home-link` size and one for the `logo` size, and infer the internal padding from these two.

didoo: btw, I don't see the `--token-app-header-home-link-logo-size` used anywhere? maybe it's still a WIP?

packages/components/src/components/hds/app-header/index.ts
didoo: I think that for the V1 version is OK, but I also think that for the V2 version the whole approach should be revisited: as discussed together, we should try to see if it's possible to move some of the viewport-related logic up to the `AppFrame` level (so it's done once, for both the `AppHeader` and the `SideNav`) and improve the resilience of the `AppHeader` (in terms of responsiveness) using `ResizeObserver`.

The reason why I am saying this is that the way in which we're doing the media querying is based on assumptions about a fixed breaking point (set by us or the consumers) but there are many factors that could make the content overflow or wrap: a different font used by the consumer, an extra long text in the org/project picker, extra content added by the consumers (with potentially variable length) after the org/project picker, etc. All of these variables make my "spider sense" tickle when I think of a fixed value as breaking point between the two modes.

As mentioned during our previous conversation, I don't have an "easy" fix or alternative, so considering the timeline we have, it's OK for the V1 release: it will take some time before the component is actually adopted and used in production, and we can use this time to come up with a more robust approach for the V2 that involves the three different pieces of this puzzle: `AppFrame`, `AppHeader` and `SideNav`.

/cc @alex-ju @Dhaulagiri

packages/components/src/components/hds/app-header/index.hbs
```diff
@@ -0,0 +1,28 @@
...
+
+  {{#if (not this.isDesktop)}}
+    <Hds::AppHeader::MenuButton
+      @onClickToggle={{this.onClickToggle}}
+      @isOpen={{this.isOpen}}
+      @menuContentId={{this.menuContentId}}
+    />
+  {{/if}}
+
+  {{#if this.showItems}}
```
didoo, line 17: I was thinking: is there maybe a way to "hide" the items when in "mobile" view (eg. using max-height + overflow hidden?) instead of removing them from the DOM? This _may_ also "solve" the problem of the `aria-controls` for the `MenuButton` below, which at this point could simply be hidden to screen readers (which have always access to the entire menu actions, it's only that they change visual presentation for the two modes, desktop and mobile)

didoo (commented):
@KristinLBradley answered to your question and left other high-level comments. Let me know if there is something you want to discuss more in detail, or rubberduck ideas/solutions.

MelSumner:
Testing the keyboard nav on the preview pages; wondering if this is expected:

CleanShot 2024-07-23 at 10 10 46@2x

MelSumner:
I'm seeing some color contrast issues here: https://hds-showcase-git-hds-3373-app-header-component-hashicorp.vercel.app/components/app-header#dropdowns-within-appheader 

Could we get the designer on this component to double check the combinations?

packages/components/src/components/hds/app-header/index.hbs
KristinLBradley: I think we could, I was following the pattern of our `Dropdown` component which actually removes the content when it's not displayed.

@MelSumner what would be the best practice in terms of accessibility?

packages/components/src/components/hds/app-header/index.ts
KristinLBradley: Ok thank you. I will go ahead and add examples to the Showcase and related tests. Then we can revisit in phase 2 to look towards a more robust solution.

packages/components/src/styles/components/app-frame.scss
```diff
@@ -24,6 +24,14 @@
 .hds-app-frame__header {
   z-index: 7;
   grid-area: header;
+
+  // NOTE: May not need to qualify this with .hds-app-header as it may not matter if AppHeader is not present
+  &:has(.hds-app-header) {
+    position: sticky;
```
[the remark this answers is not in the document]
KristinLBradley: Should I add an option to turn off the fixed behavior to the AppFrame itself then (vs the AppHeader)? (I guess maybe I'd have to...)

packages/tokens/src/products/shared/app-header.json
KristinLBradley: Ok I'll see. I think I forgot to add the property for the logo-size in the styles. I will update. (I was trying to follow some of the pattern from the SideNav HomeLink)

packages/components/src/styles/components/app-frame.scss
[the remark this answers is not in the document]
didoo: > turn off the fixed behavior to the AppFrame itself

Which fixed behaviour? 🤔

packages/components/src/components/hds/app-header/menu-button.hbs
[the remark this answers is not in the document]
MelSumner: @KristinLBradley So @shleewhite is accurate for typical guidance in JS frameworks at large but the specific approach guidance is a little different. We've chosen to keep the container that is referenced in the DOM and then insert content into that. See this demo CodePen: https://codepen.io/melsumner/pen/qBzNgWO?editors=1100

packages/components/src/components/hds/app-header/index.hbs
MelSumner: I don't recommend having duplicate content in the browser at the same time, even if some is hidden. 

- hiding interactive elements is something we've indicated folks should avoid (we should follow our own guidance)
- some browsers will expose interactive elements in the DOM and ignore anything that indicates that it shouldn't be exposed (_because_ it is an interactive element); since there are no agreed upon rules or specification for exemptions like this, they are implemented by browsers at will (and inconsistently)

Recommendation:

- Wherever possible, keep the markup the same and style for different viewports with media queries in CSS (I thought we decided on a "mobile-first" CSS approach but that is also my recommendation, I've found it easier to maintain in the long run)
- Render the correct element content to the DOM based on context

packages/components/src/components/hds/app-header/menu-button.hbs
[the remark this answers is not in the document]
KristinLBradley: Thanks, I refactored it to add an outer wrapper around the menu content with the associated id which remains in the DOM.

packages/components/src/components/hds/app-header/index.ts
KristinLBradley: I just called the new argument "breakpoint" btw, I don't know if we'd want to come up with some other name.

KristinLBradley: I added a Showcase example and an integration test.

packages/components/src/styles/components/app-frame.scss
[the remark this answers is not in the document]
KristinLBradley: I meant the "sticky" layout of the header which you said we could add an `@isHeaderFixed` option for.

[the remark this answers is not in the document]
KristinLBradley: I took a look at this, if I add an `@isHeaderFixed` I'm going to have to refactor how the SideNav is positioned as well. That's fine, but I'd rather do this in phase 2, if that's ok.

packages/tokens/src/products/shared/app-header.json
KristinLBradley: Alright, I refactored as you suggested.

packages/components/src/components/hds/app-header/index.hbs
KristinLBradley: I have since updated this to keep the container of the dropdown items in the DOM but remove the content from the DOM when it is hidden/shown as I thought you had recommended.

There isn't duplicate content currently. CSS is used to style the AppHeader content as a dropdown menu in "mobile" view vs. laying the content out within the AppHeader bar in "desktop" view.

KristinLBradley:
> Testing the keyboard nav on the preview pages; wondering if this is expected:
> 
> CleanShot 2024-07-23 at 10 10 46@2x

@MelSumner Yes because the Showcase examples can't really show the full in page experience. When the AppHeader is used within a page, there is only ever one and the open menu expands to fill the full screen. In the Showcase the menu content displays at auto height. I could remove that Showcase only style so that the menus display instead at full screen height though if that would be better. Although it's still not going to be quite the same as the actual behavior.

didoo:
> > Testing the keyboard nav on the preview pages; wondering if this is expected:
> > CleanShot 2024-07-23 at 10 10 46@2x
> 
> @MelSumner Yes, because the Showcase examples can't really show the full in-page experience. When the `AppHeader` is used within a page, there is only ever one used and the open menu expands to fill the full screen. In the Showcase, the menu content displays at auto height instead. I could remove that Showcase-only style so that the menus display instead at full screen height though if that would be better. Although it's still not going to be quite the same as the actual behavior.

@MelSumner @KristinLBradley there is an icon on the top of the `Doc::Frame` container to open the content of the iframe (the "frameless", standalone page) in a new window. that's how it's supposed to be tested (in an iFrame there are always limitations)

packages/components/src/styles/components/app-frame.scss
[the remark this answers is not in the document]
didoo: Yes, that's what I meant: make the `AppFrame` responsible for positioning and sizing, but in V2.

KristinLBradley:
> I'm seeing some color contrast issues here: https://hds-showcase-git-hds-3373-app-header-component-hashicorp.vercel.app/components/app-header#dropdowns-within-appheader
> 
> Could we get the designer on this component to double check the combinations?

@MelSumner Could you call out the specific issues? If you mean the disabled Dropdown & Button, my understanding is that disabled controls do not have to meet a11y contrast standards. We also don't expect the disabled controls to be used in practice. Styles are just included for those for completeness.

@jorytindall could you double-check?

didoo (approved):
For Phase 1 I think what we have is good enough. Up to you to decide when you're happy to merge

KristinLBradley:
> I think this is looking good, I'll address a couple of things:
> 
> @MelSumner: regarding the color contrast, are you referring to the disabled dropdown menus?
> 
> * To echo @KristinLBradley 's point these have different acceptance criteria, right?
> * We don't actually intend consumers to use these (more difficult to prevent/block them) in this context, I don't see a scenario where they would. The same thing is "supported" in Figma, but there's quite a bit of friction for a consumer to override this.
> 
> Regarding the Frameless demo, thanks for spinning that up, very helpful to see it in context! However, I was clicking through things and noticed that when the SideNav is open (on desktop) the dropdowns don't function as intended (aren't clickable). Is this a result of the change preventing the menu from being accessed when the SideNav is open on mobile? I've attached a video, I'm hoping that my comment wasn't misconstrued. Happy to sync if you need more details.
> 
>  CleanShot.2024-07-24.at.13.56.28.mp4 
> Happy to give this the green light after that confirmation, but just want to make sure

@jorytindall 
This was a bug (interactive content not working when SideNav expanded in desktop view). I've just pushed up a fix.

jorytindall (approved):
Thanks for the quick fix! :shipit:

packages/components/src/styles/components/app-header.scss
KristinLBradley: UPDATE: I deleted styles forcing Dropdown list positioning in anticipation of @alex-ju 's updates to the Dropdown.

packages/components/src/components/hds/app-header/index.hbs
KristinLBradley: I'm going to look at hiding menu content using CSS vs. rendering/not rendering in phase 1 1/2

PR #2162, opened 2024-06-13
Add Icon Library tracking events

zamoore:
#### Added tracking events for:

- Searching for icons
- Selecting an icon size

https://hashicorp.atlassian.net/browse/HDS-3382


website/docs/icons/library/index.js
```diff
@@ -127,24 +128,36 @@ export default class Index extends Component {
...
     this.router.transitionTo({
       queryParams: {
-        searchQuery: searchQuery !== '' ? searchQuery : null,
+        searchQuery,
         selectedGroupType: this.selectedGroupType,
         selectedIconSize: this.selectedIconSize,
       },
     });
+
+    if (searchQuery !== null) {
```
zamoore, line 159: @Dhaulagiri 

Do we want to track the search term being cleared? I assumed not, but can revert this if needed.

Dhaulagiri:
I'd be happy to pair on this, but for this PR I'd like to see the commit history be cleaner and only include the commits for the changes needed for this ticket. That will/should make it easier for the reviewer to understand what has changed here and why.

website/docs/icons/library/index.js
Dhaulagiri: I think that's fine not to log clearing the search. I do think we need to figure out a way to debounce or otherwise delay the event tracking more as 250ms is pretty easy to get into a situation where we are sending multiple events for the same search which is not really what we want.

website/app/services/event-tracking.js
```diff
@@ -0,0 +1,30 @@
...
+    super(...arguments);
+
+    // Only attempt to do something if we are in the right environment
+    this.isEnabled = !this.fastboot.isFastBoot && window.fathom;
+  }
+
+  @action
+  trackEvent(eventName) {
+    if (!this.isEnabled) {
+      return;
```
didoo, line 25:
```suggestion
      // comment this line if you want to test in your local environment
      return;
```

website/docs/icons/library/index.js
```diff
@@ -127,24 +128,36 @@ export default class Index extends Component {
...
       queryParams: {
+        selectedIconSize,
         searchQuery: this.searchQuery,
         selectedGroupType: this.selectedGroupType,
-        selectedIconSize: event.target.value,
       },
     });
+
+    this.eventTracking.trackEvent(
+      `Icon Library - Size Selector - ${selectedIconSize}`
```
didoo, line 142: @zamoore @Dhaulagiri should we add tracking also for the group selector?

didoo (commented):
Left a couple of comments

MelSumner (approved):
I'd probably squash/fixup the commits and re-push (`--force-with-lease`) but I don't see anything wrong with the code itself. 👍

website/docs/icons/library/index.js
Dhaulagiri: Yes. I think that was added after we had outlined the events to capture in the doc Zack was working from.

website/docs/icons/library/index.js
zamoore: We could have a separately debounced function. Let me play around with it.

Dhaulagiri: 👍  let me know what you find. I think ultimately I'd rather have no data than an overwhelming amount of partially/hopefully useful data.

zamoore:
Resubmitting in a clean branch

PR #2163, opened 2024-06-13
Adds `aria-describedby` to the icon filter input

MelSumner:
### :pushpin: Summary

If merged, this PR adds `aria-describedby` to the icon input filter that points to a message for users with screen readers to inform them that the icon list will automatically update as they type.  I noticed it was missing while reviewing #2162 

- added `aria-describedby` attribute to the component invocation
- added an accompanying non-visible message for users with a screen reader

There is no visible change to the page, only the DOM:
CleanShot 2024-06-13 at 10 02 12

***

### 👀 Component checklist

### DIFF
--- a/website/app/components/doc/icons-list/index.hbs
+++ b/website/app/components/doc/icons-list/index.hbs
@@ -10,7 +10,9 @@
     @filterQuery={{@searchQuery}}
     @onInput={{@searchIcons}}
     data-test="icons-filter"
+    aria-describedby="icons-filter-description"
   />
+  <span id="icons-filter-description" class="sr-only">Note: Icon list will automatically update as you type.</span>
   <Doc::Form::SelectGroupType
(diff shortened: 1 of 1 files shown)


website/app/components/doc/icons-list/index.hbs
```diff
@@ -10,7 +10,9 @@
     @filterQuery={{@searchQuery}}
     @onInput={{@searchIcons}}
     data-test="icons-filter"
+    aria-describedby="icons-filter-description"
   />
+  <span id="icons-filter-description" class="sr-only">Note: Icon list will automatically update as you type.</span>
```
MelSumner, line 15: Another viable option would be to place this inside of the `Doc::Form::Filter` component itself; the outcome will be the same either way.

zamoore (approved):
Nice addition!

PR #2164, opened 2024-06-13
Added conditional statement to icon list grid for h2

MelSumner:
### :pushpin: Summary

If merged, this PR adds a conditional in the icon grid list template so that the h2 is only rendered to the page if there is content for it.  

Otherwise, the headings rotor list just says "heading level two" with no text in it. This PR resolves that issue.

There is no visible change in the browser.

***

### 👀 Component checklist


alex-ju (approved):
Good spot!

zamoore (approved):

PR #2165, opened 2024-06-13
Fix related component event

zamoore:
### :pushpin: Summary

The eventName for the related component event wasn't concatenating correctly.


didoo (approved):
Haven't tested locally, but I assume you did. Code seems OK 👍

PR #2166, opened 2024-06-13
Update ember apps/addons to 5.9 blueprint

Dhaulagiri:
### :pushpin: Summary

This PR runs `npx ember-cli-update` on all our ember things to keep them in sync. All changes should be internal to us and not directly impact any consumers of HDS.

As part of this I tried to excise the blueprint-specific changes from https://github.com/hashicorp/design-system/pull/2142 so that were not related to TS so that a) this package can be relatively consistent with the others and b) to simplify the actual TypeScript diff for that work so we can cleanly review and release it.

Note: There were a number of prettier package updates prescribed as part of this, but I undid them in favor of scheduling that work in a [separate task](https://hashicorp.atlassian.net/browse/HDS-3383).


packages/components/package.json
```diff
@@ -72,27 +72,27 @@
...
-    "eslint-plugin-ember": "^11.11.1",
+    "eslint": "^8.57.0",
+    "eslint-config-prettier": "^9.1.0",
+    "eslint-plugin-ember": "^11.12.0",
     "eslint-plugin-import": "^2.29.1",
     "eslint-plugin-n": "^17.7.0",
     "eslint-plugin-prettier": "^4.2.1",
-    "eslint-plugin-qunit": "^8.0.1",
+    "eslint-plugin-qunit": "^8.1.1",
     "npm-run-all": "^4.1.5",
```
alex-ju, line 96: I believe we can now remove `npm-run-all` as we rely on `concurrently` to run commands

packages/ember-flight-icons/package.json
```diff
@@ -33,24 +33,26 @@
...
-    "eslint-config-prettier": "^8.10.0",
-    "eslint-plugin-ember": "^11.11.1",
+    "eslint": "^8.57.0",
+    "eslint-config-prettier": "^9.1.0",
+    "eslint-plugin-ember": "^11.12.0",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^4.2.1",
-    "eslint-plugin-qunit": "^8.0.1",
+    "eslint-plugin-qunit": "^8.1.1",
     "npm-run-all": "^4.1.5",
```
alex-ju, line 56: same here, regarding `npm-run-all`

alex-ju (approved):
Thanks for taking the time to do this 🙌 ! Left comments to remove `npm-run-all` as a dependency, otherwise good to go!

packages/ember-flight-icons/package.json
Dhaulagiri: Good catch. I removed it and added that to to the initial commit.

showcase/tests/helpers/index.ts
```diff
@@ -7,10 +7,9 @@ import {
   setupApplicationTest as upstreamSetupApplicationTest,
   setupRenderingTest as upstreamSetupRenderingTest,
   setupTest as upstreamSetupTest,
+  type SetupTestOptions,
 } from 'ember-qunit';
 
-import type { SetupTestOptions } from 'ember-qunit';
```
aklkv, line 12: Eve thought ember doesn't complain about this we should keep type imports in separate line for consistency and in future it might start complaining about this type of imports.

PR #2167, opened 2024-06-14
Tracking events for icon library

zamoore:
### :pushpin: Summary

If merged, this PR will add tracking events to the icon library page.

- When a user searches (debounced by 1s), we track `Icon Library - Search - SEARCH_QUERY`
- When a user selects icon size, we track `Icon Library - Size Selector - SELECTED_SIZE`
- When a user selects how they want to group icons, we track `Icon Library - Group by Selector - GROUP_BY`

### :link: External links

https://hashicorp.atlassian.net/browse/HDS-3382

### DIFF
--- a/website/app/modifiers/doc-track-event.js
+++ b/website/app/modifiers/doc-track-event.js
@@ -9,14 +9,10 @@ import { inject as service } from '@ember/service';
 import { assert } from '@ember/debug';
 
 export default class DocTrackEvent extends Modifier {
-  @service fastboot;
+  @service eventTracking;
 
   modify(element, _positional, named) {
-    if (
-      // Only attempt to do something if we are in the right environment
-      this.fastboot.isFastBoot ||
-      window.fathom == null
-    ) {
+    if (!this.eventTracking.isEnabled) {
       // comment this line if you want to test in your local environment
       return;
     }
@@ -29,8 +25,7 @@ export default class DocTrackEvent extends Modifier {
     );
 
     const handleTriggerEvent = () => {
-      // https://usefathom.com/docs/features/events
-      window.fathom?.trackEvent(eventName);
+      this.eventTracking.trackEvent(eventName);
     };
 
     element.addEventListener(triggerEvent, handleTriggerEvent);
--- a/website/app/modifiers/doc-track-event.js
+++ b/website/app/modifiers/doc-track-event.js
@@ -12,9 +12,14 @@ export default class DocTrackEvent extends Modifier {
   @service eventTracking;
 
   modify(element, _positional, named) {
+<<<<<<< HEAD
     // if the tracking is disabled, do not add the event listener
     if (!this.eventTracking.isEnabled) {
       // comment this line if you want the tracking function in the `eventTracking` service to be called even if the tracking is disabled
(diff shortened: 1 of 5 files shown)


Dhaulagiri (approved):
Nice on! Let's keep an eye on the events next week to see if the icon search gets too noisy. (I can show you how to see them if we didn't talk about that already)

website/app/modifiers/doc-track-event.js
```diff
@@ -9,14 +9,10 @@ import { inject as service } from '@ember/service';
...
+  @service eventTracking;
 
   modify(element, _positional, named) {
-    if (
-      // Only attempt to do something if we are in the right environment
-      this.fastboot.isFastBoot ||
-      window.fathom == null
-    ) {
+    if (!this.eventTracking.isEnabled) {
       // comment this line if you want to test in your local environment
```
Dhaulagiri, line 16: [nit] I think this was added via suggestion in a previous PR, but trying this out I don't see the events actually working.

didoo: 🤔

website/app/modifiers/doc-track-event.js
```diff
@@ -9,14 +9,10 @@ import { inject as service } from '@ember/service';
...
-  @service fastboot;
+  @service eventTracking;
 
   modify(element, _positional, named) {
-    if (
-      // Only attempt to do something if we are in the right environment
-      this.fastboot.isFastBoot ||
-      window.fathom == null
-    ) {
+    if (!this.eventTracking.isEnabled) {
```
Dhaulagiri, line 15: Looking at this with fresher eyes I'm not sure this entire `if` block is actually necessary as the same check now happens [here](https://github.com/hashicorp/design-system/pull/2167/files#diff-b5adddb0d69ebd2ee22d67a269141428dda6e318176c851ef8158f9fd700cbeeR18) as well

website/app/modifiers/doc-track-event.js
Dhaulagiri: Are you adding something to actually log the string or how are you seeing the event?

zamoore: AFAIK you would need to comment it out as well as adding a log.

website/app/modifiers/doc-track-event.js
zamoore: It's not strictly necessary, but it prevents the event listener from being added unnecessarily

website/app/modifiers/doc-track-event.js
didoo: Yes, as @zamoore says, I'm commenting out the call to Fathom and using a simple `console.log` instead.

website/app/modifiers/doc-track-event.js
didoo: +1 on @zamoore approach

website/app/modifiers/doc-track-event.js
Dhaulagiri: Ok. I think I was reading this too literally then as the _only_ thing you needed to do.

PR #2168, opened 2024-06-14
Tabs TS Conversion

zamoore:
### :pushpin: Summary

If merged, this PR converts the `Tabs` component and its child components into TypeScript. 

### :link: External links

https://hashicorp.atlassian.net/browse/HDS-2714


packages/components/src/components/hds/tabs/index.ts
```diff
@@ -131,33 +163,33 @@ export default class HdsTabsIndexComponent extends Component {
...
   @action
-  willDestroyTab(element) {
+  willDestroyTab(element: HTMLButtonElement) {
     this.tabNodes = this.tabNodes.filter((node) => node.id !== element.id);
     this.tabIds = this.tabIds.filter((tabId) => tabId !== element.id);
   }
 
   @action
-  didInsertPanel(element, panelId) {
+  didInsertPanel(element: HTMLElement, panelId: string) {
```
didoo, line 180: no biggie, but do you think it would make sense to have this signature come from the `HdsTabsPanelSignature['Args']['didInsertNode']` somehow? @alex-ju thoughts? I'm overthinking it?

(similar for other functions here)

packages/components/src/components/hds/tabs/index.ts
```diff
@@ -168,52 +200,52 @@ export default class HdsTabsIndexComponent extends Component {
...
-    const enterKey = 13;
-    const spaceKey = 32;
+  onKeyUp(tabIndex: number, event: KeyboardEvent) {
+    const leftArrow = 'ArrowLeft';
+    const rightArrow = 'ArrowRight';
+    const enterKey = 'Enter';
+    const spaceKey = ' ';
 
-    if (event.keyCode === rightArrow) {
+    if (event.key === rightArrow) {
```
didoo, line 209: TIL: `event.keyCode` is deprecated! 😮

packages/components/src/components/hds/tabs/index.ts
```diff
@@ -226,7 +258,7 @@ export default class HdsTabsIndexComponent extends Component {
           );
         }
       } else {
-        let message;
+        let message = '';
```
didoo, line 261: oh, we missed this one, nice catch!

packages/components/src/components/hds/tabs/panel.ts
```diff
@@ -6,14 +6,35 @@
...
 import { guidFor } from '@ember/object/internals';
 import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
 
-export default class HdsTabsPanelComponent extends Component {
+export interface HdsTabsPanelSignature {
+  Args: {
+    tabIds: string[];
+    panelIds: string[];
+    selectedTabIndex: number;
```
didoo, line 15: maybe?
```suggestion
    selectedTabIndex: HdsTabsIndexSignature['Args']['selectedTabIndex'];
```

packages/components/src/components/hds/tabs/index.ts
```diff
@@ -8,18 +8,40 @@ import { tracked } from '@glimmer/tracking';
...
+    ];
+  };
+  Element: HTMLDivElement;
+}
+
+export default class HdsTabsIndexComponent extends Component<HdsTabsIndexSignature> {
+  @tracked tabNodes: HTMLButtonElement[] = [];
+  @tracked tabIds: string[] = [];
+  @tracked panelNodes: HTMLElement[] = [];
+  @tracked panelIds: string[] = [];
```
didoo, line 41: maybe?
```suggestion
  @tracked panelIds: HdsTabsPanelSignature['Args']['panelId'] = [];
```

didoo (commented):
@zamoore I've left here and there some comments, because _maybe_ there could be a bit more type connections betweens `tabs`, `tab` and `panel` elements. But before making further changes let's see what @alex-ju says, maybe I'm overthinking it.

alex-ju (commented):
I left a few comments that would bring the component more in line with others in terms of TypeScript

packages/components/src/components/hds/tabs/index.ts
zamoore: What do you think instead about defining a type like

```
type PanelIdList = string[]

...

@tracked panelIds: PanelIdList = [];
```

And then importing that type in panels and using it?

It feels weird to me to have a parent property type defined by a child component attribute.



````

File: packages/components/src/components/hds/tabs/index.ts [hunk 6]
```diff
@@ -42,15 +64,25 @@ export default class HdsTabsIndexComponent extends Component {
     return size;
   }
 
-  constructor() {
-    super(...arguments);
+  constructor(
+    owner: HdsTabsIndexComponent,
```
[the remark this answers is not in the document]
zamoore: It appears that I was wrong. We did this in a codebase I worked on in the past, but it looks like `unknown` is the correct approach.

packages/components/src/components/hds/tabs/index.ts
```diff
@@ -61,7 +93,7 @@ export default class HdsTabsIndexComponent extends Component {
     if (this.isControlled) {
       // noop
     } else {
-      this._selectedTabIndex = value;
+      this._selectedTabIndex = value ?? 0;
```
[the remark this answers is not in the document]
zamoore: I actually don't think it's necessary anymore after adding the assertion to the getter.

packages/components/src/components/hds/tabs/index.ts
zamoore: I'd rather define shared types from more of a top-down approach if that makes sense. Curious for @alex-ju thoughts as well.

packages/components/src/components/hds/tabs/panel.ts
zamoore: This makes sense to me.

packages/components/src/components/hds/tabs/panel.ts
```diff
@@ -6,14 +6,35 @@
...
+}
+
+export default class HdsTabsPanelComponent extends Component<HdsTabsPanelSignature> {
   /**
    * Generate a unique ID for the Panel
    * @return {string}
    */
   panelId = 'panel-' + guidFor(this);
 
+  @tracked elementId: string | null = null;
```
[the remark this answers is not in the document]
zamoore: Removed. And sure thing.

packages/components/src/components/hds/tabs/index.ts
alex-ju: > I'd rather define shared types from more of a top-down approach if that makes sense.

That sounds good to me and keeps our types less coupled

packages/components/src/components/hds/tabs/tab.ts
```diff
@@ -0,0 +1,122 @@
...
+    const { willDestroyNode } = this.args;
+
+    if (typeof willDestroyNode === 'function') {
+      willDestroyNode(element);
+    }
+  }
+
+  @action
+  onClick(event: MouseEvent) {
+    const { nodeIndex } = this;
```
didoo, line 86: in the original `.js` file this was not destructured and we simply used `this.nodeIndex` in the code below. I have tried to revert this change and run `yarn lint:types` and the tests passed. 🤔

packages/components/src/components/hds/tabs/tab.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (typeof onClick === 'function' && nodeIndex !== undefined) {
+      onClick(event, nodeIndex);
+    } else {
+      return false;
+    }
+  }
+
+  @action
+  onKeyUp(event: KeyboardEvent) {
+    const { nodeIndex } = this;
```
didoo, line 98: see comment above

packages/components/src/template-registry.ts
```diff
@@ -62,6 +62,9 @@ import type HdsSideNavListLinkComponent from './components/hds/side-nav/list/lin
 import type HdsSideNavListTitleComponent from './components/hds/side-nav/list/title';
 import type HdsStepperStepIndicatorComponent from './components/hds/stepper/step/indicator';
 import type HdsStepperTaskIndicatorComponent from './components/hds/stepper/task/indicator';
+import type HdsTabsComponent from './components/hds/tabs';
+import type HdsTabsPanelComponent from './components/hds/tabs/panel';
+import type HdsTabsTabComponent from './components/hds/tabs/tab';
```
didoo, line 67: nit
```suggestion
import type HdsTabsTabComponent from './components/hds/tabs/tab';
import type HdsTabsPanelComponent from './components/hds/tabs/panel';
```

packages/components/src/template-registry.ts
```diff
@@ -289,6 +292,16 @@ export default interface HdsComponentsRegistry {
...
 
+  // Tabs
+  'Hds::Tabs': typeof HdsTabsComponent;
+  'hds/tabs': typeof HdsTabsComponent;
+
+  'Hds::Tabs::Panel': typeof HdsTabsPanelComponent;
+  'hds/tabs/panel': typeof HdsTabsPanelComponent;
+
+  'Hds::Tabs::Tab': typeof HdsTabsTabComponent;
+  'hds/tabs/tab': typeof HdsTabsTabComponent;
```
didoo, line 303:
```suggestion
  'Hds::Tabs::Tab': typeof HdsTabsTabComponent;
  'hds/tabs/tab': typeof HdsTabsTabComponent;

  'Hds::Tabs::Panel': typeof HdsTabsPanelComponent;
  'hds/tabs/panel': typeof HdsTabsPanelComponent;
```

didoo (commented):
Added more comments

packages/components/src/components/hds/tabs/index.ts
```diff
@@ -8,18 +8,44 @@ import { tracked } from '@glimmer/tracking';
...
+import type { ComponentLike } from '@glint/template';
+import type { HdsTabsTabSignature } from './tab';
+import type { HdsTabsPanelSignature } from './panel';
+import type { HdsTabsPanelIds, HdsTabsTabIds } from './types.ts';
+
+export const DEFAULT_SIZE = HdsTabsSizeValues.Medium;
+export const SIZES: HdsTabsSizeValues[] = [
+  HdsTabsSizeValues.Medium,
+  HdsTabsSizeValues.Large,
+];
```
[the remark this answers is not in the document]
alex-ju: Plus one on this, for consistency across the codebase.

packages/components/src/components/hds/tabs/index.ts
```diff
@@ -29,8 +55,8 @@ export default class HdsTabsIndexComponent extends Component {
    * @type {string}
    * @default 'medium'
    */
-  get size() {
-    let { size = DEFAULT_SIZE } = this.args;
+  get size(): HdsTabsSizeValues {
```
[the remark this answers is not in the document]
alex-ju: we sometimes do (it's one of the few inconsistencies I discovered lately); it's generally good practice to explicitly define them, especially for more complex/predefined returned types, although they can be inferred.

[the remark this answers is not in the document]
didoo: @alex-ju what do you think of leaving out them for now, and open a ticket to tackle this systematically later? like a second pass once all the components have been migrated?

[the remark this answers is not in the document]
alex-ju: sure that works for me. I believe there are only a few instances to be cleaned-up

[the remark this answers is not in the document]
alex-ju: ah, I got your suggestion the other way around 🤦 
indeed, it is recommended to have them defined explicitly, so we can add them to the components that are missing in a follow up PR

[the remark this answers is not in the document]
didoo: > it is recommended to have them defined explicitly

Ok, so @zamoore should leave them in his PR, right?

Also, should we update the wiki page about the migration, to clarify this point? (or at least make it explicit)?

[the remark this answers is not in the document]
alex-ju: sure, I can do that

packages/components/src/components/hds/tabs/index.ts
```diff
@@ -8,18 +8,44 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { assert, warn } from '@ember/debug';
 import { next, schedule } from '@ember/runloop';
+import { HdsTabsSizeValues } from './types.ts';
+import type { ComponentLike } from '@glint/template';
+import type { HdsTabsTabSignature } from './tab';
+import type { HdsTabsPanelSignature } from './panel';
+import type { HdsTabsPanelIds, HdsTabsTabIds } from './types.ts';
```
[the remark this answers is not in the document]
zamoore: What's the guide for ordering imports?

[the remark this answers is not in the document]
alex-ju: Fair question – I thought we order them alphabetically 🤷

packages/components/src/components/hds/tabs/panel.ts
```diff
@@ -6,14 +6,36 @@
 import Component from '@glimmer/component';
 import { guidFor } from '@ember/object/internals';
 import { action } from '@ember/object';
+import type { HdsTabsTabSignature } from './tab';
+import type { HdsTabsPanelIds, HdsTabsTabIds } from './types';
```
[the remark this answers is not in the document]
didoo: My bad they are correct, [they should be in alphabetical order](https://github.com/hashicorp/design-system/blob/main/wiki/TypeScript-Migration.md).

packages/components/src/components/hds/tabs/tab.ts
zamoore: Destructuring made sense to me because we're using the value from `this.nodeIndex` in more than one place now.

packages/components/src/template-registry.ts
zamoore: Should these not be in alphabetical order?

didoo: I don't know anymore 😀

packages/components/src/template-registry.ts
zamoore: I'll keep these in alphabetical order unless you want to change it up still?

packages/components/src/components/hds/tabs/index.ts
```diff
@@ -8,18 +8,44 @@ import { tracked } from '@glimmer/tracking';
...
+export default class HdsTabsIndexComponent extends Component<HdsTabsSignature> {
+  @tracked tabNodes: HTMLButtonElement[] = [];
+  @tracked tabIds: HdsTabsTabIds = [];
+  @tracked panelNodes: HTMLElement[] = [];
+  @tracked panelIds: HdsTabsPanelIds = [];
   @tracked _selectedTabIndex = this.args.selectedTabIndex ?? 0;
-  @tracked selectedTabId;
-  @tracked isControlled;
+  @tracked selectedTabId?: string;
+  @tracked isControlled = false;
```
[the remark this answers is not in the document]
zamoore: It looks like we don't need to and it can infer the type from the assignment in the constructor. Reverted.

packages/components/src/components/hds/tabs/tab.ts
zamoore: Is this ok or something you think we should revert before merging?

packages/components/src/components/hds/tabs/index.ts
```diff
@@ -8,18 +8,44 @@ import { tracked } from '@glimmer/tracking';
...
+export const SIZES: HdsTabsSizeValues[] = [
+  HdsTabsSizeValues.Medium,
+  HdsTabsSizeValues.Large,
+];
+
+interface HdsTabsSignature {
+  Args: {
+    size?: HdsTabsSizeValues;
+    onClickTab?: (event: MouseEvent, tabIndex: number) => void;
+    selectedTabIndex?: number;
```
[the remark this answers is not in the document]
zamoore: @alex-ju , curious on your opinions. To me, typing a property based on the argument type of a child component property feels odd.

Is this a case where we might want to define a type like SelectedTabIndex and use it in both the index and tab components?

packages/components/src/components/hds/tabs/index.ts
```diff
@@ -42,15 +64,25 @@ export default class HdsTabsIndexComponent extends Component {
...
+  ) {
+    super(owner, args);
 
     // this is to determine if the "selected" tab logic is controlled in the consumers' code or is maintained as an internal state
     this.isControlled = this.args.selectedTabIndex !== undefined;
   }
 
   get selectedTabIndex() {
     if (this.isControlled) {
+      assert(
```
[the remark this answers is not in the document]
zamoore: You're right. The SIZES code was a bad copy/paste from another assertion.

You're also correct that the assertion will never be false. Unfortunately, typescript is having a hard time discerning that fact and insists that `this.args.selectedTabIndex` can be `undefined`, even though this is impossible.

I've changed it to this instead of adding an assertion

`return this.args.selectedTabIndex!;`

packages/components/src/components/hds/tabs/index.ts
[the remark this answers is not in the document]
didoo: > typing a property based on the argument type of a child component property feels odd.

well, technically the parent component is _"consuming"_ the child sub-component, in the same way that a more complex component may be _"consuming"_ the Tabs (as composition with other elements/components)

packages/components/src/components/hds/tabs/tab.ts
didoo: not a blocker

packages/components/src/components/hds/tabs/index.ts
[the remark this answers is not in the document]
alex-ju: for me, it works either way. `HdsTabsTabSignature['Args']['selectedTabIndex']` has the advantage that it communicates the relationship, but it has the downside that the actual type (`number`) is not immediately surfaced.

packages/components/src/components/hds/tabs/index.ts
```diff
@@ -168,52 +190,52 @@ export default class HdsTabsIndexComponent extends Component {
...
     }
     // scroll selected tab into view (it may be out of view when activated using a keyboard with `prev/next`)
-    this.tabNodes[this.selectedTabIndex].parentNode.scrollIntoView({
+    (
+      this.tabNodes[this.selectedTabIndex]?.parentNode as HTMLElement
+    ).scrollIntoView({
       behavior: 'smooth',
       block: 'nearest',
       inline: 'nearest',
     });
```
alex-ju, line 216: I would generally prefer type checking over type casting
```suggestion
    const parentNode = this.tabNodes[this.selectedTabIndex]?.parentNode;
    if (parentNode instanceof HTMLElement) {
      parentNode.scrollIntoView({
        behavior: 'smooth',
        block: 'nearest',
        inline: 'nearest',
      });
    }
}```

alex-ju (approved):
Looking good! Not an easy component to convert. I noticed there are a few new comments from Cristiano, otherwise it's good to be merged from my perspective.

packages/components/src/components/hds/tabs/index.ts
didoo: +1

didoo:
> Looking good! Not an easy component to convert. I noticed there are a few new comments from Cristiano, otherwise it's good to be merged from my perspective.

If @alex-ju approved, consider my comments as non-blockers.

PR #2169, opened 2024-06-14
add explicit ember-get-config dependency

Dhaulagiri:
website relies on it but is getting it by accident from other project's deps


alex-ju (approved):
Rightly so

PR #2170, opened 2024-06-17
feat(ember-flight-icons): convert to typescript & publish types

aklkv:
### :pushpin: Summary

The main goal of this change is to make sure types are created and published for the Ember Flight Icon component

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-3315

Previously created PR that was massed up: #2126 

***

### 👀 Component checklist


packages/components/src/components/hds/button/index.ts
```diff
@@ -6,6 +6,7 @@
 import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
 import type { HdsInteractiveSignature } from '../interactive/';
+import type { IconName } from '@hashicorp/flight-icons/svg';
```
didoo, line 9: This needs to be updates too, to import the type from '@hashicorp/ember-flight-icons/components/flight-icon' as `FlightIconSignature`

packages/ember-flight-icons/.eslintignore
didoo, line 1: I see this change also in https://github.com/hashicorp/design-system/pull/2142/files#diff-e31a2dfa439cd2cc8aa9b1e01d10b79db320fd27009fdb3b129ea1bae67aa866 so this means that this branch should be rebased on https://github.com/hashicorp/design-system/pull/2142/ no?

didoo (commented):
Left a couple of comments

PR #2094, opened 2024-05-15
`SegmentedGroup` - Convert to TypeScript

didoo:
### :pushpin: Summary

Converts the `SegmentedGroup` component to TS

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/issues/HDS-2709

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#templates-for-npm-packages))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #2174, opened 2024-06-18
Pagination TS conversion

zamoore:
### :pushpin: Summary

If merged, this PR converts the `Hds::Pagination` component and its children to TS

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-2707


packages/components/src/components/hds/pagination/compact/index.ts
```diff
@@ -7,12 +7,38 @@ import Component from '@glimmer/component';
...
+    models?: unknown[];
+    pageSizes?: number[];
+    replace?: boolean;
+    route?: string;
+    showLabels?: boolean;
+    showSizeSelector?: boolean;
+    sizeSelectorLabel?: string;
+    onPageChange?: (page: PaginationDirection) => void;
+    onPageSizeChange?: (pageSize: number) => void;
+    // TODO: define the type of the function
```
zamoore, line 28: Remove comment

packages/components/src/components/hds/pagination/compact/index.ts
```diff
@@ -141,19 +170,20 @@ export default class HdsPaginationCompactIndexComponent extends Component {
   }
 
   @action
-  onPageChange(newPage) {
-    this.currentPage = newPage;
+  onPageChange(newPage: PaginationDirection) {
+    // TODO: I dont think this is actually used anywhere
+    // this.currentPage = newPage;
```
zamoore, line 175: This doesn't appear to be used, but I'd like confirmation

packages/components/src/components/hds/pagination/numbered/index.ts
```diff
@@ -267,15 +315,16 @@ export default class HdsPaginationNumberedIndexComponent extends Component {
...
+            page as number,
             this.currentPageSize
           ))
       );
     } else {
       routing.queryPrev = undefined;
       routing.queryNext = undefined;
-      routing.queryByPage = {};
+      // TODO: Is this used anywhere?
+      // routing.queryByPage = {};
```
zamoore, line 327: This doesn't appear to be used, but want to confirm

didoo:
@zamoore can you please move this PR to draft? the reasons I am asking are 1) we tend to avoid doing the conversion of two different/distinct components in the same PR, and 2) all the `Hds::Form`-related conversion will be done in the same batch (probably by @alex-ju).

PS: I am not sure if you have seen [this file](https://docs.google.com/spreadsheets/d/1sWzLSP8TUb3WYibYNZ5w6UfEb338DQbZkVaTjK8b1tw/edit#gid=0), it’s the list of components that need to be converted to TS, with their dependencies. If one component has a dependency, we should do the dependency first

zamoore:
> @zamoore can you please move this PR to draft? the reasons I am asking are 1) we tend to avoid doing the conversion of two different/distinct components in the same PR, and 2) all the `Hds::Form`-related conversion will be done in the same batch (probably by @alex-ju).
> 
> PS: I am not sure if you have seen [this file](https://docs.google.com/spreadsheets/d/1sWzLSP8TUb3WYibYNZ5w6UfEb338DQbZkVaTjK8b1tw/edit#gid=0), it’s the list of components that need to be converted to TS, with their dependencies. If one component has a dependency, we should do the dependency first

@didoo Sure thing, would we want to wait to merge the changes to Pagination until the dropdown work is done as part of the batched form conversion work then? Or are you comfortable with merging to dropdown work now in a separate PR?

didoo:
> > @zamoore can you please move this PR to draft? the reasons I am asking are 1) we tend to avoid doing the conversion of two different/distinct components in the same PR, and 2) all the `Hds::Form`-related conversion will be done in the same batch (probably by @alex-ju).
> > PS: I am not sure if you have seen [this file](https://docs.google.com/spreadsheets/d/1sWzLSP8TUb3WYibYNZ5w6UfEb338DQbZkVaTjK8b1tw/edit#gid=0), it’s the list of components that need to be converted to TS, with their dependencies. If one component has a dependency, we should do the dependency first
> 
> @didoo Sure thing, would we want to wait to merge the changes to Pagination until the dropdown work is done as part of the batched form conversion work then? Or are you comfortable with merging to dropdown work now in a separate PR?

I would prefer to wait if you don't mind (I made the same thing here, so don't worry: https://github.com/hashicorp/design-system/pull/2094)

alex-ju:
you probably saw this @zamoore, but I ported `form/select/base` together with the rest of the form conversions, so this will require a rebase, then we can focus on the pagination bits

packages/components/src/components/hds/pagination/compact/index.ts
```diff
@@ -7,12 +7,37 @@ import Component from '@glimmer/component';
...
+    replace?: boolean;
+    queryFunction?: (
+      page: HdsPaginationDirections,
+      pageSize?: number
+    ) => Record<string, unknown>;
+    onPageChange?: (page: HdsPaginationDirections) => void;
+    onPageSizeChange?: (pageSize: number) => void;
+  };
+  Element: HTMLDivElement;
+}
```
alex-ju, line 34: If we look at the template, most of these arguments are actually coming from subcomponents. This means we can write it as such to avoid duplicating argument types:
```suggestion
interface HdsPaginationCompactSignature {
  HdsPaginationControlArrowSignature['Args'] & HdsPaginationSizeSelectorSignature['Args'] & { [...] }  
  Element: HTMLDivElement;
}
```

packages/components/src/components/hds/pagination/numbered/index.ts
```diff
@@ -7,6 +7,35 @@ import Component from '@glimmer/component';
...
+    pageSizes?: number[];
+    currentPageSize?: number;
+    route?: string;
+    model?: unknown;
+    models?: unknown[];
+    replace?: boolean;
+    queryFunction?: (page: number, pageSize: number) => Record<string, unknown>;
+    onPageChange?: (page: number, pageSize: number) => void;
+    onPageSizeChange?: (pageSize: number) => void;
+  };
```
alex-ju, line 34: Same here about composing this from subcomponents' signatures

packages/components/src/components/hds/pagination/compact/index.ts
```diff
@@ -27,10 +52,10 @@ export default class HdsPaginationCompactIndexComponent extends Component {
   showLabels = this.args.showLabels ?? true; // if the labels for the "prev/next" controls are visible
   showSizeSelector = this.args.showSizeSelector ?? false; // if the "size selector" block is visible
 
-  constructor() {
-    super(...arguments);
+  constructor(owner: unknown, args: HdsPaginationCompactSignature['Args']) {
```
alex-ju, line 55: 👍

packages/components/src/components/hds/pagination/compact/index.ts
```diff
@@ -96,7 +121,7 @@ export default class HdsPaginationCompactIndexComponent extends Component {
    * @default [10, 30, 50]
    */
   get pageSizes() {
```
alex-ju, line 123: don't forget the return types for all functions

packages/components/src/components/hds/pagination/info/index.ts
```diff
@@ -4,8 +4,19 @@
...
 
-export default class HdsPaginationInfoComponent extends Component {
+interface HdsPaginationInfoSignature {
+  Args: {
+    itemsRangeStart: number;
+    itemsRangeEnd: number;
+    totalItems: number;
+    showTotalItems?: boolean;
+  };
+  Element: HdsTextBodySignature['Element'];
```
alex-ju, line 16: 🙌

packages/components/src/components/hds/pagination/nav/arrow.ts
```diff
@@ -6,12 +6,36 @@
 import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { assert } from '@ember/debug';
+import type { HdsPaginationDirections } from '../types.ts';
+import type { HdsInteractiveSignature } from '../../interactive/index.ts';
+import type { IconName } from '@hashicorp/flight-icons/svg';
 
-export const DIRECTIONS = ['prev', 'next'];
+export const DIRECTIONS: HdsPaginationDirections[] = ['prev', 'next'];
```
alex-ju, line 13: can we convert this into an enum and move it into `types.ts`? we have a pattern around such constants

packages/components/src/components/hds/pagination/nav/arrow.ts
```diff
@@ -6,12 +6,36 @@
...
 
-export default class HdsPaginationControlArrowComponent extends Component {
+interface HdsPaginationControlArrowContent {
+  label?: 'Previous' | 'Next';
+  icon?: IconName;
+  ariaLabel?: 'Previous page' | 'Next page';
```
alex-ju, line 18: same here, it would be good to have the `label` and `ariaLabel` in enums

alex-ju (commented):
I did a first pass and left a couple of high-level comments. We seem to have a failing test that would need investigated. Would you be able to look into that?

packages/components/src/components/hds/pagination/compact/index.ts
zamoore: I'm not sure I understand how this works in practice. `HdsPaginationControlArrowSignature['Args']` shares some, but not all of the `Args` of the pagination component, which is its parent. We might have to pair on Monday if I'm missing something.

alex-ju: I was hoping for something like this (and I understand that in some instances, such as the size selector, it may not be worthy) so we avoid declaring the types twice.
```suggestion
interface HdsPaginationCompactSignature {
  Args: Omit<HdsPaginationControlArrowSignature['Args'], 'direction'> & {
    ariaLabel?: string;
    showLabels?: boolean;
    isDisabledPrev?: boolean;
    isDisabledNext?: boolean;
    showSizeSelector?: boolean;
    sizeSelectorLabel?: string;
    pageSizes?: number[];
    currentPageSize?: number;
    queryFunction?: (
      page: HdsPaginationDirections,
      pageSize?: number
    ) => Record<string, unknown>;
    onPageChange?: (page: HdsPaginationDirections) => void;
    onPageSizeChange?: (pageSize: number) => void;
  };
  Element: HTMLDivElement;
}
```

Let me know if it doesn't make sense and I'll book us some time to go through it

alex-ju (commented):
Beware we have `{{! @glint-nocheck: not typesafe yet }}` to remove from `pagination/compact/index.hbs` and `pagination/numbered/index.hbs` and add entries to the `template-registry.ts` and the barrel export

packages/components/src/components/hds/form/select/base.ts
```diff
@@ -23,11 +23,6 @@ export interface HdsFormSelectBaseSignature {
 }
 
 export default class HdsFormSelectBaseComponent extends Component<HdsFormSelectBaseSignature> {
-  /**
-   * Get the class names to apply to the component.
-   * @method classNames
-   * @return {string} The "class" attribute to apply to the component.
-   */
```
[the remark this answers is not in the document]
zamoore: Yeah that makes sense to me.

packages/components/src/components/hds/pagination/nav/arrow.ts
```diff
@@ -0,0 +1,100 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+import { action } from '@ember/object';
+import { assert } from '@ember/debug';
+import {
+  HdsPaginationDirectionValues,
+  HdsPaginationDirectionAriaLabelValues,
```
alex-ju, line 11: Looking at the failing tests, we'd need three sets of enums for directions, instead of two.
 - one for the `prev`/`next` to match the exposed values for `direction` (currently covered by `HdsPaginationDirectionValues`)
 - one for the visible labels: `Previous` and `Next` (currently not covered)
 - and one for the `ariaLabel` (covered by `HdsPaginationDirectionAriaLabelValues`)

.changeset/hungry-cycles-drum.md
```diff
@@ -0,0 +1,6 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Hds::Pagination`
```
shleewhite, line 5: Small, but usually we've written the changelogs as:
```suggestion
`Hds::Pagination` - Converted component to Typescript
```

(instead of 2 lines)

packages/components/src/components/hds/pagination/nav/arrow.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+    if (direction === HdsPaginationDirectionValues.Next) {
+      content = {
+        label: HdsPaginationDirectionLabelValues.Next,
+        icon: 'chevron-right',
+        ariaLabel: HdsPaginationDirectionAriaLabelValues.Next,
+      };
+    }
+
+    return content as HdsPaginationControlArrowContent;
```
shleewhite, line 72: Could this be refactored so we don't need the `return content as HdsPaginationControlArrowContent`? Especially since direction should always be defined and there's only 2 options?

```typescript
    if (direction === HdsPaginationDirectionValues.Prev) {
      return {
        label: HdsPaginationDirectionLabelValues.Prev,
        icon: 'chevron-left',
        ariaLabel: HdsPaginationDirectionAriaLabelValues.Prev,
      };
    } else {
      return {
        label: HdsPaginationDirectionLabelValues.Next,
        icon: 'chevron-right',
        ariaLabel: HdsPaginationDirectionAriaLabelValues.Next,
      };
    }
```

packages/components/src/components/hds/pagination/numbered/index.hbs
```diff
@@ -33,7 +32,7 @@
               <Hds::Pagination::Nav::Ellipsis />
             {{else}}
               <Hds::Pagination::Nav::Number
-                @page={{page}}
+                @page={{this.elliptizedPageArrayItemAsNumber page}}
```
shleewhite, line 35: Out of curiosity, why did this change?

packages/components/src/components/hds/pagination/numbered/index.ts
```diff
@@ -2,28 +2,72 @@
...
+  onPageChange?: (page: number, pageSize: number) => unknown;
+  onPageSizeChange?: (pageSize: number) => unknown;
+}
+
+export interface HdsPaginationNumberedSignature {
+  Args: HdsPaginationNumberedArgs & HdsPaginationRoutingProps;
+  Element: HTMLDivElement;
+}
+
+const ELLIPSIS = '…';
```
shleewhite, line 58: Nice

packages/components/src/components/hds/pagination/numbered/index.ts
```diff
@@ -59,19 +109,18 @@ export const elliptize = ({ pages, current, limit = 7 }) => {
     const delta = (limit - 5) / 2; // this is why the limit needs to be an odd number
     // we slice the array starting at the "current" index, minus the delta, minus one because it's an array (zero-based)
     const sliceCurr = pages.slice(current - delta - 1, current + delta);
-    result = [].concat(
-      sliceStart.shift(),
-      ellipsis,
+    result = ([] as HdsPaginationElliptizedPageArray).concat(
```
shleewhite, line 112: Probably out of scope for this PR, but its odd to me that we support both number and string for these... Should we consider scoping it down to just number? Is there a specific reason we allow both?

packages/components/src/components/hds/pagination/numbered/index.ts
```diff
@@ -327,4 +376,18 @@ export default class HdsPaginationNumberedIndexComponent extends Component {
       onPageSizeChange(newPageSize);
     }
   }
+
+  elliptizedPageArrayItemAsNumber = (
```
shleewhite, line 380: Continuing from the above comment: this seems extra to have to do... shouldn't it always be a number?

shleewhite (commented):
Looks good!

packages/components/src/components/hds/pagination/numbered/index.ts
zamoore: It's either a number or an ellipsis.

packages/components/src/components/hds/pagination/numbered/index.hbs
zamoore: Glint isn't smart enough to figure out the conditional above this line where we check if the value is an ellipsis, and only render the `Hds::Pagination::Nav::Number` component if it's not. The `Number` component always expects `@page` to be a number, so we need to run the value through a function that returns the value as a number. It's a little hacky, but there's no other way to cast a type in Glimmer that I'm aware of.

packages/components/src/components/hds/pagination/nav/arrow.ts
zamoore: That's a good point. I took a look and came up with an alternative. Thanks for the suggestion.

.changeset/hungry-cycles-drum.md
```diff
@@ -0,0 +1,6 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Hds::Pagination` - Converted component to Typescript
+- Converted to Typescript
```
shleewhite, line 6: Oh, you can delete this line now.

packages/components/src/components/hds/pagination/nav/arrow.ts
```diff
@@ -0,0 +1,99 @@
...
+        ariaLabel: HdsPaginationDirectionAriaLabelValues.Prev,
+      },
+      [HdsPaginationDirectionValues.Next]: {
+        label: HdsPaginationDirectionLabelValues.Next,
+        icon: 'chevron-right',
+        ariaLabel: HdsPaginationDirectionAriaLabelValues.Next,
+      },
+    };
+
+    return hdsPaginationNavArrowContentDirectionMap[direction];
```
shleewhite, line 72: Love this implementation!

packages/components/src/components/hds/pagination/numbered/index.ts
shleewhite: Ahh... ok I misread and thought it was either a number or a string of a number.. maybe we can limit the type to be `number | ELLIPSIS` ? that way the ellipsis definition can be moved to the types file which is more inline with how we define re-usable strings?

shleewhite (approved):
Overall, seems good! The only blocking comment is the changelog one.

.changeset/hungry-cycles-drum.md
zamoore: Whoops. Did this through the Github interface 😆

packages/components/src/components/hds/pagination/numbered/index.ts
zamoore: I did try that, but it causes some issues if we try to initialize an empty object that uses keys that are of that type.

For instance:

```
queryPages: Record<HdsPaginationElliptizedPageArrayItem, string> = {};
```

The above will work if the type of `HdsPaginationElliptizedPageArrayItem` is `string | number`, but not if its `typeof ELLIPSES | number`

PR #2176, opened 2024-06-19
feat(showcase): speed up tests in CI

aklkv:
### :pushpin: Summary

This change cuts component tests in CI by 50%, today on average it takes around 4 minutes with this change we reduce it to 2 minutes.

The only potential issue I see is this message:

```
[percy] Error: This organization has exceeded the limits of the Percy BrowserStack plan. Administrators can upgrade here: https://percy.io/organizations/587d9efb/billing
```

Before:

Screenshot 2024-06-18 at 7 47 49 PM

After:

Screenshot 2024-06-18 at 7 48 27 PM

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist


Dhaulagiri:
FWIW the percy message is not unique to this PR. I see it in [others](https://github.com/hashicorp/design-system/actions/runs/9575562468/job/26400648601?pr=2176#step:9:13) as well.

showcase/testem.js
```diff
@@ -6,6 +6,7 @@
 'use strict';
 
 module.exports = {
+  parallel: process.env.CI ? -1 : 1,
```
alex-ju, line 9: for posterity, setting parallel execution to `-1` means auto-detect and use the maximum number of parallel processes possible, while setting it to 1 means executing tests serially, one after the other

alex-ju (commented):
Generally, using `ember-exam` sounds like a good idea for introducing parallel testing. 

I've only been able to compare `test:ember` (not including the VRT testing in Percy) and the differences are marginal (2m12s, compared to 2m20s for `main`); However, it could make all the difference on `test:ember:percy`, but we seem to have some challenges around using Percy so not able to test that atm.

aklkv:
> Generally, using `ember-exam` sounds like a good idea for introducing parallel testing.
> 
> I've only been able to compare `test:ember` (not including the VRT testing in Percy) and the differences are marginal (2m12s, compared to 2m20s for `main`); However, it could make all the difference on `test:ember:percy`, but we seem to have some challenges around using Percy so not able to test that atm.

I tried adding `ember-exam` to website tests but it didn't result in much savings so I thought there wouldn't be much sense in adding it at this stage. I am happy to add this instrumentation here as well for consistency.

On the other hand `showcase` test were much more cooperative and we can see ~1min saving

showcase/testem.js
aklkv: add a comment

aklkv:
> FWIW the percy message is not unique to this PR. I see it in [others](https://github.com/hashicorp/design-system/actions/runs/9575562468/job/26400648601?pr=2176#step:9:13) as well.

oh, good to know I thought `ember-exam` parallel runs triggered this warning

Dhaulagiri:
Percy is fixed now so we could re-run the comparisons to see if the improvement in perf you were seeing was related to the introduction of ember-exam or a coincidence because percy tests were no longer running

aklkv:
It seems that `ember-exam` & `percy` is no bueno, it started flaking while waiting to take a snapshot

aklkv:
@Dhaulagiri I think I got it fixed so the tests aren't flaking anymore but but speed gains are a little less we are sitting at `2m 45s` which is still a good reduction

Dhaulagiri:
@aklkv apologies for the delay on this. lost track of this over the holiday. 

the time savings is _relatively_ small but it _is_ faster overall, so this seems worth bringing in to me given the low cost to opt-in to this feature. @alex-ju any objections from your perspective?

aklkv:
@Dhaulagiri No worries at all. One thing I noticed that Percy is still in `Pending` state, I wonder if this is something related to my changes

Dhaulagiri:
🤔 
> Only 3 of the 8 parallelized build processes reported they were finished.

Dhaulagiri:
I was doing a a lot of manual starting/stopping of the tests to test this so that may be part of it, but we may also need to try following all the instructions [here](https://www.browserstack.com/docs/percy/integrate/parallel-test-suites#parallel-suites-split-across-different-machines).

alex-ju:
> the time savings is _relatively_ small but it _is_ faster overall, so this seems worth bringing in to me given the low cost to opt-in to this feature. @alex-ju any objections from your perspective?

no objections, any improvement is progress. I'll try and give this PR a fresh pair of eyes today or on Monday

PR #2177, opened 2024-06-20
fix(components): fix syncing of <F.Error/> ids to their Form elements via aria-describedby

fivetanley:
### :pushpin: Summary

Fixes all `Form::Field::*` that yield an `Error` component to properly update `aria-describedby` once an `Error` component is rendered. This previously only worked when `F.Error` was done statically. This now works when rendering `F.Error` dynamically, e.g. using `{{#if condition}}<F.Error />{{/if}}`.

This PR fixes the issue where dynamically rendering a `<F.Error/>` component did not sync the id of the `<F.Error` to the `aria-describedby` of the Form element. This didn't work for several reasons:

1. The `ariaDescribedBy` field on components only used the values in `@extraAriaDescribedBy` on initialization. If the value of `@extraAriaDescribedBy` was changed via changes to `this.args`, this was not reflected in parent components or child components. https://github.com/hashicorp/design-system/blob/eaab5b3b4a15edbea7752fb937c5475fdcd5d42d/packages/components/src/components/hds/form/field/index.js#L17
2. `setAriaDescribedBy` was only called on `{{did-insert}}`, meaning that any internal state changes to `descriptors` or `extraAriaDescribedBy` were only realized on initial render. Any subsequent changes were ignored. https://github.com/hashicorp/design-system/blob/eaab5b3b4a15edbea7752fb937c5475fdcd5d42d/packages/components/src/components/hds/form/field/index.js#L26 https://github.com/hashicorp/design-system/blob/eaab5b3b4a15edbea7752fb937c5475fdcd5d42d/packages/components/src/components/hds/form/field/index.hbs#L6
3. Update code used native Array's `push` method. Since the array used is not a [TrackedArray](https://github.com/tracked-tools/tracked-built-ins), using `push` will not cause Ember/Glimmer's rendering system to pick up changes and rerender.
4. Nodes added were never removed. While this wasn't usually an issue since screenreaders tend to ignore DOM ids that don't exist, this still was messy and could potentially cause different bugs down the line if IDs are reused (e.g. IDs get passed in instead of using the randomly generated defaults).

This PR changes the way that `aria-describedby` syncing works in a few ways:

1. Add lots of tests!
2. We take advantage of Ember/Glimmer's tracking fully. Rather than updating two properties manually, we update one and let `ariaDescribedBy` become a computed property via a getter. This allows the component to re-render when either the descriptors or `this.args.extraAriaDescribedBy` changes.
3. Introduce a new mixin via the `@ariaDescribedBy` decorator. This allows us to reduce the implementation to one file and keep fixes in one place.
4. Introduce `AriaDescribedByAttributes`, an object whose responsibility it is to keep track of components and their elements/ids and update the component in a timely and performant manner when new nodes are added/removed.
5. Remove node IDs from the `aria-describedby` when they are removed from the DOM. This aids in debugging efforts later

N/A

### :link: External links

Jira ticket: [HDS-3441](https://hashicorp.atlassian.net/browse/HDS-3441)

***

### 👀 Component checklist

[HDS-3441]: https://hashicorp.atlassian.net/browse/HDS-3441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/components/hds/form/radio-card/index.js
```diff
@@ -4,29 +4,14 @@
...
-  @tracked descriptors = [];
-
-  @action
-  setAriaDescribedBy() {
-    // we schedule this afterRender to capture all descriptors registered onInsert
-    schedule('afterRender', () => {
-      setAriaDescribedBy(this);
-    });
-  }
-
```
fivetanley, line 29: None of this seems to actually be used by the component. It seems like maybe a copy/paste from another component but it doesn't look like this has ever been used by this component: https://github.com/hashicorp/design-system/commit/5edb999bae52a5f1bf1dad68a983e57d7993e3b5

packages/components/src/components/hds/form/radio-card/index.hbs
```diff
@@ -18,8 +18,8 @@
       name={{@name}}
       checked={{@checked}}
       disabled={{@disabled}}
-      aria-describedby={{this.ariaDescribedBy}}
+      aria-describedby={{@extraAriaDescribedBy}}
```
fivetanley, line 21: since `this.ariaDescribedBy` wasn't updated via children, we can just use `@extraAriaDescribedBy` for this component

packages/components/src/components/hds/form/radio-card/index.hbs
```diff
@@ -3,7 +3,7 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<label class={{this.classNames}} {{did-insert this.setAriaDescribedBy}} {{style maxWidth=@maxWidth}}>
```
fivetanley, line 6: The descriptors are never updated dynamically and never have been. the did-insert here is un-necessary since the introduction of this component: https://github.com/hashicorp/design-system/commit/5edb999bae52a5f1bf1dad68a983e57d7993e3b5

packages/components/src/utils/hds-aria-described-by.ts
```diff
@@ -0,0 +1,96 @@
...
+  ariaDescribedBy: string;
+}
+
+// essentially a type that says we return a subclass of the given type T
+type ClassOf<T> = new () => T;
+
+export function ariaDescribedBy(
+  klass: ClassOf<AriaDescribedByComponent>
+): ClassOf<AriaDescribedByComponent> {
+  class X extends klass {
```
fivetanley, line 60: have to write this as `class X extends klass {}; return X` instead of `return class extends klass {}`. Otherwise typescript complains about using decorators for some reason 🤷

packages/components/src/utils/hds-aria-described-by.ts
```diff
@@ -0,0 +1,96 @@
...
```
alex-ju, line 60: trying to come up with more descriptive names here; what do you think about this?
```suggestion
export function ariaDescribedBy(
  BaseComponent: ClassOf<AriaDescribedByComponent>
): ClassOf<AriaDescribedByComponent> {
  class EnhancedComponent extends BaseComponent {
```

alex-ju (approved):
Looking solid! I admit, I wouldn't have thought of this myself. I like the class extension and the synchronization approach 🙌

.changeset/hip-bobcats-invent.md
```diff
@@ -0,0 +1,6 @@
+---
+"@hashicorp/design-system-components": patch
+"showcase": patch
```
Dhaulagiri, line 3:
```suggestion
```
We don't version the showcase

PR #2178, opened 2024-06-21
`Dropdown::Toggle::Icon`: Fix inconsistencies (HDS-3416)

KristinLBradley:
### :pushpin: Summary

If merged, this PR fixes inconsistencies in color and toggle icon sizing between the `Dropdown::Toggle::Icon` as compared to the `Button` component.

**Showcase:** https://hds-showcase-git-hds-3416-fix-dropdown-inconsi-809ea9-hashicorp.vercel.app/components/dropdown#icon

**BEFORE:**

image

---

**AFTER:**

image

### :link: External links

* Jira ticket: [HDS-3416](https://hashicorp.atlassian.net/browse/HDS-3416)

***

### 👀 Component checklist

[HDS-3416]: https://hashicorp.atlassian.net/browse/HDS-3416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/dropdown/toggle/icon.hbs
+++ b/packages/components/src/components/hds/dropdown/toggle/icon.hbs
@@ -17,10 +17,10 @@
       {{#if this.hasImage}}
         <img src={{@imageSrc}} alt="" role="presentation" {{on "error" this.onImageLoadError}} />
       {{else}}
-        <FlightIcon @name="user" @size={{this.iconSize}} />
+        <FlightIcon @name="user" @size="16" @stretched={{true}} />
       {{/if}}
     {{else if @icon}}
-      <FlightIcon @name={{@icon}} @size={{this.iconSize}} />
+      <FlightIcon @name={{@icon}} @size="16" @stretched={{true}} />
     {{/if}}
   </div>
   {{#if this.hasChevron}}
--- a/packages/components/src/components/hds/dropdown/toggle/icon.hbs
+++ b/packages/components/src/components/hds/dropdown/toggle/icon.hbs
@@ -17,10 +17,10 @@
       {{#if this.hasImage}}
         <img src={{@imageSrc}} alt="" role="presentation" {{on "error" this.onImageLoadError}} />
       {{else}}
-        <FlightIcon @name="user" @size="16" @stretched={{true}} />
+        <FlightIcon @name="user" @size={{this.iconSize}} />
       {{/if}}
     {{else if @icon}}
-      <FlightIcon @name={{@icon}} @size="16" @stretched={{true}} />
+      <FlightIcon @name={{@icon}} @size={{this.iconSize}} />
     {{/if}}
   </div>
   {{#if this.hasChevron}}
--- a/packages/components/src/components/hds/dropdown/toggle/icon.js
+++ b/packages/components/src/components/hds/dropdown/toggle/icon.js
@@ -84,12 +84,6 @@ export default class HdsDropdownToggleIconComponent extends Component {
       // this is the default size (notice: for the "small" variant with chevron, we set the actual size to `12px` via CSS)
       return '16';
     }
-      // if hasChevron is true, we resize the icon to 12px in CSS
-      return '16';
-    } else {
-      // if hasChevron is true, we use 16px icon
-      return this.hasChevron ? '16' : '24';
-    }
   }
 
   /**
--- a/packages/components/src/components/hds/dropdown/toggle/icon.js
+++ b/packages/components/src/components/hds/dropdown/toggle/icon.js
@@ -71,6 +71,22 @@ export default class HdsDropdownToggleIconComponent extends Component {
   }
 
   /**
+   * @param iconSize
+   * @type {string}
+   * @default 24
+   * @description ensures that the correct icon size is used
+   */
+  get iconSize() {
+    if (this.args.size === 'small') {
+      // if hasChevron is true, we resize the icon to 12px in CSS
+      return '16';
+    } else {
+      // if hasChevron is true, we use 16px icon
+      return this.hasChevron ? '16' : '24';
+    }
+  }
+
+  /**
    * Indicates if a dropdown chevron icon should be displayed; should be displayed unless the "more-horizontal" icon is used.
    *
    * @param hasChevron
@@ -112,6 +128,11 @@ export default class HdsDropdownToggleIconComponent extends Component {
     // add a class based on the @isOpen argument
     if (this.args.isOpen) {
       classes.push('hds-dropdown-toggle-icon--is-open');
+    }
+
+    // add a class based on the @hasChevron argument
+    if (this.hasChevron) {
+      classes.push('hds-dropdown-toggle-icon--has-chevron');
     }
 
     return classes.join(' ');
--- a/packages/components/src/components/hds/dropdown/toggle/icon.js
+++ b/packages/components/src/components/hds/dropdown/toggle/icon.js
@@ -77,7 +77,13 @@ export default class HdsDropdownToggleIconComponent extends Component {
    * @description ensures that the correct icon size is used
    */
   get iconSize() {
-    if (this.args.size === 'small') {
+    if (this.args.size === 'medium' && !this.hasChevron) {
+      // in this special case we use a larger SVG
+      return '24';
+    } else {
+      // this is the default size (notice: for the "small" variant with chevron, we set the actual size to `12px` via CSS)
+      return '16';
+    }
       // if hasChevron is true, we resize the icon to 12px in CSS
       return '16';
     } else {
--- a/packages/components/src/components/hds/dropdown/toggle/icon.js
+++ b/packages/components/src/components/hds/dropdown/toggle/icon.js
@@ -71,20 +71,6 @@ export default class HdsDropdownToggleIconComponent extends Component {
   }
 
   /**
-   * @param iconSize
-   * @type {string}
-   * @default 24
-   * @description ensures that the correct icon size is used
-   */
-  get iconSize() {
-    if (this.args.size === 'small') {
-      return '16';
-    } else {
-      return '24';
-    }
-  }
-
-  /**
    * Indicates if a dropdown chevron icon should be displayed; should be displayed unless the "more-horizontal" icon is used.
    *
    * @param hasChevron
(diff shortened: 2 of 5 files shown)


KristinLBradley:
> Here's where some inconsistencies between Figma and code are a bit problematic, in Figma the ToggleIcon component has a `type` property that accepts:
> 
> * `overflow`
> * `icon`
> * `image`
> * `image`
> 
> In code the method to mimic the `overflow` variant is by setting `hasChevron={{false}}`.
> 
> What we were proposing was keeping the icon size in the `overflow` variant the same as the `more-horizontal` icon looks very small within the component. Is it possible/feasible to set the size of the icon conditionally based on the `hasChevron` property? That seems a little bit strange to me, but I'm not 100% of the complexity.

Could you add some visual examples of what you want?

jorytindall (commented):
I added a couple of other comments on Friday, but I failed to submit the review 😞 .

I've also added a bit of a [matrix here](https://www.figma.com/design/9SyPujbtG4C4mcJUBFU9sN/Troubleshooting-%26-Testing?node-id=237-1131&t=NBZ2AD0Pkzt5ZlOU-1) for what is changing. Let me know if this makes sense, but essentially I think it breaks down like this:

- `Dropdown::ToggleIcon` where `hasChevron={{true}}` and has an `icon`
    - `size="small"` icon size should be `12px`
    - `size="medium"` icon size should `16px`

I think everything else should be unchanged.

packages/components/src/components/hds/dropdown/toggle/icon.hbs
```diff
@@ -17,10 +17,10 @@
       {{#if this.hasImage}}
         <img src={{@imageSrc}} alt="" role="presentation" {{on "error" this.onImageLoadError}} />
       {{else}}
-        <FlightIcon @name="user" @size={{this.iconSize}} />
+        <FlightIcon @name="user" @size="16" @stretched={{true}} />
       {{/if}}
     {{else if @icon}}
-      <FlightIcon @name={{@icon}} @size={{this.iconSize}} />
+      <FlightIcon @name={{@icon}} @size="16" @stretched={{true}} />
```
KristinLBradley, line 23: The size will always be 16px now for the underlying SVG. CSS is used to shrink the icon for the smaller 12px size variant.

packages/components/src/styles/components/dropdown.scss
```diff
@@ -73,16 +74,32 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
...
+
 .hds-dropdown-toggle-icon--size-small {
   .hds-dropdown-toggle-icon__wrapper {
-    width: 24px; // same as the `button[small]` minus 2*2px becaus there are a 1px padding and 1px border to take into account
+    width: 24px; // same as the `button[small]` minus 2*2px because there are a 1px padding and 1px border to take into account
     height: 24px;
   }
+
+  // For small variant, override default icon size except for "more-horizontal" icon
+  .flight-icon:not(.flight-icon-more-horizontal) {
```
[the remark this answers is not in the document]
KristinLBradley: Thanks, I forgot to override the size for the medium variant

KristinLBradley:
> Here's where some inconsistencies between Figma and code are a bit problematic, in Figma the ToggleIcon component has a `type` property that accepts:
> 
> * `overflow`
> * `icon`
> * `image`
> * `image`
> 
> In code the method to mimic the `overflow` variant is by setting `hasChevron={{false}}`.
> 
> What we were proposing was keeping the icon size in the `overflow` variant the same as the `more-horizontal` icon looks very small within the component. Is it possible/feasible to set the size of the icon conditionally based on the `hasChevron` property? That seems a little bit strange to me, but I'm not 100% of the complexity.

@jorytindall Ok so specifically is this what you mean should be done?
* Decrease the main icon size if there IS a chevron displayed together with it. 
(16px => 12px | 24px => 16px)
* If there is NOT a chevron, displayed together with the main icon, keep size as is.

So _all_ icons should follow the same rules instead of just adding an exception for the "more-horizontal"/ellipsis (...) icon?

packages/components/src/styles/components/dropdown.scss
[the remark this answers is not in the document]
KristinLBradley: Also I think I over-complicated it by using the :not selector...

packages/components/src/styles/components/dropdown.scss
```diff
@@ -73,16 +74,32 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
...
+    height: 12px;
+  }
 }
 
 .hds-dropdown-toggle-icon--size-medium {
   .hds-dropdown-toggle-icon__wrapper {
-    width: 32px; // same as the `button[medium]` minus 2*2px becaus there are a 1px padding and 1px border to take into account
+    width: 32px; // same as the `button[medium]` minus 2*2px because there are a 1px padding and 1px border to take into account
     height: 32px;
   }
```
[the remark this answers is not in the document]
KristinLBradley: I re-did how the logic works. Style changes have been simplified.

packages/components/src/components/hds/dropdown/toggle/icon.js
```diff
@@ -78,9 +78,11 @@ export default class HdsDropdownToggleIconComponent extends Component {
    */
   get iconSize() {
     if (this.args.size === 'small') {
+      // if hasChevron is true, we resize the icon to 12px in CSS
       return '16';
     } else {
-      return '24';
+      // if hasChevron is true, we use 16px icon
+      return this.hasChevron ? '16' : '24';
```
jorytindall, line 85: Elegant!

jorytindall (approved):
Nice! Thanks for the work iterating here!

packages/components/src/styles/components/dropdown.scss
```diff
@@ -73,16 +74,26 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
...
 }
 
+// Icon sizes
+
 .hds-dropdown-toggle-icon--size-small {
   .hds-dropdown-toggle-icon__wrapper {
-    width: 24px; // same as the `button[small]` minus 2*2px becaus there are a 1px padding and 1px border to take into account
+    width: 24px; // same as the `button[small]` minus 2*2px because there are a 1px padding and 1px border to take into account
```
didoo, line 81: I'm the one who wrote the comment (I think) and it's not clear even to me 😀
```suggestion
    width: 24px; // we use this element to provide the right size for the parent "button" so it matches the `button[small]`
```
(feel free to edit/reword it)

packages/components/src/styles/components/dropdown.scss
```diff
@@ -73,16 +74,26 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
...
+      width: 12px;
+      height: 12px;
+    }
+  }
 }
 
 .hds-dropdown-toggle-icon--size-medium {
   .hds-dropdown-toggle-icon__wrapper {
-    width: 32px; // same as the `button[medium]` minus 2*2px becaus there are a 1px padding and 1px border to take into account
+    width: 32px; // same as the `button[medium]` minus 2*2px because there are a 1px padding and 1px border to take into account
```
didoo, line 96:
```suggestion
    width: 32px; // we use this element to provide the right size for the parent "button" so it matches the `button[small]`
```

packages/components/src/styles/components/dropdown.scss
```diff
@@ -73,16 +74,26 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
...
+
 .hds-dropdown-toggle-icon--size-small {
   .hds-dropdown-toggle-icon__wrapper {
-    width: 24px; // same as the `button[small]` minus 2*2px becaus there are a 1px padding and 1px border to take into account
+    width: 24px; // same as the `button[small]` minus 2*2px because there are a 1px padding and 1px border to take into account
     height: 24px;
   }
+
+  &.hds-dropdown-toggle-icon--has-chevron {
+    // For small variant with chevron, reduce the icon size (medium size icon is set in JS)
```
didoo, line 86:
```suggestion
    // For small variant with chevron, force the icon size to custom (even if the SGV size is `16px`)
```

packages/components/src/styles/components/dropdown.scss
```diff
@@ -33,6 +33,7 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
   align-items: center;
   justify-content: center;
   padding: 1px;
+  color: var(--token-color-foreground-primary); // does not have secondary & primary variants so we use the primary color
```
didoo, line 36: I would keep the comment even more generic
```suggestion
  // this variant mimics the secondary button, but not entirely (eg. it doesn't have an elevation shadow)
  // so we need to provide the same color values but without using the `hds-button-color-secondary()` mixin
  color: var(--token-color-foreground-primary);
```
(feel free to edit/reword)

packages/components/src/components/hds/dropdown/toggle/icon.js
```diff
@@ -78,9 +78,11 @@ export default class HdsDropdownToggleIconComponent extends Component {
    */
   get iconSize() {
     if (this.args.size === 'small') {
```
didoo, line 80: maybe something like this would be more clear? 🤔
```suggestion
    if (this.args.size === 'medium' && !this.hasChevron) {
      // in this special case we use a larger SVG
      return '24';
    } else {
      // this is the default size (notice: for the "small" variant with chevron, we set the actual size to `12px` via CSS)
      return '16';
    }
```

didoo (approved):
Left a few suggestions, but nothing is a blocker.

PS: for the record: I think the `12px` icon size is too small, but that's a design decision.

packages/components/src/components/hds/dropdown/toggle/icon.js
```diff
@@ -77,10 +77,18 @@ export default class HdsDropdownToggleIconComponent extends Component {
...
+    } else {
+      // this is the default size (notice: for the "small" variant with chevron, we set the actual size to `12px` via CSS)
+      return '16';
+    }
+      // if hasChevron is true, we resize the icon to 12px in CSS
       return '16';
     } else {
-      return '24';
+      // if hasChevron is true, we use 16px icon
+      return this.hasChevron ? '16' : '24';
```
KristinLBradley, line 91: Going to need to clean this up after applying the suggested change...

PR #2183, opened 2024-06-24
[SPIKE] Hds 3402 icon

zamoore:
🚨🚨 🚨 !!! - DO NOT MERGE - !!! 🚨🚨 🚨

Exploration spike only!

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist


packages/codemods/transforms/v4/icon/index.js
```diff
@@ -0,0 +1,43 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+module.exports = function ({ source /*, path*/ }, { parse, visit }) {
+  const ast = parse(source);
+
+  const updateIsInlineBlockAttribute = (attributes) => {
+    const isInlineBlockAttr = attributes.find((a) => a.name === '@isInlineBlock');
+
+    if (isInlineBlockAttr) {
```
didoo, line 12: quick consideration: 

_in theory_, since we're switching the logic, we need to assign a `@isInlineBlock` as `true` when there's not `@isInline` argument

_in practice_ this would cause a lot of noise in the consumers codebase, and I would advocate that we can simply ignore this and simply change the default (something to discuss in the RFC and gather data about, so we can justify our decision)

for context: I did a quick test in Cloud UI, and turned off this inline behaviour for the icons, and all the UI remained the same

packages/codemods/transforms/v4/icon/index.js
```diff
@@ -0,0 +1,43 @@
...
+      isInlineBlockAttr.name = '@isInline';
+      isInlineBlockAttr.value = { true: 'false', false: 'true' }[isInlineBlockAttr.value];
```
didoo, line 14: can you double check this, there's something not clicking for me 🤔

packages/components/src/styles/components/icon.scss
```diff
@@ -0,0 +1,41 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+.flight-icon {
+  display: block;
+}
+
+.flight-icon--is-inline {
+  display: inline;
```
didoo, line 11: TBD if this needs to remain `inline-block` or instead become `inline`.

packages/components/src/styles/components/icon.scss
```diff
@@ -0,0 +1,41 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+.flight-icon {
```
didoo, line 6: don't forget that now the namespace should be `hds-icon`

(plus, we have to understand which of the previous `flight-icon` class names we want to preserve, because they're [used in the consumers codebases](https://github.com/search?q=flight-icon++user%3Ahashicorp+-repo%3Ahashicorp%2Fdesign-system+-repo%3Ahashicorp%2Fdesign-system-metrics&type=code&ref=advsearch))

didoo (commented):
@zamoore I gave a high-level look at the code, and left some comments here and there for things you may want to consider in the context of the spike

zamoore:
Closing as spike is completed

PR #2181, opened 2024-06-24
Showcase - Add `@gap` argument to `Shw::Flex` and `Shw::Grid` components

didoo:
### :pushpin: Summary

This PR cleans up many of the `{{style}}` declarations for the `Shw::Flex` and `Shw::Grid` components by making the `gap` CSS value an official argument `@gap` for the component

### :hammer_and_wrench: Detailed description

In this PR I have:
- created a codemod  to automatically remove the `gap` value in the `{{style}}` modifier (if it existed) and replace it with a `@gap` argument
- removed extra parenthesis in declarations/computations (otherwise the `codemods/bin/cli.js` runner would complain)
- run the codemod on the showcase template pages (and fixed a value that was in `px` to use `rem` for consistency)
- updated `Shw:Flex` and `Shw::Grid` components to support `@gap` argument
- removed the codemod (not necessary anymore) 

***

### 👀 Component checklist

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #2187, opened 2024-06-25
Add `gap` to `ShwFlexSignature` and `ShwGridSignature`

alex-ju:
### :pushpin: Summary

I missed these when reviewing #2181. These showcase components are not yet converted to TypeScript, but we have signatures for them to make it easier for us to write examples (via intellisense).

***


didoo (approved):
Oh I didn't know it (or maybe I knew it for some time, and then I forgot it)

PR #2207, opened 2024-07-02
Add `Hds::Icon` Component - Feature branch [MAIN]

zamoore:
### :pushpin: Summary

If merged, this PR will add the `Icon` component to `HDS`.

| Merge Order | Pull Request | Status | 
|--------|--------|--------|
| 1 | [Component](https://github.com/hashicorp/design-system/pull/2208) | 🔀 Merged |
| 2 | [Documentation](https://github.com/hashicorp/design-system/pull/2221/files) | 🔀 Merged |
| 3 | [Codemod](https://github.com/hashicorp/design-system/pull/2224/files) | 🔀 Merged |
| 4 | [@color extension](https://github.com/hashicorp/design-system/pull/2302/files) | 🔀 Merged |

### :hammer_and_wrench: Detailed description

- Adds a new `Hds::Icon` component that duplicates the behavior of the `FlightIcon` component. The one notable change from the `FlightIcon` component is the removal of the `isInlineBlock` attribute and its replacement with a new `@isInline` argument that defaults to `false`. We also extended the `@color` argument to support predefined "foreground" colors
- Updates `website` documentation to account for the new `Icon` component. The current `usage` docs are migrated to the new component docs with updates around the default display value.
- Adds a codemod to migrate the `<FlightIcon>` instances to `<Hds::Icon>`

### :link: External links

Jira ticket: [HDS-3404](https://hashicorp.atlassian.net/browse/HDS-3404)

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [ ] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#templates-for-npm-packages))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-3404]: https://hashicorp.atlassian.net/browse/HDS-3404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2208, opened 2024-07-02
`Hds::Icon` Component

zamoore:
### :pushpin: Summary

This PR is part of https://github.com/hashicorp/design-system/pull/2207

If merged, this PR will:

- Create the `Hds::Icon` component

### :link: External links

Feature branch PR: [2207](https://github.com/hashicorp/design-system/pull/2207)
Jira ticket: [HDS-3515](https://hashicorp.atlassian.net/browse/HDS-3515)

***

[HDS-3515]: https://hashicorp.atlassian.net/browse/HDS-3515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/icon/index.ts
+++ b/packages/components/src/components/hds/icon/index.ts
@@ -10,7 +10,7 @@ import { iconNames } from '@hashicorp/flight-icons/svg';
 
 import type { IconName } from '@hashicorp/flight-icons/svg';
 
-export interface FlightIconSignature {
+export interface IconSignature {
   Args: {
     name?: IconName;
     color?: string;
@@ -22,15 +22,15 @@ export interface FlightIconSignature {
   Element: SVGElement;
 }
 
-export default class FlightIcon extends Component<FlightIconSignature> {
-  constructor(owner: unknown, args: FlightIconSignature['Args']) {
+export default class Icon extends Component<IconSignature> {
+  constructor(owner: unknown, args: IconSignature['Args']) {
     super(owner, args);
 
     if (!this.args.name) {
-      assert('Please provide to <FlightIcon> a value for @name');
+      assert('Please provide to <Hds::Icon> a value for @name');
     } else if (!iconNames.includes(this.args.name)) {
       assert(
-        `The icon @name "${this.args.name}" provided to <FlightIcon> is not correct. Please verify it exists on https://helios.hashicorp.design/icons/library`
+        `The icon @name "${this.args.name}" provided to <Hds::Icon> is not correct. Please verify it exists on https://helios.hashicorp.design/icons/library`
       );
     }
   }
--- a/packages/components/src/components/hds/icon/index.ts
+++ b/packages/components/src/components/hds/icon/index.ts
@@ -1,0 +1,167 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+import { guidFor } from '@ember/object/internals';
+import { assert } from '@ember/debug';
+import { iconNames } from '@hashicorp/flight-icons/svg';
+
+import type { IconName } from '@hashicorp/flight-icons/svg';
+
+export interface FlightIconSignature {
+  Args: {
+    name?: IconName;
+    color?: string;
+    size?: '16' | '24';
+    stretched?: boolean;
+    isInline?: boolean;
+    title?: string;
+  };
+  Element: SVGElement;
+}
+
+export default class FlightIcon extends Component<FlightIconSignature> {
+  constructor(owner: unknown, args: FlightIconSignature['Args']) {
+    super(owner, args);
+
+    if (!this.args.name) {
+      assert('Please provide to <FlightIcon> a value for @name');
+    } else if (!iconNames.includes(this.args.name)) {
+      assert(
+        `The icon @name "${this.args.name}" provided to <FlightIcon> is not correct. Please verify it exists on https://helios.hashicorp.design/icons/library`
+      );
+    }
+  }
+
+  get isInline() {
+    return this.args.isInline ?? false;
+  }
+
+  /**
+   * Sets the color for the SVG
+   *
+   * @param color {string}
+   * @default 'currentColor'
+   */
+  get color() {
+    return this.args.color ?? 'currentColor';
+  }
+
+  /**
+   * Generates a unique ID for the SVG
+   *
+   * @param iconId
+   */
+  iconId = 'icon-' + guidFor(this);
+
+  /**
+   * Indicates which icon should be used. An error (in the form of an assertion)
+   * will occur if a value has not been provided.
+   *
+   * @param name {string}
+   */
+  get name() {
+    return this.args.name;
+  }
+
+  /**
+   * Gets the icon's size (16 or 24)
+   *
+   * @param size
+   * @returns the value of `size` if set
+   * @default `16`
+   */
+  get size() {
+    return this.args.size ?? '16';
+  }
+
+  /**
+   * Get the SVG width/height depending if the icon is stretched or not
+   * @method FlightIcon#svgSize
+   * @return {object} The width/height to apply to the SVG.
+   */
+  get svgSize() {
+    return {
+      width: this.args.stretched ? '100%' : this.size,
+      height: this.args.stretched ? '100%' : this.size,
+    };
+  }
+
+  /**
+   * Generates a unique ID for the title
+   *
+   * @param titleId
+   */
+  titleId = 'title-' + guidFor(this);
+
+  /**
+   * Gets the icon's title if one is set
+   *
+   * @param title
+   * @returns the value of `title` if set
+   * @default null
+   */
+  get title() {
+    return this.args.title ?? null;
+  }
+
+  /**
+   *
+   * Sets a role if a title exists
+   *
+   * @param role {string}
+   * @returns 'img' or null
+   * @default null
+   */
+  get role() {
+    if (this.args.title) {
+      return 'img';
+    } else {
+      return null;
+    }
+  }
+  /**
+   *
+   * Sets aria-labelledby if a title exists
+   *
+   * @param ariaLabelledby {string}
+   * @returns value of titleId or null
+   * @default null
+   */
+  get ariaLabelledby() {
+    if (this.args.title) {
+      return this.titleId;
+    } else {
+      return null;
+    }
+  }
+
+  /**
+   * Get the class names to apply to the icon.
+   * @return {string} The "class" attribute to apply to the component.
+   */
+  get classNames() {
+    const classes = ['hds-icon'];
+
+    // add a class based on the @name argument
+    classes.push(`hds-icon-${this.name}`);
+
+    // add a class based on the @isInlineBlock argument
+    const { isInline = false } = this.args;
+
+    if (isInline) {
+      classes.push('hds-icon--is-inline');
+    }
+
+    // add an extra class to control the animation (depends on the icon)
+    if (this.name === 'loading') {
+      classes.push('hds-icon--animation-loading');
+    } else if (this.name === 'running') {
+      classes.push('hds-icon--animation-running');
+    }
+
+    return classes.join(' ');
+  }
+}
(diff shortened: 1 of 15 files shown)


packages/components/addon-main.cjs
```diff
@@ -4,4 +4,13 @@
  */
 
 const { addonV1Shim } = require('@embroider/addon-shim');
-module.exports = addonV1Shim(__dirname);
+const flightIconSprite = require('@hashicorp/flight-icons/svg-sprite/svg-sprite-module');
```
didoo, line 7: I see the same changes as in https://github.com/hashicorp/design-system/pull/2208
Please base this second PR on the first one (#2208) not on the feature branch

(plus, make sure the tests are passing)

Let me know once done, so I can continue the review

zamoore: All set now. Thanks for reviewing.

showcase/tests/integration/components/hds/icon/index-test.js
```diff
@@ -0,0 +1,135 @@
...
+  });
+  test('it renders the 16x16 icon by default', async function (assert) {
+    await render(hbs`<Hds::Icon @name="activity" />`);
+    assert
+      .dom('svg.hds-icon.hds-icon-activity')
+      .hasStyle({ height: '16px', width: '16px' })
+      .hasAttribute('width', '16')
+      .hasAttribute('height', '16');
+  });
+  test('it renders the 24x24 icon when option is set', async function (assert) {
```
didoo, line 31: small improvement to the test description
```suggestion
  test('it renders the 24x24 icon when the "size" option is set', async function (assert) {
```

showcase/tests/integration/components/hds/icon/index-test.js
```diff
@@ -0,0 +1,135 @@
...
+  test('it sets the width/height to 100% when the "stretched" option is set to true', async function (assert) {
+    await render(
+      hbs`<Hds::Icon @name="activity" @size="24" @stretched={{true}} />`
+    );
+    assert
+      .dom('svg.hds-icon.hds-icon-activity')
+      .hasAttribute('width', '100%')
+      .hasAttribute('height', '100%');
+  });
+  test('it does not have the "hds-icon--is-inline" class if the option is not set', async function (assert) {
```
didoo, line 48: small improvement to the test description
```suggestion
  test('it does not have the "hds-icon--is-inline" class by default', async function (assert) {
```

showcase/tests/integration/components/hds/icon/index-test.js
```diff
@@ -0,0 +1,135 @@
...
+  test('it does have the "hds-icon--is-inline" class if the @isInline option is set to true', async function (assert) {
+    await render(hbs`<Hds::Icon @name="activity" @isInline={{true}} />`);
+    assert.dom('svg.hds-icon').hasClass('hds-icon--is-inline');
+  });
+  test('it does not have the "hds-icon--is-inline" class if the "stretched" option is set to true', async function (assert) {
+    await render(
+      hbs`<Hds::Icon @name="activity" @isInline={{true}} @stretched={{true}} />`
+    );
+    assert.dom('svg.hds-icon').doesNotHaveClass('hds-icon--is-inline');
+  });
```
didoo, line 61: Now that `block` is the default, we don't need anymore this test
```suggestion
```

showcase/tests/integration/components/hds/icon/index-test.js
didoo, line 1: the showcase page for the `Hds::Icon` component is missing, as well as the visual regression test for it. 
check the spike branch and see if something else is missing in this PR for the `Hds::Icon`

didoo (commented):
Left a few comments

packages/components/src/components/hds/icon/index.ts
```diff
@@ -0,0 +1,164 @@
...
+   * Get the class names to apply to the icon.
+   * @return {string} The "class" attribute to apply to the component.
+   */
+  get classNames() {
+    const classes = ['hds-icon'];
+
+    // add a class based on the @name argument
+    classes.push(`hds-icon-${this.name}`);
+
+    if (this.isInline && !this.args.stretched) {
```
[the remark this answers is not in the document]
zamoore: What behavior do we want if a user sets both to true?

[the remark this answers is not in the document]
didoo: Essentially, there is not big of an impact if consumers apply both, while before we had to make sure (read: force) to have a `block` behaviour when the icon was stretched to fill the parent container. So, we can just let the browser behave as it should, when the consumers sets the display as "inline" but also uses the stretching (which probably is not even a real use case)

showcase/app/templates/components/icon.hbs
```diff
@@ -0,0 +1,135 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+{{page-title "Icon"}}
+
+<Shw::Text::H1>Icon</Shw::Text::H1>
+
+<section data-test-percy>
```
zamoore, line 10: Is this enough for setting up tests in Percy?

showcase/app/styles/showcase-pages/icon.scss
```diff
@@ -4,8 +4,10 @@
  */
 
 // ICON
-
-body.foundations-icon {
+// The icon page is currently displayed in 2 places,
```
didoo, line 7: 👏

showcase/app/templates/components/icon.hbs
didoo: No, you have to add an entry in the file `showcase/tests/acceptance/percy-test.js`

showcase/app/templates/components/icon.hbs
```diff
@@ -0,0 +1,135 @@
...
+    <SF.Item @label="12px (stretched)">
+      <div {{style width="12px" height="12px"}}>
+        <Hds::Icon @name="bug" @size="16" @stretched={{true}} />
+      </div>
+    </SF.Item>
+    <SF.Item @label="32px (stretched)">
+      <div {{style width="32px" height="32px"}}>
+        <Hds::Icon @name="bug" @size="24" @stretched={{true}} />
+      </div>
+    </SF.Item>
```
didoo, line 37: suggestion
in this way we test also the combination `stretched` + `isInline` and make sure it still works as expected
```suggestion
    </SF.Item>
    <SF.Item @label="32px (stretched + inline)">
      <div {{style width="32px" height="32px"}}>
        <Hds::Icon @name="bug" @size="24" @stretched={{true}} @isInline={{true}} />
      </div>
    </SF.Item>
```

didoo (approved):
Left a few non-blocking comments. Apart from that LGTM 👍

packages/components/src/components/hds/icon/index.ts
```diff
@@ -0,0 +1,164 @@
...
+        `The icon @name "${this.args.name}" provided to <Hds::Icon> is not correct. Please verify it exists on https://helios.hashicorp.design/icons/library`
+      );
+    }
+  }
+
+  get isInline() {
+    return this.args.isInline ?? false;
+  }
+
+  /**
```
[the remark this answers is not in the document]
alex-ju: yes, I would not add JSDoc comments at this point

packages/components/src/components/hds/icon/index.ts
```diff
@@ -0,0 +1,164 @@
...
+import { assert } from '@ember/debug';
+import { iconNames } from '@hashicorp/flight-icons/svg';
+
+import type { IconName } from '@hashicorp/flight-icons/svg';
+
+export interface HdsIconSignature {
+  Args: {
+    name?: IconName;
+    color?: string;
+    size?: '16' | '24';
```
alex-ju, line 17: question: should we extract this as an enum to be able to use it in other components? or do we prefer to reference it from here (e.g. `HdsIconSignature['Args']['size]`)?

packages/components/src/components/hds/icon/index.ts
```diff
@@ -0,0 +1,164 @@
...
+import Component from '@glimmer/component';
+import { guidFor } from '@ember/object/internals';
+import { assert } from '@ember/debug';
+import { iconNames } from '@hashicorp/flight-icons/svg';
+
+import type { IconName } from '@hashicorp/flight-icons/svg';
+
+export interface HdsIconSignature {
+  Args: {
+    name?: IconName;
```
alex-ju, line 15: question: isn't `@name` a required argument?

packages/components/src/components/hds/icon/index.ts
```diff
@@ -0,0 +1,164 @@
...
+
+  /**
+   * Gets the icon's size (16 or 24)
+   *
+   * @param size
+   * @returns the value of `size` if set
+   * @default `16`
+   */
+  get size() {
+    return this.args.size ?? '16';
```
alex-ju, line 77: question: should we reference the `16` value here from an enum (see my other comment)?

packages/components/src/components/hds/icon/index.ts
```diff
@@ -0,0 +1,164 @@
...
+    if (!this.args.name) {
+      assert('Please provide to <Hds::Icon> a value for @name');
+    } else if (!iconNames.includes(this.args.name)) {
+      assert(
+        `The icon @name "${this.args.name}" provided to <Hds::Icon> is not correct. Please verify it exists on https://helios.hashicorp.design/icons/library`
+      );
+    }
+  }
+
+  get isInline() {
```
alex-ju, line 38: I believe we settled on adding return types for functions

packages/components/src/components/hds/icon/index.ts
```diff
@@ -0,0 +1,164 @@
...
+    if (this.isInline) {
+      classes.push('hds-icon--is-inline');
+    }
+
+    // add an extra class to control the animation (depends on the icon)
+    if (this.name === 'loading') {
+      classes.push('hds-icon--animation-loading');
+    } else if (this.name === 'running') {
+      classes.push('hds-icon--animation-running');
+    }
```
alex-ju, line 160: suggestion, in case we end up adding more animated icons
```suggestion
    if (this.name === 'loading' || this.name === 'running') {
      classes.push(`hds-icon--animation-${this.name}`);
    }
```

zamoore: I like it. Good suggestion

packages/components/src/components/hds/icon/index.ts
zamoore: What would we call the enum values? Is there another terminology for these sizes aside from a numerical size?

packages/components/src/components/hds/icon/index.ts
```diff
@@ -0,0 +1,164 @@
...
+   */
+  iconId = 'icon-' + guidFor(this);
+
+  /**
+   * Indicates which icon should be used. An error (in the form of an assertion)
+   * will occur if a value has not been provided.
+   *
+   * @param name {string}
+   */
+  get name() {
```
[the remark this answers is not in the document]
zamoore: Hah, good catch!

packages/components/src/components/hds/icon/index.ts
didoo: see these examples:
- https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/separator/types.ts#L7-L8
- https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/text/types.ts#L48-L52

didoo (approved):
👍

zamoore:
@alex-ju When you have a moment, can you taker another look please?

packages/components/src/components/hds/icon/index.ts
```diff
@@ -0,0 +1,164 @@
...
+    } else {
+      return null;
+    }
+  }
+
+  /**
+   * Get the class names to apply to the icon.
+   * @return {string} The "class" attribute to apply to the component.
+   */
+  get classNames() {
```
alex-ju, line 74:
```suggestion
  get classNames(): string {
```

alex-ju (approved):
Excellent work @zamoore!

PR #2212, opened 2024-07-04
Remove references to `dialog-polyfill` in documentation pages for `Flyout` and `Modal`

didoo:
### :pushpin: Summary

Since the `dialog-polyfill` was removed in https://github.com/hashicorp/design-system/pull/1977 we can also remove references to it in the documentation pages of the `Flyout` and `Modal` components.

***

### 👀 Component checklist


alex-ju (approved):
Good catch! For some reason, I thought we did it back then

PR #2069, opened 2024-04-30
`RichTooltip/Popover` component - Feature branch [MAIN]

didoo:
### :pushpin: Summary

This PR will contain all the sub-PRs opened for the `RichTooltip/Popover` component implementation.

It's mainly a porting (with cleanup/refactoring) of #1947 

### :hammer_and_wrench: Detailed description

This PR is a container for the following sub-PRs:
- ~~https://github.com/hashicorp/design-system/pull/2020~~
    - already merged in `main`
- ~~https://github.com/hashicorp/design-system/pull/2017~~
    - already merged in `main`
- https://github.com/hashicorp/design-system/pull/2045
- https://github.com/hashicorp/design-system/pull/2046
- https://github.com/hashicorp/design-system/pull/2048
- https://github.com/hashicorp/design-system/pull/2040

### 👓 Preview

- [PopoverPrimitive / Showcase](https://hds-showcase-git-00-popover-feature-branch-hashicorp.vercel.app/utilities/popover-primitive)
- [PopoverPrimitive / Documentation](https://hds-website-git-00-popover-feature-branch-hashicorp.vercel.app/utilities/popover-primitive)
- [RichTooltip / Showcase](https://hds-showcase-git-00-popover-feature-branch-hashicorp.vercel.app/components/rich-tooltip)
- [RichTooltip / Documentation](https://hds-website-git-00-popover-feature-branch-hashicorp.vercel.app/components/rich-tooltip)

### :link: External links

Jira tickets:
- https://hashicorp.atlassian.net/browse/HDS-1742 (Epic)
- https://hashicorp.atlassian.net/browse/HDS-3211 (Task)

Figma file: https://www.figma.com/file/noyY6dUMDYjmySpHcMjhkN/branch/ZiYg2uadc2fCNQ5tMzQqi8/HDS-Product---Components?type=design&node-id=55408-54938&mode=design&t=qDUIxW9WBWLLM1HG-0

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [x] Copyright headers have been added
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#templates-for-npm-packages))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #2214, opened 2024-07-04
Remove Popover API polyfill

didoo:
### :pushpin: Summary

Now that [Firefox officially supports the Popover API](https://caniuse.com/?search=popover) we don't need anymore a polyfill (it was introduced in https://github.com/hashicorp/design-system/pull/2069 for the RichTooltip)

screenshot_3921

In this PR I have:
- remove `@oddbird/popover-polyfill` dependency
- remove code related to `@oddbird/popover-polyfill`
- removed reference to `@oddbird/popover-polyfill` in documentation

Notice: as [agreed with the team previously](https://hashicorp.atlassian.net/browse/HDS-3276?focusedCommentId=452192) "we will consider the removal of the polyfill a non-breaking change, since we are supposed to support the last 2 major versions, and by July there will be probably 4 version supporting the API" so I've used a `patch` in the changeset, similar to what we did in https://github.com/hashicorp/design-system/pull/1977

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-3276

***

### 👀 Component checklist

### DIFF
--- a/.changeset/new-buckets-fry.md
+++ b/.changeset/new-buckets-fry.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Removed `popover-polyfill` dependency and instantiating code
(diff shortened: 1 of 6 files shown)


alex-ju (approved):
Clean removal 👏

PR #2222, opened 2024-07-09
Replaced `@required="true"` with `@required={{true}}` in `<Doc::ComponentApi>` properties declarations

didoo:
### :pushpin: Summary

Small PR to fix the way in which we declare the `@required` argument in the `<Doc::ComponentApi>` properties declarations (for context, if we declared `@required="false"` that would be interpreted as `true`)

***

### 👀 Component checklist


alex-ju (approved):
🙌

PR #2223, opened 2024-07-09
Hds 1366 prevent showdown wrapping everthing in a p tag

curthashicorp:
### :pushpin: Summary

Explore how to prevent `<p class="doc-markdown-p">` wrapping of custom elements ( Doc::  Hds::  ) in Showdown markdown transformation process.
Exclude custom elements that render inline content (we want those to remain encapsulated)


curthashicorp:
Will create a new PR with a clean git history.

PR #2217, opened 2024-07-05
Convert Form primitives to TypeScript

alex-ju:
### :pushpin: Summary

Convert Form primitives to TypeScript, enabling the conversion of Form components.

### :hammer_and_wrench: Detailed description

I added `@ts-expect-error` for the two classes using the `@ariaDescribedBy` decorator as it failed `lint:types` for `Form::Field` and `Form::Fieldset` and `@glint-expect-error` to `this.ariaDescribedBy` for the same reason. I am open to suggestions on this front, but I wouldn't block the form conversions for this alone.

<details>
<summary>console log</summary>

```log
src/components/hds/form/field/index.ts:63:2 - error TS1270: Decorator function return type 'ClassOf<AriaDescribedByComponent>' is not assignable to type 'void | typeof HdsFormFieldComponent'.
   Types of construct signatures are incompatible.
     Type 'new () => AriaDescribedByComponent' is not assignable to type 'new (owner: unknown, args: { extraAriaDescribedBy?: string | undefined; contextualClass?: string | undefined; isOptional?: boolean | undefined; isRequired?: boolean | undefined; layout?: "flag" | "vertical" | undefined; }) => HdsFormFieldComponent'.
       Type 'AriaDescribedByComponent' is missing the following properties from type 'HdsFormFieldComponent': appendDescriptor, removeDescriptor, layout, id, and 3 more.
```

</details>

### :link: External links

Jira ticket: [HDS-2692](https://hashicorp.atlassian.net/browse/HDS-2692)

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#templates-for-npm-packages))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-2692]: https://hashicorp.atlassian.net/browse/HDS-2692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2232, opened 2024-07-10
Fix way in which the `isOptional/isRequired` form primitive arguments are documented

didoo:
### :pushpin: Summary

While reviewing https://github.com/hashicorp/design-system/pull/2217/ I noticed an unusual way to document the default for `isOptional/isRequired` form primitive arguments, and since it was an easy fix I opened this PR.

***

### 👀 Component checklist


alex-ju (approved):
Thanks for updating these! They may have been the result of our automation script when we ported the properties to this format.

PR #2236, opened 2024-07-10
Website - Remove `<Doc::ListContainer>` component

didoo:
### :pushpin: Summary

While reviewing https://github.com/hashicorp/design-system/pull/2223/ I notice that @curthashicorp had to consider also the `<Doc::ListContainer>` component. Since this is used only in a specific context, and it's nothing more than a `<ul>` tag element, I'm removing it.

***

### 👀 Component checklist


MelSumner (approved):
Nice catch, and good idea.

PR #2241, opened 2024-07-12
`FlightIcon` -> `Hds::Icon` Migration

zamoore:
### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/components/hds/rich-tooltip/toggle.hbs
+++ b/packages/components/src/components/hds/rich-tooltip/toggle.hbs
@@ -17,13 +17,13 @@
     {{yield}}
   {{~else~}}
     {{~#if (and @icon (eq this.iconPosition "leading"))~}}
-      <Hds::Icon class="hds-rich-tooltip__toggle-icon" @name={{@icon}} />
+      <Hds::Icon class="hds-rich-tooltip__toggle-icon" @name={{@icon}} @isInline={{this.isInline}} />
     {{~/if~}}
     {{~#if @text~}}
       <span class="hds-rich-tooltip__toggle-text">{{~@text~}}</span>
     {{~/if~}}
     {{~#if (and @icon (eq this.iconPosition "trailing"))~}}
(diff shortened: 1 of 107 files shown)


packages/components/src/components/hds/rich-tooltip/toggle.hbs
```diff
@@ -17,13 +17,13 @@
...
   {{~else~}}
     {{~#if (and @icon (eq this.iconPosition "leading"))~}}
-      <FlightIcon class="hds-rich-tooltip__toggle-icon" @name={{@icon}} @isInlineBlock={{this.isInline}} />
+      <Hds::Icon class="hds-rich-tooltip__toggle-icon" @name={{@icon}} />
     {{~/if~}}
     {{~#if @text~}}
       <span class="hds-rich-tooltip__toggle-text">{{~@text~}}</span>
     {{~/if~}}
     {{~#if (and @icon (eq this.iconPosition "trailing"))~}}
-      <FlightIcon class="hds-rich-tooltip__toggle-icon" @name={{@icon}} @isInlineBlock={{this.isInline}} />
```
zamoore, line 26: The codemod doesn't appear to work on dynamic values for `@isInlineBlock`

didoo: What would you expect the codemod to do in this case?

zamoore: It was removing the attribute, I have since updated the codemod so that it will update correctly. I also added another test case for this scenario.

didoo: Ahh, ok. Now I understand :)

didoo: Quick question: since you've found this use case that was not covered, what do you think to do a dry run test on Cloud UI to see if other use cases are not covered? (I don't think there are, but just to be on the safe side)

zamoore:
Closing in favor of larger feature PR

PR #2252, opened 2024-07-16
[SPIKE] HDS 3542 - `ApplicationState` Updates

zamoore:
!!! SPIKE - DO NOT MERGE !!!

### :pushpin: Summary

This PR is meant as an initial development spike. It isn't meant to be merged & should only be used for the sake of discussion.

This updates the ApplicationState component to include a new yielded Media component as well as some basic design changes.

- Adds a yielded `Media` child component that takes no arguments. Yields to any content.
- Moves the error code above the title/icon line
- Changes the vertical spacing between the `footer` and `body` components
- Adds a new `align` (`left` or `center`) argument that changes the orientation of content
- Sets a max width of content inside the component (other than the media component) of 480px.
- Automatically orients the direct children of the footer element as follows
  - The first two children will display on the left side of the footer automatically.
  - Any other children will be displayed on the right.

With media, left aligned (default)
Screenshot 2024-07-18 at 4 26 34 PM

With media, center aligned
Screenshot 2024-07-18 at 4 27 04 PM

Without media
Screenshot 2024-07-18 at 4 27 38 PM

### :link: External links

Jira ticket: [HDS-3542](https://hashicorp.atlassian.net/browse/HDS-3542)
Figma file: [Figma](https://www.figma.com/design/noyY6dUMDYjmySpHcMjhkN/branch/PHekpm994tyTcf1fIY2Pgn/HDS-Components?node-id=61659-18071&t=MpLYk9U5AexVSpLw-0)

[HDS-3542]: https://hashicorp.atlassian.net/browse/HDS-3542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/styles/components/application-state.scss
```diff
@@ -7,43 +7,96 @@
...
-  gap: 8px;
-  justify-content: space-between;
+  grid-column: 1;
+  grid-row: 4;
+
+  padding-top: 12px;
+
+  & > * {
+    float: left;
+    margin-right: $hds-application-state-footer-child-gap;
```
KristinLBradley, line 91: I would use `display: flex` with `gap` for spacing vs margins. I don't think there's a reason to use float anymore for this type of layout in modern browsers which we support.

Float should be reserved for cases in which you want text content to actually flow around floated content such as an image. (Which was the original intention/purpose of float before it was co-opted for general content layout use before the advent of flexbox along with widespread browser support.)

packages/components/src/styles/components/application-state.scss
```diff
@@ -7,43 +7,96 @@
...
 
-$hds-application-state-padding: 12px 0;
+$hds-application-state-content-max-width: 480px;
+$hds-application-state-footer-child-gap: 8px;
 
 .hds-application-state {
-  width: 19.5rem;
+  display: grid;
+  grid-template-columns: minmax(0, $hds-application-state-content-max-width) 1fr 1fr;
+  grid-template-rows: auto auto auto auto;
```
KristinLBradley, line 16: The overall ApplicationState layouts (with left vs. centered alignment) are fairly simple so to me using grid to lay everything out seems overly complex and probably not necessary.

packages/components/src/styles/components/application-state.scss
```diff
@@ -7,43 +7,96 @@
...
+  }
 }
 
 .hds-application-state__body {
-  padding: $hds-application-state-padding;
-  color: var(--token-color-foreground-faint);
+  grid-column: 1;
+  grid-row: 3;
+  
+  padding: 12px 0;
```
KristinLBradley, line 79: I know the original component uses top and bottom padding to separate the "body" from the "header" and "footer" but I would confirm if this is actually desired since the bottom padding is always present even if the component is used without a footer. (So it could possibly have been a mistake to always include bottom padding.)

Also, `margin` should normally be used (vs. `padding`) for spacing apart different items while padding is used to add inner spacing around an item's content (for example if an element has a colored background, padding can be used to prevent text inside from going to the edge of the container.)

(Margin vs. padding usage is based upon my personal understanding and philosophy so other members of the team may have differing opinions...)

showcase/app/styles/showcase-pages/application-state.scss
```diff
@@ -0,0 +1,15 @@
...
+ */
+
+// APPLICATION-STATE
+
+body.application-state {
+  .shw-avatar {
+    height: 125px;
+    width: 125px;
+    border-radius: 100%;
+    overflow: hidden;
```
KristinLBradley, line 13: `overflow: hidden` shouldn't be needed as images don't have content which can overflow. (It _is_ possible for alt text to overflow but only if overflow is purposely set to `visible`.)

KristinLBradley (commented):
I left some initial comments & suggestions.

packages/components/src/styles/components/application-state.scss
zamoore: Flex doesn't work in this case specifically because we want the first 2 items on the left, and any remaining items on the right. This is the only way I could figure out how to make this work with dynamic content and without modifying the structure of the component. If you have a way you can make this work with flexbox, I would much rather use it. I also played around with a few implementations with css-grid to no avail.

KristinLBradley:
You have a linting error currently btw. Run `yarn lint` within the showcase directory locally.

packages/components/src/styles/components/application-state.scss
zamoore: This was the best way I could come up with for this alignment without modifying the structure or adding any additional elements.

I did play around with flexbox but didn't come to a conclusion that satisfied the design requirements.

MelSumner:
Some accessibility considerations:

- DOM order should match visual order 
- Consider how moving something above the title will affect how it's read via screen reader
- What will the accessible name (AccName) be?

I'll leave more thorough comments next week but just some initial thoughts to consider.

packages/components/src/styles/components/application-state.scss
didoo: Also, the current implementation is not working as expected

Before:
screenshot_3984

After
screenshot_3983

This for different reasons:
- a grid layout fills the parent container by default, while the previous implementation had a fixed width
- the grid colums don't match the content

I agree with @KristinLBradley that before jumping to using a CSS Grid, probably an attempt to see if the simple CSS Flexbox would do the trick (I mean for the root container of the ApplicationState)

One suggestion I have is: in the showcase, leave the content as it was before, and add below the "new" content, so you can spot if you're introducing breaking changes in the layout.

packages/components/src/styles/components/application-state.scss
didoo: +1 on using margin (and no, margin vs padding usage in the context of a design system is not based upon personal understanding, it has become a best practice de facto)

didoo (commented):
@zamoore I've left a couple of comments (well, mainly one) but since I don't know how much early in the spike you are, I don't know exactly what to point out and what not.

My suggestion is to use the showcase as a playground (a-la-TDD, but for UI) and incrementally make changes. 

If you want help or rubberduck or you want to pair, I am always available.

packages/components/src/styles/components/application-state.scss
KristinLBradley: Followup: Zack and I discussed together over Slack and I recommended using margin-left: auto to push over the 3rd (and following) links to the right. ([Example CodePen](https://codepen.io/KristinB/pen/PorZgRE/7a6266aae97137a707e27bbc182c8b5d))

packages/components/src/styles/components/application-state.scss
KristinLBradley: I was just trying to avoid stepping on anyone's toes if they disagreed for some reason...

didoo (commented):
Left a bunch of comments, haven't reviewed yet in depth (just a quick glance)

packages/components/src/styles/components/application-state.scss
```diff
@@ -7,43 +7,86 @@
...
+  color: var(--token-color-foreground-primary);
+  width: 100%;
+  max-width: $hds-application-state-content-max-width;
 }
 
 .hds-application-state__footer {
   display: flex;
   gap: 8px;
-  justify-content: space-between;
+  margin: 12px auto 0;
```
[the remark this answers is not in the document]
zamoore: Some old code from changing the gap between the footer and the body. Fixed.

packages/components/src/styles/components/application-state.scss
```diff
@@ -7,43 +7,89 @@
...
+    grid-column: 1 / -1;
+  }
 }
 
 .hds-application-state__body {
-  padding: $hds-application-state-padding;
-  color: var(--token-color-foreground-faint);
+  width: 100%;
+  max-width: $hds-application-state-content-max-width;
+  margin: 12px auto;
```
majedelass, line 75: Design reference: 
image

```suggestion
  margin: 12px auto 24px;
```

majedelass (commented):
Updated two areas that stood out to me.

packages/components/src/styles/components/application-state.scss
didoo: @majedelass what happens when there is a divider?

zamoore: The body has bottom-padding of 12px where the footer also has a top margin of 12px. Currently these together create a 24px gap. When the border is introduced, the gap is technically 25px (12px/1px/12px).

zamoore:
Spike completed, closing PR

PR #2146, opened 2024-06-05
feat(components): reexport all components that have backing class

aklkv:
### :link: External links

Jira ticket: [HDS-XXX](https://hashicorp.atlassian.net/browse/HDS-XXX)
Figma file: [if it applies]

***

### 👀 Component checklist

- [ ] Percy was checked for any visual regression
- [ ] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#templates-for-npm-packages))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #2253, opened 2024-07-17
Add missing component re-exports

alex-ju:
### :pushpin: Summary

I realised that we forgot about adding these re-exports for a little while.  Also revealed a naming inconsistency in the MaskedInput.

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/components.ts
+++ b/packages/components/src/components.ts
@@ -18,15 +18,56 @@ import HdsCopyButton from './components/hds/copy/button/index.ts';
 import HdsCopySnippet from './components/hds/copy/snippet/index.ts';
 import HdsDisclosurePrimitive from './components/hds/disclosure-primitive/index.ts';
 import HdsDismissButton from './components/hds/dismiss-button/index.ts';
+import HdsFormCharacterCount from './components/hds/form/character-count/index.ts';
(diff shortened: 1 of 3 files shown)


didoo:
@alex-ju does this contain all the changes that @aklkv did in https://github.com/hashicorp/design-system/pull/2146/files? if so, we can close the other then.

alex-ju:
> @alex-ju does this contain all the changes that @aklkv did in https://github.com/hashicorp/design-system/pull/2146/files? if so, we can close the other then.

This one only contains TypeScript files, and I'm tempted to leave it like that. But if you feel otherwise I'm happy to add the JavaScript files as well, all at once..

didoo:
> > @alex-ju does this contain all the changes that @aklkv did in https://github.com/hashicorp/design-system/pull/2146/files? if so, we can close the other then.
> 
> This one only contains TypeScript files, and I'm tempted to leave it like that. But if you feel otherwise I'm happy to add the JavaScript files as well, all at once..

OK, I see. Never mind.

didoo (approved):

aklkv:
@alex-ju I was contemplating an idea to add to add all components ts/js but in case of `js` it comes with a lot of errors from glint. So I would say your approach is totally valid as all new ts components would have valid signatures

PR #2270, opened 2024-07-23
Badge: update @text arg to allow numbers

shleewhite:
### :pushpin: Summary

If merged, people will be able to pass `numbers` for the `@text` argument on the Badge component

### :hammer_and_wrench: Detailed description

* updates the badge component signature to include number
* update the website documentation to match the new type
Screenshot 2024-07-23 at 3 56 46 PM

### :link: External links

Jira ticket: [HDS-3567](https://hashicorp.atlassian.net/browse/HDS-3567)

***

### 👀 Component checklist

- [ ] Percy was checked for any visual regression
- [ ] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#templates-for-npm-packages))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-3567]: https://hashicorp.atlassian.net/browse/HDS-3567?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2198, opened 2024-06-28
`Hds::Icon` Groundwork

zamoore:
### :pushpin: Summary

This PR is part of https://github.com/hashicorp/design-system/pull/2207

If merged, this PR will:

- Install `@hashicorp/flight-icons` as a direct dependency of `components`
- Port the load-sprite instance initialization functionality over from `ember-flight-icons` to `components`
- Ensure that the load-sprite initializer only runs once. (In existing and new instance initializers)

### :link: External links

Feature branch PR: [2207](https://github.com/hashicorp/design-system/pull/2207)
Jira ticket: [HDS-3515](https://hashicorp.atlassian.net/browse/HDS-3515)

***

[HDS-3515]: https://hashicorp.atlassian.net/browse/HDS-3515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2271, opened 2024-07-23
Website - Add status and version history to recently updated components

didoo:
### :pushpin: Summary

Small PR to add status ("updated") and version history for `Form::TextInput` (#2198) and `Badge` (#2270)

👉 👉 👉 Previews:
- [Badge](https://hds-website-git-website-add-missing-histories-hashicorp.vercel.app/components/badge)
- [Form::TextInput](https://hds-website-git-website-add-missing-histories-hashicorp.vercel.app/components/form/text-input)

***

### 👀 Component checklist

### DIFF
--- a/website/docs/components/badge/index.md
+++ b/website/docs/components/badge/index.md
@@ -9,6 +9,8 @@ related: ['components/badge-count','components/tag']
 previewImage: assets/illustrations/components/badge.jpg
 navigation:
   keywords: ['chip', 'pill', 'tag', 'label']
+status:
+  updated: 4.7.0
 ---
 
 <section data-tab="Guidelines">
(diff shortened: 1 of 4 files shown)


shleewhite (approved):

MelSumner (commented):
(question, non-blocker) what's our process for adding this information? Can you share in our slack chat if we have it documented? 

I see that it's for 4.7.0 but only 4.6.0 has been released. Do we have a process in place to make sure this isn't merged before release? 

Also, do we have a way to automate this at all?

didoo:
> (question, non-blocker) what's our process for adding this information? Can you share in our slack chat if we have it documented?

@MelSumner It's the first time we do (it's a **very** recent thing) I so I am acting as "guinea pig". Once I've done it a few times, and the process is "stable", I'll document it and share it with the whole team.

> I see that it's for 4.7.0 but only 4.6.0 has been released. Do we have a process in place to make sure this isn't merged before release?

We have to use the version that _will_ be released. In this case is `4.7.0`: https://github.com/hashicorp/design-system/pull/2238

> Also, do we have a way to automate this at all?

No, it needs to be done manually. Once we get used to it, it should become normal part of the process.

PR #2237, opened 2024-07-10
Hds-1366 Remove auto p tags around custom elements with showdown extension

curthashicorp:
### :pushpin: Summary

Explore how to prevent `<p class="doc-markdown-p">` wrapping of custom elements ( Doc::  Hds::  ) in Showdown markdown transformation process.
Exclude custom elements that render inline content (we want those to remain wrapped)

Doc Components without auto p tag can be reviewed here:
https://hds-website-git-hds-1366-showdown-extension-re-98bfba-hashicorp.vercel.app/testing/markdown/remove-auto-p-playground

### :link: External links

Jira ticket: [HDS-1366](https://hashicorp.atlassian.net/browse/HDS-1366)

### Evidence
Screenshot 2024-07-09 at 9 52 27 AM

### Before Fix
Screenshot 2024-07-11 at 3 22 58 PM
Screenshot 2024-07-11 at 3 24 52 PM

### After CSS Fix

Screenshot 2024-07-11 at 3 23 37 PM
Screenshot 2024-07-11 at 3 25 41 PM

[HDS-1366]: https://hashicorp.atlassian.net/browse/HDS-1366?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2277, opened 2024-07-24
fix spacing between hds-button(-set) and various components

curthashicorp:
### :pushpin: Summary

fix spacing between hds-button(-set) and various components

Stopping the default behavior of showdown [auto-wrapping all custom elements in a paragraph tag,](https://github.com/hashicorp/design-system/pull/2237) highlighted areas where spacing was not present where it should have been.

#### Before
Screenshot 2024-07-24 at 3 03 22 PM

#### After
Screenshot 2024-07-24 at 3 44 35 PM

#### Before
Screenshot 2024-07-24 at 2 58 12 PM

#### After
Screenshot 2024-07-24 at 2 58 20 PM

#### Before
Screenshot 2024-07-24 at 3 46 40 PM

#### After
Screenshot 2024-07-24 at 3 45 55 PM

#### Before
Screenshot 2024-07-24 at 3 48 39 PM

#### After
Screenshot 2024-07-24 at 3 49 15 PM

#### Before
Screenshot 2024-07-24 at 3 56 32 PM

#### After
Screenshot 2024-07-24 at 3 58 19 PM


MelSumner (approved):
It looks like you caught the issue, nice job!

didoo (commented):
Is the intention to start adding all the possible combinations and permutations of `Hds` and `Doc` components? They are a lot! (In principle, something around 100 x 10 potential combinations!)

We need a different, simpler, more scalable strategy. We already have the `Doc::Layout` component, that is used to... provide layout to elements in the docs pages, let's use that one.

Happy to discuss this in person, probably much easier.

curthashicorp:
> Is the intention to start adding all the possible combinations and permutations of `Hds` and `Doc` components? They are a lot! (In principle, something around 100 x 10 potential combinations!)
> 
> We need a different, simpler, more scalable strategy. We already have the `Doc::Layout` component, that is used to... provide layout to elements in the docs pages, let's use that one.
> 
> Happy to discuss this in person, probably much easier.

Yes much cleaner this way!

website/docs/components/accordion/partials/code/how-to-use.md
```diff
@@ -101,12 +101,12 @@ With a link in the toggle block and a form in the content.
         </div>
       </:toggle>
       <:content>
-        <p>
+        <Doc::Layout @spacing="16px" @direction="vertical">
```
didoo, line 104: This is a code snippet that people will copy & paste, we can't use `Doc` components here. We just have to leave the `<p>` and let it collapse with the button. It's an example, it's OK even if it's not perfect.

didoo (commented):
Left a comment.

website/docs/components/accordion/partials/code/how-to-use.md
curthashicorp: Ah good point, will adjust.

curthashicorp: good point! I removed.

didoo (approved):
👍

PR #2282, opened 2024-07-25
Doc Image Caption Component

curthashicorp:
### :pushpin: Summary

Implement captions for images on the Helios website

### Figma design
figma-spec

### New Component
Screenshot 2024-07-25 at 5 23 01 PM

### :link: External links

Jira ticket: [HDS-3547](https://hashicorp.atlassian.net/browse/HDS-3547)
Figma file: [Image Captions](https://www.figma.com/design/Ky0qWjvHZR3je1lCBlzNB1/HDS-Website-%5BInitial-Build%5D?node-id=5246-198172&t=4rcAGdYpQRcAUEXF-0)

***

[HDS-3547]: https://hashicorp.atlassian.net/browse/HDS-3547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/components/doc/image-caption/index.hbs
+++ b/website/app/components/doc/image-caption/index.hbs
@@ -1,0 +1,3 @@
+<div class="doc-image-caption doc-text-body-small" ...attributes>
+  {{@text}}
+</div>
--- a/website/app/styles/doc-components/image-caption.scss
+++ b/website/app/styles/doc-components/image-caption.scss
@@ -1,0 +1,12 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// IMAGE-CAPTION
+
+.doc-image-caption {
+  margin-top: 4px;
+  color: var(--doc-color-gray-300);
+  font-style: italic;
+}
--- a/website/docs/testing/markdown/more-complex-images.md
+++ b/website/docs/testing/markdown/more-complex-images.md
@@ -4,7 +4,8 @@ title: More complex images
 
 A simple small image (using markdown syntax).
 
-![](/assets/testing/small_cat.png)
+![Captions improve user experience by adding context to images. They should be clear, concise, and visually distinct with proper styling like background colors, padding, and rounded corners. Effective captions enhance readability and engagement.](/assets/testing/small_cat.png)
+<Doc::ImageCaption @text="Captions improve user experience by adding context to images. They should be clear, concise, and visually distinct with proper styling like background colors, padding, and rounded corners. Effective captions enhance readability and engagement."/>
 
 A larger image (using markdown syntax).
 
@@ -12,11 +13,11 @@ A larger image (using markdown syntax).
 
 A larger image (using markdown syntax) scaled down to a specific size (uses a [custom syntax](https://github.com/showdownjs/showdown/wiki/Showdown-Options#parseimgdimensions)).
 
-![](/assets/testing/big_cat.png =120x120)
+![a scaled down kitten image](/assets/testing/big_cat.png =120x120)
 
 An image with alt text and title (using markdown syntax).
 
-![a nice kitten for you](/assets/testing/small_cat.png "this is a kitten")
+![each alt text should be unique](/assets/testing/small_cat.png "this is a kitten")
 
 An image with alt text and title (using markdown syntax) **and** size.
 
@@ -25,7 +26,7 @@ An image with alt text and title (using markdown syntax) **and** size.
 
 An reference style image with alt text and title (using markdown syntax) **and** size.
 
-![Alt text][cat]
+![Example of kitten image with a alt and a title][cat]
 
 [cat]: /assets/testing/small_cat.png =250x250  "Optional title attribute"
 
@@ -33,6 +34,4 @@ An reference style image with alt text and title (using markdown syntax) **and**
 
 Some images declared as inline `<img>` HTML.
 
-<img width="400" alt="kitten" src="/assets/testing/small_cat.png">
-
-<img width="1200" alt="kitten" src="http://placekitten.com/g/1200/800/">
+<img width="400" alt="kitten here in this image" src="/assets/testing/small_cat.png">
--- a/website/app/styles/doc-components/index.scss
+++ b/website/app/styles/doc-components/index.scss
@@ -16,6 +16,7 @@
 @import "font-helpers-list";
 @import "form";
 @import "icons-list";
+@import "image-caption";
 @import "layout";
 @import "link-with-icon";
 @import "meta-row";
--- a/website/docs/components/accordion/partials/guidelines/guidelines.md
+++ b/website/docs/components/accordion/partials/guidelines/guidelines.md
@@ -98,13 +98,11 @@ Use the `flush` variant where space is limited such as within a [Card](/componen
 
 The toggle accepts many different types of content, from text-based content to nested HDS components or generic content.
 
-#### Text
+![Toggle type with basic text](/assets/components/accordion/toggle-content-text.png =690x*)
+<Doc::ImageCaption @text="Toggle type with basic text"/>
 
-![Example of toggle content](/assets/components/accordion/toggle-content-text.png =690x*)
-
-#### Generic 
-
-![Example of toggle content](/assets/components/accordion/toggle-content-custom.png =690x*)
+![Toggle type with generic content](/assets/components/accordion/toggle-content-custom.png =690x*)
+<Doc::ImageCaption @text="Toggle type with generic content"/>
 
 ### Interactive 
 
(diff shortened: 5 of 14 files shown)


website/app/components/doc/image-caption/index.js
```diff
@@ -0,0 +1,17 @@
...
+
+import Component from '@glimmer/component';
+
+export default class DocImageCaptionComponent extends Component {
+  constructor() {
+    super(...arguments);
+  }
+
+  get classNames() {
+    const classes = ['doc-image-caption'];
```
didoo, line 14: When there is only one class name, or there is no "logic" associated with the class names, we write directly the classes in the Handlebars template. See for example here: https://github.com/hashicorp/design-system/blob/main/website/app/components/doc/npm-version.hbs#L6

website/app/components/doc/image-caption/index.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+
+export default class DocImageCaptionComponent extends Component {
+  constructor() {
```
didoo, line 9: We don't need a constructor here

website/app/components/doc/image-caption/index.js
didoo, line 1: Given that we don't need a constructor, and we can add the classname directly to the template, you can completely remove this file

website/app/components/doc/image-caption/index.hbs
```diff
@@ -0,0 +1,3 @@
+<div class={{this.classNames}} ...attributes>
```
didoo, line 1: See comments below
```suggestion
<div class="doc-image-caption doc-text-body-small" ...attributes>
```

website/app/styles/doc-components/image-caption.scss
```diff
@@ -0,0 +1,16 @@
...
+// IMAGE-CAPTION
+
+@use "../typography/mixins" as *;
+
+.doc-image-caption {
+    @include doc-font-style-body-small();
+    margin-top: 4px;
+    color: var(--doc-color-gray-300);
+    font-style: italic;
+    letter-spacing: 0.0125em;
```
didoo, line 15: We don't change the letter spacing or kerning in the fonts we use (it's a Figma thing)
```suggestion
```

website/app/components/doc/image-caption/index.hbs
```diff
@@ -0,0 +1,3 @@
...
+  {{yield}}
```
didoo, line 2: @andgen404 do we allow to have also structured content inside a caption (eg. a bold text, a link, a list, etc) or should it be text-only?
- if it's text only, instead of yielding we should expose a `@text` argument to prevent passing structured content
- if we allow to pass structured content, then we should provide basic formatting (expecially for the link)

didoo: @curthashicorp on a second thought: I did this test:
```
<Doc::ImageCaption>
Lorem ipsum
- ipsum
- dolor
</Doc::ImageCaption>
```
and guess what?
screenshot_4016

So to avoid people adding markdown content that works in some cases but not others, better to be on the safe side and use just a `@text `argument, instead of the `yield`.

website/app/styles/doc-components/image-caption.scss
```diff
@@ -0,0 +1,16 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// IMAGE-CAPTION
+
+@use "../typography/mixins" as *;
+
+.doc-image-caption {
+    @include doc-font-style-body-small();
```
didoo, line 11: You can simplify it even further by using the class `doc-text-body-small` in the template
```suggestion
.doc-image-caption {
```

didoo (commented):
Left a few suggestions.

website/app/components/doc/image-caption/index.js
curthashicorp: Understood will follow this pattern going forward!

website/app/components/doc/image-caption/index.js
curthashicorp: Good point. Will follow this pattern going forward.

website/app/components/doc/image-caption/index.js
curthashicorp: 💯

website/app/components/doc/image-caption/index.hbs
curthashicorp: Understood will follow this pattern going forward!

website/app/styles/doc-components/image-caption.scss
curthashicorp: ok

website/app/components/doc/image-caption/index.hbs
curthashicorp: Great point.  This would add a lot of confusion for content authoring.  I have implement @text argument.

website/app/styles/doc-components/image-caption.scss
curthashicorp: Nice, don't even need the mixin! Keeping it simple.

MelSumner:
So when I use VoiceOver, and pull up the rotor (sometimes called the dialog box in some spec guidance), I see that some images do not have the `alt` attribute nor are they marked presentational. Others have duplicate `alt` attribute values, so they can't be distinguished from each other. 

`alt` attribute values should be unique and should describe what is in the image (nothing less, nothing more). The `alt` attribute should be present _and_ empty (e.g. `` if the image is presentational. 

Screenshot: 
CleanShot 2024-07-26 at 09 38 53

MelSumner:
It would be good to have the caption associated with the `img` element.

This is what I see:
CleanShot 2024-07-26 at 09 42 55


This is what I'd like to see: 
CleanShot 2024-07-26 at 09 44 45

This is what I would expect to see in the accessibility tree: 
CleanShot 2024-07-26 at 09 45 43

didoo:
Per [conversation](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1722000706499809) with @MelSumner @curthashicorp: given the time constraints, we agreed that an OK solution for this is to:

- we implement the caption as an independent text with custom styling
- content editors will make sure to add an alt text for the image (which is already what we do) which matches the caption

@curthashicorp in this PR can you also add the changes to the existing cases where there is a "caption" below an image (I think @heatherlarsen can point you to which cases [she was referring to](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1684791122475699?thread_ts=1684789012.006419&cid=C025N5V4PFZ))

website/app/components/doc/image-caption/index.hbs
```diff
@@ -0,0 +1,5 @@
+<div class="doc-image-caption doc-text-body-small" ...attributes>
+  {{#if @text}}
+    {{@text}}
+  {{/if}}
```
didoo, line 4: If you think about it, you don't need the if ;)

website/app/styles/doc-components/image-caption.scss
```diff
@@ -0,0 +1,12 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// IMAGE-CAPTION
+
+.doc-image-caption {
```
didoo, line 8: check the indentation (we use 2 spaces)

website/docs/testing/markdown/more-complex-images.md
```diff
@@ -5,6 +5,7 @@ title: More complex images
 A simple small image (using markdown syntax).
 
 ![](/assets/testing/small_cat.png)
+<Doc::ImageCaption @text="Captions improve user experience by adding context to images. They should be clear, concise, and visually distinct with proper styling like background colors, padding, and rounded corners. Effective captions enhance readability and engagement."></Doc::ImageCaption>
```
didoo, line 8: we can also use self-closing tags in Ember
```suggestion
<Doc::ImageCaption @text="Captions improve user experience by adding context to images. They should be clear, concise, and visually distinct with proper styling like background colors, padding, and rounded corners. Effective captions enhance readability and engagement." />
```

didoo (commented):
As mentioned in the other comment, make sure to make the changes to the existing use cases (so also we test the solution "[in context](https://helios.hashicorp.design/about/principles#:~:text=04-,Design%20in%20context,-We%20meet%20consumers)")

website/app/components/doc/image-caption/index.hbs
curthashicorp: smart

curthashicorp:
> So when I use VoiceOver, and pull up the rotor (sometimes called the dialog box in some spec guidance), I see that some images do not have the `alt` attribute nor are they marked presentational. Others have duplicate `alt` attribute values, so they can't be distinguished from each other.
> 
> `alt` attribute values should be unique and should describe what is in the image (nothing less, nothing more). The `alt` attribute should be present _and_ empty (e.g. `` if the image is presentational.
> 
@MelSumner Appreciate the review! I updated the test cases on this page.

didoo (approved):
👍 🎉

didoo:
For future reference, here's the list prepared by @heatherlarsen https://docs.google.com/document/d/12g9YGkTFFnHJG3Y6kzeC8ZyllW5Frl9ajizHyIyT_s8

PR #2295, opened 2024-07-29
`SegmentedGroup`, `Dropdown` - mark optional args

alex-ju:
### :pushpin: Summary

Mark optional args as such. We missed them in their conversion PRs, but we haven't released yet so we can patch without changelog entries.

***

### 👀 Component checklist


didoo (approved):
🤦‍♂️

PR #2298, opened 2024-07-29
fix: move keyboard focus when click scroll to top button

shleewhite:
### :pushpin: Summary

If merged, this PR would make it so when a user clicks the scroll back to top on the website, their keyboard focus moves to the `<main/>` element.

### 🛠️ Detailed description

I ended up changing the implementation from a button to an anchor. I initially made it where the button searched the DOM for all the interactive elements and focus the first one, but this had a few downsides:
* for some pages, when the first interactive element isn't close to the top, the scroll wouldn't go all the way up
* if there are no interactive elements, I tried programmatically focusing the main element but screen readers wouldn't consistently move the screen reader cursor
* overall, the approach seemed over complicated - especially when the skip to main content link works well and announces correctly.

Other changes: 
* set `aria-labelledby` for the `<main />` element on all pages except the home page to id of the h1

### :link: External links

Jira ticket: [HDS-3596](https://hashicorp.atlassian.net/browse/HDS-3596)

***

### 👀 Component checklist

[HDS-3596]: https://hashicorp.atlassian.net/browse/HDS-3596?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/components/doc/scroll-to-top/index.hbs
+++ b/website/app/components/doc/scroll-to-top/index.hbs
@@ -8,7 +8,6 @@
   aria-label="Scroll back to top"
   class="doc-scroll-to-top {{if this.isVisible 'doc-scroll-to-top--is-visible'}}"
   tabindex={{if this.isVisible 0 -1}}
-  {{on "click" this.scrollToTop}}
   {{doc-track-event eventName="Navigation - Scroll to Top"}}
 >
   <FlightIcon @name="arrow-up" @color="white" @size="24" @isInlineBlock={{false}} />
--- a/website/app/components/doc/scroll-to-top/index.hbs
+++ b/website/app/components/doc/scroll-to-top/index.hbs
@@ -6,8 +6,8 @@
 <a
   href="#main"
   aria-label="Scroll back to top"
-  class="doc-scroll-to-top {{if this.isLinkVisible 'doc-scroll-to-top--is-visible'}}"
-  tabindex={{if this.isLinkVisible 0 -1}}
+  class="doc-scroll-to-top {{if this.isVisible 'doc-scroll-to-top--is-visible'}}"
+  tabindex={{if this.isVisible 0 -1}}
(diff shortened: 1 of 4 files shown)


didoo (approved):
All good from the code perspective, so I am approving it.

One consideration, though, it that now with this focus shift, in some pages (eg. the [Principles](https://hds-website-git-shleewhite-hds-3596fix-scroll-to-top-hashicorp.vercel.app/about/principles)) the "Scroll to top" doesn't scroll to the very top anymore (at times I see it also in other pages too, but inconsistently).

website/app/components/doc/scroll-to-top/index.hbs
```diff
@@ -3,13 +3,13 @@
...
 }}
 
-<button
+<a
+  href="#main"
   aria-label="Scroll back to top"
-  class="doc-scroll-to-top {{if this.isButtonVisible 'doc-scroll-to-top--is-visible'}}"
-  type="button"
-  tabindex={{if this.isButtonVisible 0 -1}}
+  class="doc-scroll-to-top {{if this.isLinkVisible 'doc-scroll-to-top--is-visible'}}"
```
didoo, line 9: the fact that is a button or a link should be an implementation detail. what do you think of renaming the property to `isVisible` to avoid future refactorings if we change again our approach :)

didoo (approved):
I love this solution! 🙌 👏

(left only a suggestion, don't need to re-review if you make the changes)

website/app/components/doc/scroll-to-top/index.hbs
shleewhite: Good point! I'll update.

Dhaulagiri (approved):
👏  so fresh. so clean (clean).

PR #2299, opened 2024-07-29
`AppFrame`: “Enterprise Nav” phase 1.5 updates (HDS-3612)

KristinLBradley:
### APP FRAME UPDATES

---

### :pushpin: Summary

If merged, this PR updates the `AppFrame` to:

* Update `AppFrame` Showcase to add new framed example & longer content to test scrolling.
* Refactor positioning styles to move responsibility to `AppFrame` vs. child components.
* "Unstick" `AppHeader` when viewport height is under 480px in height.
* Style inoperable actions as disabled within `AppHeader` (when `SideNav` is expanded in mobile view).
* Remove the `withAppHeader` argument from the `SideNav` along with refactors removing the need for it.

👉 **AppFrame Showcase:** https://hds-showcase-git-hds-3612-app-frame-updates-hashicorp.vercel.app/layouts/app-frame
👉 [AppFrame "framed" example including AppHeader & SideNav components](https://hds-showcase-git-hds-3612-app-frame-updates-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-app-header)

**Related PR:**

* AppHeader updates PR: https://github.com/hashicorp/design-system/pull/2306

### :link: External links

* Jira ticket: [HDS-3612](https://hashicorp.atlassian.net/browse/HDS-3612)

***

### 👀 Component checklist

[HDS-3612]: https://hashicorp.atlassian.net/browse/HDS-3612?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


showcase/app/styles/showcase-components/frame.scss
```diff
@@ -6,9 +6,9 @@
 
 $shw-frame-navigation-bar-height: 48px;
 
-.shw-frame {
-  margin-bottom: 2rem;
-}
+// .shw-frame {
+//   margin-bottom: 2rem;
+// }
```
KristinLBradley, line 11: I updated AppFrame Showcase to use frame examples inside `Shw::Grid` so this style was adding an extra margin I didn't want/need. Would it be ok to delete this style?

showcase/app/styles/showcase-components/frame.scss
```diff
@@ -52,4 +52,11 @@ $shw-frame-navigation-bar-height: 48px;
   width: 100%;
   height: calc(100% - #{$shw-frame-navigation-bar-height});
   border: none;
+}
+
+// TESTING
+// Make frame wider so desktop view can be seen when included in a showcase page
+// TODO: Consider updating --main-width for all pages
+.showcase-app:has(.shw-frame) {
+  --main-width: calc(1088px + 2 * 2rem);
```
KristinLBradley, line 61: The framed examples displayed within the AppFrame Showcase weren't wide enough to see the Desktop view. You could only see Mobile view. I wondered if we could modify the width of the main content to make it 1088px wide?

(NOTE: When/if the responsive layout behavior of the AppHeader is refactored this may resolve this issue although it could still affect the SideNav examples if they continue to use a fixed breakoint of 1088px.)

showcase/app/templates/layouts/app-frame/index.hbs
```diff
@@ -182,7 +182,7 @@
 
   <Shw::Text::H2>Content</Shw::Text::H2>
 
-  <Shw::Grid @columns={{1}} {{style gap="2rem"}} as |SG|>
+  {{!-- <Shw::Grid @columns={{1}} {{style gap="2rem"}} as |SG|>
```
KristinLBradley, line 185: I was thinking that these non-framed examples may not be useful/needed so I tried commenting them out and leaving just the framed examples.

Let me know what you think and I'll either go ahead and delete these or else put them back.

packages/components/src/styles/components/side-nav/main.scss
```diff
@@ -8,25 +8,19 @@
...
+  // position: sticky;
+  // top: 0;
+  // z-index: 20; // needs to stay above the main content
   width: var(--hds-app-sidenav-width-fixed); // "default" width used by the `SideNav::Base` subcomponent (that is not responsive)
-  height: 100vh;
-  min-height: 100vh;
+  // height: 100vh;
+  // min-height: 100vh;
+  height: 100%;
+  min-height: 100%;
```
KristinLBradley, line 19: I'm testing moving the above styles related to the "sticky" layout to app-frame.scss styles for the "sidebar" container. (To give the AppFrame responsibility for the sticky/fixed layout of child components.)

KristinLBradley:
I added open questions I have to the top of the PR description. Please let me know your opinions and I can update this PR accordingly. Thanks.

jorytindall (commented):
For me the interaction between all of these elements looks good, the overlay isn't very perceivable when it's on top of the `AppHeader` (the elements show through a bit), but I think that's fine.

I'm not sure if it's directly related to this PR, but in testing on my phone the interactive (specifically `active`) states of the menu button and the collapse button are seeming to persist in an odd way (attached screen recording)

https://github.com/user-attachments/assets/d0a8d090-c892-46c1-85ac-376f442bf9a9

KristinLBradley:
> For me the interaction between all of these elements looks good, the overlay isn't very perceivable when it's on top of the `AppHeader` (the elements show through a bit), but I think that's fine.
> 
> I'm not sure if it's directly related to this PR, but in testing on my phone the interactive (specifically `active`) states of the menu button and the collapse button are seeming to persist in an odd way (attached screen recording)
> 
>  ScreenRecording_07-31-2024.14-42-47_1.MP4

@jorytindall Do you have any ideas on how it could be improved? To me the overlay just has a slight transition between visible/not visible which looks fine to me but if it bothers you how would you want it changed?

jorytindall:
> @jorytindall Do you have any ideas on how it could be improved? To me the overlay just has a slight transition between visible/not visible which looks fine to me but if it bothers you how would you want it changed?

Sorry, I'll clarify, I think the transition of the overlay is fine, if anything were to change it would maybe be increasing the opacity of the overlay (making it less opaque), but that's a larger discussion that would affect other things as well, so I'd say let's not touch it.

> I'm not quite getting what you mean by the active states persisting? I don't notice anything jumping out as overly weird personally.

I only noticed it on my phone, but when I press either of these buttons (menu or collapse), the persist in what looks like their hover state as I scroll around the page.

packages/components/src/styles/components/app-frame.scss
```diff
@@ -16,32 +16,47 @@
...
-    position: sticky;
-    top: 0;
-    right: 0;
-    left: 0;
-  }
 }
 
 .hds-app-frame__sidebar {
-  z-index: 6;
+  position: sticky;
```
didoo, line 50: See comment above. For me this should be an `opt-in` behaviour.

packages/components/src/styles/components/app-frame.scss
```diff
@@ -16,32 +16,47 @@
...
+  top: 0;
+  z-index: 20; // needs to stay above the main content
   grid-area: sidebar;
+  height: 100vh;
+  min-height: 100vh;
+
+  &:has(.hds-side-nav--is-not-minimized) {
+    z-index: 22; // makes overlay appear above header when sidenav is expanded to block interaction
```
didoo, line 58: This feels weird to me (and "semantically" not correct: the secondary navigation obscures/is above the primary navigation).

Definitely a designers' call, but that's my 2 cents.

didoo: Also if we remove this, it means we remove the `has(.hds-side-nav--is-not-minimized)` selector which "smells" for me (why should the `AppFrame` know about its children in this way? should be more agnostic, for me)

packages/components/src/styles/components/side-nav/main.scss
```diff
@@ -8,25 +8,13 @@
 //
 
 .hds-side-nav {
```
didoo, line 10: I would leave a `position: relative` here, so the `SideNav::ToggleButton` position is determined in relation to this component, not its parent.
```suggestion
.hds-side-nav {
  position: relative;
```

showcase/app/styles/showcase-components/frame.scss
didoo: When I implemented the `Shw::Frame` I imagined they would always be used in isolation, because of their nature, and because I would not imagine them used side by side. 
Hence the margin bottom. 

Why do you want to use them in a grid? Because of the label? In that case `Shw::Frame` supports a `@label` argument (as well as `@width` and `@height` if needed)

showcase/app/styles/showcase-components/frame.scss
didoo: No objections in principle. I have opened this PR https://github.com/hashicorp/design-system/pull/2312 to see what is the impact on the existing pages and the visual regressions.
If we are OK with it we can promote that to production, and you can remove this override here.

showcase/app/templates/layouts/app-frame/index.hbs
didoo: I see your point. Given that we can't do visual regression testing on them, but we may in the future (I just have to understand how to solve a problem in how the URLs are computed in the tests) if it's OK for you I would prefer to leave the code commented out (it's easy to remove it if we want, one day).

showcase/app/templates/layouts/app-frame/index.hbs
```diff
@@ -338,24 +337,38 @@
...
+  {{!-- <Shw::Divider @level={{2}} /> --}}
 
   <Shw::Text::H3>Framed</Shw::Text::H3>
 
-  <Shw::Frame
-    @id="demo-full-app-frame-with-app-heade"
-    @src="/layouts/app-frame/frameless/demo-full-app-frame-with-app-header"
-    @label="Full AppFrame with AppHeader"
-  />
+  <Shw::Grid @columns={{1}} {{style gap="2rem"}} as |SG|>
```
didoo, line 349: See comment above about grids and labels

showcase/tests/integration/components/hds/app-frame/index-test.js
```diff
@@ -74,6 +74,24 @@ module('Integration | Component | hds/app-frame/index', function (hooks) {
...
+
+  // @isHeaderFixed hds-app-frame--is-header-fixed
+  test('it should have the hds-app-frame--is-header-fixed class by default', async function (assert) {
+    await render(hbs`
+        <Hds::AppFrame id="test-app-frame" />
+      `);
+    assert.dom('#test-app-frame').hasClass('hds-app-frame--is-header-fixed');
+  });
+
+  test('it should not add the hds-app-frame--is-header-fixed class when @isHeaderFixed is false', async function (assert) {
```
didoo, line 86: probably I would test instead of `false` the case when `isHeaderFixed` is not set

didoo (commented):
@KristinLBradley I've two main comments, about how I see the "problem" but I haven't tested my hypothesis in code, so... be mindful of this, I may have missed/not considered something important. 

That said, overall the "new" approach followed in this PR makes sense to me. Of course, is not tested (yet) with production-code, and the proof is in the pudding, bur for now I would say let's continue the exploration.

Now I'll move on the second PR.

Re. your questions:
> Should I go ahead and remove withAppHeader arg from SideNav entirely in this PR?
I would do it in another PR stacked on top of this so it's cleaner (potentially you could have a stacked set of PRs on top of a "feature" branch/PR.

> Is the behavior of the SideNav overlay rendering over the AppHeader ok?
I've replied also in a couple of places where it was in context, but in general I am not a big fan of this approach.

Let me know if you want to discuss my comments in person, or you need clarifications (or just for rubberducking). You know where to always find me :)

didoo:
@alex-ju it would be great to have your feedback on this direction too.

alex-ju:
> @alex-ju it would be great to have your feedback too, on the direction taken in this PR.

sure, I'll try and get myself up to speed with the AppHeader/AppFrame workstreams

packages/components/src/components/hds/app-frame/index.ts
```diff
@@ -17,6 +17,7 @@ export interface HdsAppFrameSignature {
     hasMain?: boolean;
     hasModals?: boolean;
     hasSidebar?: boolean;
+    isHeaderFixed?: boolean;
```
alex-ju, line 20: I was tempted to name this `hasFixedHeader`, but seeing it next to the other args, I can see how it can be confusing

alex-ju (commented):
I had a quick look and the general approach looks good to me. 

There is an issue with some of the examples overlapping that you may already be aware of
Screenshot 2024-08-02 at 14 49 19

packages/components/src/styles/components/app-frame.scss
```diff
@@ -16,32 +16,47 @@
...
+      position: sticky;
+      top: 0;
+      right: 0;
+      left: 0;
+    }
+  
+    // Modify sidenav layout when used together with fixed app-header
+    .hds-app-frame__sidebar {
+      top: var(--token-app-header-height);
+      height: calc(100vh - var(--token-app-header-height));
```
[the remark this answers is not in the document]
jorytindall: I have a couple of thoughts:

Should we be more opinionated on this? I know various consumers have brought up this being something they might want control over, but this component isn't used in the same way as others. This probably will be set and likely not changed.

Is the request from consumers because they _actually_ want to change this interaction (make the `AppHeader` not sticky)? If our guidance is to have both nav elements sticky, do they intend to go against this (and do they have a good enough reason to)?

To me, this request seems a little bit like consumers want the appearance of control over this interaction pattern, but don't actually intend to go against our guidance (and if they are, they we should be asking why)

jorytindall (commented):
I'll eagerly await @didoo's testing! I dropped a comment that we may want to discuss (potentially in team sync, potentially as a smaller group) about the sticky nature of these elements. To me, I'd like to learn more from consumers of their intentions here and whether they have a valid use case for the AppHeader to _not_ be sticky.

showcase/app/styles/showcase-components/frame.scss
KristinLBradley: Ok I'll change it back. I just thought it seemed redundant since we already had components and patterns for adding labels & spacing

showcase/app/templates/layouts/app-frame/index.hbs
KristinLBradley: Ok thanks, so I will leave as is with the code commented out. I will just add a comment with an explanation.

showcase/app/styles/showcase-components/frame.scss
KristinLBradley: I'm undoing the overrides, thank you.

packages/components/src/components/hds/app-frame/index.ts
KristinLBradley: If anyone has ideas for better names, I'm fine with changing it

packages/components/src/styles/components/app-frame.scss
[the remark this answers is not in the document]
KristinLBradley: Currently the SideNav is fixed/sticky and that's how it's behaved from its initial release and I haven't known of complaints about that behavior (although I may just be ignorant of complaints.)

The consensus so far from designers seems to be to make the AppHeader fixed/sticky and not give consumers a built in option to override this behavior for sake of consistency.

I tend to agree that these type of cross-product standard navigation components should behave as consistently as practical.

So I'm going to leave all this alone awhile longer to see if there's any further discussion... (And then I will remove the option to "unfix" the header if that's the true consensus.)

packages/components/src/styles/components/app-frame.scss
KristinLBradley: Leaving this alone for now until there's more discussion and a firmer decision.

packages/components/src/styles/components/app-frame.scss
KristinLBradley: I agree this still needs more input on how this behavior should work. I asked @jorytindall to test and give his feedback. I would also like input from others @hashicorp/hds-design @hashicorp/hds-engineering 

**Problem:**
How should the AppHeader's `MenuButton` appear and behave when the SideNav is expanded in mobile view?

[Test the interactions for yourself](https://hds-showcase-git-hds-3612-app-frame-updates-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-app-header)

1. Narrow your browser window until the "Menu" button appears in the top left corner of the AppHeader.
2. Click the SideNav tab to expand the SideNav
3. With the SideNav open, attempt to interact with the "Menu" button.

Is the interaction ok? Confusing? How would you expect the Menu button to behave instead (if differently) and how should it appear visually?

packages/components/src/components/hds/app-frame/index.ts
KristinLBradley: although it also seems that this may go away entirely dependent upon decisions...

showcase/tests/integration/components/hds/app-frame/index-test.js
KristinLBradley: I clarified the wording

packages/components/src/styles/components/app-frame.scss
KristinLBradley: I do agree it needs a bit more thought. I asked @jorytindall to test and give his feedback of the behavior. I would appreciate input from other @hashicorp/hds-design people as well.

**To clarify the questions:**
👉 We are trying to figure out the best way of handling the AppHeader's `MenuButton` when the SidNav is expanded (open) in mobile view. How should it appear visually? What should its behavior be?

KristinLBradley:
> I had a quick look and the general approach seems fine. I'll try and follow-up with a more in-depth review.
> 
> There is an issue with some of the examples overlapping that you may already be aware of Screenshot 2024-08-02 at 14 49 19

Just noticed this, thanks. Will look into it

KristinLBradley:
> > @jorytindall Do you have any ideas on how it could be improved? To me the overlay just has a slight transition between visible/not visible which looks fine to me but if it bothers you how would you want it changed?
> 
> @KristinLBradley Sorry, I'll clarify, I think the transition of the overlay is fine, if anything were to change it would maybe be increasing the opacity of the overlay (making it less opaque), but that's a larger discussion that would affect other things as well, so I'd say let's not touch it.
> 
> > I'm not quite getting what you mean by the active states persisting? I don't notice anything jumping out as overly weird personally.
> 
> I only noticed it on my phone, but when I press either of these buttons (menu or collapse), the persist in what looks like their hover state as I scroll around the page.

Do you think the hover styles should be turned off for mobile users? Do other HDS Buttons have this issue in mobile devices? (If it's a broader issue then I'd be more inclined to looking at it separately.)

showcase/app/styles/showcase-components/frame.scss
KristinLBradley: `main` has been merged in with the updated width for Showcase examples, thanks!

packages/components/src/styles/components/app-frame.scss
KristinLBradley: Should we perhaps _not_ disable the `AppHeader` `MenuButton` and `HomeLink` controls at all when the mobile SideNav is expanded?

packages/components/src/styles/components/app-frame.scss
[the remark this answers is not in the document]
didoo: For future reference (copy&pasting&rewording from a thread in Slack)

> Deciding that the AppFrame, which is a very generic, high-level/page-level layout component, has by default a fixed header and sidebar, I mean engrained in the code implementation, instead of being an argument/prop, is an antipattern. We can easily guide our consumers to set it to true (or do it ourselves, since it’s likely in the consumers codebase there is [a very limited number of instances](https://github.com/search?q=%3CHds%3A%3AAppFrame+user%3Ahashicorp+repo%3Ahashicorp%2Fcloud-ui&type=code&ref=advsearch)).
> [Pre-assigning sticky behaviour to the Header/Sidebar in the AppFrame] is contrast with how we’ve implemented most of our components (in code) in our design system. Components should be as agnostic as possible, be black boxes that know only (ideally) about themselves, and not about their parents or their children. Their interface with the inside/outside world is via their API.

That said, my suggestion is to escalate the decision to the Core Squad Lead(s).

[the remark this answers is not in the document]
KristinLBradley: From discussion, I've gone ahead and removed the `isHeaderFixed` argument along with related code

packages/components/src/styles/components/app-frame.scss
KristinLBradley: There doesn't appear to be a way to trap focus in both the SideNav & AppHeader simultaneously that I was able to discover. I've instead made a number of refactors to the CSS to include disabled styles for interactive elements in the AppHeader when the SideNav is expanded in mobile view. Using the disabled styles should make it more obvious to users that the MenuButton and HomeLink can't be interacted with.

KristinLBradley:
@didoo I removed the no longer necessary `withAppHeader` argument from the `SideNav` in a separate PR branched off of this one as you recommended: https://github.com/hashicorp/design-system/pull/2324

(I searched around but couldn't find your original comment about this.)

jorytindall:
@KristinLBradley 

> Do you think the hover styles should be turned off for mobile users? Do other HDS Buttons have this issue in mobile devices? (If it's a broader issue then I'd be more inclined to looking at it separately.)

I don't think so for now as this would be somewhat inconsistent with the rest of our interactive elements which rely on hover as opposed to touch. I think it's low risk and should probably be looked at more holistically in the future.

jorytindall (approved):
These changes to the buttons in the `AppHeader` work for me given the technical constraints of controlling focus between two separate components. Since being able to control the focus order more broadly across both the `AppHeader` and `SideNav` would be a better experience long-term, let's consider looking into that in Phase 2. I'll add it to the list of priorities.

.changeset/rich-actors-speak.md
```diff
@@ -0,0 +1,14 @@
...
+`AppFrame`:
+- Modified sticky/fixed position to turn off when viewport height is under 480px in height
+- Refactored styles to make `AppFrame` responsible for sticky/fixed layout of `SideNav` and `AppHeader`
+
+`AppHeader`:
+- Styled inoperable actions as disabled (which occurs when the `SideNav` is expanded in mobile view)
+
+`SideNav`:
+- Removed need for `withAppHeader` option when used together with AppHeader
+- Removed the `withAppHeader` option as it is no longer needed.
```
didoo, line 14: [todo] Are these two lines duplicated somehow?

packages/components/src/components/hds/app-frame/index.ts
```diff
@@ -33,47 +33,30 @@ export interface HdsAppFrameSignature {
...
-   * @type {boolean}
-   * @default true
-   */
+  // Indicates if the "modals" container should be displayed
   get hasModals(): boolean {
     return this.args.hasModals ?? true;
   }
+
+  // Get the class names to apply to the component.
+  get classNames(): string {
```
didoo, line 57: I think at this point you can revert the changes to the getter and move the single class back to the template file

packages/components/src/styles/components/app-frame.scss
```diff
@@ -16,32 +18,39 @@
...
+  // Make header position sticky/fixed if the viewport height is greater than 480px
+  @media (height >= 480px) {
     position: sticky;
     top: 0;
     right: 0;
     left: 0;
+
+    // Modify sidenav layout when used together with fixed app-header
+    &:has(.hds-app-header) {
+      + .hds-app-frame__sidebar {
```
didoo, line 38: [thought/non-blocking] for me having this here is confusing: it took me a while to realize this rule applies to the sidebar not the header (it's in the "header" block).

an option would be to move it down inside the `.hds-app-frame__sidebar`, something like this (rough code):

```

.hds-app-frame__sidebar {
  // rules here

  @media (height >= 480px) {
    .hds-app-frame__header:has(.hds-app-header) + & {
      // top/height/min-height here
    }
  }
}
```

Another option to explore is to see if we could use instead the `hasHeader`/`hasSidebar` booleans to determine when both are present, and add a class accordingly (in which case, the `classNames` getter can be left in the backing class).

packages/components/src/styles/components/app-header.scss
```diff
@@ -76,6 +75,30 @@
...
+    &:disabled,
+    &[disabled],
+    &.mock-disabled,
+    &:disabled:focus,
+    &[disabled]:focus,
+    &.mock-disabled:focus,
+    &:disabled:hover,
+    &[disabled]:hover,
+    &.mock-disabled:hover {
+      @include hds-interactive-dark-theme-state-disabled();
```
didoo, line 89: [todo] I think the `@use "../mixins/interactive-dark-theme" as *;` should be added to this file.

packages/components/src/styles/components/app-header.scss
```diff
@@ -76,6 +75,30 @@
...
+    }
+  }
+}
+
+// prevent interaction with AppHeader when SideNav is open/expanded
```
didoo, line 94: For the record: I still think this is not the right solution, but the majority of the people decided otherwise so I am not blocking moving forward with for this.

didoo (commented):
Left a few comments

.changeset/rich-actors-speak.md
KristinLBradley: So first I refactored to remove the need for the `withAppHeader` argument and then I actually removed the actual argument from the code. That's what I was trying to express but I agree that's confusing so I will edit, thanks.

packages/components/src/styles/components/app-header.scss
KristinLBradley: What do you think is the right solution?

packages/components/src/styles/components/app-frame.scss
```diff
@@ -7,6 +7,8 @@
 // APP-FRAME
 //
 
+@use "../mixins/interactive-dark-theme" as *;
+
```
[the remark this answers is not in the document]
KristinLBradley: Removing locally

packages/components/src/styles/components/app-header.scss
didoo: Briefly commented here: https://github.com/hashicorp/design-system/pull/2299#discussion_r1701757266
Essentially, for me the "primary" navigation (the `AppHeader`) should not be blocked/obscured/disabled by the "secondary" navigation (the `SideNav`).
See how the IBM Carbon `AppShell` (not the exact name, but it's what it would be called if we had to implement something similar) works at small viewports. For me this behaviour makes more sense.

packages/components/src/styles/components/app-frame.scss
KristinLBradley: Added a conditional classname (.hds-app-frame--has-header-with-sidebar) to the AppFrame and used it to enable positioning related styles.

packages/components/src/styles/components/app-header.scss
KristinLBradley: > [todo] I think the `@use "../mixins/interactive-dark-theme" as *;` should be added to this file.

It's there above. I would have gotten an error with the `@include`s if it had been missing.

KristinLBradley: I updated the Showcase to include disabled button and HomeLink examples.

packages/components/src/styles/components/app-header.scss
KristinLBradley: The overlay/grayscreen of the `SideNav` now no longer overlays the `AppHeader`. I had thought that was your main objection in that old thread which I had resolved since I agreed it probably wasn't the best solution.

I had wanted to keep the `AppHeader`'s child actions still enabled when the `SideNav` was expanded in mobile view, and I explored that solution, but the way the `focus-trap` the `SideNav` is using works I couldn't find a way to do that.

We could look to find another solution in the next major phase of this project but I think it's going to be much more involved (so for me I think the solution of making the actions more visibly look "disabled" is ok for this intermediate phase).

packages/components/src/styles/components/app-frame.scss
KristinLBradley: UPDATE: For now, we have decided to add disabled styling to the inoperable controls within the AppHeader to make it clearer to users that they aren't functional.

packages/components/src/styles/components/app-header.scss
didoo: OK, I misunderstood then. Better so :)

packages/components/src/styles/components/side-nav/header.scss
```diff
@@ -47,6 +47,19 @@
   width: 100%;
   height: 100%;
   padding: calc(var(--token-side-nav-header-home-link-padding) - 1px); // by design -  we take in account the transparent border
+
+  // disabled state:
+  &:disabled,
```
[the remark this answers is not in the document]
KristinLBradley: Sorry, I missed the other controls. Adding disabled styles for them as well now.

I had separated out the disabled styles as a separate mixin so they could be applied separately.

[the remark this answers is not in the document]
KristinLBradley: too many "separate"s... :P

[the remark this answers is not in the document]
KristinLBradley: I updated the SideNav Showcase as well now to include examples of disabled interactive content

showcase/app/styles/showcase-pages/app-frame.scss
```diff
@@ -66,6 +66,18 @@ body.layouts-app-frame {
       width: 100%;
       height: 100%;
     }
+
+    // remove sticky positioning from examples
+    .hds-app-frame--is-header-fixed .hds-app-frame__header,
+    .hds-app-frame__sidebar {
+      position: relative;
+      top: 0;
```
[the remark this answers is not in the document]
KristinLBradley: when I remove it, there is a gap above the sidebar block

KristinLBradley:
@didoo Thanks for your review, I pushed up changes to address your comments. I think I've responded to everything now so would appreciate another re-review when you have time.

MelSumner:
When testing at browser zoom 200%, navigating with my keyboard is allowing me to focus on elements that are not in the viewport. 

Relevant WCAG:
- https://www.w3.org/WAI/WCAG22/Understanding/focus-visible
- https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum

MelSumner:
I'm also seeing instances of elements getting cut off on browser zoom, but I wonder if @alex-ju 's PR fixes this issue? 

Screenshot: 
CleanShot 2024-08-08 at 16 05 49@2x

KristinLBradley:
> I'm also seeing instances of elements getting cut off on browser zoom, but I wonder if @alex-ju 's PR fixes this issue?
> 
> Screenshot: CleanShot 2024-08-08 at 16 05 49@2x

@MelSumner The Dropdown list positioning issues will be fixed once Alex's PR is merged in, yes.

KristinLBradley:
> When testing at browser zoom 200%, navigating with my keyboard is allowing me to focus on elements that are not in the viewport.
> 
> Relevant WCAG:
> 
> * https://www.w3.org/WAI/WCAG22/Understanding/focus-visible
> * https://www.w3.org/WAI/WCAG22/Understanding/focus-not-obscured-minimum

@MelSumner Could you please tell me which elements are the issue? In testing myself I wasn't seeing this issue. Unless you mean the Dropdown content which should be fixed by [Alex's PR](https://github.com/hashicorp/design-system/pull/2309).

didoo (approved):
LGTM 👍

PR #2302, opened 2024-07-30
`Hds::Icon` - Add support for predefined "foreground" colors in`@color` argument

didoo:
### :pushpin: Summary

While reviewing some code related to icons, I realized that we don't support pre-defined "foreground" colors for the `Hds::Icon`, in the same way as we do for the `Hds::Text` component. 

This PR addresses this difference (while the `Hds::Icon` has not yet been released, so it's just an extension of the existing API, without impact on our consumers).

Notice: I have tested the change in Cloud UI and it worked as expected

In this PR I have:
- added support for predefined “foreground” colors to the `@color` argument
- updated “colors” section of the showcase page
- updated `@color` in "Component API" section
- updated “Color” sub-section in “How to use” section

Preview: https://hds-showcase-git-hds-icon-color-argument-hashicorp.vercel.app/components/icon#color

Notice: I haven't added a changelog to this PR because lives on top of the `Hds::Icon` ones, so the changelog is provided by them.

screenshot_4025

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/components/hds/icon/types.ts
+++ b/packages/components/src/components/hds/icon/types.ts
@@ -4,3 +4,27 @@ export enum HdsIconSizeValues {
 }
 
 export type HdsIconSizes = `${HdsIconSizeValues}`;
+
+export enum HdsIconColorValues {
+  Primary = 'primary',
+  Strong = 'strong',
+  Faint = 'faint',
+  Disabled = 'disabled',
+  HighContrast = 'high-contrast',
+  Action = 'action',
+  ActionHover = 'action-hover',
+  ActionActive = 'action-active',
+  Highlight = 'highlight',
+  HighlightOnSurface = 'highlight-on-surface',
+  HighlightHighContrast = 'highlight-high-contrast',
+  Success = 'success',
+  SuccessOnSurface = 'success-on-surface',
+  SuccessHighContrast = 'success-high-contrast',
+  Warning = 'warning',
+  WarningOnSurface = 'warning-on-surface',
+  WarningHighContrast = 'warning-high-contrast',
+  Critical = 'critical',
+  CriticalOnSurface = 'critical-on-surface',
+  CriticalHighContrast = 'critical-high-contrast',
+}
+export type HdsIconColors = `${HdsIconColorValues}`;
--- a/showcase/app/templates/components/icon.hbs
+++ b/showcase/app/templates/components/icon.hbs
@@ -42,46 +42,71 @@
     </SF.Item>
   </Shw::Flex>
 
-  <Shw::Divider @level={{2}} />
+  <Shw::Divider />
 
   <Shw::Text::H2>Color</Shw::Text::H2>
 
+  <Shw::Text::H4 @tag="h3">Color inheritance</Shw::Text::H4>
   <Shw::Flex @direction="column" as |SF|>
     <SF.Item as |SFI|>
-      <SFI.Label>Unspecified color (<code>currentColor</code>)</SFI.Label>
-      <Hds::Icon @name="bookmark" />
-    </SF.Item>
-    <SF.Item as |SFI|>
-      <SFI.Label>Custom red color, specified via <code>@color</code> argument</SFI.Label>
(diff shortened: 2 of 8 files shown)


website/docs/components/icon/partials/code/how-to-use.md
```diff
@@ -36,13 +36,21 @@ The default size is 16px. To use the alternative 24px icon size, set the `@size`
...
 
 ```handlebars
-<Hds::Icon @name="zap" @color="var(--token-color-foreground-success)" />
+<Hds::Icon @name="zap" @color="success" />
 ```
 
-Other accepted values include named colors and color values themselves (e.g., hex, rgb, etc).
+For the list of possible foreground colors supported, refer to the [Component API](#component-api) section for details.
+
+It’s also possible to provide a CSS color as string (in this case the color will be applied as SVG `fill` property). The string can be a CSS `var()` that uses one of the [predefined color tokens](/foundations/colors?tab=palette):
```
zamoore, line 47:
```suggestion
It’s also possible to provide a CSS color as a string (in this case the color will be applied as SVG `fill` property). The string can be a CSS `var()` that uses one of the [predefined color tokens](/foundations/colors?tab=palette):
```

zamoore (approved):
Looks good to me. Added a minor grammar suggestion, but I'm glad that icon color is in alignment with the rest of the components in the system.

showcase/app/templates/components/icon.hbs
```diff
@@ -42,46 +42,71 @@
     </SF.Item>
   </Shw::Flex>
 
-  <Shw::Divider @level={{2}} />
+  <Shw::Divider />
 
   <Shw::Text::H2>Color</Shw::Text::H2>
 
+  <Shw::Text::H4 @tag="h3">Color inheritance</Shw::Text::H4>
```
shleewhite, line 49: Nit: you could just do `<Shw::Text::H3>Color inheritance</Shw::Text::H4>`

shleewhite (approved):

showcase/app/templates/components/icon.hbs
KristinLBradley: This should be changed to H3 as to not skip a heading level

KristinLBradley: Missed that the @tag was changed initially. If you want the visual style of the smaller H4 here for some reason then this is fine.

packages/components/src/components/hds/icon/types.ts
```diff
@@ -0,0 +1,30 @@
+export enum HdsIconSizeValues {
+  Sixteen = '16',
+  TwentyFour = '24',
+}
+
+export type HdsIconSizes = `${HdsIconSizeValues}`;
+
+export enum HdsIconColorValues {
```
MelSumner, line 8: (question): wouldn't we want all of these to be HdsColorValues anyway? Is there a need to have them specifically in IconColorValues? It seems reasonable that if these are our ColorValues we'd want them available for everywhere that we use these values. WDYT?

KristinLBradley (approved):

showcase/app/templates/components/icon.hbs
didoo: > Nit: you could just do <Shw::Text::H3>Color inheritance</Shw::Text::H4>

`<Shw::Text::H4 @tag="h3">` means render an `<h3>` tag with the visual style of a `.shw-text-h4`

packages/components/src/components/hds/icon/types.ts
didoo: I thought about it, but I ended up keeping them separate for different reasons:

1) we may want to add also the "brand" colors for the icon (but not the text, I don't see use cases for that); at the moment thought is not clear _how_ we would do it: have [different brand colors](https://helios.hashicorp.design/foundations/tokens?searchQuery=brand), which ones would be the ones that _could_ be used for an icon?
2) this would require to add a global `types.ts` file, under `packages/components/src/components/hds`; not a big deal but maybe something we want to consider a bit more

so my thinking is that we can always do it, it's not something for this specific PR

PR #2304, opened 2024-07-31
Convert Breadcrumb to TypeScript

shleewhite:
### :pushpin: Summary

Converted breadcrumb to typescript.

### :link: External links

Jira ticket: [HDS-2685](https://hashicorp.atlassian.net/browse/HDS-2685)

***

### 👀 Component checklist

[HDS-2685]: https://hashicorp.atlassian.net/browse/HDS-2685?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/menu-primitive/index.ts
+++ b/packages/components/src/components/hds/menu-primitive/index.ts
@@ -17,7 +17,7 @@ export interface MenuPrimitiveSignature {
     toggle?: [
       {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        onClickToggle?: (event: MouseEvent, ...args: any[]) => void;
+        onClickToggle: (event: MouseEvent, ...args: any[]) => void;
         isOpen?: boolean;
       },
     ];
--- a/.changeset/olive-eagles-add.md
+++ b/.changeset/olive-eagles-add.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Breadcrumb` - Converted component to TypeScript
--- a/packages/components/src/components.ts
+++ b/packages/components/src/components.ts
@@ -11,6 +11,9 @@ import HdsAppFrame from './components/hds/app-frame/index.ts';
 import HdsApplicationState from './components/hds/application-state/index.ts';
 import HdsBadge from './components/hds/badge/index.ts';
 import HdsBadgeCount from './components/hds/badge-count/index.ts';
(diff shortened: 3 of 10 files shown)


packages/components/src/components/hds/menu-primitive/index.ts
```diff
@@ -17,7 +17,7 @@ export interface MenuPrimitiveSignature {
     toggle?: [
       {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        onClickToggle?: (event: MouseEvent, ...args: any[]) => void;
+        onClickToggle: (event: MouseEvent, ...args: any[]) => void;
```
shleewhite, line 20: Question: this is probably a me misunderstanding ember components issue, but when I converted the breadcrumb component to TS - [the button for the truncation component](https://github.com/hashicorp/design-system/blob/73316fdeb281ba329724415098ac2d52b614d1e3/packages/components/src/components/hds/breadcrumb/truncation.hbs#L13) had a type error because the onClick event couldn't be undefined.

```ts
  <:toggle as |t|>
      <button
        type="button"
        class="hds-breadcrumb__truncation-toggle"
        aria-label={{this.ariaLabel}}
        aria-expanded={{if t.isOpen "true" "false"}}
        {{on "click" t.onClickToggle}} // where the error happened
      >
        <FlightIcon @name="more-horizontal" @size="16" @isInlineBlock={{false}} />
      </button>
    </:toggle>
```

When I made onClickToggle in the menu-primitive, the error went away. Is this the correct fix? is there a way to conditionally pass an onClick event to a button? Also, if this is the correct fix, is it a breaking change? It doesn't break any of the HDS usage of menu primitive but I don't know how many people are using it outside of our team.

alex-ju: In both instances where we use the `MenuPrimitive` (`Dropdown` and `Breadcrumb::Truncation`) `onClickToggle` is provided. I typed the `MenuPrimitive` as such when I converted the Dropdown component to TypeScript (and I was not aware of the usage in `Breadcrumb::Truncation` at the time). Now that we know it's required it's ok to adjust the `MenuPrimitiveSignature` as needed, as you did.

shleewhite: Cool, thanks!

alex-ju (approved):
Looking good! We're missing return types for functions, but not a blocker

PR #2305, opened 2024-07-31
Page load optimization

curthashicorp:
### :pushpin: Summary

This PR simply moves the prism highlighting out of the build-time compile stage, stops blocking, and is applied after the view has rendered (highlighting time after view has rendered is only 53ms). This reduces the time to load each tab from 3 seconds to 1 second. A one second delay is now visible when clicking from tab to tab vs 3+ seconds.

### Before

https://github.com/user-attachments/assets/eca931b4-1d14-40c6-b764-831d254935f8

### After

https://github.com/user-attachments/assets/d0738fac-123c-4661-ba7c-46ba16986d0c

### :link: External links

Jira ticket: [HDS-2501](https://hashicorp.atlassian.net/browse/HDS-2501)


website/app/initializers/showdown-extensions.js
```diff
@@ -93,13 +93,8 @@ export function initialize(/* application */) {
           language = 'shell';
         }
 
-        if (Prism.languages[language]) {
-          // if the specified language is available in Prism we highlight the codeblock
-          highlightedCodeBlock = Prism.highlight(codeblock, Prism.languages[language], language) + end;
```
Dhaulagiri, line 98: More thinking out loud here, but the previous function here used the specified language to highlight the block so I was wondering if the new approach might not highlight different languages correctly. In practice it seems to be working ok though, at least looking at `hbs` and `js` code blocks.

website/app/controllers/show.js
```diff
@@ -202,6 +225,21 @@ export default class ShowController extends Controller {
     // console.log('show didInsert', this.sections, this.tabs, this.tocs);
   };
 
+  @action
```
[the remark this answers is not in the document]
curthashicorp: ahh makes sense.

website/app/initializers/showdown-extensions.js
curthashicorp: Yeah lang still gets added as css selector and the Prism.highlightAllUnder() call in the show controller looks for language for syntax highlighting.  I will test a couple additional langs and make sure this new method is exactly the same.

Dhaulagiri: Nice. That's what I think I understood from reading the docs, but was good to confirm.

curthashicorp: Double verified and we can add additional language support in ./website/ember-cli-build.js I tested by adding support for html and scss and highlighting was applied with old prism calls and our new one triggered in didRender.  Highlighting seems to be theme dependent though.  We are passing theme: 'dracula' to our 'ember-prism' prop in appConfig and this particular theme doesn't support every language. Eg. tested with php and json and no highlighting support available before this change or after. So in the future if we want to support more langs we will need to make the the prism theme we are using has appropriate css selectors.

website/app/controllers/show.js
```diff
@@ -202,6 +204,20 @@ export default class ShowController extends Controller {
     // console.log('show didInsert', this.sections, this.tabs, this.tocs);
   };
 
+  applyCodeHighlighting() {
+    const allSections = document.querySelectorAll(
```
didoo, line 208: suggestion: since what we're highlighting are code blocks (or code snippets, to be more precise) and since the word "section" is often used to indicate different parts of the pages (and is also an HTML tag) maybe we could call these entities `codeBlocks` or `codeSnippets` and a single item would be `codeBlock` or `codeSnippet`. What do you think?

didoo (approved):
Code is OK, I have left a couple of comments about variables naming.
👍

website/app/initializers/showdown-extensions.js
```diff
@@ -93,13 +93,8 @@ export function initialize(/* application */) {
...
-        } else {
-          // otherwise we encode the codeblock and present it as is (without highlight)
-          highlightedCodeBlock = Prism.util.encode(codeblock) + end;
-        }
+        // we encode the codeblock and present it as is (without highlight)
+        highlightedCodeBlock = Prism.util.encode(codeblock) + end;
```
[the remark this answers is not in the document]
curthashicorp: ok

alex-ju (approved):
The only downside I can see with this approach is that we don't have the highlight in a no-js experience (this also affects the `CodeBlock` component, as there can only be one Prism global). I consider that a minor tradeoff considering the improved experience resulting from this change.

PR #2306, opened 2024-07-31
`AppHeader`: “Enterprise Nav” phase 1.5 updates  (HDS-3613)

KristinLBradley:
### :pushpin: Summary

If merged, this PR updates the `AppHeader` to:

* Hide the closed menu in mobile view via CSS vs. rendering/not rendering the content
* Add a Skip link with associated args which points to "#main" by default
* ~~**_WIP:_** Rough implementation using resizeObserver to switch views instead of fixed breakpoint~~ _-Removed_

👉 **AppHeader Showcase:** https://hds-showcase-git-hds-3613-app-header-updates-hashicorp.vercel.app/components/app-header
👉 **Enterprise Nav Framed Example:** https://hds-showcase-git-hds-3613-app-header-updates-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-app-header

**Related PR:**

* ~~AppFrame updates PR: https://github.com/hashicorp/design-system/pull/2299~~ _- merged into `main` then rebased onto this branch_

### :link: External links

Jira ticket: [HDS-3613](https://hashicorp.atlassian.net/browse/HDS-3613)

***

### 👀 Component checklist

[HDS-3613]: https://hashicorp.atlassian.net/browse/HDS-3613?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/components/hds/app-header/index.ts
```diff
@@ -42,6 +50,18 @@ export default class HdsAppHeaderComponent extends Component<HdsAppHeaderSignatu
...
+    // * Should hasA11yRefocus default to "true"? (I assume yes)
+    // * Should a11yRefocusSkipTo default to a value? (Could add an id to the SideNav then point to that id by default)
+
+    // if a11yRefocus is enabled, ensure that a11yRefocusSkipTo value is set
+    // if (this.hasA11yRefocus) {
+    //   assert(
+    //     '@a11yRefocusSkipTo for NavigatorNarrator (a11y-refocus) in "Hds::SideNav" must have a valid value',
+    //     this.args.a11yRefocusSkipTo !== undefined
+    //   );
+    // }
```
KristinLBradley, line 64: I took this mainly from the `SideNav`.

However, the SideNav was checking `this.args.hasA11yRefocus` which is not true if it's not explicitly passed in (which it shouldn't have to be since the default is supposed to be `true`.)

I commented this out for now because I would need to pass an id into `a11yRefocusSkipTo` in all my examples otherwise they blow up as expected. But I wanted to confirm what behavior we actually want & want to enforce on consumers.

KristinLBradley:
@didoo I pushed up a rough implementation using the resizeObserver to switch views. I commented out all the previous code related to using a fixed breakpoint (so the code is in. messy state currently).

See what you think. The resize observer is kind of working but it's glitchy.

didoo (commented):
@KristinLBradley I looked at this a lot, and I can't figure out a "simple" solution (for the ResizeObserver, I mean). The reason why is glitchy is because of a fundamental flaw in the approach (that I suggested, so it's not your fault, it's mine): when the content becomes larger than the parent container (`if (entry.target.scrollWidth > entry.target.clientWidth)`) the CSS switches to "mobile", but at this point the content becomes smaller than the parent container, so it switches back to "desktop", but at this point is agina larger, so it switches to mobile, and so on back and forth. If you put a `console.log` in the code, you will see how many times it's called! (almost an infinite loop)

Now, there may be a way to make the logic more complex, to handle this, but I think we should look at the problem in a different way. If that's OK for you I an going to try a quick experiment, to explore an alternative approach where the ResizeObserver observes the AppFrame parts and sets CSS variables about the size of the different elements; and in the experiment I also want to try something about the "mobile" menu).
Let me know it that's OK, or if you prefer to do it yourself (I can describe in more details what I am thinking, if you want)

/cc @alex-ju

packages/components/src/components/hds/app-header/index.ts
KristinLBradley: I had thought that we would want to provide a skip link for both the AppHeader and the SideNav controls. If a user activated the AppHeader skip link I thought it would make sense to jump them to the SideNav where they could either skip again or else use the SideNav controls. @MelSumner how do you think this should work?

I think it could make sense to create an HDS `SkipLink` component in any case...

KristinLBradley:
> @KristinLBradley I looked at this a lot, and I can't figure out a "simple" solution (for the ResizeObserver, I mean). The reason why is glitchy is because of a fundamental flaw in the approach (that I suggested, so it's not your fault, it's mine): when the content becomes larger than the parent container (`if (entry.target.scrollWidth > entry.target.clientWidth)`) the CSS switches to "mobile", but at this point the content becomes smaller than the parent container, so it switches back to "desktop", but at this point is agina larger, so it switches to mobile, and so on back and forth. If you put a `console.log` in the code, you will see how many times it's called! (almost an infinite loop)
> 
> Now, there may be a way to make the logic more complex, to handle this, but I think we should look at the problem in a different way. If that's OK for you I an going to try a quick experiment, to explore an alternative approach where the ResizeObserver observes the AppFrame parts and sets CSS variables about the size of the different elements; and in the experiment I also want to try something about the "mobile" menu). Let me know it that's OK, or if you prefer to do it yourself (I can describe in more details what I am thinking, if you want)
> 
> /cc @alex-ju

I tried combining a preset breakpoint with overflow logic using the ResizeObserver but the logic is very convoluted and I don't think it provides a whole lot of benefit.

I actually think just setting a fixed breakpoint with the option to override with a custom one works well enough for the majority of content. But let me know what you think.

didoo:
> I actually think just setting a fixed breakpoint with the option to override with a custom one works well enough for the majority of content.

At this point, all considered, makes sense.

packages/components/src/components/hds/app-header/index.ts
KristinLBradley: I still don't know what to do about the SkipLinks.

I don't think the existing SideNav one works quite right, as I said, since it doesn't actually require consumers to add an id value for the Skip Link.

If we want to remove the built-in Skip Links and instead ask consumers to add them should I do that as part of this PR or is that something we should look at separately?

KristinLBradley: **Q: How about I do the following?**

1. Create a reusable `HdsSkipLink` component.
2. Add this new SkipLink into the `AppHeader`.
3. Update the `SideNav` to replace its current built-in skip link with the new component.
4. When the SideNav is pared with the AppHeader, turn off its skip link and instead turn on the AppHeader skip link.

shleewhite: I can't remember if we talked about this in sync, but are there times that the SideNav is used without the AppHeader?

KristinLBradley: @shleewhite Currently the [SideNav](https://helios.hashicorp.design/components/side-nav) is always used without the AppHeader as the AppHeader is new and hasn't been adopted by products yet.

We plan to help teams in adopting the new AppHeader to pair it with the SideNav.

didoo: > How about I do the following?

Makes sense to me

MelSumner: We need to have the page navigation message there too, is there a reason we can't use `ember-a11y-refocus` like we've been using? That has a skip link (bypass block) built in.

MelSumner:
> Or what we could do is to provide an <Hds::NavigationNarrator> component, that is built on top of the <NavigationNarrator> but provides by default the positioning and visual styling (see CSS below for .ember-a11y-refocus-skip-link)

@didoo I would be okay with having the consumer do this work, but a few things:

- If they are still using our AppFrame component, we should make sure that there is an `id="main"` on the `<main>` element, since that's the default behavior (just like the other options in the addon, it can be customized on invocation with `@skipTo` but we should try to use the defaults if we're asking them to invoke it themselves, IMO)
- we don't need to build a whole new component, we can use the styling classes in the navigation narrator component itself for the CSS we want to use, it's been explicitly namespaced just for this exact reason.

didoo:
> * If they are still using our AppFrame component, we should make sure that there is an `id="main"` on the `<main>` element, since that's the default behavior (just like the other options in the addon, it can be customized on invocation with `@skipTo` but we should try to use the defaults if we're asking them to invoke it themselves, IMO)

@MelSumner The first app I tested (Cloud UI) doesn't have an `id="main` but `id="app-main-content"`. They may have had their own reasons to call it this way, there may be tests associated with it, or CSS or JS selectors, etc. We should respect their decision in this case. "Guidance over control".

modified

MelSumner:
@didoo isn't this coming from our AppFrame component? 🤔

KristinLBradley:
> > Or what we could do is to provide an Hds::NavigationNarrator component, that is built on top of the  but provides by default the positioning and visual styling (see CSS below for .ember-a11y-refocus-skip-link)
> 
> @didoo I would be okay with having the consumer do this work, but a few things:
> 
> * If they are still using our AppFrame component, we should make sure that there is an `id="main"` on the `<main>` element, since that's the default behavior (just like the other options in the addon, it can be customized on invocation with `@skipTo` but we should try to use the defaults if we're asking them to invoke it themselves, IMO)
> * we don't need to build a whole new component, we can use the styling classes in the navigation narrator component itself for the CSS we want to use, it's been explicitly namespaced just for this exact reason.

@MelSumner if we just use the navigation narrator class names wouldn't we run the risk of overriding custom styles consumers may have in their codebases? If we wrap it to create an `HdsSkipLink` component then we could scope styles to the component and it would also be available for consumers to use and include if they are building their own custom navigation components for whatever reason.

packages/components/src/components/hds/app-header/index.ts
KristinLBradley: Ok, so what I've gone ahead and done for now, is to include a `NavigationNarrator` in the `AppHeader` component which is included by default. And, I set a default value of "#main" for `@skipTo` / `@a11yRefocusSkipTo`.

I _did not_ create a new SkipTo component as there doesn't seem to be alignment around that. (We can always do this later if desired.)

didoo (approved):
Nothing specific to raise, LGTM 👍

KristinLBradley:
Note: When I merged in the AppFrame updates branch I discovered an issue with the AppHeader skip link.

The AppFrame PR had introduced longer content into the AppFrame framed examples. I discovered that with this longer content when I used the skip link it jumped the main content to the top which hid content underneath the fixed AppHeader. 

I added a CSS fix to prevent this from happening.

jorytindall (approved):
I'm not sure if these are blockers (probably not because they're kind of caused by the broader experience and other focus/keyboard related things), but noting here to keep in mind:

- When the mobile menu is open, using the skip link moves the scroll, but the menu doesn't close, so I can't see the page content or where I'm focused. Is this a known issue?
- Not sure what specifically this is related to, but the SkipLink seems pretty slow to appear. It took me a while to notice it.
- On smaller viewports, as I'm  navigating through elements with the keyboard there are a lot of elements that I think I'm focusing on but are either hidden or not visible in the current view. Is that intended?
- When the SideNav is collapsed, I _think_ I'm still tabbing through the list items, but they aren't visible. I think this is a known issue that was discussed previously, but just calling it out. 

I know there are still some things in flight for this work, but just capturing these points in my testing as we're going so they aren't missed.

KristinLBradley:
> I'm not sure if these are blockers (probably not because they're kind of caused by the broader experience and other focus/keyboard related things), but noting here to keep in mind:
> 
> * When the mobile menu is open, using the skip link moves the scroll, but the menu doesn't close, so I can't see the page content or where I'm focused. Is this a known issue?
**KB:** _I pushed up an update to not render the skip button when the menu is open for now._

> * Not sure what specifically this is related to, but the SkipLink seems pretty slow to appear. It took me a while to notice it.
**KB:** _We can consider design related updates including the Skip link animation as part of larger design updates in the next major phase._

> * On smaller viewports, as I'm  navigating through elements with the keyboard there are a lot of elements that I think I'm focusing on but are either hidden or not visible in the current view. Is that intended?
> * When the SideNav is collapsed, I _think_ I'm still tabbing through the list items, but they aren't visible. I think this is a known issue that was discussed previously, but just calling it out.

**KB:** _Going through the 2 points above now to test more, will update this comment..._


> I know there are still some things in flight for this work, but just capturing these points in my testing as we're going so they aren't missed.

PR #2297, opened 2024-07-29
`Table` TypeScript migration - Feature branch

didoo:
### :pushpin: Summary

This is the feature branch for the migration of the `Hds::Table` component to TypeScript.

The order of the PRs is the following:
- https://github.com/hashicorp/design-system/pull/2292
- https://github.com/hashicorp/design-system/pull/2289
- https://github.com/hashicorp/design-system/pull/2287
- https://github.com/hashicorp/design-system/pull/2288
- https://github.com/hashicorp/design-system/pull/2290
- https://github.com/hashicorp/design-system/pull/2291
- https://github.com/hashicorp/design-system/pull/2293
- https://github.com/hashicorp/design-system/pull/2311

### :link: External links

Jira ticket: [HDS-2713](https://hashicorp.atlassian.net/browse/HDS-2713)

***

[HDS-2713]: https://hashicorp.atlassian.net/browse/HDS-2713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2311, opened 2024-08-01
Table TypeScript migration - Table component [05]

curthashicorp:
### :pushpin: Summary

Feature branch: https://github.com/hashicorp/design-system/pull/2297

`Table` component TS Migration

### :link: External links

Jira ticket: [HDS-3604](https://hashicorp.atlassian.net/browse/HDS-3604)

***

### 👀 Component checklist

[HDS-3604]: https://hashicorp.atlassian.net/browse/HDS-3604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/components/hds/table/index.ts
```diff
@@ -4,29 +4,110 @@
...
+        setSortBy: (column: string) => void;
+      },
+    ];
+    body?: [
+      {
+        Td: ComponentLike<HdsTableTdArgs>;
+        Tr: ComponentLike<HdsTableTrArgs>;
+        Th: ComponentLike<HdsTableThArgs>;
+        ThSort?: ComponentLike<HdsTableThSortArgs>;
+        data?: string;
```
zamoore, line 86: Should this be a string, or something like `unknown`? It should be the same type as `@model` members I think.

packages/components/src/components/hds/table/index.ts
```diff
@@ -4,29 +4,110 @@
...
-const DEFAULT_VALIGN = 'top';
-
-export default class HdsTableIndexComponent extends Component {
+export default class HdsTableIndexComponent extends Component<HdsTableArgs> {
   @tracked sortBy = this.args.sortBy;
-  @tracked sortOrder = this.args.sortOrder || 'asc';
-  @tracked selectAllCheckbox = undefined;
-  selectableRows = [];
-  @tracked isSelectAllCheckboxSelected = undefined;
+  @tracked sortOrder = this.args.sortOrder || HdsTableThSortOrderValues.Asc;
```
zamoore, line 97:
```suggestion
  @tracked sortOrder = this.args.sortOrder ?? HdsTableThSortOrderValues.Asc;
```

packages/components/src/components/hds/table/index.ts
```diff
@@ -4,29 +4,110 @@
...
+  @tracked selectAllCheckbox:
```
zamoore, line 98: I think this could be

`@tracked selectAllCheckbox?: HdsFormCheckboxBaseSignature['Element'] = undefined`

packages/components/src/components/hds/table/index.ts
```diff
@@ -4,29 +4,110 @@
...
+    | HdsFormCheckboxBaseSignature['Element']
+    | undefined = undefined;
+  selectableRows: HdsTableSelectableRow[] = [];
+  @tracked isSelectAllCheckboxSelected: boolean | undefined = undefined;
```
zamoore, line 102:
```suggestion
  @tracked isSelectAllCheckboxSelected?: boolean = undefined;
```

packages/components/src/components/hds/table/index.ts
```diff
@@ -4,29 +4,110 @@
...
 
   /**
```
zamoore, line 104: We're removing all of the jsdoc style comments. You can delete them as part of the conversion.

packages/components/src/components/hds/table/index.ts
```diff
@@ -4,29 +4,110 @@
...
+  HdsTableColumn,
+  HdsTableDensities,
+  HdsTableHorizontalAlignment,
+  HdsTableSelectableRow,
+  HdsTableSortingFunction,
+  HdsTableThSortOrder,
+  HdsTableVerticalAlignment,
+} from './types';
+
+export const DENSITIES: string[] = Object.values(HdsTableDensityValues);
```
zamoore, line 32: Could this instead be

`export const DENSITIES: HDsTableDensity[] = Object.values(HdsTableDensityValues);`

packages/components/src/components/hds/table/index.ts
```diff
@@ -4,29 +4,110 @@
...
+export const DEFAULT_DENSITY = HdsTableDensityValues.Medium;
+
+export const VALIGNMENTS: string[] = Object.values(
```
zamoore, line 35: Same question as above regarding more specific typing

packages/components/src/components/hds/table/index.ts
```diff
@@ -4,29 +4,110 @@
...
+    sortOrder?: HdsTableThSortOrder;
+    valign?: HdsTableVerticalAlignment;
+  };
+  Blocks: {
+    head?: [
+      {
+        Tr: ComponentLike<HdsTableTrArgs>;
+        Th: ComponentLike<HdsTableThArgs>;
+        ThSort: ComponentLike<HdsTableThSortArgs>;
+        sortBy: string | undefined;
```
zamoore, line 75:
```suggestion
        sortBy?: string;
```

packages/components/src/components/hds/table/index.ts
```diff
@@ -4,29 +4,110 @@
...
+      },
+    ];
+    body?: [
+      {
+        Td: ComponentLike<HdsTableTdArgs>;
+        Tr: ComponentLike<HdsTableTrArgs>;
+        Th: ComponentLike<HdsTableThArgs>;
+        ThSort?: ComponentLike<HdsTableThSortArgs>;
+        data?: string;
+        sortBy?: string | undefined;
```
zamoore, line 87:
```suggestion
        sortBy?: string;
```

zamoore (commented):
Looking pretty good so far! Appreciate you adding the return types to functions.

packages/components/src/components/hds/table/index.ts
curthashicorp: It is actually the identityKey which is a string
{{#each (sort-by this.getSortCriteria @model) key=this.identityKey as |record|}}

packages/components/src/components/hds/table/index.ts
curthashicorp: Since it is possible any of our consumers are passing in an empty string for @sortOrder I am going to leave current code in place and not do any refactors that could possibly break something.

- **Logical OR (`||`)**: Defaults to the right-hand side value if the left-hand side value is any falsy value (e.g., `false`, `0`, `''`, `null`, `undefined`).

- **Nullish Coalescing (`??`)**: Defaults to the right-hand side value only if the left-hand side value is `null` or `undefined`.

### Example
```typescript
@tracked sortOrder = this.args.sortOrder || HdsTableThSortOrderValues.Asc;
// Uses default if this.args.sortOrder is falsy (false, 0, '', null, undefined)

@tracked sortOrder = this.args.sortOrder ?? HdsTableThSortOrderValues.Asc;
// Uses default only if this.args.sortOrder is null or undefined
```

Use `||` for broader defaulting and `??` for more precise defaulting.

packages/components/src/components/hds/table/index.ts
curthashicorp: It could be, but I like using a Union Type with undefined because it explicitly states the possible types, making it clear that undefined is a valid value. But might be a good topic for engineering sync, if we want to define best practices around what we use shorthand for and what we do not.

packages/components/src/components/hds/table/index.ts
curthashicorp: same as above

packages/components/src/components/hds/table/index.ts
curthashicorp: Yes! 
```
export const DENSITIES: HdsTableDensities[] = Object.values(
  HdsTableDensityValues
);
```

packages/components/src/components/hds/table/index.ts
curthashicorp: Gonna dig deeper into the ember/syntax and see. Good call out.

packages/components/src/components/hds/table/index.ts
```diff
@@ -0,0 +1,313 @@
...
+        sortOrder?: HdsTableThSortOrder;
+        setSortBy: (column: string) => void;
+      },
+    ];
+    body?: [
+      {
+        Td: ComponentLike<HdsTableTdArgs>;
+        Tr: ComponentLike<HdsTableTrArgs>;
+        Th: ComponentLike<HdsTableThArgs>;
+        data?: Array<Record<string, unknown>>;
```
alex-ju, line 78: question: checking my understanding here, isn't `data` a single record, instead of an array?

packages/components/unpublished-development-types/global.d.ts
```diff
@@ -25,5 +25,11 @@ declare module '@glint/environment-ember-loose/registry' {
     // we have to use `ember-stargate` version `0.4.3` because version `0.5.0` causes a break in the tests for `ember-lts-3.28`
     Portal: typeof Portal;
     PortalTarget: typeof PortalTarget;
+    'sort-by': HelperLike<{
+      Args: {
+        Positional: [string[], unknown[]];
+      };
+      Return: unknown[];
+    }>;
```
alex-ju, line 33: question: no luck getting this from `ember-composable-helpers`, right?

alex-ju (approved):
Looks good overall! I left two questions, for clarification, but I believe we can get it in to the feature branch then we can do a once over review there

packages/components/unpublished-development-types/global.d.ts
curthashicorp: yeah, unfortunately the library is all JS and I couldn't find any types for it

packages/components/src/components/hds/table/index.ts
curthashicorp: yes, fixing now.

packages/components/unpublished-development-types/global.d.ts
alex-ju: fair

PR #2313, opened 2024-08-02
`CodeBlock`: convert component to TypeScript

shleewhite:
### :pushpin: Summary

If merged, this PR will convert the Code Block component to TypeScript

Only other thing is that I added `@types/prismjs` as a dev dependency.

### :link: External links

Jira ticket: [HDS-3550](https://hashicorp.atlassian.net/browse/HDS-3550)

***

### 👀 Component checklist

[HDS-3550]: https://hashicorp.atlassian.net/browse/HDS-3550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/quick-socks-look.md
+++ b/.changeset/quick-socks-look.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Code Block`: Converted component to TypeScript
--- a/.changeset/quick-socks-look.md
+++ b/.changeset/quick-socks-look.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": minor
 ---
 
-`Code Block`: Converted component to TypeScript
+`CodeBlock`: Converted component to TypeScript
--- a/packages/components/src/components/hds/code-block/index.ts
+++ b/packages/components/src/components/hds/code-block/index.ts
@@ -13,6 +13,12 @@ import { guidFor } from '@ember/object/internals';
 
 import Prism from 'prismjs';
 
+import type { SafeString } from '@ember/template/-private/handlebars';
+import type { ComponentLike } from '@glint/template';
+
+import type { HdsCodeBlockTitleSignature } from './title';
+import type { HdsCodeBlockDescriptionSignature } from './description';
+
 import 'prismjs/plugins/line-numbers/prism-line-numbers';
 import 'prismjs/plugins/line-highlight/prism-line-highlight';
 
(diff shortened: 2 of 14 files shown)


packages/components/src/components/hds/code-block/index.ts
```diff
@@ -97,22 +135,22 @@ export default class HdsCodeBlockIndexComponent extends Component {
...
-    const grammar = Prism.languages[language];
+    const grammar = language ? Prism.languages[language] : undefined;
 
     if (code) {
       next(() => {
         if (language && grammar) {
           this.prismCode = htmlSafe(Prism.highlight(code, grammar, language));
         } else {
-          this.prismCode = htmlSafe(Prism.util.encode(code));
+          this.prismCode = htmlSafe(Prism.util.encode(code).toString());
```
shleewhite, line 153: Question: this is the one bit I'm not super sure works as expected. `Prism.util` returns a type TokenString which doesn't match string which `htmlSafe` expects. Looking at the showcase example for no language set (https://hds-showcase-git-hds-3550-ts-codeblock-hashicorp.vercel.app/components/code-block#language), it seems fine but not sure if other code snippets could break.

shleewhite: the "hasLineWrapping=true with long string" example also doesn't have a language set and it works: https://hds-showcase-git-hds-3550-ts-codeblock-hashicorp.vercel.app/components/code-block#line-wrapping

packages/components/src/components/hds/code-block/index.ts
```diff
@@ -31,8 +37,40 @@ import 'prismjs/components/prism-yaml';
...
-  @tracked prismCode = '';
+type Language =
+  | 'bash'
+  | 'go'
+  | 'hcl'
+  | 'json'
+  | 'log'
+  | 'ruby'
+  | 'shell-session'
+  | 'yaml';
```
alex-ju, line 48: suggestions: I know it may look excessive, but for consistency on how we define argument values in components, could this be an enum in `types.ts` with an exported `HdsCodeBlockLanguages` type?

packages/components/src/components/hds/code-block/index.ts
```diff
@@ -31,8 +37,40 @@ import 'prismjs/components/prism-yaml';
...
+      {
+        Title?: ComponentLike<HdsCodeBlockTitleSignature>;
+        Description?: ComponentLike<HdsCodeBlockDescriptionSignature>;
+      },
+    ];
+  };
+  Element: HTMLDivElement;
+}
+
+export default class HdsCodeBlockIndexComponent extends Component<HdsCodeBlockIndexSignature> {
```
alex-ju, line 72: suggestion: for brevity and consistency with other components
```suggestion
export default class HdsCodeBlockComponent extends Component<HdsCodeBlockSignature> {
```

yarn.lock
```diff
@@ -4116,6 +4116,7 @@ __metadata:
     "@types/ember-qunit": "npm:^6.1.1"
     "@types/ember-resolver": "npm:^9.0.0"
     "@types/ember__destroyable": "npm:^4.0.5"
+    "@types/prismjs": "npm:^1.26.4"
```
alex-ju, line 4119: 🙌

.changeset/quick-socks-look.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Code Block`: Converted component to TypeScript
```
alex-ju, line 5: nit
``` suggestion
`CodeBlock`: Converted component to TypeScript
```

alex-ju (approved):
Looking great! Left a few suggestions

packages/components/src/components/hds/code-block/index.ts
shleewhite: Oh, totally! Updated.

PR #2315, opened 2024-08-03
Website Speed Optimizations - Model Caching

curthashicorp:
### :pushpin: Summary

Further optimize website speed/performance by model caching for show route.


website/app/routes/show.js
```diff
@@ -128,7 +155,7 @@ export default class ShowRoute extends Route {
           });
         }
 
-        return {
+        this.currentModelCache = {
```
didoo, line 158: Consider having this as an object, instead of a single variable whose content is constantly replaced (and so, lost) every time the user goes to another page:
```suggestion
        this.currentModelCache[path] = {
```
In this way you can also remove the `activate()` logic, which is not needed anymore.

didoo (commented):
@curthashicorp overall the direction seems OK. I've left only a suggestion.
Let me know when it's ready to review.

website/app/routes/show.js
```diff
@@ -19,7 +19,16 @@ export default class ShowRoute extends Route {
     },
   };
 
+  modelCache = {};
+
   model(params) {
+    const extractComponentName = (path) => path?.split('/').pop();
+    const componentName = extractComponentName(params.path);
+    // If the model is cached, return it
+    if (this.modelCache[componentName]) {
```
Dhaulagiri, line 28: I _think_ you could safely use `params.path` here directly and avoid the (unlikely) future situation where we have something with the same name in, say, `components/` and `patterns/`. Or did you find a specific reason you wanted to split off just the last part?

curthashicorp: ahh good point, didn't think of that edge case.

didoo: The path is already extracted below (line 33)

website/app/routes/show.js
```diff
@@ -128,7 +135,7 @@ export default class ShowRoute extends Route {
           });
         }
 
-        return {
+        const model = {
```
didoo, line 138: you can remove one computation, if you want
```suggestion
        this.modelCache[path] = {
```

website/app/routes/show.js
```diff
@@ -141,6 +148,11 @@ export default class ShowRoute extends Route {
           hasSidecar,
           relatedComponents,
         };
+
+        // Cache the model
+        this.modelCache[path] = model;
+
+        return model;
```
didoo, line 155: see suggestion above
```suggestion
        return this.modelCache[path];
```

didoo (approved):
Left a suggestion. Apart from that, all good for me 👍

website/app/components/dynamic-template.js
```diff
@@ -25,9 +25,9 @@ export default Component.extend({
     this.templateMap = templateMap;
   },
 
-  componentName: computed('templateString', 'componentId', function() {
+  componentName: computed('templateString', 'componentId', function () {
```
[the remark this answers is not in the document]
curthashicorp: Prettier format on save turned on in vscode. I touched the file to add performance logging and then removed it.  I left them in since they seemed to be reasonable lint cleanups.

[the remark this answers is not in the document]
Dhaulagiri: Ahh. I see what's happening now. For some reason we have prettier disabled at the top of the file (and a few others that were part of that initial scaffolding) which is why this hasn't come up before/isn't failing the lining checks. Sort of puzzling why your editor ignored that and linted the file. It would be cleaner to extract that into a separate commit/PR and we could turn on linting for that file and a few others moving forward.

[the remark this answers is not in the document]
curthashicorp: makes sense, for sure, will do!

Dhaulagiri (approved):
🚀  👏 

*but would encourage continuing to poke around the deployed app _just_ to be sure things are working as any bugs here would likely be pretty subtle)

PR #2320, opened 2024-08-06
feat(components): simplify reexport and add type reexport example

aklkv:
### :pushpin: Summary

* simplify reexport for components (might need to split and clean up some non top level reexports as intentions of this files is to only export top level components, non top level component will never be imported directly)
* type reexport (enum specifically)

### :link: External links

Reference: https://hashicorp.slack.com/archives/C06NMSYDZPA/p1722372651081339

Figma file: [if it applies]

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/components.ts
+++ b/packages/components/src/components.ts
@@ -11,18 +11,28 @@
 
 // Accordion
 export { default as HdsAccordion } from './components/hds/accordion/index.ts';
+export { default as HdsAccordionItem } from './components/hds/accordion/item/index.ts';
 export * from './components/hds/accordion/types.ts';
 
 // Alert
 export { default as HdsAlert } from './components/hds/alert/index.ts';
+export { default as HdsAlertDescription } from './components/hds/alert/description.ts';
+export { default as HdsAlertTitle } from './components/hds/alert/title.ts';
 export * from './components/hds/alert/types.ts';
 
 // AppFooter
 export { default as HdsAppFooter } from './components/hds/app-footer/index.ts';
+export { default as HdsAppFooterCopyright } from './components/hds/app-footer/copyright.ts';
+export { default as HdsAppFooterItem } from './components/hds/app-footer/item.ts';
+export { default as HdsAppFooterLegalLinks } from './components/hds/app-footer/legal-links.ts';
+export { default as HdsAppFooterLink } from './components/hds/app-footer/link.ts';
+export { default as HdsAppFooterStatusLink } from './components/hds/app-footer/status-link.ts';
 export * from './components/hds/app-footer/types.ts';
 
 // AppHeader
 export { default as HdsAppHeader } from './components/hds/app-header/index.ts';
+export { default as HdsAppHeaderHomeLink } from './components/hds/app-header/home-link.ts';
+export { default as HdsAppHeaderMenuButton } from './components/hds/app-header/menu-button.ts';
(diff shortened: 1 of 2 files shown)


didoo:
@aklkv is still in progress or ready to review?

aklkv:
@didoo got confused half way through when I realized that at some point this file started to contain more then just tope level components, which was initial intent. I conceptually this PR is ready for initial feedback but I need to clean it up to only contain top level components and it's types.

didoo:
> @didoo got confused half way through when I realized that at some point this file started to contain more then just tope level components, which was initial intent. I conceptually this PR is ready for initial feedback but I need to clean it up to only contain top level components and it's types.

@aklkv some HDS sub-components are used in code, so we can't export only the top-level ones (that's why we have them in the file). See one example of many here: https://github.com/hashicorp/cloud-ui/blob/226302069218edb3d7e2fc542fc6aad0746a9000/engines/waypoint/addon/components/preview-pane.hbs

aklkv:
@didoo I would need your help to identify which contextual component you know for a fact that would be consumed as template imports world directly. Or an other option is we keep everything exposed

didoo:
> @didoo I would need your help to identify which contextual component you know for a fact that would be consumed as template imports world directly. Or an other option is we keep everything exposed

I think this is something that was in the TODO list for @alex-ju. Alex can you help @aklkv here? Thanks

alex-ju:
> > @didoo I would need your help to identify which contextual component you know for a fact that would be consumed as template imports world directly. Or an other option is we keep everything exposed
> 
> I think this is something that was in the TODO list for @alex-ju. Alex can you help @aklkv here? Thanks

@aklkv's suggestion looks good! I see that you [preserved subcomponents where added](https://github.com/hashicorp/design-system/pull/2320/files#diff-42bc542d3aa5be20e25b64912d937dc931c872cc0523b881c2034b1a313e1886R26-R30). I'm not sure we have an exhausted list as of now, so the best start is to preserve what we have (which you've already done, it seems).

didoo:
@aklkv I've updated a bit the code by re-organizing the imports/exports and adding the missing sub-components (practically every sub-component is public, since it's documented).

didoo (approved):

packages/components/src/components.ts
```diff
@@ -52,7 +62,7 @@ export { default as HdsButton } from './components/hds/button/index.ts';
 export { default as HdsButtonSet } from './components/hds/button-set/index.ts';
 
 // Card
-export { default as HdsCard } from './components/hds/card/container.ts';
+export { default as HdsCardContainer } from './components/hds/card/container.ts';
```
aklkv, line 65: I though we have renamed this component to just be HdsCard a while back? I might be mistaken tho

didoo: No, always been `CardContainer`, in case one day we introduce a generic `Card` component (similar to what is happening for the `StepperStepIndicator/StepperTaskIndicator` elements)

aklkv: Oh ok, I guess we missed it when we introduced component.ts reexport file. I have some fixing to do on the next upgrade in cloud-ui 🤣

shleewhite (approved):
So much more readable than before!

didoo:
@aklkv don't forget to add a changeset to the PR

PR #2321, opened 2024-08-06
Showcase - Remove "color" helper classes

didoo:
### :pushpin: Summary

While working on a PR to port the Showcase implementation to HDS+ I noticed that we could remove the "color" helper classes because they're not actually used

In this PR I have:
- removed logic to add a “predefined” color to the `Shw::Text` component (never implemented, it was ported from the `Hds::Text` implementation)
- removed “color” helper classes in showcase (never used)

***

### 👀 Component checklist


alex-ju (approved):
Nice cleanup!

PR #2326, opened 2024-08-08
`Dropdown::ListItem::Interactive` - MAIN

zamoore:
### :pushpin: Summary

If merged, this PR will allow users to yield dynamic content within the `Hds::Dropdown::ListItem` component rather than through providing a `@text` argument.

This will also deprecate the `@text` argument in favor of yielding a content block.

Placeholder

Placeholder

### :link: External links

Jira ticket: [HDS-3620](https://hashicorp.atlassian.net/browse/HDS-3620)
Figma file: [Figma](https://www.figma.com/design/noyY6dUMDYjmySpHcMjhkN/branch/rdDQOAXN7HX1sj0SyO5ciW/HDS-Components?node-id=2-10&m=dev)

### Child PRs

| Link | Status |
| :--- | :----- |
| [**Documentation**](https://github.com/hashicorp/design-system/pull/2332)| 👁️ In-review |
| [**Codemod**](https://github.com/hashicorp/design-system/pull/2331) | ✅ Merged |

***

### 👀 Component checklist

[HDS-3620]: https://hashicorp.atlassian.net/browse/HDS-3620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/components/hds/dropdown/list-item/interactive.hbs
```diff
@@ -30,7 +30,11 @@
         </span>
       {{/if}}
       <Hds::Text::Body class="hds-dropdown-list-item__interactive-text" @tag="span" @size="200" @weight="medium">
-        {{this.text}}
+        {{#if (has-block)}}
+          {{yield (hash Badge=(component "hds/badge" class="hds-dropdown-list-item__interactive-badge"))}}
```
didoo, line 34: [suggestion] since we're yielding any content anyway, and we're not passing any specific argument to the badge (and the class name `hds-dropdown-list-item__interactive-badge` doesn't have a CSS style) what is the benefit (for the consumers, as well as for us) of yielding as contextual component the badge?

```suggestion
          {{yield}}
```

zamoore: That makes sense, and I considered that, but then I thought of the ApplicationState::Footer component. We specifically yield the intended children there, even though it doesn't provide any benefit. I thought it was more to inform the intention of the component API.

didoo: I see your point 🤔. Maybe a difference (debatable) is that in the `ApplicationState::Footer` we know and want to have only these three yielded components (actions). Is the same here? (in my mind, I expect consumers to start yielding other things, but maybe we want to "guide" them in yielding only a badge. OK, resolving the conversation.

zamoore: That's a good question too. I'm going to talk to Lilith and Heather.

Dhaulagiri:
> This will also deprecate the @text argument in favor of yielding a content block.

A few months ago we had previously settled on actually using [`deprecate`](https://api.emberjs.com/ember/release/functions/@ember%2Fdebug/deprecate) to signal the deprecation for consumers, is that something we can use here to nudge folks away from `@text` usage?

packages/components/src/components/hds/dropdown/list-item/interactive.ts
```diff
@@ -25,15 +27,41 @@ export interface HdsDropdownListItemInteractiveSignature {
...
+    args: HdsDropdownListItemInteractiveSignature['Args']
+  ) {
+    super(owner, args);
+
+    if (args.text !== undefined) {
+      deprecate(
+        'The `@text` argument for "Hds::Dropdown::ListItem::Interactive" has been deprecated. Please put text in the yielded block.',
+        false,
+        {
+          id: 'hds.dropdown.list-item.interactive',
```
Dhaulagiri, line 52: I think this is lefover from the suggestion from copilt?

```suggestion
          id: 'hds.dropdown.list-item.text',
```

zamoore: Maybe it should be `hds.dropdown.list-item.interactive.text`

Dhaulagiri: oh yea. it could be. in practice it _really_ doesn't matter all that much so that seems good to me.

showcase/tests/integration/components/hds/dropdown/list-item/interactive-test.js
```diff
@@ -95,7 +95,7 @@ module(
 
     // CONTENT
 
-    test('it should render the text passed as @text prop', async function (assert) {
+    test('it should not render the text passed as @text prop', async function (assert) {
```
alex-ju, line 98: question: aren't we aiming for a backward-compatible approach? In this case the `@text` should continue to work (even if deprecated), right?

zamoore: Yes, I think this was a mistake in the name of the test. The test is actually checking the opposite.

alex-ju: I think the name of the test was correct, _it should render the text passed as @text prop_, even if it's deprecated, we still support this case.

zamoore: Yes, I meant that I made the mistake and need to revert it. Sorry, that was unclear.

zamoore:
Closed in favor of https://github.com/hashicorp/design-system/pull/2347

PR #2328, opened 2024-08-08
Showcase: Convert showcase components to TypeScript

shleewhite:
### :pushpin: Summary

If merged, this PR converts all the components in the showcase to TypeScript.

Components are:
- Autoscrollable
- Divider
- Flex & FlexItem
- Grid & GridItem
- Label
- Logo
- Outliner
- Placeholder
- Text & all variants

There are also a few changes to the `globals.d.ts` file:
- removed the component types defined in the file and instead import them from a template-registry
- make the types extend `EmberTruthRegistry` and `RenderModifiersRegistry` because some of the components use `eq` and `did-insert`
- removed the eslint ignores because they're not necessary

### :link: External links

Jira ticket: [HDS-3657](https://hashicorp.atlassian.net/browse/HDS-3657)

[HDS-3657]: https://hashicorp.atlassian.net/browse/HDS-3657?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


showcase/app/components/shw/text/index.ts
```diff
@@ -58,18 +71,18 @@ export default class ShwTextIndexComponent extends Component {
...
    * @param variant
+   * @return {TextVariants}
    */
-  get variant() {
-    let { variant } = this.args;
+  get variant(): TextVariants {
+    const { variant } = this.args;
 
-    if (variant) {
-      variant = variant.toLowerCase();
```
shleewhite, line 69: I'm not sure this check is needed anymore. As we convert pages to typescript, if the variant prop is typed incorrectly we can just fix it.

showcase/app/components/shw/text/index.ts
```diff
@@ -115,14 +128,14 @@ export default class ShwTextIndexComponent extends Component {
...
   get classNames() {
-    let classes = ['shw-text'];
+    const classes = ['shw-text'];
 
     // Add a class based on the @variant or @tag arguments
     classes.push(`shw-text-${this.variant}`);
 
     // Add a class based on the @weight argument
     if (this.weight && this.weight !== 'inherit') {
-      classes.push(`shw-text-weight-${this.variant.weight}`);
```
shleewhite, line 125: `this.variant.weight` doesn't exist and judging from how the align part works below, I'm guessing it should just be `this.weight`.

Not sure if this would change the showcase but we'll see!

showcase/types/global.d.ts
```diff
@@ -2,138 +2,29 @@
...
-    background?: string;
-    flex?: string;
-  };
-  Element: HTMLDivElement;
-  Blocks: {
-    default: [];
-  };
-}
-export class ShwPlaceholder extends Component<ShwPlaceholderSignature> {}
+import ShwAutoscrollable from '../app/components/shw/autoscrollable';
```
aklkv, line 11: could we possibly create separate file and import it here as we do with hds and others

showcase/app/components/shw/autoscrollable/index.ts
```diff
@@ -0,0 +1,66 @@
...
+      scheduleOnce('afterRender', this, centerScrollableArea, {
+        element: element,
+        direction: this.args.direction ?? 'both',
+        horizontalShift: this.args.horizontalShift ?? 0,
+        verticalShift: this.args.verticalShift ?? 0,
+      });
+    },
+    // @glint-expect-error
+    // @ts-expect-error - https://github.com/ember-modifier/ember-modifier/issues/774
+    { eager: false }
```
aklkv, line 64: this option should be removed @didoo and I had a PR that dropped it in HDS components

showcase/types/global.d.ts
shleewhite: I can definitely put this in a separate file! And interesting about not needing to add them to the registry, I did notice that when the `shw/label` and `shw/flex/item` weren't included, lines like [this](https://github.com/hashicorp/design-system/blob/01bc8ab3e06744f231d45c0f6726839b630420ab/showcase/app/components/shw/grid/index.hbs#L10) had type errors 

```hbs
  {{yield (hash Label=(component "shw/label"))}}
```

showcase/app/components/shw/text/index.ts
shleewhite: I just confirmed that no instance of Shw::Text passes and incorrect variant prop.

showcase/types/global.d.ts
alex-ju: Indeed, we likely need them at least for contextual components

showcase/app/components/shw/text/index.ts
```diff
@@ -32,15 +25,35 @@ export const MAPPING_VARIANT_TO_TAG = {
...
+}
+
+export default class ShwTextIndexComponent extends Component<ShwTextIndexComponentSignature> {
   /**
    * Get a tag to render based on the `@tag` argument passed or the value of `this.size` (via mapping)
    *
    * @method #tag
    * @return {string} The html tag to use in the dynamic render of the component
    */
   get tag() {
```
alex-ju, line 54:
```suggestion
  get tag(): TextTags {
```

showcase/app/components/shw/text/index.ts
```diff
@@ -115,14 +128,14 @@ export default class ShwTextIndexComponent extends Component {
    * @return {string} The "class" attribute to apply to the component.
    */
   get classNames() {
```
alex-ju, line 129:
```suggestion
  get classNames(): string {
```

alex-ju (approved):
Looking great – thanks for taking the time to get this done

showcase/app/components/shw/flex/index.ts
```diff
@@ -0,0 +1,48 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+import { htmlSafe } from '@ember/template';
+
+import type { ComponentLike } from '@glint/template';
+import type { SafeString } from '@ember/template/-private/handlebars';
```
aklkv, line 10:
```suggestion
import type { SafeString } from '@ember/template';
```

showcase/app/components/shw/frame/index.ts
```diff
@@ -7,8 +7,25 @@ import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
 import { htmlSafe } from '@ember/template';
 
-export default class FrameComponent extends Component {
-  get src() {
+import type { SafeString } from '@ember/template/-private/handlebars';
```
aklkv, line 10:
```suggestion
import type { SafeString } from '@ember/template';
```

showcase/app/components/shw/grid/index.ts
```diff
@@ -0,0 +1,57 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+import { htmlSafe } from '@ember/template';
+import { assert } from '@ember/debug';
+
+import type { ComponentLike } from '@glint/template';
+import type { LabelComponentSignature } from '../label';
+import type { SafeString } from '@ember/template/-private/handlebars';
```
aklkv, line 12:
```suggestion
import type { SafeString } from '@ember/template';
```

showcase/app/components/shw/placeholder/index.ts
```diff
@@ -6,15 +6,31 @@
 import Component from '@glimmer/component';
 import { htmlSafe } from '@ember/template';
 
-export default class PlaceholderIndexComponent extends Component {
+import type { SafeString } from '@ember/template/-private/handlebars';
```
aklkv, line 9:
```suggestion
import type { SafeString } from '@ember/template';
```

PR #2338, opened 2024-08-13
`FlightIcon` => `Hds::Icon` Internal migration - `showcase` & `website`

zamoore:
### :pushpin: Summary

If merged, this PR updates instances of `FlightIcon` within `website` & `showcase` to use the `Hds::Icon` component.

### :link: External links

Jira ticket: [HDS-3707](https://hashicorp.atlassian.net/browse/HDS-3707)
Jira ticket: [HDS-3708](https://hashicorp.atlassian.net/browse/HDS-3708)

### Known Percy issues
- Tooltip - Text change
Screenshot 2024-08-20 at 9 35 46 AM
- Icon - Added a border to the icons in the "Display" section that was missing previously
Screenshot 2024-08-20 at 9 36 43 AM

***

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-3707]: https://hashicorp.atlassian.net/browse/HDS-3707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HDS-3708]: https://hashicorp.atlassian.net/browse/HDS-3708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2339, opened 2024-08-13
`FlightIcon` => `Hds::Icon` Migration - `packages/components`

zamoore:
### :pushpin: Summary

If merged, this PR updates instances of `FlightIcon` within `components` to use the `Hds::Icon` component.

### :link: External links

Jira ticket: [HDS-3709](https://hashicorp.atlassian.net/browse/HDS-3709)

***

[HDS-3707]: https://hashicorp.atlassian.net/browse/HDS-3707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HDS-3708]: https://hashicorp.atlassian.net/browse/HDS-3708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HDS-3709]: https://hashicorp.atlassian.net/browse/HDS-3709?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aklkv:
A few spots you might want to remove references to ember-flight-icons packages:

https://github.com/hashicorp/design-system/blob/64a0cc615e5b2e55506f762b573e5c6e2b5a82f1/packages/components/package.json#L23

https://github.com/hashicorp/design-system/blob/64a0cc615e5b2e55506f762b573e5c6e2b5a82f1/packages/components/package.json#L42

https://github.com/hashicorp/design-system/blob/64a0cc615e5b2e55506f762b573e5c6e2b5a82f1/packages/components/rollup.config.mjs#L45

https://github.com/hashicorp/design-system/blob/64a0cc615e5b2e55506f762b573e5c6e2b5a82f1/packages/components/README.md#L40


once all other PRs are merged CI references needs to be cleaned up as well

packages/components/src/components/hds/alert/index.ts
```diff
@@ -40,7 +40,7 @@ export interface HdsAlertSignature {
   Args: {
     type: HdsAlertTypes;
     color?: HdsAlertColors;
-    icon?: FlightIconSignature['Args']['name'] | false;
+    icon?: HdsIconSignature['Args']['name'] | false;
```
aklkv, line 43: I am seeing a number of variations like `boolean | undefined` would it make sense to expand `HdsIconSignature` @alex-ju @didoo ?

alex-ju: I will assess each instance as I review the code but, if my memory serves me well, only the Alert component should have this option of false (as we want to provide consumers with a way to suppress the default icon, if needed).

packages/components/src/components/hds/application-state/header.hbs
```diff
@@ -11,7 +11,7 @@
   {{/if}}
   {{#if @icon}}
     <div class="hds-application-state__icon">
-      <FlightIcon @color="var(--token-color-foreground-strong)" @name={{@icon}} @size="24" />
+      <Hds::Icon @color="var(--token-color-foreground-strong)" @name={{@icon}} @size="24" />
```
alex-ju, line 14: blocker: since we flipped the flag, similarly to how we removed all `@isInlineBlock={{false}}` instances, where we don't have an `@isInlineBlock` argument we must add `@isInline={{true}}`.

```suggestion
      <Hds::Icon @color="var(--token-color-foreground-strong)" @name={{@icon}} @size="24" @isInline={{true}} />
```

Otherwise we get misaligned icons, also highlighted by VRT

The proposed changed
hds-showcase-git-3709-internal-icon-migrationc-9c4900-hashicorp vercel app_components_application-state

With `@isInline={{true}}`
localhost_3000_components_application-state

.changeset/soft-badgers-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+Migration from `FlightIcon` to `Hds::Icon`
```
alex-ju, line 5: suggestion: we aim to use past tense for the changelog entry

```suggestion
Migrated all internal instances of `FlightIcon` to `Hds::Icon`
```

packages/components/src/components/hds/app-footer/status-link.ts
```diff
@@ -58,16 +58,11 @@ export default class HdsAppFooterStatusLinkComponent extends Component<HdsAppFoo
...
-  get statusIcon(): FlightIconSignature['Args']['name'] {
-    if (this.status && !this.args.statusIcon) {
-      return STATUSES[this.status]?.iconName;
-    }
-    return this.args.statusIcon;
+  get statusIcon(): HdsIconSignature['Args']['name'] | undefined {
+    return (
+      this.args.statusIcon ??
+      (this.status !== undefined ? STATUSES[this.status]?.iconName : undefined)
+    );
```
alex-ju, line 65: thought: I don't think the new logic is equivalent to the old one, can you double check?

packages/components/src/components/hds/badge/index.ts
```diff
@@ -121,7 +121,7 @@ export default class HdsBadgeComponent extends Component<HdsBadgeSignature> {
    * @type {string|null}
    * @default null
    */
-  get icon(): FlightIconSignature['Args']['name'] | null {
+  get icon(): HdsIconSignature['Args']['name'] | null {
```
alex-ju, line 124: should probably be `undefined`
```suggestion
  get icon(): HdsIconSignature['Args']['name'] | undefined {
```

alex-ju (commented):
Did a first pass on this. The main thing is that we're missing a trick (a few instances where `@isInline={{true}}` should be set), otherwise looks promising! Very happy to see this change! Let me know when you address these and I'll have a final look.

packages/components/src/components/hds/flyout/header.ts
```diff
@@ -5,14 +5,14 @@
...
-import type { FlightIconSignature } from '@hashicorp/ember-flight-icons/components/flight-icon';
+import type { HdsIconSignature } from '../icon';
 
 export interface HdsFlyoutHeaderSignature {
   Args: {
     id?: string;
     tagline?: string;
     onDismiss: (event: MouseEvent) => void;
-    icon?: FlightIconSignature['Args']['name'] | false;
+    icon?: HdsIconSignature['Args']['name'] | false;
```
alex-ju, line 15: we don't/shouldn't support `false`
```suggestion
    icon?: HdsIconSignature['Args']['name'];
```

packages/components/src/components/hds/icon-tile/index.ts
```diff
@@ -171,22 +140,10 @@ export default class HdsIconTileComponent extends Component<HdsIconTileSignature
...
 
-  /**
-   * Sets the "secondary" icon name (one of the FlightIcons)
-   *
-   * @param iconSecondary
-   * @type {string|null}
-   * @default null
-   */
-  get iconSecondary(): FlightIconSignature['Args']['name'] | null {
+  get iconSecondary(): HdsIconSignature['Args']['name'] | null {
```
alex-ju, line 143:
```suggestion
  get iconSecondary(): HdsIconSignature['Args']['name'] | undefined {
```

packages/components/src/components/hds/modal/header.ts
```diff
@@ -5,14 +5,14 @@
...
-import type { FlightIconSignature } from '@hashicorp/ember-flight-icons/components/flight-icon';
+import type { HdsIconSignature } from '../icon';
 
 export interface HdsModalHeaderSignature {
   Args: {
     id?: string;
     tagline?: string;
     onDismiss: (event: MouseEvent) => void;
-    icon?: FlightIconSignature['Args']['name'] | false;
+    icon?: HdsIconSignature['Args']['name'] | false;
```
alex-ju, line 15: similarly, we don't support false
```suggestion
    icon?: HdsIconSignature['Args']['name'];
```

packages/components/src/components/hds/application-state/header.hbs
zamoore: We only want to do that in specific locations where using them inline was intentional. It was expected that we might have some visual regressions to clean up, but most vertical alignment is done via flexbox, so we shouldn't see too many issues. We do have an option on the codemod for doing what you're suggesting, but we don't suggest using it.

packages/components/src/components/hds/app-footer/status-link.ts
zamoore: Sure, here's some test cases I put together with copilot

```
1. Both status and statusIcon are undefined:
Input: this.status = undefined, this.args.statusIcon = undefined
Expected Output: undefined

2. status is defined, statusIcon is undefined:
Input: this.status = 'operational', this.args.statusIcon = undefined
Expected Output: STATUSES['operational']?.iconName

3. status is undefined, statusIcon is defined:
Input: this.status = undefined, this.args.statusIcon = 'custom-icon'
Expected Output: 'custom-icon'

4. status and statusIcon are both defined:
Input: this.status = 'ON_TIME', this.args.statusIcon = 'custom-icon'
Expected Output: 'custom-icon'

5.status is null, statusIcon is undefined:
Input: this.status = null, this.args.statusIcon = undefined
Expected Output: undefined

6. status is a valid string that does not exist in STATUSES, and statusIcon is undefined:
Input: this.status = 'UNKNOWN', this.args.statusIcon = undefined
Expected Output: undefined (assuming STATUSES['UNKNOWN'] does not exist)

7. status is undefined, statusIcon is a string with a space:
Input: this.status = undefined, this.args.statusIcon = ' '
Expected Output: ' ' (assuming it's a valid icon name)
```

packages/components/src/components/hds/badge/index.ts
zamoore: Probably no reason to keep the getter then, right?

packages/components/src/components/hds/app-footer/status-link.ts
alex-ju: @KristinLBradley would you be able to help check this change when you have the time (I believe you're the initial author)

packages/components/src/components/hds/badge/index.ts
alex-ju: right, good point, we may be able to get rid of it

showcase/app/styles/app.scss
```diff
@@ -13,6 +13,9 @@
 @import "./typography";
 @import "./globals";
 
+// this file comes from the 'ember-flight-icons' package
+@import "icons";
+
```
alex-ju, line 18: question: we plan to remove this in #2338, right? if so, don't we temporarily need it in `website` too?

packages/components/src/components/hds/app-footer/status-link.ts
alex-ju: the old code reads easier to me, but not strongly opinionated; this is how I read it

> if we have a status but no custom icon (`statusIcon`) return the predefined icon for that status.
> otherwise return the custom icon (`statusIcon`)

I'd ask Kristin or someone else to assess the legibility of the new code.

.changeset/soft-badgers-obey.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
```
alex-ju, line 2: 🤔 the way I see it this is a patch, as we're not introducing any new features

packages/components/src/components/hds/app-footer/status-link.ts
zamoore: Tagged Kristin to take a look.

The way I see it is:

- If we have a `@statusIcon`, we always want to use it.
- If we dont:
  - If we have a status, use it to access available STATUSES, but if not, return nothing

KristinLBradley: @zamoore For me, it's a little harder to read the logic but it's more explicit about what it returns so that's a benefit.

zamoore: @KristinLBradley Thanks for taking a look. That was exactly the intention. It became easier to satisfy typescript after the change. If there are no hard objections, I'm going to leave it as is.

I appreciate the input from both of you!

showcase/app/styles/app.scss
zamoore: We actually don't need it here anymore either after removing the foundations/icon page

PR #2342, opened 2024-08-15
add explicit dotenv dependency

Dhaulagiri:
### :pushpin: Summary

We added a [reference to dotenv](https://github.com/hashicorp/design-system/blob/64a0cc615e5b2e55506f762b573e5c6e2b5a82f1/.changeset/changelog-hds.cjs.js#L13) but are relying on it existing from dependencies in other parts of the monorepo. This makes it more explicit.

```bash
⋊> ~/c/w/design-system on main ◦ npx depcheck                                                                                                                   
Missing dependencies
* dotenv: ./.changeset/changelog-hds.cjs.js
 ```

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
   ],
   "devDependencies": {
     "@changesets/cli": "^2.26.2",
-    "@changesets/get-github-info": "^0.5.2"
+    "@changesets/get-github-info": "^0.5.2",
+    "dotenv": "^16.3.1"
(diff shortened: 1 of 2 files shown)


alex-ju (approved):
Good catch! The value of doing a `depcheck` every once in a while

Dhaulagiri:
I was today many years old when I learned this very helpful command existed 😎

PR #2353, opened 2024-08-21
Accessible Headings: Add tag arguments to components

shleewhite:
### :pushpin: Summary

If merged, this PR will add a `tag` or `titleTag` argument to the following components:
* Accordion
* Alert
* ApplicationState
* CodeBlock
* DialogPrimitive

**Note for reviewers:** It is probably easiest to review this PR by commit because each component was done in a separate commit.

The new argument accepts: "div" or "p", depending on which the component currently uses, and `"h1" | "h2" | "h3" | "h4" | "h5" | "h6"`

**Accordion**
* Add titleTag argument

**Alert**
* Add tag argument to Alert Title

**ApplicationState**
* Update the titleTag argument to only allow: `"div" | "h1" | "h2" | "h3" | "h4" | "h5" | "h6"`

**CodeBlock**
* Add tag argument to CodeBlock Header

**DialogPrimitive**
* add titleTag argument to DialogPrimitive Header
* update styles so changing the tag doesn't add margin

Also updates the showcase so there are examples of customized tags.

**The rendered HTML for Accordion when you pass `titleTag`**
Screenshot 2024-08-21 at 4 08 21 PM

### :link: External links

RFC: https://go.hashi.co/rfc/ds-087
Jira ticket: [HDS-3772](https://hashicorp.atlassian.net/browse/HDS-3772)

***

### 👀 Component checklist

[HDS-3772]: https://hashicorp.atlassian.net/browse/HDS-3772?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/gentle-cows-remain.md
+++ b/.changeset/gentle-cows-remain.md
@@ -2,12 +2,12 @@
 "@hashicorp/design-system-components": minor
 ---
 
-`Accordion`: Add `@titleTag` argument
+`Accordion`: Added `@titleTag` argument
 
-`Alert`: Add `@tag` argument to `[A].Title`
+`Alert`: Added `@tag` argument to `[A].Title`
 
-`ApplicationState`: Update the `@titleTag` argument to only accept `"div" | "h1" | "h2" | "h3" | "h4" | "h5" | "h6"`
+`ApplicationState`: Updated the `@titleTag` argument to only accept `"div" | "h1" | "h2" | "h3" | "h4" | "h5" | "h6"`
 
-`CodeBlock`: Add `@tag` argument to `[CB].Title`
+`CodeBlock`: Added `@tag` argument to `[CB].Title`
 
-`DialogPrimitive`: Add `@titleTag` argument to `DialogPrimitive::Header`
+`DialogPrimitive`: Added `@titleTag` argument to `DialogPrimitive::Header`
--- a/.changeset/gentle-cows-remain.md
+++ b/.changeset/gentle-cows-remain.md
@@ -1,0 +1,13 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Accordion`: Add `@titleTag` argument
+
+`Alert`: Add `@tag` argument to `[A].Title`
+
+`ApplicationState`: Update the `@titleTag` argument to only accept `"div" | "h1" | "h2" | "h3" | "h4" | "h5" | "h6"`
+
+`CodeBlock`: Add `@tag` argument to `[CB].Title`
+
+`DialogPrimitive`: Add `@titleTag` argument to `DialogPrimitive::Header`
--- a/packages/components/src/components/hds/accordion/types.ts
+++ b/packages/components/src/components/hds/accordion/types.ts
@@ -22,7 +22,7 @@ export enum HdsAccordionForceStateValues {
 }
 export type HdsAccordionForceStates = `${HdsAccordionForceStateValues}`;
 
-export enum HdsAccordionItemTagValues {
+export enum HdsAccordionItemTitleTagValues {
   Div = 'div',
   H1 = 'h1',
   H2 = 'h2',
@@ -32,4 +32,4 @@ export enum HdsAccordionItemTagValues {
   H6 = 'h6',
 }
 
-export type HdsAccordionItemTags = `${HdsAccordionItemTagValues}`;
+export type HdsAccordionItemTitleTags = `${HdsAccordionItemTitleTagValues}`;
--- a/packages/components/src/components/hds/accordion/types.ts
+++ b/packages/components/src/components/hds/accordion/types.ts
@@ -21,3 +21,15 @@ export enum HdsAccordionForceStateValues {
   Close = 'close',
 }
 export type HdsAccordionForceStates = `${HdsAccordionForceStateValues}`;
+
+export enum HdsAccordionItemTagValues {
+  Div = 'div',
+  H1 = 'h1',
+  H2 = 'h2',
+  H3 = 'h3',
+  H4 = 'h4',
+  H5 = 'h5',
+  H6 = 'h6',
+}
+
+export type HdsAccordionItemTags = `${HdsAccordionItemTagValues}`;
--- a/packages/components/src/components/hds/alert/title.hbs
+++ b/packages/components/src/components/hds/alert/title.hbs
@@ -2,4 +2,8 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<div class="hds-alert__title hds-font-weight-semibold" ...attributes>{{yield}}</div>
+{{! IMPORTANT: we removed any extra newlines before/after the `let` to reduce the issues with unexpected whitespaces (see https://github.com/hashicorp/design-system/pull/1652) }}
+{{#let (element this.componentTag) as |Tag|}}<Tag
+    class="hds-alert__title hds-typography-body-200 hds-font-weight-semibold"
+    ...attributes
+  >{{yield}}</Tag>{{/let}}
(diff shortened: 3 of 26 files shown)


packages/components/src/styles/components/dialog-primitive.scss
```diff
@@ -89,6 +89,7 @@ $hds-dialog-primitive-vertical-padding: 16px;
 
 .hds-dialog-primitive__title {
   flex-grow: 1;
+  margin: 0;
```
shleewhite, line 92: **Question:** this is the only line in this file I wanted to change.. but I guess my editor reformatted a bunch of stuff. Should I revert to only change this line?

Dhaulagiri: I see the same thing locally, but given these changes don't show up when running `yarn lint` it's possible I gave you the wrong editor settings for vscode 🤔 . I'm curious to see what others say/what their setting for this is:

```
"[scss]": {
        "editor.defaultFormatter": "esbenp.prettier-vscode",
        "editor.formatOnSave": true
    },
```

showcase/app/templates/components/alert.hbs
```diff
@@ -244,4 +244,27 @@
     </SG.Item>
   </Shw::Grid>
 
+  <Shw::Text::H2>With a custom title tag</Shw::Text::H2>
```
alex-ju, line 247: thought: it probably makes more sense for this section to live under 'Content', with a `Shw::Text::H3`

packages/components/src/components/hds/alert/types.ts
```diff
@@ -18,3 +18,15 @@ export enum HdsAlertColorValues {
   Critical = 'critical',
 }
 export type HdsAlertColors = `${HdsAlertColorValues}`;
+
+export enum HdsAlertTagValues {
```
alex-ju, line 22: question: considering this file is shared across Alert-related files, should we be more explicit with the naming (e.g. `HdsAlertTitleTagValues`)?

packages/components/src/components/hds/accordion/types.ts
```diff
@@ -21,3 +21,15 @@ export enum HdsAccordionForceStateValues {
   Close = 'close',
 }
 export type HdsAccordionForceStates = `${HdsAccordionForceStateValues}`;
+
+export enum HdsAccordionItemTagValues {
```
alex-ju, line 25: similarly, `HdsAccordionItemTitleTagValues`

.changeset/gentle-cows-remain.md
```diff
@@ -0,0 +1,13 @@
...
+
+`Accordion`: Add `@titleTag` argument
+
+`Alert`: Add `@tag` argument to `[A].Title`
+
+`ApplicationState`: Update the `@titleTag` argument to only accept `"div" | "h1" | "h2" | "h3" | "h4" | "h5" | "h6"`
+
+`CodeBlock`: Add `@tag` argument to `[CB].Title`
+
+`DialogPrimitive`: Add `@titleTag` argument to `DialogPrimitive::Header`
```
alex-ju, line 13: I suggest we also add `Modal` and `Flyout` here, as consumers may not be familiar with the composition

alex-ju (commented):
Generally looks good! I left a couple of questions and recommendations.

More generally, it would be good to add tests for the newly introduced arguments. We'd also need to update the API in `website` and update some examples (or add new ones), but that can be done in a separate PR if you prefer.

.changeset/gentle-cows-remain.md
shleewhite: Hmm.. we could, but I think in the next section of work we should hard code the value for modal and flyout to be H1.

shleewhite:
> Generally looks good! I left a couple of questions and recommendations.
> 
> More generally, it would be good to add tests for the newly introduced arguments. We'd also need to update the API in `website` and update some examples (or add new ones), but that can be done in a separate PR if you prefer.

@alex-ju I was planning on updating the website in a separate PR! I wanted to take a bit of time to think about how to explain / demonstrate the correct usage.

.changeset/gentle-cows-remain.md
alex-ju: ah, right! then this should be enough for now

packages/components/src/styles/components/dialog-primitive.scss
alex-ju: > Should I revert to only change this line?

yes, please, until we fire it out. it looks like the linter couldn't find the config and used the defaults

alex-ju (approved):
Excellent work!

packages/components/src/styles/components/dialog-primitive.scss
alex-ju: your setup looks correct, Brian

shleewhite: 👍🏽 when I removed that config it didn't make the changes, so maybe there is something I need to fix.

packages/components/src/components/hds/accordion/types.ts
```diff
@@ -21,3 +21,15 @@ export enum HdsAccordionForceStateValues {
   Close = 'close',
 }
 export type HdsAccordionForceStates = `${HdsAccordionForceStateValues}`;
+
+export enum HdsAccordionItemTitleTagValues {
```
zamoore, line 25: @alex-ju I know we often define types like this on a per-component basis. Would it make sense to define these in one spot as HDS header types?

packages/components/src/components/hds/dialog-primitive/header.ts
```diff
@@ -24,6 +27,15 @@ export interface HdsDialogPrimitiveHeaderSignature {
 const NOOP = (): void => {};
 
 export default class HdsDialogPrimitiveHeaderComponent extends Component<HdsDialogPrimitiveHeaderSignature> {
+  /**
```
zamoore, line 30: We opted to discontinue the jsdoc style comments, so you can remove these and don't need to add them in the future.

zamoore (approved):
Great work!

packages/components/src/components/hds/accordion/types.ts
alex-ju: good question. let's bring this to the engineering sync. there may be an opportunity to 'promote' some of our types at library level

shleewhite: Added to the list of topics for a future eng sync!

packages/components/src/components/hds/alert/title.hbs
```diff
@@ -2,4 +2,8 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<div class="hds-alert__title hds-font-weight-semibold" ...attributes>{{yield}}</div>
\ No newline at end of file
+{{! IMPORTANT: we removed any extra newlines before/after the `let` to reduce the issues with unexpected whitespaces (see https://github.com/hashicorp/design-system/pull/1652) }}
```
KristinLBradley, line 5: This is fine (and I see this is already merged) but just to let you know, another way to remove whitespace within Ember HBS files is to use tilde symbols "~" (sometimes referred to as "squishies" in Ember)

An example of using squishies in our code base: https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/rich-tooltip/index.hbs

shleewhite: Oh! Cool! Danger of copy/pasting from older files haha, I have another branch to update the default tags for Modal/Flyout so maybe I can clean that up in that PR.

PR #2355, opened 2024-08-22
Adjust 'full app' examples to be conformant

alex-ju:
### :pushpin: Summary

Updated the 'full app' examples adding `hds-side-nav-hide-when-minimized` class to elements that should be hidden when the SideNav is collapsed.

### :link: External links

Jira ticket: [Slack thread](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1724274496952149)

***

### DIFF
--- a/showcase/app/templates/layouts/app-frame/frameless/demo-full-app-frame-with-app-header.hbs
+++ b/showcase/app/templates/layouts/app-frame/frameless/demo-full-app-frame-with-app-header.hbs
@@ -53,10 +53,10 @@
   <Frame.Sidebar>
     <Hds::SideNav @isResponsive={{true}} @hasA11yRefocus={{false}} @isCollapsible={{true}}>
       <:body>
-        <Hds::SideNav::List aria-label="Dashboard" as |SNL|>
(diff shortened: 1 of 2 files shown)


alex-ju:
> Thanks for adding these.
> 
> When I tab through the SideNav only framed example the Toggle tab gets skipped over though. Is this expected or am I perhaps doing something wrong?
> 
> See: https://hds-showcase-git-alex-ju-adjust-full-app-examples-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-side-nav

Ah, the Toggle tab is the first element in the SideNav, so if you start from the top of the page it should come after the skip link

KristinLBradley (approved):

KristinLBradley:
> > Thanks for adding these.
> > When I tab through the SideNav only framed example the Toggle tab gets skipped over though. Is this expected or am I perhaps doing something wrong?
> > See: https://hds-showcase-git-alex-ju-adjust-full-app-examples-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-side-nav
> 
> Ah, the Toggle tab is the first element in the SideNav, so if you start from the top of the page it should come after the skip link

Thanks, I just realized that I had been accidentally skipping over those.

PR #2240, opened 2024-07-12
`Checkbox`, `Radio`, `RadioCard`, `Toggle` - Convert to TypeScript

alex-ju:
### :pushpin: Summary

Converted form components with `::Group` variant to TypeScript.

### :hammer_and_wrench: Detailed description

The conversion revealed two things in the `::Group` variants:
 - `isRequired` and `isOptional` are not required to be passed down to the `Legend` as the `Fieldset` component receives these arguments and applies them to the Legend itself (this mirrors/aligns with the setup we use in the `::Field` variant)
 - the `@name` argument does not need to be passed into the `Fieldset` component as it's not being used there

### :link: External links

Jira ticket: [HDS-2693](https://hashicorp.atlassian.net/browse/HDS-2693), [HDS-2696](https://hashicorp.atlassian.net/browse/HDS-2696), [HDS-2697](https://hashicorp.atlassian.net/browse/HDS-2697), [HDS-2701](https://hashicorp.atlassian.net/browse/HDS-2701).

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#templates-for-npm-packages))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-2693]: https://hashicorp.atlassian.net/browse/HDS-2693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HDS-2696]: https://hashicorp.atlassian.net/browse/HDS-2696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HDS-2697]: https://hashicorp.atlassian.net/browse/HDS-2697?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[...]

PR #2356, opened 2024-08-22
Convert `SuperSelect` components to TypeScript

alex-ju:
### :pushpin: Summary

Convert `SuperSelect` components to TypeScript

- [x] Converted `SuperSelect::Single` components to TypeScript
- [x] Converted `SuperSelect::Multiple` components to TypeScript

Notable changes:
  - moved from extending the `PowerSelectComponent` class to using the `publicAPI` (this is because we now need to extend the glimmer `Component`; it also makes more sense to uncouple the classes and use the API instead)
  - fixed default results message in `SuperSelect::Single` (changed from `results` to `total` as per design)

### :link: External links

Jira ticket: [HDS-2723](https://hashicorp.atlassian.net/browse/HDS-2723)

***

### 👀 Component checklist

[HDS-2723]: https://hashicorp.atlassian.net/browse/HDS-2723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/afraid-guests-relax.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+Converted `SuperSelect` components to TypeScript
```
shleewhite, line 5: Nit: most of the other ts conversion change sets have been like:
```md
`Table` - Converted component and sub-components to TypeScript
```

packages/components/src/components/hds/form/super-select/types.ts
```diff
@@ -0,0 +1,26 @@
...
+export type HdsFormSuperSelectHorizontalPositions =
+  `${HdsFormSuperSelectHorizontalPositionValues}`;
+
+// map SuperSelect's `horizontalPosition` values to anchoredPositionModifier's `placement` values
+export const HdsFormSuperSelectHorizontalPositionToPlacementValues: Record<
+  // SuperSelect's `horizontalPosition` values
+  HdsFormSuperSelectHorizontalPositionValues,
+  FloatingUIOptions['placement']
+> = {
+  [HdsFormSuperSelectHorizontalPositionValues.Left]: 'bottom-start',
```
shleewhite, line 23: Praise: very slick

packages/components/src/components/hds/form/super-select/single/base.ts
```diff
@@ -0,0 +1,144 @@
...
+    return this.args.searchPlaceholder ?? 'Search';
+  }
+
+  /**
+   * Get the maxWidth to apply to the dropdown
+   * @param dropdownMaxWidth
+   * @type {string}
+   * @default 'none'
+   */
+  get dropdownMaxWidthStyle() {
```
shleewhite, line 115: Nit: missing this return type

packages/components/src/components/hds/form/super-select/multiple/base.ts
```diff
@@ -109,7 +141,7 @@ export default class HdsSuperSelectMultipleBaseComponent extends PowerSelectComp
    * @default 'none'
    */
   get dropdownMaxWidthStyle() {
```
shleewhite, line 143: Nit: missing this return type

packages/components/src/components/hds/form/super-select/multiple/base.ts
```diff
@@ -3,46 +3,78 @@
...
-      placement: this.args.horizontalPosition
-        ? HORIZONTAL_POSITION_MAPPING[this.args.horizontalPosition]
-        : DEFAULT_HORIZONTAL_POSITION,
+      placement: HORIZONTAL_POSITION_MAPPING[this.horizontalPosition],
       offsetOptions: 4,
       enableCollisionDetection: true,
     });
     // prevent PowerSelect from setting position
-    return {};
+    return {
```
shleewhite, line 73: Out of curiosity, why did the return value change?

packages/components/src/components/hds/form/super-select/multiple/field.hbs
```diff
@@ -65,7 +63,7 @@
       @triggerComponent={{@triggerComponent}}
       @triggerId={{F.id}}
       @triggerRole={{@triggerRole}}
-      @typeAheadMatcher={{@typeAheadMatcher}}
+      @typeAheadOptionMatcher={{@typeAheadOptionMatcher}}
```
shleewhite, line 66: Out of curiosity, why did this get renamed?

packages/components/src/components/hds/form/super-select/single/field.hbs
```diff
@@ -12,8 +11,7 @@
   as |F|
 >
   {{! Notice: the order of the elements is not relevant here, because is controlled at "Hds::Form::Field" component level }}
-  {{yield (hash Label=F.Label isRequired=F.isRequired isOptional=F.isOptional)}}
-  {{yield (hash HelperText=F.HelperText Error=F.Error)}}
+  {{yield (hash Label=F.Label HelperText=F.HelperText Error=F.Error)}}
```
shleewhite, line 14: Do you not need to pass isRequired/isOptional anymore?

shleewhite (commented):
Mostly questions! Couple of small things and I think you need to add an export to the `components.ts` file.

packages/components/src/components/hds/form/super-select/multiple/field.hbs
alex-ju: The argument is [wrongly documented as `@typeAheadMatcher`](https://ember-power-select.com/docs/api-reference). [In practice, it is `@typeAheadOptionMatcher`](https://github.com/cibernox/ember-power-select/blob/master/ember-power-select/src/components/power-select.ts#L117). I'll raise this with the author to fix the docs.

packages/components/src/components/hds/form/super-select/single/field.hbs
alex-ju: [We recently discovered that these were not needed](https://github.com/hashicorp/design-system/pull/2240/files#diff-9b906f806254a94975bab30d6ba8378b8d41572aff5eac578645ec7911bfff04R7), as they are part of the label/legend API

packages/components/src/components/hds/form/super-select/multiple/base.ts
alex-ju: We previously returned an empty object, forcing the PowerSelect to fall back to the default values. Now that we moved to TypeScript we need to provide the bare minimum for the `CalculatePositionResult` object to have a type match. I provided some default values for the mandatory properties, but the computation is done by the `anchoredPositionModifier` so it doesn't make much of a difference.

packages/components/src/components/hds/form/super-select/single/base.ts
alex-ju: Good catch! Added in https://github.com/hashicorp/design-system/pull/2356/commits/08c99afae38f74beb19294c618baf00e1eb2ee5e

packages/components/src/components/hds/form/super-select/multiple/base.ts
alex-ju: Good catch! Added in https://github.com/hashicorp/design-system/pull/2356/commits/08c99afae38f74beb19294c618baf00e1eb2ee5e

.changeset/afraid-guests-relax.md
alex-ju: Updated in https://github.com/hashicorp/design-system/pull/2356/commits/217c865c29b101cebdb60b7f19fd91be31f57dac

shleewhite (approved):

PR #2359, opened 2024-08-22
Enterprise Nav: `SideNav` & `AppHeader`—Improve scoping for dark theme styling

KristinLBradley:
### :pushpin: Summary

If merged, this PR adds additional scoping to dark theme styles for `Button` & `Dropdown` components within the `SideNav` & `AppHeader` so when nested within a Dropdown they will not inherit the dark theme.

**Included changes:**

* Scoped styles to apply dark theme styling to Buttons and Dropdowns within SideNav & AppHeader only when they are not nested within another Dropdown
* Added Showcase examples of Button & Dropdown nested within another Dropdown within SideNav & AppHeader

SIDE-NAV NESTED ELEMENTS

image

APP-HEADER NESTED ELMENTS

image

### :link: External links

* Jira ticket: [HDS-3793](https://hashicorp.atlassian.net/browse/HDS-3793)
* Bug example in Vault: https://github.com/hashicorp/vault/pull/28151

***

### 👀 Component checklist

[HDS-3793]: https://hashicorp.atlassian.net/browse/HDS-3793?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/styles/components/app-header.scss
+++ b/packages/components/src/styles/components/app-header.scss
@@ -112,7 +112,6 @@
   .hds-dropdown-toggle-button,
   .hds-dropdown-toggle-icon {
     // Apply dark theme to child interactive components not within a nested dropdown
-    // NOTE: Testing overrides for ember-basic-dropdown as well although likely not something we want to do
     &:not(.hds-dropdown * *, .ember-basic-dropdown-trigger * *, .ember-basic-dropdown-content * *) {
       @include hds-interactive-dark-theme();
 
--- a/packages/components/src/styles/components/app-header.scss
+++ b/packages/components/src/styles/components/app-header.scss
@@ -107,23 +107,27 @@
     gap: inherit;
   }
 
-  // Apply dark theme to child interactive components
+  // Dropdown & Button color theming overrides
   .hds-button,
   .hds-dropdown-toggle-button,
   .hds-dropdown-toggle-icon {
-    @include hds-interactive-dark-theme();
+    // Apply dark theme to child interactive components not within a nested dropdown
+    // NOTE: Testing overrides for ember-basic-dropdown as well although likely not something we want to do
+    &:not(.hds-dropdown * *, .ember-basic-dropdown-trigger * *, .ember-basic-dropdown-content * *) {
+      @include hds-interactive-dark-theme();
 
-    // disabled state:
-    &:disabled,
-    &[disabled],
-    &.mock-disabled,
-    &:disabled:focus,
-    &[disabled]:focus,
-    &.mock-disabled:focus,
-    &:disabled:hover,
-    &[disabled]:hover,
-    &.mock-disabled:hover {
-      @include hds-interactive-dark-theme-state-disabled();
+      // disabled state:
+      &:disabled,
+      &[disabled],
+      &.mock-disabled,
+      &:disabled:focus,
+      &[disabled]:focus,
+      &.mock-disabled:focus,
+      &:disabled:hover,
+      &[disabled]:hover,
+      &.mock-disabled:hover {
+        @include hds-interactive-dark-theme-state-disabled();
+      }
     }
   }
 }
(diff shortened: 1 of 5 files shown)


packages/components/src/styles/components/app-header.scss
```diff
@@ -107,23 +107,27 @@
...
-    &:disabled:focus,
-    &[disabled]:focus,
-    &.mock-disabled:focus,
-    &:disabled:hover,
-    &[disabled]:hover,
-    &.mock-disabled:hover {
-      @include hds-interactive-dark-theme-state-disabled();
+    // Apply dark theme to child interactive components not within a nested dropdown
+    // NOTE: Testing overrides for ember-basic-dropdown as well although likely not something we want to do
+    &:not(.hds-dropdown * *, .ember-basic-dropdown-trigger * *, .ember-basic-dropdown-content * *) {
```
KristinLBradley, line 116: I included selectors to target ember-basic-dropdown since the Vault team chose to create their own custom Dropdown-like component using it instead of using the HDS Dropdown. However, I'm not sure if it's really advisable to include class names from third party components like this. (_**Please let me know your opinions.**_)

**Note:** These complex scoping styles aren't ideal but seem to be a workable solution for now. Ideally, once we begin to implement theming, we can refactor to remove the need for them.

packages/components/src/styles/components/side-nav/header.scss
```diff
@@ -71,27 +71,29 @@
...
-    &:disabled:focus,
-    &[disabled]:focus,
-    &.mock-disabled:focus,
-    &:disabled:hover,
-    &[disabled]:hover,
-    &.mock-disabled:hover {
-      @include hds-interactive-dark-theme-state-disabled();
+    // Apply dark theme to child interactive components not within a nested dropdown
+    // NOTE: Testing overrides for ember-basic-dropdown as well although likely not something we want to do
+    &:not(.hds-dropdown * *, .ember-basic-dropdown-trigger * *, .ember-basic-dropdown-content * *) {
```
KristinLBradley, line 82: I included selectors to target ember-basic-dropdown since the Vault team chose to create their own custom Dropdown-like component using it instead of using the HDS Dropdown. However, I'm not sure if it's really advisable to include class names from third party components like this. (_**Please let me know your opinions.**_)

**Note:** These complex scoping styles aren't ideal but seem to be a workable solution for now. Ideally, once we begin to implement theming, we can refactor to remove the need for them.

packages/components/src/styles/components/app-header.scss
alex-ju: Well, we do use these third-party classes in the SuperSelect, so I wouldn't worry that much.

alex-ju (approved):
I think this is a pragmatic solution. Would you be able to test this in Vault to make sure we're addressing their issue? Then maybe reach out to them after the release to let them know they can safly upgrade?

packages/components/src/styles/components/app-header.scss
alex-ju: Well, we do use these third-party classes in the SuperSelect, so I wouldn't worry that much.

shleewhite (approved):
Looks good to me!

KristinLBradley:
> I think this is a pragmatic solution. ~Would you be able to test this in Vault to make sure we're addressing their issue?~ (just read your comments in Slack).
> 
> And don't forget to add a changelog entry for this fix

Unfortunately, this will not resolve their particular issue as they used custom components with custom styling for their dropdown and buttons within the SideNav. However, if they update their code to use HDS components in the future it will fix the issue. So while this update will not fix Vault's immediate issue on its own, it will prevent similar issues for other product teams (as long as they use HDS components) and it will help the Vault team if they update their code in the future.

PR #2366, opened 2024-08-27
Update Vagrant primary gradient tokens

jorytindall:
### :pushpin: Summary

In the Figma v2.0 library work a discrepancy was uncovered between design and code.

- In code/`style-dictionary` the Vagrant primary gradient stops aliased `Vagrant / 100` and `Vagrant / 200`
- In Figma the gradient "aliases" (not technical an alias) `Vagrant / 300` and `Vagrant / 400`.

Double checking in [HDS Global Foundations](https://www.figma.com/design/rR2zONuQi54Hqeq1RAVt48/HDS-Global-Foundations?m=auto&node-id=0-1&t=6C2j7PSymXWWEUab-1) the implementation seems incorrect.

Before
Screenshot 2024-08-27 at 9 48 11 AM

After
Screenshot 2024-08-27 at 9 47 53 AM

***

### 👀 Component checklist

### DIFF
--- a/.changeset/many-lies-leave.md
+++ b/.changeset/many-lies-leave.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/design-system-tokens': patch
+---
+
+Updated design tokens for Vagrant primary gradient to match Figma.
--- a/packages/tokens/dist/cloud-email/helpers/color.css
+++ b/packages/tokens/dist/cloud-email/helpers/color.css
@@ -1,6 +1,6 @@
 /**
  * Do not edit directly
- * Generated on Wed, 24 Jul 2024 00:26:23 GMT
+ * Generated on Tue, 27 Aug 2024 16:41:13 GMT
  */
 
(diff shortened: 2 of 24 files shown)


jorytindall:
A large number of changes here from the generated file timestamps, I hope I did that right (I built the tokens for a quick reference on the website). I also marked this as a `patch` change, but in looking at the rest of the token package releases we generally have only used `minor` changes, happy to change that if necessary.

didoo:
@Dhaulagiri any idea why there is no Percy test? because it's run only if the `components` folder has changes?

didoo (approved):
👍

Dhaulagiri:
> because it's run only if the components folder has changes?

Yes. It looks like I completely overlooked changes to the tokens package when compiling [this matrix](https://hashicorp.atlassian.net/l/cp/pzc5Pmvn). We could/should update the filter to also include changes to the tokens package to at least trigger the component tests/percy.

PR #2367, opened 2024-08-27
Remove ember-flight-icons as trigger for component CI

Dhaulagiri:
### :pushpin: Summary

Since we no longer rely on ember-flight-icons directly in the components package it's no longer necessary to use it as a trigger for running the component tests/percy


MelSumner (approved):
It makes sense.

zamoore (approved):

PR #2371, opened 2024-08-28
SPIKE Table - Sort by selected

zamoore:
### :pushpin: Summary

THIS PR SHOULD NOT BE MERGED - SPIKE ONLY

### Component Capabilities:

The Table component currently supports both internal and external sorting.
It supports only external pagination, meaning the pagination logic is handled outside the component.

#### Awareness in tr Component:
We need to enable the tr component to recognize when it is being used within a thead. This awareness is crucial because we only want to add the sort button to the selection cell in the thead and not elsewhere.

#### Awareness in Table Component:
The table component must be aware of the selectedItemKey for selected records. This will allow it to manage internal sorting of these items effectively.

#### Internal Sorting and Pagination:
Internal sorting currently doesn’t work well with pagination. Since the table isn’t internally aware of the full dataset (paginated and unpaginated), it only sorts the current page, which can lead to misleading results.

_Recommendation_: Internal sorting should be avoided when using pagination. Instead, external sorting should be used for tables with pagination, especially when selection state needs to be tracked across pages.

### Approach Taken:
- The row element will determine whether its parent container is `thead` when it's inserted and use this information to determine whether or not to add the sort button to the selection cell.
- We use a sentinel value, `__selected__`, as the internal sort key for sorting by selected. This prevents collision with user-provided sort values.

### Review Considerations:
- This is unstyled and has intentionally broken some styles for ease of testing
- I don't love the approach of climbing the DOM to determine the parentage of the `tr` component, so I'm certainly open to suggestions

https://hashicorp.atlassian.net/browse/HDS-3592


packages/components/src/components/hds/table/index.ts
```diff
@@ -92,11 +95,48 @@ export default class HdsTableIndexComponent extends Component<HdsTableArgs> {
...
+        { selectedModels: [], unselectedModels: [] }
+      );
+
+      console.log(selectedModels, unselectedModels);
+
+      return [...selectedModels, ...unselectedModels] as typeof this.args.model;
+    }
+
+    if (typeof sortCriteria === 'string') {
+      return A([...model]).sortBy(sortCriteria);
```
aklkv, line 124: even though direct usage of ember array is allowed it is discouraged, additionally usage of ember array prototype extensions is deprecated and will be removed. https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-array-prototype-extensions.md

Disregard this comment if this is only done do demo but I thought I would call this out

zamoore: Thanks for calling that out, but yeah, this is just a little hacky demo.

packages/components/src/components/hds/table/index.hbs
```diff
@@ -13,6 +13,9 @@
   <thead class="hds-table__thead">
     {{#if @columns}}
       <Hds::Table::Tr
+        @sortOrder={{if (eq "__selected__" this.sortBy) this.sortOrder}}
```
zamoore, line 16: Will probably end up just comparing this to the selectedItemKey directly instead of using this string constant.

packages/components/src/components/hds/table/index.ts
```diff
@@ -109,6 +112,73 @@ export default class HdsTableIndexComponent extends Component<HdsTableArgs> {
     }
   }
 
+  sortBySelected(
```
zamoore, line 115: Going to condense sort to one function and not use the reducer implementation.

zamoore:
Going another direction. See this draft PR, but feel free to leave comments in either place https://github.com/hashicorp/design-system/pull/2387/files

zamoore:
Closing in favor of linked component PR

packages/components/src/components/hds/table/index.ts
alex-ju: Nonetheless, `eslint-plugin-ember` does need updating. Will make a note of that

PR #2358, opened 2024-08-22
`Modal`, `Flyout`: make title an H1

shleewhite:
### :pushpin: Summary

If merged, this PR would change the HTML element wrapping the Modal and Flyout tagline + title in the heading to an H1 (the previous element was a div).

### :hammer_and_wrench: Detailed description

* update the Modal and Flyout title to be H1s
* add tests to ensure the correct tag is rendered
* update the Flyout showcase page heading structure so it doesn't jump from H1 to H3 (this causes an automated a11y test failure)

### :camera_flash: Screenshots

**New Modal markup:**
Screenshot 2024-08-22 at 12 10 24 PM

### :link: External links

RFC: https://go.hashi.co/rfc/ds-087
Jira ticket: [HDS-3773](https://hashicorp.atlassian.net/browse/HDS-3773)

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#templates-for-npm-packages))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-3773]: https://hashicorp.atlassian.net/browse/HDS-3773?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2365, opened 2024-08-27
Website: add documentation for `tag` and `titleTag` arguments

shleewhite:
### :pushpin: Summary

If merged, this PR would add content in each of the component's "How to use" section and include the new argument in the "Component API" section.

Changed pages:
* [Accordion](https://hds-website-3w235i59d-hashicorp.vercel.app/components/accordion?tab=code#titletag)
* [Alert](https://hds-website-3w235i59d-hashicorp.vercel.app/components/alert?tab=code#tag)
* [ApplicationState](https://hds-website-3w235i59d-hashicorp.vercel.app/components/application-state?tab=code#titletag)
* [CodeBlock](https://hds-website-3w235i59d-hashicorp.vercel.app/components/code-block?tab=code#title-tag)
* [DialogPrimitive](https://hds-website-3w235i59d-hashicorp.vercel.app/utilities/dialog-primitive?tab=code#header-titletag)

### :link: External links

RFC: https://go.hashi.co/rfc/ds-087
Jira ticket: [HDS-3774](https://hashicorp.atlassian.net/browse/HDS-3774)

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-3774]: https://hashicorp.atlassian.net/browse/HDS-3774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2375, opened 2024-08-29
fix: ignore heading order rule on modal/flyout docs pages

shleewhite:
### :pushpin: Summary

If merged, this PR should fix the failing website tests on the modal and flyout pages introduced in this PR: https://github.com/hashicorp/design-system/pull/2358

To fix the failing tests, instead of changing the components - we decided to turn off [the heading-order](https://dequeuniversity.com/rules/axe/4.8/heading-order?application=axeAPI)  rule because it is better for the component to be correct in product than have correct heading structure on the docs site.

### DIFF
--- a/website/tests/acceptance/components/flyout-test.js
+++ b/website/tests/acceptance/components/flyout-test.js
@@ -19,7 +19,13 @@ module('Acceptance | components/flyout', function (hooks) {
   test('Components/flyout page passes automated a11y checks', async function (assert) {
     await visit('/components/flyout');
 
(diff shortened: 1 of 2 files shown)


MelSumner (approved):
Thanks for the fast follow-up!

didoo:
@shleewhite I don't know if you've seen this comment https://github.com/hashicorp/design-system/pull/2365#pullrequestreview-2267721888 but one option (I _think_) could have been to disable the rendering of the code snippets

PR #2383, opened 2024-09-03
`Accordion`: make toggle button names unique

shleewhite:
### :pushpin: Summary

If merged, this PR would update the accordion toggle to make the button names unique.

To make `aria-label` not required on each item but still create unique labels for the toggle buttons, I decided to use `aria-labelledby`. `aria-labelledby` is set to the id of the wrapper around the toggle block and I removed the `aria-label` default value. This makes it so the accessible name is just the title of the accordion. 

If the developer passes an `aria-label` to the accordion item, `aria-labelledby` isn't set.

**NOTE:**

I used to have a different implementation where I also included a visually hidden piece of text that said "Expand" so the button name would be "Expand {title}", but after discussing with @MelSumner we decided that wasn't even necessary because the fact that it shows/hide content is implied by `aria-expanded`. When you navigate to the button with VoiceOver it says "{title}, collapsed, button" and then when you press the button it says "expanded". this makes including the word "expand" in the button name redundnant.

**DOM after changes:**

### :link: External links
Jira ticket: [HDS-3809](https://hashicorp.atlassian.net/browse/HDS-3809)
***

### 👀 Component checklist

[HDS-3809]: https://hashicorp.atlassian.net/browse/HDS-3809?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/accordion/item/button.hbs
+++ b/packages/components/src/components/hds/accordion/item/button.hbs
@@ -10,6 +10,7 @@
   aria-controls={{@contentId}}
   aria-expanded={{if @isOpen "true" "false"}}
   aria-label={{@ariaLabel}}
+  aria-labelledby={{@ariaLabelledBy}}
   ...attributes
 >
   <Hds::Icon @name="chevron-down" @size={{if (eq @size "large") "24" "16"}} />
--- a/.changeset/khaki-bats-admire.md
+++ b/.changeset/khaki-bats-admire.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Accordion`: changed the default name of the `Accordion` item toggles. Now, they are labelled by the content in the `Accordion` title.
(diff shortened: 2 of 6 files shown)


zamoore (commented):
Cool change!

What do you think about adding an integration test to make sure the property value is set correctly?

shleewhite:
> Cool change!
> 
> What do you think about adding an integration test to make sure the property value is set correctly?

@zamoore for sure! just added a couple tests.

packages/components/src/components/hds/accordion/item/button.hbs
```diff
@@ -10,7 +10,9 @@
   aria-controls={{@contentId}}
   aria-expanded={{if @isOpen "true" "false"}}
   aria-label={{@ariaLabel}}
+  aria-labelledby={{@ariaLabelledBy}}
   ...attributes
 >
+  <div id={{@id}} aria-hidden="true" class="sr-only">Expand</div>
```
KristinLBradley, line 16: Would it be possible/practical to update the Button aria-label instead of adding a nested element?

If you do use a nested element like this it should be changed to a `span` tag as block-level tags shouldn't be nested inside HTML buttons as they are inline-level.

KristinLBradley (approved):
Looks good!

PR #2384, opened 2024-09-03
`AppSideNav` - Component branch (HDS-3800)

KristinLBradley:
_**STATUS:** Updated to deprioritize component adoption and unpublish AppSideNav & AppHeader. **—Ready for release barring any last feedback**._

----

### :pushpin: Summary

If merged, this PR will add the `AppSideNav` component to HDS but will not publish it. It will also un-publish the related `AppHeader` component.

The majority of code was duplicated from the `SideNav` and then modified to implement the new `AppSideNav` component.

**Major differences of note:**

* The `AppSideNav` has an updated visual design using lighter colors than the `SideNav`
  - Dark theme styles are no longer added to nested Button & Dropdown components
* The AppSideNav structure was simplified:
  - Structural named blocks were removed: `:header`, `:footer`, `:body`, `:root`
  - The `:logo` & `:actions` child named blocks were removed (used in old :header)
  - The `Header` and `Header::HomeLink` child components were removed.
* The Navigator Narrator (skip link) was removed as the AppSideNav is meant to always be paired with the `AppHeader` which already includes this
* Bug with overflow scrolling was fixed (caused by navigating from panel with long content to panel with short content)

([Spike of restructured List and ListTitle](https://github.com/hashicorp/design-system/pull/2478/files) which may be explored in later iterations)

(Internal references: [RFC](https://hermes.hashicorp.services/document/1GcQVJZjZKgS_S9OFDUxUpwn5UJEJR5YhmrPyA5YSWfs), [Meta CRD](https://docs.google.com/document/d/11Dg4AQMfYssWfMuZJtpVG1WE6N1HbARatCmjkaO1rp4/edit#heading=h.8ifqih738u0a))

### :link: External links

- Jira ticket: [HDS-3800](https://hashicorp.atlassian.net/browse/HDS-3800)
- [Figma file](https://www.figma.com/design/noyY6dUMDYjmySpHcMjhkN/branch/WVcyo3y0R8Cf5ZatlfUkuk/HDS-Components?node-id=62374-9779&node-type=CANVAS&t=ByhuMTHaX9kgaHdk-0)

***

### 👀 Component checklist

- ~~[ ] A changelog entry was added via [Changesets]~~ (_Removed changelog as component will not be published_) (https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

[HDS-3800]: https://hashicorp.atlassian.net/browse/HDS-3800?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/components/hds/app-side-nav/portal/target.ts
```diff
@@ -0,0 +1,193 @@
...
+  Element: HTMLDivElement;
+}
+
+export default class HdsAppSideNavPortalTargetComponent extends Component<HdsAppSideNavPortalTargetSignature> {
+  @service router!: Services['router'];
+
+  @tracked numSubnavs = 0;
+  @tracked lastPanelEl: Element | undefined;
+
+  static get prefersReducedMotionOverride(): boolean {
```
shleewhite, line 43: Blocking: since Ember.testing is going to be deprecated, can you match what SideNav has now? https://github.com/hashicorp/design-system/blob/8d459d3326a16173674de1ab68d6a694816729e7/packages/components/src/components/hds/side-nav/portal/target.ts#L43

website/tests/acceptance/components/app-side-nav-test.js
```diff
@@ -0,0 +1,37 @@
...
+  //       'duplicate-id-active': { enabled: false },
+  //     },
+  //   });
+
+  //   await visit('/components/app-side-nav');
+
+  //   await a11yAudit();
+
+  //   assert.ok(true, 'a11y automation audit passed');
+  // });
```
KristinLBradley, line 36: This is being re-enabled in the docs branch. I accidentally added the file here initially.

packages/components/src/components/hds/app-side-nav/index.ts
```diff
@@ -0,0 +1,195 @@
...
+  get showToggleButton(): boolean {
+    return (this.isResponsive && !this.isDesktop) || this.isCollapsible;
+  }
+
+  /**
+   * @param ariaLabel
+   * @type {string}
+   * @default 'close menu'
+   */
+  get ariaLabel(): string {
```
shleewhite, line 105: Nit: maybe this could be `toggleButtonAriaLabel` just so its clearer what this is for.

packages/components/src/components/hds/app-side-nav/index.ts
```diff
@@ -0,0 +1,195 @@
...
+    if (this.isMinimized) {
+      return this.args.ariaLabel ?? 'Open menu';
```
shleewhite, line 107: Issue: since this isn't actually a menu, I think better alt text would be "App navigation" or something like that and then use `aria-expanded` to let users know if it is expanded or not... like a reveal.

showcase/tests/acceptance/components/hds/app-side-nav-test.js
```diff
@@ -0,0 +1,33 @@
...
+import { a11yAudit } from 'ember-a11y-testing/test-support';
+
+module('Acceptance | Component | hds/app-side-nav', function (hooks) {
+  setupApplicationTest(hooks);
+
+  test('Components/app-side-nav page passes a11y automated checks', async function (assert) {
+    let axeOptions = {
+      rules: {
+        listitem: {
+          enabled: false,
```
shleewhite, line 18: Question: is there a selector you could use to not disable this on the whole page?

packages/components/src/components/hds/app-side-nav/base.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import TemplateOnlyComponent from '@ember/component/template-only';
+
+export interface HdsAppSideNavBaseSignature {
+  Blocks: {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    root?: any;
```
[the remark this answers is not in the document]
KristinLBradley: I can update it. The majority of the code was taken from the old SideNav.

packages/components/src/components/hds/app-side-nav/list/link.hbs
```diff
@@ -0,0 +1,50 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<Hds::AppSideNav::List::Item>
```
jorytindall, line 6: Not sure exactly where to comment these, but I have a few notes from testing all related to list items (I think):

1. (Bear with me, somewhat confusing language) for the standard list items (ones that aren't active/current), the active/mouse down state is the same as the resting/default state for the active/current list item, where they should follow the interactive states outlined [here](https://www.figma.com/design/noyY6dUMDYjmySpHcMjhkN/branch/WVcyo3y0R8Cf5ZatlfUkuk/HDS-Components?m=auto&node-id=64810-18282&t=Pt2sSNKzYnjO4dMl-1). Let me know any questions about this, I tried to outline it a bit in Figma.

2. The Active/Current link still has a focus state which I found somewhat strange, I think we determined that there wouldn't be interactive states for this element because it doesn't change the view/route in the UI. Additionally, the design changes somewhat when it's focused.
Screenshot 2024-09-17 at 4 04 41 PM

3. It looks like the focus ring on the "standard" list items is getting cut off a bit by the list item below. I think this is caused by the list items having a background color in their default state (which is the same as the background color for the container of the component. What do you think about removing that and instead having a transparent background in the resting state?
Screenshot 2024-09-17 at 3 42 33 PM

4. When the components expands from it's collapsed state, the blue border for the active/current link looks like it's stretched? It's kind of a strange visual bug, probably because the width of the list item is expanding which causes the height to decrease. I tried to capture a screen recording of it:
CleanShot 2024-09-17 at 15 47 14

packages/components/src/styles/components/app-side-nav/toggle-button.scss
```diff
@@ -0,0 +1,116 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+//
+// APP-SIDE-NAV > TOGGLE BUTTON
+//
+
+.hds-app-side-nav__toggle-button {
```
jorytindall, line 10: Bit of a nit: There's a small visual bug in the hover and active states of this collapse toggle button that appears to be related to some of those issues we were discussing with the svg clip path. Very hard and not really perceptible if you aren't paying attention.

Hover:
Screenshot 2024-09-17 at 3 38 34 PM

Active:
Screenshot 2024-09-17 at 3 38 46 PM

jorytindall (commented):
I did a bunch of testing here and found a few weird visual bugs and small things. Let me know if you want to pair on these or go into more detail!

packages/components/src/components/hds/app-side-nav/index.hbs
```diff
@@ -0,0 +1,37 @@
...
+      />
+    {{/if}}
+  </:root>
+  <:header as |Header|>
+    {{~yield (hash Header=Header isMinimized=this.isMinimized) to="header"~}}
+  </:header>
+  <:body as |Body|>
+    {{~yield (hash Body=Body isMinimized=this.isMinimized) to="body"~}}
+  </:body>
+  <:footer as |Footer|>
```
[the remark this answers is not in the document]
KristinLBradley: I thought it would be more flexible/friendly for consumers if we left optional named blocks where they could insert "header" or "footer" content if they really needed/wanted to.

For the header area, I did remove two child components we had that were originally used within it but I left the `:header` named block itself.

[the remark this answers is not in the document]
KristinLBradley: @alex-ju (or others reading this) Do you have an opinion on whether we should keep the `:header` & `:footer` named blocks or if it would be best to remove them?

packages/components/src/components/hds/app-side-nav/index.ts
KristinLBradley: Let's leave discussion of changing this for later because this is what it was in the SideNav.

@alex-ju Do you have any opinions on the wording?

packages/components/src/components/hds/app-side-nav/list/link.hbs
```diff
@@ -0,0 +1,50 @@
...
+  <Hds::Interactive
+    class="hds-app-side-nav__list-item-link {{if @isActive 'active'}}"
```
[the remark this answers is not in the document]
KristinLBradley: Consumers would have to do this I believe. We could mention something in the documentation though to give guidance.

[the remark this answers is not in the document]
KristinLBradley: Actually let me see. I was thinking about this in terms of the generic component not the "active" one...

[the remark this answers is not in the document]
KristinLBradley: I'm adding it & updating the related test. thanks

packages/components/src/components/hds/app-side-nav/list/link.ts
```diff
@@ -0,0 +1,30 @@
...
+import TemplateOnlyComponent from '@ember/component/template-only';
+
+import type { HdsIconSignature } from '../../icon';
+import type { HdsInteractiveSignature } from '../../interactive';
+
+export interface HdsAppSideNavListLinkSignature {
+  Args: HdsInteractiveSignature['Args'] & {
+    icon?: HdsIconSignature['Args']['name'];
+    iconColor?: string; // TODO: Document in API web docs
+    text?: string;
```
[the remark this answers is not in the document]
KristinLBradley: It uses `Hds::Interactive` under the hood so it's not necessarily literally a link. So I will leave as is.

packages/components/src/components/hds/app-side-nav/portal/target.ts
```diff
@@ -0,0 +1,193 @@
...
+
+  @action
+  didUpdateSubnav(element: HTMLElement, [count]: [number]): void {
+    this.animateSubnav(element, [count]);
+  }
+
+  @action
+  animateSubnav(element: HTMLElement, [count]: [number]): void {
+    /*
+     * Here is what the layout looks like for this setup
```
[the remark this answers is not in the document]
KristinLBradley: This is all from @didoo :)

packages/components/src/components/hds/app-side-nav/list/title.hbs
```diff
@@ -0,0 +1,11 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<Hds::AppSideNav::List::Item>
+  <div
```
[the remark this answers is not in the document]
KristinLBradley: This is valid but let's discuss more. @hashicorp/hds-engineering 

We could possibly do this as a follow on as well.

showcase/tests/acceptance/components/hds/app-side-nav-test.js
KristinLBradley: There may be but let's look at improvements to this as a follow on. This is all taken from the existing SideNav.

packages/components/src/components/hds/app-side-nav/index.ts
```diff
@@ -0,0 +1,195 @@
...
+        Footer?: HdsAppSideNavBaseSignature['Blocks']['footer'];
+        isMinimized?: boolean;
+      },
+    ];
+  };
+  Element: HdsAppSideNavBaseSignature['Element'];
+}
+
+export default class HdsAppSideNavComponent extends Component<HdsAppSideNavSignature> {
+  @tracked isResponsive = this.args.isResponsive ?? true; // controls if the component reacts to viewport changes
```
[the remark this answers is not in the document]
KristinLBradley: Should this be done now or is it ok to do as a follow on?

packages/components/src/components/hds/app-side-nav/list/link.hbs
KristinLBradley: Trying to go through these one by one... Btw, it may be helpful to look at the Showcase examples for these as well: https://hds-showcase-git-hds-3800-app-sidenav-component-hashicorp.vercel.app/components/app-side-nav

* For the focus state on the "active" link, that's there because it was in the original SideNAv. I'll look to remove.

packages/components/src/components/hds/app-side-nav/list/link.hbs
```diff
@@ -0,0 +1,52 @@
...
+    tabindex={{if @isActive "-1"}}
```
KristinLBradley, line 9: @shleewhite @alex-ju I added a negative tab index to prevent tabbing to the "active" Link from Jory's feedback above. Is this an acceptable solution?

packages/components/src/styles/components/app-side-nav/toggle-button.scss
KristinLBradley: Would it be possible to not use a semi-transparent border color for the AppSideNav? That would fix issues like this without the need for fiddly CSS like the clip-path I added.

packages/components/src/components/hds/app-side-nav/index.ts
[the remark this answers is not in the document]
shleewhite: I've done more research on the rule and actually I think this implementation will work. It was just a problem if you set 

```
@tracked isResponsive = this.args.isResponsive
```

Having the default be `true` makes it work.

packages/components/src/components/hds/app-side-nav/list/link.hbs
shleewhite: Hmm.. I think we should discuss this. It would be an a11y violation to remove a non-disabled interactive element from the tab order. What are the concerns with having it focusable? Just that a ring is visible?

KristinLBradley: Perhaps it would be better to try to change it from a link/button to plain text when it's "active"? I don't know the reasoning behind why it was kept as a link in the SideNav when it links to the same page and is thus not useful (or I assume it's not useful).

@alex-ju do you have any thoughts around this?

packages/components/src/components/hds/app-side-nav/list/link.hbs
jorytindall: Ok for the first point, everything looks like it's mapped correctly, but when I click on a link (mouse down), it appears that that active state is is the same as the active/current link. But I wonder if this is a showcase issue because all of the links are local. Might not be an issue outside of the showcase, but attaching a screen recording just in case (excuse the jumping to a new page).

CleanShot 2024-09-19 at 10 15 51

KristinLBradley: Pushed up a style change that seems to fix this.

KristinLBradley: @jorytindall When you have a chance, could you please test again and let me know if it needs any more adjusting?

* [AppSideNav Showcase](https://hds-showcase-git-hds-3800-app-sidenav-component-hashicorp.vercel.app/components/app-side-nav)
* [Framed example used in context](https://hds-showcase-git-hds-3800-app-sidenav-component-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-app-header-and-app-side-nav)

packages/components/src/styles/components/app-side-nav/toggle-button.scss
KristinLBradley: @heatherlarsen Do you have any opinions/input you could share? There are issues with the semi-transparent border that affect the toggle tab design as shown. I came up with a partial mitigation but it's fiddly and there are remaining visual artifacts that are very difficult to fix as Jory screen-shotted above (I can keep experimenting though and will post an update here if I'm successful).

jorytindall: I think for this scenario it makes sense to use a neutral color in place of the `Border/Primary` to solve this issue of transparency and overlap. In this case the closest is `Neutral/200`, I've already applied this to the component in [Figma](https://www.figma.com/design/noyY6dUMDYjmySpHcMjhkN/branch/WVcyo3y0R8Cf5ZatlfUkuk/HDS-Components?m=auto&node-id=59981-56567&t=NYJ9XGY2GiVrrpJZ-1) and it cleans up some of those problems already, so I'm good with making the slightly less systematic decision.

packages/components/src/components/hds/app-side-nav/list/link.hbs
jorytindall: Nice, yep, that seems to fix it!

packages/components/src/styles/components/app-side-nav/toggle-button.scss
KristinLBradley: Ok, I zoomed in and kept tweaking values in the browser until I got the clip paths looking better and figured out an issue with box-shadows that was also impacting rendering. I think I managed to come up with something that improves the rendering a good bit now but see what you think. @jorytindall

KristinLBradley: I didn't see your comment above before I posted above, sorry. So should I just update the border color to use the solid color now then?

jorytindall: Yeah, I think so, this is a bit more of a simple solution and is probably a bit more maintainable. Thanks for throwing that together though!

KristinLBradley: Updated the border color in tokens and removed CSS which was no longer needed.

packages/components/src/components/hds/app-side-nav/list/link.hbs
KristinLBradley: I discussed with @jorytindall and I'm going to undo this. This means that the "active" link will have a focus border again as it did in the SideNav.

We can discuss if we want to change this behavior as a follow on. (To avoid over-complicating this PR.)

showcase/app/templates/components/app-side-nav.hbs
```diff
@@ -0,0 +1,883 @@
...
+    <Nav.Item><Shw::Placeholder @height="200px" @text="portaled content" /></Nav.Item>
+    <Nav.extraAfter><Shw::Placeholder @height="72px" @text="extraAfter" @background="#f3d9c5" /></Nav.extraAfter>
+  </Hds::AppSideNav::Portal>
+
+</section>
+
+<Shw::Divider @level={{2}} />
+
+<section data-test-percy>
+  <Shw::Text::H2>Examples of sidebar navigation</Shw::Text::H2>
```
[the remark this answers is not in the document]
KristinLBradley: I added an example. I'll see if I can add other more realistic use case examples later on as I get to testing more within product apps.

packages/components/src/components/hds/app-side-nav/base.ts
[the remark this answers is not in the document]
alex-ju: Indeed, as discussed during the last engineering sync; we'll do the work to remove all `any` instances (either by replacing them with `unknow` or `[]`

showcase/app/templates/layouts/app-frame/frameless/demo-full-app-frame-with-app-header-and-app-side-nav.hbs
```diff
@@ -0,0 +1,280 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+{{! @glint-nocheck: not typesafe yet }}
```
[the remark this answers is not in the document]
alex-ju: this has been corrected in other frameless examples; still having it here is a sign we'd need to _rebase_ this branch to `main` before merging it in

packages/components/src/components/hds/app-side-nav/list/link.ts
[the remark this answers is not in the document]
alex-ju: for context, @shleewhite: the interactive Interactive allows multiple ways of creating an actual `href` value (via `route`, `query`, `model`, etc.)

packages/components/src/components/hds/app-side-nav/list/title.hbs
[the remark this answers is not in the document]
alex-ju: The hint in the class name (`hds-app-side-nav__list-title`) would suggest the same. We can write a follow-up task for this if you prefer, Kristin

packages/components/src/components/hds/app-side-nav/portal/target.ts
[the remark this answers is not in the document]
alex-ju: I think it precedes Cristiano as well; [credits](https://github.com/hashicorp/ember-shared-components/commit/0b4ed0cbec442ba537b226a7e588c58ad6864f92#diff-c5e6e580dd41c6a359f2e282eb72879fa123e11d2df1b0bd5c9d613441fea782R52) to @meirish for the ascii art

packages/components/src/components/hds/app-side-nav/index.ts
alex-ju: Lee's suggestion makes sense to me, but we can only change it if we add the `aria-expanded` attribute. If you'd rather do this as a follow-up, we can.

packages/components/src/components/hds/app-side-nav/index.hbs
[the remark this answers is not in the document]
alex-ju: I don't have the complete historical context of these, but I believe the `:footer` was added for the org selector (that was custom across products at the time), likewise for the `:header` (elements that are now in the AppHeader). I would be tempted to audit the current usage, and if we don't find them in use to simplify the API.

packages/components/src/components/hds/app-side-nav/list/link.hbs
alex-ju: Thanks for the update, Kristin. If we don't want it to be focused I see two options: either render it as a disabled button (meaning `aria-current="page"` will not be surfaced), or it could be plain text (or even better the h2 that would be overarching the rest of the SideNav titles, as Lee suggested in another comment above).

alex-ju (commented):
I did a high-level review and looked at all the open conversations and here are my thoughts:
 - investigate if we need the named blocks in `Hds::AppSideNav` to accommodate all products or if we can remove them and converge _(KB: I would like to look at this in a follow-on if ok.)_
 - ✅ ~~there is a fix for the inert synchronization that would need to be ported to this PR~~ _(KB: Ported code over)_
 - ✅ ~~the skip link doesn't work in `demo-full-app-frame-with-app-header-and-app-side-nav` (I suspect a rebase before merge will resolve this)~~ _(KB: Fixed)_
 - ✅ ~~we don't seem to be having enough contrast in the AppSideNav titles~~ _(KB: Investigated w/ @jorytindall's help & updated value)_

Screenshot 2024-09-20 at 12 09 51

packages/components/src/components/hds/app-side-nav/list/title.hbs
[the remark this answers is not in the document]
KristinLBradley: I would prefer to do this as a follow on as this is how it was in the SideNav. (I want to be able to wrap up the main work related to visual details for this PR and then do improvements like this as a fast follow on task.)

showcase/app/templates/layouts/app-frame/frameless/demo-full-app-frame-with-app-header-and-app-side-nav.hbs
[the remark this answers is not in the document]
KristinLBradley: I merged in the latest `main`. This is now gone.

packages/components/src/components/hds/app-side-nav/list/title.hbs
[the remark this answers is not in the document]
KristinLBradley: I'm making a list and will add this to it as a follow on.

packages/components/src/components/hds/app-side-nav/index.ts
KristinLBradley: I'm making a list and will add this to it as a follow on.

showcase/tests/acceptance/components/hds/app-side-nav-test.js
KristinLBradley: I will add this to the list of things to look at in a follow on. (Will investigate more at that time if there's any improvement we could do that would make sense.)

packages/components/src/components/hds/app-side-nav/index.hbs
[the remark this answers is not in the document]
KristinLBradley: Alright, I'm thinking to perhaps do this in the follow on. Would that be ok? I'm making a list of smaller optimizations/improvements like this which I could complete as a fast follow on.

packages/components/src/components/hds/app-side-nav/index.ts
```diff
@@ -0,0 +1,188 @@
+/**
```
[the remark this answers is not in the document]
KristinLBradley: Thanks! I ported everything over.

KristinLBradley:
@alex-ju I pushed up fixes for issues. For some items I would like to look at them as a follow on if that's ok. (Listed below.)

**AppSideNav Follow-ons:**

* Change div for list "title" to an html header tag such as h3
	https://github.com/hashicorp/design-system/pull/2384#discussion_r1764009486
* Improve ariaLabel defaults for AppSideNav toggle button state (open/closed)
	https://github.com/hashicorp/design-system/pull/2384#discussion_r1764017267
* Look to improve app-side-nav-test.js so it doesn't disable the whole page 		
	https://github.com/hashicorp/design-system/pull/2384#discussion_r1764023460
* Audit usage of :header & :footer and look at removing them from the AppSideNav
	https://github.com/hashicorp/design-system/pull/2384#discussion_r1764121069

KristinLBradley:
There's a test failing I will investigate...

packages/components/src/styles/components/app-side-nav/content.scss
```diff
@@ -0,0 +1,177 @@
...
+  }
+}
+
+
+// LIST ITEM > INNER ELEMENTS
+
+.hds-app-side-nav__list-item-text {
+  min-width: 0;
+  max-width: 100%;
+  color: var(--token-app-side-nav-color-foreground-primary);
```
jorytindall, line 165: Now we know why the question between Section Titles and List Items was looking different in code and design, because it was. So now after looking more closely I do think a change needs to be made, I'm not sure exactly how you want to do it so I won't suggest, but I think the color of the ListItems should be changed to `Foreground / Strong` just to give us a bit of extra breathing room.

jorytindall (approved):
I think this is good to go, short of one note about the list item color that should probably be a quick fix. Let me know what you think or any other questions!

_KB: Updated the color, thanks for reviewing!_

packages/components/src/styles/components/app-side-nav/content.scss
KristinLBradley: Updated, thanks

shleewhite (commented):
Issue (blocking): when testing [the full page example](https://hds-showcase-git-hds-3800-app-sidenav-component-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-app-header-and-app-side-nav) I can still navigate to the hidden content when the nav is collapsed.

packages/components/src/components/hds/app-side-nav/index.hbs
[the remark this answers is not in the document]
alex-ju: this is the one bit from the open conversations that defines the component API. technically, moving _away from named blocks_ or _to named blocks_ would both mean a breaking change. I'm ok if we want to do it 'behind the scenes' (assuming consumers will not adopt the component before we decide on the API) but I want it to be a conscious decision and not an afterthought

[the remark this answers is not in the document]
KristinLBradley: It's looking like everything that was in the :header and the :footer should go in the AppHeader so I'll just remove those named blocks from the AppSideNav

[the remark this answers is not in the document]
KristinLBradley: @alex-ju I went ahead and removed the :header and :footer blocks, updated related tests, & the Showcase. I'm wondering now if the :body block still makes sense. There is a :root and a :body which the original SideNav had though: https://helios.hashicorp.design/components/side-nav?tab=code#sidenavbase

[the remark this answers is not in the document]
alex-ju: indeed. considering the current implementation, I'd be tempted to only have the default block (no named blocks) to keep things simple. and if we find additional needs we can decide how we want to go about them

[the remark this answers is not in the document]
KristinLBradley: @alex-ju I ended up reverting removal of the named blocks because it was just becoming over-complicated to refactor all the underlying code. I think it should be done as a follow on if we do it so that it can be reviewed properly in more isolation from all the other pre-existing code.

packages/components/src/components/hds/app-side-nav/index.ts
```diff
@@ -0,0 +1,195 @@
...
+    this.desktopMQ = window.matchMedia(`(min-width:${this.desktopMQVal})`);
+    this.addEventListeners();
+    registerDestructor(this, (): void => {
+      this.removeEventListeners();
+    });
+  }
+
+  addEventListeners(): void {
+    document.addEventListener('keydown', this.escapePress, true);
+    this.desktopMQ.addEventListener('change', this.updateDesktopVariable, true);
```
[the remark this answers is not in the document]
KristinLBradley: The current SideNav included Dropdowns in the :header and :footer blocks which consumers could choose to add. They functioned as expected and didn't interfere with the SideNav,

showcase/tests/acceptance/components/hds/app-side-nav-test.js
alex-ju: We haven't really nailed the axe config to narrow down exceptions. We can track this in a separate task

packages/components/src/components/hds/app-side-nav/list/index.hbs
```diff
@@ -0,0 +1,19 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<nav class="hds-app-side-nav__list-wrapper" ...attributes>
+  {{yield (hash ExtraBefore=(component "hds/yield"))}}
+  <ul class="hds-app-side-nav__list" role="list">
```
MelSumner, line 8: Is there a way for the title to be added to the list as the value for an `aria-labelledby` attribute? 

Otherwise it will continue to show up in rotors as "list" instead of having a meaningful name. It will also be indistinguishable from other lists on the page.

packages/components/src/components/hds/app-side-nav/portal/target.ts
MelSumner, line 71: Please add to the comment that this is ascii art. See https://www.w3.org/TR/WCAG20-TECHS/H86.html

packages/components/src/components/hds/app-side-nav/toggle-button.hbs
```diff
@@ -0,0 +1,7 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<button class="hds-app-side-nav__toggle-button" type="button" ...attributes>
```
MelSumner, line 5: Issue, blocker: this seems to be missing an accessible name.

MelSumner:
https://github.com/user-attachments/assets/7f0ebf12-837b-44a3-b3ed-4cacbd4cc388

This video is demonstrating that when browser zoom is used, the page in the background continues to scroll when the sidebar is open. The page underneath should not respond when the sidebar is open and overlapping the underlying content.

packages/components/src/components/hds/app-side-nav/list/index.hbs
MelSumner: https://github.com/user-attachments/assets/cd34a7a5-855d-4482-b0dd-684c92fbf098

Here's a video to show what I mean.

packages/components/src/components/hds/app-side-nav/toggle-button.hbs
MelSumner: Additional issue, blocker: when I have VoiceOver open and I toggle the menu open or closed, there is no `aria-expanded` attribute telling me if it's open or closed.

MelSumner:
https://github.com/user-attachments/assets/202788d6-f9de-4134-9c6c-ba27d2c7c020

Issue, blocker: 
The attached video demonstrates that I can still keyboard navigate to the sidebar items even when the sidebar is closed.

packages/components/src/components/hds/app-side-nav/list/title.hbs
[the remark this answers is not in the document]
KristinLBradley: I agree this needs to be done but I would like to make these kind of updates as a follow on since what it has now is what is currently already implemented in the SideNav.

Doing it as a follow on would make it easier for others to see what's been changed and to review since this current PR takes the majority of its code directly from the current SideNav code (which is a lot of code).

packages/components/src/components/hds/app-side-nav/list/index.hbs
KristinLBradley: I added it to the list. I would like to make updates like this as a follow on if that's ok.

packages/components/src/components/hds/app-side-nav/toggle-button.hbs
KristinLBradley: I will check into these, thanks.

KristinLBradley:
> Issue (blocking): when testing [the full page example](https://hds-showcase-git-hds-3800-app-sidenav-component-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-app-header-and-app-side-nav) I can still navigate to the hidden content when the nav is collapsed.

Fixed. It's dependent on a class name and I still had the old classname.

packages/components/src/components/hds/app-side-nav/toggle-button.hbs
KristinLBradley: In index.ts there is code which adds an ariaLabel to the toggleButton:

```
get toggleButtonAriaLabel(): string {
    if (this.isMinimized) {
      return this.args.toggleButtonAriaLabel ?? 'Open menu';
    }
    return this.args.toggleButtonAriaLabel ?? 'Close menu';
  }
```

I had already added an item to the list of follow-on tasks (in PR description) to improve this (partially by adding aria-expanded as mentioned in linked comment from @shleewhite ). It looks like you implemented the above in the original SideNav code btw.

KristinLBradley:
> CleanShot.2024-09-25.at.13.39.35.mp4 
> This video is demonstrating that when browser zoom is used, the page in the background continues to scroll when the sidebar is open. The page underneath should not respond when the sidebar is open and overlapping the underlying content.

I'm seeing this issue with the old SideNav as well. Would it be ok to look at a fix in a fast followon?

alex-ju:
> > CleanShot.2024-09-25.at.13.39.35.mp4
> > This video is demonstrating that when browser zoom is used, the page in the background continues to scroll when the sidebar is open. The page underneath should not respond when the sidebar is open and overlapping the underlying content.
> 
> I'm seeing this issue with the old SideNav as well. Would it be ok to look at a fix in a fast followon?

yes, we seem to have this issue in the SideNav as well – we could potentially use the same solution we have in Modal/Flyout (this is aligning more and more with the Flyout anyway)

showcase/tests/integration/components/hds/app-side-nav/index-test.js
```diff
@@ -0,0 +1,350 @@
...
+  });
+
+  test('it should render the component with a CSS class that matches the component name', async function (assert) {
+    await render(
+      hbs`<Hds::AppSideNav id="test-app-side-nav" @hasA11yRefocus={{false}} />`
+    );
+    assert.dom('#test-app-side-nav').hasClass('hds-app-side-nav');
+  });
+
+  // CONTENT
```
shleewhite, line 56: Thought: I know we haven't really done this with other components, but I think it would be beneficial to add a gut check test to ensure keyboard focus is working correctly. Like when the sidenav is collapsed, you can't tab to the contents.

alex-ju:
This branch is in a good state for its scope (porting the SideNav + visual adjustments). Kristin's suggestion to continue the work by addressing the open threads (many of them improvements of the current `SideNav`) as the follow-up PRs makes sense to me.

packages/components/src/components/hds/app-side-nav/list/link.ts
```diff
@@ -0,0 +1,30 @@
...
+
+import TemplateOnlyComponent from '@ember/component/template-only';
+
+import type { HdsIconSignature } from '../../icon';
+import type { HdsInteractiveSignature } from '../../interactive';
+
+export interface HdsAppSideNavListLinkSignature {
+  Args: HdsInteractiveSignature['Args'] & {
+    icon?: HdsIconSignature['Args']['name'];
+    iconColor?: string;
```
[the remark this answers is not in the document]
KristinLBradley: Checked with @jorytindall and I'm going to go ahead and just remove this option as the final design ended up not including color icons anyway.

showcase/tests/integration/components/hds/app-side-nav/index-test.js
KristinLBradley: On line 183 there's already a test for this:
```
assert.dom('.hds-app-side-nav-hide-when-minimized').hasAttribute('inert');
```

The framed example wasn't preventing tabbing as it should previously (before I fixed it) because I had missed updating the class names in the file (The classes have to be added manually to Lists in the AppSideNav content as in the SideNav).

packages/components/src/components/hds/app-side-nav/list/title.hbs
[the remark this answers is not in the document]
KristinLBradley: Started a new branch for "follow-ons". Please take a look and let me know any feedback

- Branch: https://github.com/hashicorp/design-system/pull/2463
- Commit: https://github.com/hashicorp/design-system/pull/2463/commits/dbe1568f4afebae23ac02e55ad80ee4f6913d6a8

(I'll do the smaller updates like the above just in this one branch and if there's anything more involved I may create other branches...)

packages/components/src/components/hds/app-side-nav/list/index.hbs
KristinLBradley: I looked at this and I don't see a way to really do it with the current structure. I believe it would requiring completely restructuring the code for the "Title" in relation to the List and List Items.

I worked on the SideNav in the very early beginning stages and had structured the title to be outside the list instead of inside it but was overruled. The structure and functionality of the current SideNav (and now the AppSideNav which is based off it) is very complex so I'm wary of making too many changes to the structure if it can be avoided.

packages/components/src/components/hds/app-side-nav/index.ts
KristinLBradley: - PR Branch: https://github.com/hashicorp/design-system/pull/2463/commits/43e57208bec1381d33b4ecb98438b9e63e567ae4
- Commit: https://github.com/hashicorp/design-system/pull/2463/commits/43e57208bec1381d33b4ecb98438b9e63e567ae4

packages/components/src/components/hds/app-side-nav/toggle-button.hbs
KristinLBradley: - PR Branch: https://github.com/hashicorp/design-system/pull/2463/commits/43e57208bec1381d33b4ecb98438b9e63e567ae4
- Commit: https://github.com/hashicorp/design-system/pull/2463/commits/43e57208bec1381d33b4ecb98438b9e63e567ae4

packages/components/src/components/hds/app-side-nav/index.hbs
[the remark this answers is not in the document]
KristinLBradley: Update: This is being worked on in a separate branch

https://github.com/hashicorp/design-system/pull/2477

alex-ju (approved):
Individual changes were reviewed in separate PRs. I would suggest rebasing this PR to main to resolve the one conflict highlighted in the website docs around the deprecation badge.

website/docs/components/app-side-nav/partials/version-history/4.10.0.md
```diff
@@ -0,0 +1,5 @@
+## 4.10.0
```
alex-ju, line 1: filename as well

```suggestion
## 4.14.0
```

website/docs/components/app-side-nav/index.md
```diff
@@ -0,0 +1,38 @@
...
+<section data-tab="Specifications">
+  @include "partials/specifications/anatomy.md"
+</section>
+
+<section data-tab="Accessibility">
+  @include "partials/accessibility/accessibility.md"
+</section>
+
+<section data-tab="Version history">
+  @include "partials/version-history/4.10.0.md"
```
alex-ju, line 36:
```suggestion
  @include "partials/version-history/4.14.0.md"
```

website/docs/components/app-side-nav/partials/version-history/4.10.0.md
KristinLBradley: Updated

packages/components/src/components/hds/app-side-nav/index.ts
```diff
@@ -0,0 +1,210 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { registerDestructor } from '@ember/destroyable';
+
+interface HdsAppSideNavSignature {
```
KristinLBradley, line 11: Should I export the signature? @alex-ju

alex-ju: Yes, please, that will allow be to merge #2499 now

zamoore (approved):
Really clean code and good tests. I know this has been a journey, but I think this is very good work.

packages/components/src/components/hds/app-side-nav/index.ts
KristinLBradley: Done

KristinLBradley:
Set this PR back to draft to prevent accidental merge as planned adoption of Enterprise Nav related components was put off.

.changeset/silver-planes-occur.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Hds::AppSideNav` - Added the component. It is meant to replace the standalone `SideNav` component and be paired together with the `AppHeader` which should both be contained within the `AppFrame` layout component.
```
alex-ju, line 5: This changelog entry should be removed

packages/components/src/styles/@hashicorp/design-system-components.scss
```diff
@@ -17,6 +17,7 @@
 @use "../components/app-footer";
 @use "../components/app-frame";
 @use "../components/app-header";
+@use "../components/app-side-nav";
```
alex-ju, line 20: it would be a good idea to comment these out; or not include them

showcase/tests/integration/components/hds/app-side-nav/portal/index-test.js
```diff
@@ -0,0 +1,151 @@
...
+import { setupRenderingTest } from 'ember-qunit';
+// import { render, waitUntil } from '@ember/test-helpers';
+// import { hbs } from 'ember-cli-htmlbars';
+
+// NOTICE: we combine the tests for `PortalTarget` and `Portal` in a single file, because the two components are strictly interconnected
+
+module('Integration | Component | hds/app-side-nav/portal', function (hooks) {
+  setupRenderingTest(hooks);
+
+  // DISABLEtest('it should render the component with a CSS class that matches the component name', async function (assert) {
```
alex-ju, line 16: recommend `skip` here as well

showcase/tests/integration/components/hds/app-side-nav/index-test.js
```diff
@@ -0,0 +1,296 @@
...
+      await settled();
+    };
+  });
+
+  hooks.afterEach(function () {
+    resetOnerror();
+    window.matchMedia = this.__matchMedia;
+  });
+
+  // DISABLEtest('it should render the component with a CSS class that matches the component name', async function (assert) {
```
alex-ju, line 49: recommend `skip`

alex-ju (commented):
left a few suggestions. considering the age of this branch, I suggest we rebase before we get this in. let me know if you'd like me to handle this on your behalf

showcase/tests/integration/components/hds/app-side-nav/portal/index-test.js
KristinLBradley: Updated

alex-ju (approved):

website/docs/components/app-side-nav/index.md
```diff
@@ -0,0 +1,38 @@
...
+links:
+  figma: 
+  github: https://github.com/hashicorp/design-system/tree/main/packages/components/src/components/hds/app-side-nav
+related: ['components/side-nav', 'components/breadcrumb','components/tabs','layouts/app-frame','components/app-header']
+previewImage: assets/illustrations/components/app-side-nav.jpg
+navigation:
+  hidden: true
+  keywords: ['navigation', 'side navigation', 'sidenav', 'sidebar']
+status:
+  added: 4.14.0
```
alex-ju, line 14: we should remove this badge until we actually publish the component. same for the version history

website/docs/components/app-side-nav/index.md
```diff
@@ -0,0 +1,38 @@
...
+  @include "partials/specifications/anatomy.md"
+</section>
+
+<section data-tab="Accessibility">
+  @include "partials/accessibility/accessibility.md"
+</section>
+
+<section data-tab="Version history">
+  @include "partials/version-history/4.14.0.md"
+</section>
```
alex-ju, line 35:
```suggestion
```

KristinLBradley: Oops missed that. Thanks!

PR #2387, opened 2024-09-04
Table - Sort by selected

zamoore:
### :pushpin: Summary

If merged, this PR implements new "sort-by-selected" behavior in the `Hds::Table` component.

### ⚙️ Details

There is a new argument that need to be provided to the component to enable this behavior:
- `@sortBySelectedItemKey`: `string`
  - Flag to enable SBS behavior

  - Sets which `@model` property is used to store the record's selection state

Screenshot 2024-09-06 at 1 30 51 PM

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-3592

### DIFF
--- a/.changeset/large-spiders-protect.md
+++ b/.changeset/large-spiders-protect.md
@@ -1,0 +1,6 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Hds::Table`
+- Added `@sortBySelectedItemKey` argument which enables sorting by row selection state
--- a/.changeset/large-spiders-protect.md
+++ b/.changeset/large-spiders-protect.md
@@ -12,4 +12,4 @@
 
 `Hds::Table::ThSelectable`
 - Added `@onClickSortBySelected` argument which is the callback for the sort button in the selected item column.
-- Added `@sortOrder` (`undefined` | `asc` | `desc`) argument which determines the state of the sort button in the selected item column.
+- Added `@sortBySelectedOrder` argument which determines the state of the sort button in the selected item column.
--- a/.changeset/large-spiders-protect.md
+++ b/.changeset/large-spiders-protect.md
@@ -3,4 +3,4 @@
 ---
 
 `Hds::Table`
-- Added `@sortBySelectedItemKey` argument which enables sorting by row selection state
+- Added `@selectableColumnKey` argument which enables sorting by row selection state
--- a/.changeset/large-spiders-protect.md
+++ b/.changeset/large-spiders-protect.md
@@ -3,4 +3,13 @@
 ---
 
 `Hds::Table`
-- Added `@selectableColumnKey` argument which enables sorting by row selection state
+- Added `@selectableColumnKey` argument which enables sorting by row selection state and specifies the corresponding selection state key.
+
+`Hds::Table::Tr`
+- Added `@selectableColumnKey` argument which enables sorting by row selection state and specifies the corresponding selection state key.
+- Added `@sortBySelectedOrder` (`undefined` | `asc` | `desc`) argument which determines the state of the sort button in the selected item column.
+- Added `@onClickSortBySelected` argument which is the callback for the sort button in the selected item column.
+
+`Hds::Table::ThSelectable`
+- Added `@onClickSortBySelected` argument which is the callback for the sort button in the selected item column.
+- Added `@sortOrder` (`undefined` | `asc` | `desc`) argument which determines the state of the sort button in the selected item column.
--- a/.changeset/large-spiders-protect.md
+++ b/.changeset/large-spiders-protect.md
@@ -7,7 +7,7 @@
 
 `Hds::Table::Tr`
 - Added `@selectableColumnKey` argument which enables sorting by row selection state and specifies the corresponding selection state key.
-- Added `@sortBySelectedOrder` (`undefined` | `asc` | `desc`) argument which determines the state of the sort button in the selected item column.
+- Added `@sortBySelectedOrder` argument which determines the state of the sort button in the selected item column.
 - Added `@onClickSortBySelected` argument which is the callback for the sort button in the selected item column.
 
 `Hds::Table::ThSelectable`
--- a/packages/components/src/components/hds/table/index.hbs
+++ b/packages/components/src/components/hds/table/index.hbs
@@ -48,10 +48,8 @@
         (hash
           Tr=(component
             "hds/table/tr"
-            sortOrder=@sortOrder
             selectionScope="col"
             isSelectable=@isSelectable
-            onClickSortBySelected=(if @selectableColumnKey (fn this.setSortBy @selectableColumnKey))
             onSelectionChange=this.onSelectionAllChange
             didInsert=this.didInsertSelectAllCheckbox
             willDestroy=this.willDestroySelectAllCheckbox
--- a/packages/components/src/components/hds/table/index.hbs
+++ b/packages/components/src/components/hds/table/index.hbs
@@ -14,6 +14,7 @@
     {{#if @columns}}
       <Hds::Table::Tr
         @sortOrder={{if (eq this.sortBy @selectedItemKey) this.sortOrder}}
+        @onClickSort={{fn this.setSortBy @selectedItemKey}}
         @selectionScope="col"
         @isSelectable={{@isSelectable}}
         @onSelectionChange={{this.onSelectionAllChange}}
@@ -79,7 +80,6 @@
               "hds/table/tr"
               selectionScope="row"
               isSelectable=@isSelectable
-              onClickSort=(fn this.setSortBy @selectedItemKey)
               onSelectionChange=this.onSelectionRowChange
               didInsert=this.didInsertRowCheckbox
               willDestroy=this.willDestroyRowCheckbox
--- a/packages/components/src/components/hds/table/index.hbs
+++ b/packages/components/src/components/hds/table/index.hbs
@@ -13,7 +13,6 @@
   <thead class="hds-table__thead">
     {{#if @columns}}
       <Hds::Table::Tr
-        @sortBySelectedItemKey={{@sortBySelectedItemKey}}
         @sortOrder={{if (eq this.sortBy @sortBySelectedItemKey) this.sortOrder}}
         @selectionScope="col"
         @onClickSortBySelected={{if @sortBySelectedItemKey (fn this.setSortBy @sortBySelectedItemKey)}}
@@ -49,7 +48,6 @@
         (hash
           Tr=(component
             "hds/table/tr"
-            sortBySelectedItemKey=@sortBySelectedItemKey
             sortOrder=@sortOrder
             selectionScope="col"
             isSelectable=@isSelectable
--- a/packages/components/src/components/hds/table/index.hbs
+++ b/packages/components/src/components/hds/table/index.hbs
@@ -13,9 +13,9 @@
   <thead class="hds-table__thead">
     {{#if @columns}}
       <Hds::Table::Tr
-        @canSortBySelectedItemKey={{@canSortBySelectedItemKey}}
-        @sortOrder={{if (eq this.sortBy @selectedItemKey) this.sortOrder}}
-        @onClickSort={{fn this.setSortBy @selectedItemKey}}
+        @sortBySelectedItemKey={{@sortBySelectedItemKey}}
+        @sortOrder={{if (eq this.sortBy @sortBySelectedItemKey) this.sortOrder}}
+        @onClickSort={{fn this.setSortBy @sortBySelectedItemKey}}
         @selectionScope="col"
         @isSelectable={{@isSelectable}}
         @onSelectionChange={{this.onSelectionAllChange}}
--- a/packages/components/src/components/hds/table/index.hbs
+++ b/packages/components/src/components/hds/table/index.hbs
@@ -13,9 +13,9 @@
   <thead class="hds-table__thead">
     {{#if @columns}}
       <Hds::Table::Tr
-        @sortBySelectedOrder={{if (eq this.sortBy @selectableColumnKey) this.sortOrder}}
         @selectionScope="col"
         @onClickSortBySelected={{if @selectableColumnKey (fn this.setSortBy @selectableColumnKey)}}
+        @sortBySelectedOrder={{if (eq this.sortBy @selectableColumnKey) this.sortOrder}}
         @isSelectable={{@isSelectable}}
         @onSelectionChange={{this.onSelectionAllChange}}
         @didInsert={{this.didInsertSelectAllCheckbox}}
--- a/packages/components/src/components/hds/table/index.hbs
+++ b/packages/components/src/components/hds/table/index.hbs
@@ -13,9 +13,9 @@
   <thead class="hds-table__thead">
     {{#if @columns}}
       <Hds::Table::Tr
-        @sortOrder={{if (eq this.sortBy @sortBySelectedItemKey) this.sortOrder}}
+        @sortOrder={{if (eq this.sortBy @selectableColumnKey) this.sortOrder}}
         @selectionScope="col"
-        @onClickSortBySelected={{if @sortBySelectedItemKey (fn this.setSortBy @sortBySelectedItemKey)}}
+        @onClickSortBySelected={{if @selectableColumnKey (fn this.setSortBy @selectableColumnKey)}}
         @isSelectable={{@isSelectable}}
         @onSelectionChange={{this.onSelectionAllChange}}
         @didInsert={{this.didInsertSelectAllCheckbox}}
@@ -51,7 +51,7 @@
             sortOrder=@sortOrder
             selectionScope="col"
             isSelectable=@isSelectable
-            onClickSortBySelected=(if @sortBySelectedItemKey (fn this.setSortBy @sortBySelectedItemKey))
+            onClickSortBySelected=(if @selectableColumnKey (fn this.setSortBy @selectableColumnKey))
             onSelectionChange=this.onSelectionAllChange
             didInsert=this.didInsertSelectAllCheckbox
             willDestroy=this.willDestroySelectAllCheckbox
--- a/packages/components/src/components/hds/table/index.hbs
+++ b/packages/components/src/components/hds/table/index.hbs
@@ -13,6 +13,7 @@
   <thead class="hds-table__thead">
     {{#if @columns}}
       <Hds::Table::Tr
+        @sortOrder={{if (eq this.sortBy @selectedItemKey) this.sortOrder}}
         @selectionScope="col"
         @isSelectable={{@isSelectable}}
         @onSelectionChange={{this.onSelectionAllChange}}
@@ -78,6 +79,7 @@
               "hds/table/tr"
               selectionScope="row"
               isSelectable=@isSelectable
+              onClickSort=(fn this.setSortBy @selectedItemKey)
               onSelectionChange=this.onSelectionRowChange
               didInsert=this.didInsertRowCheckbox
               willDestroy=this.willDestroyRowCheckbox
--- a/packages/components/src/components/hds/table/index.hbs
+++ b/packages/components/src/components/hds/table/index.hbs
@@ -54,6 +54,8 @@
             didInsert=this.didInsertSelectAllCheckbox
             willDestroy=this.willDestroySelectAllCheckbox
             selectionAriaLabelSuffix="all rows"
+            onClickSortBySelected=(if @selectableColumnKey (fn this.setSortBy @selectableColumnKey))
+            sortBySelectedOrder=(if (eq this.sortBy @selectableColumnKey) this.sortOrder)
           )
           Th=(component "hds/table/th")
           ThSort=(component "hds/table/th-sort")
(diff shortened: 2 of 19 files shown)


packages/components/src/components/hds/table/th-selectable.hbs
```diff
@@ -4,13 +4,18 @@
 }}
 
 <Hds::Table::Th class="hds-table__th--is-selectable" @scope={{@selectionScope}} ...attributes>
```
shleewhite, line 6: Issue: the th is missing aria-sort='ascending/descending' if the column is sorted by the selected items. you can see it for the other sortable columns here: https://github.com/hashicorp/design-system/blob/a5e990eda421353d1f20f9c546c99fa8e5547780/packages/components/src/components/hds/table/th-sort.hbs#L7

packages/components/src/components/hds/table/th-selectable.hbs
```diff
@@ -4,13 +4,18 @@
...
+      id={{this.checkboxId}}
+      class="hds-table__checkbox"
+      checked={{@isSelected}}
+      aria-label={{this.ariaLabel}}
+      {{did-insert this.didInsert}}
+      {{will-destroy this.willDestroyNode}}
+      {{on "change" this.onSelectionChange}}
+    />
+    {{#if @showSortButton}}
+      <Hds::Table::ThButtonSort @sortOrder={{@sortOrder}} @onClick={{@onClickSort}} />
```
shleewhite, line 18: Issue: something is missing because the label for aria-labelledby is only saying "sort by ascending". i think its because there is no visible label in the cell.  it should say "sort by selected ascending".

https://github.com/hashicorp/design-system/blob/a5e990eda421353d1f20f9c546c99fa8e5547780/packages/components/src/components/hds/table/th-button-sort.hbs#L9

^^ here you can see the labelledby is the prefix label id + the label id + the suffix id.

shleewhite (commented):
Did an accessibility review - just a couple issues.

didoo (commented):
@zamoore this is the first part of the review, so you have something you can start to think/work on; in the meantime I'll continue testing the implementation itself (the overall direction I _think_ is OK, it's what I would have imagined, but there are a few things that don't click - yet - for me).

packages/components/src/components/hds/table/index.hbs
```diff
@@ -13,6 +13,9 @@
   <thead class="hds-table__thead">
     {{#if @columns}}
       <Hds::Table::Tr
+        @sortBySelectedItemKey={{@sortBySelectedItemKey}}
```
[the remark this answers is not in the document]
didoo: @zamoore ignore my last comment: `isSelected` as model's **key** is actually set by the consumer (now I see where, I was confused by the two similar examples in the showcase demo).

packages/components/src/components/hds/table/index.hbs
```diff
@@ -13,6 +13,9 @@
...
+        @sortOrder={{if (eq this.sortBy @sortBySelectedItemKey) this.sortOrder}}
```
didoo, line 17: [note] we may be able to simplify this logic/arguments drilling, but I'll give it another look once the other things has been fixed/simplified so there's less code surface to check.

didoo (commented):
@zamoore I did a second pass, in particular re. some of the arguments drilling/forwarding. I decided to wait for some of the fixes to be merged, so the surface area to touch/review is reduced.

let me know when is ready for a new review

packages/components/src/components/hds/table/index.hbs
```diff
@@ -13,6 +13,9 @@
...
+        @onClickSort={{fn this.setSortBy @sortBySelectedItemKey}}
```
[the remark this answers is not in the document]
zamoore: Ah yes, I could see how that could be confusing. Good suggestion.

packages/components/src/components/hds/table/index.ts
```diff
@@ -190,7 +192,11 @@ export default class HdsTable extends Component<HdsTableArgs> {
   }
 
   @action
-  setSortBy(column: string): void {
+  setSortBy(column?: string): void {
+    if (column === undefined) {
```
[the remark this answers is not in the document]
zamoore: I think you're right, and I can't recall the specific reason this change was made, so it was likely part of some intermediary state of the code. Thanks for spotting it.

packages/components/src/components/hds/table/th-button-tooltip.ts
```diff
@@ -16,11 +16,6 @@ export interface HdsTableThButtonTooltipArgs {
 }
 
 export default class HdsTableThButtonTooltip extends Component<HdsTableThButtonTooltipArgs> {
-  /**
-   * Generates a unique ID for the (hidden) "label prefix" <span> element
```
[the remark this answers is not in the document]
zamoore: Makes sense to me

packages/components/src/components/hds/table/index.hbs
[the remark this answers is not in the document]
zamoore: Nice. Sounds good. I agree that we should add tests for that case though. Let me see what I can do.

packages/components/src/components/hds/table/th-selectable.ts
```diff
@@ -18,13 +19,16 @@ export interface HdsTableThSelectableArgs {
       selectionKey?: string
     ) => void;
     isSelected?: boolean;
+    onClickSort: HdsTableTrArgs['Args']['onClickSort'];
```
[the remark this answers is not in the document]
zamoore: We need to pass an argument to the function though, and the signature for ThButtonSort onClick does not accept any arguments.

showcase/app/templates/components/table.hbs
```diff
@@ -142,6 +141,72 @@
...
+  </Hds::Table>
+
+  <Shw::Text::H4 @tag="h3">Sortable table with no model, sorting by selected item</Shw::Text::H4>
+  <Shw::Text::Body>⚠️
+    <em>Notice: in this example the "sort" button doesn't work, is for presentation purpose only.</em></Shw::Text::Body>
+
+  <Hds::Table
+    @caption="a sortable table with no model defined that can sort by selection state"
+    @isSelectable={{true}}
+    @onSelectionChange={{this.onSelectionChangeSortBySelectedWithoutModel__demo6}}
```
didoo, line 181: [todo] since this example is "for presentation only" we don't need this custom "sorting" function (you could use `@onSelectionChange={{this.onSelectionChangeLogArguments}}` like on other places; if instead the intent is to validate that it's possible to sort by selected row even when there's no `@model/@columns` then the example should be moved under the "Custom sorting" section, but in that case the logic/code should be functional

didoo (changes requested):
left a few comments (some are blockers)

showcase/app/templates/components/table.hbs
zamoore: I'm not certain we need to validate that it's possible as almost all of the logic for implementing it will be outside of the component. I think we've already validated elsewhere that the sortProperty will be passed to the custom onSort function, but I can add a functional example with custom sort by selected behavior if you think it will add value.

showcase/tests/integration/components/hds/table/tr-test.js
```diff
@@ -77,6 +77,24 @@ module('Integration | Component | hds/table/tr', function (hooks) {
     assert.strictEqual(key, 'row123');
   });
 
+  test('it should render a sort button in the checkbox cell if `@isSelectable` is `true` and `@sortBySelectedItemKey` is provided', async function (assert) {
```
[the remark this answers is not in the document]
zamoore: I think I probably pushed the fix for this right as you were reviewing. Saw it this morning and was hoping to push up before reviews. Sorry about that, but all set now.

packages/components/src/components/hds/table/th-selectable.hbs
```diff
@@ -3,14 +3,28 @@
...
+    <Hds::Form::Checkbox::Base
+      id={{this.checkboxId}}
+      class="hds-table__checkbox"
+      checked={{@isSelected}}
+      aria-label={{this.ariaLabel}}
+      {{did-insert this.didInsert}}
+      {{will-destroy this.willDestroyNode}}
+      {{on "change" this.onSelectionChange}}
+    />
+    {{#if @onClickSort}}
```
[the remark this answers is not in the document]
zamoore: Not a bad idea. I set up a getter that makes this less awkward.

didoo (commented):
left another comment

showcase/app/templates/components/table.hbs
didoo: no, I don't think we need to test that use case (the logic is client-side, and as you say above we know the sortProperty is passed to the custom onSort function).

packages/components/src/components/hds/table/th-selectable.hbs
```diff
@@ -3,14 +3,29 @@
...
+      id={{this.checkboxId}}
+      class="hds-table__checkbox"
+      checked={{@isSelected}}
+      aria-label={{this.ariaLabel}}
+      {{did-insert this.didInsert}}
+      {{will-destroy this.willDestroyNode}}
+      {{on "change" this.onSelectionChange}}
+    />
+    {{#if this.isSortable}}
+      <span id={{this.labelId}} class="sr-only">selection state</span>
```
zamoore, line 23: Should we make this the dynamic sort key? Probably not meaningful to the end user.

packages/components/src/components/hds/table/index.hbs
```diff
@@ -13,7 +13,10 @@
   <thead class="hds-table__thead">
     {{#if @columns}}
       <Hds::Table::Tr
+        @sortBySelectedItemKey={{@sortBySelectedItemKey}}
```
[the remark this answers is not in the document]
zamoore: I agree and I'm fine with selectableColumnKey. Changed.

packages/components/src/components/hds/table/th-selectable.hbs
didoo: for consistency with the other similar implementation, probably leaving the `labelId` makes sense

showcase/app/controllers/components/table.js
```diff
@@ -438,4 +466,9 @@ export default class ComponentsTableController extends Controller {
       user.isAnimated = false;
     });
   }
+
+  @action
+  noop() {
```
aklkv, line 471: might be a good idea to extract into simple helper

packages/components/src/components/hds/table/th-selectable.hbs
zamoore: @didoo You mean like this?

`<span id={{this.labelId}} class="sr-only">{{this.labelId}}</span>`

didoo (commented):
I'll leave this comment for now, and look at the tests

packages/components/src/components/hds/table/index.hbs
```diff
@@ -13,7 +13,9 @@
   <thead class="hds-table__thead">
     {{#if @columns}}
       <Hds::Table::Tr
+        @sortOrder={{if (eq this.sortBy @selectableColumnKey) this.sortOrder}}
```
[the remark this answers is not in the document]
zamoore: It is the `sortOrder` of the table component.

It's the same way we pass the value to the `ThSort` component like

```
<Hds::Table::ThSort
              @sortOrder={{if (eq column.key this.sortBy) this.sortOrder}}
              @onClickSort={{fn this.setSortBy column.key}}
              @align={{column.align}}
              @width={{column.width}}
              @tooltip={{column.tooltip}}
            >```


It will either be `undefined`, `asc`, or `desc`. When it's `undefined`, that column is unsorted, and when it's defined, it's sorted by the specified `sortOrder`

**Review by didoo — COMMENTED** [review]
@zamoore OK, I've reviewed everything and, apart from the previous comments, I've nothing else to add; it's been quite a journey, but we're at the end now 😅 so when they're addressed/resolved I am more than happy to approve (great job 🙌)

I would suggest to have someone from the "core" squad to approve too though

File: packages/components/src/components/hds/table/index.hbs (hunk 15 above)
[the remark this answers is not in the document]
Comment (zamoore) [reply packages/components/src/components/hds/table/index.hbs:16]: Nevermind, I see what you mean. sortOrder is misleading because it will be undefined when sorting by another value, which works but is confusing and incorrect

File: packages/components/src/components/hds/table/th-sort.ts [hunk 16]
```diff
@@ -26,7 +27,7 @@ export const DEFAULT_ALIGN = HdsTableHorizontalAlignmentValues.Left;
 export interface HdsTableThSortArgs {
   Args: {
     align?: HdsTableHorizontalAlignment;
-    onClickSort?: () => void;
+    onClickSort?: HdsTableThButtonSortArgs['Args']['onClick'];
```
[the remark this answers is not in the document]
zamoore: Thanks!

packages/components/src/components/hds/table/tr.ts
```diff
@@ -25,6 +28,7 @@ export interface BaseHdsTableTrArgs {
       selectionKey?: string
     ) => void;
     willDestroy: () => void;
+    onClickSortBySelected?: () => void;
```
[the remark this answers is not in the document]
zamoore: That makes sense and seems to work correctly

packages/components/src/components/hds/table/th-selectable.ts
```diff
@@ -49,6 +62,22 @@ export default class HdsTableThSelectable extends Component<HdsTableThSelectable
...
+
+    switch (this.args.sortOrder) {
+      case HdsTableThSortOrderValues.Asc:
+        return HdsTableThSortOrderLabelValues.Asc;
+      case HdsTableThSortOrderValues.Desc:
+        return HdsTableThSortOrderLabelValues.Desc;
+      default:
+        // none is the default per the spec.
+        return HdsTableThSortOrderLabelValues.None;
+    }
```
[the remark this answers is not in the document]
zamoore: Ah, that makes sense. I like it. Thanks.

packages/components/src/components/hds/table/th-selectable.ts
```diff
@@ -18,13 +26,15 @@ export interface HdsTableThSelectableArgs {
       selectionKey?: string
     ) => void;
     isSelected?: boolean;
+    onClickSort?: () => void;
```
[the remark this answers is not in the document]
zamoore: Yeah, that makes sense.

.changeset/large-spiders-protect.md
```diff
@@ -0,0 +1,15 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Hds::Table`
+- Added `@selectableColumnKey` argument which enables sorting by row selection state and specifies the corresponding selection state key.
+
+`Hds::Table::Tr`
+- Added `@selectableColumnKey` argument which enables sorting by row selection state and specifies the corresponding selection state key.
+- Added `@sortBySelectedOrder` (`undefined` | `asc` | `desc`) argument which determines the state of the sort button in the selected item column.
```
didoo, line 10: [suggestion] we usually don't list the values of an enum in the release notes, unless necessary
```suggestion
- Added `@sortBySelectedOrder` argument which determines the state of the sort button in the selected item column.
```

.changeset/large-spiders-protect.md
```diff
@@ -0,0 +1,15 @@
...
+- Added `@onClickSortBySelected` argument which is the callback for the sort button in the selected item column.
+
+`Hds::Table::ThSelectable`
+- Added `@onClickSortBySelected` argument which is the callback for the sort button in the selected item column.
+- Added `@sortOrder` (`undefined` | `asc` | `desc`) argument which determines the state of the sort button in the selected item column.
```
didoo, line 15: [question] should `@sortOrder` become `sortBySelectedOrder` also for the `ThSelectable`, at this point?

showcase/app/controllers/components/table.js
zamoore: I agree. We have a lot of repetition throughout our tests/examples that can probably be condensed down into a reusable toolkit. I think that's out of scope for this PR, but I'm going to bring it up in our next engineering sync for discussion. Thanks for the input!

didoo (approved):
Left a couple of minor comments, but apart from that is GTG for me. 🚀

majedelass (approved):
From a purely design side, this is already following the design patterns established previously (8px gap between buttons and text), using the right convention with the arrow up/down for letting the user know it's sortable, colored when selected, etc.

👍 Nicely done

packages/components/src/components/hds/table/index.hbs
```diff
@@ -13,7 +13,9 @@
   <thead class="hds-table__thead">
     {{#if @columns}}
       <Hds::Table::Tr
+        @sortBySelectedOrder={{if (eq this.sortBy @selectableColumnKey) this.sortOrder}}
         @selectionScope="col"
+        @onClickSortBySelected={{if @selectableColumnKey (fn this.setSortBy @selectableColumnKey)}}
```
alex-ju, line 18: nit: should we align these with what we have further down the file in the tr section?
```suggestion
        @selectionScope="col"
        @onClickSortBySelected={{if @selectableColumnKey (fn this.setSortBy @selectableColumnKey)}}
        @sortBySelectedOrder={{if (eq this.sortBy @selectableColumnKey) this.sortOrder}}
```

packages/components/src/components/hds/table/index.ts
```diff
@@ -50,11 +51,12 @@ export interface HdsTableArgs {
     isFixedLayout?: boolean;
     isSelectable?: boolean;
     isStriped?: boolean;
-    model: Array<Record<string, unknown>>;
+    model: HdsTableModel;
```
alex-ju, line 54: suggestion: to ease the rebase with #2421 (marking the model as optional)
```suggestion
    model?: HdsTableModel;
```

packages/components/src/components/hds/table/th-selectable.hbs
```diff
@@ -3,14 +3,29 @@
...
-    {{will-destroy this.willDestroyNode}}
-    {{on "change" this.onSelectionChange}}
-  />
+<Hds::Table::Th
+  class="hds-table__th--is-selectable"
+  aria-sort={{if this.isSortable this.ariaSort}}
+  @scope={{@selectionScope}}
+  ...attributes
+>
+  <div class="hds-table__th-content">
```
alex-ju, line 12: praise: nice re-use of existing classes and conventions!

showcase/app/controllers/components/table.js
```diff
@@ -67,8 +68,14 @@ export default class ComponentsTableController extends Controller {
   @tracked multiSelectUsersCurrentPage_demo3 = 1;
   @tracked multiSelectUsersCurrentPageSize_demo3 = 4;
   @deepTracked multiSelectUserData__demo4 = [...this.model.userDataDemo4];
-
-  debugger;
```
alex-ju, line 71: 😬

.changeset/large-spiders-protect.md
```diff
@@ -0,0 +1,15 @@
...
+- Added `@selectableColumnKey` argument which enables sorting by row selection state and specifies the corresponding selection state key.
+
+`Hds::Table::Tr`
+- Added `@selectableColumnKey` argument which enables sorting by row selection state and specifies the corresponding selection state key.
+- Added `@sortBySelectedOrder` argument which determines the state of the sort button in the selected item column.
+- Added `@onClickSortBySelected` argument which is the callback for the sort button in the selected item column.
+
+`Hds::Table::ThSelectable`
+- Added `@onClickSortBySelected` argument which is the callback for the sort button in the selected item column.
+- Added `@sortBySelectedOrder` argument which determines the state of the sort button in the selected item column.
```
alex-ju, line 15: thought: some of these are for internal use; I don't think we plan on documenting them all (but I might be wrong) so I suggest we add to the changelog only the arguments we will expose in the component API

alex-ju (approved):
Sorry for the delayed review – it took me a while, but it eventually clicked 😄 .
Couldn't spot anything off in my testing, so I think it's a go.

.changeset/large-spiders-protect.md
zamoore: That was what I had originally, but @didoo suggested that I document all of the additions to subcomponents as well. Curious to hear his thoughts

showcase/app/controllers/components/table.js
zamoore: 😬

packages/components/src/components/hds/table/th-selectable.hbs
zamoore: Thanks!

packages/components/src/components/hds/table/index.ts
zamoore: Agreed

.changeset/large-spiders-protect.md
didoo: In terms of changelog, we all agree the changes should be mentioned. 
For the website, personally I would document them anyway (and in case add a note clarifying that it's meant for internal use). That said, I totally get that the Table API documentation is already heavy so if you decide differently I would not oppose (but consider that, at times, we use the component API documentation too, to understand what a certain argument does).

PR #2403, opened 2024-09-10
`SideNav`: remove `Ember.testing` because it will be deprecated

shleewhite:
### :pushpin: Summary

If merged, this PR would remove usage of `Ember.testing` because it will be deprecated and replace it with [the suggested replacement](https://github.com/emberjs/rfcs/blob/3e032300e02f1222a0d1a0788b3e1936fc33a143/text/1003-deprecation-import-ember-from-ember.md?plain=1#L207).

### :link: External links

Jira ticket: [HDS-3813](https://hashicorp.atlassian.net/browse/HDS-3813)

***

### 👀 Component checklist

[HDS-3813]: https://hashicorp.atlassian.net/browse/HDS-3813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/side-nav/portal/target.ts
+++ b/packages/components/src/components/hds/side-nav/portal/target.ts
@@ -41,11 +41,7 @@ export default class HdsSideNavPortalTarget extends Component<HdsSideNavPortalTa
   @tracked lastPanelEl: Element | undefined;
 
   static get prefersReducedMotionOverride(): boolean {
-    if (macroCondition(isTesting())) {
-      return true;
-    }
-
-    return false;
+    return macroCondition(isTesting()) ? true : false;
   }
 
   prefersReducedMotionMQ = window.matchMedia(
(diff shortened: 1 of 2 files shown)


packages/components/src/components/hds/side-nav/portal/target.ts
```diff
@@ -41,7 +41,11 @@ export default class HdsSideNavPortalTarget extends Component<HdsSideNavPortalTa
   @tracked lastPanelEl: Element | undefined;
 
   static get prefersReducedMotionOverride(): boolean {
-    return Ember.testing;
+    if (macroCondition(isTesting())) {
```
shleewhite, line 44: I tried just doing `return macroCondition(isTesting())` and the build failed 🤷🏽

didoo: what about this? does it fail as well?
```
const isTesting = macroCondition(isTesting());
return isTesting;
```

(also, what does `macroCondition(isTesting())` return?)

alex-ju (approved):
Looks in line with the guidance. Odd choice to make it predicative, but they may have their reasons 🤷

PR #2407, opened 2024-09-11
Showcase - Normalize naming of variables/types in `Shw` components

didoo:
### :pushpin: Summary

Another small PR (on top of #2406) to normalize how we call the variables and types in the `Shw` components.

Inspired by the normalization we did when we migrated the HDS components to TypeScript.
The intent is to bring these changes later to the HDS+ repository too (so we can keep things consistent)

### 👀 Component checklist


KristinLBradley (approved):

didoo:
> I like this change! Is Ember smart enough to know that the renamed components are the same? Like we don't need to change the name where the components are used?

@shleewhite my understanding is that Ember uses the file name/path to determine the component name, so everything should remain the same.

shleewhite (approved):

PR #2409, opened 2024-09-11
chore: remove unneeded flight icons ci action

shleewhite:
### :pushpin: Summary

If merged, this PR would remove the ember-flight-icons ci step because we don't plan on releasing the package anymore and it only runs lint.

Also removes reference to it in the README.


Dhaulagiri (approved):
🫡

PR #2415, opened 2024-09-12
Website: Change out the "updated" badges and add 4.10.0 version history

shleewhite:
### :pushpin: Summary

If merged, this PR would:
* clear out the updated status metadata from before 4.9.0
* add updated status metadata for components changed in 4.10.0
* add version history to components changed in 4.10.0
* update the component tabs acceptance tests to not fail if the "Alert" link in the sidebar has an "updated" tag.

### DIFF
--- a/website/docs/components/app-header/index.md
+++ b/website/docs/components/app-header/index.md
@@ -9,8 +9,6 @@ related: ['components/side-nav', 'components/app-footer', 'layouts/app-frame']
 previewImage: assets/illustrations/components/app-header.jpg
 navigation:
   keywords: ['navigation', 'header', 'navbar', 'menubar', 'topbar']
-status:
-  added: 4.8.0
 ---
 
 <section data-tab="Guidelines">
--- a/website/docs/components/app-header/index.md
+++ b/website/docs/components/app-header/index.md
@@ -9,6 +9,8 @@ related: ['components/side-nav', 'components/app-footer', 'layouts/app-frame']
 previewImage: assets/illustrations/components/app-header.jpg
 navigation:
   keywords: ['navigation', 'header', 'navbar', 'menubar', 'topbar']
+status:
+  updated: 4.10.0
 ---
 
 <section data-tab="Guidelines">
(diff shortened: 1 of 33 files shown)


website/docs/components/icon/index.md
```diff
@@ -9,8 +9,6 @@ related: ['icons/library']
 previewImage: assets/illustrations/components/icon.jpg
 navigation:
   keywords: ['icon', 'glyph', 'flight']
-status:
-  added: 4.9.0
```
alex-ju, line 13: thought: I appreciate the diligence here, and I'm aware we don't have any guidance around when a status should be updated, but it feels a bit too early to remove this badge, considering we're still in the adoption/migration phase for this new component.

website/docs/components/app-header/index.md
```diff
@@ -32,5 +32,6 @@ status:
 </section>
 
 <section data-tab="Version history">
+  @include "partials/version-history/4.10.0.md"
   @include "partials/version-history/4.8.0.md"
```
alex-ju, line 36: thought: not sure we want to maintain these component-level version docs in perpetuity 🤔  @didoo any thoughts on this?

alex-ju (approved):
Thanks for taking the time to get this up to date! Left two overarching questions about how we manage tags and component-level docs.

website/docs/components/app-header/index.md
didoo: My though when implementing this feature was that yes, we could maintain them

- it's an easy reference to see how a component evolved, and honestly they don't add much weight to the page itself (and if a user don't click on the tab, they don't see the content anyway)
- the alternative would be the release notes on the website, but 1) they are truncated anyway and 2) it's harder to find content specific for a component

website/docs/components/icon/index.md
shleewhite: That's fair, I will add it back. 

I just noticed that because so many components changed in this release there were a lot of badges everywhere.

PR #2419, opened 2024-09-13
Cleanup color tokens

jorytindall:
### :pushpin: Summary

As discovered in the Figma v2.0 Library project:

- Updates the icon colors in the `Copy::Button` to match Figma and the interactive states of the component
- Updates the `Stepper::Indicator::Step` to use a surface token instead of a palette token.
- Fixes a bug in the `secondary` `Copy::Snippet` where the icon had it's own interactive state, now preventing the color of the icon from changing with the component status = `success` or `error`. This matches the experience for the `primary` variant.

No visual change in the Stepper, but there is a visual change in the CopyButton:

**Before**
Screenshot 2024-09-13 at 12 21 08 PM

**After**
Screenshot 2024-09-13 at 12 20 23 PM

### :link: External links

Jira ticket: [HDS-3845](https://hashicorp.atlassian.net/browse/HDS-3845), [HDS-3845](https://hashicorp.atlassian.net/browse/HDS-3845)
Figma file: [CopyButton](https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/%5BWIP%5D-HDS-Components-v2.0?node-id=39372-123324&t=rq4Nnx1RFT1CKnGV-1)

***

### 👀 Component checklist

[HDS-3845]: https://hashicorp.atlassian.net/browse/HDS-3845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/styles/components/copy/snippet.scss
+++ b/packages/components/src/styles/components/copy/snippet.scss
@@ -58,10 +58,10 @@
     background-color: var(--token-color-surface-interactive);
     border-color: var(--token-color-border-strong);
 
-    // Change the color of the icon in hover state, except if the component has a status of success or error
     .hds-copy-snippet__icon:not(
-        .hds-copy-snippet--status-success .hds-copy-snippet__icon
-      ):not(.hds-copy-snippet--status-error .hds-copy-snippet__icon) {
+        .hds-copy-snippet--status-success .hds-copy-snippet__icon,
+        .hds-copy-snippet--status-error .hds-copy-snippet__icon
+      ) {
       color: var(--token-color-foreground-action-hover);
     }
   }
@@ -73,8 +73,9 @@
 
     // Change the color of the icon in active status, except if the component has a status of success or error
     .hds-copy-snippet__icon:not(
-        .hds-copy-snippet--status-success .hds-copy-snippet__icon
-      ):not(.hds-copy-snippet--status-error .hds-copy-snippet__icon) {
+        .hds-copy-snippet--status-success .hds-copy-snippet__icon,
+        .hds-copy-snippet--status-error .hds-copy-snippet__icon
+      ) {
       color: var(--token-color-foreground-action-active);
     }
   }
--- a/packages/components/src/styles/components/copy/snippet.scss
+++ b/packages/components/src/styles/components/copy/snippet.scss
@@ -49,31 +49,33 @@
   color: var(--token-color-foreground-primary);
   background-color: transparent;
 
+  .hds-copy-snippet__icon {
+    color: var(--token-color-foreground-action);
+  }
+
   &:hover,
   &.mock-hover {
     background-color: var(--token-color-surface-interactive);
     border-color: var(--token-color-border-strong);
+
+    // Change the color of the icon in hover state, except if the component has a status of success or error
+    .hds-copy-snippet__icon:not(
+        .hds-copy-snippet--status-success .hds-copy-snippet__icon
+      ):not(.hds-copy-snippet--status-error .hds-copy-snippet__icon) {
+      color: var(--token-color-foreground-action-hover);
+    }
   }
 
   &:active,
   &.mock-active {
     background-color: var(--token-color-surface-interactive-active);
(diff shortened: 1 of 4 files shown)


packages/components/src/styles/components/copy/snippet.scss
```diff
@@ -49,32 +49,34 @@
...
+    color: var(--token-color-foreground-action);
+  }
+
   &:hover,
   &.mock-hover {
     background-color: var(--token-color-surface-interactive);
     border-color: var(--token-color-border-strong);
+
+    // Change the color of the icon in hover state, except if the component has a status of success or error
+    .hds-copy-snippet__icon:not(
```
jorytindall, line 62: I'm not sure if this is the cleanest solution, but it does prevent some repetition in declaring the hover states.

jorytindall: Throws an error in the linter, so I'm thinking this is not the best approach.

didoo: not necessarily. I think you have a linter different from what is used in the repo. I'll have a look on monday

jorytindall: I reworked this a bit passed a comma-separated list of classes rather than chaining multiple `:not()` pseudo classes together which seems to satisfy the linter.

didoo: [suggestion] instead of doing all these `not()` overrides, I would follow the same approach used for the `Copy::Button` which is to increase the specificity for the "success/error" state classes (and in this case, since it would be used twice, I would extract it to a local mixin):

```
@mixin hds-copy-snippet-success-error() {
  &.hds-copy-snippet--status-success {
    background-color: var(--token-color-surface-interactive);

    .hds-copy-snippet__icon {
      color: var(--token-color-foreground-success);
    }
  }

  &.hds-copy-snippet--status-error {
    background-color: var(--token-color-surface-interactive);

    .hds-copy-snippet__icon {
      color: var(--token-color-foreground-critical);
    }
  }
}

.hds-copy-snippet {
  @include hds-focus-ring-with-pseudo-element();

  display: flex;
  gap: 8px;
  align-items: center;
  padding: 6px 4px;
  text-align: left;
  border: 1px solid transparent;
  border-radius: 5px;
  cursor: pointer;
}

.hds-copy-snippet--color-primary {
  color: var(--token-color-foreground-action);
  background-color: transparent;

  &:hover,
  &.mock-hover {
    color: var(--token-color-foreground-action-hover);
    background-color: var(--token-color-surface-interactive);
    border-color: var(--token-color-border-strong);
  }

  &:active,
  &.mock-active {
    color: var(--token-color-foreground-action-active);
    background-color: var(--token-color-surface-interactive-active);
    border-color: var(--token-color-border-strong);
  }

  @include hds-copy-snippet-success-error();
}

.hds-copy-snippet--color-secondary {
  background-color: transparent;

  .hds-copy-snippet__icon {
    color: var(--token-color-foreground-action);
  }

  &:hover,
  &.mock-hover {
    background-color: var(--token-color-surface-interactive);
    border-color: var(--token-color-border-strong);

    .hds-copy-snippet__icon {
      color: var(--token-color-foreground-action-hover);
    }
  }

  &:active,
  &.mock-active {
    background-color: var(--token-color-surface-interactive-active);
    border-color: var(--token-color-border-strong);

    .hds-copy-snippet__icon {
      color: var(--token-color-foreground-action-active);
    }
  }

  @include hds-copy-snippet-success-error();
}

```

didoo (approved):
Left a suggestion. Technically not a blocker so I am approving the PR 👍

packages/components/src/styles/components/copy/snippet.scss
jorytindall: This is a great suggestion, much cleaner!

LilithJames-HDS (approved):

PR #2421, opened 2024-09-16
Remove glint nocheck from remaining templates and fix revealed type issues

alex-ju:
### :pushpin: Summary

This PR will help [unblock HCP from upgrading to the latest HDS](https://hashicorp.slack.com/archives/C06NMSYDZPA/p1726276284759049) (tested this branch in `cloud-ui` locally)

### 🛠️ Detailed description

- Removed `@glint-nocheck` from a series of templates in `showcase` (these would've been ideally removed at conversion) and fixed a few type issues this change reveals
 - Fixed optional arguments in component signatures and mark `@icon` as required in `AppHeader::HomeLink` (we seem to have quite a bit of misalignment between what we mark as required in our docs and what's required via TypeScript, so I tried aligning the two, preventing many potential blockers in adoption where our signatures were stricter than needed)
 - Fixed absolute paths to signatures (we have a few instances where because of absolute paths, some files were not resolved by the ember CLI at build time)

### :link: External links

Jira ticket: [HDS-3847](https://hashicorp.atlassian.net/browse/HDS-3847)

***

### 👀 Component checklist

[HDS-3847]: https://hashicorp.atlassian.net/browse/HDS-3847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/form/super-select/multiple/field.ts
+++ b/packages/components/src/components/hds/form/super-select/multiple/field.ts
@@ -6,14 +6,30 @@
 import Component from '@glimmer/component';
 import { ID_PREFIX } from '../../label/index.ts';
 
+import type { ComponentLike } from '@glint/template';
+import type { HdsFormErrorSignature } from '../../error/index.ts';
 import type { HdsFormFieldSignature } from '../../field/index.ts';
+import type { HdsFormHelperTextSignature } from '../../helper-text/index.ts';
+import type { HdsFormLabelSignature } from '../../label';
 import type { HdsFormSuperSelectMultipleBaseSignature } from './base.ts';
+import type { Select as PowerSelect } from 'ember-power-select/components/power-select';
+import type { HdsYieldSignature } from 'src/components/hds/yield/index.ts';
 
 interface HdsFormSuperSelectMultipleFieldSignature {
   Args: HdsFormSuperSelectMultipleBaseSignature['Args'] &
     HdsFormFieldSignature['Args'];
   Blocks: {
-    default: [unknown];
+    default: [
+      {
+        Label?: ComponentLike<HdsFormLabelSignature>;
+        HelperText?: ComponentLike<HdsFormHelperTextSignature>;
+        Error?: ComponentLike<HdsFormErrorSignature>;
+        Options?: ComponentLike<HdsYieldSignature>;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        options?: any;
+        select?: PowerSelect;
+      },
+    ];
   };
   Element: HdsFormFieldSignature['Element'];
 }
--- a/packages/components/src/components/hds/form/super-select/multiple/field.ts
+++ b/packages/components/src/components/hds/form/super-select/multiple/field.ts
@@ -13,7 +13,7 @@ import type { HdsFormHelperTextSignature } from '../../helper-text/index.ts';
 import type { HdsFormLabelSignature } from '../../label';
 import type { HdsFormSuperSelectMultipleBaseSignature } from './base.ts';
 import type { Select as PowerSelect } from 'ember-power-select/components/power-select';
-import type { HdsYieldSignature } from 'src/components/hds/yield/index.ts';
+import type { HdsYieldSignature } from '../../../yield/index.ts';
 
 interface HdsFormSuperSelectMultipleFieldSignature {
   Args: HdsFormSuperSelectMultipleBaseSignature['Args'] &
--- a/packages/components/src/components/hds/form/super-select/multiple/field.ts
+++ b/packages/components/src/components/hds/form/super-select/multiple/field.ts
@@ -25,8 +25,7 @@ interface HdsFormSuperSelectMultipleFieldSignature {
         HelperText?: ComponentLike<HdsFormHelperTextSignature>;
         Error?: ComponentLike<HdsFormErrorSignature>;
         Options?: ComponentLike<HdsYieldSignature>;
-        // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        options?: any;
+        options?: unknown;
         select?: PowerSelect;
       },
     ];
--- a/showcase/app/templates/components/form/super-select/mock-components/selected-component-single.hbs
+++ b/showcase/app/templates/components/form/super-select/mock-components/selected-component-single.hbs
@@ -1,8 +1,8 @@
-{{! @glint-nocheck: not typesafe yet }}
 {{!
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
 
 {{! This is not an HDS component, but a supporting file for `form/super-select.hbs` which requires a component to be passed in for the showcase }}
+{{! @glint-expect-error - option is known to be an object in this mock component }}
 <Hds::Text::Body @tag="span" @size="200">{{@option.size}}</Hds::Text::Body>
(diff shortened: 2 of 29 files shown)


alex-ju:
We're currently getting [glint errors](https://github.com/hashicorp/design-system/actions/runs/10881593235/job/30190877787?pr=2421#step:7:13) related to the `<AF.LegalLinks />` line in both full-page demo apps in `showcase`

packages/components/src/components/hds/form/super-select/multiple/field.ts
```diff
@@ -6,14 +6,30 @@
...
   Blocks: {
-    default: [unknown];
+    default: [
+      {
+        Label?: ComponentLike<HdsFormLabelSignature>;
+        HelperText?: ComponentLike<HdsFormHelperTextSignature>;
+        Error?: ComponentLike<HdsFormErrorSignature>;
+        Options?: ComponentLike<HdsYieldSignature>;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        options?: any;
```
shleewhite, line 29: Could the options be unknown?

packages/components/src/components/hds/tag/index.ts
```diff
@@ -16,7 +16,7 @@ export const DEFAULT_COLOR = HdsTagColorValues.Primary;
 interface HdsTagSignature {
   Args: HdsInteractiveSignature['Args'] & {
     color?: HdsTagColors;
-    text: string;
+    text?: string;
```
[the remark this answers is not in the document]
alex-ju: Right, I see the assert on Badge and Tag (it was me who added it a while back 🤦 ), which means that we should switch to updating the documentation to flag it as required. However, I can't see an assert for the Breadcrumb item 🤔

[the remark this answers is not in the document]
shleewhite: oh I totally thought I was looking at badge, not breadcrumb 🤦 ignore that one

packages/components/src/components/hds/form/super-select/multiple/field.ts
alex-ju: It could, but I'm not convinced we want to enforce type-checking on our consumers. To me, it's a more explicit way of saying that they can use any format for the `options` object. If you feel strongly about it, I'm happy to go with `unknown` – I know it's preferred, but not convinced in this scenario.

packages/components/src/components/hds/tag/index.ts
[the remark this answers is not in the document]
alex-ju: Updated the docs for Badge, BadgeCount, Breadcrumb, and Tag

shleewhite:
> We're currently getting [glint errors](https://github.com/hashicorp/design-system/actions/runs/10881593235/job/30190877787?pr=2421#step:7:13) related to the `<AF.LegalLinks />` line in both full-page demo apps in `showcase`
> 
> Update: for now, I'm going to set a glint to ignore on these two templates and add an action item to investigate further

@alex-ju its super weird. I looked into it a bit and no matter what AF sub components you use it throws errors.  nothing is erroring on the component page. I have to think it is something in a ts config (even tho I couldn't find anything there) or because the component is in the `layouts` subfolder and not `components` because all the frameless stuff was added more recently.

packages/components/src/components/hds/pagination/size-selector/index.ts
```diff
@@ -12,7 +12,7 @@ import type { HdsFormSelectBaseSignature } from '../../form/select/base';
 
 interface HdsPaginationSizeSelectorSignature {
   Args: {
-    pageSizes: number[];
+    pageSizes?: number[];
```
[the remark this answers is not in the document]
alex-ju: I ended up reverting this as well – thanks for the check!

showcase/app/templates/components/form/super-select/mock-components/selected-component-single.hbs
```diff
@@ -1,8 +1,8 @@
-{{! @glint-nocheck: not typesafe yet }}
 {{!
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
 
 {{! This is not an HDS component, but a supporting file for `form/super-select.hbs` which requires a component to be passed in for the showcase }}
+{{! @glint-expect-error - option is known to be an object in this mock component }}
 <Hds::Text::Body @tag="span" @size="200">{{@option.size}}</Hds::Text::Body>
```
aklkv, line 8: breaking lines might offer better understanding where the issues is coming from
```suggestion
<Hds::Text::Body @tag="span" @size="200">
  {{! @glint-expect-error - option is known to be an object in this mock component }}
  {{@option.size}}
</Hds::Text::Body>
```

packages/components/src/components/hds/code-block/index.ts
```diff
@@ -50,7 +50,7 @@ export interface HdsCodeBlockSignature {
     isStandalone?: boolean;
     language?: HdsCodeBlockLanguages;
     maxHeight?: string;
-    value: string;
+    value?: string;
```
shleewhite, line 53: Issue: Another arg I think should be required.. in the code getter we check that value is not undefined.

packages/components/src/components/hds/form/super-select/multiple/field.ts
shleewhite: 🤷🏽 I think it is a decision we should make as a team. are most people using typescript? if so, it doesn't seem unreasonable to make them type check the options.

packages/components/src/components/hds/code-block/index.ts
alex-ju: cool, I will update the docs and mark it as required then

alex-ju: updated

packages/components/src/components/hds/form/super-select/multiple/field.ts
alex-ju: Agree, let's bring it to the eng sync. However, I'm afraid I need a 'yes' sooner to unblock HCP, so I f9241750ca60b76f08ede92725e5be6d8cf81ef7 in the hope of an approval

showcase/app/templates/layouts/app-frame/frameless/demo-full-app-frame-with-side-nav.hbs
```diff
@@ -3,12 +3,11 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-{{! @glint-nocheck: not typesafe yet }}
 {{page-title "AppFrame Component - Frameless"}}
 
 <Hds::AppFrame @hasHeader={{false}} as |Frame|>
   <Frame.Sidebar>
-    <Hds::SideNav @isResponsive={{true}} @hasA11yRefocus={{true}} @isCollapsible={{true}} @hasHeader={{false}}>
+    <Hds::SideNav @isResponsive={{true}} @hasA11yRefocus={{true}} @isCollapsible={{true}}>
```
KristinLBradley, line 10: Thanks for finding & removing this. I overlooked it before.

packages/components/src/components/hds/form/super-select/multiple/field.ts
alex-ju: Agree, let's bring it to the eng sync. However, I'm afraid I need a 'yes' sooner to unblock HCP, so I f9241750ca60b76f08ede92725e5be6d8cf81ef7 in the hope of an approval

shleewhite (approved):

alex-ju:
Rebased to resolve conflicts after #2387

PR #2427, opened 2024-09-17
feat: update eslint-plugin-ember to latest major version

shleewhite:
### :pushpin: Summary

If merged, this PR would upgrade eslint-plugin-ember to the latest version and includes some fixes to get the lint to pass:
- add specific overrides for `gts` files in showcase and components 
- add ignores to showcase, website, and components to make the linting checks pass

### :link: External links

Jira ticket: [HDS-3805](https://hashicorp.atlassian.net/browse/HDS-3805)

***

### 👀 Component checklist

[HDS-3805]: https://hashicorp.atlassian.net/browse/HDS-3805?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/loud-grapes-refuse.md
+++ b/.changeset/loud-grapes-refuse.md
@@ -1,0 +1,7 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Upgraded `eslint-plugin-ember` to `12.2.0`
+
+`AppFooter`, `AppFrame`, `SideNav` - refactored subcomponents to use `TemplateOnlyComponent` instead of empty classes.
--- a/packages/components/.eslintrc.cjs
+++ b/packages/components/.eslintrc.cjs
@@ -36,7 +36,8 @@ module.exports = {
         'plugin:@typescript-eslint/recommended',
       ],
       rules: {
-        // Add any custom rules here
+        // ember doesn't support template only components that have type signatures so we need to use empty component classes
+        'ember/no-empty-glimmer-component-classes': 'off',
       },
     },
     // require relative imports use full extensions
--- a/packages/components/.eslintrc.cjs
+++ b/packages/components/.eslintrc.cjs
@@ -36,8 +36,7 @@ module.exports = {
         'plugin:@typescript-eslint/recommended',
       ],
       rules: {
-        // ember doesn't support template only components that have type signatures so we need to use empty component classes
-        'ember/no-empty-glimmer-component-classes': 'off',
+        // Add any custom rules here
       },
     },
     // require relative imports use full extensions
--- a/packages/components/src/components/hds/app-footer/index.ts
+++ b/packages/components/src/components/hds/app-footer/index.ts
@@ -7,10 +7,10 @@ import Component from '@glimmer/component';
 
 import type { ComponentLike } from '@glint/template';
 import type { HdsYieldSignature } from '../yield';
-import type HdsAppFooterItemSignature from './legal-links';
-import type HdsAppFooterLegalLinksSignature from './legal-links';
-import type HdsAppFooterLinkSignature from './link';
-import type HdsAppFooterStatusLinkSignature from './status-link';
+import type { HdsAppFooterItemSignature } from './item.ts';
+import type { HdsAppFooterLegalLinksSignature } from './legal-links';
+import type { HdsAppFooterLinkSignature } from './link';
+import type { HdsAppFooterStatusLinkSignature } from './status-link';
 import type { HdsAppFooterThemeTypes } from './types.ts';
 
 export interface HdsAppFooterSignature {
--- a/packages/components/src/components/hds/code-block/index.ts
+++ b/packages/components/src/components/hds/code-block/index.ts
@@ -140,6 +140,7 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
     const grammar = language ? Prism.languages[language] : undefined;
 
     if (code) {
+      // eslint-disable-next-line ember/no-runloop
       next(() => {
         if (language && grammar) {
           this.prismCode = htmlSafe(Prism.highlight(code, grammar, language));
@@ -158,6 +159,7 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
         // Context: https://github.com/hashicorp/design-system/pull/1749#discussion_r1374288785
         if (this.args.highlightLines) {
           // we need to delay re-evaluating the context for prism-line-highlight for as much as possible, and `afterRender` is the 'latest' we can use in the component lifecycle
+          // eslint-disable-next-line ember/no-runloop
           schedule('afterRender', () => {
             // we piggy-back on the plugin's `resize` event listener to trigger a new call of the `highlightLines` function: https://github.com/PrismJS/prism/blob/master/plugins/line-highlight/prism-line-highlight.js#L337
             if (window) window.dispatchEvent(new Event('resize'));
--- a/packages/components/src/components/hds/table/index.ts
+++ b/packages/components/src/components/hds/table/index.ts
@@ -87,8 +87,7 @@ export interface HdsTableSignature {
 }
 
 export default class HdsTable extends Component<HdsTableSignature> {
-  // eslint-disable-next-line ember/no-tracked-properties-from-args
-  @tracked sortBy = this.args.sortBy;
+  @tracked sortBy = this.args.sortBy ?? undefined;
   @tracked sortOrder = this.args.sortOrder || HdsTableThSortOrderValues.Asc;
   @tracked selectAllCheckbox?: HdsFormCheckboxBaseSignature['Element'] =
     undefined;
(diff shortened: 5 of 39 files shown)


packages/components/src/components/hds/table/index.ts
```diff
@@ -85,6 +85,7 @@ export interface HdsTableArgs {
 }
 
 export default class HdsTable extends Component<HdsTableArgs> {
+  // eslint-disable-next-line ember/no-tracked-properties-from-args
   @tracked sortBy = this.args.sortBy;
```
aklkv, line 89: I wonder what is the goal of such property?

aklkv (approved):
Looks good  👍🏻 One rule I would investigate `ember/no-tracked-properties-from-args` and maybe potentially fix

shleewhite:
Approved Percy because the changes found were just that there are new table sort by selected examples.

didoo:
> Approved Percy because the changes found were just that there are new table sort by selected examples.

@shleewhite for future reference: I _think_ that rebasing on `main` would have automatically fixed the Percy tests

.changeset/loud-grapes-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Upgrade eslint-plugin-ember to ^12.2.0
```
alex-ju, line 5: nit: we aim to use past tense for changelogs
```suggestion
Upgraded `eslint-plugin-ember` to `12.2.0`
```

packages/components/.eslintrc.cjs
```diff
@@ -36,7 +36,8 @@ module.exports = {
         'plugin:@typescript-eslint/recommended',
       ],
       rules: {
-        // Add any custom rules here
+        // ember doesn't support template only components that have type signatures so we need to use empty component classes
+        'ember/no-empty-glimmer-component-classes': 'off',
```
alex-ju, line 40: blocker: we don't have to turn this off in `packages/component`; we should fix this by updating the (8?) instances where we have empty backing classes with `TemplateOnlyComponent`s

packages/components/src/components/hds/code-block/index.ts
```diff
@@ -140,6 +140,7 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
     const grammar = language ? Prism.languages[language] : undefined;
 
     if (code) {
+      // eslint-disable-next-line ember/no-runloop
```
alex-ju, line 143: question: should we aim to migrate these to `ember-lifeline` or another recommended solution?

packages/components/unpublished-development-types/global.d.ts
```diff
@@ -9,6 +9,7 @@ import type EmberStargateRegistry from 'ember-stargate/template-registry';
 import type EmberStyleModifierRegistry from 'ember-style-modifier/template-registry';
 import type EmberTruthRegistry from 'ember-truth-helpers/template-registry';
 import type PowerSelectRegistry from 'ember-power-select/template-registry';
+// eslint-disable-next-line ember/no-at-ember-render-modifiers
```
alex-ju, line 12: question: what's the alternative to `ember-render-modifiers`?

alex-ju (changes requested):
`eslint-plugin-ember` v12 brings a fair amount of pivotal changes in how we should write glimmer component. while it is ok to skip some rules while we align our components with the new listing rules I want to make sure we understand these and we plan follow-up work accordingly.

one specific rule that I want us to address as part of this upgrade is `empty-glimmer-component-classes` – template-only Glimmer components without a backing class are much faster and lighter-weight than those with a backing classes

packages/components/unpublished-development-types/global.d.ts
aklkv: we have not found suitable solution but here a resource that talks in more details about this https://nullvoxpopuli.com/avoiding-lifecycle/

packages/components/src/components/hds/code-block/index.ts
shleewhite: Yeah - we should definitely migrate off `ember-runloop` and `ember-render-modifiers`. I can add tickets to the backlog since those changes could be more involved...

packages/components/.eslintrc.cjs
shleewhite: This is needed for the `gts` conversion. There is currently no way for the gts components to have types and only a template. The showcase text components are examples of this ([Shw::Text::Body](https://github.com/hashicorp/design-system/blob/main/showcase/app/components/shw/text/body.gts)).

I can make it more specific so the rule is only turned off for `.gts` files and not `.ts` files too.

packages/components/src/components/hds/code-block/index.ts
didoo: > Yeah - we should definitely migrate off `ember-runloop` and `ember-render-modifiers`. I can add tickets to the backlog since those changes could be more involved...

@shleewhite @zamoore @alex-ju my 2 cents: before going all-in with such change, I would recommend doing a thorough analysis of the impact/effort (with associated RFC): it may not be as easy/straightforward as it may seem (from what I remember) plus it may require breaking changes in the components' APIs.

packages/components/unpublished-development-types/global.d.ts
shleewhite: I was doing some research as well and it seems pretty complicated to migrate off `ember-render-modifiers`.

packages/components/src/components/hds/code-block/index.ts
shleewhite: 100% agree

packages/components/.eslintrc.cjs
alex-ju: ok, I understand this may be a must for showcase (a link to the issue, if possible, would be of help so that we keep track of this), but is it required for the components as well? (as we don't have `.gts` components in our library)

packages/components/src/components/hds/code-block/index.ts
alex-ju: yes, to be clear, I was not suggesting we make these changes in this PR.

I find it paradoxical that the ember community suggests these are not recommended approaches when there are few alternatives (if any) out there

packages/components/.eslintrc.cjs
alex-ju: from where I stand, this rule found quite a few instances in `components` where we've not been very thorough in how we converted our components to TypeScript, so I'd be keen to correct them

packages/components/src/components/hds/code-block/index.ts
shleewhite: Tickets: 
- https://hashicorp.atlassian.net/browse/HDS-3884
- https://hashicorp.atlassian.net/browse/HDS-3885

packages/components/.eslintrc.cjs
shleewhite: I corrected the failures in the components and. removed the override!

showcase/app/templates/layouts/app-frame/frameless/demo-full-app-frame-with-app-header.hbs
```diff
@@ -256,7 +256,6 @@
 
   <Frame.Footer>
     <Hds::AppFooter as |AF|>
-      {{! @glint-expect-error - TODO: investigate why this line fails }}
```
shleewhite, line 259: @alex-ju haha fixing the empty class components to templateOnly components revealed an issue with how the AppFooter component imported the types, which I fixed and then I think that fixed the weird typing issue with these files! Who woulda thought.

packages/components/src/components/hds/app-footer/index.ts
```diff
@@ -7,10 +7,10 @@ import Component from '@glimmer/component';
...
 import type { ComponentLike } from '@glint/template';
 import type { HdsYieldSignature } from '../yield';
-import type HdsAppFooterItemSignature from './legal-links';
-import type HdsAppFooterLegalLinksSignature from './legal-links';
-import type HdsAppFooterLinkSignature from './link';
-import type HdsAppFooterStatusLinkSignature from './status-link';
+import type { HdsAppFooterItemSignature } from './item.ts';
+import type { HdsAppFooterLegalLinksSignature } from './legal-links';
+import type { HdsAppFooterLinkSignature } from './link';
+import type { HdsAppFooterStatusLinkSignature } from './status-link';
```
alex-ju, line 13: right!!

alex-ju (approved):
Thanks for going the extra mile here – it paid off!

.changeset/loud-grapes-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Upgraded `eslint-plugin-ember` to `12.2.0`
```
alex-ju, line 5:
```suggestion
Upgraded `eslint-plugin-ember` to `12.2.0`

`AppFooter`, `AppFrame`, `SideNav` - refactored to use `TemplateOnlyComponent`
```

PR #2433, opened 2024-09-19
fix(components): add `instance-initializers` to appReexports

aklkv:
### :pushpin: Summary

We need to add `instance-initializers` otherwise it will not be registered by consuming app there may be side effects if consumer uses `ember-flight-icons` still

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist


alex-ju (approved):
I've been running into this issue in #2439 as well – thanks for raising this PR!

PR #2434, opened 2024-09-20
Add more documentation for select all behavior

zamoore:
### :pushpin: Summary

If merged, this will provide a documented example of "Select all row" behavior with an external action.

TBD

TBD

### :link: External links

Jira ticket: [HDS-3699](https://hashicorp.atlassian.net/browse/HDS-3699)
Figma file: [if it applies]

***

### 👀 Component checklist

[HDS-3699]: https://hashicorp.atlassian.net/browse/HDS-3699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/docs/components/table/index.js
+++ b/website/docs/components/table/index.js
@@ -189,4 +189,29 @@ export default class Index extends Component {
       }
     });
   }
+
+  @action
+  selectAll(isSelected) {
+    this.demoSourceData.forEach((record) => {
+      record.isSelected = isSelected;
+    });
+  }
+
+  @action
+  selectAllExternally() {
+    this.selectAll(true);
+  }
+
+  @action
+  demoOnSelectionChangeWithPagination2({
+    selectionKey,
+    selectionCheckboxElement,
+    selectableRowsStates,
+  }) {
+    if (selectionKey === 'all') {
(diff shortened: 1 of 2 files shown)


didoo:
[suggestion] @zamoore my 2 cents tip: I would start to implement this in the showcase, because it's much simpler to follow the flow of the data/logic (and potentially isolate the example, while developing it) and debug if necessary; in this way we can make sure we're covering all the use cases, and later choose the ones that should be "ported" to the website documentation (which becomes much easier, at that point)

One further suggestion, if you don't mind: to validate different way of selecting rows, I would have four different external actions: "select all" + "deselect all" + "select even" + "select odd" (where even/odd is the ID of the row).

And last but not least tip: I would try to throw in the mix also the pagination, to make sure it works also in that case (not just on the "visible" rows).

Essentially, something similar to this example, but with a dropdown instead of the button:
screenshot_4185

zamoore:
Closing this spike as it has been deemed that external selection is not currently possible.

https://hashicorp.atlassian.net/browse/HDS-3699?focusedCommentId=545950

PR #2211, opened 2024-07-04
`DialogPrimitive` with `Modal` and `Flyout` (non-breaking)

didoo:
### :pushpin: Summary

This PR is build using https://github.com/hashicorp/design-system/pull/1918 as base, and modified in a way that the chances can be released as non-breaking for our consumers.

For context about why the previous implementation was put on hold, see: https://docs.google.com/document/d/1yixV5kfGzrao8LMT0YHRrUkdN24nxf1II0-lC6pd9Vk/edit

### :hammer_and_wrench: Detailed description

In this PR I have:
- ported the `DialogPrimitive` component from the previous implementation
- updated `DialogPrimitive::Header` and `DialogPrimitive::Overlay` to support contextual classes (used to pass `modal/flyout` specific class names, to avoid breaking changes)
- fixed an issue with `DialogPrimitive::Footer` showing a border when “empty” (no child nodes, apart from whitespaces)
- remove extra whitespace from the `DialogPrimitive` templates
- ported integration and acceptance tests for `DialogPrimitive` from previous implementation (also fixed a wrong name in another test)
- updated `Modal` and `Flyout` components to use `DialogPrimitive` sub-components
- updated `Modal` and `Flyout` styles to reflect new structure based on `DialogPrimitive` sub-components
- added deprecation notes to old `Modal` and `Flyout` sub-components
- ported showcase page for `DialogPrimitive` from previous implementation
- fixed an a11y issue in the showcase (“[Scrollable region must have keyboard access](https://dequeuniversity.com/rules/axe/4.8/scrollable-region-focusable)”)
- small tweaks to the showcase page for the “Modal” page

✅ Note: we tested the changes in Cloud UI https://github.com/hashicorp/cloud-ui/pull/10195 and all the tests are passing, and both the `Hds::Modal` and `Hds::Flyout` work as expected within the application.

📍Note: the conversion of the components to TS will be done in separate tickets.

### 👉👉 👉 **Showcase previews:**
- [DialogPrimitive](https://hds-showcase-git-dialog-primitive-non-breaking-hashicorp.vercel.app/utilities/dialog-primitive)
- [Flyout](https://hds-showcase-git-dialog-primitive-non-breaking-hashicorp.vercel.app/components/flyout)
- [Modal](https://hds-showcase-git-dialog-primitive-non-breaking-hashicorp.vercel.app/components/modal)

### 🪢 Deprecation impact
- Flyout subcomponents
[...]

PR #2442, opened 2024-09-23
`DialogPrimitive` - Fix issue with `box-sizing` inheritance

didoo:
### :pushpin: Summary

In https://github.com/hashicorp/design-system/pull/2211 I (specifically @didoo) have introduced a change to the `Modal/Flyout` components, to use a newly created `DialogPrimitive` that could be used as standalone component/primitive as well as sub-component 

Because of this, I decided to reset the default styles applied to the "root" element of the `DialogPrimitive`, which is a `<dialog>` HTML element, using a `all: unset` CSS rule, and then apply the desired styles at `Modal/Flyout` level (while sharing the common styles in the `DialogPrimitive` CSS styles, declaring them just after the `all: unset` reset:

```
:where(.hds-dialog-primitive__wrapper) {
  // -----------
  all: unset; // removes default dialog & inherited styles. Must be the first style to be applied
  // -----------
  display: flex;
  flex-direction: column;
  width: 100%;
  height: 100%;
  margin: 0;
  padding: 0;
  background: var(--token-color-surface-primary);
}
```

The change was described as "non-breaking"m b

Thanks to @aklkv raising the issue [in a Slack thread](https://hashicorp.slack.com/archives/C7KTUHNUS/p1726871393498119), we've come to realize that this actually introduced a regression in the UI of the Modal in some of our components: depending on where the CSS reset for the `box-sizing` property was declared in the order of imports, some of the UI elements assumed a `box-sizing: content-box` property (thanks to the `*, *::before, *::after { box-sizing: inherit; }` rule) and with in some cases the padding applied to them was added to the height, causing them to blow up (see button in the screenshot below):

image

We missed this detail, both in implementation as well as in review, because the HDS showcase the reset is declared _after_ the import of the HDS components styles, so the `box-sizing` applied to `<dialog>` is inherited from the `<html>` element, which is set to `border-box` in the reset:

```
@import "@hashicorp/design-system-components";
@import "@hashicorp/design-system-power-select-overrides";

// global declarations

@import "./tokens";
@import "./layout";
@import "./typography";
@import "./globals";  // ← the box-sizing CSS reset is here
```

while in Cloud UI is declared _before_ the import of the CSS 

```
@use "resets";  // ← the box-sizing CSS reset is here
@use "lists";

// Components and helpers provided by the design system
// Notice: the full path is declared in the `hcp/ember-cli-build.js` file, under `sassOptions.includePaths`
@use "helpers/typography.css";
@use "@hashicorp/design-system-components.css";
```

In this PR I have:
- removed `all: unset` declaration from `hds-dialog-primitive__wrapper` and replaced with explicit overrides
- updated version-history and tags in website documentation

_Notice: there was a small visual regression in Percy but only in a mobile viewport and only for a use case, and was likely a fluke._

Preview: https://hds-showcase-git-3869-fix-dialog-primitive-box-sizing-hashicorp.vercel.app/components/modal

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-3869

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/styles/components/dialog-primitive.scss
+++ b/packages/components/src/styles/components/dialog-primitive.scss
@@ -13,17 +13,17 @@ $hds-dialog-primitive-vertical-padding: 16px;
 
 // WRAPPER CONTAINER (<dialog>)
 
-// The special declaration below is used to allow the `<dialog>` to unset
-// all its predefined styles and set them at `Modal/Flyout` level
+// The special declaration below is used to override the `<dialog>` predefined styles
+// Note: some of these style will then be set (again) at `Modal/Flyout` level too
 //
 // Since the `:where()` selector has a specificity of `0` (see https://developer.mozilla.org/en-US/docs/Web/CSS/:where)
 // the styles declared below are applied to the `<dialog>` element, but are overwritten by any other style applied via classname
 //
 
 :where(.hds-dialog-primitive__wrapper) {
-  // -----------
-  all: unset; // removes default dialog & inherited styles. Must be the first style to be applied
-  // -----------
+  // override default `<dialog>` styles
+  position: initial;
+  inset: initial;
   display: flex;
   flex-direction: column;
   width: 100%;
@@ -31,6 +31,7 @@ $hds-dialog-primitive-vertical-padding: 16px;
   margin: 0;
   padding: 0;
   background: var(--token-color-surface-primary);
+  border: none;
 }
 
 
--- a/.changeset/hot-timers-complain.md
+++ b/.changeset/hot-timers-complain.md
@@ -1,0 +1,7 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`DialogPrimitive` - Fixed issue with `box-sizing` inheritance
+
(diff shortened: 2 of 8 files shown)


KristinLBradley (approved):

packages/components/src/styles/components/dialog-primitive.scss
```diff
@@ -13,24 +13,25 @@ $hds-dialog-primitive-vertical-padding: 16px;
...
 // Since the `:where()` selector has a specificity of `0` (see https://developer.mozilla.org/en-US/docs/Web/CSS/:where)
 // the styles declared below are applied to the `<dialog>` element, but are overwritten by any other style applied via classname
 //
 
 :where(.hds-dialog-primitive__wrapper) {
-  // -----------
-  all: unset; // removes default dialog & inherited styles. Must be the first style to be applied
-  // -----------
+  // override default `<dialog>` styles
+  position: initial;
```
shleewhite, line 25: I like that we're being explicit about what we are overriding from `<dialog>`, makes it easier to understand!

shleewhite (approved):

PR #2443, opened 2024-09-23
`Dropdown`: added `isOpen` state to dropdown content

shleewhite:
### :pushpin: Summary

If merged, this PR would expose the `isOpen` state to the yielded content. This would allow implementations like:
```hbs
<Hds::Dropdown @listPosition="bottom-left" @width="200px" as |D|>
  <D.ToggleButton @color="secondary" @text="Menu" />
  {{#if D.isOpen}}
    <D.Interactive @href="#">
      Lorem ipsum dolor sit amet
    </D.Interactive>
    <D.Interactive @href="#">
      Consectetur adipisicing elit
    </D.Interactive>
  {{/if}}
</Hds::Dropdown>
```

* adds `isOpen` to the yielded content
* adds a test that the `isOpen` state works correctly
* adds a showcase example using it
* update website docs to make yielded close and isOpen more clear

### :link: External links

Jira ticket: [HDS-3871](https://hashicorp.atlassian.net/browse/HDS-3871)

***

### 👀 Component checklist

[HDS-3871]: https://hashicorp.atlassian.net/browse/HDS-3871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/docs/components/dropdown/partials/code/component-api.md
+++ b/website/docs/components/dropdown/partials/code/component-api.md
@@ -56,12 +56,12 @@ The Dropdown component is composed of different child components each with their
   <C.Property @name="<[D].Footer>" @type="yielded component">
     `Dropdown::Footer` yielded as contextual component (see below).
   </C.Property>
-  <C.Property @name="<[D].close>" @type="function">
+  <C.Property @name="[D].close" @type="function">
     Function to programmatically close the Dropdown yielded to the content.
     <br/><br/>
     If this function is invoked using an `\{{on "click"}}` modifier applied to the `ListItem::Interactive` element, there is a quirky behavior of the Ember `<LinkTo>` component which requires a workaround to have the events executed in the right order (this happens only if it has a `@route` argument). Read more about the issue and a possible solution [in this GitHub comment](https://github.com/hashicorp/design-system/pull/399#issuecomment-1171186772).
   </C.Property>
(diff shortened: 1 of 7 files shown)


website/docs/components/dropdown/partials/code/component-api.md
```diff
@@ -56,6 +56,14 @@ The Dropdown component is composed of different child components each with their
   <C.Property @name="<[D].Footer>" @type="yielded component">
     `Dropdown::Footer` yielded as contextual component (see below).
   </C.Property>
+  <C.Property @name="<[D].close>" @type="function">
```
didoo, line 59: [suggestion] we could consider adding something also to the "How to use section", in which we explain the fact (maybe not obvious) that the dropdown content is injected in the DOM when rendered, and if a user needs to prevent this behaviour tehy can use the `isOpen` variable (eg. because it triggers an async call to a backend or some similar action that would be better to defer to when the user actually toggles the dropdown)

didoo (approved):
Left a couple of non-blocking comments. LGTM 👍

website/docs/components/dropdown/partials/code/component-api.md
shleewhite: Good point! Added one.

PR #2445, opened 2024-09-23
Fix `RadioCard` styles and interactive states

jorytindall:
### :pushpin: Summary

This PR fixes the tokens applied to the `RadioCard` in it's selected state and updates the color of text elements within a disabled `RadioCard`.

Previously the `RadioCard` was using `focus-action-internal` for the `border-color` when a `RadioCard` was selected. Not only is this the incorrect token, but it differs from Figma and the interactive states of the component. This change updates the border color for a selected `RadioCard` to:

- `blue-300` for resting/default state
- `blue-400` for hover state

Additionally, we are proposing in design to change the tokens for text elements (namely the icon, `description`, and `label` within the `RadioCard`) to use `form-control-disabled-foreground-color` to semantically match the rest of the tokens applied when the component is disabled.

**Before**
Screenshot 2024-09-23 at 1 24 32 PM

**After**
Screenshot 2024-09-23 at 1 24 06 PM

### :link: External links

Jira ticket: [HDS-3868](https://hashicorp.atlassian.net/browse/HDS-3868)
Figma file: [Update RadioCard disabled colors](https://www.figma.com/design/noyY6dUMDYjmySpHcMjhkN/branch/3ctSmHIlVaRYqa8E6OaIaG/HDS-Components?m=auto&node-id=16838-51020&t=JkKW71GXhXPqXkeX-1)

***

### 👀 Component checklist

[HDS-3868]: https://hashicorp.atlassian.net/browse/HDS-3868?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/styles/components/form/radio-card.scss
+++ b/packages/components/src/styles/components/form/radio-card.scss
@@ -96,7 +96,7 @@
     }
 
     .hds-form-radio-card__content {
-      opacity: 0.6;
+      opacity: 0.5;
     }
   }
 }
--- a/packages/components/src/styles/components/form/radio-card.scss
+++ b/packages/components/src/styles/components/form/radio-card.scss
@@ -79,10 +79,10 @@
 
   &--disabled,
   &.mock-disabled {
+    color: var(--token-form-control-disabled-foreground-color);
     background-color: var(--token-color-surface-interactive-disabled);
     border-color: var(--token-color-border-primary);
     box-shadow: none;
-    color: var(--token-form-control-disabled-foreground-color);
     cursor: not-allowed;
 
     .hds-form-radio-card__label,
--- a/packages/components/src/styles/components/form/radio-card.scss
+++ b/packages/components/src/styles/components/form/radio-card.scss
@@ -98,7 +98,7 @@
     /* Set the opacity of nested components to make them appear disabled */
     .hds-badge,
     .hds-icon {
-      opacity: 50%;
+      opacity: 0.5;
     }
   }
 }
--- a/packages/components/src/styles/components/form/radio-card.scss
+++ b/packages/components/src/styles/components/form/radio-card.scss
@@ -95,10 +95,8 @@
       border-color: var(--token-color-border-primary);
     }
 
-    /* Set the opacity of nested components to make them appear disabled */
-    .hds-badge,
-    .hds-icon {
-      opacity: 0.5;
+    .hds-form-radio-card__content {
+      opacity: 0.6;
     }
   }
 }
--- a/packages/components/src/styles/components/form/radio-card.scss
+++ b/packages/components/src/styles/components/form/radio-card.scss
@@ -7,7 +7,6 @@
 // FORM > RADIO-CARD
 //
 
-
 // RadioCard Group
 
 .hds-form-group--radio-cards {
@@ -38,7 +37,8 @@
   display: flex;
   flex-direction: column;
   background-color: var(--token-color-surface-primary);
-  border: var(--token-form-radiocard-border-width) solid var(--token-color-border-primary);
+  border: var(--token-form-radiocard-border-width) solid
+    var(--token-color-border-primary);
   border-radius: var(--token-form-radiocard-border-radius);
   box-shadow: var(--token-elevation-mid-box-shadow);
   cursor: pointer;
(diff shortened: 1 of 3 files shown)


packages/components/src/styles/components/form/radio-card.scss
```diff
@@ -64,21 +64,32 @@
 
   &--checked,
   &.mock-checked {
-    border-color: var(--token-color-focus-action-internal);
+    border-color: var(--token-color-palette-blue-300);
```
jorytindall, line 67: Update the default border color for a selected `RadioCard`

packages/components/src/styles/components/form/radio-card.scss
```diff
@@ -64,21 +64,32 @@
...
 
     .hds-form-radio-card__control-wrapper {
       background-color: var(--token-color-surface-action);
       border-color: var(--token-color-border-action);
     }
+
+    &:hover,
+    &.mock-hover {
+      border-color: var(--token-color-palette-blue-400);
```
jorytindall, line 76: Add the interactive style on hover.

packages/components/src/styles/components/form/radio-card.scss
```diff
@@ -64,21 +64,32 @@
...
+    }
   }
 
   &--disabled,
   &.mock-disabled {
     background-color: var(--token-color-surface-interactive-disabled);
     border-color: var(--token-color-border-primary);
     box-shadow: none;
+    color: var(--token-form-control-disabled-foreground-color);
```
jorytindall, line 85: Set the base color for child elements within the `RadioCard` in it's disabled state.

packages/components/src/styles/components/form/radio-card.scss
```diff
@@ -64,21 +64,32 @@
...
     cursor: not-allowed;
 
+    .hds-form-radio-card__label,
+    .hds-form-radio-card__description {
+      color: var(--token-form-control-disabled-foreground-color);
```
jorytindall, line 90: Overwrite with specificity the `label` and `description` text elements.

heatherlarsen (approved):
lgtm 👍🏻

alex-ju (approved):
Code changes look good!

One note, as you mentioned, the disabled overrides are only applied to labels and descriptions; a coloured icon or badge (or any custom element for that matter) would still have the same aspect.

Screenshot 2024-09-24 at 15 14 45

jorytindall:
@alex-ju 
> One note, as you mentioned, the disabled overrides are only applied to labels and descriptions; a coloured icon or badge (or any custom element for that matter) would still have the same aspect.

Yes, I did think about this and there didn't really seem like a systematic way to approach that so I kept the scope specific to those elements.

didoo:
> @alex-ju
> 
> > One note, as you mentioned, the disabled overrides are only applied to labels and descriptions; a coloured icon or badge (or any custom element for that matter) would still have the same aspect.
> 
> Yes, I did think about this and there didn't really seem like a systematic way to approach that so I kept the scope specific to those elements.

Don't have full context (and I may have mentioned elsewhere, so forgive me if that's the case) but one trick I used to "emulate" a disabled state of something was to change its opacity to 50/40%, so it would blend in with the background; maybe something that can be done to the icon? worth a try?

jorytindall:
@didoo interesting suggestion, I played around with this and while I don't think it's feasible to target every nested component (especially when generic content is passed), it seems pragmatic to target the icon and the badge directly with this method. I updated the example in the showcase for reference, let me know what you think of the method and I'll revert it before committing (unless you see a benefit in including an example like this).

didoo:
> @didoo interesting suggestion, I played around with this and while I don't think it's feasible to target every nested component (especially when generic content is passed), it seems pragmatic to target the icon and the badge directly with this method. I updated the example in the showcase for reference, let me know what you think of the method and I'll revert it before committing (unless you see a benefit in including an example like this).

@jorytindall (+ @alex-ju) you may consider changing the opacity of the whole `.hds-form-radio-card__content` container, instead of targeting its children one by one (which means leaving the text foreground color as “strong/primary”); this would result in something like this:
screenshot_4220

alex-ju:
if acceptable from a design standpoint, changing the opacity for the entire card seems like a more robust approach

PR #2453, opened 2024-09-24
`DialogPrimitive`: add a guard so the yielded `close` function is always defined

shleewhite:
### :pushpin: Summary

If merged, this PR would create a `onDismiss` getter in the DialogPrimitive so the close function that is yielded to the Footer is always defined.

### :link: External links

Jira ticket: [HDS-3889](https://hashicorp.atlassian.net/browse/HDS-3889)

***

### 👀 Component checklist

- [X] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

[HDS-3889]: https://hashicorp.atlassian.net/browse/HDS-3889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/dialog-primitive/footer.ts
+++ b/packages/components/src/components/hds/dialog-primitive/footer.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import templateOnlyComponent from '@ember/component/template-only';
+import Component from '@glimmer/component';
 
 export interface HdsDialogPrimitiveFooterSignature {
   Args: {
@@ -15,14 +15,26 @@ export interface HdsDialogPrimitiveFooterSignature {
     default: [
       {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        close?: (event: MouseEvent, ...args: any[]) => void;
+        close: (event: MouseEvent, ...args: any[]) => void;
       },
     ];
   };
   Element: HTMLDivElement;
 }
 
-const HdsDialogPrimitiveFooter =
-  templateOnlyComponent<HdsDialogPrimitiveFooterSignature>();
+const NOOP = (): void => {};
 
-export default HdsDialogPrimitiveFooter;
+export default class HdsDialogPrimitiveFooter extends Component<HdsDialogPrimitiveFooterSignature> {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  get onDismiss(): (event: MouseEvent, ...args: any[]) => void {
+    const { onDismiss } = this.args;
+
+    // notice: this is to make sure the function is always defined when consumers add `{{on 'click' F.close}}` to a button in the DialogFooter.
+    // in reality it's always used inside the main components as a yielded component, so the onDismiss handler is always defined
+    if (typeof onDismiss === 'function') {
+      return onDismiss;
+    } else {
+      return NOOP;
+    }
+  }
+}
--- a/.changeset/tricky-forks-lay.md
+++ b/.changeset/tricky-forks-lay.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`DialogPrimitive`: added a guard so the yielded close function is always defined'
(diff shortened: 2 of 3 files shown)


packages/components/src/components/hds/dialog-primitive/footer.ts
```diff
@@ -15,14 +15,26 @@ export interface HdsDialogPrimitiveFooterSignature {
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  get onDismiss(): (event: MouseEvent, ...args: any[]) => void {
+    const { onDismiss } = this.args;
+
+    // notice: this is to make sure the function is always defined when consumers add `{{on 'click' F.close}}` to a button in the DialogFooter.
+    // in reality it's always used inside the main components as a yielded component, so the onDismiss handler is always defined
+    if (typeof onDismiss === 'function') {
+      return onDismiss;
+    } else {
+      return NOOP;
```
didoo, line 37: [question] I see that in the `Hds::Alert` component we have a very similar argument, but we use a slightly different approach (we simply return `false` in the getter, without the need to define a `noop` operation): https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/alert/index.ts 

Which one is more "correct™" (if there is one)?
Should we align this approach across the HDS components?

packages/components/src/components/hds/dialog-primitive/footer.ts
```diff
@@ -15,14 +15,26 @@ export interface HdsDialogPrimitiveFooterSignature {
     default: [
       {
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
-        close?: (event: MouseEvent, ...args: any[]) => void;
+        close: (event: MouseEvent, ...args: any[]) => void;
```
alex-ju, line 18: question: not sure I follow why we enforce the `close` function to exist in a block

shleewhite: There were several type errors where HdsButtons in the modal footer complained that the close function could be undefined.

Screenshot 2024-09-25 at 10 07 29 AM

packages/components/src/components/hds/dialog-primitive/footer.ts
shleewhite: Interesting... I got this implementation from [the DialogHeader](https://github.com/hashicorp/design-system/blob/36a45f0be2f6ea595e372e27e1d34a23774d19f9/packages/components/src/components/hds/dialog-primitive/header.ts#L40).

Looking at where the type errors are happening, if it returned false it wouldn't resolve the type issues.

packages/components/src/components/hds/dialog-primitive/footer.ts
alex-ju: hm.. interesting how only by making this change I get the `modal-*.gts` components in HCP to pass linting 🤔

alex-ju: could it be the case that we don't really need the fallback noop function?

alex-ju: @shleewhite and I discussed this and making the `close` mandatory would not make for a flexible implementation, considering the `DisclosurePrimitive::Footer` may not contain actions; we'll stick with defining a `noop` as a fallback

shleewhite: @alex-ju and I talked about this more, we decided that to make close always defined without a wrapping function would make onDismiss also required. This is a larger change that kind of goes against how the DialogPrimitive is supposed to be used, so we will stick with the wrapping function approach.

alex-ju (approved):

packages/components/src/components/hds/dialog-primitive/footer.ts
alex-ju: the `onDismiss` function in Alert is slightly different in a sense that it is not forwarded as `F.close` (or similar) so proper checks for it to actually exists and be a function are not required

didoo: Makes sense, I've resolved the comment then

PR #2456, opened 2024-09-25
Rebuild declaration files on the fly

alex-ju:
### :pushpin: Summary

Improve the developer experience by also rebuilding the declaration files on the fly when running `yarn start` in `packages/components`

***


packages/components/package.json
```diff
@@ -21,7 +21,9 @@
     "build:js": "rollup --config",
     "build:types": "glint --declaration",
     "prepublishOnly": "yarn build && test -f 'dist/styles/@hashicorp/design-system-components.css' || (echo 'The pre-compiled CSS file was not found' && exit 1)",
-    "start": "rollup --config --watch --environment development",
+    "start": "concurrently 'yarn:start:*'",
+    "start:js": "rollup --config --watch --environment development",
```
aklkv, line 25: I wonder what `--environment development` flag is used for. Here is the same lines from the blueprint https://github.com/embroider-build/addon-blueprint/blob/main/files/__addonLocation__/package.json#L29

alex-ju: we have a distinct setup when working with this package locally to overcome an issue we had where the build fails when a js file is deleted ([more context here](https://github.com/hashicorp/design-system/pull/1946))

aklkv: cool 👍🏻 if it works for you!

PR #2460, opened 2024-09-26
chore: remove ember-keyboard dependency

shleewhite:
### :pushpin: Summary

If merged, this PR would remove the dependency on ember-keyboard. It seems like this was a relic dependency from [the CTA component](https://github.com/hashicorp/design-system/commit/01d4b033bfa2a35968208da5abcec2fa22fefd8a), but it has not been used in a long time.

I looked into using It for the dataGrid spike, but there is not a strong reason to use it over the native addEventListener plus it isn't typesafe yet.

- [X] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4293,7 +4293,6 @@ __metadata:
     ember-element-helper: "npm:^0.8.5"
     ember-focus-trap: "npm:^1.1.0"
     ember-get-config: "npm:^2.1.1"
-    ember-keyboard: "npm:^8.2.1"
     ember-modifier: "npm:^4.1.0"
     ember-power-select: "npm:^8.2.0"
     ember-stargate: "npm:^0.4.3"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -13194,7 +13193,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"ember-destroyable-polyfill@npm:^2.0.1, ember-destroyable-polyfill@npm:^2.0.2, ember-destroyable-polyfill@npm:^2.0.3":
+"ember-destroyable-polyfill@npm:^2.0.1, ember-destroyable-polyfill@npm:^2.0.2":
```
didoo, line 13196: [thought] @alex-ju @shleewhite would it be worth to run a `yarn dedupe` in general (in a different PR, maybe; or maybe not)? as far as I remember, @alex-ju did this operation with some cadence

alex-ju: yes, we do `yarn dedupe` when upgrading packages. we're due an update soon

PR #2463, opened 2024-09-27
`AppSideNav` - Follow-on improvements for headers & aria (HDS-3897)

KristinLBradley:
### :pushpin: Summary

If merged, this PR will make improvements to the `AppSideNav` component related to `List::Title` headers and the `ToggleButton` aria properties. 

(It will merge into the component branch, not directly into the main branch.)

**Implemented changes:**

* Change `List::Title` to h3 & add visually hidden h2 to AppSideNav - [Reference comment](https://github.com/hashicorp/design-system/pull/2384#discussion_r1764009486)
* Improve ariaLabel for `ToggleButton` state & use aria-expanded - [Reference comment](https://github.com/hashicorp/design-system/pull/2384#discussion_r1764017267)

### :link: External links

* Jira ticket: [HDS-3897](https://hashicorp.atlassian.net/browse/HDS-3897)

***

### 👀 Component checklist

- ~~[ ] A changelog entry was added via [Changesets]~~(https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

[HDS-3897]: https://hashicorp.atlassian.net/browse/HDS-3897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/app-side-nav/base.hbs
+++ b/packages/components/src/components/hds/app-side-nav/base.hbs
@@ -4,6 +4,8 @@
 }}
 {{! IMPORTANT: we need to add "squishies" here (~) because otherwise the whitespace added by Ember causes the empty element to still have visible padding - See https://handlebarsjs.com/guide/expressions.html#whitespace-control }}
 <div class="hds-app-side-nav" ...attributes>
+  <h2 class="sr-only">Local navigation</h2>
+
   <div class="hds-app-side-nav__wrapper">
     {{yield to="root"}}
     <div class="hds-app-side-nav__wrapper-header">
--- a/packages/components/src/components/hds/app-side-nav/base.hbs
+++ b/packages/components/src/components/hds/app-side-nav/base.hbs
@@ -4,7 +4,7 @@
 }}
 {{! IMPORTANT: we need to add "squishies" here (~) because otherwise the whitespace added by Ember causes the empty element to still have visible padding - See https://handlebarsjs.com/guide/expressions.html#whitespace-control }}
 <div class="hds-app-side-nav" ...attributes>
-  <h2 class="sr-only">Local navigation</h2>
+  <h2 class="sr-only" id="hds-app-side-nav-header">Application local navigation</h2>
 
   <div class="hds-app-side-nav__wrapper">
     {{yield to="root"}}
--- a/packages/components/src/components/hds/app-side-nav/index.hbs
+++ b/packages/components/src/components/hds/app-side-nav/index.hbs
@@ -19,7 +19,7 @@
       <div class="hds-app-side-nav__overlay" {{on "click" this.toggleMinimizedStatus}} />
       {{! template-lint-enable no-invalid-interactive}}
       <Hds::AppSideNav::ToggleButton
-        aria-label="App navigation"
+        aria-labelledby="hds-app-side-nav-header"
         aria-expanded={{if this.isMinimized "false" "true"}}
         @icon={{if this.isMinimized "chevrons-right" "chevrons-left"}}
         {{on "click" this.toggleMinimizedStatus}}
--- a/packages/components/src/components/hds/app-side-nav/index.hbs
+++ b/packages/components/src/components/hds/app-side-nav/index.hbs
@@ -19,7 +19,8 @@
       <div class="hds-app-side-nav__overlay" {{on "click" this.toggleMinimizedStatus}} />
       {{! template-lint-enable no-invalid-interactive}}
       <Hds::AppSideNav::ToggleButton
-        aria-label={{this.toggleButtonAriaLabel}}
+        aria-label="App navigation"
+        aria-expanded={{if this.isMinimized "false" "true"}}
         @icon={{if this.isMinimized "chevrons-right" "chevrons-left"}}
         {{on "click" this.toggleMinimizedStatus}}
       />
--- a/packages/components/src/components/hds/app-side-nav/list/index.hbs
+++ b/packages/components/src/components/hds/app-side-nav/list/index.hbs
@@ -5,12 +5,12 @@
 
 <nav class="hds-app-side-nav__list-wrapper" aria-labelledby="hds-app-side-nav-header" ...attributes>
   {{yield (hash ExtraBefore=(component "hds/yield"))}}
-  <ul class="hds-app-side-nav__list" role="list">
+  <ul class="hds-app-side-nav__list" role="list" aria-labelledby={{this.titleIds}}>
     {{yield
       (hash
         Item=(component "hds/app-side-nav/list/item")
         BackLink=(component "hds/app-side-nav/list/back-link")
-        Title=(component "hds/app-side-nav/list/title")
+        Title=(component "hds/app-side-nav/list/title" didInsertTitle=this.didInsertTitle)
         Link=(component "hds/app-side-nav/list/link")
       )
     }}
--- a/packages/components/src/components/hds/app-side-nav/list/index.hbs
+++ b/packages/components/src/components/hds/app-side-nav/list/index.hbs
@@ -3,7 +3,7 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-<nav class="hds-app-side-nav__list-wrapper" ...attributes>
+<nav class="hds-app-side-nav__list-wrapper" aria-labelledby="hds-app-side-nav-header" ...attributes>
   {{yield (hash ExtraBefore=(component "hds/yield"))}}
   <ul class="hds-app-side-nav__list" role="list">
     {{yield
--- a/packages/components/src/components/hds/app-side-nav/list/title.hbs
+++ b/packages/components/src/components/hds/app-side-nav/list/title.hbs
@@ -4,8 +4,8 @@
 }}
 
 <Hds::AppSideNav::List::Item>
-  <div
+  <h3
     class="hds-app-side-nav__list-title hds-typography-body-100 hds-font-weight-semibold"
     ...attributes
-  >{{~yield~}}</div>
+  >{{~yield~}}</h3>
 </Hds::AppSideNav::List::Item>
--- a/packages/components/src/components/hds/app-side-nav/list/title.hbs
+++ b/packages/components/src/components/hds/app-side-nav/list/title.hbs
@@ -4,10 +4,10 @@
 }}
 
 <Hds::AppSideNav::List::Item>
-  <h3
+  <div
     class="hds-app-side-nav__list-title hds-typography-body-100 hds-font-weight-semibold"
     id={{this.titleId}}
     {{did-insert this.didInsertTitle}}
     ...attributes
-  >{{~yield~}}</h3>
+  >{{~yield~}}</div>
 </Hds::AppSideNav::List::Item>
--- a/packages/components/src/components/hds/app-side-nav/list/title.hbs
+++ b/packages/components/src/components/hds/app-side-nav/list/title.hbs
@@ -6,6 +6,8 @@
 <Hds::AppSideNav::List::Item>
   <h3
     class="hds-app-side-nav__list-title hds-typography-body-100 hds-font-weight-semibold"
+    id={{this.titleId}}
+    {{did-insert this.didInsertTitle}}
     ...attributes
   >{{~yield~}}</h3>
 </Hds::AppSideNav::List::Item>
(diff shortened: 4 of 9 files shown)


packages/components/src/components/hds/app-side-nav/base.hbs
```diff
@@ -4,6 +4,8 @@
 }}
 {{! IMPORTANT: we need to add "squishies" here (~) because otherwise the whitespace added by Ember causes the empty element to still have visible padding - See https://handlebarsjs.com/guide/expressions.html#whitespace-control }}
 <div class="hds-app-side-nav" ...attributes>
+  <h2 class="sr-only">Local navigation</h2>
```
KristinLBradley, line 7: Let me know if this header should use different wording, be moved elsewhere in the DOM, and/or have the text exposed in case consumers would want to override it, etc.

KristinLBradley: @shleewhite @MelSumner

packages/components/src/components/hds/app-side-nav/list/title.hbs
```diff
@@ -4,8 +4,8 @@
 }}
 
 <Hds::AppSideNav::List::Item>
-  <div
+  <h3
     class="hds-app-side-nav__list-title hds-typography-body-100 hds-font-weight-semibold"
     ...attributes
-  >{{~yield~}}</div>
+  >{{~yield~}}</h3>
```
KristinLBradley, line 10: I change the div tags to h3 tags for now.

KristinLBradley: I'm not sure how I would be able to associate the ul with the title using aria-labelled by with the current structure. I think the whole structure would have to be redone to make it work. 

(I worked on the SideNav in the early stages and wanted to structure it as headers outside of uls which served as the title of the uls but was overruled in favor of the current structure. The structure has complexities due to some of how it functions within Products I believe.)

packages/components/src/components/hds/app-side-nav/index.hbs
```diff
@@ -19,7 +19,8 @@
       <div class="hds-app-side-nav__overlay" {{on "click" this.toggleMinimizedStatus}} />
       {{! template-lint-enable no-invalid-interactive}}
       <Hds::AppSideNav::ToggleButton
-        aria-label={{this.toggleButtonAriaLabel}}
+        aria-label="App navigation"
```
KristinLBradley, line 22: Let me know if this label is ok or if it should be changed, moved, and/or exposed so consumers can override

packages/components/src/components/hds/app-side-nav/list/title.hbs
KristinLBradley: If we really need to do this (aria-labelled changes to associate the Title headers with the Lists) then I could make another branch of the component PR as it's going to require more extensive changes and testing than the other smaller updates I've made in this current branch. Please let me know what you think.

packages/components/src/components/hds/app-side-nav/base.hbs
MelSumner: Should this be "Application navigation" ? 

Also, is there an aria-label on the nav element in the sidebar? It might be redundant to have both; I'd update in these ways:

1. Add an `id` to the `h2`
2. change the `aria-label` in the aside nav to be `aria-labelledby` and give it the value of the `h2`'s `id` attribute.

packages/components/src/components/hds/app-side-nav/list/title.hbs
MelSumner: 1. The lists do need to have accNames or else they just show up as "list" in the rotor
2. I think it would be sufficient to have this list item given an `id` attribute and then use `aria-labelledby` on the `ul`. 

I don't think these should be h3's.

packages/components/src/components/hds/app-side-nav/base.hbs
KristinLBradley: The `nav` element which wraps each list doesn't have an aria-label (I could add an aria-labelledby though/). There can be multiple `nav` elements potentially and likely. The `appFrame` `sidebar` child component where the `AppSideNav` slots into doesn't have an aria-label.

Does it matter if the h2 labels multiple nav elements?

KristinLBradley: I updated the h2 text to "Application local navigation" since the AppSideNav is supposed to contain page level navigation and to be paired with the AppHeader which contains "global" & "utility" navigation. Let me know if I should change the text to something else again.

I added a hard-coded id to the h2 and aria-labelledby with the id valu to the `nav` elements wrapping lists of links. Again there are likely to be multiple nav elements. (I hard-coded the ids instead of generating since it didn't seem like we would really need to generate them but I can change that too if desired.)

Please let me know if these changes are ok or if I should make more. @MelSumner @shleewhite @alex-ju

packages/components/src/components/hds/app-side-nav/list/title.hbs
KristinLBradley: Can you take a look at the code and let me know if you have ideas on how to do it with the current structure? Redoing the current structure if something I really want to avoid unless absolutely necessary.

If the current structure is unacceptable on the AppSideNav it will need to be updated on the SideNav too.

@shleewhite @alex-ju Please let me know if you have any input on this.

packages/components/src/components/hds/app-side-nav/index.hbs
shleewhite: Hmm.. I think it should match what the h2 calls it for consistency - so "Application local navigation"

packages/components/src/components/hds/app-side-nav/list/title.hbs
shleewhite: Hmmm.. I hadn't realized before that it is just one list with everything inside it... this is tough because visually theyre either separate lists or sub-lists, both of which would require refactoring parts of the structure. 

Can you clarify why you would need to update it in sidenav too? Isn't it going to be deprecated?

packages/components/src/components/hds/app-side-nav/base.hbs
shleewhite: I think its fine.... does the AppHeader navigation have an accessible name?

packages/components/src/components/hds/app-side-nav/index.hbs
alex-ju: in this case, I suggest we point to the heading to avoid duplicating content
```suggestion
        aria-labelledby="hds-app-side-nav-header"
```

alex-ju (commented):
This looks good overall! I will wait to see if we come up with a pragmatic solution for [titles in the list](https://github.com/hashicorp/design-system/pull/2463/files#r1779208065). Otherwise we'll record the need for improvements in a follow-up tasks.

packages/components/src/components/hds/app-side-nav/list/title.hbs
KristinLBradley: Update: From squad discussions, we will avoid making new updates the the SideNav unless absolutely necessary.

packages/components/src/components/hds/app-side-nav/base.hbs
KristinLBradley: If the AppHeader needs to have a header or aria-label or similar added could you please open a new issue or PR? I want to focus on the AppSideNav for the current open PRs.

packages/components/src/components/hds/app-side-nav/list/title.hbs
KristinLBradley: Let me know if you have ideas on improvements that could be made to the current structure or if the updates I already made are ok for now.

KristinLBradley:
I'm getting this error now since there are multiple AppSideNav components on the Showcase page although in actual usage there would only ever be one per page:

```
Error: The page should have no accessibility violations. Violations:
            [moderate]: Ensures landmarks are unique 
            Violated 1 time. Offending nodes are: 
            <nav class="hds-app-side-nav__list-wrapper" aria-labelledby="hds-app-side-nav-header" aria-label="Primary on portal demo 1">
```

alex-ju:
> I'm getting this error now since there are multiple AppSideNav components on the Showcase page although in actual usage there would only ever be one per page:
> 
> ```
> Error: The page should have no accessibility violations. Violations:
>             [moderate]: Ensures landmarks are unique 
>             Violated 1 time. Offending nodes are: 
>             <nav class="hds-app-side-nav__list-wrapper" aria-labelledby="hds-app-side-nav-header" aria-label="Primary on portal demo 1">
> ```
> 
> I suppose I could generate a unique id instead of hard-coding as I've done currently but I would only be doing it for the Showcase instead of a real use case so that seems like overkill. Let me know what you think though. @alex-ju

Indeed, setting a unique ID will not have any benefits for our consumers.

The only way around it would be to disable the check for this specific element in our automated tests:

```
'landmark-unique': {
  enabled: false,
  selectors: [['.hds-app-side-nav__list-wrapper']],
}
```

packages/components/src/components/hds/app-side-nav/list/index.hbs
```diff
@@ -3,7 +3,7 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-<nav class="hds-app-side-nav__list-wrapper" ...attributes>
+<nav class="hds-app-side-nav__list-wrapper" aria-labelledby="hds-app-side-nav-header" ...attributes>
   {{yield (hash ExtraBefore=(component "hds/yield"))}}
   <ul class="hds-app-side-nav__list" role="list">
```
MelSumner, line 8: This is where you'd add an aria-labelledby attribute, something like:

```suggestion
  <ul class="hds-app-side-nav__list" role="list" aria-labelledby="listTitle-id">
```

And then in the `app-side-nav/list-title`, add a `guidFor` that produces a unique `id` for the title. 

If we associate these correctly, the user would not have to change anything on their end.

KristinLBradley: Thanks I pushed up changes to add an id to each Title and generate a list of TitleIds that gets inserted in the aria-labelledby of the ul.

Let me know if I should remove the h3s I added earlier for the Titles.

packages/components/src/components/hds/app-side-nav/list/title.hbs
KristinLBradley: I pushed up some changes to add ids to Titles and associate them with the `ul`.

packages/components/src/components/hds/app-side-nav/list/index.hbs
KristinLBradley: @MelSumner @shleewhite When you have time could you please review the latest changes I made from @MelSumner's feedback above?

packages/components/src/components/hds/app-side-nav/list/title.hbs
shleewhite: After reviewing it again, I think we should just take the work for labelling the lists by the titles out of scope for now. The reason being, now there is still one large list called "Services Default Org". 

We need to come back to this and restructure the components slightly as demonstrated in this spike: https://github.com/hashicorp/design-system/pull/2478/files

Unfortunately, I don't think there are any small wins possible for the sublists/titles without changing the structure.

I can approve this PR once it is scoped down to just adding the visually hidden H2 and labelling the nav by the H2.

KristinLBradley: Thanks @shleewhite 

@MelSumner could you please let me know your input? Are you ok with what Lee said above? I'd like to know any specific changes I should make to this current PR so I can get it merged into the component branch (if you would like changes other than what Lee recommended).

Thanks!

KristinLBradley: For now, I changed the `h3` I had added wrapping the Title content back into a `div`

MelSumner (approved):
This approach seems to work, I think we could ship this for now but also discuss Lee's sideNav spike and compare to see if we get any performance wins one way or the other. Thanks for taking the time to be thoughtful about the accessibility for this component!

KristinLBradley:
Going to merge this into the component branch as @MelSumner approved. (There will still be time to make last minor tweaks before the component branch is merged if there's any additional feedback.)

PR #2472, opened 2024-10-01
`AppSideNav` - overlay scrolling bug fix (HDS-3914)

KristinLBradley:
### :pushpin: Summary

If merged, this PR fixes an issue with the body content scrolling when the `AppSideNav` is expanded as an overlay in mobile view. The fix prevents the body content from scrolling while maintaining the previous scroll position.

(It will merge into the component branch, not directly into the main branch.)

**Preview of fix in action:** https://hds-showcase-git-hds-3914-app-side-nav-overlay-d5df13-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-app-header-and-app-side-nav

**To test:**

1. Narrow browser window until you are in "mobile" view
2. With the AppSideNav still minimizing scroll the main content a noticeable amount
3. Use the Toggle Button to expand the AppSideNav
4. Attempt to scroll within the main content under the gray screen overlay.

**Expected:** The main content should not react when you try to scroll. The previous scroll position should be maintained instead of jumping back to the top of the main content.

### :link: External links

* Jira ticket: [HDS-3914](https://hashicorp.atlassian.net/browse/HDS-3914)

***

### 👀 Component checklist

[HDS-3914]: https://hashicorp.atlassian.net/browse/HDS-3914?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/app-side-nav/index.ts
+++ b/packages/components/src/components/hds/app-side-nav/index.ts
@@ -136,6 +136,18 @@ export default class HdsAppSideNav extends Component<HdsAppSideNavSignature> {
     });
   }
 
+  lockBodyScroll(): void {
+    document.body.style.top = `-${window.scrollY}px`;
+    document.body.style.position = 'fixed';
+  }
+
+  unlockBodyScroll(): void {
+    const verticalScrollPosition = document.body.style.top;
+    document.body.style.position = '';
+    document.body.style.top = '';
+    window.scrollTo(0, parseInt(verticalScrollPosition || '0') * -1);
+  }
+
   @action
   escapePress(event: KeyboardEvent): void {
     if (event.key === 'Escape' && !this.isMinimized && !this.isDesktop) {
@@ -147,13 +159,18 @@ export default class HdsAppSideNav extends Component<HdsAppSideNavSignature> {
   @action
   toggleMinimizedStatus(): void {
     this.isMinimized = !this.isMinimized;
-
     this.synchronizeInert();
 
     const { onToggleMinimizedStatus } = this.args;
 
     if (typeof onToggleMinimizedStatus === 'function') {
       onToggleMinimizedStatus(this.isMinimized);
+    }
+
+    if (this.isMinimized) {
+      this.unlockBodyScroll();
+    } else {
+      this.lockBodyScroll();
     }
   }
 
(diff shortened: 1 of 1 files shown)


alex-ju (commented):
We took a slightly different approach in [Modal](https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/modal/index.ts) (and Flyout), changing the `overflow` value and then resetting it. Would that approach not work in this scenario?

KristinLBradley:
> We took a slightly different approach in [Modal](https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/modal/index.ts) (and Flyout), changing the `overflow` value and then resetting it. Would that approach not work in this scenario?

@alex-ju I can look at it, thanks.

KristinLBradley:
@alex-ju I pushed up refactored code following the Modal code.

jorytindall (approved):
This works well to me!

packages/components/src/components/hds/app-side-nav/index.ts
```diff
@@ -136,6 +138,30 @@ export default class HdsAppSideNav extends Component<HdsAppSideNavSignature> {
     });
   }
 
+  lockBodyScroll(): void {
+    if (this.body) {
+      // Store the initial `overflow` value of `<body>` so we can reset to it
+      this.bodyInitialOverflowValue =
+        this.body.style.getPropertyValue('overflow');
```
alex-ju, line 145: thought: the initial value could be stored only once, in the `didInsert`, right?

alex-ju (approved):
Left a comment that would save some unnecessary assignments – otherwise looks good. Don't forget to add a changelog entry

KristinLBradley:
> Left a comment that would save some unnecessary assignments – otherwise looks good. Don't forget to add a changelog entry

@alex-ju 
Should I add a changelog when this is an update to a new component (AppSideNav) and I'm planning to merge this into the component branch to be released together with it?

packages/components/src/components/hds/app-side-nav/index.ts
KristinLBradley: Pushed up refactored code

alex-ju:
> > Left a comment that would save some unnecessary assignments – otherwise looks good. Don't forget to add a changelog entry
> 
> @alex-ju 
> Should I add a changelog when this is an update to a new component (AppSideNav) and I'm planning to merge this into the component branch to be released together with it?

you're right, no changelog needed!

PR #2486, opened 2024-10-08
`AppSideNav` - Replace tracked variables from args

KristinLBradley:
### :pushpin: Summary

If merged, this PR replaces tracked variables from args formerly used in `AppSideNav`.

(This branch merges into the AppSideNav component branch.)

### :link: External links

* [Slack thread](https://hashicorp.slack.com/archives/C7KTUHNUS/p1728404760159249)

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/components/hds/app-side-nav/index.ts
+++ b/packages/components/src/components/hds/app-side-nav/index.ts
@@ -69,12 +69,14 @@ export default class HdsAppSideNav extends Component<HdsAppSideNavSignature> {
     );
   }
 
+  // controls if the component reacts to viewport changes
   get isResponsive(): boolean {
-    return this.args.isResponsive ?? true; // controls if the component reacts to viewport changes
+    return this.args.isResponsive ?? true;
   }
 
+  // controls if users can collapse the appsidenav on 'desktop' viewports
   get isCollapsible(): boolean {
-    return this.args.isCollapsible ?? false; // controls if users can collapse the appsidenav on 'desktop' viewports
+    return this.args.isCollapsible ?? false;
   }
 
   get shouldTrapFocus(): boolean {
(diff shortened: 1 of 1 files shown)


KristinLBradley:
I created a separate branch for these changes so it's easier to review apart from other AppSideNav code in the component branch.

packages/components/src/components/hds/app-side-nav/index.ts
```diff
@@ -39,6 +38,7 @@ export default class HdsAppSideNav extends Component<HdsAppSideNavSignature> {
 
   constructor(owner: unknown, args: HdsAppSideNavSignature['Args']) {
     super(owner, args);
+    this.isMinimized = this.args.isMinimized ?? false; // sets the default state on 'desktop' viewports
```
didoo, line 41: [question] I infer this has not  been converted to a getter because we don't want consumers to update dynamically its value, once the component is rendered; right?

didoo (approved):
Left a couple of minor comments. Not blockers.

packages/components/src/components/hds/app-side-nav/index.ts
KristinLBradley: It seems like it should only be set once from the args, if passed, then updated via the actions.

KristinLBradley: If it used a getter instead then it would also need a setter. (I was getting an error when I initially changed it to use a getter so decided to just set it in the constructor instead.)

Let me know if you think it would be better to use a getter & setter though.

didoo: No, makes sense. Thanks for the clarification.

PR #2487, opened 2024-10-08
`AppSideNav` - Fix bug with scrolling caused by hidden panels (HDS-3892)

KristinLBradley:
### :pushpin: Summary

If merged, this PR will prevent unwanted scrollbars from being triggered by hidden panels.
(Merges into `AppSideNav` component branch.)

[AppSideNav within framed layout](https://hds-showcase-git-hds-3892-app-sidenav-scrolling-bug-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-app-header-and-app-side-nav), notice you can still scroll all content as before.

Tested locally within Terraform/Atlas.

**Fix within Terraform/Atlas**
(_Lime green outline added around visible panel. Purple dashed outlines show layout of hidden panels. Their grid area heights are determined by the height of the visible panel._)

image

_NOTE: I tested mainly using the SideNav within Terraform. (Will continue testing with AppSideNav to confirm but should be the same.)_

### :link: External links

* Jira ticket: [HDS-3892](https://hashicorp.atlassian.net/browse/HDS-3892)

***

### 👀 Component checklist

[HDS-3892]: https://hashicorp.atlassian.net/browse/HDS-3892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/styles/components/app-side-nav/content.scss
+++ b/packages/components/src/styles/components/app-side-nav/content.scss
@@ -34,7 +34,14 @@
 }
 
 .hds-app-side-nav__content-panel {
+  // Note: height is impacted by hidden panels if max-height is not set (may not matter but this prevents it)
+  height: max-content;
   padding: 0 var(--token-app-side-nav-wrapper-padding-horizontal);
+  overflow: hidden; // the panel itself does not need to be scrollable
+
+  &[aria-hidden="true"] {
+    max-height: 250px; // prevents hidden panels from causing scrolling
+  }
 }
 
 // (LIST) TITLE
--- a/packages/components/src/styles/components/app-side-nav/content.scss
+++ b/packages/components/src/styles/components/app-side-nav/content.scss
@@ -34,13 +34,11 @@
 }
 
 .hds-app-side-nav__content-panel {
-  // Note: height is impacted by hidden panels if max-height is not set (may not matter but this prevents it)
-  height: max-content;
   padding: 0 var(--token-app-side-nav-wrapper-padding-horizontal);
   overflow: hidden; // the panel itself does not need to be scrollable
 
   &[aria-hidden="true"] {
-    max-height: 220px; // prevents hidden panels from causing scrolling
+    max-height: 0; // prevents hidden panels from causing scrolling
   }
 }
 
--- a/packages/components/src/styles/components/app-side-nav/content.scss
+++ b/packages/components/src/styles/components/app-side-nav/content.scss
@@ -40,7 +40,7 @@
   overflow: hidden; // the panel itself does not need to be scrollable
 
   &[aria-hidden="true"] {
-    max-height: 250px; // prevents hidden panels from causing scrolling
+    max-height: 220px; // prevents hidden panels from causing scrolling
   }
 }
 


KristinLBradley:
Not sure how/if I could create a useful test case for the Showcase for the fix.

didoo:
> Not sure how/if I could create a useful test case for the Showcase for the fix.

@KristinLBradley I don't know if in the showcase you can _emulate_ the behaviour of route changes in the sidenav like in a production environment. I think I tried, and I came to the conclusion is not possible with the current setup.

One thing I would test is what happens when the panels have very different heights, and they are scrolled, in the different scenarios. The reason I am saying is that by setting `max-height: 0` all of a sudden, in some cases/scenarios it may make the sidenav content jump: I remember I saw this effect when testing the possible solutions for this problem, a few days ago, hence my suggestions about having transitions for this. But I don't know if your solution has this side-effect, so I'll leave it to you to to test this condition (you have to do it in a product environment, unfortunately, unless you come up with a way to emulate this behaviour without a route change.

Below a schema (let me know if it's not clear what I mean 😀) of these possible conditions (there may be more):
screenshot_4279

Take everything with a pinch of salt, I may be completely wrong, so please test before making changes if there is such effect, and if it's worth to mitigate.

KristinLBradley:
> > Not sure how/if I could create a useful test case for the Showcase for the fix.
> 
> @KristinLBradley I don't know if in the showcase you can _emulate_ the behaviour of route changes in the sidenav like in a production environment. I think I tried, and I came to the conclusion is not possible with the current setup.
> 
> One thing I would test is what happens when the panels have very different heights, and they are scrolled, in the different scenarios. The reason I am saying is that by setting `max-height: 0` all of a sudden, in some cases/scenarios it may make the sidenav content jump: I remember I saw this effect when testing the possible solutions for this problem, a few days ago, hence my suggestions about having transitions for this. But I don't know if your solution has this side-effect, so I'll leave it to you to to test this condition (you have to do it in a product environment, unfortunately, unless you come up with a way to emulate this behaviour without a route change.
> 
> Take everything with a pinch of salt, I may be completely wrong, so please test before making changes if there is such effect, and if it's worth to mitigate.

In testing, if you've scrolled within the SidNav and then click on a link that navigates to a new panel and then go back to the previous panel then it does not maintain the position you scrolled to so you may have to scroll again to find it. However without the fix, the issues seem worse. In some cases, all the SideNav links were hidden when I navigated making it look empty and I had to scroll to see the links.

didoo:
> In testing, if you've scrolled within the SidNav and then click on a link that navigates to a new panel and then go back to the previous panel then it does not maintain the position you scrolled to so you may have to scroll again to find it. However without the fix, the issues seem worse. In some cases, all the SideNav links were hidden when I navigated making it look empty and I had to scroll to see the links.

Could we leverage something like `scrollIntoView` or `scrollTo`?

KristinLBradley:
> > In testing, if you've scrolled within the SidNav and then click on a link that navigates to a new panel and then go back to the previous panel then it does not maintain the position you scrolled to so you may have to scroll again to find it. However without the fix, the issues seem worse. In some cases, all the SideNav links were hidden when I navigated making it look empty and I had to scroll to see the links.
> 
> Could we leverage something like `scrollIntoView` or `scrollTo`?

Maybe? It's very difficult to figure out how to test everything. I'm poking around within Atlas Terraform more though to see if I can figure anything out.

With what I've added so far in this PR though I still think it's an improvement at least.

KristinLBradley:
> Thanks for investigating this, Kristin! Would you be able to test the change in HCP and Vault as well?

Tested in HCP, also tested current behavior in Vault which has the same issue.

Without the fix, if you navigate from a long list of links in the side panel to a short list of links, the side panel can appear empty until you scroll back to the top of the content (which may not be obvious that you need to do.) Once you do scroll back and then click to return to the previous screen the list of links displays as before and does not preserve the position you scrolled to within the long list of links.

With the fix, if you navigate from a long list of links to a short one, the short list of links displays as expected (i.e., you don't have to scroll to see the links). If you navigate back to the previous list of links, any position you had scrolled to previously is not preserved and the links display with the scroll position at the top which is the same as the current behavior.

So from the above, I don't see any drawback from the current fix. I think it's a worthwhile improvement as it fixes a serious usability problem which has the potential to confuse and annoy users.

alex-ju (approved):

PR #2490, opened 2024-10-09
`Dropdown` - fix content being preserved in the DOM when closed

alex-ju:
### :pushpin: Summary

Removes the Dropdown content from the DOM when closed and reverts hashicorp/design-system#2443 (removes the `[D].isOpen` argument, as it would not have any effects). This is in relation to the change of behaviour when the Dropdown is closed introduced in https://github.com/hashicorp/design-system/pull/2309.

### :link: External links

[Slack conversation](https://hashicorp.slack.com/archives/C11JCBJTW/p1728421028141439)

***

### 👀 Component checklist

### DIFF
--- a/.changeset/curly-apples-share.md
+++ b/.changeset/curly-apples-share.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Dropdown` - Fixed content being preserved in the DOM when closed
--- a/.changeset/curly-apples-share.md
+++ b/.changeset/curly-apples-share.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-`Dropdown` - Fixed content being preserved in the DOM when closed
+`Dropdown` - Fixed content being preserved in the DOM when closed and removed the `isOpen` yielded argument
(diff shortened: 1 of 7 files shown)


.changeset/curly-apples-share.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Dropdown` - Fixed content being preserved in the DOM when closed
```
didoo, line 5: [suggestion] we should also mention that we're removing the `isOpen` yielded argument
```suggestion
`Dropdown` - Fixed content being preserved in the DOM when closed and removed the `isOpen` yielded argument
```

didoo (approved):
Left only a small suggestion about the changeset.

FYI I am going to open a parallel PR to see if we can "have it both ways": maintain the old behaviour while at the same time give consumers the option to opt-in the new one via a dedicated argument. So we can compare and decide if makes sense or not.

didoo:
@alex-ju @zamoore here's the PR I mentioned above: https://github.com/hashicorp/design-system/pull/2491

zamoore (approved):

PR #2491, opened 2024-10-10
`Dropdown` - Small changes to logic for content being preserved in the DOM (on top of #2490)

didoo:
### :pushpin: Summary

_⚠️ This PR is built on top of #2490, which is still in review. ⚠️_

The intent of this PR is to propose a middle ground between the previous implementation, where the content was always present in the DOM, even when the dropdown is closed, and the changes (rollback) done in #2490, where the content is never present in the DOM when the dropdown is closed.

The idea is to give consumers a way - through a dedicated `@preserveDomContent` argument (name TBD) - to opt-in in rendering the list content also when the dropdown is closed (and potentially have a finer control of what is rendered and what is not via the `isOpen` yielded status.

In this PR I have:
- introduced `preserveDomContent` argument to control rendering of the list in the DOM (name TBD)
- added back the `isOpen` yielded status that was removed in #2490
- added back a showcase demo where we use the `@preserveDomContent` argument in combination with the `isOpen` state
- updated the documentation
- updated the changeset description

**Preview**: https://hds-showcase-git-revert-2443-hds-3871-dropdown-29f41d-hashicorp.vercel.app/components/dropdown#demo

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/components/hds/dropdown/index.ts
+++ b/packages/components/src/components/hds/dropdown/index.ts
@@ -44,6 +44,7 @@ export interface HdsDropdownSignature {
     listPosition?: HdsDropdownPositions;
     width?: string;
     enableCollisionDetection?: FloatingUIOptions['enableCollisionDetection'];
+    preserveDomContent?: boolean;
   };
   Blocks: {
     default: [
--- a/packages/components/src/components/hds/dropdown/index.ts
+++ b/packages/components/src/components/hds/dropdown/index.ts
@@ -63,6 +63,7 @@ export interface HdsDropdownSignature {
         ToggleButton?: ComponentLike<HdsDropdownToggleButtonSignature>;
         ToggleIcon?: ComponentLike<HdsDropdownToggleIconSignature>;
         close?: () => void;
+        isOpen?: boolean;
       },
     ];
   };
--- a/website/docs/components/dropdown/partials/code/component-api.md
+++ b/website/docs/components/dropdown/partials/code/component-api.md
@@ -61,6 +61,9 @@ The Dropdown component is composed of different child components each with their
     <br/><br/>
     If this function is invoked using an `\{{on "click"}}` modifier applied to the `ListItem::Interactive` element, there is a quirky behavior of the Ember `<LinkTo>` component which requires a workaround to have the events executed in the right order (this happens only if it has a `@route` argument). Read more about the issue and a possible solution [in this GitHub comment](https://github.com/hashicorp/design-system/pull/399#issuecomment-1171186772).
   </C.Property>
+  <C.Property @name="[D].isOpen" @type="boolean" @default="false" @values={{array "true" "false"}}>
+    The state of the Dropdown yielded to the content.
+  </C.Property>
   <C.Property @name="listPosition" @type="string" @values={{array "bottom-left" "bottom-right" "top-left" "top-right" }} @default="bottom-right">
     _Note: If `@enableCollisionDetection` is set, the list will automatically flip position to remain visible when near the edges of the screen regardless of the starting placement._
   </C.Property>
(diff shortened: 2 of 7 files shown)


packages/components/src/components/hds/dropdown/index.ts
```diff
@@ -44,6 +44,7 @@ export interface HdsDropdownSignature {
     listPosition?: HdsDropdownPositions;
     width?: string;
     enableCollisionDetection?: FloatingUIOptions['enableCollisionDetection'];
+    preserveDomContent?: boolean;
```
didoo, line 47: what do you think of this name? does it convey what its intent is? do you have other ideas/suggestions?

meirish:
:wave: just an outside observer here, but wondering what is the thinking here? 
That folks would opt-in to the new way of rendering? If they do, do they need to test and make sure they don't run into the resize observer error? 

And then in a 5.x release does this arg get removed?

My gut reaction to this change is that you're leaking/exposing implementation details. And in doing so you're increasing the burden on consumers to know about those implementation details. The args to the component are now not just about how the component acts or looks, but directly about about it treats yielded content when it's rendered.

meirish:
> That PR is reverting the behaviour to how it was before (undoing the "breaking change"). This PR is just giving the consumers an option to choose a different behaviour, as in to have the content _preserved_ in the DOM if they want/need, instead of destroying it every time. The new argument would remain available, it's an extra option, it's not a "fix" for something.

Sorry I may not have the details straight - I thought this was allowing users to opt-in to the behavior that's being reverted, is that not the case? If it's not, then please disregard. If it is, does that mean there won't be a major release with that change as the default, it would only be opt-in?

didoo:
> Sorry I may not have the details straight - I thought this was allowing users to opt-in to the behavior that's being reverted

In https://github.com/hashicorp/design-system/pull/2490, the "main" PR, the code is being reverted to the (old) behaviour [content is deleted from the DOM when the dropdown is closed]. This will go live no matter what, once approved/merged.

In this PR, which is a proposal to extend the API, we add an option to opt-in via a dedicated argument to the ("new") behaviour [content is always present in the DOM]. Since we yield the `isOpen` status variable, if a consumer decides to opt-in for the new behaviour, they can also decide more granularly if something is injected in the DOM only when opened, if they want.

> does that mean there won't be a major release with that change as the default, it would only be opt-in?

There will be a release surely with #2490, and maybe with this PR if approved. We are not sure yet of the semver to use for the release, we're still discussing internally what are the options.

packages/components/src/components/hds/dropdown/index.ts
zamoore: What do you think about `persistOptionsInDom`?

zamoore (approved):

packages/components/src/components/hds/dropdown/index.ts
didoo: they are not options :( 
they are not even "items" because this includes also the header and footer elements
maybe `persistDomContent`?

meirish:
> There will be a release surely with #2490, and maybe with this PR if approved. We are not sure yet of the semver to use for the release, we're still discussing internally what are the options.

I was trying to understand why the previously default behavior would now be opt-in. Are all of the benefits of the refactor (collision detection, a11y of list not being in the DOM, etc) still gained without rendering the contents when the dropdown is closed? And is that resize error still something that could happen if they do opt-in to always rendering the contents?

alex-ju:
Trying to answer a few questions here before I review it.

> I was trying to understand why the previously default behavior would now be opt-in.

The newly introduced behaviour of preserving the Dropdown content (that seems to be causing issues in products) will be opt-in, not the previous one.

> Are all of the benefits of the refactor (collision detection, a11y of list not being in the DOM, etc) still gained without rendering the contents when the dropdown is closed?

yes, we found a way to preserve all the benefits without enforcing the Dropdown content to be preserved.

> And is that resize error still something that could happen if they do opt-in to always rendering the contents?

we are planning on investigating this before the release

website/docs/components/dropdown/partials/code/component-api.md
```diff
@@ -79,8 +82,11 @@ The Dropdown component is composed of different child components each with their
   <C.Property @name="height" @type="string" @valueNote="any valid CSS height (px, rem, etc)">
     If a `@height` parameter is provided then the list will have a max-height.
   </C.Property>
-  <C.Property @name="onClose" @type="function">
-    Callback function invoked when the Dropdown is closed, if provided.
+  <C.Property @name="height" @type="string" @valueNote="any valid CSS height (px, rem, etc)">
+    If a `@height` parameter is provided then the list will have a max-height.
+  </C.Property>
```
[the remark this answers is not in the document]
didoo: 🤦‍♂️ I copy&pasted one block, but then edited the other

didoo:
> I am up for introducing a `preserveDomContent` (or `preserveContentInDom`, or however we decide to call it). Howver, I'm not convinced about re-introducing `D.isOpen`. I think it will add a lot of complexity to the Dropdown API and could be misused. Since we don't have a use case for it, I would stick to only allowing the Dropdown content to be either preserved or not.

@alex-ju I am going to close this one, and open a fresh one with just the changes related to `preserveContentInDom`.

PR #2496, opened 2024-10-11
Showcase - Added demo to `Modal` and `Flyout` pages in which the modal is opened via a `Dropdown`

didoo:
### :pushpin: Summary

While working on the `HdsPlus::ChallengeModal` one argument that is missing from our `Hds::Modal`, looking at the implementation of similar components in our consumers' codebases, is the `returnFocusTo` ([see here](https://github.com/hashicorp/cloud-ui/blob/main/addons/core/package/src/components/modal-dialog.ts#L236-L241)). I am thinking to introduce a similar functionality in our Modal component too (and potentially the Flyout). 

This is the first step, used to show how the focus is "lost" once the modal is closed: is not returned to the original element that opened the modal, since it's deleted from the DOM, so it's assigned to the `body`.

In this PR I have:
- added a small demo to `Modal` showcase page in which the modal is opened via a `Dropdown`
- added a small demo to `Flyout` showcase page in which the modal is opened via a `Dropdown`

👀  **Previews**: 
- Modal: https://hds-showcase-git-showcase-modal-add-extra-use-case-hashicorp.vercel.app/components/modal#demo
- Flyout: https://hds-showcase-git-showcase-modal-add-extra-use-case-hashicorp.vercel.app/components/flyout#demo

### 🎬 Video

You can see how the focus is returned to the `body` once the Modal is closed (look at the `document.activeElement` live variable in the devtools on the right)

https://github.com/user-attachments/assets/5ad62dc0-3bfa-40f4-a25f-548aac9e44f6

***

### 👀 Component checklist

### DIFF
--- a/showcase/app/controllers/components/flyout.js
+++ b/showcase/app/controllers/components/flyout.js
@@ -10,6 +10,7 @@ import { tracked } from '@glimmer/tracking';
 export default class FlyoutController extends Controller {
   @tracked mediumFlyoutActive = false;
   @tracked largeFlyoutActive = false;
+  @tracked dropdownInitiatedFlyoutActive = false;
 
   @action
   activateFlyout(Flyout) {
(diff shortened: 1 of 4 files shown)


alex-ju (approved):

didoo:
Related PR: https://github.com/hashicorp/design-system/pull/2497

PR #2500, opened 2024-10-11
`Alert` - don't use role or aria-live for non-valid alert usages (HDS-3919)

KristinLBradley:
_**STATUS:**_
* _Component updates have been reviewed & approved. THANKS!_
* _[Doc updates have been PRed separately](https://github.com/hashicorp/design-system/pull/2509) **PLEASE REVIEW**_

---

### :pushpin: Summary

If merged, this PR changes the `Alert` to not set a default role or aria-live value for "neutral" or "highlight" alerts.

👉 **Docs branch:** NEW - https://github.com/hashicorp/design-system/pull/2509
(OLD - ~~https://github.com/hashicorp/design-system/pull/2507~~)

* **Showcase preview:** https://hds-showcase-git-hds-3919-alert-non-alert-usag-8010ea-hashicorp.vercel.app/components/alert (_no visual changes; inspect the code for different color values_)

### :link: External links

* Jira ticket: [HDS-3919](https://hashicorp.atlassian.net/browse/HDS-3919)

***

### 👀 Component checklist

[HDS-3919]: https://hashicorp.atlassian.net/browse/HDS-3919?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KristinLBradley:
Let me know if the approach I've taken in this PR is not how we want to handle this (or any other feedback).

packages/components/src/components/hds/alert/index.ts
```diff
@@ -178,8 +155,20 @@ export default class HdsAlert extends Component<HdsAlertSignature> {
...
-    if (actions.length) {
+
+    const isRealAlert: boolean =
+      this.color === 'warning' ||
+      this.color === 'critical' ||
+      this.color === 'success';
+
+    /*
+      Q: If the alert contains actions but is not a "real alert", should it be an alertdialog? 
+      (Currently assuming it is not but don't know if that's correct)
```
KristinLBradley, line 166: @MelSumner I wasn't sure about what role Alerts should have if they are used for info or promo purposes but contain actions.

MelSumner: They should not have any role if they are used only for informational or promotional purposes. 

The difference between `alert` and `alertdialog` (as we've chosen to define it) is that if the component contains an interactive element, then it should be `alertdialog`. An element with the role of `alert` should contain string content (when flattened).

MelSumner:
Another thing to consider, if you haven't already, is how this might affect the Toast component.

KristinLBradley:
> Another thing to consider, if you haven't already, is how this might affect the Toast component.

@MelSumner Should the `Toast` behave differently than the `Alert` do you think?

i.e., for a `Toast` using `@color="neutral"` or `@color="highlight"` which is used for informational or promotional purposes, should it have `role="alert"` and `aria-live="polite"`?

majedelass:
@KristinLBradley Toasts, from a UX perspective, behave differently. For a neutral/highlight toast, a user may need to still need to be aware of it even if it's simply communicating a "process has started." Especially because a Toast is considered "temporary" (either times out or is dimissed). I do not believe anyone currently uses it for promotional purposes (nor should they ever) as they are never contextual. So I think the two are different use cases.

packages/components/src/components/hds/alert/index.ts
```diff
@@ -156,11 +137,7 @@ export default class HdsAlert extends Component<HdsAlertSignature> {
     }
   }
 
-  /**
-   * Get the class names to apply to the component.
-   * @method Alert#classNames
-   * @return {string} The "class" attribute to apply to the component.
-   */
+  // The "class" attribute to apply to the component.
```
didoo, line 140: [suggestion/non-blocker] I would leave these comments only for snippets of code for which it may not be clear at first sight what they do; in this case, it's a pretty common/widespread function.

packages/components/src/components/hds/alert/index.ts
```diff
@@ -178,8 +155,16 @@ export default class HdsAlert extends Component<HdsAlertSignature> {
     const actions = element.querySelectorAll(
       `${CONTENT_ELEMENT_SELECTOR} button, ${CONTENT_ELEMENT_SELECTOR} a`
     );
-    if (actions.length) {
+
+    const isRealAlert: boolean =
```
didoo, line 159: [thought] what does "real" mean in this context? is there a more semantic way to call this variable?

showcase/tests/integration/components/hds/alert/index-test.js
```diff
@@ -145,37 +145,144 @@ module('Integration | Component | hds/alert/index', function (hooks) {
 
   // A11Y
 
-  test('it should render with the correct semantic tags and aria attributes', async function (assert) {
-    await render(hbs`<Hds::Alert @type="inline" id="test-alert" />`);
+  // Alert and non-alert usages
```
didoo, line 148: [thought] Personally, I'm find a bit confusing this idea of "Alert non-alert"; is there a way to describe these different kind of alerts in a different way, with a different term? (something about being announced, or similar)

didoo (commented):
Left a couple of minor, non-blocking comments. Apart from that, great PR. 🙌

packages/components/src/components/hds/alert/index.ts
KristinLBradley: It means an Alert which actually alerts users (has role="alert" & aria-live="polite").

I didn't actually know what to call it and realize the name may seem a bit "weird" perhaps so I'm open to better naming ideas.

showcase/tests/integration/components/hds/alert/index-test.js
KristinLBradley: I updated comments to attempt to clarify

packages/components/src/components/hds/alert/index.ts
KristinLBradley: @alex-ju @zamoore Do you have ideas for clearer naming for this variable (if you don't think it's clear)?

MelSumner: `isSemanticAlert` maybe? Everything else I immediately thought of was funny but probably not useful.

MelSumner (approved):
This looks good to me. Nice test additions, too. I think that the boolean for the alert could be thought about a little more but I don't consider it a blocker. Naming things is hard.

packages/components/src/components/hds/alert/index.ts
KristinLBradley: @MelSumner Well both you and Cristiano mentioned the word "semantic" so maybe I'll just change it to that. Thanks!

KristinLBradley: Updated it. Thanks for the feedback.

PR #2503, opened 2024-10-14
`Dropdown` - Small changes to logic for content being preserved in the DOM (on top of #2490)

didoo:
### :pushpin: Summary

_⚠️ This PR is built on top of #2490, which is still in review. ⚠️_

The intent of this PR is to propose a middle ground between the previous implementation, where the content was always present in the DOM, even when the dropdown is closed, and the changes (rollback) done in #2490, where the content is never present in the DOM when the dropdown is closed.

The idea is to give consumers a way - through a dedicated `@ preserveContentInDom ` argument - to opt-in in rendering the list content also when the dropdown is closed.

_Notice: in a previous iteration (see https://github.com/hashicorp/design-system/pull/2491) we also reintroduced the yielded `isOpen` status, but in review feedback was decided to remove it, hence this new PR._

In this PR I have:
- introduced `preserveContentInDom` argument to control rendering of the list in the DOM
- added a showcase demo where we use the `@preserveContentInDom` argument
- updated the changeset description

**Preview**: TODO

***

### 👀 Component checklist

### DIFF
--- a/.changeset/curly-apples-share.md
+++ b/.changeset/curly-apples-share.md
@@ -2,4 +2,8 @@
 "@hashicorp/design-system-components": patch
 ---
 
-`Dropdown` - Fixed content being preserved in the DOM when closed and removed the `isOpen` yielded argument
+`Dropdown`
+
+- Fixed content being preserved in the DOM when closed
+- Removed the `isOpen` yielded argument
+- Added `@preserveContentInDom` to optionally control rendering of the content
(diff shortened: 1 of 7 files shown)


alex-ju (approved):
Looking good! I believe this will help us migrate instances where consumers need to do extra work to preserve (or bring back) the Dropdown content

alex-ju:
hey @zamoore – I saw you're tagged for review. I'll merge this in so I can test a pre-release, but feel free to leave comments after your review and we can open a follow-up if needed

PR #2506, opened 2024-10-15
`Dropdown` - fix dropdown content not being preserved when interacted with

alex-ju:
### :pushpin: Summary

Fixed an issue where the dropdown content was not preserved when interacted with.

The PopoverPrimitive assumes the popover element is focusable and [registers a `focusout` event](https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/popover-primitive/index.ts#L86-L88) on the main popover container to determine when to hide the container. When the popover element is not focusable, any interaction will trigger a `focusout` and close the popover. To resolve this we make the popover container focusable using `tabindex="-1"`.

### :link: External links

Jira ticket: [HDS-3994](https://hashicorp.atlassian.net/browse/HDS-3994)

***

### 👀 Component checklist

[HDS-3994]: https://hashicorp.atlassian.net/browse/HDS-3994?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/dropdown/index.hbs
+++ b/packages/components/src/components/hds/dropdown/index.hbs
@@ -13,6 +13,7 @@
       )
     }}
     <div
+      tabindex="-1"
       class={{this.classNamesContent}}
       {{style width=@width max-height=@height}}
       {{PP.setupPrimitivePopover anchoredPositionOptions=this.anchoredPositionOptions}}
--- a/.changeset/seven-taxis-develop.md
+++ b/.changeset/seven-taxis-develop.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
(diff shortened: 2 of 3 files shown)


packages/components/src/components/hds/dropdown/index.hbs
```diff
@@ -13,6 +13,7 @@
       )
     }}
     <div
+      tabindex="-1"
```
didoo, line 16: That's what I noticed the other day when doing the code review, that this `tabindex` was missing: https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/menu-primitive/index.hbs#L16 

(but then I thought we would have noticed if it was important, so I didn't say anything; well, we noticed 🤦‍♂️)

PR #2511, opened 2024-10-18
Fixing Error in `Hds::Flyout` & `Hds::Modal` subcomponents

zamoore:
### :pushpin: Summary

Fixes an error in the deprecated `HdsDialogPrimitive::Header/Body/Description/Footer` components as well as the `Hds::Modal::Body` component. Passes `args` from constructor call to `super` within constructor.

I discovered an error caused by the recent TS conversion. When we converted the `Hds::Flyout::Body` component, we incorrectly passed `{}` as the `args` argument to `super` in the constructor, when we should use the argument that is passed from the constructor. To make that clearer, here's an example:

Code causing error:
```
export interface Signature {
  Blocks: {
    default: [];
  };
  Element: HTMLDivElement;
}

export default class MyComponent extends Component<Signature> {
  constructor(owner: unknown) {
    // object is passed to super, args not utilized from constructor
    super(owner, {});
  }
}
```

Should be
```
export interface Signature {
  Blocks: {
    default: [];
  };
  Element: HTMLDivElement;
}

export default class MyComponent extends Component<Signature> {
  constructor(owner: unknown, args: Record<string, never>) {
    super(owner, args);
  }
}
```

### :link: External links

[Slack discussion](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1729279817534249)

***

### DIFF
--- a/.changeset/dull-lamps-retire.md
+++ b/.changeset/dull-lamps-retire.md
@@ -2,10 +2,8 @@
 "@hashicorp/design-system-components": minor
 ---
 
-`Hds::DialogPrimitive`
-- Fixed error in `Description` and `Body` subcomponents, by not passing the `args` argument from the constructor to `super` 
-- Added missing arguments in `Arg` TypeScript signature object
+`Hds::Flyout`
+- Fixed error in `Description` and `Body` subcomponents, caused by not passing the `args` argument from the constructor to `super` 
 
 `Hds::Modal`
 - Fixed error in `Body` subcomponent, caused by not passing the `args` argument from the constructor to `super`
-- Added missing arguments in `Arg` TypeScript signature object
--- a/.changeset/dull-lamps-retire.md
+++ b/.changeset/dull-lamps-retire.md
@@ -2,7 +2,10 @@
 "@hashicorp/design-system-components": minor
 ---
 
-`Hds::DialogPrimitive::Description`
-`Hds::DialogPrimitive::Body`
-`Hds::Modal::Body`
-- Fixed error caused by not passing the `args` argument from the constructor to `super`
+`Hds::DialogPrimitive`
+- Fixed error in `Description` and `Body` subcomponents, by not passing the `args` argument from the constructor to `super` 
+- Added missing arguments in `Arg` TypeScript signature object
+
+`Hds::Modal`
+- Fixed error in `Body` subcomponent, caused by not passing the `args` argument from the constructor to `super`
+- Added missing arguments in `Arg` TypeScript signature object
--- a/.changeset/dull-lamps-retire.md
+++ b/.changeset/dull-lamps-retire.md
@@ -2,6 +2,7 @@
 "@hashicorp/design-system-components": minor
 ---
 
-`Hds::DialogPrimitive::Description` 
-`Hds::DialogPrimitive::Body` 
+`Hds::DialogPrimitive::Description`
+`Hds::DialogPrimitive::Body`
+`Hds::Modal::Body`
 - Fixed error caused by not passing the `args` argument from the constructor to `super`
--- a/.changeset/dull-lamps-retire.md
+++ b/.changeset/dull-lamps-retire.md
@@ -1,5 +1,5 @@
 ---
-"@hashicorp/design-system-components": minor
+"@hashicorp/design-system-components": patch
 ---
 
 `Hds::Flyout`
--- a/.changeset/dull-lamps-retire.md
+++ b/.changeset/dull-lamps-retire.md
@@ -1,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Hds::DialogPrimitive::Description` 
+`Hds::DialogPrimitive::Body` 
+- Fixed error caused by not passing the `args` argument from the constructor to `super`
--- a/packages/components/src/components/hds/flyout/body.ts
+++ b/packages/components/src/components/hds/flyout/body.ts
@@ -7,9 +7,7 @@ import Component from '@glimmer/component';
 import { deprecate } from '@ember/debug';
 
 export interface HdsFlyoutBodySignature {
-  Args: {
-    contextualClass?: string;
-  };
+  Args: never;
   Blocks: {
     default: [];
   };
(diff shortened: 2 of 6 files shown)


packages/components/src/components/hds/flyout/body.ts
```diff
@@ -7,15 +7,18 @@ import Component from '@glimmer/component';
 import { deprecate } from '@ember/debug';
 
 export interface HdsFlyoutBodySignature {
+  Args: {
+    contextualClass?: string;
+  };
```
didoo, line 12: [praise] 🙌

.changeset/dull-lamps-retire.md
```diff
@@ -0,0 +1,8 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Hds::DialogPrimitive::Description`
+`Hds::DialogPrimitive::Body`
+`Hds::Modal::Body`
+- Fixed error caused by not passing the `args` argument from the constructor to `super`
```
didoo, line 8: [suggestion] should we split this in multiple entries, per component?

```suggestion
`Hds::DialogPrimitive`
- Fixed error in `Description` and `Body` subcomponents, by not passing the `args` argument from the constructor to `super` 
- Added missing arguments in `Arg` TypeScript signature object

`Hds::Modal`
- Fixed error in `Body` subcomponent, caused by not passing the `args` argument from the constructor to `super`
- Added missing arguments in `Arg` TypeScript signature object
```

didoo (approved):
Left a minor suggestion (feel free to ignore or edit as you see fit).

packages/components/src/components/hds/flyout/footer.ts
```diff
@@ -8,6 +8,7 @@ import { deprecate } from '@ember/debug';
 
 export interface HdsFlyoutFooterSignature {
   Args: {
+    contextualClass?: string;
```
alex-ju, line 11: thought: we should probably add this to `hds/modal/footer.ts` as well

packages/components/src/components/hds/flyout/body.ts
alex-ju: question: hm.. the `hds/flyout/body` doesn't have or make use of any `contextualClass`; `hds/dialog-primitive/body` has a `contextualClass`. unless I'm missing something all these signature additions are not needed in `flyout` and `modal`.

alex-ju (commented):
the constructor changes look good. the signature adjustments in flyout/modal subcomponents (the deprecated body, description, footer, header) are not needed. the `contextualClass`es are only used on the `dialog-primitive` subcomponents

packages/components/src/components/hds/flyout/body.ts
zamoore: We are passing `contextualClass` in as an argument in the index.hbs for those components.

```
  <:body>
    {{yield (hash Body=(component "hds/dialog-primitive/body" contextualClass="hds-modal__body"))}}
  </:body>
  ```
  
If we're not using those classes we can either remove them, or actually implement them. If they're not doing anything, I would lean toward removing them.

File: packages/components/src/components/hds/flyout/body.ts [hunk 4]
```diff
@@ -7,15 +7,16 @@ import Component from '@glimmer/component';
 import { deprecate } from '@ember/debug';
 
 export interface HdsFlyoutBodySignature {
+  Args: never;
```
didoo, line 10: [question] as a pattern, do we prefer to have `Args` completely missing (I think I have done something like this, recently), or `Args: never` like here? which one is more "correct™"?

/cc @alex-ju

didoo: Or is it because below we need to have a `HdsFlyoutBodySignature['Args']`?

alex-ju: > [question] as a pattern, do we prefer to have Args completely missing (I think I have done something like this, recently), or Args: never like here? which one is more "correct™"?

not having `Args` is the established pattern. however, considering [the error we encounter in HCP](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1729279817534249) for this subcomponent, I don't see a way of avoiding `never`.

we could use `never` straight in the constructor, but we have this pattern of using the signature in constructors, so what we have in this PR feels more consistent.

```js
constructor(owner: unknown, args: never) {
```

.changeset/dull-lamps-retire.md
```diff
@@ -0,0 +1,11 @@
...
+"@hashicorp/design-system-components": minor
+---
+
+`Hds::DialogPrimitive`
+- Fixed error in `Description` and `Body` subcomponents, by not passing the `args` argument from the constructor to `super` 
+- Added missing arguments in `Arg` TypeScript signature object
+
+`Hds::Modal`
+- Fixed error in `Body` subcomponent, caused by not passing the `args` argument from the constructor to `super`
+- Added missing arguments in `Arg` TypeScript signature object
```
alex-ju, line 11: The changelog would need to be updated, as we're not making any changes to the `DialogPrimitive` and there are no missing arguments anymore.

packages/components/src/components/hds/flyout/body.ts
didoo: > not having Args is the established pattern

@alex-ju Good. So nothing to change on my side :) 

> however, considering [the error we encounter in HCP](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1729279817534249) for this subcomponent, I don't see a way of avoiding never.

@alex-ju @zamoore since this is a deviation from the pattern, should we add a comment above these `Args: never` are not following the established pattern? (and potentially have also a link to this thread, for future reference)

zamoore: I can add a comment, but I think this isn't as much as a deviation as a new rule in the pattern. If the component utilizes the constructor, but also does not take any args, define the type for args as `never`.

alex-ju (approved):

.changeset/dull-lamps-retire.md
```diff
@@ -0,0 +1,9 @@
+---
+"@hashicorp/design-system-components": minor
```
alex-ju, line 2:
```suggestion
"@hashicorp/design-system-components": patch
```

PR #2513, opened 2024-10-21
Fix Dropdown-related deprecation warnings in `website` and `showcase`

alex-ju:
### :pushpin: Summary

Replace the deprecated `@text` argument in `Hds::Dropdown::ListItem::Interactive` to silence warnings.
Alternatively, we could've updated the `deprecation-workflow.js` to silence these warnings until fixed.

### :link: External links

Jira ticket: [HDS-3999](https://hashicorp.atlassian.net/browse/HDS-3999)

***

[HDS-3999]: https://hashicorp.atlassian.net/browse/HDS-3999?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/config/deprecation-workflow.js
+++ b/website/config/deprecation-workflow.js
@@ -11,6 +11,15 @@ self.deprecationWorkflow.config = {
     { handler: "silence", matchId: "remove-owner-inject" },
     { handler: "silence", matchId: "ember-modifier.function-based-options" },
     { handler: "throw", matchId: "deprecate-auto-location" },
-    { handler: "silence", matchId: "ember-string.add-package" }
+    { handler: "silence", matchId: "ember-string.add-package" },
+    { handler: "throw", matchId: "hds.dropdown.list-item.interactive" },
+    { handler: "throw", matchId: "hds.components.flyout.body" },
+    { handler: "throw", matchId: "hds.components.flyout.description" },
+    { handler: "throw", matchId: "hds.components.flyout.footer" },
+    { handler: "throw", matchId: "hds.components.flyout.header" },
+    { handler: "throw", matchId: "hds.components.modal.body" },
+    { handler: "throw", matchId: "hds.components.modal.footer" },
+    { handler: "throw", matchId: "hds.components.modal.header" },
+    { handler: "throw", matchId: "hds.components.sidenav.header.iconbutton" },
   ]
 };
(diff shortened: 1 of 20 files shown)


Dhaulagiri:
is there any value in us adding these as `throw` in `deprecation-workflow.js` to avoid us re-adding any back, or is the risk of that effectively already zero?

alex-ju:
> is there any value in us adding these as `throw` in `deprecation-workflow.js` to avoid us re-adding any back, or is the risk of that effectively already zero?

we could. however, from my checks, `throw` logs a console error that doesn't get picked up by our tests. so it would cause errors for end users, with no benefit to our automation testing

Dhaulagiri:
Huh, that sounds undesirable (and also not expected).

shleewhite (approved):
Thanks for fixing this!

alex-ju:
> > is there any value in us adding these as `throw` in `deprecation-workflow.js` to avoid us re-adding any back, or is the risk of that effectively already zero?
> 
> we could. however, from my checks, `throw` logs a console error that doesn't get picked up by our tests. so it would cause errors for end users, with no benefit to our automation testing

I went back to this and did a bit more digging and it seems like I was partly wrong in my initial exploration.

Adding `{ handler: "throw", matchId: "hds.dropdown.list-item.interactive" }` in `showcase/config/deprecation-workflow.js` does fail tests. However, we can't do this, because we'd still [need a test to make sure the `@text` works as expected](https://github.com/hashicorp/design-system/blob/main/showcase/tests/integration/components/hds/dropdown/list-item/interactive-test.js#L98) (even if deprecated). Similarly for other deprecated subcomponents.

Adding `{ handler: "throw", matchId: "hds.dropdown.list-item.interactive" }` in `website/config/deprecation-workflow.js` does cause tests to fail (or timeout), so I added 8faad90 to prevent us from introducing deprecated subcomponents in our documentation website.

I'm going to go ahead and merge this, but let me know if you think any follow-up work might be valuable.

website/config/deprecation-workflow.js
```diff
@@ -11,6 +11,15 @@ self.deprecationWorkflow.config = {
     { handler: "silence", matchId: "remove-owner-inject" },
     { handler: "silence", matchId: "ember-modifier.function-based-options" },
     { handler: "throw", matchId: "deprecate-auto-location" },
-    { handler: "silence", matchId: "ember-string.add-package" }
+    { handler: "silence", matchId: "ember-string.add-package" },
+    { handler: "throw", matchId: "hds.dropdown.list-item.interactive" },
```
alex-ju, line 15: this change revealed the dropdown id doesn't follow convention (missing `.components`), but I might address this in a separate PR

PR #2522, opened 2024-10-28
`Dropdown`: fix height for small variant

shleewhite:
### :pushpin: Summary

Currently, the small (non-icon) variant of Dropdown is 30px tall when it should be 28px. This was because the chevron icon was 16px tall instead of 12px.

This was because the styles to make the icon 12px were only applied to the chevron icon in the ToggleIcon Dropdown. To solve this, I also added those styles to the ToggleButton.

Showcase: https://hds-showcase-git-hds-3996-dropdown-size-hashicorp.vercel.app/components/dropdown

New dimensions for both icon only toggle and toggle with text: 

Screenshot 2024-10-28 at 2 43 29 PM
Screenshot 2024-10-28 at 2 43 42 PM

### :link: External links

Jira ticket: [HDS-3996](https://hashicorp.atlassian.net/browse/HDS-3996)

***

### 👀 Component checklist

[HDS-3996]: https://hashicorp.atlassian.net/browse/HDS-3996?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/violet-donuts-matter.md
+++ b/.changeset/violet-donuts-matter.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Dropdown`: fix the height of the chevron in `ToggleButton`.
--- a/.changeset/violet-donuts-matter.md
+++ b/.changeset/violet-donuts-matter.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-`Dropdown`: fix the height of the chevron in `ToggleButton`.
+`Dropdown`: fixed the height of the chevron in `ToggleButton`.
--- a/packages/components/src/styles/components/dropdown.scss
+++ b/packages/components/src/styles/components/dropdown.scss
@@ -136,6 +136,12 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
 
 .hds-dropdown-toggle-button--size-small {
   padding-right: 0.375rem;
+
+  // For small variant with chevron, force the icon size to custom (even if the SGV size is `16px`)
+  .hds-icon {
+    width: 12px;
+    height: 12px;
+  }
 }
 
 .hds-dropdown-toggle-button--size-medium {


shleewhite:
Looks like the AppFooter percy screenshot changed because there is a Dropdown in one of the examples.

jorytindall (approved):
More simple solution than I was expecting! Thanks for the quick update on this :shipit:

.changeset/violet-donuts-matter.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Dropdown`: fix the height of the chevron in `ToggleButton`.
```
didoo, line 5: [suggestion] conventionally, we use past tenses in these changeset comments
```suggestion
`Dropdown`: fixed the height of the chevron in `ToggleButton`.
```

didoo (approved):
👍

packages/components/src/styles/components/dropdown.scss
```diff
@@ -136,6 +136,12 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
 
 .hds-dropdown-toggle-button--size-small {
   padding-right: 0.375rem;
+
+  // For small variant with chevron, force the icon size to custom (even if the SGV size is `16px`)
+  .hds-icon {
```
alex-ju, line 141: It would be wiser for us to limit this override to the toggle chevron, otherwise it will affect any icon in the dropdown toggle (e.g. the `@icon`, an icon in the badge, both currently set at `0.75rem`, which coincidentally translates to `12px`, but we can't rely on that from a systems perspective).
```suggestion
  .hds-dropdown-toggle-chevron .hds-icon {
```

alex-ju (changes requested):
I think what we have in this PR might be a bit too loose for our needs.

packages/components/src/styles/components/dropdown.scss
didoo: My bad I missed this, I was tricked by the fact that the parent selector `.hds-dropdown-toggle-button--size-small` so I thought it was safe to use just `.hds-icon`, but I forgot that the Dropdown toggle looks like a Button but is not a Button, and can contain more things 🤦‍♂️

(that said, the badge is always small (so the badge icon is `12px` anyway) and the toggle icon is also `12px` when small, so visually nothing changed, and that's why probably the VRT tests didn't fail)

alex-ju: yes not a big deal, really. a fix was added in #2533

PR #2518, opened 2024-10-25
`AppSideNav` - deprioritize adoption

KristinLBradley:
### :pushpin: Summary

Updates following memo: https://hermes.hashicorp.services/document/1kA2xrhm3HciF8UkOMcPEqsBvUP45Dh8LTt31MgdtpOY

If merged, this PR will revert deprecation of the SideNav and de-publicize components related to the Enterprise Nav project.

**Related:**

* Component branch: https://github.com/hashicorp/design-system/pull/2384

### :hammer_and_wrench: Detailed description

* ✅ Revert deprecation of `SideNav`
* ✅ Unlink documentation for AppSideNav
* ✅ Unlink documentation for AppHeader
* ✅ Add banner to AppSideNav docs advising consumers not to use it at this time.
* ✅ Add banner to AppHeader docs advising consumers not to use it at this time.
* ✅ Comment out links to AppSideNav & AppHeader Showcase pages and examples within them.
* ✅ Update rollup config so AppSideNav & AppHeader don't get re-exported into consumer apps
* ✅ Update Showcase links for AppSideNav & AppHeader to section them off from other links & update routing
* ✅ Comment out framed examples which include AppHeader and AppSideNav
* ✅ In SideNav docs next to illustrations & video which show AppHeader, update to remove them
* ✅ Delete changelog

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- ~~[ ] A changelog entry was added via [Changesets]~~(https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #2523, opened 2024-10-29
Showcase - Small fixes for `Shw::Grid` component

didoo:
### :pushpin: Summary

This small PR fixes a couple of issues spotted in these failing tests:
https://github.com/hashicorp/design-system/actions/runs/11525889346/job/32089039125
that were mentioned in this comment:
https://github.com/hashicorp/design-system/pull/2518#issuecomment-2443801163

In this PR I have:
- made the `gap` argument optional in the `Shw::Grid` signature
- added support for 1-column layout in `Shw::Grid` component

***

### 👀 Component checklist

### DIFF
--- a/showcase/app/components/shw/grid/index.gts
+++ b/showcase/app/components/shw/grid/index.gts
@@ -20,7 +20,7 @@ interface ShwGridSignature {
   Args: {
     columns: 2 | 3 | 4 | 5 | 6 | 7;
     forceMinWidth?: boolean;
-    gap: string;
+    gap?: string;
     grow?: boolean;
     label?: string;
   };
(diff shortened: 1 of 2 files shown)


alex-ju (approved):
Thanks! These were the only `Shw::Grid`-related errors I've seen in #2518. I'm still not sure why these are not being picked up in `app-header.hbs`, likely `glint` still doesn't work as expected in `showcase`.

alex-ju:
> > `glint` still doesn't work as expected in `showcase`
> 
> Yes, something we should look into. If we can't spot errors in the typing of our components in the showcase, it means these errors will end up in released code, potentially hindering/blocking successive adoption.

indeed, it happened in the past. we fixed some of the configuration, but it seems like not all of it. I captured this in [HDS-4090(]https://hashicorp.atlassian.net/browse/HDS-4090)

PR #2524, opened 2024-10-29
`AppFrame` - Add support for `sticky-footer`

didoo:
> [!IMPORTANT]
> Update the version number in the documentation before merging

### :pushpin: Summary

While working on https://hashicorp.atlassian.net/browse/HDS-3658  we’ve come to the realization that to have a proper responsive layout for the HDS+ `StickyFooter` component/pattern (see https://github.com/hashicorp/hds-plus/pull/31), that at the same time it didn’t come with the accessibility violations of the current implementation in Cloud UI, we had to extend the HDS `AppFrame` so that it would expose a dedicated `sticky-footer` slot/grid area, with “sticky” behaviour, to contain the HDS+ `StickyFooter` component (which in turn would become more agnostic about its position, size and location).

This effort is supported by an official RFC (to be published soon) to collect feedback from the HDS Engineers as well as from the HDS stakeholders.

In this PR I have:

- added `StickyFooter` container to `AppFrame`
- updated integration tests for `AppFrame`
- updated showcase for `AppFrame` with sticky-footer related examples
    - added Ember Stargate to the template registry
- updated the documentation for the `AppFrame` to include the sticky footer
- updated version history for `AppFrame`

### 👀 Previews:

**Showcase** 

- AppFrame page: https://hds-showcase-git-app-frame-sticky-footer-hashicorp.vercel.app/layouts/app-frame
- Frameless page: https://hds-showcase-git-app-frame-sticky-footer-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-content-with-sticky-footer

**Documentation**

- Code page: https://hds-website-git-app-frame-sticky-footer-hashicorp.vercel.app/layouts/app-frame#sticky-footer-container

https://github.com/user-attachments/assets/a04d1ecb-8d41-475b-9114-9b7ed1c68dde

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-4081

RFC document: TODO

***

### 👀 Component checklist


jorytindall (commented):
I don't have a problem with the implementation for this (it's nice to support it directly in this component!), but I don't think we have alignment on the layout strategy for the StickyFooter. A couple of thoughts:
- At a minimum I think I would expect the the StickyFooter to render above the `AppFooter`, though I know this is tricky with the implementation. The current stacking is confusing from a hierarchy perspective.
- I think we need a broader discussion around whether continuing to include the `AppFooter` when the `StickyFooter` is present is the right path. @evankerrigan brings up some valid points, are we able to get some clarity on whether or not excluding it will result in an accessibility violation, and what that violating criteria is? It may be that either direction we choose has downsides for usability and experience, but it would be helpful to better understand the violating criteria if that is the case.

didoo:
> At a minimum I think I would expect the the StickyFooter to render above the AppFooter, though I know this is tricky with the implementation. The current stacking is confusing from a hierarchy perspective.

@jorytindall do you mean like in this screenshot?
screenshot_4373
in that case don't worry, it's only artificial, because of how we override some layout properties of the AppFrame to be able to show it in the showcase page; in reality it will work like this: https://hds-showcase-git-app-frame-sticky-footer-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-content-with-sticky-footer

Plus, in the DOM the order is "correct": main, sticky-footer, footer.

> I think we need a broader discussion around whether continuing to include the AppFooter when the StickyFooter is present is the right path. @evankerrigan brings up some valid points, are we able to get some clarity on whether or not excluding it will result in an accessibility violation, and what that violating criteria is? It may be that either direction we choose has downsides for usability and experience, but it would be helpful to better understand the violating criteria if that is the case.

Can you raise this point in the RFC once it's published (soon), so we can have a public conversation around this? I too have mixed feelings about it, but since "we build the Helios Design System with accessibility as a core requirement" ([ref](https://helios.hashicorp.design/about/accessibility-statement)) there should be a common conversation and decision around this "hot" topic.

packages/components/src/components/hds/app-frame/index.ts
```diff
@@ -53,6 +56,11 @@ export default class HdsAppFrame extends Component<HdsAppFrameSignature> {
     return this.args.hasModals ?? true;
   }
 
+  // Indicates if the "sticky footer" container should be displayed
+  get hasStickyFooter(): boolean {
+    return this.args.hasStickyFooter ?? true;
```
philrenaud, line 61: `false` might be a more sensible default here, given this is a new component.

packages/components/src/styles/components/app-frame.scss
```diff
@@ -65,6 +66,20 @@
...
 }
 
+// STICKY FOOTER "CONTAINER"
+
+.hds-app-frame__sticky-footer {
+  position: sticky;
+  bottom: 0;
+  grid-area: sticky-footer;
+
+  // note: this doesn't alwasy work, because of the whitespace issues in Ember
```
philrenaud, line 76:
```suggestion
  // note: this doesn't always work, because of the whitespace issues in Ember
```

philrenaud (approved):

packages/components/src/styles/components/app-frame.scss
```diff
@@ -12,8 +12,9 @@
   grid-template-areas:
     "header header"
     "sidebar main"
-    "sidebar footer";
-  grid-template-rows: auto 1fr auto;
+    "sidebar footer"
+    "sidebar sticky-footer";
```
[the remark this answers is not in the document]
didoo: > If there's a possibility that this grid will eventually have gap properties

@philrenaud I don't see this as a possibility. If spacing will be added, (**if**), it will be in the "main" content area
- or in the form of internal padding (I doubt it, but may as well be if we decide to have standard, responsive padding for all the consumers, at page level) 
- or delegated to a dedicated component (imagine like a page-level "main" layout, similar to what it's done in cloud UI with the `<Core::Layout::Main>` component.

packages/components/src/components/hds/app-frame/index.ts
alex-ju: thought: I tend to agree with Phil here, maybe it should be an opt-in. although I see how it will not harm if it's true but doesn't have content

showcase/app/templates/layouts/app-frame/frameless/demo-full-app-frame-content-with-sticky-footer.hbs
```diff
@@ -0,0 +1,151 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+{{page-title "AppFrame Component - Frameless"}}
+
+<Hds::AppFrame @hasHeader={{false}} as |Frame|>
```
alex-ju, line 8: noting that if #2527 gets merged first this could use `Mock::App`

showcase/types/global.d.ts
```diff
@@ -16,8 +18,13 @@ declare module '@glint/environment-ember-loose/registry' {
   export default interface Registry
     extends EmberStyleModifierRegistry,
       EmberTruthRegistry,
+      EmberStargateRegistry,
       HdsComponentsRegistry,
       PageTitle,
       RenderModifiersRegistry,
-      ShowcaseTemplateRegistry {}
+      ShowcaseTemplateRegistry {
+    // we have to use `ember-stargate` version `0.4.3` because version `0.5.0` causes a break in the tests for `ember-lts-3.28`
```
alex-ju, line 26: thought: it looks like 0.4.x is common across products as well

.changeset/fast-rabbits-think.md
```diff
@@ -0,0 +1,8 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`AppFrame`
+
+- Added support for `sticky-footer` slot in the grid layout
```
alex-ju, line 7: thought: not sure the CSS grid is part of the AppFrame API, but `@hasStickyFooter` should be mentioned, as it's part of the API
```suggestion
- Added `@hasStickyFooter`to control the rendering of `sticky-footer` container
```

website/docs/layouts/app-frame/index.md
```diff
@@ -15,5 +15,6 @@ navigation:
 </section>
 
 <section data-tab="Version history">
+  @include "partials/version-history/4.15.0.md"
```
alex-ju, line 18: thought: next is 4.14.0; unless you would rather have this published afterward/separately?
```suggestion
  @include "partials/version-history/4.15.0.md"
```

alex-ju (commented):
The code looks solid! My only concern here is around the focus order not matching the visual order. Let's see what comes out of the feedback on the RFC. As I read through the comments it sounds like one option currently considered by designers is to not have the main footer when the sticky footer is present.

website/docs/layouts/app-frame/index.md
didoo: I didn't know if and when it was going to be released, so I tried to be conservative. I'll update before merging, if approved.

didoo:
Since there is no consensus around any of the solutions proposed in the related RFC (hence considered not approved), I am closing this PR. We will reconsider the problem in the future, if necessary.

PR #2530, opened 2024-10-31
`Dropdown` - add `matchToggleWidth`

alex-ju:
### :pushpin: Summary

Adds a `@matchToggleWidth` argument to the Dropdown component.

We found instances in our products where the Dropdown list should match the toggle width. Although not documented or recommended, this was previously achieved by setting `@width="100%"`, as the list was positioned relative to the outermost element of this component. After the `Dropdown` refactor to use the `PopoverPrimitive` the list can only be positioned relative to `document` (as any `popover` element). Setting `@width="100%"` in this case results in a Dropdown list that overflows the width of the page.

To address this use case we introduce a `@matchToggleWidth` argument that configures Floating UI (the underlying library handling positioning) to replicate the toggle width in the dropdown list.

**Showcase**
 - [Examples](https://hds-showcase-pvyizomzz-hashicorp.vercel.app/components/dropdown#width)

**Website**
 - [How to use](https://hds-website-fhftzpgzs-hashicorp.vercel.app/components/dropdown?tab=code#list-size)
 - [API](https://hds-website-fhftzpgzs-hashicorp.vercel.app/components/dropdown?tab=code#dropdown)

Test in `atlas`

Screenshot 2024-11-06 at 16 14 48

### :link: External links

Jira ticket: [HDS-4093](https://hashicorp.atlassian.net/browse/HDS-4093)

***

### 👀 Component checklist

[HDS-4093]: https://hashicorp.atlassian.net/browse/HDS-4093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/dropdown/index.ts
+++ b/packages/components/src/components/hds/dropdown/index.ts
@@ -143,7 +143,7 @@ export default class HdsDropdown extends Component<HdsDropdownSignature> {
     // context: https://github.com/hashicorp/design-system/pull/2309#discussion_r1706941892
     classes.push(`hds-dropdown__content--position-${this.listPosition}`);
 
-    // add a class based on the @width argument
+    // add a class based on the @width or @matchToggleWidth arguments
     if (this.args.width || this.args.matchToggleWidth) {
       classes.push('hds-dropdown__content--fixed-width');
     }
--- a/packages/components/src/components/hds/dropdown/index.ts
+++ b/packages/components/src/components/hds/dropdown/index.ts
@@ -45,6 +45,7 @@ export interface HdsDropdownSignature {
     width?: string;
     enableCollisionDetection?: FloatingUIOptions['enableCollisionDetection'];
     preserveContentInDom?: boolean;
+    matchToggleWidth?: boolean;
   };
   Blocks: {
     default: [
@@ -93,10 +94,15 @@ export default class HdsDropdown extends Component<HdsDropdownSignature> {
     return this.args.enableCollisionDetection ?? false;
   }
 
+  get matchToggleWidth(): FloatingUIOptions['matchToggleWidth'] {
+    return this.args.matchToggleWidth ?? false;
+  }
+
   get anchoredPositionOptions(): {
     placement: FloatingUIOptions['placement'];
     offsetOptions: FloatingUIOptions['offsetOptions'];
     enableCollisionDetection: FloatingUIOptions['enableCollisionDetection'];
+    matchToggleWidth: FloatingUIOptions['matchToggleWidth'];
   } {
     // custom options specific for the `RichTooltip` component
     // for details see the `hds-anchored-position` modifier
@@ -104,6 +110,7 @@ export default class HdsDropdown extends Component<HdsDropdownSignature> {
       placement: HdsDropdownPositionToPlacementValues[this.listPosition],
       offsetOptions: 4,
       enableCollisionDetection: this.enableCollisionDetection ? 'flip' : false,
+      matchToggleWidth: this.matchToggleWidth,
     };
   }
 
@@ -137,7 +144,7 @@ export default class HdsDropdown extends Component<HdsDropdownSignature> {
     classes.push(`hds-dropdown__content--position-${this.listPosition}`);
 
     // add a class based on the @width argument
-    if (this.args.width) {
+    if (this.args.width || this.args.matchToggleWidth) {
       classes.push('hds-dropdown__content--fixed-width');
     }
 
--- a/showcase/app/styles/showcase-pages/dropdown.scss
+++ b/showcase/app/styles/showcase-pages/dropdown.scss
@@ -23,7 +23,7 @@ body.components-dropdown {
 
   .shw-component-dropdown-fixed-height-container {
     min-height: 200px;
-    outline: 1px dotted #d3d3d3;
+    outline: 1px dotted var(--shw-color-gray-400);
   }
 
   .shw-component-dropdown-states-matrix {
--- a/showcase/app/styles/showcase-pages/dropdown.scss
+++ b/showcase/app/styles/showcase-pages/dropdown.scss
@@ -23,6 +23,7 @@ body.components-dropdown {
 
   .shw-component-dropdown-fixed-height-container {
     min-height: 200px;
+    outline: 1px dotted #d3d3d3;
   }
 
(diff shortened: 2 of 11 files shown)


packages/components/src/components/hds/dropdown/index.ts
```diff
@@ -93,17 +94,23 @@ export default class HdsDropdown extends Component<HdsDropdownSignature> {
     return this.args.enableCollisionDetection ?? false;
   }
 
+  get matchToggleWidth(): FloatingUIOptions['matchToggleWidth'] {
```
zamoore, line 97: I really like this ts pattern for default arg value getters.

showcase/app/styles/showcase-pages/dropdown.scss
```diff
@@ -23,6 +23,7 @@ body.components-dropdown {
 
   .shw-component-dropdown-fixed-height-container {
     min-height: 200px;
+    outline: 1px dotted #d3d3d3;
```
zamoore, line 26: do we have variables for these colors somewhere?

alex-ju: not yet for the outline/borders, but we do have [a minimal set](https://github.com/hashicorp/design-system/blob/main/showcase/app/styles/_tokens.scss) of token in `showcase`. can take some time to abstract `#d3d3d3` too

alex-ju: addressed in ff16764 by replacing with the closest grey available as token

alex-ju: Cristiano asked me to move this in a separate PR so https://github.com/hashicorp/design-system/pull/2540 is the nbetter place to look for this change

packages/components/src/components/hds/dropdown/index.ts
```diff
@@ -137,7 +144,7 @@ export default class HdsDropdown extends Component<HdsDropdownSignature> {
     classes.push(`hds-dropdown__content--position-${this.listPosition}`);
 
     // add a class based on the @width argument
```
didoo, line 146: [nit]
```suggestion
    // add a class based on the @width or @matchToggleWidth arguments
```

showcase/app/templates/components/dropdown.hbs
```diff
@@ -13,18 +13,99 @@
...
-          <D.Interactive @href="#">Edit</D.Interactive>
-        </Hds::Dropdown>
+      <SG.Item @label={{capitalize position}}>
+        <div {{style padding="6em 12em"}} class="shw-outliner">
+          <Hds::Dropdown @isOpen={{true}} @listPosition={{position}} as |D|>
+            <D.ToggleButton @color="secondary" @text="Menu" />
+            <D.Interactive @href="#">Create</D.Interactive>
+            <D.Interactive @href="#">Edit</D.Interactive>
+          </Hds::Dropdown>
+        </div>
```
didoo, line 23: [suggestion] you can use directly the `Shw::Outliner` component here
```suggestion
        <Shw::Outliner {{style padding="6em 12em"}}>
          <Hds::Dropdown @isOpen={{true}} @listPosition={{position}} as |D|>
            <D.ToggleButton @color="secondary" @text="Menu" />
            <D.Interactive @href="#">Create</D.Interactive>
            <D.Interactive @href="#">Edit</D.Interactive>
          </Hds::Dropdown>
        </Shw::Outliner>
```

website/docs/components/dropdown/partials/code/component-api.md
```diff
@@ -75,10 +75,14 @@ The Dropdown component is composed of different child components each with their
   </C.Property>
   <C.Property @name="width" @type="string" @valueNote="any valid CSS width (px, rem, etc)">
     By default, the Dropdown List has a `min-width` of `200px` and a `max-width` of `400px`, so it adapts to the content size. If a `@width` parameter is provided then the list will have a fixed width.
+    <br/><br/>We discourage the use of percentage values for this argument. The Dropdown list is [a `popover` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover), so relative units use `document` as a reference (not the parent element), meaning percentage values act similar to `vw`.
```
didoo, line 78: [praise] nice! 👌

didoo: [question] should we specify here (and potentially in the "how to use" example) that if a consumer sets `@matchToggleWidth={{true}}` the `@width` argument is ~~ignored~~ irrelevant?

packages/components/src/components/hds/dropdown/index.ts
```diff
@@ -137,7 +144,7 @@ export default class HdsDropdown extends Component<HdsDropdownSignature> {
...
-    if (this.args.width) {
+    if (this.args.width || this.args.matchToggleWidth) {
       classes.push('hds-dropdown__content--fixed-width');
```
didoo, line 148: [question] in the template we're setting `{{style width=@width max-height=@height}}` independently from the fact that `@matchToggleWidth` is true or false; should we instead avoid assigning a `width` if the value is true?

(practically it doesn't change, because _I think_ what is currently happening is that it sets an initial width equal to the `@width` value but then the Popover API takes over and assigns the toggle width value.

No strong opinions, just wanted to gather your thought on what is better to do in this case.

didoo (approved):
All good for me. Left a few minor questions/suggestions, but nothing is a blocker.

packages/components/src/components/hds/dropdown/index.ts
alex-ju: > (practically it doesn't change, because I think what is currently happening is that it sets an initial width equal to the @width value but then the Popover API takes over and assigns the toggle width value.

yes, that's how it behaves. I gave this a thought but wasn't sure it was worth interfering with the style modifier to avoid an override

showcase/app/styles/showcase-pages/dropdown.scss
KristinLBradley: In the past we haven't worried so much about standardizing color values like this in the Showcase. (Just to mention, not saying we shouldn't.)

alex-ju: indeed, `showcase` was generally seen as a scrappy playground

PR #2536, opened 2024-10-31
`SideNav` - A11y improvements

KristinLBradley:
### :pushpin: Summary

If merged, this PR makes a11y related improvements to the `SideNav` related to List::Title headers and the ToggleButton aria properties following work done in the deprioritized `AppSideNav`.

**Implemented changes:**

* Change `List::Title` to h3 & add visually hidden h2 to AppSideNav - [Reference comment](https://github.com/hashicorp/design-system/pull/2384#discussion_r1764009486)
* Improve ariaLabel for `ToggleButton` state & use aria-expanded - [Reference comment](https://github.com/hashicorp/design-system/pull/2384#discussion_r1764017267)

### :link: External links

* Related AppSideNav PR: https://github.com/hashicorp/design-system/pull/2463

***

### 👀 Component checklist


shleewhite (approved):
Thank you for porting these over!

alex-ju (commented):
We should check for usage of `@ariaLabel` in `<Hds::SideNav />` in our products before removing it. If no instances are found we should update the docs as well.

A more careful approach would be to keep the `@ariaLabel` argument, update the getter to return `this.args.ariaLabel`, and deprecate it.

KristinLBradley:
@alex-ju I couldn't find uses of the `ariaLabel` on the Hds::SideNav when I searched but I went ahead and brought it back and deprecated it.

(Search: [https://github.com/search?q=org%3Ahashicorp+%22%3CHds%3A%3ASideNav+%22&type=code](https://github.com/search?q=org%3Ahashicorp+%22%3CHds%3A%3ASideNav+%22&type=code))

packages/components/src/components/hds/side-nav/index.ts
```diff
@@ -105,6 +127,13 @@ export default class HdsSideNav extends Component<HdsSideNavSignature> {
     return (this.isResponsive && !this.isDesktop) || this.isCollapsible;
   }
 
+  get ariaLabel(): string {
+    if (this.isMinimized) {
+      return this.args.ariaLabel ?? 'Open menu';
+    }
+    return this.args.ariaLabel ?? 'Close menu';
+  }
+
```
alex-ju: at this point (since we don't need the old labels) it can even be
```suggestion
      return this.args.ariaLabel;
  }
```

packages/components/src/components/hds/side-nav/index.ts
```diff
@@ -72,6 +77,23 @@ export default class HdsSideNav extends Component<HdsSideNavSignature> {
...
+      deprecate(
+        'The `@ariaLabel` argument for "Hds::SideNav" has been deprecated. It is replaced by aria-labelledby and aria-expanded on the toggle button',
+        false,
+        {
+          id: 'hds.sidenav',
+          until: '5.0.0',
+          url: 'https://helios.hashicorp.design/components/sidenav?tab=version%20history#4130',
+          for: '@hashicorp/design-system-components',
+          since: {
+            available: '4.13.0',
```
alex-ju, line 91: suggestion: next release is going to be 4.14 (needs updated in the website docs as well)
```suggestion
            available: '4.14.0',
```

packages/components/src/components/hds/side-nav/index.ts
KristinLBradley: Updated this. thanks

alex-ju (approved):
Looking good! 🚀

PR #2546, opened 2024-11-07
`RichTooltip` - Standardize private properties and methods

dchyun:
### :pushpin: Summary

- Provides written standards for the usage of private properties and methods into the new component checklist guide.
- Standardizes the private properties on the `RichTooltip` as a test component for these changes that will be made throughout the components. 

This is the first PR in a series of updates that will implement a consistent naming convention for private properties and methods across the component library. The goal is to follow the TypeScript + underscore prefix naming convention (`private _propertyName`), and extend this across the whole library. 

### :link: External links

Jira ticket: [HDS-2508](https://hashicorp.atlassian.net/browse/HDS-2508)

***

### 👀 Component checklist

(https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

[HDS-2508]: https://hashicorp.atlassian.net/browse/HDS-2508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/NEW-COMPONENT-CHECKLIST.md
+++ b/packages/components/NEW-COMPONENT-CHECKLIST.md
@@ -112,6 +112,7 @@ The engineering checklist has six parts: creating the feature branch, component 
   - write API comments in the [JS doc](https://jsdoc.app/) format (look at existing components for examples)
   - use the same naming as the Figma file for the components API, unless it conflicts with a pre-existing HTML attribute. If that is the case, document the difference in a comment.
   - booleans should start with a verb (is/has/etc)
+  - private properites and methods should be declared as `private` and start with an `_`, e.g., `private _elementId`
   - assertions should match the content style of the other components, e.g., `'@text for "Hds::Button" must have a valid value'`,
   - program with intent; think about the invocation for the developer who will use the component. The goal is a terse invocation, but we also want to consider the big picture. Try to get feedback when you can.
 - [ ] **component style**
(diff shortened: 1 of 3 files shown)


shleewhite (approved):
This looks good to me too! The only question I have is do we just make everything private? are there any situations we should keep in mind when deciding if a property should be public/protected/private?

dchyun:
> This looks good to me too! The only question I have is do we just make everything private? are there any situations we should keep in mind when deciding if a property should be public/protected/private?

Yeah this is a great point. Maybe it's on a case by case basis. For example with this tooltip, if there's a situation where consumers would need to have access to these generated ids, we would need to keep them public. @alex-ju do you ever see that being the case?

alex-ju:
> > This looks good to me too! The only question I have is do we just make everything private? are there any situations we should keep in mind when deciding if a property should be public/protected/private?
> 
> Yeah this is a great point. Maybe it's on a case by case basis. For example with this tooltip, if there's a situation where consumers would need to have access to these generated ids, we would need to keep them public. @alex-ju do you ever see that being the case?

It's a good question – hence the recommendation to do it on a component so we all get a taste of what it would look like. 

Most properties will likely be marked as private; [sometimes it helps differentiate them from the getters](https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/disclosure-primitive/index.ts#L38-L49), which comes in handy when managing internal states. sometimes, they will not bring much value (aside from being aligned with other components).

PR #2547, opened 2024-11-07
Fix position for autocomplete container

alex-ju:
### :pushpin: Summary

Fix position for the autocomplete container

Before
Screenshot 2024-11-07 at 18 07 33

After
Screenshot 2024-11-07 at 18 07 26

### :link: External links

[Slack thread](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1730984851735729)

***

### DIFF
--- a/website/app/styles/doc-components/algolia-search/index.scss
+++ b/website/app/styles/doc-components/algolia-search/index.scss
@@ -11,6 +11,7 @@
 
 // "autocomplete" target container - not used
 .doc-algolia-search__autocomplete-container {
+  // we assign it a position absolute to remove it from the layout flow, so the `gap` property of the parent flex container doesn't get applied to this as well
   position: absolute;
 }
 
--- a/website/app/styles/doc-components/algolia-search/index.scss
+++ b/website/app/styles/doc-components/algolia-search/index.scss
@@ -10,7 +10,9 @@
 @import "./results";
 
 // "autocomplete" target container - not used
-// .doc-algolia-search__autocomplete-container {}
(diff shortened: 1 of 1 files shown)


shleewhite (approved):

KristinLBradley (approved):
Looks good.

I was curious about how the layout structure was working so investigated a bit:

image

website/app/styles/doc-components/algolia-search/index.scss
```diff
@@ -10,7 +10,9 @@
 @import "./results";
 
 // "autocomplete" target container - not used
-// .doc-algolia-search__autocomplete-container {}
+.doc-algolia-search__autocomplete-container {
+  position: absolute;
```
didoo, line 14: [suggestion]
```suggestion
  // we assign it a position absolute to remove it from the layout flow, so the `gap` property of the parent flex container doesn't get applied to this as well
  position: absolute;
```

didoo: [suggestion] a possible alternative approach to solving the problem
```suggestion
  // we use this trick so the `gap` property of the parent flex container doesn't get applied to this as well
  display: contents;
```

didoo (approved):
left a couple of suggestions (one of them an alternative approach, just in case)

website/app/styles/doc-components/algolia-search/index.scss
alex-ju: a good alternative, indeed. not sure which one is 'better', so I'll stick with `position: absolute`

PR #2548, opened 2024-11-07
Website - sticky tabs

dchyun:
### :pushpin: Summary

Implementing sticky behavior for page tabs on the website to remain visible as the user scrolls down the page. 

In this PR I have
- Updated `Doc::Page::Tabs` to have a sticky behavior
- Moved `Doc::Page::Tabs` outside of `Doc::Page::Cover` to be a direct child of `Doc::Page::Stage`. This was necessary to implement the sticky interaction
- Updated the styles of `Doc::Page::Cover` to account for `Doc::Page::Tabs` no longer being a direct child
- Adjusted the scroll behavior of the header links and `Doc::Page::Sidecar` to account for the sticky tabs
- Updated the `z-index` of the sidecar and tabs to account for when the tabs overlay the code blocks

### :link: External links

Jira ticket: [HDS-3297](https://hashicorp.atlassian.net/browse/HDS-3297)

***

[HDS-3297]: https://hashicorp.atlassian.net/browse/HDS-3297?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/styles/pages/application/stage.scss
+++ b/website/app/styles/pages/application/stage.scss
@@ -40,8 +40,8 @@
       auto;
     margin: 0 auto 0 0;
   }
+}
 
-  &.doc-page-stage--tabs {
-    --doc-page-offset-height: calc(var(--doc-page-header-height) + var(--doc-page-tabs-height));
-  }
+.doc-page-stage--tabs {
+  --doc-page-offset-height: calc(var(--doc-page-header-height) + var(--doc-page-tabs-height));
 }
--- a/website/app/styles/pages/application/stage.scss
+++ b/website/app/styles/pages/application/stage.scss
@@ -8,10 +8,13 @@
 @use "../../breakpoints" as breakpoint;
 
 .doc-page-stage {
+  --doc-page-offset-height: var(--doc-page-header-height);
+
   display: grid;
   grid-area: stage;
   grid-template-areas:
     "cover"
+    "tabs"
     "content";
   grid-template-rows: auto 1fr;
   grid-template-columns: 1fr;
@@ -25,6 +28,7 @@
     --max-content-width: calc(var(--doc-page-content-max-width) + 2 * var(--doc-page-stage-gutter-large));
     grid-template-areas:
       "cover cover cover"
+      "tabs tabs tabs"
       "content sidecar _extra";
     grid-template-columns:
       // content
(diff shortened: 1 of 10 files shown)


andgen404 (commented):
Design wise this looks good to me 👍 really like the scroll to top interaction when switching between tabs.

shleewhite (commented):
When testing with a keyboard, if I click a tab my focus gets put back at the top of the page. Seems like this is an existing issue with the docs site, but this might be a good opportunity to improve this. Ideally, it would focus the newly visible tabpanel.

dchyun:
> When testing with a keyboard, if I click a tab my focus gets put back at the top of the page. Seems like this is an existing issue with the docs site, but this might be a good opportunity to improve this. Ideally, it would focus the newly visible tabpanel.

Created [HDS-4150](https://hashicorp.atlassian.net/browse/HDS-4150) to look into this issue. An a11y decision needs to be made around what the behavior needs to be. The tab changes currently trigger a route change, and for all route changes right now the focus is reset to the beginning of the page using the [ember-a11y-refocus](https://github.com/ember-a11y/ember-a11y-refocus) library.

website/app/styles/pages/application/stage.scss
```diff
@@ -25,6 +28,7 @@
     --max-content-width: calc(var(--doc-page-content-max-width) + 2 * var(--doc-page-stage-gutter-large));
     grid-template-areas:
       "cover cover cover"
+      "tabs tabs tabs"
```
alex-ju, line 31: I like how this reads! 😄

website/app/styles/pages/application/stage.scss
```diff
@@ -36,4 +40,8 @@
       auto;
     margin: 0 auto 0 0;
   }
+
+  &.doc-page-stage--tabs {
+    --doc-page-offset-height: calc(var(--doc-page-header-height) + var(--doc-page-tabs-height));
+  }
```
alex-ju, line 46: suggestion: my preference for class modifiers is to live as siblings. I don't think we need specificity here, but I appreciate that un-nesting it means we rely on definition order

dchyun: Implemented your suggestion! Let me know if it looks okay.

alex-ju: Yes, all good from me

MelSumner (approved):
from an accessibility POV this navigates correctly and never blocks any other interactive content. 👍

PR #2554, opened 2024-11-12
`Table` component - Expose the `@each` index

didoo:
### :pushpin: Summary

In the context of this conversation https://hashicorp.slack.com/archives/C7KTUHNUS/p1731002512029259 I've opened a quick PR, so the Core team can evaluate if it makes sense to add it to our component, or not (and rely on consumers using a custom `@each` loop as explained [here](https://helios.hashicorp.design/components/table?tab=code#using-each-to-loop-over-records-to-create-rows)).

In this PR I have:
- yielded the `index` of the `each` loop as `rowIndex` to the Table content
- updated the table showcase to have at least one of the examples to use the yielded `rowIndex`
- updated integration test for `Table` to cover the yielded `rowIndex`
- updated the Component API documentation for the `Table` to add the yielded `rowIndex`

***

### 👀 Component checklist


showcase/app/templates/components/table.hbs
```diff
@@ -21,7 +21,7 @@
     }}
   >
     <:body as |B|>
-      <B.Tr>
+      <B.Tr id={{B.rowIndex}}>
```
KristinLBradley, line 24: Maybe you could add an example that displays the rowIndex visually inside a `td`?

KristinLBradley (approved):
Looks good. Just added one comment about maybe displaying the rowIndex values visually within a Showcase example.

showcase/app/templates/components/table.hbs
didoo: I though about it, but what would it add? this is not a UI-related thing, is a functional/code thing, and it's already covered in the integration test, which checks that we're actually outputting/yielding the `rowIndex`value. 
So my thought is that it was not the case to add it to the showcase; but if there are strong feelings about it, I can do it.

KristinLBradley: Just a suggestion. I thought it would work for visual regression and just making it more obvious that it exists in the Showcase context. I don't have strong feelings about it.

alex-ju (approved):
Looking good. We can add the 'component' changelog entry too (if not we can do it before the release)

website/docs/components/table/partials/code/component-api.md
```diff
@@ -24,6 +24,9 @@ The Table component itself is where most of the options will be applied. However
   <C.Property @name="<:body>" @type="named block">
     This is a named block where the content for the table body (`<tbody>`) is rendered.
     <Doc::ComponentApi as |C|>
+      <C.Property @name="[B].rowIndex" @type="number">
+        The value of the index asssociated with the `@each` loop.
```
alex-ju, line 28: thought: I was thinking if we should mention that this is available only if `@model`/`@columns` exists, but that's true for most of the arguments

didoo:
> We can add the 'component' changelog entry too (if not we can do it before the release)

👍 (done in https://github.com/hashicorp/design-system/pull/2554/commits/72fd07b48a94962705f17ab252ab6f51aa14250b)

shleewhite (approved):

PR #2570, opened 2024-11-18
add designers as approved codeowners for docs

Dhaulagiri:
### :pushpin: Summary

This would allow for designers to independently approve and merge documentation changes

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,2 +1,5 @@
 # Default owners
 *       @hashicorp/hds-engineering
+
+# co-ownership of website docs
+/website/docs/ @hashicorp/hds-engineering @hashicorp/hds-design
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,4 +2,5 @@
 *       @hashicorp/hds-engineering
 
 # co-ownership of website docs & assets
-/website/docs/ /website/public/assets/ @hashicorp/hds-engineering @hashicorp/hds-design
+/website/docs/ @hashicorp/hds-engineering @hashicorp/hds-design
+/website/public/assets/ @hashicorp/hds-engineering @hashicorp/hds-design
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1,2 +1,5 @@
 # Default owners
-*       @hashicorp/hds-engineering
\ No newline at end of file
+*       @hashicorp/hds-engineering
+
+# co-ownership of website docs
+/website/docs/ @hashicorp/hds-engineering @hashicorp/hds-design
```
alex-ju, line 5: meaning that any changes to files within this directory will require approval from members of either of these teams

alex-ju (dismissed):

.github/CODEOWNERS
Dhaulagiri: is it "either" or "both"? I believe it's both _can_ but only 1 is required to unblock merging?

didoo, line 5: should we include also the `website/public/assets` folder? usually they add images for the documentation there, too

Dhaulagiri: good call. I added that as well.

.github/CODEOWNERS
```diff
@@ -1,2 +1,5 @@
 # Default owners
-*       @hashicorp/hds-engineering
\ No newline at end of file
+*       @hashicorp/hds-engineering
+
+# co-ownership of website docs & assets
+/website/docs/ /website/public/assets/ @hashicorp/hds-engineering @hashicorp/hds-design
```
shleewhite, line 5: issue: I think docs and assets need to be done on separate lines, github says its an invalid owner
Screenshot 2024-11-19 at 10 26 07 AM

Dhaulagiri: 🤦

alex-ju: I was about to leave the same comment – it's a shame you can't chain them

PR #2573, opened 2024-11-19
HDS::CodeEditor with Code Mirror 6

zamoore:
### :pushpin: Summary

This PR adds a code editor utilizing Code Mirror 6 to the HDS component library. The code editor can be instantiated via:
- The `hds-code-editor` modifier, which can be added to any element to transform that element into a code editor instance
- The `Hds::CodeEditor` component, which uses the `hds-code-editor` modifier under the hood, but adds support for title, description, and various tools.

#### `hds-code-editor` modifier

This modifier will instantiate a Code Mirror editor on the element that it's added to. In order to reduce the up-front cost of loading the editor, we only load the CodeMirror package when the editor is scrolled into view. Because of this, there may be an exceptionally brief loading state before the editor is ready when it is viewed by the user.

```
// Example
<div {{hds-code-editor
  value=@value
  language=@language
  onInput=@onInput
  onSetup=this.onSetup
}} />
```
`value`: The initial value of the editor
`language`: optional, one of 4 possible supported languages (json, sql, go, hcl)
`onInput`: callback function when the code editor changes
`onSetup`: callback function when the modifier has completed loading dependencies and instantiating the editor, receives CodeMirror Editor instance.

This editor features support for:

- Custom title
- Custom description
- Copy editor content
- Fullscreen toggle
- Keyboard support including directional navigation
- History (support for `cmd+z` and other history commands)

#### `Hds::CodeEditor` component

This component uses the `hds-code-editor` modifier in combination with a customizable header and support for a fullscreen mode.

```
// Example
<Hds::CodeEditor
  @language="go"
  @value="test"
  @hasExpandButton={{true}}
  @hasCopyButton={{true}}
  @onInput={{this.noop}}
  @onSetup={{this.noop}}
  as |CE|
>
  <CE.Title>Code editor with title</CE.Title>
  <CE.Description>This is a code editor with a description</CE.Description>
  <Hds::Button @text="Custom button" />
  <Hds::Icon @name="code" @color="#fff" />
</Hds::CodeEditor>
```
`value`: The initial value of the editor
`hasExpandButton`: Whether the editor can be expanded into fullscreen mode
`hasCopyButton`: Whether the editor has a button that adds the current value to the clipboard
`language`: optional, one of 4 possible supported languages (json, sql, go, hcl)
`onInput`: callback function when the code editor changes
`onSetup`: callback function when the component has completed loading dependencies and instantiating the editor, receives CodeMirror Editor instance.

Screenshot 2024-12-17 at 5 22 00 PM

### :link: External links

Jira ticket: [HDS-4016](https://hashicorp.atlassian.net/browse/HDS-4016)
[Figma](https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/branch/ZGXtIMxAYf16lgGLcspMty/HDS-Components-v2.0?node-id=70828-8699&p=f&m=dev)

***

### 👀 Component checklist

[HDS-4016]: https://hashicorp.atlassian.net/browse/HDS-4016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -0,0 +1,132 @@
...
+  drawSelection,
+  highlightActiveLine,
+  EditorView,
+  keymap,
+} from '@codemirror/view';
+import { EditorState, type Extension } from '@codemirror/state';
+import { javascript } from '@codemirror/lang-javascript';
+import { json } from '@codemirror/lang-json';
+import { sql } from '@codemirror/lang-sql';
+import { go } from '@codemirror/lang-go';
```
alex-ju, line 21: suggestion: investigate loading languages asynchronously, to reduce the size of loaded assets for CodeEditor instances

packages/components/package.json
```diff
@@ -35,14 +35,24 @@
...
     "@ember/render-modifiers": "^2.0.5",
     "@ember/string": "^3.1.1",
     "@ember/test-waiters": "^3.1.0",
     "@embroider/addon-shim": "^1.8.7",
     "@floating-ui/dom": "^1.6.3",
     "@hashicorp/design-system-tokens": "^2.2.1",
     "@hashicorp/flight-icons": "^3.7.0",
+    "@lezer/highlight": "^1.2.1",
     "clipboard-polyfill": "^4.1.0",
+    "codemirror": "^5.65.12",
```
alex-ju, line 55: the main `codemirror` dependency must be removed (this is covered by `@codemirror/state` and `@codemirror/view`, as detailed in the RFC)
```suggestion
    "codemirror": "^5.65.12",
```

packages/components/package.json
```diff
@@ -75,6 +85,7 @@
     "@glint/template": "^1.4.0",
     "@rollup/plugin-babel": "^6.0.4",
     "@tsconfig/ember": "^3.0.8",
+    "@types/codemirror": "^5.60.15",
```
alex-ju, line 88: this should not be required (`@codemirror/state` and `@codemirror/view` have their type declarations in the packages)

packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -0,0 +1,132 @@
...
+import {
+  lineNumbers,
+  highlightActiveLineGutter,
+  highlightSpecialChars,
+  drawSelection,
+  highlightActiveLine,
+  EditorView,
+  keymap,
+} from '@codemirror/view';
+import { EditorState, type Extension } from '@codemirror/state';
```
alex-ju, line 17: suggestion: investigate if we can defer importing these until setup.

alex-ju (commented):
Thank you for opening this PR for early feedback! The functionalities look very promising.

The main comment at this stage (when the design is finalized I can do an in-depth review of the component API) is to look into how can we leverage dynamic imports (leveraging `ember-auto-import`) to defer loading the third-party code until consumers start using the component on the page

packages/components/src/styles/components/code-editor.scss
```diff
@@ -0,0 +1,92 @@
...
+//
+
+.hds-code-editor {
+  $border-strong: rgba(178, 182, 189, 40%);
+
+  background-color: var(--token-color-code-surface-primary);
+  border: 1px solid;
+  border-radius: 6px;
+
+  &--is-fullscreen {
```
didoo, line 17: [todo] we don't use this kind of nesting in HDS (on purpose), we prefer a much flatten structure, and avoid using the `&` operator to create class names, in favour of more explicit (and easy to find) names

packages/tokens/src/global/color/semantic-code.json
```diff
@@ -0,0 +1,42 @@
+{
+    "color": {
+        "code": {
```
didoo, line 3: [issue/blocker] these tokens don't exist (as variables) in Figma, as far as I know. If we want to use variables for the CSS we should use local Sass variables. Also, these would not be semantic tokens, but component-level tokens.

Plus, it's not clear how these tokens (as in `--token-color`) would work in the context of theming (something [we're still investigating](https://hashicorp.atlassian.net/browse/HDS-4239)). So my suggestion is to keep them local to the file, as `Sass` variables.

(there's an [RFC here](https://docs.google.com/document/d/1Mjoo8FQo3cJWUzf8j3Yh0m4PZoM3Ixw0qwREkgPiKmQ/) but is a bit dated; otherwise happy to jump on a call to explain more in detail how the design tokens are organized)

didoo (commented):
Left a couple of high-level comments, about something that stood out for me (I hope you won't mind).

packages/tokens/src/global/color/semantic-code.json
didoo: I've just seen @majedelass [comment in Slack](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1733752232153229?thread_ts=1733751911.485639&cid=C025N5V4PFZ) and [this comment in the CRD](https://docs.google.com/document/d/1fmWOUtajtJHsThKG_uozcpSWGJ-O4MmtqzMPEZRqUeg/edit?disco=AAABZPv0994), so I have looked at the file in Figma and I see that the styles and variables seem to be local, not public (@majedelass correct me if I'm wrong):
screenshot_4479
screenshot_4480

So, this confirms my understanding that these should be local variables (Sass) and not public design tokens.

Happy to discuss more in detail, if needed.

/cc @alex-ju @zamoore @hashicorp/hds-design

majedelass: They are local, you are correct.

packages/components/src/modifiers/hds-code-editor.ts
zamoore: Good call. In the latest version we only load the editor code and the language package when the code editor is in-view. We also only load the language if one is specified.

packages/components/src/modifiers/hds-code-editor.ts
zamoore: Good call, done in the latest.

packages/components/src/styles/components/code-editor.scss
zamoore: For sure. This was just shorthand for me during development.

didoo:
@zamoore I checked the Percy output, and apart from diffs due to the table sub-pixel shift (probably you need to rebase on `main`) I see a change in the `Accordion` page which is not expected. Can you double check if it's a false positive or something that changed (in the position/margin of the button).

packages/components/src/components/hds/code-editor/index.hbs
```diff
@@ -0,0 +1,53 @@
...
+              @textToCopy={{@value}}
+            />
+          {{/if}}
+          {{#if @hasExpandButton}}
+            <Hds::CodeEditor::Button
+              class="hds-code-editor__expand-button"
+              @isIconOnly={{true}}
+              @size="small"
+              @icon="maximize"
+              @text="Fullscreen"
```
shleewhite, line 36: Haven't done a full review yet, but just noticed this.

The label should be "Toggle full screen view" and then the button needs to have `aria-pressed` set to true or false depending on the state.

packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -0,0 +1,189 @@
...
+      ] = await Promise.all([
+        import('@codemirror/view'),
+        import('@codemirror/state'),
+        import('@codemirror/commands'),
+        import('@codemirror/language'),
+      ]);
+
+      this.onInput = onInput;
+
+      const languageExtension = await this.loadLanguageTask.perform(language);
```
alex-ju, line 143: suggestion: we have a chain of promises here: the perform action as well as the async loading of modules within the `loadLanguageTask`. I suggest we add some error handling to prevent a JS crash in case things go wrong along the way.

packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -0,0 +1,189 @@
...
+          rootMargin: LOADER_HEIGHT,
+        }
+      );
+
+      this.observer.observe(element);
+    }
+  }
+
+  willRemove() {
+    this.observer.disconnect();
```
alex-ju, line 75: 👏

packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -0,0 +1,189 @@
...
+        hdsDarkTheme,
+        keymap.of([...defaultKeymap, ...historyKeymap]),
+        bracketMatching(),
+        syntaxHighlighting(hdsDarkHighlightStyle),
+        history(),
+      ];
+
+      if (languageExtension !== undefined) {
+        extensions = [languageExtension, ...extensions];
+      }
```
alex-ju, line 174: thought: this can be extracted into a separate `buildExtensions` method; not a priority, can be refactored afterwards.

packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -0,0 +1,189 @@
...
+
+  loadLanguageTask = task(
+    { drop: true },
+    async (language?: HdsCodeEditorLanguages) => {
+      if (language === undefined) {
+        return;
+      }
+
+      let module: CodemirrorLanguageModule | null = null;
+      let languageFunction = null;
```
alex-ju, line 86: suggestion: it would  be great if we can give this a type as well

packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -0,0 +1,189 @@
...
+          module = await import('@codemirror/lang-sql');
+          languageFunction = (module as CodemirrorSqlModule).sql;
+          break;
+        case 'hcl':
+          module = await import('codemirror-lang-hcl');
+          languageFunction = (module as CodemirrorHclModule).hcl;
+          break;
+        default:
+          throw new Error(`Language ${language} is not supported`);
+      }
```
alex-ju, line 107: suggestion: we can wrap this into a try-catch, with some sort of 'Failed to load language module for ${language}'

packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -0,0 +1,189 @@
...
+
+      return languageFunction();
```
alex-ju, line 109: suggestion: in case things go wrong above
```suggestion
      return languageFunction?.();
```

alex-ju (commented):
Awesome work here! I did a high-level review of this and here are my initial findings:

**Issue**: Copy button doesn't work (doesn't seem to be wired up correctly); we should add a test for this as well (aware it's missing from CodeBlock too)

**Issue**: Focus state on toolbar buttons is not consistent
Screenshot 2024-12-19 at 12 00 54

**Issue**: In full screen mode, if the editor content is longer than the viewport height, the header is not visible/accessible in the viewport anymore, resulting in user being trapped (as the full screen toggle button is not visible/accessible anymore).

**Suggestion**: When in full screen mode, it would  be great if escape key would switch to 'compact' mode; a feature that can be added afterwards.

packages/components/src/components/hds/code-editor/index.ts
```diff
@@ -0,0 +1,68 @@
...
+        Button?: ComponentLike<HdsButtonSignature>;
+      },
+    ];
+  };
+  Element: HTMLDivElement;
+}
+
+export default class HdsCodeEditor extends Component<HdsCodeEditorSignature> {
+  @tracked isFullscreen = false;
+  @tracked isSetupComplete = false;
```
alex-ju, line 40:
```suggestion
  @tracked private _isSetupComplete = false;
```

alex-ju, line 40:
```suggestion
  @tracked private _isFullscreen = false;
  @tracked private _isSetupComplete = false;
```

packages/components/src/components/hds/code-editor/index.hbs
```diff
@@ -0,0 +1,53 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<div class={{this.classNames}} ...attributes>
+  {{#if (and this.isSetupComplete (or this.hasToolbarButton (has-block)))}}
```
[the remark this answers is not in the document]
zamoore: It's part of the design at the moment, but we don't need to hide them for any real reason.

@majedelass - How do you want to handle the header during loading?

packages/components/src/styles/components/code-editor/theme.scss
```diff
@@ -0,0 +1,18 @@
...
+.hds-code-editor--theme-dark {
+  // COLORS
+
+  // Base UI colors:
+  --hds-code-editor-color-border-strong: rgba(178, 182, 189, 40%);
+  --hds-code-editor-color-foreground-primary: #d5d7db;
+  --hds-code-editor-color-foreground-faint: #878a8f;
+  --hds-code-editor-color-foreground-high-contrast: #fff;
+  --hds-code-editor-color-surface-faint: #15181e;
+  --hds-code-editor-color-surface-primary: #0D0E12;
```
majedelass, line 17: Is there a reason why these colors are RGBA/HEX values, vs tokens (local) that are shared among this component and the CodeBlock? Both of them are visually the same, and wouldn't creating shared tokens make maintenance easier?

packages/components/src/components/hds/code-editor/index.hbs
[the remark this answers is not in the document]
alex-ju: right – happy to keep it as is if it was a design decision 👍

packages/components/src/components/hds/code-editor/index.hbs
```diff
@@ -0,0 +1,53 @@
...
+          (hash
+            Title=(component "hds/code-editor/title")
+            Description=(component "hds/code-editor/description")
+            Button=(component "hds/code-editor/button")
+          )
+        }}
+      </div>
+
+      {{#if this.hasToolbarButton}}
+        <div class="hds-code-editor__header-toolbar">
```
majedelass, line 20: Can we change this name to "hasActions?"

This is how it looks like in the figma properties
image

packages/components/src/components/hds/code-editor/index.hbs
majedelass: looping @alex-ju into this because we discussed how it looks in Figma. How we want this to represent should be similar (or I can change to mimic), but this is what it looks like in the properties (status has both maximize and minimize as options):

image

majedelass (commented):
Added some high level feedback.

packages/components/src/components/hds/code-editor/index.hbs
majedelass: (hasActions is the container -- which would potentially become `hds-code-editor__header-actions`)

shleewhite (commented):
Nice work! Most of the feedback I have is focused on accessibility.

packages/components/src/components/hds/code-editor/index.hbs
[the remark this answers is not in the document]
majedelass: I think we should not show it while it's loading, unless there is an accessibility implication to that decision.

packages/components/src/components/hds/code-editor/index.hbs
```diff
@@ -0,0 +1,53 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<div class={{this.classNames}} ...attributes>
```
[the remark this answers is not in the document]
alex-ju: +1 to this. I suggest we leverage `ember-focus-trap` to address this

packages/components/src/components/hds/code-editor/index.hbs
```diff
@@ -0,0 +1,53 @@
...
+            Button=(component "hds/code-editor/button")
+          )
+        }}
+      </div>
+
+      {{#if this.hasToolbarButton}}
+        <div class="hds-code-editor__header-toolbar">
+          {{#if @hasCopyButton}}
+            <Hds::Copy::Button
+              class="hds-code-editor__button hds-code-editor__copy-button"
```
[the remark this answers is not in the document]
zamoore: `.hds-code-editor__button` also sets the dark styles for any additional buttons that the user adds to the header with the yielded `Button` component.

packages/components/src/components/hds/code-editor/description.hbs
```diff
@@ -0,0 +1,9 @@
+<Hds::Text::Body
+  class="hds-code-editor__description hds-code-editor__header-content-text"
+  @tag="p"
+  @size="100"
+  @color="var(--hds-code-editor-color-foreground-faint)"
```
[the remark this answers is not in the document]
zamoore: Definitely. I wasn't sure on what the preference was for use of the color argument

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,106 @@
...
+  .hds-code-editor__loader {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    height: 164px;
+    background-color: var(--hds-code-editor-color-surface-primary);
+  }
+
+  .hds-code-editor__button,
+  .hds-code-editor__button.hds-copy-button {
```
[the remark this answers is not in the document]
zamoore: We need the extra specificity to override the copy button styles.

packages/components/src/modifiers/hds-code-editor.ts
zamoore: Good idea!

packages/components/src/styles/components/code-editor/theme.scss
zamoore: I'm not sure what you mean by local tokens. This is also how we set up the colors for the CodeBlock. Do you have something you could point me to?

packages/components/src/components/hds/code-editor/index.hbs
```diff
@@ -0,0 +1,53 @@
...
+        </div>
+      {{/if}}
+    </div>
+  {{/if}}
+  <div
+    class="hds-code-editor__editor"
+    {{hds-code-editor value=@value language=@language onInput=@onInput onSetup=this.onSetup}}
+  />
+  {{#unless this.isSetupComplete}}
+    <div class="hds-code-editor__loader">
```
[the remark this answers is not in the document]
zamoore: Agree on the aria/role additions. I'm not sure we can do the loading state in the showcase unless we want to add another external argument to override the loading state. Right now it's just reflective of the loading task state inside the component.

showcase/app/styles/app.scss
```diff
@@ -30,16 +30,15 @@
 // START COMPONENT PAGES IMPORTS
 @import "./showcase-pages/accordion";
 @import "./showcase-pages/alert";
-@import "./showcase-pages/app-header";
```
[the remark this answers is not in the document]
zamoore: That must have been from a bad merge or something 🙅

packages/components/src/modifiers/hds-code-editor.ts
zamoore: That makes sense. I have it set up to warn now as the code editor is still technically functional without loading the language. Do you think it should be an error?

packages/components/src/components/hds/code-editor/index.hbs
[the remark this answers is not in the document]
shleewhite: There's a couple examples in the showcase like [dropdown](https://github.com/hashicorp/design-system/blob/5102e27eaa446d1e3d6028caa72148ffecc41d5c/showcase/app/templates/components/dropdown.hbs#L456) where we kinda fake a state by using the hds classes instead of the full component. Could be a way around adding a new external arg?

packages/components/src/types/hds-code-editor.types.ts
```diff
@@ -0,0 +1,36 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+import type { Extension } from '@codemirror/state';
```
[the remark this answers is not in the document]
zamoore: That was the intention, but it looks like these types are exclusively used in the modifier, so I've moved them there. Thanks for the question.

packages/components/src/components/hds/code-editor/description.hbs
```diff
@@ -0,0 +1,9 @@
+<Hds::Text::Body
+  class="hds-code-editor__description hds-code-editor__header-content-text"
```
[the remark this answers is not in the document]
zamoore: Sounds good. I've reworked things to fit these conventions.

packages/components/src/styles/components/code-editor/index.scss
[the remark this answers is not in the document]
zamoore: I changed this to be a style that targets Hds::Button elements within the code editor.

packages/components/src/components/hds/code-editor/index.hbs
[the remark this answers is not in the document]
zamoore: After double checking, there's not a way to do this in the showcase that I can think of. We can pass in an optional override that tells the component it's loading, but it will still display the code editor once it's loaded which doesn't provide a clear indication of what the loading state will look like.

packages/components/src/components/hds/code-editor/index.hbs
```diff
@@ -0,0 +1,49 @@
...
+      {{/if}}
+    </div>
+  {{/if}}
+  <div
+    class="hds-code-editor__editor"
+    {{hds-code-editor value=@value language=@language onInput=this.onInput onSetup=this.onSetup}}
+  />
+  {{#unless this._isSetupComplete}}
+    <div class="hds-code-editor__loader" aria-live="polite" role="status">
+      <Hds::Icon @name="loading" @size="24" />
```
shleewhite, line 46: Issue: sorry, I think this got lost in the last comment, but you also need to add a visually hidden span inside the div that says "loading" otherwise there is no text in the aria-live region.

```html
<span class="sr-only">Loading</span>
```

shleewhite (commented):
Looks great! Just one issue with the loading state.

packages/components/src/components/hds/code-editor/index.hbs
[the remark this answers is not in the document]
shleewhite: Thats ok! Its so short it should be fine

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,120 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+//
+// CODE-EDITOR
+//
+
+@import "theme";
```
dchyun, line 10:
```suggestion
@use "theme";
```
Small change to avoid the sass deprecation warnings, and align with other scss files.

dchyun (commented):
Looks great! Just caught one small thing in the styles.

packages/components/src/components/hds/code-editor/index.hbs
```diff
@@ -0,0 +1,50 @@
...
+            <Hds::CodeEditor::FullScreenButton
+              @isFullScreen={{this._isFullScreen}}
+              @onToggleFullScreen={{this.toggleFullScreen}}
+            />
+          {{/if}}
+        </div>
+      {{/if}}
+    </div>
+  {{/if}}
+  <div
```
shleewhite, line 40: Issue: Looking at the showcase again, `aria-label` should be required (or at least optional - but I don't know what a good default would be). The code editor div has `role="textbox"`, which makes it an input and needs an accessible name.

packages/components/src/modifiers/hds-code-editor.ts
alex-ju: A warning is probably best indeed.

packages/components/src/styles/components/code-editor/theme.scss
alex-ju: I believe Majed is wondering if we can share tokens/styles between the two components (CodeBlock and CodeEditor) instead of re-declaring them

packages/components/src/components/hds/code-editor/index.ts
```diff
@@ -0,0 +1,107 @@
...
+  @action
+  toggleFullScreen(): void {
+    this._isFullScreen = !this._isFullScreen;
+  }
+
+  @action
+  onInput(newVal: string): void {
+    this._value = newVal;
+    this.args.onInput?.(newVal);
+  }
```
alex-ju, line 93: nit: we're not saving anything by abbreviating 'value'
```suggestion
  onInput(newValue: string): void {
    this._value = newValue;
    this.args.onInput?.(newValue);
  }
```

packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -0,0 +1,248 @@
...
+    super(owner, args);
+
+    registerDestructor(this, () => this.observer?.disconnect());
+  }
+
+  modify(
+    element: HTMLElement,
+    positional: PositionalArgs<HdsCodeEditorSignature>,
+    named: NamedArgs<HdsCodeEditorSignature>
+  ): void {
```
alex-ju, line 96: 👏 
note to self: we should update other modifiers to make use of `ember-modifier` types

packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -0,0 +1,248 @@
...
+    // the intersection observer makes loading unreliable in tests
```
alex-ju, line 97: question: curious to hear more about the issues you experienced here

packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -0,0 +1,248 @@
...
+        return LANGUAGES[language].load();
+      } catch (error) {
+        warn(
+          `\`hds-code-editor\` modifier - Failed to dynamically import the CodeMirror language module for '${language}'. Error: ${JSON.stringify(
+            error
+          )}`,
+          {
+            id: 'hds-code-editor.load-language-task.import-failed',
+          }
+        );
```
alex-ju, line 146: 🙌

packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -0,0 +1,248 @@
...
+      language?: HdsCodeEditorLanguages;
+      value?: string;
+      onInput?: (newVal: string) => void;
+      onBlur?: (editor: EditorView, event: FocusEvent) => void;
+      onSetup?: (editor: EditorView) => unknown;
+    };
+  };
+}
+
+async function defineStreamLangugage(streamParser: StreamParser<unknown>) {
```
alex-ju, line 34: typo (needs update in a few places)
```suggestion
async function defineStreamLanguage(streamParser: StreamParser<unknown>) {
```

packages/components/src/components/hds/code-editor/index.hbs
alex-ju: Agree. We don't have to provide a default label (by providing one we risk having multiple identical names), but we should require one and leave the consumer decide the best name (we can aid with guidance).

alex-ju (commented):
Did a fresh new round of testing and left a few comments. Looks very good – it almost makes me code in it just for fun!

A few things that stood out for me and I'd like to highlight:
 - assertions and error handling looks much better after the recent updates; I particularly like the warning for language module fails, and error if CodeMirror fails.
 - intersection observer used to lazily load the editor when it comes into view – a good performance optimization
 - tasks with `ember-concurrency` preventing race conditions
 - dynamic imports for language modules used to load only the necessary code, which is good for reducing the initial bundle size

One thing that I discovered in my testing:
 - if HDS icons are lazy loaded, on a slow connection the loading icon is not visible as it loads after the CodeEditor is setup (when it's not visible/needed anymore). this was observed in `showcase`, have not tested in products, but I expect the same behaviour

packages/components/src/modifiers/hds-code-editor.ts
zamoore: 😬 🙇

packages/components/src/modifiers/hds-code-editor.ts
zamoore: That's a good idea. I'll make an issue.

.changeset/long-poets-wonder.md
```diff
@@ -0,0 +1,6 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Hds::CodeEditor` - Added new CodeMirror 6 supported code editor component
+`hds-code-editor` modifier - Added new code editor modifier which converts the element it is applied to into a CodeMirror6 code editor
```
KristinLBradley, line 6: Super nit: Should be consistent as to whether there is a space before the "6" or not in "CodeMirror6".

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,120 @@
...
+  background-color: var(--hds-code-editor-color-surface-primary);
+  border: 1px solid var(--hds-code-editor-color-border-strong);
+  border-radius: 6px;
+
+  &.hds-code-editor--is-full-screen {
+    position: fixed;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
```
KristinLBradley, line 25: Tip: You can replace the top, right, etc. styles with `inset: 0;` to set them all at once.

https://developer.mozilla.org/en-US/docs/Web/CSS/inset

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,120 @@
...
+    > :not(.hds-code-editor__title, .hds-code-editor__description):has(+ *) {
+      margin-right: 8px;
+    }
+  }
+
+  .hds-code-editor__title,
+  .hds-code-editor__description {
+    width: 100%;
+
+    &:has(+ *) {
```
KristinLBradley, line 60: Could be helpful to explain what kind of element(s) this could apply to & why it's needed.

KristinLBradley: I would also consider grouping the title & description together (wrapping with a div) then set spacing for them while providing another div whether custom content (if any) would get inserted. We could set the space between the title/description group & the custom content area but leave layout/spacing of elements within the custom content area to consumers.

Just a thought.

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,120 @@
...
+      margin-bottom: 12px;
+    }
+
+    &:has(+ .hds-code-editor__title, + .hds-code-editor__description) {
+      margin-bottom: 4px;
+    }
```
KristinLBradley, line 66: Comment: Btw, `:has` (& :not used above) are very powerful & useful but it can be difficult to reason out the behavior when reading through code so I would tend to avoid using if there are other ways to accomplish the same thing which may be clearer/simpler.

(This is just a general comment.)

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,120 @@
...
+  .hds-code-editor__header-content {
+    display: flex;
+    flex-grow: 1;
+    flex-wrap: wrap;
+    align-items: center;
+
+    // child element that is not the title/description and has subsequent siblings
+    > :not(.hds-code-editor__title, .hds-code-editor__description):has(+ *) {
+      margin-right: 8px;
+    }
```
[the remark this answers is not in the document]
KristinLBradley: Comment: Chaining together :not & :has like this is a little tricky to read and reason though.

KristinLBradley (commented):
Left some comments regarding some of the CSS spacing/layout structure that I thought could perhaps be simplified/improved. (Especially in regards to custom content.)

packages/components/src/styles/components/code-editor/index.scss
zamoore: Yeah, I struggled with this for a bit.

I originally had the title and description together in a div, but with the optional yielded content there were a number of vertical spacing issues. This is the best configuration I could come up with that works for all the different combinations of content with the specified spacing.

If you want to set something up that you think might work I'd be happy to pair on it or take a look.

zamoore: Also an issue was that if there wasn't a Title or Description that were yielded, we were left with an empty div that caused issues with the flex layout.

packages/components/src/styles/components/code-editor/index.scss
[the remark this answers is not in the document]
zamoore: I believe we're being pretty open-ended on what content can go here.

@majedelass - Can you provide some guidance on default spacing?

packages/components/src/styles/components/code-editor/index.scss
zamoore: I agree, and it's not a tool I typically reach for. Given the variability in content configurations in the header, I had to get a little more creative with selectors than I typically would.

packages/components/src/modifiers/hds-code-editor.ts
zamoore: Experienced some flakiness early on. I think the intersection observer is well tested enough that we can expect it to fire and test downstream of that.

alex-ju: Fair enough – thanks for expanding on this

packages/components/src/components/hds/code-editor/index.hbs
shleewhite: Ah sorry, I thought because the modifier is added to this element when you add the aria-label/labelledby it would apply to the div with role textbox. It looks like in reality, that div gets added inside this one so the textbox is still unlabelled.

Screenshot 2025-01-08 at 1 42 53 PM

I think you could use [the editorAttributes](https://codemirror.net/docs/ref/#view.EditorView^editorAttributes) option to customize what aria-attributes the editor has?

packages/components/src/styles/components/code-editor/index.scss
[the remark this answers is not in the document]
KristinLBradley: If we are being open-ended on custom content that can go there then I think we should leave spacing and layout of the content entirely up to the consumer. We could just provide a few typical examples with recommended spacing perhaps in the Guidance docs.

packages/components/src/styles/components/code-editor/index.scss
KristinLBradley: I tried checking out your branch to take a closer look but I can't get it to build & run. Could you see if you can rebase onto the latest `main` and get it to run?

showcase/app/styles/showcase-pages/code-editor.scss
KristinLBradley, line 1: If you don't need this file you should delete

packages/components/src/components/hds/code-editor/index.hbs
zamoore: Should be all set now

showcase/app/styles/showcase-pages/code-editor.scss
zamoore: I was getting errors without it for some reason, but I can try again. I assumed it was necessary for our build or something.

packages/components/src/styles/components/code-editor/index.scss
zamoore: Was having some issues after implementing some aria stuff. Should be good to go now.

showcase/app/styles/showcase-pages/code-editor.scss
alex-ju: If you remove it from [app.scss](https://github.com/hashicorp/design-system/pull/2573/files#diff-3d50f715ac106bace9fa3a330462db5f134047068a7464fdf8495d1a273d4d40R43) as well it should be fine

packages/components/src/styles/components/code-editor/index.scss
KristinLBradley: Thanks! I started experimenting a bit yesterday. If I come up with something that seems worthwhile I'll share.

KristinLBradley: Ok, here is an experimental spike branch I created, see what you think:
https://github.com/hashicorp/design-system/pull/2641

packages/components/src/styles/components/code-editor/theme.scss
zamoore: I'm going to move this to its own quick followup that we can do between merging this and releasing it. I started working on it, but given that it touches the CodeBlock component as well, it feels like distinct work.

zamoore: https://hashicorp.atlassian.net/browse/HDS-4347

showcase/app/styles/showcase-pages/code-editor.scss
zamoore: using it now thanks to kristin's changes

packages/components/src/styles/components/code-editor/index.scss
zamoore: I like it. Pulled all of it in and adjusted it to fit the existing code. Thanks!

packages/components/src/styles/components/code-editor/index.scss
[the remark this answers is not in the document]
zamoore: That makes sense to me. I kept that part of your suggestion from the other branch.

KristinLBradley:
@zamoore One case I forgot in my spike was Generic content together with a Title only & no Description. I would add another Showcase example and then add styling like below:

```
.hds-code-editor__title + .hds-code-editor__header-generic {
    margin-top: 12px;
}
```

packages/components/src/modifiers/hds-code-editor.ts
alex-ju: I drafted this task https://hashicorp.atlassian.net/browse/HDS-4349, but let me know if you already had one.

packages/components/src/styles/components/code-editor/theme.scss
```diff
@@ -0,0 +1,18 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// THEMING
+
+.hds-code-editor--theme-dark {
+  // COLORS
+
+  // Base UI colors:
+  --hds-code-editor-color-border-strong: rgba(178, 182, 189, 40%);
```
majedelass, line 12: We need to add an additional border color for primary:
```suggestion
  --hds-code-editor-color-border-strong: rgba(178, 182, 189, 40%);
  --hds-code-editor-color-border-primary: rgba(178, 182, 189, 20%);
```

This is from the component in Figma:
image

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  .hds-code-editor__header {
+    display: flex;
+    gap: 12px;
+    align-items: start;
+    padding: 16px;
+    background-color: var(--hds-code-editor-color-surface-faint);
+    border-bottom: 1px solid var(--hds-code-editor-color-border-strong);
```
majedelass, line 38: To lighter border:
```suggestion
    border-bottom: 1px solid var(--hds-code-editor-color-border-primary);
```

packages/components/src/modifiers/hds-code-editor/palettes/hds-dark-palette.ts
```diff
@@ -0,0 +1,17 @@
...
+export const HDS_CODE_BLOCK_GREEN = '#86ff13';
+export const HDS_CODE_BLOCK_ORANGE = '#ffa800';
+export const HDS_CODE_BLOCK_PURPLE = '#c76cff';
+export const HDS_CODE_BLOCK_RED = '#ff3b20';
+export const HDS_CODE_BLOCK_CYAN = '#32fff7';
+
+export const HDS_CODE_BLOCK_LINE_HIGHLIGHT = 'rgba(0, 74, 222, 0.2)';
+export const HDS_CODE_BLOCK_LINE_HIGHLIGHT_BORDER = '#1b5fe5';
+
+export const HDS_CODE_EDITOR_COLOR_BORDER_STRONG = 'rgba(178, 182, 189, 40%)';
```
majedelass, line 12: Add another const:
```suggestion
export const HDS_CODE_EDITOR_COLOR_BORDER_STRONG = 'rgba(178, 182, 189, 40%)';
export const HDS_CODE_EDITOR_COLOR_BORDER_PRIMARY = 'rgba(178, 182, 189, 20%)';
```

packages/components/src/modifiers/hds-code-editor/themes/hds-dark-theme.ts
```diff
@@ -0,0 +1,71 @@
...
+      color: HDS_CODE_EDITOR_COLOR_SURFACE_PRIMARY,
+    },
+    '.cm-content': {
+      padding: '16px 0',
+    },
+    '.cm-gutters': {
+      backgroundColor: HDS_CODE_EDITOR_COLOR_SURFACE_PRIMARY,
+    },
+    '.cm-gutter': {
+      borderRight: `1px solid ${HDS_CODE_EDITOR_COLOR_BORDER_STRONG}`,
```
majedelass, line 36:
```suggestion
      borderRight: `1px solid ${HDS_CODE_EDITOR_COLOR_BORDER_PRIMARY}`,
```

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,106 @@
...
+    color: var(--hds-code-editor-color-foreground-primary);
+    background-color: var(--hds-code-editor-color-surface-primary);
+  }
+
+  .hds-button {
+    color: var(--hds-code-editor-color-foreground-primary);
+    background-color: var(--hds-code-editor-color-surface-faint);
+    border: 1px solid var(--hds-code-editor-color-border-strong);
+
+    &:active,
```
majedelass, line 92: There should be an active color, we do this for the [CodeBlock too](https://hds-showcase-git-alex-ju-code-editor-codemirror-spike-hashicorp.vercel.app/components/code-block#copy-button).

```suggestion
    &:active {
        background-color: var(--hds-code-editor-color-surface-interactive-active);
    },
```

This seems to be the code block's variable too: 
image

packages/components/src/styles/components/code-editor/theme.scss
majedelass, line 17: Need to add the variable for our button's active state:
```suggestion
  --hds-code-editor-color-surface-primary: #0D0E12;
  --hds-code-editor-color-surface-interactive-active: #2B303C;
```

majedelass (commented):
Added some updates to the code to align with what we designed.

Added a few tokens and update their reference in the code.

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,106 @@
...
+      margin-top: 4px;
+    }
+  }
+
+  .hds-code-editor__description {
+    color: var(--hds-code-editor-color-foreground-faint);
+    
+    + .hds-code-editor__header-generic {
+      margin-top: 12px;
+    }
```
alex-ju, line 60: Issue: With the current styles, if the description is missing, there will be no space between Title and Generic.

Screenshot 2025-01-14 at 14 36 31


Suggestion: Add generic content examples to the Content section (all combinations, only Generic, Title+Description, Title+Generic, Title+Description+Generic, etc) using the `<Shw::Placeholder>` component (see an example in Alert below). This will help you spot any potential CSS issues and fix them. Having a button in the generic content is helpful to test its styles are correct, but that can be kept in separate section, 'Custom content' as you have it now.

Screenshot 2025-01-14 at 14 39 41

packages/components/src/components/hds/code-editor/index.hbs
```diff
@@ -0,0 +1,69 @@
...
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<div
+  id={{this._id}}
+  class={{this.classNames}}
+  {{! @glint-expect-error - https://github.com/josemarluedke/ember-focus-trap/issues/86 }}
+  {{focus-trap isActive=this._isFullScreen}}
+  {{this._handleEscape}}
+  ...attributes
```
majedelass, line 12: I believe this is missing the `isStandalone` property that if is `false` will remove the border radius from the component. Similar to the [CodeBlock](https://helios.hashicorp.design/components/code-block?tab=code#codeblock). Unless I'm mistaken, we want this in two places, the modifier and the component?

alex-ju: Indeed taking a similar approach with `CodeBlock` in terms of how `isStandalone` being set to `true` by default and setting it to false will remove the rounder border makes the most sense in keeping the two APIs in sync. It would be great to have [a section in the showcase](https://hds-showcase.vercel.app/components/code-block#standalone) to test this as well.

zamoore: I can add this one in in any case

zamoore: Added

majedelass: It looks like it was only added in the .ts file but not the .hbs file?

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,113 @@
...
+    background-color: var(--hds-code-editor-color-surface-faint);
+    border: 1px solid var(--hds-code-editor-color-border-strong);
+
+    &:active {
+      background-color: var(--hds-code-editor-color-surface-interactive-active)
+    }
+  
+    &:focus,
+    &:hover {
+      color: var(--hds-code-editor-color-foreground-high-contrast);
```
majedelass, line 102: I found one last thing from the design side:
```suggestion
      color: var(--hds-code-editor-color-foreground-high-contrast);
      background-color: var(--hds-code-editor-color-surface-primary);
```

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,113 @@
...
+    background-color: var(--hds-code-editor-color-surface-primary);
+  }
+
+  .hds-button {
+    color: var(--hds-code-editor-color-foreground-primary);
+    background-color: var(--hds-code-editor-color-surface-faint);
+    border: 1px solid var(--hds-code-editor-color-border-strong);
+
+    &:active {
+      background-color: var(--hds-code-editor-color-surface-interactive-active)
```
majedelass, line 97: Just adding a semicolon here
```suggestion
      background-color: var(--hds-code-editor-color-surface-interactive-active);
```

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,113 @@
...
+    }
+  
+    &:focus,
+    &:hover {
+      color: var(--hds-code-editor-color-foreground-high-contrast);
+      
+      .hds-button__icon {
+        color: var(--hds-code-editor-color-foreground-high-contrast);
+      }
```
majedelass, line 106: I think remove this too?

```suggestion

```

packages/tokens/dist/cloud-email/tokens.scss
```diff
@@ -1,10 +1,6 @@
 
 // Do not edit directly, this file was auto-generated.
 
-$token-border-radius-x-small: 3px;
```
shleewhite, line 4: Question: is this supposed to change?

shleewhite (approved):
Looks great!!!

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,109 @@
...
+    &:focus,
+    &:hover {
+      background-color: var(--hds-code-editor-color-surface-primary);
+    }
+
+    .hds-button__icon {
+      color: var(--hds-code-editor-color-foreground-primary);
+    }
+  }
+}
```
majedelass, line 109: I found two minor issues, but not sure where to place this, so doing it here because I figure its related to something here.

The CopyButton on hover is blue:
image

When using my mouse for either of the two `.hds-button` elements, the active state isn't showing up, but when I trigger it from the console it does. 

My gut instinct is telling me if you reverse the order of the hover and active it will show (because of the cascading effect -- the hover is taking priority over the active).

packages/components/src/modifiers/hds-code-editor/themes/hds-dark-theme.ts
```diff
@@ -0,0 +1,72 @@
...
+    '.cm-lineNumbers': {
+      color: HDS_CODE_EDITOR_COLOR_FOREGROUND_FAINT,
+    },
+    '.cm-lineNumbers .cm-gutterElement': {
+      borderLeft: '4px solid transparent',
+      padding: '0 16px',
+    },
+    '.cm-gutterElement.cm-activeLineGutter': {
+      backgroundColor: HDS_CODE_EDITOR_COLOR_SURFACE_PRIMARY,
+      color: HDS_CODE_EDITOR_COLOR_FOREGROUND_HIGH_CONTRAST,
```
majedelass, line 47: I'm not sure if this is what's affecting this, but when not focused in the Editor, the line number (in this case its 1) is still white. Wondering if we can keep that as the rest of the numbers until the user is focused in?
image

majedelass (commented):
There are three minor things I caught. Let me know when you address them so I can approve the PR 👍

packages/components/src/styles/components/code-editor/index.scss
alex-ju: A tangent on this. The success/error icon has the same color (`--hds-code-block-color-foreground-primary`) whereas in the CodeBlock we use `--hds-code-block-color-foreground-success` and `--hds-code-block-color-foreground-critical`

alex-ju (commented):
I tested the functionality again. One thing I noticed is that in full-screen mode the border-radius is preserved (not sure if it was intentional or not). Otherwise all looks good! Happy to get it in, and if anything comes up in testing we can address separately.

alex-ju (approved):

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  .hds-button {
+    color: var(--hds-code-editor-color-foreground-primary);
+    background-color: var(--hds-code-editor-color-surface-faint);
+    border: 1px solid var(--hds-code-editor-color-border-strong);
+
+    &:active {
+      background-color: var(--hds-code-editor-color-surface-interactive-active)
+    }
```
majedelass, line 98: Active still isn't showing up properly. I believe this needs to be moved below the hover pseudo class so it will show up properly on mouse click.

majedelass (approved):
I'm approving because there is one last minor thing with the active state (in CSS). I'd move the active state below the hover state in CSS so it will show up properly on mouse click.

packages/components/src/styles/components/code-editor/index.scss
zamoore: Adressing in a followup

PR #2574, opened 2024-11-20
`Dropdown` - Focus ring on click for `ToggleIcon` fix

dchyun:
### :pushpin: Summary

If merged, this PR would fix an open issue with the `ToggleIcon` variation of the `Dropdown` where the focus ring does not appear on mouse click on the button. 

### 🛠️ Detailed description

Currently the `ToggleIcon` variation of the `Dropdown` only shows a focus ring when the menu is opened, if it is opened via the keyboard, not the mouse. This interaction differs from that of the `ToggleButton` variation of the `Dropdown` and from the generic `Button` component. 

The approach for this variation has been update to align with those of other button components in the library. It now will show the focus ring on both mouse click and keyboard interaction. 

Current behavior on mouse click
Screenshot 2024-11-20 at 9 09 42 AM

New behavior on mouse click
Screenshot 2024-11-20 at 9 09 35 AM

### :link: External links

Jira ticket: [HDS-4096](https://hashicorp.atlassian.net/browse/HDS-4096)

***

### 👀 Component checklist

[HDS-4096]: https://hashicorp.atlassian.net/browse/HDS-4096?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/styles/components/dropdown.scss
+++ b/packages/components/src/styles/components/dropdown.scss
@@ -63,6 +63,10 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
   &.mock-active {
     background-color: var(--token-color-surface-interactive-active);
     border-color: var(--token-color-border-strong);
+
+    &::before {
+      border-color: transparent;
+    }
   }
 
   &:disabled,
--- a/packages/components/src/styles/components/dropdown.scss
+++ b/packages/components/src/styles/components/dropdown.scss
@@ -52,6 +52,7 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
   &:focus,
   &.mock-focus {
     @include hds-button-state-focus();
+    border-color: var(--token-color-focus-action-internal);
 
     &::before {
       border-color: var(--token-color-focus-action-external);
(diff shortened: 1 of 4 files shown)


alex-ju (commented):
The active (+ focus) state should ideally match the one we have for Dropdown Toggle (and Button); which is essentially the active state.

dropdown-focus

packages/components/src/styles/components/dropdown.scss
```diff
@@ -28,6 +28,7 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
 // TOGGLE/ICON
 
 .hds-dropdown-toggle-icon {
+  position: relative;
```
alex-ju, line 31: noting here that this element does not contain any custom content (we control all the elements nested within) to there is not risk of introducing layout/positioning issue to our consumers.

alex-ju (approved):
Works as expected and the CSS implementation seems in line with the Button component! Let's get some design eyes on it and then we can get it in.

zamoore (approved):
Nice work

PR #2584, opened 2024-11-25
 Helios website - update body font (HDS-4046)

KristinLBradley:
### :pushpin: Summary

If merged, this PR will update the body font for the Helios website.

**Preview:** https://hds-website-git-hds-4046-update-website-font-hashicorp.vercel.app/

**_BEFORE:_**
image

**_AFTER:_**
image

### :link: External links

Jira ticket: [HDS-4046](https://hashicorp.atlassian.net/browse/HDS-4046)

***

### 👀 Component checklist

[HDS-4046]: https://hashicorp.atlassian.net/browse/HDS-4046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/styles/typography/mixins.scss
+++ b/website/app/styles/typography/mixins.scss
@@ -14,12 +14,10 @@
   @if ($family == "hashicorp-sans") {
     font-family: hashicorp-sans-variable, Geneva, Tahoma, Helvetica, Verdana, sans-serif;
     font-synthesis: none; // Safari makes this variable font too bold, compared to Chrome, and this fixes the issue
-  } @else if ($family == "sans") {
-    font-family: metro-web, Metro, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen, Ubuntu, Cantarell, "Open Sans", "Helvetica Neue", sans-serif;
   } @else if ($family == "mono") {
     font-family: ui-monospace, Menlo, Consolas, monospace;
   } @else if ($family == "system") {
-    font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Oxygen, Ubuntu, Cantarell, "Open Sans", "Helvetica Neue", sans-serif;
+    font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
   }
 }
 
@@ -140,7 +138,7 @@
 // BASE
 
 @mixin doc-font-style-body {
-  @include doc-font-family("sans");
+  @include doc-font-family("system");
   font-size: 1.063rem; // 17px;
   line-height: 1.765; // 30px
 }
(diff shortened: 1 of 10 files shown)


website/app/styles/typography/mixins.scss
```diff
@@ -14,12 +14,10 @@
   @if ($family == "hashicorp-sans") {
     font-family: hashicorp-sans-variable, Geneva, Tahoma, Helvetica, Verdana, sans-serif;
     font-synthesis: none; // Safari makes this variable font too bold, compared to Chrome, and this fixes the issue
-  } @else if ($family == "sans") {
```
didoo, line 17: [todo] If we decided to not use anymore the "metro web" custom font, we should remove it also from the `assets/fonts` folder, and from the `website/app/styles/typography/font-faces.scss` file.

didoo: @heatherlarsen @KristinLBradley general comment: for me there was a specific reason to use a different font from the HDS body as body of the website: to visually differentiate what was an HDS component (inlined as image, or as code snippet preview) from what is actual body copy of the website.

In the same way that we don't use HDS component to build the HDS website, we should not use the HDS font (even if it's the system font) to build the HDS website.

But that's my opinion.

didoo (commented):
Left a couple of comments

shleewhite (approved):
Nice!

website/app/styles/typography/mixins.scss
didoo: Also, I _think_ that the custom font allowed a bit more granularity in terms of font-weight than the system font.

KristinLBradley: Your reasoning above makes sense to me but will wait on hearing from @heatherlarsen and other @hashicorp/hds-design people on what they want to do. (If we want to keep this change then I'll go ahead and delete the metro web font file)

jorytindall: I tend to agree that using a body font that is different from what is used in our components for differentiation purposes is logical. I took a brief skim through the deploy preview and have a couple of takeaways:

To me (and this may be because we spend a lot time looking at the site), there is a pretty noticeable difference between the two fonts, especially in the navigation (side bar specifically). The weights might need some additional adjustment, but they look very different.

**Before**
Screenshot 2024-11-25 at 1 08 07 PM

**After**
Screenshot 2024-11-25 at 1 08 33 PM

Additionally, the type-scale is different between what is on the website vs what is used in our components, I get a bit of a feeling like there's a mistake when components are are paired with body text. It's almost like they are intended to be the same, but are off by a small amount which stands out to me.

Again, might be due to visual familiarity with the website, and I'm not proposing a different typeface, just noting my first impression at the visual change.

LilithJames-HDS: I agree with @jorytindall here. I think the difference is noticeable since we look at the site a lot but would still be enough for a casual user to notice something is "off"

I also agree with @didoo on wanting to differentiate the body content with components but I don't think this font is quite "different enough" to do that.

KristinLBradley:
Closing PR as @hashicorp/hds-design has decided to re-examine the need for the web font update.

PR #2588, opened 2024-12-04
Table: add borders around the cells

shleewhite:
### :pushpin: Summary

If merged, this PR would add a full border around all the cells. This is in an effort to improve the styling when cells span multiple rows or columns.

The existing borders were made with the `::before` pseudo selector to make the borders not the full height and a border around the whole table. Now, it just uses borders on the cells.

#### Before
Screenshot 2024-12-04 at 10 27 28 AM
Screenshot 2024-12-04 at 10 27 41 AM

#### After
Screenshot 2024-12-04 at 10 21 17 AM
Screenshot 2024-12-04 at 10 21 32 AM

### :link: External links

Jira ticket: [HDS-3956](https://hashicorp.atlassian.net/browse/HDS-3956)
Figma file: https://www.figma.com/design/FfPRXgqDb67jlXNhMrvfub/ME%2BAG%2FAdvancedTable?node-id=16727-59704

***

### 👀 Component checklist

[HDS-3956]: https://hashicorp.atlassian.net/browse/HDS-3956?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/gorgeous-days-call.md
+++ b/.changeset/gorgeous-days-call.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Table` - Updated the visual design of `Table` cells by adding borders, making them more distinguishable when spanning rows or columns.
--- a/packages/components/src/styles/components/table.scss
+++ b/packages/components/src/styles/components/table.scss
@@ -35,6 +35,14 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
   table-layout: fixed;
 }
 
+// table border
+
+.hds-table, 
+.hds-table__th, 
+.hds-table__td {
+  border: calc(#{$hds-table-border-width} / 2) solid $hds-table-border-color;
+  // box-shadow: 0 0 0 calc(#{$hds-table-border-width} / 2) $hds-table-border-color;
+}
 
 // ----------------------------------------------------------------
 
@@ -48,13 +56,13 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
     .hds-table__th {
       padding: $hds-table-cell-padding-medium;
       text-align: left;
-      border-top: $hds-table-border-width solid $hds-table-border-color;
-      border-left: $hds-table-border-width solid $hds-table-border-color;
+      // border-top: $hds-table-border-width solid $hds-table-border-color;
+      // border-left: $hds-table-border-width solid $hds-table-border-color;
     }
 
-    .hds-table__th:last-child {
-      border-right: $hds-table-border-width solid $hds-table-border-color;
-    }
+    // .hds-table__th:last-child {
+    //   border-right: $hds-table-border-width solid $hds-table-border-color;
+    // }
 
     // horizontal alignment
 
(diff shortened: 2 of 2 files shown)


.changeset/gorgeous-days-call.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
```
shleewhite, line 2: @alex-ju based on your survey, should this be a patch or a minor?

alex-ju: the survey was somewhat nonconclusive on this (at least not a consensus). minor seems appropriate to me.

shleewhite: Yeah, I think @andgen404 will communicate this more widely on slack before we merge so people know its happening.

packages/components/src/styles/components/table.scss
```diff
@@ -225,7 +210,10 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
     .hds-table--density-tall & {
       padding: $hds-table-cell-padding-tall;
     }
+  }
 
+  .hds-table__td:last-child {
+    border-right: $hds-table-border-width solid $hds-table-border-color;
```
KristinLBradley, line 216: Would something like this work for the border styling? (I don't know how well it would work cross-browser though or with the table grid structure...)

```
.hds-table, 
.hds-table__th, 
.hds-table__td {
    border: calc(#{$hds-table-border-width} / 2) solid $hds-table-border-color;
}
```

shleewhite: Yeah, I tried that originally and it looks fine in Safari but in Chrome it results in a 2px border.

shleewhite: Screenshot 2024-12-16 at 10 50 51 AM

shleewhite: Updated: testing more, if I do border-collapse: collapse, it works as expected across browsers - but the border-radius is lost.

KristinLBradley: Thanks for testing. I tested just by editing within the Chrome browser. I first unchecked all the existing border styles then added back a .5px border to everything. I compared with another tab in which I left the styles unedited and the borders seemed to be the same width.

image

shleewhite: Funny! The borders also look incorrect in percy: https://percy.io/HashiCorp/web/hds-components/builds/37844826/changed/2058208695?browser=chrome&browser_ids=59&group_snapshots_by=similar_diff&subcategories=unreviewed%2Cchanges_requested&utm_source=github_status_private&viewLayout=overlay&viewMode=original&width=1280&widths=375%2C1280

KristinLBradley: It does cause a slight shift of some borders but it doesn't seem to effect the width of them.

I created a simplified test and checked in Chrome, Safari, Firefox, & Edge and applying .5px width borders seemed to work well in all: https://codepen.io/KristinB/pen/NPKdgxv/f5c2ee448f0e40e60d806252b57a9691

packages/components/src/styles/components/table.scss
```diff
@@ -225,9 +204,8 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
     .hds-table--density-tall & {
       padding: $hds-table-cell-padding-tall;
     }
-
   }
-
+
```
alex-ju, line 208:
```suggestion

```

alex-ju (approved):

packages/components/src/styles/components/table.scss
shleewhite: Just to wrap up this thread: it turns out it is because of my display. We are moving forward with this solution because the potential of a thicker border is worth the simpler/more flexible implementation.

alex-ju (approved):

KristinLBradley (approved):

andgen404 (approved):
Lgtm! 🎉

PR #2589, opened 2024-12-05
[Spike] Get theming “support” in the Showcase app

didoo:
# 🚨 DO NOT MERGE 🚨

### :pushpin: Summary

Branch used for exploring what needs to be done to have "support" for theming in the Showcase app.

In the process, we will use the Showcase app as proxy to understand what a consumer would need from HDS to use the themes provided by the system (eg. should HDS provide a "toggle" or "switcher" for themes? should provide a service that returns the themes, sets the theme in the hosting app, saves the user preference in a local storage?)

### :link: External links

Jira tickets: 
- Epic: https://hashicorp.atlassian.net/browse/HDS-3595
- Task: https://hashicorp.atlassian.net/browse/HDS-4232

### DIFF
--- a/showcase/app/components/shw/theme/switcher.gts
+++ b/showcase/app/components/shw/theme/switcher.gts
@@ -1,0 +1,41 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+import { on } from '@ember/modifier';
+import { action } from '@ember/object';
+
+interface ShwThemeSwitcherSignature {
+  // Args: {};
+  Element: HTMLDivElement;
+}
+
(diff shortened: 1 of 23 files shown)


didoo:
The first iteration (this one) is now outdated (we're working in the branch `spike-tokens-aggregated-scrappy`, so we can close it.

showcase/app/components/shw/theme/switcher.gts
```diff
@@ -0,0 +1,74 @@
...
+
+interface ShwThemeSwitcherSignature {
+  // Args: {};
+  Element: HTMLDivElement;
+}
+
+export default class ShwThemeSwitcher extends Component<ShwThemeSwitcherSignature> {
+  @service declare readonly theming: ShwThemingService;
+
+  @action
```
aklkv, line 21: in `.gts` files `@action` can be replaced with an arrow function

PR #2109, opened 2024-05-16
feat(showcase): add typescript tooling

aklkv:
### :pushpin: Summary

* add typescript tooling
* update app blueprint
* fix in what order GHA runs build/lint/test for showcase as we need components and flight-icons to be built

### :link: External links

Jira ticket: [HDS-XXX](https://hashicorp.atlassian.net/browse/HDS-XXX)
Figma file: [if it applies]

***

### 👀 Component checklist

- [ ] Percy was checked for any visual regression
- [ ] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#templates-for-npm-packages))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #2592, opened 2024-12-09
Upgrade `@glint/environment-ember-loose` to `1.5.0`

alex-ju:
### :pushpin: Summary

Upgrade `@glint/environment-ember-loose` to `1.5.0` in `devDependencies` across the monorepo.

We used to rely on a patch introduced in #2109, but there's no need to use it anymore. It also clears the warnings in yarn.

Before

Screenshot 2024-12-09 at 21 43 48

After

Screenshot 2024-12-09 at 21 44 50

***

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
   },
   "packageManager": "yarn@4.4.0",
   "resolutions": {
-    "broccoli-asset-rewrite@^2.0.0": "patch:broccoli-asset-rewrite@npm%3A2.0.0#./.yarn/patches/broccoli-asset-rewrite-npm-2.0.0-c4ce42084a.patch",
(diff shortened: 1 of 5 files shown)


alex-ju:
This PR unveils ~1589 lint errors in `showcase`; likely related to https://hashicorp.atlassian.net/browse/HDS-4090.
Will mark it as draft for now.

aklkv:
@alex-ju I am not sure if my patch was release as part of `v1.5` since you have removed it it might be the reason for so many lint errors. If you want I can help to reapply this patch

alex-ju:
> @alex-ju I am not sure if my patch was released as part of `v1.5` here is [ref](https://github.com/typed-ember/glint/blob/1.5.0/packages/environment-ember-loose/-private/environment/index.ts#L65) where the code I have removed in my patch is still in place, since you have removed it it might be the reason for so many lint errors. If you want I can help to reapply this patch, or we just skip this upgrade all together for now.

right, I was under the impression that we used the patch for an upcoming feature. thanks for flagging this! If I understand the patch correctly, we'd need to use pod-style templates to be able to move from the patch to v1.5? or is it more than just that?

PR #2596, opened 2024-12-11
`Table`: make checkbox aria-labels stay the same regardless of checkbox state

shleewhite:
### :pushpin: Summary

If merged, this PR would update the select all and select row checkbox labels so the visually hidden label is always `Select {selectionAriaLabelSuffix}` and update the default of selectionAriaLabelSuffix to be 'row'.

It is not compliant to have the label of a form input change based on the state of the input because it makes it harder for users to find again / understand what the input is for. Instead, it is better to keep the input name consistent and let the screen reader use the html attributes on the input to convey the state (in this case, checked or unchecked).

**Before**

Screenshot 2024-12-11 at 3 29 32 PM
Screenshot 2024-12-11 at 3 29 23 PM
Screenshot 2024-12-11 at 3 29 14 PM
Screenshot 2024-12-11 at 3 29 04 PM

**After**
Screenshot 2024-12-11 at 3 32 30 PM
Screenshot 2024-12-11 at 3 32 22 PM
Screenshot 2024-12-11 at 3 32 10 PM
Screenshot 2024-12-11 at 3 31 53 PM

### :link: External links

Jira ticket: [HDS-4258](https://hashicorp.atlassian.net/browse/HDS-4258)

***

### 👀 Component checklist

[HDS-4258]: https://hashicorp.atlassian.net/browse/HDS-4258?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/showcase/tests/integration/components/hds/table/index-test.js
+++ b/showcase/tests/integration/components/hds/table/index-test.js
@@ -627,6 +627,12 @@ module('Integration | Component | hds/table/index', function (hooks) {
 
     assert.dom(selectAllCheckboxSelector).hasAria('label', 'Select all rows');
     assert.dom(rowCheckboxesSelector).hasAria('label', 'Select row');
+
+    await click(selectAllCheckboxSelector);
+    await click(rowCheckboxesSelector);
+
+    assert.dom(selectAllCheckboxSelector).hasAria('label', 'Select all rows');
+    assert.dom(rowCheckboxesSelector).hasAria('label', 'Select row');
   });
 
   test('it renders the expected `aria-label` for rows with `@selectionAriaLabelSuffix`', async function (assert) {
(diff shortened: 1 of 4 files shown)


showcase/tests/integration/components/hds/table/index-test.js
```diff
@@ -606,7 +606,30 @@ module('Integration | Component | hds/table/index', function (hooks) {
...
+          <B.Tr @selectionKey={{B.data.id}}>
+            <B.Td>{{B.data.artist}}</B.Td>
+            <B.Td>{{B.data.album}}</B.Td>
+            <B.Td>{{B.data.year}}</B.Td>
+          </B.Tr>
+        </:body>
+      </Hds::Table>
+    `);
+
+    assert.dom(selectAllCheckboxSelector).hasAria('label', 'Select all rows');
```
didoo, line 628: [question] should be test that the label doesn't change when selected/unselected?

shleewhite: Good point! Added explicit checks for both tests.

dchyun (approved):

dchyun:
Is this something we would want to add for 4.15?

shleewhite:
> [Question] Is this something we would want to add for 4.15?

@dchyun it can go out in the next release with all of the other table/advanced table stuff

KristinLBradley (approved):
Nice!

PR #2615, opened 2024-12-18
`AdvancedTable`: create component

shleewhite:
### :pushpin: Summary

If merged, this PR would add the AdvancedTable component and related sub-components

**[AdvancedTable showcase](https://hds-showcase-git-hds-3965-advanced-table-clean-hashicorp.vercel.app/components/advanced-table)**

The easiest way to review this would be to step through each commit. I have also left a couple comments/questions inline.

#### **feat: copy Hds::Table and rework to Hds::AdvancedTable**
Changes of note:
* component, class, and type names change
* components use `div` elements with roles instead of table elements
* styled with CSS Grid
  * note: some styles changed because of the grid, for example the cell background color is set on each cell instead of the row because the row now has `display: contents` which doesn't allow background-color
* Th scope attribute now maps to role "rowheader" or "columnheader"

#### **feat: copy tests from Table and rework for AdvancedTable**
Changes of note:
* changed selectors to the new classes
* omitted tests where the Table was defined "inline" (aka without model/columns)
* Created tests for rowspan/colspan that make sure aria-rowspan/colspan and grid-row/column are correct
* updated scope tests to check for correct role

#### **feat: copy showcase from Table and rework for AdvancedTable**
Changes of note:
* Omitted examples where Table was defined "inline" (aka without model/columns)
* Examples where it was impossible to have a model/columns (ie. Th/Td isolation tests), built fake advanced table with HTML and the component classes
* Copied relevant javascript for the route and controller from Table

#### **feat: add keyboard navigation + related tests**
Changes of note:
* new helpers file that exports
  * a function for when a grid cell is inserted in the dom. the function adds event handlers to any focusable content inside the cell for keyboard navigation
  * a function that adds keyboard events for grid cells
* directions for how to use the keyboard navigation in the showcase
* a new test file for the helpers file

#### **feat: add @hasStickyHeader argument and related showcase example, test**
Changes of note:
* adds @hasStickyHeader argument to advanced table
* adds an intersecton observer so the border radius can change when the header is pinned

#### **feat: add nested rows, @childrenKey arg**
Changes of note:
* adds new ExpandableTrGroup component
* AdvancedTable infers if there are nested rows based on the @childrenKey argument
* Adds guardrails so you can't have nested rows with isStriped argument, isSelectable argument, or if the columns are sortable
  * Note: these can be added later, but design/eng have found that there are not any use cases for them yet 

#### **chore: skip sticky header test for now**
Will improve the sticky header test as a follow up pr.

#### **fix: add cleanup for IntersectionObserver**
* Adds a `will-destroy` lifecycle event to clean up the intersection observer.
* Doesn't add the intersection observer if in a testing environment because it isn't defined. Can circle back to this at the same time as the stikcy header test.

#### **feat: use ember-focus-trap and tabbable libraries**
* adds tabbable as a dependency (doesn't increase the bundle size because it is already a dependency of ember-focus-trap)
* use tabbable instead of custom functions to find all focusable or tabbable elements
* use ember-focus-trap to handle focus trapping within the th, td, and thSort components
* fix keyboard bug @alex-ju found with expandable rows by only doing the arrow key navigation if the target is the cell (was previously being run on focusable elements inside the cell too, which broke keyboard pattern)

### :link: External links

Jira ticket: [HDS-3965](https://hashicorp.atlassian.net/browse/HDS-3965)
Figma file: [if it applies]

***

### 👀 Component checklist

[HDS-3965]: https://hashicorp.atlassian.net/browse/HDS-3965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


showcase/app/templates/components/advanced-table.hbs
```diff
@@ -0,0 +1,1710 @@
...
+        <SF.Item @label="Indeterminate {{if bool ' with sort'}}">
+
+          <div class="hds-advanced-table hds-advanced-table--density-medium" {{style gridTemplateColumns="auto 1fr"}}>
+            <div class="hds-advanced-table__thead">
+              <div class="hds-advanced-table__tr">
+                <Hds::AdvancedTable::ThSelectable
+                  @selectionScope="col"
+                  @isSelected={{true}}
+                  @onClickSortBySelected={{if bool this.noop}}
+                  @didInsert={{this.mockIndeterminateState}}
```
shleewhite, line 1345: Question: if you look at the preview for this, the indeterminate state example is broken. does anyone have any thoughts as to what could be going on? 

Screenshot 2024-12-18 at 5 11 33 PM

showcase/tests/integration/components/hds/advanced-table/index-test.js
```diff
@@ -0,0 +1,951 @@
...
+    </B.Tr>
+  </:body>
+</Hds::AdvancedTable></div>`
+    );
+
+    assert
+      .dom('#data-test-advanced-table .hds-advanced-table__thead')
+      .hasClass('hds-advanced-table__thead--sticky');
+
+    // commenting out for now because it is flaky
```
shleewhite, line 326: Question: this test worked... fairly reliably when I ran `ember t -s`, but with just `ember t` it always fails. Should I try to debug more? or is it ok to not test that the pinned class gets added when you scroll with a sticky header.

packages/components/src/components/hds/advanced-table/index.hbs
```diff
@@ -0,0 +1,107 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+<div
+  class={{this.classNames}}
+  ...attributes
+  role="grid"
+  aria-describedby={{this._captionId}}
```
shleewhite, line 10: @MelSumner heads up: I ended up using aria-describedby instead of aria-description for the advanced table caption because it needs to be aria-live="polite"

packages/components/src/components/hds/advanced-table/helpers.ts
```diff
@@ -0,0 +1,246 @@
...
+          event.preventDefault();
+          newTarget = allCellFocusableChildren[0];
+        }
+
+        if (newTarget) {
+          newTarget.focus();
+        }
+      }
+
+      if (key === 'Escape') {
```
shleewhite, line 55: @MelSumner another heads up, after looking into the options for the issue with escape closing both the tooltip and re-focusing the cell - there is no option that scales well here to try and prevent that. It would be very hacky (like making a new tooltip just for advanced table) and wouldn't work if a consumer put something in a cell that also has an escape key handler.

Leaving the behavior as is.

packages/components/src/components/hds/advanced-table/helpers.ts
```diff
@@ -0,0 +1,246 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import type { HdsAdvancedTableTdSignature } from './td';
+import type { HdsAdvancedTableThSignature } from './th';
+
+const getAllFocusableChildren = (element: HTMLElement): NodeListOf<Element> => {
+  return element.querySelectorAll(
+    'button:not([disabled]), a[href], input:not([disabled]), select:not([disabled]), textarea:not([disabled]), [tabindex]:not([tabindex="-1"]):not([disabled])'
```
alex-ju, line 11: noting here that this list is not exhaustive, but rather elements we expect to find in a table

packages/components/src/components/hds/advanced-table/helpers.ts
```diff
@@ -0,0 +1,246 @@
...
+
+const handleGridCellChildKeyPress = (event: KeyboardEvent): void => {
+  const { key, target } = event;
+
+  if (target instanceof HTMLElement) {
+    const cell = getParentCell(target);
+    if (cell instanceof HTMLElement) {
+      const allCellFocusableChildren = Array.from(
+        getAllFocusableChildren(cell)
+      ) as HTMLElement[];
```
alex-ju, line 33: question: would this list become obsolete/outdated if elements are added to/removed from this cell?

showcase/tests/integration/components/hds/advanced-table/index-test.js
```diff
@@ -202,6 +202,39 @@ module('Integration | Component | hds/advanced-table/index', function (hooks) {
       .hasAttribute('aria-label', 'data test table');
   });
 
+  test('it should render with a CSS class appropriate for the @hasStickyHeader argument', async function (assert) {
```
alex-ju, line 302:
```suggestion
  skip('it should render with a CSS class appropriate for the @hasStickyHeader argument', async function (assert) {
```

showcase/tests/integration/components/hds/advanced-table/index-test.js
alex-ju: if you'd like, you can replace `test(` with `skip(` to further investigate a solution but not block the PR on it

packages/components/src/components/hds/advanced-table/th-button-expand.hbs
```diff
@@ -0,0 +1,19 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+{{! template-lint-disable no-unsupported-role-attributes }}
+{{! ember template lint doesnt support ARIA 1.3 yet, including aria-description - https://github.com/A11yance/aria-query/pull/557 }}
```
alex-ju, line 6: 👍

showcase/package.json
```diff
@@ -75,6 +75,7 @@
     "ember-concurrency": "^4.0.2",
     "ember-deep-tracked": "^2.0.0",
     "ember-load-initializers": "^2.1.2",
+    "ember-math-helpers": "^3.0.0",
```
alex-ju, line 78: question: I've not seen this in use; am I missing something?

alex-ju (commented):
Thanks for cleaning-up the git history for this work! 🙌 

I did a first high-level pass, skimming through the code and checking the component in showcase. Here are my main findings so far:

**Issue**: keyboard navigation + expandable rows: navigate inside a cell (pressing Enter) and moving focus from the expand button to the next in-cell element (by pressing the right arrow key); at this stage pressing escape will not return focus to the table cell; the user has to return to the expand button and then press escape.

Screenshot 2024-12-19 at 14 22 27

**Issue**: keyboard navigation + sticky header: navigating upwards (pressing the up key) results in a tricky state where the focused cell is overlapping the header

Screenshot 2024-12-19 at 14 21 44

**Issue**: keyboard navigation: in-cell navigation works as expected for left/right arrows, but moves the focus outside the cell when up/down arrows are pressed.

showcase/package.json
shleewhite: Its just used for this example in the showcase: https://hds-showcase-git-hds-3965-advanced-table-clean-hashicorp.vercel.app/components/advanced-table#with-nested-rows-and-cells-that-span-rows

https://github.com/hashicorp/design-system/blob/0ecbd25af1e65f290eee90d525f8acff0228f65c/showcase/app/templates/components/advanced-table.hbs#L126

alex-ju: right, I missed that instance – thanks!

packages/components/src/components/hds/advanced-table/helpers.ts
shleewhite: Yeah, I was actually going to ask about this. I know we use ember-focus-trap, which is dependent on [focus-trap](https://github.com/focus-trap/focus-trap?tab=readme-ov-file) and [tabbable](https://github.com/focus-trap/tabbable).

I can't use ember-focus trap in this implementation, but would it make sense to use normal focus-trap and tabbable to be more confident in the focus trapping within the cell and finding all focusable children?

alex-ju: I believe it's worth exploring it, considering it may reduce the code we need to maintain, and we use `ember-focus-trap` elsewhere, so it's not increasing our bundle size.

alex-ju: It may also solve the issue flagged above as 'keyboard navigation + expandable rows'

showcase/app/templates/components/advanced-table.hbs
```diff
@@ -0,0 +1,1712 @@
...
+                December 15, 2022,<br />16:40:09 PM
+              </Hds::TooltipButton>
+            </Hds::AdvancedTable::Td>
+          </div>
+        </div>
+      </div>
+    </SF.Item>
+  </Shw::Flex>
+
+  <Shw::Flex @label="Vertical alignment" as |SF|>
```
shleewhite, line 1516: Question: I think the vertical alignment got a little wonky along the way.... maybe the grid styles affected it? Looking at the showcase, the CSS property is being added but the top aligned example doesn't look right. What is expected with this argument?
Screenshot 2024-12-19 at 2 23 37 PM

packages/components/src/components/hds/advanced-table/helpers.ts
shleewhite: Update: I refactored so the cells use ember-focus-trap and the helpers functions use tabbable now! Also, the commit fixes the issues with arrow keys breaking the nav when youre inside of a cell (tldr: the arrow keys should only work when focused on a cell, so now you can only tab between elements inside of a cell)

zamoore (approved):

packages/components/src/components/hds/advanced-table/index.hbs
```diff
@@ -0,0 +1,76 @@
...
+        {{/if}}
+      {{/each}}
+    </Hds::AdvancedTable::Tr>
+  </div>
+
+  <div class="hds-advanced-table__tbody" role="rowgroup">
+    {{! ----------------------------------------------------------------------------------------
+        IMPORTANT: we loop on the `model` array and for each record
+        we yield the Tr/Td/Th elements _and_ the record itself as `data`
+        this means the consumer will *have to* use the `data` key to access it in their template
```
zamoore, line 55: question - What do you think about linking to an example in the docs for how this works?

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -0,0 +1,346 @@
...
+      `@valign for "Hds::Table" must be one of the following: ${VALIGNMENTS.join(
+        ', '
+      )}; received: ${valign}`,
+      VALIGNMENTS.includes(valign)
+    );
+
+    return valign;
+  }
+
+  get gridTemplateColumns(): string {
```
zamoore, line 240: [suggestion] - what do you think about adding a comment that expands a little on what kind of string this returns?

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -0,0 +1,346 @@
...
+      onSort(this._sortBy, this._sortOrder);
+    }
+  }
+
+  onSelectionChangeCallback(
+    checkbox?: HdsFormCheckboxBaseSignature['Element'],
+    selectionKey?: string
+  ): void {
+    const { onSelectionChange } = this.args;
+    if (typeof onSelectionChange === 'function') {
```
zamoore, line 333: [suggestion][nitpick] - what do you think about doing the inverse check here and returning early? imo less nesting is easier to read.

showcase/app/templates/components/advanced-table.hbs
shleewhite: Looking into vertical align more, it doesnt work because advanced table doesn't use the table elements. Should we just remove this argument for now?

packages/components/src/components/hds/advanced-table/td.ts
```diff
@@ -0,0 +1,99 @@
...
+    // add a class based on the @align argument
+    if (this.align) {
+      classes.push(`hds-advanced-table__td--align-${this.align}`);
+    }
+
+    return classes.join(' ');
+  }
+
+  @action onFocusTrapDeactivate(): void {
+    this._shouldTrapFocus = false;
```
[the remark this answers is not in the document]
shleewhite: 🤦🏽‍♀️ a part I lost when switching to ember-focus-trap. It is fixed in the commit `fix: restore tabindex correctly when switch between mode`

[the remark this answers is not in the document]
dchyun: Looks great now! Thanks for fixing it.

packages/components/src/components/hds/advanced-table/index.ts
shleewhite: Sure! This was copied over from Table, but that makes sense to do.

packages/components/src/components/hds/advanced-table/index.hbs
shleewhite: Hm.. not sure what you mean. Since the model and columns arguments are required, all examples work this way.

zamoore:
[question] Something I noticed when trying this out in the showcase;

The row expansion toggle button doesn't have a visual focus state. Is that intentional?

shleewhite:
> [question] Something I noticed when trying this out in the showcase;
> 
> The row expansion toggle button doesn't have a visual focus state. Is that intentional?

@zamoore it actually has the same behavior as the sorting and tooltip buttons, the focus ring isn't visible on click, but if you focus it with a keyboard it is visible. This also comes from the existing Table component.


Screenshot 2025-01-03 at 10 52 22 AM

packages/components/src/components/hds/advanced-table/index.hbs
zamoore: Ah ok, looking at this in more depth that makes sense.

showcase/tests/integration/components/hds/advanced-table/index-test.js
```diff
@@ -0,0 +1,656 @@
...
+  id='data-test-advanced-table'
+  @model={{this.model}}
+  @columns={{this.columns}}
+  @valign='middle'
+/>`
+    );
+
+    assert
+      .dom('#data-test-advanced-table')
+      .hasClass('hds-advanced-table--valign-middle');
```
zamoore, line 239: [suggestion] If you want to expand this test, you can update the `valign` value with `this.set` and check the classes for each `valign` value/similar values in other tests.

packages/components/src/components/hds/advanced-table/helpers.ts
shleewhite: I will check this - but I wonder if it is solved by using the 3rd party library instead.

showcase/app/templates/components/advanced-table.hbs
KristinLBradley: I think the behavior should probably be emulated if practical.

alex-ju (commented):
I left a few comments from manual testing. I'll try and wrap up the actual code review soon

showcase/app/templates/components/advanced-table.hbs
dchyun: Found the issue. In the `AdvancedTable::ThSelectable` component the action is named `didInsertCheckbox` not `didInsert`. Changing the showcase action call to `didInsertCheckbox` fixed the issue. 

`Table::ThSelectable` calls the action `didInsert` so may want to align with that.

packages/components/src/components/hds/advanced-table/th-sort.hbs
```diff
@@ -0,0 +1,27 @@
...
+  aria-sort={{this.ariaSort}}
+  role="columnheader"
+  aria-rowspan={{@rowspan}}
+  aria-colspan={{@colspan}}
+  {{did-insert this.didInsert}}
+  {{! @glint-expect-error - https://github.com/josemarluedke/ember-focus-trap/issues/86 }}
+  {{focus-trap
+    isActive=this._shouldTrapFocus
+    shouldSelfFocus=true
+    focusTrapOptions=(hash onDeactivate=this.onFocusTrapDeactivate initialFocus=this.getInitialFocus)
```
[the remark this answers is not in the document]
shleewhite: This makes sense! Will update.

showcase/app/templates/components/advanced-table.hbs
shleewhite: Updated so it emulates the table behavior!

showcase/app/templates/components/advanced-table.hbs
```diff
@@ -0,0 +1,1712 @@
...
+    {{/if}}
+  </div>
+
+  <Shw::Text::H4>Execute action on selected rows</Shw::Text::H4>
+
+  <Shw::Text::Body>This demo emulates, for example, when a user needs to download the selected files.</Shw::Text::Body>
+
+  <div class="shw-component-table-demo-topbar">
+    <div class="shw-component-table-demo-topbar__action">
+      <Hds::Button @text="Animate users" @icon="play" {{on "click" this.multiSelectAnimateSelectedUsers_demo4}} />
```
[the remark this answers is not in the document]
shleewhite: Ah, I missed copying over styles from the table docs page. the example should work now.

packages/components/src/components/hds/advanced-table/helpers.ts
shleewhite: Added an example to the showcase, seems like it works as expected! https://hds-showcase-git-hds-3965-advanced-table-clean-hashicorp.vercel.app/components/advanced-table#with-dynamic-focusable-content-in-cells

dchyun: Not sure if this is just me, but seeing two issues with the new dynamic example. I think some of the `tabindex` adjustments aren't getting triggered correctly.

Seeing these issues after using the keyboard and entering into the "show more buttons" cell, and then triggering the button.

- When new buttons are added to cells outside of the cell that has focus, those elements aren't having `tabindex="-1"` set on them, making them focusable when they shouldn't be.
- If I use `esc` and take focus out of the cell, all of the new buttons in that cell get `tabindex="-1"` correctly added to them. But if I reenter the cell, the `tabindex` is not updated to `0` on the new buttons, making them unreachable.

shleewhite: Ok... I think this should be better now. I ended up adding a MutationObserver to detect when the cell's children changes.

dchyun: This looks perfect! Working great now!

alex-ju (approved):
Awesome work! I couldn't spot anything off. Left a few minor comments. Let's wait for feedback from design (and potentially other engineers if they're up for it), but good to go for me.

P.S. It looks like it needs a rebase

packages/components/src/components/hds/advanced-table/td.ts
```diff
@@ -0,0 +1,70 @@
...
+      )}; received: ${align}`,
+      ALIGNMENTS.includes(align)
+    );
+    return align;
+  }
+
+  get classNames(): string {
+    const classes = [
+      'hds-advanced-table__td',
+      'hds-typography-body-200',
```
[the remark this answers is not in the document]
didoo: > The final value would be dependant on the import order. We import @use "helpers/typography"; before the components, and we can probably rely on that to be stable, but _if we ever change the import order in the main file_

That would be very, very, **very** unlikely. It would go against how CSS files need to be imported (the classic "piramid"). 

Another advantage of using composition instead of re-adding the styles in the CSS is that we don't increase the CSS output size.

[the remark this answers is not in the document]
alex-ju: I'm more worried about [the import order in apps](https://github.com/hashicorp/cloud-ui/blob/main/hcp/app/styles/app.scss#L7-L8), rather than our package.

The output size after compression is insignificant.

Not a hill I'm prepared to die on 😄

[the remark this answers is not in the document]
didoo: > I'm more worried about [the import order in apps](https://github.com/hashicorp/cloud-ui/blob/main/hcp/app/styles/app.scss#L7-L8), rather than our package.

Same. If they were to swap that two imports, I would not be surprised if all hell broke loose, which would be a sign that they should **not** do it :)

packages/components/src/components/hds/advanced-table/helpers.ts
```diff
@@ -0,0 +1,212 @@
...
+          if (nextCell instanceof HTMLElement) {
+            changeActiveCell(target, nextCell);
+          }
+        }
+      }
+    } else if (event.key === 'Home' || event.key === 'End') {
+      event.preventDefault();
+      if (event.ctrlKey) {
+        const table = target.parentElement?.closest('[role="grid"]');
+        const allRows = table?.querySelectorAll('[role="row"]');
```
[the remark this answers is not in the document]
shleewhite: No reason.... I went back and forth on this. my instinct was to use roles as selectors initially because it felt more "semantic" and felt easier to scan for me, but then I realized in most places we use classes. I'm fine with either.

majedelass (commented):
I found a few things. I believe Andrew will do an independent review too. But here are a few ones. Let me know if you need anything.

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -0,0 +1,346 @@
...
+        data?: Record<string, unknown>;
+        rowIndex?: number;
+      },
+    ];
+  };
+  Element: HTMLDivElement;
+}
+
+export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignature> {
+  @tracked private _sortBy = this.args.sortBy ?? undefined;
```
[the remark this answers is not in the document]
shleewhite: Oh... i dont think its ever null. I ported this from table, so it should probably be addressed in both places.

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -0,0 +1,344 @@
...
+      }
+    }
+  }
+}
+
+// multi-select (isSelectable=true)
+.hds-advanced-table__th-content {
+  display: flex;
+  gap: 8px;
+  align-items: center;
```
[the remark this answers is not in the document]
shleewhite: The same styles handle the placement of the expand button, the tooltip, and the sorting button. Do you want all of them to be top aligned? From the table, the tooltip and sorting button are center aligned.

Screenshot 2025-01-16 at 11 22 15 AM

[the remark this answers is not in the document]
shleewhite: If we want to allow flexibility, that can be a fast follow pr.

packages/components/src/components/hds/advanced-table/index.hbs
```diff
@@ -0,0 +1,108 @@
...
+            (hash
+              Tr=(component "hds/advanced-table/tr" isParentRow=T.isExpandable depth=T.depth)
+              Th=(component
+                "hds/advanced-table/th"
+                scope="row"
+                isExpandable=T.isExpandable
+                newLabel=T.id
+                parentId=T.parentId
+                onClickToggle=T.onClickToggle
+                isExpanded=T.isExpanded
```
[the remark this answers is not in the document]
majedelass: @alex-ju I'd like to hear your thoughts on this. What do you think?

packages/components/src/styles/components/advanced-table.scss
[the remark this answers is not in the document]
majedelass: Ok, instead of allowing flexibility, can we top align it?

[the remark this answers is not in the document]
majedelass: I see, ok. I wonder if we should separate the `thead` and the `isExpandable` row so that the expandable row's button is aligned to the content. Thoughts @andgen404 and @alex-ju?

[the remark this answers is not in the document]
majedelass: If doing this delays things significantly, it's not as important and we can create a ticket as a fast follow.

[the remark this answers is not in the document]
shleewhite: No, I can easily just change it for the expand button. Just wanted to check if the other cases should be changed too.

packages/components/src/components/hds/advanced-table/index.hbs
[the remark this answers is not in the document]
alex-ju: I see how expanded is semantically (fits better with the idea of rows unfolding to reveal more details) and technically (`aria-expanded`) closer to the spec but for the sake of keeping aligned with the Accordion it's probably a good call to change it to `isOpen`. I would only chage it for the public api through (we should keep `@isExpandable` and `ExpandableTrGroup`. Sorry for not catching this @shleewhite!

packages/components/src/styles/components/advanced-table.scss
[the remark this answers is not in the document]
alex-ju: This is a design decision – as Lee mentioned, happy to change it

packages/components/src/components/hds/advanced-table/index.hbs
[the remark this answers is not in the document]
shleewhite: ok! ill update now.

packages/components/src/components/hds/advanced-table/th-button-expand.hbs
```diff
@@ -0,0 +1,19 @@
...
+<button
+  type="button"
+  class={{this.classNames}}
+  {{on "click" this.onClick}}
+  aria-labelledby="{{this._prefixLabelId}} {{@labelId}}"
+  aria-expanded="{{this.isExpanded}}"
```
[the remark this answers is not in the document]
shleewhite: Resolving because we're only going to change the public instances of isExpanded.

packages/components/src/components/hds/advanced-table/th.hbs
```diff
@@ -0,0 +1,52 @@
...
+  {{#if @isVisuallyHidden}}
+    <span class="sr-only">{{yield}}</span>
+  {{else}}
+    {{#if @tooltip}}
+      <div class="hds-advanced-table__th-content">
+        {{#if @isExpandable}}
+          <Hds::AdvancedTable::ThButtonExpand
+            @labelId={{this._labelId}}
+            @onToggle={{@onClickToggle}}
+            @isExpanded={{@isExpanded}}
```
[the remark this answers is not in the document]
shleewhite: Resolving because we're only going to change the public instances of isExpanded.

packages/components/src/components/hds/advanced-table/th.hbs
```diff
@@ -0,0 +1,52 @@
...
+        <span id={{this._labelId}} class="hds-typography-body-200 hds-font-weight-semibold">{{yield}}</span>
+        <Hds::AdvancedTable::ThButtonTooltip @tooltip={{@tooltip}} @labelId={{this._labelId}} />
+      </div>
+    {{else}}
+      <div class="hds-advanced-table__th-content">
+        {{#if @isExpandable}}
+          <Hds::AdvancedTable::ThButtonExpand
+            @labelId={{this._labelId}}
+            @onToggle={{@onClickToggle}}
+            @isExpanded={{@isExpanded}}
```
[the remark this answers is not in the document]
shleewhite: Resolving because we're only going to change the public instances of isExpanded.

packages/components/src/styles/components/advanced-table.scss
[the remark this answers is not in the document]
majedelass: No, it should only be for the expand button (and thanks for making the update 🙌 ). Keep the table header the way it is.

packages/components/src/components/hds/advanced-table/index.hbs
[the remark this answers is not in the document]
majedelass: Ok!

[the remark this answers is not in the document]
shleewhite: Updated.

alex-ju (approved):
I did a re-check on functionality. All looks good! I'd say let's get it in, and if anything comes up in testing we can address it separately.

MelSumner:
While keyboard navigating the grid, I noticed that the underlying browser keyboard defaults are still in place. 

So pressing the up/down arrow keys are not only navigating the grid cells but also moving the page up/down. Additionally, if you are in the bottom cell of a grid, and press the down arrow key again, nothing should happen, but the page continues to scroll down. This is probably not what we want.

majedelass (approved):
🎊 this is looking superb. Really excited about this shipping!!!

andgen404 (approved):
Looks good to me! Great work on this 🎉

PR #2618, opened 2024-12-19
Upgrade Ember to 5.12.0

dchyun:
### :pushpin: Summary

If merged, this PR would upgrade Ember from `5.9.0` to the latest stable released version of `5.12.0`.

### :link: External links

Jira ticket: [HDS-4248](https://hashicorp.atlassian.net/browse/HDS-4248)

***

### 👀 Component checklist


packages/components/config/ember-cli-update.json
```diff
@@ -3,7 +3,7 @@
   "packages": [
     {
       "name": "ember-cli",
-      "version": "5.9.0",
+      "version": "5.12.0",
```
aklkv, line 6: this config is wrong it should be updated to similar to this https://github.com/hashicorp/cloud-ui/blob/main/addons/vault-common/config/ember-cli-update.json
and if you are updating it you should run `ember-cli-update` to update the blueprint as well same goes for the apps unless you are planning to do follow up PR

packages/components/package.json
```diff
@@ -89,26 +89,26 @@
     "babel-plugin-ember-template-compilation": "^2.2.4",
     "concurrently": "^8.2.2",
     "ember-basic-dropdown": "^8.1.0",
-    "ember-source": "~5.9.0",
+    "ember-source": "~5.12.0",
```
aklkv, line 92: since it's already in `peerDeps` you do not need it as a `devDeps`

packages/ember-flight-icons/config/ember-cli-update.json
```diff
@@ -3,7 +3,7 @@
   "packages": [
     {
       "name": "ember-cli",
-      "version": "5.9.0",
+      "version": "5.12.0",
```
aklkv, line 6: you shouldn't bump this manually but by running ember-cli-update

showcase/config/ember-cli-update.json
```diff
@@ -3,7 +3,7 @@
   "packages": [
     {
       "name": "ember-cli",
-      "version": "5.9.0",
+      "version": "5.12.0",
```
aklkv, line 6: you shouldn't bump this manually but by running `ember-cli-update`

website/config/ember-cli-update.json
```diff
@@ -3,7 +3,7 @@
   "packages": [
     {
       "name": "ember-cli",
-      "version": "5.9.0",
+      "version": "5.12.0",
```
aklkv, line 6: you shouldn't bump this manually but by running `ember-cli-update`

packages/components/config/ember-cli-update.json
dchyun: Thanks for the feedback. These changes across the packages were made using `ember-cli-update`. If there is something about the changes that makes that not seem the case, please let me know. 

Also in terms of the formatting of the file, can you clarify what alignments you would like to see? Should they have a different format because of the v2 addon conversion?

packages/components/package.json
Dhaulagiri: thinking out loud, but it seems likely this is leftover from when there was an actual dummy app inside this package. now that it's moved elsewhere, there's no need to specify a specific ember version?

packages/components/package.json
```diff
@@ -51,7 +51,7 @@
     "ember-element-helper": "^0.8.5",
     "ember-focus-trap": "^1.1.0",
     "ember-get-config": "^2.1.1",
-    "ember-modifier": "^4.1.0",
+    "ember-modifier": "^4.2.0",
```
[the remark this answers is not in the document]
alex-ju: yeah, better to tackle this in [HDS-4250](https://hashicorp.atlassian.net/browse/HDS-4250)

packages/components/package.json
aklkv: correct, this line should be moved to `peerDeps` and provide wide ranges that are supported, so that types can be resolved

dchyun: Removed this from `devDependencies` and left the `peerDependencies` range as it is currently as `"ember-source": "^3.28.0 || ^4.0.0 || ^5.3.0"`.

packages/components/package.json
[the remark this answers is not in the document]
dchyun: Reverted this upgrade for now. Should this be addressed in [HDS-4252](https://hashicorp.atlassian.net/browse/HDS-4252) since it's a dependency?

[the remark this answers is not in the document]
Dhaulagiri: yes

packages/components/config/ember-cli-update.json
aklkv: this config is most likely remained unchanged since hds migrated to v2 addon, but v2 addon follows it's own blueprint, so this whole config needs to be updated to similar format I linked above

packages/components/package.json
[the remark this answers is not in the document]
alex-ju: > Reverted this upgrade for now. Should this be addressed in [HDS-4252](https://hashicorp.atlassian.net/browse/HDS-4252) since it's a dependency?

could you update the changelog entry as well to remove the `ember-modifier` bit?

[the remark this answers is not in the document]
dchyun: Updated the changelog entry

.changeset/curvy-colts-applaud.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Upgraded Ember.js to latest stable release `5.12`
```
Dhaulagiri, line 5: [suggestion] since there are no actual user-facing changes at this point, the changelog could be removed.

alex-ju (approved):
I ran `ember-cli-update` and got similar results with the current `ember-cli-update` (except for `@ember/string` being bumped to 4.x, but we don't want to enforce this as all our consumers are still on 3.x). Let's get this in and draft a card to look into re-aligning the `ember-cli-update` blueprint separately.

packages/components/package.json
alex-ju: for posterity, the reason we added `ember-source` to `devDependencies ` was to silence the following post-resolution yarn warning
```bash
➤ YN0002: │ @hashicorp/design-system-components@workspace:packages/components doesn't provide ember-source (pcc80c), requested by @ember/render-modifiers and other dependencies.
```

it's fine either way, as long as we're not bothered by the warning and we make sure the setup is correct.

PR #2619, opened 2024-12-19
Upgrade Nodejs to 22

dchyun:
### :pushpin: Summary

- Shifted supported version of Node.js from `16* || >= 18` to `>=18`
- Upgraded Node.js version of CI workflows from `20.x` to the latest LTS of `22.x`

Removing support for Node.js 16 due to the following factors

- [v16 is end-of-life](https://github.com/nodejs/release?tab=readme-ov-file#end-of-life-releases) as of Sep. 2023
- [Ember 5.12](https://github.com/emberjs/ember.js/blob/v5.12.0/package.json#L174) supports Node versions of `>18`
- [Yarn 4.5.3](https://github.com/yarnpkg/berry/blob/2fe7d635e8eb163c1b3e56ab76e652b6bae8c160/package.json#L73) supports Node versions of `>=18.12.0`

### :link: External links

Jira ticket: [HDS-4265](https://hashicorp.atlassian.net/browse/HDS-4265)

***

### 👀 Component checklist


packages/components/package.json
```diff
@@ -330,6 +330,6 @@
     "dist"
   ],
   "engines": {
-    "node": "16.* || >= 18"
+    "node": ">= 18"
```
Dhaulagiri, line 333: @alex-ju I'm curious your opinion on this. Technically this should be considered a breaking change, but AFAICT we don't have any known consumers using Node 16. On the other hand, there is no real _need_ for us to make this change either.

alex-ju: According to the latest findings around the API our consumers are mostly concerned about the changes not affecting them, rather than following a rigid versioning policy. Based on this, and considering Ember dropped support for Node 16 in v5.11.0, so I'd say we can do it in a minor release too.

shleewhite (approved):

zamoore (approved):
Nice work

PR #2612, opened 2024-12-17
`SuperSelect` - search position (HDS-4254)

KristinLBradley:
### :pushpin: Summary

If merged, this PR adds the `searchFieldPosition` to `SuperSelect` enabling `SuperSelect::Multiple` to display the search field within the before options area instead of in the trigger area.

* **Showcase:** https://hds-showcase-git-hds-4254-super-select-search-484150-hashicorp.vercel.app/components/form/super-select

#### Related:
* **Docs PR:** https://github.com/hashicorp/design-system/pull/2620

### :hammer_and_wrench: Detailed description

* Updates ember-power-select version to 8.6.0 & above
* Adds `searchFieldPosition` argument to SuperSelect components
  - Sets value of "before-options" for argument

### :camera_flash: Screenshots

image

### :link: External links

* Jira ticket: [HDS-4254](https://hashicorp.atlassian.net/browse/HDS-4254)

***

### 👀 Component checklist

- [ ] Percy was checked for any visual regression
- [ ] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-4254]: https://hashicorp.atlassian.net/browse/HDS-4254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2620, opened 2024-12-19
`SuperSelect` - docs for search position (HDS-4254)

KristinLBradley:
### :pushpin: Summary

If merged, this PR adds documentation for the `SuperSelect` `searchFieldPosition` feature.

* **Web docs:** https://hds-website-git-hds-4254-super-select-search-p-2680bc-hashicorp.vercel.app/components/form/super-select

#### Related:

* **Component PR:** https://github.com/hashicorp/design-system/pull/2612

### :link: External links

* Jira ticket: [HDS-4254](https://hashicorp.atlassian.net/browse/HDS-4254)

***

### 👀 Component checklist

[HDS-4254]: https://hashicorp.atlassian.net/browse/HDS-4254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


website/docs/components/form/super-select/partials/code/component-api.md
```diff
@@ -79,6 +79,13 @@ The default values of some [ember-power-select](https://ember-power-select.com/d
     <br><br>
     <strong>Note</strong>: It cannot be overridden.
   </C.Property>
+  <C.Property @name="searchFieldPosition" @type="string">
+    Determines whether the search field, when enabled, renders either within the before options content or in the “trigger” area.
+    <br><br>
+    Set to `"before-options"` to fix a `PowerSelect` accessibility issue. (Rendering the search input within the trigger area creates a nested interaction.)
```
KristinLBradley, line 85: Is there any more a11y-related content related to this that we would want to add within the SuperSelect docs?  
(See: [Docs preview](https://hds-website-git-hds-4254-super-select-search-p-2680bc-hashicorp.vercel.app/components/form/super-select))

@MelSumner @shleewhite

shleewhite: I don't think we need to necessarily link out to anything, you could just elaborate a bit more. Something like this:

```suggestion
    Set to `"before-options"` to fix a `PowerSelect` accessibility issue. Rendering the search input within the trigger is not complaint because it nests an interactive element within another interactive element.
```

website/docs/components/form/super-select/partials/code/component-api.md
```diff
@@ -211,6 +218,13 @@ The default values of some [ember-power-select](https://ember-power-select.com/d
     <br><br>
     Note: It cannot be overridden.
   </C.Property>
+  <C.Property @name="searchFieldPosition" @type="string">
+    Determines whether the search field, when enabled, renders either within the before options content or in the “trigger” area.
+    <br><br>
+    Set to `"before-options"` to fix a `PowerSelect` accessibility issue. (Rendering the search input within the trigger area creates a nested interaction.)
```
KristinLBradley, line 224:
```suggestion
    Set to `"before-options"` to fix a `PowerSelect` accessibility issue.  Rendering the search input within the trigger is not compliant because it nests an interactive element within another interactive element.
```

KristinLBradley:
@majedelass Do you think any of the illustrations or design related docs need updating? I didn't see anything obvious when I looked through but wanted to double-check.

KristinLBradley:
What version would this be released as part of? (So I can add the version history.)

Dhaulagiri:
I would think 4.16

jorytindall:
I went down the rabbit hole with https://github.com/hashicorp/design-system/pull/2612 to better understand where this content was coming from, and I personally found it a bit confusing to read through. Given the context it looks like we're documenting another argument along with the rest of the component API, but the note about it not being possible to override kind of makes this prop useless to consumers. That is, unless we determine that it's useful for consumers to know that we are setting this argument explicitly for accessibility reasons being that it's coming from `PowerSelect`.

At a minimum it might be helpful to make that more clear immediately in the component API (leading with the content about not being able to override it and why), then capturing the rest of the details about the property if necessary.

Not overly opinionated on the topic and I don't have all of the background, just noting a couple of thoughts.

KristinLBradley:
> I went down the rabbit hole with #2612 to better understand where this content was coming from, and I personally found it a bit confusing to read through. Given the context it looks like we're documenting another argument along with the rest of the component API, but the note about it not being possible to override kind of makes this prop useless to consumers. That is, unless we determine that it's useful for consumers to know that we are setting this argument explicitly for accessibility reasons being that it's coming from `PowerSelect`.
> 
> At a minimum it might be helpful to make that more clear immediately in the component API (leading with the content about not being able to override it and why), then capturing the rest of the details about the property if necessary.
> 
> Not overly opinionated on the topic and I don't have all of the background, just noting a couple of thoughts.

@jorytindall @hashicorp/hds-engineering 

I originally wasn't positive whether we should document this new prop but thought it made sense from the standpoint of consumers who are used to using Ember PowerSelect since it differs from what they may expect. And since this is what we followed for the `renderInPlace` prop which also can't be overridden.

I'm open to removing or altering the documentation of these props though if others have ideas on what may make more sense.

KristinLBradley:
I removed documentation for `renderInPlace` and ``searchFieldPosition` within the API docs properties list and added mention of them within an existing related Info banner.

**Preview:** https://hds-website-git-hds-4254-super-select-search-p-2680bc-hashicorp.vercel.app/components/form/super-select?tab=code#component-api

image

What do you think? (The Info banner is kind of long now I think...) @jorytindall @shleewhite

shleewhite (dismissed):

website/docs/components/form/super-select/partials/code/component-api.md
```diff
@@ -13,7 +13,12 @@ While the [ember-power-select documentation](https://ember-power-select.com/docs
 
 !!! Info
 
-The default values of some [ember-power-select](https://ember-power-select.com/docs/api-reference) options have been overridden in some cases to either resolve accessibility issues or to customize according to our needs. These options are documented below.
+The default values of some [ember-power-select](https://ember-power-select.com/docs/api-reference) options have been overridden to either resolve accessibility issues or to customize according to our needs.
+
+Overrides to fix accessibility issues:
+
+- `renderInPlace`: Always set as `true`. (The opener element or “trigger” should be next to dropdown content it hides & reveals.)
```
shleewhite, line 20: Super small nit, but i'd remove the parenthesis on this line and the following line.
```suggestion
- `renderInPlace`: Always set as `true`. The opener element or “trigger” must be next to dropdown content it hides & reveals.
```

website/docs/components/form/super-select/partials/code/component-api.md
```diff
@@ -13,7 +13,12 @@ While the [ember-power-select documentation](https://ember-power-select.com/docs
...
+- `searchFieldPosition`: Always set to `"before-options"`. (Rendering the search input within the trigger is not compliant because it nests interactive elements within each other.)
```
shleewhite, line 21:
```suggestion
- `searchFieldPosition`: Always set to `"before-options"`. Rendering the search input within the trigger is not compliant because it nests interactive elements within each other.
```

jorytindall:
> I removed documentation for renderInPlace and ``searchFieldPosition` within the API docs properties list and added mention of them within an existing related Info banner.

@KristinLBradley personally, I think this is a better solution than including them in the component API list which seemed to create a bit of a false equivalency. Thanks for making this change, curious what other folks think!

shleewhite (approved):

majedelass (dismissed):
This looks good. I removed the warning within the guidelines about the nested interactive of the previous version.

alex-ju (dismissed):

shleewhite (approved):
website changes look good! But why are there changes to yarn.lock?

website/docs/components/form/super-select/partials/version-history/version-history.md
dchyun, line 1: With the new changelog scripts I think these updates can be removed, as they'll be generated later.

dchyun (approved):

PR #2627, opened 2024-12-26
`Tag` - Overflow with longer text fix

dchyun:
### :pushpin: Summary

If merged, this PR would fix an issue observed in the `Tag` where the dismiss button background overflows the border when the text wraps to multiple lines.

It was observed in the `Tag` that when the text exceeds 3 lines, the background of the dismiss button obscures the border of the element. 

The fix provided solves the issue by adding `overflow: hidden` to the tag. When the dismiss button is focused, the focus ring of the button goes outside of the `Tag` border, so for the `:focus` state, the `overflow: hidden` is removed. 

There are design efforts underway in [HDS-4317](https://hashicorp.atlassian.net/browse/HDS-4317) to provide more guidance or future changes around this issue. This PR is an immediate fix for the present issue. 

`Tag` - Before
Screenshot 2024-12-26 at 10 24 59 AM

`Tag` - After
Screenshot 2025-01-06 at 8 41 19 AM

### :link: External links

Jira ticket: [HDS-3854](https://hashicorp.atlassian.net/browse/HDS-3854)

***

### 👀 Component checklist

[HDS-3854]: https://hashicorp.atlassian.net/browse/HDS-3854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HDS-4317]: https://hashicorp.atlassian.net/browse/HDS-4317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/styles/components/form/super-select.scss
+++ b/packages/components/src/styles/components/form/super-select.scss
@@ -7,7 +7,7 @@
 
 // Copied from components/src/styles/components/tag.scss styles
 // we set a higher value than the line-height (~13px) to accommodate cases where the text wraps
-$hds-tag-border-radius: 50px; // This variable is repeated from tag.scss
+$hds-tag-border-radius: 20px; // This variable is repeated from tag.scss
 $hds-super-select-item-height: 36px;
 
 // Style overrides:
--- a/packages/components/src/styles/components/form/super-select.scss
+++ b/packages/components/src/styles/components/form/super-select.scss
@@ -7,7 +7,7 @@
 
 // Copied from components/src/styles/components/tag.scss styles
 // we set a higher value than the line-height (~13px) to accommodate cases where the text wraps
-$hds-tag-border-radius: 20px; // This variable is repeated from tag.scss
+$hds-tag-border-radius: 50px; // This variable is repeated from tag.scss
 $hds-super-select-item-height: 36px;
 
 // Style overrides:
--- a/website/docs/components/tag/partials/version-history/4.16.0.md
+++ b/website/docs/components/tag/partials/version-history/4.16.0.md
@@ -1,5 +1,0 @@
-## 4.16.0
-
-### Updated
-
-Fixed issue with overflow of dismiss button with longer content
--- a/website/docs/components/tag/partials/version-history/4.16.0.md
+++ b/website/docs/components/tag/partials/version-history/4.16.0.md
@@ -1,0 +1,5 @@
+## 4.16.0
+
+### Updated
+
+Fixed issue with overflow of dismiss button
--- a/website/docs/components/tag/partials/version-history/4.16.0.md
+++ b/website/docs/components/tag/partials/version-history/4.16.0.md
@@ -2,4 +2,4 @@
 
 ### Updated
 
-Fixed issue with overflow of dismiss button
+Fixed issue with overflow of dismiss button with longer content
--- a/.changeset/heavy-camels-crash.md
+++ b/.changeset/heavy-camels-crash.md
@@ -1,0 +1,6 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Tag` - Fixed issue with overflow of dismiss button
+`SuperSelect` - Fixed issue with overflow of tags in `SuperSelect::Multiple`
--- a/.changeset/heavy-camels-crash.md
+++ b/.changeset/heavy-camels-crash.md
@@ -2,5 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-`Tag` - Fixed issue with overflow of dismiss button
-`SuperSelect` - Fixed issue with overflow of tags in `SuperSelect::Multiple`
+`Tag` - Fixed issue with overflow of dismiss button with longer content
--- a/packages/components/src/styles/components/tag.scss
+++ b/packages/components/src/styles/components/tag.scss
@@ -15,12 +15,12 @@ $hds-tag-border-radius: 50px;
 .hds-tag {
   display: inline-flex;
   align-items: stretch;
+  overflow: hidden;
   line-height: 1rem; // 16px - override `body-100`
   vertical-align: middle;
   background-color: var(--token-color-surface-interactive);
   border: 1px solid var(--token-color-border-strong);
   border-radius: $hds-tag-border-radius;
-  overflow: hidden;
 }
 
 .hds-tag:focus-within,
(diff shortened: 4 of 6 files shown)


KristinLBradley:
An alternative to changing the border-radius could be adding `overflow: hidden;` to the wrapping span tag so that the nested button background won't poke out overlapping the border.

KristinLBradley (commented):
Commented with another possible option for fixing

dchyun:
> An alternative to changing the border-radius could be adding `overflow: hidden;` to the wrapping span tag so that the nested button background won't poke out overlapping the border. (That way the current design won't be altered btw.)
> 
> (@jorytindall @LilithJames-HDS Tagging you both so that you are aware of this option)
> 
> image

I tried this approach, and the only issue with it is that this results in the focus ring getting cut off when the tag has focus. I also tried an approach with setting the button background with the `::before` element, but this only solved the overlap issue for the default state. The active state still had the overlap. 

Screenshot 2025-01-02 at 11 39 32 AM
Screenshot 2025-01-02 at 11 39 39 AM

packages/components/src/styles/components/form/super-select.scss
```diff
@@ -7,7 +7,7 @@
 
 // Copied from components/src/styles/components/tag.scss styles
 // we set a higher value than the line-height (~13px) to accommodate cases where the text wraps
-$hds-tag-border-radius: 50px; // This variable is repeated from tag.scss
```
jorytindall, line 10: [rhetorical question] I don't have the historical context here (this component was created before I joined), but I'm wondering why this was the value set in the first place? It's probably different from Figma for a reason, but `50px` seems arbitrary.

KristinLBradley:
> > An alternative to changing the border-radius could be adding `overflow: hidden;` to the wrapping span tag so that the nested button background won't poke out overlapping the border. (That way the current design won't be altered btw.)
> > (@jorytindall @LilithJames-HDS Tagging you both so that you are aware of this option)
> > image
> 
> I tried this approach, and the only issue with it is that this results in the focus ring getting cut off when the tag has focus. I also tried an approach with setting the button background with the `::before` element, but this only solved the overlap issue for the default state. The active state still had the overlap.
> 
> Screenshot 2025-01-02 at 11 39 32 AM Screenshot 2025-01-02 at 11 39 39 AM

I see. I think your fix is the best option in that case.

KristinLBradley (dismissed):

jorytindall (approved):
Thanks for tackling this @dchyun ! The solution makes sense to me, and from a design perspective things are looking solid.

packages/components/src/styles/components/form/super-select.scss
KristinLBradley: I'm thinking it was set to a high value to ensure it would still look rounded if text lines wrapped but for some reason the issue with the button background poking out when it wraps to 3 lines+ was missed..

(If you want to ensure borders stay fully rounded for elements no matter what size they are you can set a super high border-radius value such as 500px.)

heatherlarsen:
@dchyun Thanks for catching this issue! However, to distinguish between Badge and Tag, Tag should always be fully rounded (a pill shape). The current proposed solution (rectangle with a very soft border-radius) does not fit within our visual language.  

I tested the Figma component and it does what you have proposed, but I don't believe we expected there to be tags that wrapped to this many lines. I'd like design to take a look at this before moving forward to see if there's an alternative solution that lets us keep the pill shape. I've created a Jira ticket to track this work, [HDS-4317](https://hashicorp.atlassian.net/browse/HDS-4317).

dchyun:
Reverting back to a Draft PR until design can explore further in [HDS-4317](https://hashicorp.atlassian.net/browse/HDS-4317)

packages/components/src/styles/components/form/super-select.scss
KristinLBradley: How about trying something like:

```
.hds-tag {overflow: hidden;}

.hds-tag:focus-within,
.hds-tag:has(.mock-focus) {
    overflow: visible;
}
```

dchyun: @KristinLBradley this is a great find! I think this would fix the issue. @heatherlarsen this fix would solve the problem without any design changes. The focus state would still have the overflow issue, but I think it's not noticeable, especially when zoomed out. Would design be okay with this?

You can take a look at a [preview of the fix](https://hds-showcase-git-dchyun-tag-overflow-fix-hashicorp.vercel.app/components/tag)

Screenshot 2025-01-03 at 11 20 05 AM

heatherlarsen: This is a great find and resolves the immediate issue! ❤️ 

We may still want to advise against showing this many lines or finding an alternative solution, so I'll keep the ticket open but adjust the language.

shleewhite (approved):
Nice!

KristinLBradley:
@dchyun If you merge in the latest `main` it should take care of one of the Percy visual diffs.

The other visual diff is from the fixes you made in this PR so if those are as expected they can be approved. (Merge in `main` first though.)

dchyun:
> @dchyun If you merge in the latest `main` it should take care of one of the Percy visual diffs.
> 
> The other visual diff is from the fixes you made in this PR so if those are as expected they can be approved. (Merge in `main` first though.)

Rebased to main and just have the one Percy change needing review from design.

dchyun:
> I also noticed the active state looks like this: image
> 
> Active is ignoring the overflow hidden parent rule.

Thanks for catching this. Fixed the issue by updating the CSS selectors. 

The issue was that we were relying on `:focus-within` to set the `overflow: visible`, but setting the focus ring with `:focus-visible` on the button. This led to situations where a click with a mouse would trigger the `:focus-within` selector, but not the `:focus-visible` selector.

website/docs/components/tag/partials/version-history/4.16.0.md
```diff
@@ -0,0 +1,5 @@
+## 4.16.0
```
shleewhite, line 1: Suggestion: should this be removed now that we have automation to handle it?

shleewhite (commented):
Fix is good! 

Also looks like main needs to be merged in so the multi-select super select percy test updates.

dchyun:
Reverting back to draft until @majedelass completes their exploration of the issue from the design side in [HDS-4317](https://hashicorp.atlassian.net/browse/HDS-4317).

Current solution solves issues with inactive, active, and hover states, but overflow still occurs on focus.
Screenshot 2025-01-14 at 3 58 15 PM

PR #2629, opened 2024-12-27
feat: switch website scripts from npm-run-all to concurrently

shleewhite:
### :pushpin: Summary

Currently, the website is the only package that still uses `npm-run-all` instead of `concurrently`. This PR switches it over to `concurrently` and removes the dependency on `npm-run-all`.

This fixes an issue where the website has not been properly linting for some period of time. Example run: https://github.com/hashicorp/design-system/actions/runs/12510182387/job/34900621801?pr=2628 (need to expand the lint step to see that it didn't actually run).

**Before fix**
Screenshot 2024-12-26 at 5 55 39 PM

**After fix**
Screenshot 2025-01-02 at 9 38 25 AM

This PR also fixes some spelling errors that have cropped up and updates the dictionaries to include more words that should be allowed.

***

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #2630, opened 2024-12-27
fix: lint:css:fix command in components and showcase

dchyun:
### :pushpin: Summary

If merged, this PR would fix an issue with the `lint:css:fix` command in the components package and showcase.

In reviewing, #2629 it was observed that the `lint:css:fix` command was not properly working across the repo. If a CSS error to a file is introduced, such as swapping the order of two properties, the linter will catch the issue, but `--fix` was not applying the fix.

Output after creating a linting error in a CSS file.

Before
Screenshot 2024-12-27 at 7 40 25 AM

After
Screenshot 2024-12-27 at 7 40 30 AM


zamoore (approved):
Nice catch

shleewhite (approved):

PR #1367, opened 2023-05-19
Update spacing between `Table` and `Pagination` in showcase

didoo:
### :pushpin: Summary

See thread: https://hashicorp.slack.com/archives/C025N5V4PFZ/p1684521629847999

***

### 👀 Reviewer's checklist:

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #2638, opened 2025-01-09
HDS-4320 Dropdown Toggle Icon: Fix z-index issue with focus ring

KristinLBradley:
### :pushpin: Summary

If merged, this PR adds a style to Dropdown Toggle Icon to isolate it from its background and solve a z-index issue in which the focus ring was going behind parent containers (such as when the Dropdown is used within a Table).

**Showcase:** https://hds-showcase-git-hds-4320-dropdown-focus-ring-bug-hashicorp.vercel.app/components/table#custom-sorting (Tab through nested Dropdowns to test)

**_Before (focus ring of toggle is not visible):_**
image

---

**_After (focus ring of toggle is visible):_**
image

### :link: External links

* Jira ticket: [HDS-4320](https://hashicorp.atlassian.net/browse/HDS-4320)

***

### 👀 Component checklist

[HDS-4320]: https://hashicorp.atlassian.net/browse/HDS-4320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/smooth-months-repair.md
+++ b/.changeset/smooth-months-repair.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Dropdown` - Fixed z-index bug which caused the focus ring of the toggle icon to not be visible when the component was nested in a container.
--- a/.changeset/smooth-months-repair.md
+++ b/.changeset/smooth-months-repair.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-`Dropdown` - Fixed z-index bug which caused the focus ring of the toggle icon to not be visible when the component was nested in a container.
+`Dropdown` - Fixed `z-index` bug which caused the focus ring of the toggle icon to not be visible when the component was nested in a container.
--- a/packages/components/src/styles/components/dropdown.scss
+++ b/packages/components/src/styles/components/dropdown.scss
@@ -42,6 +42,7 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
   border-radius: $hds-dropdown-toggle-border-radius;
   outline-style: solid; // used to avoid double outline+focus-ring in Safari (see https://github.com/hashicorp/design-system-components/issues/161#issuecomment-1031548656)
   outline-color: transparent; // We need this to be transparent for a11y
+  isolation: isolate; // used to create a new stacking context (needed to have the pseudo element below text/icon but not the parent container)
 
   &:hover,
   &.mock-hover {


KristinLBradley:
Should I add some sort of Showcase example?

packages/components/src/styles/components/dropdown.scss
```diff
@@ -42,6 +42,7 @@ $hds-dropdown-toggle-border-radius: $hds-button-border-radius;
   border-radius: $hds-dropdown-toggle-border-radius;
   outline-style: solid; // used to avoid double outline+focus-ring in Safari (see https://github.com/hashicorp/design-system-components/issues/161#issuecomment-1031548656)
   outline-color: transparent; // We need this to be transparent for a11y
+  isolation: isolate; // used to create a new stacking context (needed to have the pseudo element below text/icon but not the parent container)
```
KristinLBradley, line 45: Btw, we've used this as a fix for similar issues else where so I think it's very low risk. But does anyone have any concerns? I can try testing within consumer platforms but may be difficult to tell if anything's effected.

didoo: Mmm, if you look at [this very old PR](https://github.com/hashicorp/design-system/pull/1367), you [will see](https://hds-showcase-git-update-pagination-spacing-hashicorp.vercel.app/components/table) that the focus was correctly rendered (I was puzzled how we could have missed) while in the [current showcase](https://hds-showcase.vercel.app/components/table#custom-sorting) the focus ring is incomplete.

Before adding this fix, personally I would investigate what happened in between these two (or you can go back in time and see when it started to appear), to discover what is causing this issue, at least to understand if there's an alternative solution. 

Not that the one you're applying is not OK, but it may be tricky because we have seen issues in some cases with it, in Cloud UI for example, because of how they use the `z-index` in their pages, in particular in relation to the PageHeader and their layout component for the "main" content (even though, in that case, was the dropdown menu causing the problem).

KristinLBradley: So when I look at the Showcase in your old branch I'm seeing the `isolation: isolate` style on the button:
image

That style isn't in the newest version though. I'm trying to figure out when/how it was removed or what happened...

KristinLBradley: I tried looking around in HCP to see what they have in their now and when I inspect I see `isolation: isolate;` on `.hds-dropdown-toggle-icon`.

Here it's crossed out because it's overridden by another style (below) which also applies isolation:
image

image

didoo: > I don't understand why I see the style when I inspect that old Showcase but I can't find it within the actual dropdown.scss styles so far in older versions of the file.

@KristinLBradley if you check out for example the commit [5012cd9](https://github.com/hashicorp/design-system/commit/5012cd96f609b4951927bf5568bdb461305481dd) you will see that the `isolation` comes from the `hds-focus-ring-with-pseudo-element` mixin:





Now, if you look at today's file you will see that there's not anymore such mixin:
screenshot_4582

The change has been introduced in this PR https://github.com/hashicorp/design-system/pull/2574, see https://github.com/hashicorp/design-system/pull/2574/files#diff-264737dac22665650a57892f74bdfc36ec22fefc6e29dc76a47a858c2d4a33b7

/cc @alex-ju @dchyun

didoo: > I tried looking around in HCP to see what they have in their now and when I inspect I see `isolation: isolate;` on `.hds-dropdown-toggle-icon`.
> 
> Here it's crossed out because it's overridden by another style (below) which also applies isolation: image
> 
> image

Btw, for context, this complex selector is coming from here:
screenshot_4583

KristinLBradley: So since the removed `hds-focus-ring-with-pseudo-element` mixin included the `isolation: isolate;` style, this PR would just add that style back. Unless there was a specific reason that style wasn't included in the changes in this PR...

https://github.com/hashicorp/design-system/pull/2574/files#diff-264737dac22665650a57892f74bdfc36ec22fefc6e29dc76a47a858c2d4a33b7

@dchyun @alex-ju could you confirm whether this was an intended change (removing isolate style)?

alex-ju: I don't think the intention was to remove the isolation. It was likely an unintended effect of that fix. I'm ok with your solution here.

.changeset/smooth-months-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Dropdown` - Fixed z-index bug which caused the focus ring of the toggle icon to not be visible when the component was nested in a container.
```
alex-ju, line 5:
```suggestion
`Dropdown` - Fixed `z-index` bug which caused the focus ring of the toggle icon to not be visible when the component was nested in a container.
```

packages/components/src/styles/components/dropdown.scss
KristinLBradley: @dchyun confirmed that it was unintended to remove this style. https://hashicorp.slack.com/archives/C025N5V4PFZ/p1736880167697849?thread_ts=1736879810.252109&cid=C025N5V4PFZ

alex-ju (approved):

PR #2643, opened 2025-01-13
`DisclosurePrimitive` - Implement a11y toggled content pattern

dchyun:
### :pushpin: Summary

This PR implements usage of the `aria-controls` attribute in the `DisclosurePrimitive` component to align show/hide behavior across the repo as follows in the initiative from [HDS-3581](https://hashicorp.atlassian.net/browse/HDS-3581).

The `DisclosurePrimitive` is leveraged in the `Accordion` and `Reveal`.

Currently in the `DisclosurePrimitive`, the toggled content does not follow the proper a11y structure for toggled content, or leverage `aria-controls` in the preferred way. 

As per a11y guidance in [HDS-3581](https://hashicorp.atlassian.net/browse/HDS-3581), all togged content should follow a pattern leveraging `aria-controls`, and all toggled containers should not be removed from the DOM on each toggle. They should always be present, and only the content inside them is added and removed. 

Currently the `DisclosurePrimitive` follows a pattern where the entire content block is added and removed on each toggle, and any `aria-controls` attributes are set to the `id` of this removable block. In this PR the `hds-disclosure-primitive__content` block is now in the DOM at all times, and the yielded content is removed or added based on the toggle.

In this PR, in the `Accordion` and `Reveal`, the `aria-controls` attribute of the toggle elements are now set equal to a `contentId` provided by the `DisclosurePrimitive`. They are no longer set to the `id` of content inside the yielded block.

### Smoke Testing

Since this PR involves updating the DOM structure of the `DisclosurePrimitive` it has been smoke tested in various products to ensure they are not relying on the existing structure in any of their tests. 

- ✅  [Cloud UI](https://github.com/hashicorp/cloud-ui/pull/12287) 
- ✅  [Atlas](https://github.com/hashicorp/atlas/pull/23357) 
  - Note: The one failing test is unrelated to these changes
- ✅  Nomad 
  - There was one test case relying on the old structure, but that test has been updated in a [PR](https://github.com/hashicorp/nomad/pull/25905)

**Before - `Accordion`**

Closed
Screenshot 2025-01-10 at 3 08 59 PM

Open
Screenshot 2025-01-10 at 3 09 14 PM

**After - `Accordion`**

Closed
Screenshot 2025-01-10 at 3 16 11 PM

Open
Screenshot 2025-01-10 at 3 16 23 PM

### :link: External links

Jira ticket: [HDS-4324](https://hashicorp.atlassian.net/browse/HDS-4324)

***

### 👀 Component checklist

[HDS-3581]: https://hashicorp.atlassian.net/browse/HDS-3581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HDS-3581]: https://hashicorp.atlassian.net/browse/HDS-3581?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HDS-4324]: https://hashicorp.atlassian.net/browse/HDS-4324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/disclosure-primitive/index.hbs
+++ b/packages/components/src/components/hds/disclosure-primitive/index.hbs
@@ -4,11 +4,11 @@
 }}
 <div class="hds-disclosure-primitive" {{did-update this.onStateChange @isOpen}} ...attributes>
   <div class="hds-disclosure-primitive__toggle">
-    {{yield (hash onClickToggle=this.onClickToggle isOpen=this.isOpen) to="toggle"}}
+    {{yield (hash onClickToggle=this.onClickToggle isOpen=this.isOpen contentId=this._contentId) to="toggle"}}
   </div>
-  {{#if this.isOpen}}
-    <div class="hds-disclosure-primitive__content">
+  <div class="hds-disclosure-primitive__content" id={{this._contentId}}>
+    {{#if this.isOpen}}
       {{yield (hash close=this.close) to="content"}}
-    </div>
-  {{/if}}
+    {{/if}}
+  </div>
 </div>
--- a/.changeset/seven-eyes-whisper.md
+++ b/.changeset/seven-eyes-whisper.md
@@ -3,5 +3,7 @@
 ---
 
 `Accordion` - Set `aria-controls` of `Accordion::Item::Button` to `contentId` from `DisclosurePrimitve` for a11y improvements with toggled content
+
 `DisclosurePrimitve` - Changed DOM structure of content section and exposed `contentId` for a11y improvements with toggled content
+
 `Reveal` - Set `aria-controls` of `Reveal::Toggle` to `contentId` from `DisclosurePrimitve` for a11y improvements with toggled content
--- a/.changeset/seven-eyes-whisper.md
+++ b/.changeset/seven-eyes-whisper.md
@@ -1,0 +1,7 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Accordion` - Set `aria-controls` of `Accordion::Item::Button` to `contentId` from `DisclosurePrimitve` for a11y improvements with toggled content
+`DisclosurePrimitve` - Changed DOM structure of content section and exposed `contentId` for a11y improvements with toggled content
+`Reveal` - Set `aria-controls` of `Reveal::Toggle` to `contentId` from `DisclosurePrimitve` for a11y improvements with toggled content
--- a/packages/components/src/components/hds/accordion/item/index.hbs
+++ b/packages/components/src/components/hds/accordion/item/index.hbs
@@ -14,7 +14,7 @@
       <Hds::Accordion::Item::Button
         @isOpen={{t.isOpen}}
         @onClickToggle={{t.onClickToggle}}
-        @contentId={{this._contentId}}
+        @contentId={{t.contentId}}
         @ariaLabel={{@ariaLabel}}
         @ariaLabelledBy={{this.ariaLabelledBy}}
         @size={{this.size}}
@@ -35,14 +35,7 @@
   </:toggle>
 
   <:content as |c|>
-    <Hds::Text::Body
-      class="hds-accordion-item__content"
-      @tag="div"
-      @size="200"
-      @weight="regular"
-      @color="primary"
-      id={{this._contentId}}
-    >
+    <Hds::Text::Body class="hds-accordion-item__content" @tag="div" @size="200" @weight="regular" @color="primary">
       {{yield (hash close=c.close) to="content"}}
     </Hds::Text::Body>
   </:content>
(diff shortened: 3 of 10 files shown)


shleewhite (dismissed):
This looks good, but we should definitely test this in products before we ship. Last year we changed something similar-ish in Dropdown and it broke tests (and led us to add `@preserveContentInDom`): https://github.com/hashicorp/design-system/blob/735b7bd11aab9f5101b59fc21a9839bcd0880c21/packages/components/src/components/hds/dropdown/index.hbs#L21)

dchyun:
> This looks good, but we should definitely test this in products before we ship. Last year we changed something similar-ish in Dropdown and it broke tests (and led us to add `@preserveContentInDom`):
> 
> https://github.com/hashicorp/design-system/blob/735b7bd11aab9f5101b59fc21a9839bcd0880c21/packages/components/src/components/hds/dropdown/index.hbs#L21
> 
> )

Tested the changes across `atlas` and `cloud-ui` and didn't run into any test failures. But I think this is an open question if these kind of DOM changes could be considered breaking or not. @alex-ju do you see any potential issues with this or think it could possibly be a breaking change based on your recent RFC? 

Similar DOM changes will have to be made for the `hds-tooltip` updates, so can follow whatever we decide here on that as well.

packages/components/src/components/hds/disclosure-primitive/index.hbs
```diff
@@ -4,11 +4,11 @@
 }}
 <div class="hds-disclosure-primitive" {{did-update this.onStateChange @isOpen}} ...attributes>
   <div class="hds-disclosure-primitive__toggle">
-    {{yield (hash onClickToggle=this.onClickToggle isOpen=this.isOpen) to="toggle"}}
+    {{yield (hash onClickToggle=this.onClickToggle isOpen=this.isOpen contentId=this._contentId) to="toggle"}}
   </div>
-  {{#if this.isOpen}}
-    <div class="hds-disclosure-primitive__content">
+  <div class="hds-disclosure-primitive__content" id={{this._contentId}}>
```
alex-ju, line 9: This could a breaking change if consumers use `hds-disclosure-primitive__content` selector to check if disclosed content is visible or not. If we find through testing that it does not affect tests then we're good. Otherwise we'd either have to find another way or defer this feature to a major release.

dchyun: Yeah I agree it could be seen as breaking. I tested it out in atlas locally and ran it through tests in [cloud-ui](https://github.com/hashicorp/cloud-ui/pull/11475) and nothing seems to be breaking, but not sure about other consumers.

alex-ju: Smoke testing sounds good. Searching for this class being used as a selector I could only find [a few instances in one test in nomad](https://github.com/hashicorp/nomad/blob/main/ui/tests/pages/jobs/job/actions.js). I would try and check that test suite and if it's all good we can consider releasing this as a feature.

dchyun: Tests seem to pass for me. If anything it may make the test they have never fail. They are testing that `hds-disclosure-primitive__content` is present on the page, and with the new approach it will always be present and only the inner content will be added / removed.

alex-ju: Great! Thanks for checking! On that basis we can class it as a patch indeed.

dchyun: Went back through this after I realized my changes from the testing branch weren't updating the dist for local testing in Nomad. There does seem to be a test failure now on that test you highlighted. The test is related to a [component](https://github.com/hashicorp/nomad/blob/main/ui/app/components/actions-dropdown.hbs) they have that uses a reveal inside of a dropdown. So we may have to hold off on this change...

Screenshot 2025-01-15 at 8 54 44 PM

alex-ju: We can hold this PR for now and work it from another angle. Maybe we can update their tests to not need this class, then get this change in.

dchyun:
Converted back to a draft pr until [issues with failing tests](https://github.com/hashicorp/design-system/pull/2643#discussion_r1917566826) in consumer's repos can be sorted.

packages/components/src/components/hds/disclosure-primitive/index.hbs
dchyun: Update: This test has now been updated in Nomad through our [PR](https://github.com/hashicorp/nomad/pull/25905) which was merged.

PR #2650, opened 2025-01-14
`Table`: Multi-select bug - selected rows removed when model updates (HDS-4273)

KristinLBradley:
### :pushpin: Summary

If merged, this PR fixes the issue with selectableRows state not being maintained properly after model update.

_(**NOTE:** Initial commits include test code from Michael Lange which will be removed before merge.)_

👉 See [related PR](https://github.com/hashicorp/design-system/pull/2667) to update docs for multi select table used with a model.

TODO: Add/update related tests.

**PREVIEW:** https://hds-showcase-git-hds-4273-table-multi-select-bug-hashicorp.vercel.app/components/table-multi-select-bug

**Bug:** Originally, clicking "Force new data" button in the first example wiped out the selectableRows array due to timing of didInsert & willDestroy calls.

**Expected:** SelectableRows array should have the proper set of checkbox items after model update and willDestroy calls finish.

### :link: External links

* Jira ticket: [HDS-4273](https://hashicorp.atlassian.net/browse/HDS-4273)

***

### 👀 Component checklist

[HDS-4273]: https://hashicorp.atlassian.net/browse/HDS-4273?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/table/index.ts
+++ b/packages/components/src/components/hds/table/index.ts
@@ -300,6 +300,11 @@ export default class HdsTable extends Component<HdsTableSignature> {
     );
 
     if (selectionKey) {
+      // Remove any existing item with the same `selectionKey` identifier (this may occur if the model is updated and the rows re-rendered)
+      this._selectableRows = this._selectableRows.filter(
+        (row): boolean => row.selectionKey !== selectionKey
+      );
+      // Add the new item
       this._selectableRows.push({ selectionKey, checkbox });
     }
     this.setSelectAllState();
@@ -314,22 +319,6 @@ export default class HdsTable extends Component<HdsTableSignature> {
       this._selectableRows.map((row): string => row.selectionKey)
     );
 
-    // Fix for selectableRows state not being maintained properly when model updates:
-    // Delete the first row with the given selection key you find from the selectableRows array
-    // Search for the index of the row to delete. If found, delete it.
-    // (Fixes issue of not rows not being properly selectable including "select all" functionality)
-    const rowToDeleteIndex = this._selectableRows.findIndex(
-      (row): boolean => row.selectionKey === selectionKey
-    );
-
-    if (rowToDeleteIndex > -1) {
-      this._selectableRows.splice(rowToDeleteIndex, 1);
-    }
-
-    // ORIGINAL:
-    // this._selectableRows = this._selectableRows.filter(
-    //   (row): boolean => row.selectionKey !== selectionKey
-    // );
     this.setSelectAllState();
   }
 
--- a/packages/components/src/components/hds/table/index.ts
+++ b/packages/components/src/components/hds/table/index.ts
@@ -105,7 +105,7 @@ export default class HdsTable extends Component<HdsTableSignature> {
   get getSortCriteria(): string | HdsTableSortingFunction<unknown> {
     // get the current column
     const currentColumn = this.args?.columns?.find(
-      (column) => column.key === this.sortBy
+      (column): boolean => column.key === this.sortBy
     );
     if (
       // check if there is a custom sorting function associated with the current `sortBy` column (we assume the column has `isSortable`)
@@ -124,7 +124,9 @@ export default class HdsTable extends Component<HdsTableSignature> {
     if (this.args.identityKey === 'none') {
       return undefined;
     } else {
-      return this.args.identityKey ?? '@identity';
+      // We return a default key to cause an update operation instead of replacement when the model changes to side-step
+      // a bug in Ember which would cause the table to lose its selection state
+      return this.args.identityKey ?? 'hds-table-key';
     }
   }
 
@@ -289,6 +291,13 @@ export default class HdsTable extends Component<HdsTableSignature> {
     checkbox: HdsFormCheckboxBaseSignature['Element'],
     selectionKey?: string
   ): void {
+    // TEST:
+    console.log(
+      `[${this.args.caption}] Did insert row checkbox`,
+      selectionKey,
+      this._selectableRows.map((row) => row.selectionKey)
+    );
+
     if (selectionKey) {
       this._selectableRows.push({ selectionKey, checkbox });
     }
@@ -297,8 +306,15 @@ export default class HdsTable extends Component<HdsTableSignature> {
 
   @action
   willDestroyRowCheckbox(selectionKey?: string): void {
+    // TEST:
+    console.log(
+      `[${this.args.caption}] Will destroy row checkbox`,
+      selectionKey,
+      this._selectableRows.map((row) => row.selectionKey)
+    );
+
     this._selectableRows = this._selectableRows.filter(
-      (row) => row.selectionKey !== selectionKey
+      (row): boolean => row.selectionKey !== selectionKey
     );
     this.setSelectAllState();
   }
@@ -308,7 +324,7 @@ export default class HdsTable extends Component<HdsTableSignature> {
     if (this._selectAllCheckbox) {
       const selectableRowsCount = this._selectableRows.length;
       const selectedRowsCount = this._selectableRows.filter(
-        (row) => row.checkbox.checked
+        (row): boolean => row.checkbox.checked
       ).length;
 
       this._selectAllCheckbox.checked =
--- a/packages/components/src/components/hds/table/index.ts
+++ b/packages/components/src/components/hds/table/index.ts
@@ -126,7 +126,7 @@ export default class HdsTable extends Component<HdsTableSignature> {
     } else {
       // We return a default key to cause an update operation instead of replacement when the model changes to side-step
       // a bug in Ember which would cause the table to lose its selection state
-      return this.args.identityKey ?? 'hds-table-key';
+      return this.args.identityKey ?? '@identity';
     }
   }
 
@@ -295,7 +295,7 @@ export default class HdsTable extends Component<HdsTableSignature> {
     console.log(
       `[${this.args.caption}] Did insert row checkbox`,
       selectionKey,
-      this._selectableRows.map((row) => row.selectionKey)
+      this._selectableRows.map((row): string => row.selectionKey)
     );
 
     if (selectionKey) {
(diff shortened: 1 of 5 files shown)


packages/components/src/components/hds/table/index.ts
```diff
@@ -124,7 +124,9 @@ export default class HdsTable extends Component<HdsTableSignature> {
     if (this.args.identityKey === 'none') {
       return undefined;
     } else {
-      return this.args.identityKey ?? '@identity';
+      // We return a default key to cause an update operation instead of replacement when the model changes to side-step
+      // a bug in Ember which would cause the table to lose its selection state
+      return this.args.identityKey ?? 'hds-table-key';
```
KristinLBradley, line 129: I added a default value here to prevent the bug with selections being wiped out when the model updates. Are there any potential issues with this approach to consider?

I wasn't sure why the original value included "@". @MelSumner ?

KristinLBradley: **Answer:** This is a special key value in Ember: https://api.emberjs.com/ember/6.1/classes/Ember.Templates.helpers/methods/each?anchor=each

I am investigating a different approach.

KristinLBradley: Pushed an update

packages/components/src/components/hds/table/index.ts
```diff
@@ -297,9 +306,29 @@ export default class HdsTable extends Component<HdsTableSignature> {
...
+    // Delete the first row with the given selction key you find from the selectableRows array
+    // Search for the index of the row to delete. If found, delete it.
+    // (Fixes issue of not rows not being properly selectable including "select all" functionality)
+    const rowToDeleteIndex = this._selectableRows.findIndex(
+      (row): boolean => row.selectionKey === selectionKey
     );
+
+    if (rowToDeleteIndex > -1) {
+      this._selectableRows.splice(rowToDeleteIndex, 1);
+    }
```
KristinLBradley, line 326: This fixes what seems to be the actual problem. Would appreciate feedback though. (I'll clean up all the messy commented out & test code later.)

packages/components/src/components/hds/table/index.ts
```diff
@@ -124,6 +124,8 @@ export default class HdsTable extends Component<HdsTableSignature> {
     if (this.args.identityKey === 'none') {
       return undefined;
     } else {
+      // We return a default key to cause an update operation instead of replacement when the model changes to side-step
+      // a bug in Ember which would cause the table to lose its selection state
```
didoo, line 128: [issue?] I am not sure this comment is correct. We return the `@identity`
https://api.emberjs.com/ember/6.1/classes/Ember.Templates.helpers/methods/each?anchor=each

```suggestion
      // We return `@identity` as default to cause an update operation instead of replacement when the model changes to avoid unexpected side effects
      // see: "Specifying Keys" here: https://api.emberjs.com/ember/6.1/classes/Ember.Templates.helpers/methods/each?anchor=each
      // see also: https://github.com/hashicorp/design-system/pull/1160 + https://github.com/hashicorp/cloud-ui/pull/5178 + https://hashicorp.atlassian.net/browse/HDS-4273
```

showcase/app/controllers/components/table-multi-select-bug.js
```diff
@@ -0,0 +1,37 @@
...
+  { key: 'twelve', quantity: 12, type: 'musician, technically', line: 'Twelve drummers drumming' },
+];
+
+export default class extends Controller {
+  @tracked showTable = false;
+  @tracked lyrics = daysOfChristmas.slice();
+
+  @action dirtyModel() {
+    // Deep copy to destroy references
+    this.lyrics = this.lyrics.map((r) => ({ ...r }));
```
didoo, line 32: [suggestion] this will make problems even more evident (if there are)
```suggestion
    this.lyrics = this.lyrics
      // Deep copy to destroy references
      .map((r) => ({ ...r }))
      // Randomly sort to see what happens in this case
      .sort(() => Math.random() - 0.5);
```

packages/components/src/components/hds/table/index.ts
didoo: [though] One consideration: instead of trying to fix things once the array has been populated with duplicates, what if we avoid adding duplicates in the first place? We could update the `didInsertRowCheckbox` method in the `table/index.ts` file, in this way:

```
  @action
  didInsertRowCheckbox(
    checkbox: HdsFormCheckboxBaseSignature['Element'],
    selectionKey?: string
  ): void {
    if (selectionKey) {
      // Remove any existing item with the same `selectionKey` identifier (this may occurr if the model is updated and the rows re-rendered)
      this._selectableRows = this._selectableRows.filter(
      (row) => row.selectionKey !== selectionKey
      );
      // Add the new item
      this._selectableRows.push({ selectionKey, checkbox });
    }
    this.setSelectAllState();
  }
```

didoo (commented):
@KristinLBradley I added a few thoughts/comments, but I want to double check with @DingoEatingFuzz if I'm understanding correctly the problem (I may be missing something).

didoo:
@KristinLBradley @alex-ju I had a quick call with @DingoEatingFuzz (thanks 🙏) to see if I was understanding correctly the problem and if the solution(s) proposed are be what he would expect to see, as a "consumer" of HDS. 

A few points below, to summarize what we discussed:
- the solution(s) proposed in this PR (be it to remove only the "old" row from the `_selectableRows` on `willDestroyRowCheckbox`, or to remove the duplicate row in the `didInsertRowCheckbox`) will solve the main problem, which is having the "select all" checkbox in an unusable state; but the problem that the checkbox status is lost when the model is updated remains
- using the `@identityKey` argument solves this problem, but Michael (as consumer) didn't know that he had to add the `@identityKey` for this, and it took him some time to figure it out, and that's not great
- he would prefer to see the HDS component take care of this, but (personally) I am not sure how to do it without something that would amount to a refactoring of the internals/logic of the multi-select table; I'll get back to this in a sec
- we have also checked how many instances of the selectable table there are ([not many](https://github.com/search?q=%40isSelectable+user%3Ahashicorp+-repo%3Ahashicorp%2Fdesign-system&type=code&ref=advsearch)) so this is more of a special case (the one in which consumers _have to_ add the `identityKey` to fix something that doesn't work as expected)

I'll leave it to him chime in, in case I've forgotten/misinterpreted something, or if he wants to add more context or his point og view as HDS "consumer".

didoo:
Re. the problem of the loss of the checkbox state when the model is updated but there is no `@identityKey` specified (so it defaults to `@identity`) one option I've tried is to just recycle the existing item in the "_selectableRows", but even in that case Ember re-renders the rows (exactly because it sees them as new "rows" in the model, so it's not a viable option to preserve the state without having an `@identityKey`.

didoo:
@KristinLBradley @alex-ju + cc @DingoEatingFuzz at this point I think we have limited options in front of us:
- refactor the component, to avoid "reading" the state from the checkbox, but having an internal proxy for it [big job]
- fix "just" the problem of the select all checkbox getting in the messy state by
   - remove only the "old" row from the `_selectableRows` on `willDestroyRowCheckbox`
   - remove the duplicate row in the `didInsertRowCheckbox` (my preferred solution)

Thoughts? Alternatives?

packages/components/src/components/hds/table/index.ts
KristinLBradley: What if the set of new keys doesn't completely correspond to the set of old keys?

For example, if in the model update there are fewer rows than previously. Would that be a concern or can we assume that the set of keys will be the same between updates?

KristinLBradley: Thanks, I updated the code.

I was wondering though if it's an issue if the new model has a different number of rows or different keys or something. I added another example to the Showcase page to test this and it doesn't work correctly but I don't know if it's an actual realistic concern.

KristinLBradley:
> @KristinLBradley @alex-ju + cc @DingoEatingFuzz at this point I think we have limited options in front of us:
> 
> * refactor the component, to avoid "reading" the state from the checkbox, but having an internal proxy for it [big job]
> * fix "just" the problem of the select all checkbox getting in the messy state by
>   
>   * remove only the "old" row from the `_selectableRows` on `willDestroyRowCheckbox`
>   * remove the duplicate row in the `didInsertRowCheckbox` (my preferred solution)
> 
> Thoughts? Alternatives?

@didoo @alex-ju + cc @DingoEatingFuzz

Removing the duplicate row seems to work as long as there is a one to one correspondance between the new rows and the old rows. But it breaks if there are a different number of rows and/or key changes. So can we be certain that the rows won't change between model updates?

KristinLBradley:
> @KristinLBradley @alex-ju I had a quick call with @DingoEatingFuzz (thanks 🙏) to see if I was understanding correctly the problem and if the solution(s) proposed are be what he would expect to see, as a "consumer" of HDS.
> 
> A few points below, to summarize what we discussed:
> 
> * the solution(s) proposed in this PR (be it to remove only the "old" row from the `_selectableRows` on `willDestroyRowCheckbox`, or to remove the duplicate row in the `didInsertRowCheckbox`) will solve the main problem, which is having the "select all" checkbox in an unusable state; but the problem that the checkbox status is lost when the model is updated remains
> * using the `@identityKey` argument solves this problem, but Michael (as consumer) didn't know that he had to add the `@identityKey` for this, and it took him some time to figure it out, and that's not great
> * he would prefer to see the HDS component take care of this, but (personally) I am not sure how to do it without something that would amount to a refactoring of the internals/logic of the multi-select table; I'll get back to this in a sec
> * we have also checked how many instances of the selectable table there are ([not many](https://github.com/search?q=%40isSelectable+user%3Ahashicorp+-repo%3Ahashicorp%2Fdesign-system&type=code&ref=advsearch)) so this is more of a special case (the one in which consumers _have to_ add the `identityKey` to fix something that doesn't work as expected)
> 
> I'll leave it to him chime in, in case I've forgotten/misinterpreted something, or if he wants to add more context or his point og view as HDS "consumer".

@didoo In addition to the current code changes, we could also update the documentation to mention that if you want the state of the checkboxes to persist after the model updates, you'll need to add an identityKey.

Do you think this the current changes together with updating the docs would be enough to address the main problem?

didoo:
> Removing the duplicate row seems to work as long as there is a one to one correspondance between the new rows and the old rows. But it breaks if there are a different number of rows and/or key changes. So can we be certain that the rows won't change between model updates?

@KristinLBradley I've played quite some time with the code yesterday, and I wasn't able to find an OK solution. The root of the problem is the timing of `did-insert` vs `will-destroy` (the first one is deterministic, the second one is not). The solution that you have found works "by chance" (and the one I proposed works only in certain cases) but I personally think both are not "the right one" from a code logic perspective. If you want I can discuss more (maybe in a pairing session) or I can have a look and propose something when we're back from the offsite. In the meantime I would suggest to put on hold on this task (apart from the documentation update, see below).

> @didoo In addition to the current code changes, we could also update the documentation to mention that if you want the state of the checkboxes to persist after the model updates, you'll need to add an identityKey.

Yes, this is something that I would do anyway, independently from this ticket/task, that is meant to fix the case for when the consumer is _not_ adding the `identityKey` (even if technically is not a fix, it's a more graceful failure to maintain the state)

> Do you think this the current changes together with updating the docs would be enough to address the main problem?

See my comments above.

/cc @alex-ju @DingoEatingFuzz

KristinLBradley:
> > @didoo In addition to the current code changes, we could also update the documentation to mention that if you want the state of the checkboxes to persist after the model updates, you'll need to add an identityKey.
> 
> Yes, this is something that I would do anyway, independently from this ticket/task, that is meant to fix the case for when the consumer is _not_ adding the `identityKey` (even if technically is not a fix, it's a more graceful failure to maintain the state)

@didoo

How is this? I opened a PR: https://github.com/hashicorp/design-system/pull/2667

KristinLBradley:
Closing as stale

PR #2655, opened 2025-01-17
`Tag` - Text truncation for overflow fix

dchyun:
### :pushpin: Summary

This PR fixes an overflow issue present in the `Tag` when text wraps to > 3 lines. It implements the intended design change to address this, which truncates any text that goes to multiple lines. When truncation occurs, an ellipsis is rendered, and a tooltip showing the full text is added.

It also adds `width: fit-content;` to the tag to ensure a tag's width never goes beyond its content.

It was observed in the Tag in [HDS-3854](https://hashicorp.atlassian.net/browse/HDS-3854) that when the text exceeds 3 lines, the background of the dismiss button obscures the border of the element. 

Per design guidance, the chosen solution has been to truncate any text in the tag that goes beyond one line, and add an ellipsis and tooltip with the full text when this truncation occurs. 

The text is truncated when the text would wrap to multiple lines using `overflow: hidden` and `-webkit-line-clamp: 1`. When this occurs a tooltip is also added using the `hds-tooltip` modifier, and the `Hds::TooltipButton`.

There are several other changes related to this fix.

- The `max-width` of the text is set to `150px`
- A new property `@tooltipPlacement` has been added to allow for customization of the tooltip placement
- `width: fit-content` has been added to prevent the tag width from ever exceeding its content. This was present previously when the tag had a parent with `display: grid`. 

#### Text Truncation

**Before**
Screenshot 2025-01-24 at 9 24 43 AM

**After**
Screenshot 2025-02-10 at 4 27 06 PM

**After - Hover / Focus**
Screenshot 2025-02-10 at 4 25 52 PM

#### Width

**Before**
Screenshot 2025-02-05 at 9 28 22 AM

**After**
Screenshot 2025-02-10 at 4 26 07 PM

### :link: External links

Jira ticket: [HDS-4317](https://hashicorp.atlassian.net/browse/HDS-4317)
Figma file: [Link](https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/branch/TQ8Xv7T6DPmLtpFMbe9Bv7/HDS-Components-v2.0?m=auto&node-id=72325-1458&t=2szZ94wgmze0xjgO-1)

***

### 👀 Component checklist

[HDS-4317]: https://hashicorp.atlassian.net/browse/HDS-4317?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[HDS-3854]: https://hashicorp.atlassian.net/browse/HDS-3854?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/brown-wolves-admire.md
+++ b/.changeset/brown-wolves-admire.md
@@ -2,6 +2,6 @@
 "@hashicorp/design-system-components": minor
 ---
 
-`Tag` - Truncate any text that is longer than 150px and add a tooltip with the full text when truncation occurs
+`Tag` - Truncate any text that is longer than about 20 characters (160px) and add a tooltip with the full text when truncation occurs
 
 `Tag` - Added `@tooltipPlacement` argument
--- a/.changeset/brown-wolves-admire.md
+++ b/.changeset/brown-wolves-admire.md
@@ -3,3 +3,5 @@
 ---
 
 `Tag` - Truncate any text that wraps to multiple lines and add a tooltip with the full text when truncation occurs
+
+`Tag` - Added `@tooltipPlacement` argument
--- a/.changeset/brown-wolves-admire.md
+++ b/.changeset/brown-wolves-admire.md
@@ -2,6 +2,6 @@
 "@hashicorp/design-system-components": minor
 ---
 
-`Tag` - Truncate any text that is longer than about 20 characters (160px) and add a tooltip with the full text when truncation occurs
+`Tag` - Truncate any text that is longer than about 20 characters, and add a tooltip with the full text when truncation occurs
 
 `Tag` - Added `@tooltipPlacement` argument
--- a/.changeset/brown-wolves-admire.md
+++ b/.changeset/brown-wolves-admire.md
@@ -2,6 +2,6 @@
 "@hashicorp/design-system-components": minor
 ---
 
-`Tag` - Truncate any text that wraps to multiple lines and add a tooltip with the full text when truncation occurs
+`Tag` - Truncate any text that is longer than 150px and add a tooltip with the full text when truncation occurs
 
 `Tag` - Added `@tooltipPlacement` argument
--- a/.changeset/brown-wolves-admire.md
+++ b/.changeset/brown-wolves-admire.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Tag` - Truncate any text that wraps to multiple lines and add a tooltip with the full text when truncation occurs
--- a/packages/components/src/components/hds/tag/index.hbs
+++ b/packages/components/src/components/hds/tag/index.hbs
@@ -2,14 +2,23 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<Hds::Text::Body class={{this.classNames}} @tag="span" @size="100" @weight="medium" @color="primary" ...attributes>
+<Hds::Text::Body
+  class={{this.classNames}}
+  @tag="span"
+  @size="100"
+  @weight="medium"
+  @color="primary"
+  {{did-insert this.didInsert}}
+  {{will-destroy this.willDestroyNode}}
+  ...attributes
+>
   {{#if this.onDismiss}}
     <button class="hds-tag__dismiss" type="button" aria-label={{this.ariaLabel}} {{on "click" this.onDismiss}}>
       <Hds::Icon class="hds-tag__dismiss-icon" @name="x" @size="16" />
     </button>
   {{/if}}
   {{#if (or @href @route)}}
-    {{#if this.isTextTruncated}}
+    {{#if this._isTextOverflow}}
       <Hds::Interactive
         class="hds-tag__link"
         @current-when={{@current-when}}
@@ -22,7 +31,9 @@
         @isHrefExternal={{@isHrefExternal}}
         {{hds-tooltip this.text options=(hash placement="right")}}
       >
-        {{this.text}}
+        <div class="hds-tag__text-container">
+          {{this.text}}
+        </div>
       </Hds::Interactive>
     {{else}}
       <Hds::Interactive
@@ -42,15 +53,12 @@
       </Hds::Interactive>
     {{/if}}
   {{else}}
-    {{#if this.isTextTruncated}}
-      <span
-        class="hds-tag__text"
-        {{hds-tooltip this.text options=(hash placement="right")}}
-      >
+    {{#if this._isTextOverflow}}
+      <Hds::TooltipButton class="hds-tag__text" @text={{this.text}} @placement="right">
         <div class="hds-tag__text-container">
           {{this.text}}
         </div>
-      </span>
+      </Hds::TooltipButton>
     {{else}}
       <span class="hds-tag__text">
         <div class="hds-tag__text-container">
--- a/packages/components/src/components/hds/tag/index.hbs
+++ b/packages/components/src/components/hds/tag/index.hbs
@@ -8,8 +8,7 @@
   @size="100"
   @weight="medium"
   @color="primary"
-  {{did-insert this.didInsert}}
-  {{will-destroy this.willDestroyNode}}
+  {{this._setUpObserver}}
   ...attributes
 >
   {{#if this.onDismiss}}
--- a/packages/components/src/components/hds/tag/index.hbs
+++ b/packages/components/src/components/hds/tag/index.hbs
@@ -36,7 +36,9 @@
         @href={{@href}}
         @isHrefExternal={{@isHrefExternal}}
       >
-        {{this.text}}
+        <div class="hds-tag__text-container">
+          {{this.text}}
+        </div>
       </Hds::Interactive>
     {{/if}}
   {{else}}
@@ -45,11 +47,15 @@
         class="hds-tag__text"
         {{hds-tooltip this.text options=(hash placement="right")}}
       >
-        {{this.text}}
+        <div class="hds-tag__text-container">
+          {{this.text}}
+        </div>
       </span>
     {{else}}
       <span class="hds-tag__text">
-        {{this.text}}
+        <div class="hds-tag__text-container">
+          {{this.text}}
+        </div>
       </span>
     {{/if}}
   {{/if}}
--- a/packages/components/src/components/hds/tag/index.hbs
+++ b/packages/components/src/components/hds/tag/index.hbs
@@ -28,7 +28,7 @@
         @isRouteExternal={{@isRouteExternal}}
         @href={{@href}}
         @isHrefExternal={{@isHrefExternal}}
-        {{hds-tooltip this.text options=(hash placement="right")}}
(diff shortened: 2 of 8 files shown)


majedelass (commented):
This is looking great, although I did find an interesting "bug"
image

The tooltip is still there even if the text is fully revealed. Wonder if there is a way to fix this?

dchyun:
> This is looking great, although I did find an interesting "bug" image
> 
> The tooltip is still there even if the text is fully revealed. Wonder if there is a way to fix this?

Did you notice this after doing some resizing of the page, or on load? One open issue with the fix is that the tooltip will be added if the text goes from not cut off to cut off on a resize, but not removed if the text goes from cut off to shown. 

I figure this isn't a use case that will come up often as users aren't usually resizing their browser back and forth, but will see if I can find some fix.

MelSumner:
Def want to call out that the there needs to be documentation that informs consumers that truncated text is not accessible for all users and they should prefer limiting the text length before the need for truncation exists.

majedelass:
> Def want to call out that the there needs to be documentation that informs consumers that truncated text is not accessible for all users and they should prefer limiting the text length before the need for truncation exists.

Can you elaborate on this? Truncated text will always have tooltips associated to it. Also from my understanding, truncation is a CSS based visual indicator that should still be accessible by a screen reader. One call out is to make sure that the text isn't read out twice. Is there something else I'm not seeing?

dchyun:
> > Def want to call out that the there needs to be documentation that informs consumers that truncated text is not accessible for all users and they should prefer limiting the text length before the need for truncation exists.
> 
> Can you elaborate on this? Truncated text will always have tooltips associated to it. Also from my understanding, truncation is a CSS based visual indicator that should still be accessible by a screen reader. One call out is to make sure that the text isn't read out twice. Is there something else I'm missing?

Yes the truncation is only via CSS. In the DOM the text will still be un-truncated. I did test with voiceover and this does mean that the text does get read twice, which isn't ideal, but no content is lost. It reads the text of the button and then the tooltip associated with it. 

Screenshot 2025-01-24 at 10 04 31 AM
Screenshot 2025-01-24 at 10 07 01 AM

majedelass:
> > > Def want to call out that the there needs to be documentation that informs consumers that truncated text is not accessible for all users and they should prefer limiting the text length before the need for truncation exists.
> > 
> > 
> > Can you elaborate on this? Truncated text will always have tooltips associated to it. Also from my understanding, truncation is a CSS based visual indicator that should still be accessible by a screen reader. One call out is to make sure that the text isn't read out twice. Is there something else I'm missing?
> 
> Yes the truncation is only via CSS. In the DOM the text will still be un-truncated. I did test with voiceover and this does mean that the text does get read twice, which isn't ideal, but no content is lost. It reads the text of the button and then the tooltip associated with it.
> 
> Screenshot 2025-01-24 at 10 04 31 AM Screenshot 2025-01-24 at 10 07 01 AM

I wonder if there is a way to assign the .sr-only class name to truncated text and then remove it if its un-truncated? Seems like a whole level of complexity.

packages/components/src/components/hds/tag/index.hbs
```diff
@@ -2,29 +2,69 @@
...
 }}
-<Hds::Text::Body class={{this.classNames}} @tag="span" @size="100" @weight="medium" @color="primary" ...attributes>
+<Hds::Text::Body
+  class={{this.classNames}}
+  @tag="span"
+  @size="100"
+  @weight="medium"
+  @color="primary"
+  {{did-insert this.didInsert}}
+  {{will-destroy this.willDestroyNode}}
```
zamoore, line 12: Suggestion: If you're up for it, I would try to set this up with a custom modifier. Happy to pair on it.

packages/components/src/components/hds/tag/index.ts
```diff
@@ -104,4 +109,33 @@ export default class HdsTag extends Component<HdsTagSignature> {
...
+  }
+
+  @action
+  willDestroyNode(): void {
+    super.willDestroy();
+    this._observer.disconnect();
+  }
+
+  private _isOverflow(el: Element): boolean {
+    if (el.scrollHeight > el.clientHeight) {
```
zamoore, line 135: I would suggest refactoring this to
`return el.scrollHeight > el.clientHeight;`

zamoore (commented):
This makes sense and looks great, but I think we can improve it with custom modifiers. Let's set up some time next week to work on this if you'd like. Or if you'd prefer to explore the modifier conversion on your own, that's cool too.

packages/components/src/components/hds/tag/index.hbs
dchyun: I was thinking the same thing after we talked about custom modifiers last week. I've refactored it now to use a custom modifier. Lmk if it looks like there are any issues with the implementation.

zamoore: Looks great!

zamoore (approved):
Nice work. Good use of Ember modifiers!

packages/components/src/styles/components/tag.scss
```diff
@@ -15,6 +15,7 @@ $hds-tag-border-radius: 50px;
 .hds-tag {
   display: inline-flex;
   align-items: stretch;
+  width: fit-content;
```
KristinLBradley, line 18: You should test if it could possibly exceed the width of its container with this style set. May want to add `max-width: 100%`

KristinLBradley: I did a bit of quick testing and it doesn't seem very likely that the tag would overflow it's container but it _is_ possible with containers that have very narrow widths. So you _could_ add `.hds-tag, .hds-tag__link {max-width: 100%;}` to be extra cautious although it could be overkill so up to you...

KristinLBradley: Thanks, just be sure to test yourself to confirm the effect as I just did quick in browser tests.

alex-ju (approved):
Left one suggestion – otherwise looking great!

packages/components/src/styles/components/tag.scss
```diff
@@ -76,6 +86,18 @@ $hds-tag-border-radius: 50px;
   }
 }
 
+.hds-tooltip-button.hds-tag__text {
```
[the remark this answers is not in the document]
dchyun: Great catch! Updated to add this in.

showcase/tests/integration/components/hds/tag/index-test.js
```diff
@@ -82,4 +87,22 @@ module('Integration | Component | hds/tag/index', function (hooks) {
...
+  test('it should not render a tooltip if the text does not overflow', async function (assert) {
+    await render(hbs`
+        <Hds::Tag @text="My text tag" id="test-tag"/>
+    `);
+    assert.dom('.hds-tooltip-button').doesNotExist();
+  });
+
+  test('it should render a tooltip if the text overflows', async function (assert) {
+    await render(hbs`
+      <div style="width: 50px;">
```
dchyun, line 102: [Question] This tests seems to cause failures in the CI in some cases, but passes when re-ran manually. Should it be removed? Or is there another pattern to follow for testing in narrow viewports we use?

packages/components/src/components/hds/tag/index.ts
```diff
@@ -25,6 +27,25 @@ export interface HdsTagSignature {
 }
 
 export default class HdsTag extends Component<HdsTagSignature> {
+  @tracked private _isTextOverflow!: boolean;
+  private _observer!: ResizeObserver;
+
+  private _setUpObserver = modifier((element: HTMLElement) => {
```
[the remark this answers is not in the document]
shleewhite: all credit to dylan, I followed his pattern :)

alex-ju (approved):

.changeset/brown-wolves-admire.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Tag` - Truncate any text that wraps to multiple lines and add a tooltip with the full text when truncation occurs
```
shleewhite, line 5: Suggestion: might be good to explicitly call out the new max-width

```suggestion
`Tag` - Truncate any text that is longer than 150px and add a tooltip with the full text when truncation occurs
```

showcase/app/templates/components/tag.hbs
```diff
@@ -33,12 +33,22 @@
   </Shw::Flex>
 
   <Shw::Flex @label="With long text" as |SF|>
-    <SF.Item {{style width="200px"}}>
+    <SF.Item {{style width="176px"}}>
```
shleewhite, line 36: Suggestion: These examples don't need the `width` set now that there is a `max-width` on the tag text.

Screenshot 2025-02-11 at 9 48 39 AM

shleewhite (commented):
Just a couple small suggestions, looks good!

shleewhite (approved):

.changeset/brown-wolves-admire.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Tag` - Truncate any text that is longer than 150px and add a tooltip with the full text when truncation occurs
```
shleewhite, line 5:
```suggestion
`Tag` - Truncate any text that is longer than 160px and add a tooltip with the full text when truncation occurs
```

Is this more accurate now?

shleewhite (dismissed):

shleewhite (dismissed):

majedelass (approved):
Visually looks good, and the max width values you put in are the same as in Figma :)

packages/components/src/styles/components/tag.scss
```diff
@@ -41,12 +43,23 @@ $hds-tag-border-radius: 50px;
...
+  overflow: hidden;
+  word-break: break-all;
+  -webkit-box-orient: vertical;
+  -webkit-line-clamp: 1;
+  line-clamp: 1;
+}
+
 .hds-tag__dismiss ~ .hds-tag__text,
 .hds-tag__dismiss ~ .hds-tag__link {
+  max-width: 160px;
```
KristinLBradley, line 62: I would add a brief comment here explaining why you are overriding the 160px width set above. (e.g., to account for the width of the dismiss button)

KristinLBradley (dismissed):
Looks good overall. I added a minor suggestion to add a comment to one of the styles.

Note: If the visual diff is as expected you can approve the Percy diff to get rid of the related test failure.

packages/components/src/styles/components/tag.scss
dchyun: Good suggestion. The intended width is 160, but we bump it to 166 to account for the excess padding on the text in the non-dismissible variant, so added a comment on the other `max-width` line.

KristinLBradley (approved):

shleewhite (approved):
🚢

majedelass (approved):
Looks great! :shipit:

PR #2613, opened 2024-12-18
Add script to generate component changelog entries

alex-ju:
### :pushpin: Summary

This PR proposes a set of changes aiming to automate the component changelog entries.
 
#### Group component changelog entries in a single markdown file

For illustration, I did this for the components updated in 4.15.0. ~~If agreed I will update the rest of the components.~~ After the initial review I updated all components.

```diff
- @include "partials/version-history/4.15.0.md"
- @include "partials/version-history/4.10.0.md"
+ @include "partials/version-history/version-history.md"
```

#### Update the main component changelog to follow convention

Adjusted a few changelog entries in 4.15.0. I propose we follow a one-line format moving forward, as much as possible, to make the automation more reliable:
```md
`Component` - Added `@argument` to `Component::Subcomponent`
`Component` - Fixed something
`Component` - Deprecated `@oldArgument`
```

#### Add script to generate component changelog entries

Running the following command in `changeset-release/main` when preparing to release will append changelog entries to each component, as shown in https://github.com/hashicorp/design-system/commit/e4dcdd6b19c660c898a0767e2f62bb036ecbd8f4.

```bash
yarn run generate-component-changelog-entries
```

Once the component changelogs are generated they can be adjusted as needed before committing to the changeset branch.

### :link: External links

Jira ticket: [HDS-4215](https://hashicorp.atlassian.net/browse/HDS-4215)

***

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-4215]: https://hashicorp.atlassian.net/browse/HDS-4215?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2656, opened 2025-01-20
Automate component version badges

alex-ju:
### :pushpin: Summary

Follow-up on #2613 to automate the removal and addition of version badges for each component.

 - e82ea680a5d1635ee1430008ad05341553668ffe update component changelog script to handle badges 
 - e82ea680a5d1635ee1430008ad05341553668ffe as we use `gray-matter` (a slim wrapper of `js-yaml`) to read/write the front matter we updated components' front matter to match the standard schema

#### Outcome

Running the automated script revealed a few instances where the 4.15 badge was missing. Any `added` or `updated` badge before 4.15 is removed.

> [!NOTE]
> This PR will be updated to match the agreed proposal outlined by @zamoore in the release cycle RFC.

#### Limitations

The script doesn't generate `added` or `deprecated` badges (only `updated`). When you add a new component (or deprecate one), you have to either update the front matter status when updating the associated docs or change the badge type after running the automated script. These details will be covered in the confluence documentation updates.

### :link: External links

Jira ticket: [HDS-4299](https://hashicorp.atlassian.net/browse/HDS-4299)

***

[HDS-4299]: https://hashicorp.atlassian.net/browse/HDS-4299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/docs/components/badge/index.md
+++ b/website/docs/components/badge/index.md
@@ -1,14 +1,22 @@
 ---
 title: Badge
-description: Concise, non-interactive labels that represent metadata.
-caption: Concise, non-interactive labels that represent metadata.
+description: 'Concise, non-interactive labels that represent metadata.'
+caption: 'Concise, non-interactive labels that represent metadata.'
 links:
-  figma: https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/HDS-Components-v2.0?node-id=67237-59854&t=w8xQlWxzH7bwXLe2-1
-  github: https://github.com/hashicorp/design-system/tree/main/packages/components/src/components/hds/badge
-related: ['components/badge-count','components/tag']
+  figma: >-
+    https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/HDS-Components-v2.0?node-id=67237-59854&t=w8xQlWxzH7bwXLe2-1
+  github: >-
+    https://github.com/hashicorp/design-system/tree/main/packages/components/src/components/hds/badge
+related:
+  - components/badge-count
+  - components/tag
 previewImage: assets/illustrations/components/badge.jpg
 navigation:
-  keywords: ['chip', 'pill', 'tag', 'label']
+  keywords:
+    - chip
+    - pill
+    - tag
+    - label
 ---
 
 <section data-tab="Guidelines">
--- a/website/docs/components/card/index.md
+++ b/website/docs/components/card/index.md
@@ -1,13 +1,18 @@
 ---
 title: Card
-description: A block container that provides styling for elevation, border, and background.
-caption: A block container that provides styling for elevation, border, and background.
+description: 'A block container that provides styling for elevation, border, and background.'
+caption: 'A block container that provides styling for elevation, border, and background.'
 links:
-  figma: https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/HDS-Components-v2.0?node-id=67251-87362&t=w8xQlWxzH7bwXLe2-1
-  github: https://github.com/hashicorp/design-system/tree/main/packages/components/src/components/hds/card
+  figma: >-
+    https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/HDS-Components-v2.0?node-id=67251-87362&t=w8xQlWxzH7bwXLe2-1
+  github: >-
+    https://github.com/hashicorp/design-system/tree/main/packages/components/src/components/hds/card
 previewImage: assets/illustrations/components/card.jpg
 navigation:
-  keywords: ['tile', 'container', 'box']
+  keywords:
+    - tile
+    - container
+    - box
 ---
 
 <section data-tab="Guidelines">
--- a/website/docs/components/table/index.md
+++ b/website/docs/components/table/index.md
@@ -1,19 +1,24 @@
 ---
 title: Table
-description: Used to display organized, two-dimensional tabular data.
-caption: Used to display organized, two-dimensional tabular data.
+description: 'Used to display organized, two-dimensional tabular data.'
+caption: 'Used to display organized, two-dimensional tabular data.'
 links:
-  figma: https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/HDS-Components-v2.0?node-id=67216-35163&t=w8xQlWxzH7bwXLe2-1
-  github: https://github.com/hashicorp/design-system/tree/main/packages/components/src/components/hds/table
+  figma: >-
+    https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/HDS-Components-v2.0?node-id=67216-35163&t=w8xQlWxzH7bwXLe2-1
+  github: >-
+    https://github.com/hashicorp/design-system/tree/main/packages/components/src/components/hds/table
 related:
-  [
-    'components/pagination',
-    'patterns/filter-patterns',
-    'patterns/table-multi-select',
-  ]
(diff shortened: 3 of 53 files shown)


website/docs/components/badge/index.md
```diff
@@ -1,14 +1,22 @@
 ---
 title: Badge
-description: Concise, non-interactive labels that represent metadata.
-caption: Concise, non-interactive labels that represent metadata.
+description: 'Concise, non-interactive labels that represent metadata.'
```
shleewhite, line 3: Question: why do these caption and description have quotes? it seems like most of the others dont.

website/docs/components/card/index.md
```diff
@@ -1,13 +1,18 @@
 ---
 title: Card
-description: A block container that provides styling for elevation, border, and background.
-caption: A block container that provides styling for elevation, border, and background.
+description: 'A block container that provides styling for elevation, border, and background.'
```
shleewhite, line 3: Same question about the quotes here.

website/docs/components/table/index.md
```diff
@@ -1,19 +1,24 @@
 ---
 title: Table
-description: Used to display organized, two-dimensional tabular data.
-caption: Used to display organized, two-dimensional tabular data.
+description: 'Used to display organized, two-dimensional tabular data.'
```
shleewhite, line 3: Another example where the quotes are randomly added.

website/docs/components/badge/index.md
shleewhite: Looking at the deployed site (https://hds-website-git-alex-ju-automate-component-ver-2d7be8-hashicorp.vercel.app/components/badge) seems like it doesn't randomly add the quotes to the page so I guess its fine

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -99,6 +100,75 @@ const updateComponentVersionHistory = (componentChangelogEntries, version) => {
   });
 };
 
+const updateComponentFrontMatter = (componentChangelogEntries, version) => {
+  Object.keys(componentChangelogEntries).forEach((componentName) => {
+    const indexPath = `${allComponentsPath[componentName]}/index.md`;
+
+    if (fs.existsSync(indexPath)) {
+      // Fetch the index markdown file
```
shleewhite, line 108: Really appreciate the comments!

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -99,6 +100,75 @@ const updateComponentVersionHistory = (componentChangelogEntries, version) => {
...
+        parsedFrontMatter.data
+      );
+
+      // Write the updated content back to the index markdown file
+      fs.writeFileSync(indexPath, updatedContent);
+    }
+  });
+};
+
+const cleanComponentFrontMatter = (components) => {
```
shleewhite, line 137: Suggestion: I think version needs to be added to the args here? Otherwise I don't see where it is defined.

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -99,6 +100,75 @@ const updateComponentVersionHistory = (componentChangelogEntries, version) => {
...
+
+      // Parse the index file content
+      const parsedFrontMatter = matter(indexContent);
+
+      // Update the front matter
+      if (!parsedFrontMatter.data.status) {
+        parsedFrontMatter.data.status = {};
+      }
+      if (
+        parsedFrontMatter.data.status.added !== version &&
```
shleewhite, line 119: Question: going forward, will we still need to manually add the `added` tag?

shleewhite (commented):
Looks good! Just left a couple questions and I think one thing to change.

website/docs/components/badge/index.md
alex-ju: According to the schema used by `js-yaml` strings over a certain length are moved to a block (hence the `>-` notation and the short ones get wrapped in single quotes to avoid parse errors.

website/lib/changelog/generate-component-changelog-entries.mjs
alex-ju: indeed. either add it manually, or change it from Updated to Added where needed. We can try to add some logic there to distinguish additions from updates, but it's not going to be a clean cut and wasn't sure it's worth the effort.

website/lib/changelog/generate-component-changelog-entries.mjs
alex-ju: right, yes, I believe I'm missing the second argument here – will update!

website/lib/changelog/generate-component-changelog-entries.mjs
shleewhite: Makes sense! We add stuff much less often.

website/docs/components/badge/index.md
shleewhite: Ok! I just noticed that there were places where the file was updated, but caption/description stayed the same (didn't add the quotes) so I thought it was interesting that it only got added 3 times.

ex: Accordion didnt have quotes added around the caption
Screenshot 2025-01-22 at 10 10 58 AM

website/lib/changelog/generate-component-changelog-entries.mjs
alex-ju: fixed in [15e50a9](https://github.com/hashicorp/design-system/pull/2656/commits/15e50a923b3135793dcd7fe047e9df93aa43b500) thanks for spotting this!

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -99,6 +100,75 @@ const updateComponentVersionHistory = (componentChangelogEntries, version) => {
...
+        parsedFrontMatter.data.status.updated !== version
+      ) {
+        parsedFrontMatter.data.status.updated = version;
```
dchyun, line 122: [Suggestion] This would defiantly make the implementation more complicated 😅, and is probably best raised in the doc, but for patch releases do we want to remove all "updated" badges from a previous minor release.

Just thinking of the situation where we do a minor release, the badges get added, and then we have some need to do a patch release quickly after and the badges get removed before consumers are really even aware they were there.

dchyun (commented):
Everything working great! Tested it with a few different scenarios and all worked out. Just one comment I left.

website/lib/changelog/generate-component-changelog-entries.mjs
alex-ju: It's good point. This is one of the many reasons I couldn't completely automate the action and it would need to be dispatched manually on minor releases.

alex-ju: That's a good point. Let me think about it, maybe we can prevent this from code rather than docs 🤔

alex-ju: Addressed this usecase in [a10e7cf](https://github.com/hashicorp/design-system/pull/2656/commits/a10e7cfcf7fae8746672062145e92f4946e1db75) by only cleaning added and updated badges on non-patch versions (minors and majors).

dchyun (approved):

website/lib/changelog/generate-component-changelog-entries.mjs
dchyun: Works great! Thanks for adding that.

PR #2658, opened 2025-01-21
Automate Vercel alias

alex-ju:
### :pushpin: Summary

TBA

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist


.github/workflows/create-vercel-alias.yml
```diff
@@ -0,0 +1,32 @@
...
+
+jobs:
+  create-vercel-alias:
+    name: 'Create Vercel Alias'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Extract Current Version
+        id: extract_version
+        run: echo "VERSION=$(jq -r .version packages/components/package.json | sed 's/\./-/g')" >> $GITHUB_ENV
```
alex-ju, line 18: Grabs the version from `package.json` and change format from dot-separator to dash-separator (e.g. `4.15.0` becomes `4-15-0`)

.github/workflows/create-vercel-alias.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Get Latest Vercel Deployment
+        id: get_vercel_deployment
+        run: |
+          DEPLOYMENT_ID=$(curl -s -H "Authorization: Bearer ${{ secrets.VERCEL_TOKEN }}" \
+          "https://api.vercel.com/v6/deployments?teamId=${{ secrets.VERCEL_TEAM_ID }}&projectId=${{ secrets.VERCEL_WEBSITE_PROJECT_ID }}" \
+          | jq -r --arg branch "${{ github.head_ref }}" '.deployments[] | select(.meta.githubCommitRef == $branch) | .uid' | head -n 1)
+          echo "DEPLOYMENT_ID=$DEPLOYMENT_ID" >> $GITHUB_ENV
```
alex-ju, line 25: Get the latest deployment id for the current branch (the branch on which the workflow is run) using Vercel's REST API.

.github/workflows/create-vercel-alias.yml
```diff
@@ -0,0 +1,32 @@
...
+      - name: Alias Vercel Deployment
+        id: alias_vercel_deployment
+        run: |
+          curl -X POST "https://api.vercel.com/v2/deployments/${{ env.DEPLOYMENT_ID }}/aliases?slug=hashicorp&teamId=${{ secrets.VERCEL_TEAM_ID }}" \
+          -H "Authorization: Bearer ${{ secrets.VERCEL_TOKEN }}" \
+          -H "Content-Type: application/json" \
+          -d "{\"alias\": \"hds-website-${{ env.VERSION }}.vercel.app\", \"teamId\": \"${{ secrets.VERCEL_TEAM_ID }}\", \"slug\": \"hashicorp\"}"
```
alex-ju, line 32: Create the alias for the latest deployment on the current branch. Note that the alias can only be created if the latest deployment is completed. If not completed, this step will fail and you'd need to re-run the action.

dchyun (approved):
Looks great to me! The test looks like it worked from the log. 

I like the idea of having it run after each update to `changeset-release/main` to keep things automated. My one thought is we may have to be careful about the order of how we merge in docs updates. 

There could be a situation where a feature gets merged to `main` and then the alias would get updated, but new docs are held off from being merged until closer to the release. Then when the docs branch gets merged the alias wouldn't automatically get updated because the docs branches don't cause updates to `changeset-release/main`.

alex-ju:
> Looks great to me! The test looks like it worked from the log.
> 
> I like the idea of having it run after each update to `changeset-release/main` to keep things automated. My one thought is we may have to be careful about the order of how we merge in docs updates.
> 
> There could be a situation where a feature gets merged to `main` and then the alias would get updated, but new docs are held off from being merged until closer to the release. Then when the docs branch gets merged the alias wouldn't automatically get updated because the docs branches don't cause updates to `changeset-release/main`.

Indeed, this is one of the scenarios that made be think a manual approach for now would be a safer start.

PR #2659, opened 2025-01-21
`AdvancedTable`: follow ups

shleewhite:
### :pushpin: Summary

- Remove utility classes from the Td component
- Prevent default when user presses arrow keys while focused on a cell
- Move repeated code from `td`, `th`, and `th-sort` to a new `hds-advanced-table-cell` modifier

### :link: External links

Jira ticket: [HDS-4381](https://hashicorp.atlassian.net/browse/HDS-4381)

***

### 👀 Component checklist

[HDS-4381]: https://hashicorp.atlassian.net/browse/HDS-4381?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/components/hds/advanced-table/td.hbs
```diff
@@ -8,12 +8,14 @@
...
-  {{will-destroy this.willDestroy}}
+  {{hds-advanced-table-cell
+    handleEnableFocusTrap=this.enableFocusTrap
+    shouldTrapFocus=this._shouldTrapFocus
+    setCellElement=this.setElement
+  }}
   {{! @glint-expect-error - https://github.com/josemarluedke/ember-focus-trap/issues/86 }}
   {{focus-trap
     isActive=this._shouldTrapFocus
-    shouldSelfFocus=true
```
shleewhite, line 16: removed `shouldSelfFocus` because it doesn't do anything when initialFocus is set (see within focusTrapOptions)

shleewhite:
I went back and forth on testing that the `is-pinned` class gets added correctly on scroll and ultimately got a solution that works most of the time locally, but it still fails in the headless browser in CI. Removing that part of the test so at least theres a test that the one class gets added.

dchyun (commented):
Great refactor to a custom modifier, left one small changeset comment.

.changeset/thin-masks-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`AdvancedTable`: Refactored keyboard navigation to a new modifier `hds-advanced-table-cell` for reusability.
```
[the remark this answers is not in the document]
shleewhite: Oh! I totally blanked. Thanks for catching!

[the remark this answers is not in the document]
shleewhite: Oh! I totally blanked. Thanks for catching!

dchyun (dismissed):

packages/components/src/components/hds/advanced-table/helpers.ts
```diff
@@ -3,10 +3,7 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
-import { focusable, tabbable } from 'tabbable';
-
-import type { HdsAdvancedTableTdSignature } from './td.ts';
-import type { HdsAdvancedTableThSignature } from './th.ts';
+import { tabbable } from 'tabbable';
 
 export const onFocusTrapDeactivate = (cell: HTMLDivElement) => {
```
alex-ju, line 8: question: should this helper be moved to `dom-management` too? it does operate on dom notes (changing the  `tabindex` attribute) as the other functions do

packages/components/src/modifiers/hds-advanced-table-cell.ts
```diff
@@ -0,0 +1,79 @@
...
+function cleanup(instance: HdsAdvancedTableCellModifier): void {
+  const { _observer } = instance;
+  if (_observer) {
+    _observer.disconnect();
+  }
+}
+
+export default class HdsAdvancedTableCellModifier extends Modifier<HdsAdvancedTableCellModifierSignature> {
+  // track shouldTrapFocus internally so the correct value is used when update tabbable children
+  @tracked private _shouldTrapFocus = false;
```
alex-ju, line 32: question: do we need the `@tracked` decorator here? as far as I know, modifiers do not have a rendering lifecycle as components do
```suggestion
  private _shouldTrapFocus = false;
```

packages/components/src/modifiers/hds-advanced-table-cell.ts
```diff
@@ -0,0 +1,79 @@
+import Modifier from 'ember-modifier';
+import type { ArgsFor, PositionalArgs } from 'ember-modifier';
+import { registerDestructor } from '@ember/destroyable';
+import { tracked } from '@glimmer/tracking';
```
alex-ju, line 4: I'm questioning the use of this – see comment below
```suggestion
```

alex-ju (approved):
I like the new modifier pattern. It's a shame we still need the `setElement` action, but I can't think of an alternative. I left two suggestions/questions otherwise looking great

packages/components/src/components/hds/advanced-table/helpers.ts
shleewhite: I went back and forth on moving this function haha. You're right, its totally fair to just move it all to the dom management file because it does modify the dom.

packages/components/src/modifiers/hds-advanced-table-cell.ts
shleewhite: You're correct! I updated it.

packages/components/src/components/hds/advanced-table/helpers.ts
shleewhite: Updated.

packages/components/src/modifiers/hds-advanced-table-cell/keyboard-navigation.ts
```diff
@@ -124,6 +50,7 @@ export const handleGridCellKeyPress = (
         enableFocusTrap();
       }
     } else if (event.key === 'ArrowRight' || event.key === 'ArrowLeft') {
+      event.preventDefault();
```
alex-ju, line 53: Leaving a comment here, for posterity, that we're introducing these to prevent the page from moving when reaching a boundary cell (I suspect, correct me if I'm wrong).

alex-ju (approved):

packages/components/src/modifiers/hds-advanced-table-cell/keyboard-navigation.ts
shleewhite: That is the case, @MelSumner requested that behavior.

dchyun (approved):

PR #2660, opened 2025-01-21
`Hds::CodeEditor` - Support Sentinel language

zamoore:
### :pushpin: Summary

If merged, this PR will add support for the Sentinel language in the `Hds::CodeEditor` component.

- Adds a [StreamParser](https://codemirror.net/docs/ref/#language.StreamParser) for Sentinel.
- Dynamically loads the parser when `sentinel` is specified as the `@language` argument.
- Adds showcase example for Sentinel syntax highlighting.

image

### :link: External links

Jira ticket: [HDS-4309](https://hashicorp.atlassian.net/browse/HDS-4309)

[HDS-4309]: https://hashicorp.atlassian.net/browse/HDS-4309?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/modifiers/hds-code-editor/languages/sentinel.ts
+++ b/packages/components/src/modifiers/hds-code-editor/languages/sentinel.ts
@@ -20,19 +20,25 @@ const wordOperators = wordRegexp(['and', 'or', 'not']);
 
 // keywords
 const sentinelKeywords = [
-  'import',
-  'main',
-  'rule',
-  'precondition',
-  'if',
-  'else',
-  'for',
-  'while',
+  'all',
+  'any',
+  'as',
   'break',
+  'case',
   'continue',
+  'default',
+  'else',
+  'empty',
+  'filter',
+  'for',
+  'func',
+  'if',
+  'import',
+  'map',
+  'param',
   'return',
-  'in',
-  'each',
+  'rule',
+  'when',
 ];
 
 // built-ins / constants
--- a/packages/components/src/modifiers/hds-code-editor/languages/sentinel.ts
+++ b/packages/components/src/modifiers/hds-code-editor/languages/sentinel.ts
@@ -42,7 +42,23 @@ const sentinelKeywords = [
 ];
 
 // built-ins / constants
-const sentinelBuiltins = ['true', 'false', 'null'];
+const sentinelBuiltins = [
+  'true',
+  'false',
+  'null',
+  'undefined',
+  'length',
+  'append',
+  'delete',
+  'keys',
+  'values',
+  'range',
+  'print',
+  'int',
+  'float',
+  'string',
+  'bool',
+];
 
 const keywords = wordRegexp(sentinelKeywords);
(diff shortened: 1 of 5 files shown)


packages/components/src/modifiers/hds-code-editor/languages/sentinel.ts
```diff
@@ -0,0 +1,182 @@
...
+  'precondition',
+  'if',
+  'else',
+  'for',
+  'while',
+  'break',
+  'continue',
+  'return',
+  'in',
+  'each',
```
alex-ju, line 35: we seem to be missing a few keywords here
```suggestion
  'all',
  'any',
  'as',
  'break',
  'case',
  'continue',
  'default',
  'else',
  'empty',
  'filter',
  'for',
  'func',
  'if',
  'import',
  'map',
  'param',
  'return',
  'rule',
  'when',
```

packages/components/src/modifiers/hds-code-editor/languages/sentinel.ts
```diff
@@ -0,0 +1,182 @@
...
+interface SentinelState {
+  tokenize: (stream: StringStream, state: SentinelState) => string | null;
+}
+
+function wordRegexp(words: string[]): RegExp {
+  return new RegExp('^((' + words.join(')|(') + '))\\b');
+}
+
+// logical operators
+const wordOperators = wordRegexp(['and', 'or', 'not']);
```
alex-ju, line 19:
```suggestion
const wordOperators = wordRegexp([
  'and',
  'contains',
  'else',
  'in',
  'is',
  'matches',
  'not',
  'or',
  'xor',
]);
```

packages/components/src/modifiers/hds-code-editor/languages/sentinel.ts
```diff
@@ -0,0 +1,182 @@
...
+  'while',
+  'break',
+  'continue',
+  'return',
+  'in',
+  'each',
+];
+
+// built-ins / constants
+const sentinelBuiltins = ['true', 'false', 'null'];
```
alex-ju, line 39: adding built-in functions and type convertors
```suggestion
const sentinelBuiltins = [
  'true',
  'false',
  'null',
  'undefined',
  'length',
  'append',
  'delete',
  'keys',
  'values',
  'range',
  'print',
  'int',
  'float',
  'string',
  'bool',
];
```

alex-ju (commented):
Not proficient in sentinel, but the highlight seem a bit off, so I checked [the docs for the syntax](https://developer.hashicorp.com/sentinel/docs/language/spec) and suggested some adjustments.

Highlighted the impact on the current example below, but tested with other snippets as well.

#### Before
Screenshot 2025-01-22 at 12 24 21

#### After 
Screenshot 2025-01-22 at 12 28 07

PR #2675, opened 2025-01-29
fix start:prember command

Dhaulagiri:
This wasn't always working due to the build not happening first


alex-ju (approved):
Thanks for that!

PR #2683, opened 2025-02-04
Adding missing guidance

heatherlarsen:
### :pushpin: Summary

If merged, this PR will add the missing Stepper details from the previous design stickersheet.

**👉🏻 Preview:** https://hds-website-git-hl-stepper-guidelines-hashicorp.vercel.app/components/stepper

**Before:**
Screenshot 2025-02-03 at 9 19 28 PM

**After:**
screenshot-hds-website-git-hl-stepper-guidelines-hashicorp_vercel_app-2025_02_03-21_20_13

### :link: External links

Jira ticket: [HDS-4379](https://hashicorp.atlassian.net/browse/HDS-4379)

***

### 👀 Component checklist

[HDS-4379]: https://hashicorp.atlassian.net/browse/HDS-4379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shleewhite (approved):

website/docs/components/stepper/partials/guidelines/guidelines.md
```diff
@@ -9,6 +9,34 @@
...
+
+Non-interactive steps are a part of a list of steps that can be completed sequentially, out of order, or asynchronously. They cannot be interacted with.
+
+![](/assets/components/stepper/step-noninteractive.png)
+
+## Tasks
+
+### Interactive tasks
+
+Interactive tasks are part of a sequential flow and can be interacted with as a navigational device.
```
heatherlarsen, line 30: @jorytindall I pulled this from the old stickersheet, but I'm unsure if it's correct. Do tasks need to be in a sequential flow or is that just for steps? Do you have any ideas on what should be included here if they aren't for sequential flows?

jorytindall: I'm trying to remember the reasoning behind this guidance, now that I'm re-reading this it seems a little to restrictive and pointed, especially given that this more or less only a decorative element. The point about the interactive **steps** makes sense, but **tasks** less so. Maybe it would be best to share the content about sequential/out of order/async between both task types, WDYT?

website/docs/components/stepper/partials/guidelines/guidelines.md
```diff
@@ -9,6 +9,34 @@
...
+
+![](/assets/components/stepper/task-interactive.png)
+
+### Non-interactive tasks
+
+Non-interactive tasks are part of a list of tasks which can be completed sequentially, out of order, or asynchronously. They cannot be interacted with.
```
LilithJames-HDS, line 36:
```suggestion
Non-interactive tasks are part of a list of tasks which can be completed in any order. They are not interactive but serve as a visual list item indicator. 
```

LilithJames-HDS (approved):
One suggestion to make the non-interactive stepper language clearer.

website/docs/components/stepper/partials/guidelines/guidelines.md
heatherlarsen: Love this! Thank you

website/docs/components/stepper/partials/guidelines/guidelines.md
heatherlarsen: Yea, I can use the language @LilithJames-HDS suggested. Thanks!

jorytindall (approved):

alex-ju (approved):
In this new guidance, we speak about steps and tasks, but show only the indicators (presuming the step is made of an indicator and some sort of content). I'm ok to get it in as it is for now, but it must be reviewed when the `Stepper::Step` and `Stepper::Task` are defined.

The interactive part is a separate can of worms, that will hopefully be addressed with the upcoming work – the indicator is not technically an interactive element on its own, only when wrapped by an interactive element (potentially the `Stepper::Step` and `Stepper::Task`)

heatherlarsen:
Understood. Thanks @alex-ju! We'll definitely address these in the upcoming Stepper project and I imagine this documentation will change quite a bit.

PR #2685, opened 2025-02-04
`Clipboard` & `Copy Button` - Fix handling of empty string and zero values (HDS-4447)

KristinLBradley:
### :pushpin: Summary

If merged, this PR adds a fix to allow empty strings and zero number values passed into the Copy Button to copy the values.

* Update clipboard modifier used by Copy Button to handle passed in empty string and zero number values
* Add Showcase examples
* Add integration tests

### :link: External links

* Jira ticket: [HDS-4447](https://hashicorp.atlassian.net/browse/HDS-4447)

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/modifiers/hds-clipboard.ts
+++ b/packages/components/src/modifiers/hds-clipboard.ts
@@ -37,7 +37,7 @@ export const getTextToCopy = (text: TextToCopy): string => {
       textToCopy = text.toString();
     } else {
       assert(
-        `\`hds-clipboard\` modifier - \`text\` argument must be a string - provided: ${typeof text}`
+        `\`hds-clipboard\` modifier - \`text\` argument must be a string or number - provided: ${typeof text}`
       );
     }
   }
@@ -109,7 +109,7 @@ export const writeTextToClipboard = async (
 ): Promise<boolean> => {
   // finally copy the text to the clipboard using the Clipboard API
   // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard_API
-  if (textToCopy) {
+  if (textToCopy || textToCopy === '') {
     try {
       // notice: the "clipboard-write" permission is granted automatically to pages when they are in the active tab
       // https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/write
@@ -151,7 +151,11 @@ export const copyToClipboard = async (
 ): Promise<boolean> => {
   let textToCopy: string = '';
 
-  if (text) {
+  if (text === '') {
+    textToCopy = '';
+  } else if (text === 0) {
+    textToCopy = '0';
+  } else if (text) {
     textToCopy = getTextToCopy(text);
   } else if (target) {
     const targetElement = getTargetElement(target);
--- a/showcase/app/templates/components/copy/button.hbs
+++ b/showcase/app/templates/components/copy/button.hbs
@@ -142,6 +142,17 @@
     </SF.Item>
   </Shw::Flex>
 
+  <Shw::Text::H4>Edge cases</Shw::Text::H4>
+
+  <Shw::Flex as |SF|>
+    <SF.Item>
+      <Hds::Copy::Button @text="Copy an empty string" @textToCopy="" />
+    </SF.Item>
+    <SF.Item>
+      <Hds::Copy::Button @text="Copy the number '0'" @textToCopy={{0}} />
+    </SF.Item>
+  </Shw::Flex>
+
   <Shw::Divider @level={{2}} />
 
   <Shw::Text::H3>With <code>target</code> element</Shw::Text::H3>
--- a/showcase/tests/integration/components/hds/copy/button/index-test.js
+++ b/showcase/tests/integration/components/hds/copy/button/index-test.js
@@ -46,6 +46,22 @@ module('Integration | Component | hds/copy/button/index', function (hooks) {
     assert.true(this.success);
   });
 
+  test('it should copy an empty string value passed in', async function (assert) {
+    await render(
+      hbs`<Hds::Copy::Button id="test-copy-button" @text="Copy empty string"  @textToCopy="" @onSuccess={{this.onSuccess}} @onError={{this.onError}} />`
+    );
+    await click('button#test-copy-button');
+    assert.true(this.success);
+  });
+
+  test('it should copy a zero number value passed in', async function (assert) {
+    await render(
+      hbs`<Hds::Copy::Button id="test-copy-button" @text="Copy zero number value"  @textToCopy={{0}} @onSuccess={{this.onSuccess}} @onError={{this.onError}} />`
+    );
+    await click('button#test-copy-button');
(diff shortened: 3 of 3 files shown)


.changeset/early-knives-fly.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Copy Button` - Fixed issue preventing copying of empty string and zero number values.
```
didoo, line 5: [question] we're changing the `hds-clipboard` modifier, which is used also in the `Copy::Snippet` component; should this mentioned too in the changeset?

showcase/tests/integration/components/hds/copy/button/index-test.js
didoo, line 1: [blocker] I think we should also update/add tests in `showcase/tests/integration/modifiers/hds-clipboard-test.js`

showcase/tests/integration/components/hds/copy/button/index-test.js
```diff
@@ -46,6 +46,22 @@ module('Integration | Component | hds/copy/button/index', function (hooks) {
     assert.true(this.success);
   });
 
+  test('it should copy an empty string value passed in', async function (assert) {
```
didoo, line 49: [thought] should these just live directly in the tests for the underlying modifier? in this way both `Copy::Button` and `Copy::Snippet` are covered

packages/components/src/modifiers/hds-clipboard.ts
```diff
@@ -151,7 +151,11 @@ export const copyToClipboard = async (
 ): Promise<boolean> => {
   let textToCopy: string = '';
 
-  if (text) {
+  if (text === '') {
+    textToCopy = '';
+  } else if (text === 0) {
+    textToCopy = '0';
+  } else if (text) {
```
didoo, line 158: [suggestion] what if we just check that the value is defined here? (potentially we could do the same also for `else if (target !== undefined)` below. 

My reasoning is: it seems more "correct" in relation to the logic in this function (the actual check about the type is done anyway inside the `getTextToCopy()` function, here is just a logical switch between the two arguments `@text` or `@target` to decide which function to invoke

```suggestion
  if (text !== undefined) {
```

didoo: In this way you can then remove the `if (text)` check inside the `getTextToCopy()` function and the `if (textToCopy)` check inside the `writeTextToClipboard()` function

As point in favour of this: the updated logic would reflect the typing definition of the functions, that are not expecting undefined values

showcase/app/templates/components/copy/button.hbs
```diff
@@ -142,6 +142,17 @@
     </SF.Item>
   </Shw::Flex>
 
+  <Shw::Text::H4>Edge cases</Shw::Text::H4>
```
didoo, line 145: [suggestion] At this point, they're not "technically" edge, more of special values that we have to handle correctly

```suggestion
  <Shw::Text::H4>Special cases</Shw::Text::H4>
```

didoo (commented):
@KristinLBradley left a few comments; let me know if you want to discuss them in a call if it's not clear or doesn't makes sense to you what I'm suggesting

.changeset/early-knives-fly.md
KristinLBradley: Edited

showcase/tests/integration/modifiers/hds-clipboard-test.js
```diff
@@ -278,6 +290,18 @@ module('Integration | Modifier | hds-clipboard', function (hooks) {
     assert.true(this.success);
   });
 
+  test('it should copy a zero number value provided as a `@text` argument', async function (assert) {
+    await render(
+      hbs`<button id="test-button" {{hds-clipboard
+        text=0
```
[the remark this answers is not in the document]
KristinLBradley: I followed how numbers were defined in other tests and I thought nested {{}} wouldn't work anyway. (I did initially wonder about it but then realized it would create {{}} so didn't try it)

packages/components/src/modifiers/hds-clipboard.ts
KristinLBradley: I made the changes & pushed up. I also had to remove a unit test for passing an undefined value to `writeTextToClipboard`.

showcase/tests/integration/modifiers/hds-clipboard-test.js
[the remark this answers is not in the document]
KristinLBradley: Yeah it fails when you nest the curly braces

showcase/app/templates/components/copy/button.hbs
```diff
@@ -142,6 +142,17 @@
     </SF.Item>
   </Shw::Flex>
 
+  <Shw::Text::H4>Special cases</Shw::Text::H4>
```
didoo, line 145: [suggestion] since both the `Copy::Button` and the `Copy::Snippet` have been technically updated, we should add similar use cases to the `Copy::Snippet` showcase page too

didoo (commented):
left a comment around missing showcase use cases for `Copy::Snippet`. Apart from that, everything is fine for me

showcase/app/templates/components/copy/button.hbs
KristinLBradley: Just added.

KristinLBradley: https://hds-showcase-git-hds-4447-copybutton-empty-string-hashicorp.vercel.app/components/copy/snippet#demo

packages/components/src/modifiers/hds-clipboard.ts
didoo: @KristinLBradley check the "Copy a Bigint number" in the `Copy::Snippet`

[You may also consider adding the two missing cases, empty string and zero, directly to the first "Content" section]

KristinLBradley: Sorry, I missed that. I removed that section and added to the Content section examples instead.

showcase/app/templates/components/copy/button.hbs
KristinLBradley: Updated again

KristinLBradley: https://hds-showcase-git-hds-4447-copybutton-empty-string-hashicorp.vercel.app/components/copy/snippet#content

packages/components/src/modifiers/hds-clipboard.ts
KristinLBradley: https://hds-showcase-git-hds-4447-copybutton-empty-string-hashicorp.vercel.app/components/copy/snippet#content

didoo (commented):
LGTM 👍

Nice job! 👏

didoo (approved):

PR #2687, opened 2025-02-05
Website: Scroll to top goes to top of page

dchyun:
### :pushpin: Summary

This PR fixes an issue in the website back-to-top button, where upon clicking the button the page was scrolled only until the page title was in view, not the entire way back to the top of the page. 

[Preview page](https://hds-website-git-dchyun-website-scroll-to-top-i-64561a-hashicorp.vercel.app/components/accordion)

Screenshots show appearance after scrolling down page and pressing back-to-top button.

**Before**
Screenshot 2025-02-05 at 2 34 53 PM

**After**
Screenshot 2025-02-05 at 2 35 27 PM

### :link: External links

Jira ticket: [HDS-4472](https://hashicorp.atlassian.net/browse/HDS-4472)

***

[HDS-4472]: https://hashicorp.atlassian.net/browse/HDS-4472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/styles/pages/application/stage.scss
+++ b/website/app/styles/pages/application/stage.scss
@@ -19,6 +19,7 @@
   grid-template-rows: auto auto 1fr;
   grid-template-columns: 1fr;
   width: 100%; // needed (otherwise it collapses, because is in a grid)
+  scroll-margin-top: var(--doc-page-stage-gutter-large);
 
   @include breakpoint.with-fixed-sidebar() {
     padding-left: var(--doc-page-sidebar-width); // we need to compensate for the fixed position of the sidebar
--- a/website/app/styles/pages/application/stage.scss
+++ b/website/app/styles/pages/application/stage.scss
@@ -19,7 +19,7 @@
   grid-template-rows: auto auto 1fr;
   grid-template-columns: 1fr;
   width: 100%; // needed (otherwise it collapses, because is in a grid)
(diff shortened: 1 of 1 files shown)


shleewhite (approved):
Nice! Tested on a variety of pages and it seems fixed!

website/app/styles/pages/application/stage.scss
```diff
@@ -19,6 +19,7 @@
   grid-template-rows: auto auto 1fr;
   grid-template-columns: 1fr;
   width: 100%; // needed (otherwise it collapses, because is in a grid)
+  scroll-margin-top: var(--doc-page-stage-gutter-large);
```
didoo, line 22: [suggestion] my understanding is that the scroll is done on the `main` container, but because the page header has a sticky position, the page doesn't scroll to the _visual_ top. which means, for me, that this should be instead `--doc-page-header-height` (which is the same value in `px`, but semantically more correct, and resilient/future proof)

```suggestion
  scroll-margin-top: var(--doc-page-header-height);
```

dchyun: Thanks for the suggestion! Great catch. Updated now.

didoo (approved):
👍

PR #2688, opened 2025-02-06
`AdvancedTable`: add expand all button

shleewhite:
### :pushpin: Summary

If merged, this PR would add an Expand All button to AdvancedTables with nested rows. To add the Expand All button, consumers have to indicate which column is expandable in the columns arg by adding `isExpandable: true`.

**Changes of note:**
1. Updated the `ExpandableTrGroup` so now the rows are always in the DOM. This is to make it so all the expand buttons are always in the DOM, which makes it easier to toggle all of their states at the same time when the expand all button is clicked.

2. Added a modifier to the AdvancedTable for when `Th` is inserted in the header. The modifier checks if the cell has an expand all button and if there is, it saves a reference to the expand all button element. Initially I had it more similar to how the select all button works, but it felt excessive because for the select all checkbox we pass a didInsert and willDestroy function from the index file -> Tr -> ThSelectable.

3. There's also a couple clean up things for the select all stuff in AdvancedTable: I removed the custom event that I also removed from Table and changed the didInsert/willDestroy in ThSelectable to a modifier

4. Also one clean up thing for the advanced table cell modifier, I forgot to remove the event listener on the cell. so I restructured the modifier a bit to actually clean that up.

### :link: External links

Jira ticket: [HDS-4382](https://hashicorp.atlassian.net/browse/HDS-4382)
Figma file: [if it applies]

***

### 👀 Component checklist

[HDS-4382]: https://hashicorp.atlassian.net/browse/HDS-4382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -425,9 +444,71 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
+  }
+
+  @action
+  willDestroyExpandButton(): void {
+    this.setExpandAllState();
+  }
+
+  @action
+  setExpandAllState(): void {
+    if (this._expandAllButton && this._element) {
```
alex-ju, line 462: suggestion: not a future-proof solution, but to make the logic here work, you would need to delay it until the next cycle (you can wrap the content of this if statement in a `next(() => { /*if content hoes here*/ });`). It will require `import { next } from '@ember/runloop';` and likely a `// eslint-disable-next-line ember/no-runloop`, but it should resolve the state issue.

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -425,9 +444,71 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
+      console.log('parentRows', expandButtons);
+      console.log(
+        'expandedRows',
+        expandButtons.filter(
+          (button) => button.getAttribute('aria-expanded') === 'true'
+        )
+      );
+
+      if (parentRowsCount === expandedRowsCount) expandAllState = true;
+      if (expandedRowsCount === 0) expandAllState = false;
```
alex-ju, line 485: nit: should this be an `else if`? doesn't make a different in the grand scheme of things, but it could make sense since the conditionals are related

packages/components/src/components/hds/advanced-table/index.ts
shleewhite: Cool, it does fix the issue!

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -316,6 +316,10 @@ $hds-advanced-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px differenc
 
   // nested rows
 
+  .hds-advanced-table__tr--hidden {
+    display: none;
```
dchyun, line 320: [Issue] Caught a keyboard nav issue related to this change. In the nested example, using the down arrow after focusing on the table, the focus is stopped at "Policy Set 1" and prevented from navigating further.

After the focus is on "Policy Set 1" and pressing the down arrow, the nested step within "Policy Set 1" is set to `tabindex="0"`, but this class is on the parent of that nested step. The `display: none` is hiding the focused step from the DOM.  

This seems to have to do with the change made to keep the nested steps in the DOM at all times vs adding and removing them when activated. Maybe there's a way in the `findNewRow` function in the keyboard nav modifier to ignore hidden rows. 

Screenshot 2025-02-14 at 4 00 56 PM
Screenshot 2025-02-14 at 4 01 11 PM

shleewhite:
> I'm not sure where to place this, but I noticed on the showcase that the expand all starts like this: image
> 
> But upon clicking it, it changes to the standard chevron image
> 
> And then clicking again changes again: image
> 
> Shouldn't the expand/collapse all remain consistent throughout?

@andgen404 can weigh in here, my understanding was that the button should reflect the current state of the table similar to the select all checkbox. the first icon is a "mixed" state (some rows are open, some are closed), then when you click expand all it expands all and the icon turns into the expanded variant, then when you collapse all it looks collapsd.

packages/components/src/styles/components/advanced-table.scss
shleewhite: Ah! thanks for catching this! It should be fixed now, just had to update a couple selectors.

dchyun: Working great! Thanks for fixing.

andgen404 (dismissed):
I checked interaction/design, and it looks good to me! 🎉

jorytindall:
Screenshot 2025-02-27 at 10 17 28 AM

I'm not sure if it happened as a part of this work, but it looks like there's a small visual bug in the table with nesting. The bottom left and right border radii are clipped. It's not a problem when all of the rows are expanded, but is when they are collapsed.

showcase/app/templates/components/advanced-table.hbs
dchyun, line 1: [Issue] Finding an odd issue in the showcase in the [2nd nested row example](https://hds-showcase-git-hds-4382-advanced-table-expand-all-hashicorp.vercel.app/components/advanced-table#with-nested-rows-and-custom-children-key) with the 2nd level nested row not being operable with keyboard.

Steps to reproduce (using keyboard)
1.  Open "Policy set 2"
2. Open "test-advisory-pass.sentinel"
3. Move to expand button
4. Click expand button to expand all rows
5. Click expand button again to close all rows
6. Open "Policy set 2"
7. Open "test-advisory-pass.sentinel"

On the last step, clicking enter will not move the focus to the expand button inside the row, but the `tabindex` will still get updated to `-1` for the cell. Leaving the table using `tab` and going back to it using `shift + tab` will also show the table can no longer be reached using focus. I think the `tabindex="0"` is getting lost somehow and everything in the table is `-1`. 
Screenshot 2025-03-05 at 11 49 41 AM

packages/components/src/modifiers/hds-advanced-table-cell/dom-management.ts
```diff
@@ -78,3 +78,31 @@ export const didInsertGridCell = (
     }
   }
 };
+
+export const updateLastRowClass = (element: HTMLDivElement) => {
+  console.log('in update last row class');
```
dchyun, line 83:
```suggestion
```

packages/components/src/modifiers/hds-advanced-table-cell/dom-management.ts
```diff
@@ -78,3 +78,31 @@ export const didInsertGridCell = (
...
+  const previousLastRow = element.querySelectorAll(
+    '.hds-advanced-table__tr--last-row'
+  );
+
+  console.log(previousLastRow);
```
dchyun, line 88:
```suggestion
```

.changeset/swift-elephants-play.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`AdvancedTable`: added an expand all button to `AdvancedTable`s with nested rows. Changed the structure of the component so now nested rows are always in the DOM, even when they are not visible.
+
+To add the expand all button, add `isExpandable: true` to the desired column in the `@columns` argument.
```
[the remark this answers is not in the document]
dchyun: @alex-ju will the parser still pick up these changes correctly having them in this format?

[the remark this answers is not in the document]
alex-ju: > @alex-ju will the parser still pick up these changes correctly having them in this format?

unfortunately not, we only support `-` separator and one-line description
there's room for enhancements, but to benefit from automation we'd have to stick to one pattern

andgen404 (commented):
Screenshot 2025-03-05 at 6 04 29 PM

It looks like there's a visual bug in the ColSpan example, the radius on the bottom left and the cells to the right.

packages/components/src/components/hds/advanced-table/expandable-tr-group.hbs
```diff
@@ -8,13 +8,16 @@
...
     parentId=@parentId
     rowIndex=this.rowIndex
+    didInsertExpandButton=@didInsertExpandButton
+    willDestroyExpandButton=@willDestroyExpandButton
+    shouldDisplayChildRows=@shouldDisplayChildRows
+    onClickToggle=this.onClickToggle
   )
 }}
-{{#if (and this.hasChildren this._isExpanded)}}
+{{#if (and this.hasChildren)}}
```
zamoore, line 20:
```suggestion
{{#if this.hasChildren}}
```

packages/components/src/components/hds/advanced-table/expandable-tr-group.ts
```diff
@@ -26,17 +30,20 @@ export interface HdsAdvancedTableExpandableTrGroupSignature {
         id?: string;
         parentId?: string;
         depth: number;
-        onClickToggle?: () => void;
-        isExpanded?: boolean;
+        onClickToggle?: (newValue?: boolean | 'mixed') => void;
```
zamoore, line 33: nit: could define a type for this since it shows up in a few places

shleewhite:
> It looks like there's a visual bug in the ColSpan example, the radius on the bottom left and the cells to the right.

@andgen404 I ended up removing this example entirely - its not possible to get the border radius correct in this use case and it is not something we have ever seen consumers need. Generally, we would advise not using rowspan this way.

showcase/app/templates/components/advanced-table.hbs
shleewhite: Ah I figured out the issue, the function that initializes all the focusable content within advanced table cells didnt count elements that are visually hidden as tabbable. Updated to look for tabbable elements that are also hidden now.

dchyun: Looks great! Thanks for fixing it!

.changeset/swift-elephants-play.md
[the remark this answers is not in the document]
shleewhite: Good point. I updated the changset to follow the pattern that the automation expects.

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -297,16 +307,36 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
+        this._intersectionObserver.disconnect();
+      }
+
+      if (this._mutationObserver) {
+        this._mutationObserver.disconnect();
       }
     };
   });
 
+  private _setUpExpandAllButton = modifier((element: HTMLDivElement) => {
```
zamoore, line 326: Is there any way this modifier can be put on the "expand all" button itself? If you need to set the value higher up, you can pass a function down to where the "expand all" button is rendered and call it in the modifier. It's more "embery" to do it this way rather than query the DOM.

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -425,9 +454,68 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
+  setExpandAllState(): void {
+    if (this._expandAllButton && this._element) {
+      const expandButtons = Array.from(
+        this._element.querySelectorAll(
+          '.hds-advanced-table__tbody .hds-advanced-table__th-button--expand'
+        )
       );
+
+      // eslint-disable-next-line ember/no-runloop
+      next(() => {
```
zamoore, line 480: What was going on without doing this in the next run loop?

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -425,9 +454,68 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
+        updateLastRowClass(this._element);
+      });
+    }
+  }
+
+  @action
+  onExpandAllClick(): void {
+    if (this._expandAllButton && this._element) {
+      const expandButtons = Array.from(
+        this._element.querySelectorAll(
```
zamoore, line 502: Another way to maintain a collection of these elements might be to insert/remove them from a tracked array with a modifier on the expand button.

I generally try to use vanilla DOM code as little as possible in the framework.

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -425,9 +454,68 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
+          '.hds-advanced-table__tbody .hds-advanced-table__th-button--expand'
+        )
+      );
+
+      const newState =
+        this._expandAllButton.getAttribute('aria-expanded') === 'true'
```
zamoore, line 508: The DOM shouldn't be the source of truth of the expansion state if we can avoid it. Is there a way we can track the expanded state and use that to determine the newState?

packages/components/src/components/hds/advanced-table/index.ts
shleewhite: The expand all button's state was always 1 step behind what it should actually be.

packages/components/src/components/hds/advanced-table/index.ts
shleewhite: Sure! Can update this.

packages/components/src/components/hds/advanced-table/index.ts
shleewhite: ahh true, I already have a `this._expandAllButtonState`. Not sure why im not already using it.

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -10,6 +10,7 @@ import { tracked } from '@glimmer/tracking';
 import type { ComponentLike } from '@glint/template';
 import { guidFor } from '@ember/object/internals';
 import { modifier } from 'ember-modifier';
+// import { next } from '@ember/runloop';
```
dchyun, line 13:
```suggestion
```

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -297,12 +308,18 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
+
     return () => {
-      if (this._observer) {
-        this._observer.disconnect();
+      if (this._intersectionObserver) {
+        this._intersectionObserver.disconnect();
+      }
+
+      if (this._mutationObserver) {
+        this._mutationObserver.disconnect();
```
dchyun, line 322: [Question] Is the `_mutationObserver` ever getting set up? I see it being disconnected here, but don't see it being connected. Can we remove it if it's not being used?

packages/components/src/components/hds/advanced-table/th-button-expand.ts
```diff
@@ -6,33 +6,45 @@
...
+    onToggle?: (newValue?: HdsAdvancedTableExpandState) => void;
+    isExpandAll?: boolean;
   };
   Element: HTMLButtonElement;
 }
 
 export default class HdsAdvancedTableThButtonExpand extends Component<HdsAdvancedTableThButtonExpandSignature> {
   // Generates a unique ID for the (hidden) "label prefix" <span> element
   private _prefixLabelId = 'prefix-' + guidFor(this);
+  _observer: MutationObserver | undefined = undefined;
```
dchyun, line 29: [Question] Is this variable being used?

packages/components/src/components/hds/advanced-table/th-button-expand.ts
```diff
@@ -6,33 +6,45 @@
...
+  HdsAdvancedTableThExpandIconValues,
+  type HdsAdvancedTableExpandState,
+} from './types.ts';
 import type { HdsAdvancedTableThSortExpandIcons } from './types.ts';
 export interface HdsAdvancedTableThButtonExpandSignature {
   Args: {
     labelId?: string;
-    isExpanded?: boolean;
-    onToggle?: () => void;
+    isExpanded?: boolean | 'mixed';
```
dchyun, line 19:
```suggestion
    isExpanded?: HdsAdvancedTableExpandState;
```

packages/components/src/components/hds/advanced-table/th-button-expand.ts
```diff
@@ -6,33 +6,45 @@
...
   Element: HTMLButtonElement;
 }
 
 export default class HdsAdvancedTableThButtonExpand extends Component<HdsAdvancedTableThButtonExpandSignature> {
   // Generates a unique ID for the (hidden) "label prefix" <span> element
   private _prefixLabelId = 'prefix-' + guidFor(this);
+  _observer: MutationObserver | undefined = undefined;
 
-  get isExpanded(): boolean {
+  get isExpanded(): boolean | 'mixed' {
```
dchyun, line 31:
```suggestion
  get isExpanded(): HdsAdvancedTableExpandState {
```

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -217,22 +217,33 @@ $hds-advanced-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px differenc
...
-      .hds-advanced-table__td:last-child {
-        border-bottom-right-radius: $hds-advanced-table-inner-border-radius;
-      }
+    // a <th> will never be last child, only first child
+    .hds-advanced-table__td:last-child {
+      border-bottom-right-radius: $hds-advanced-table-inner-border-radius;
     }
   }
 
+  // .hds-advanced-table__tr:has(+ .hds-advanced-table__tr--hidden) {
```
dchyun, line 235: [Question] Can this be removed?

dchyun (commented):
Looking good! Just a couple small things.

packages/components/src/components/hds/advanced-table/index.ts
shleewhite: Note: talked to Zack about this more offline, going to leave next in here for now. We have a ticket for us to refactor the component which may make it so we dont need it anymore.

packages/components/src/styles/components/advanced-table.scss
shleewhite: Ah, yes - can definitely remove this.

packages/components/src/components/hds/advanced-table/th-button-expand.ts
shleewhite: ah! another good catch - I tried a bunch of different approaches to get this to work. Can remove this.

packages/components/src/components/hds/advanced-table/index.ts
shleewhite: Ah yes, definitely can remove.

website/docs/components/table/advanced-table/partials/code/component-api.md
```diff
@@ -61,6 +61,9 @@ The Advanced Table component itself is where most of the options will be applied
       <C.Property @name="tooltip" @type="string">
         Text string which will appear in [`Tooltip`](/components/tooltip). May contain basic HTML tags for formatting text such as `strong` and `em` tags. Not intended for multi-paragraph text or other more complex content. May not contain interactive content such as links or buttons. The `placement` and `offset` are automatically set and can’t be overwritten.
       </C.Property>
+      <C.Property @name="isExpandable" @type="boolean">
```
dchyun, line 64:
```suggestion
      <C.Property @name="isExpandable" @type="boolean" @default="false">
```
[Nit] Add default value

dchyun (approved):
Looks great to me! Just one docs comment.

PR #2690, opened 2025-02-06
`Table`: remove last of update aria-label code

shleewhite:
### :pushpin: Summary

I noticed in more testing that there was still a bit of code remaining for the changing aria-label stuff in th-selectable. It currently doesn't do anything because the state of the checkbox is not included in the checkbox label anymore. Removing it to reduce confusion.

***

### 👀 Component checklist

### DIFF
--- a/.changeset/three-walls-lay.md
+++ b/.changeset/three-walls-lay.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Table`: Removed unused updateAriaLabel function and event listener
--- a/.changeset/three-walls-lay.md
+++ b/.changeset/three-walls-lay.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-`Table`: Removed unused updateAriaLabel function and event listener
+`Table` – Removed unused `updateAriaLabel` function and event listener
--- a/packages/components/src/components/hds/table/th-selectable.ts
+++ b/packages/components/src/components/hds/table/th-selectable.ts
@@ -82,7 +82,7 @@ export default class HdsTableThSelectable extends Component<HdsTableThSelectable
   }
 
   @action
-  willDestroyNode(checkbox: HdsFormCheckboxBaseSignature['Element']): void {
+  willDestroyNode(): void {
     super.willDestroy();
     const { willDestroy } = this.args;
     if (typeof willDestroy === 'function') {
--- a/packages/components/src/components/hds/table/th-selectable.ts
+++ b/packages/components/src/components/hds/table/th-selectable.ts
@@ -78,14 +78,6 @@ export default class HdsTableThSelectable extends Component<HdsTableThSelectable
     const { didInsert } = this.args;
     if (typeof didInsert === 'function') {
       didInsert(checkbox, this.args.selectionKey);
-      // we need to use a custom event listener here because changing the `checked` value via JS
-      // (and this happens with the "select all") doesn't trigger the `change` event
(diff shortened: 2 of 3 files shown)


packages/components/src/components/hds/table/th-selectable.ts
```diff
@@ -115,10 +100,4 @@ export default class HdsTableThSelectable extends Component<HdsTableThSelectable
       onSelectionChange(target, this.args.selectionKey);
     }
   }
-
-  updateAriaLabel(event: Event): void {
-    // Assert event.target as HTMLInputElement to access the 'checked' property
-    const target = event.target as HdsFormCheckboxBaseSignature['Element'];
-    this.isSelected = target.checked;
```
shleewhite, line 122: Its safe to remove this line because the state is already updated in the `onSelectionChange` function.

alex-ju: Isn't this also the case for the AdvancedTable?

alex-ju (dismissed):
Makes sense to me. Wondering if things are different in any way for the AdvancedTable or we should remove `updateAriaLabel` from there as well. 

I understand the changelog rationale, and if we haven't released the previous change I would agree it's not required. However, if anything goes wrong because of this change (hypothetically) it will be more difficult for us and our consumers to understand what caused the change, so I'd be tempted to add a brief note in the changelog every time we touch component code, even if it's along the lines of 'Removed unused code'.

I also noticed that we don't mark isSelected as private here (and we do in the AdvancedTable), so I suspect `@tracked isSelected: boolean;` should likely become `@tracked private _isSelected: boolean`, but can be tackled separately.

packages/components/src/components/hds/table/th-selectable.ts
shleewhite: Yes, I updated it in this pr: https://github.com/hashicorp/design-system/pull/2688

alex-ju: Nice! Thank you!

shleewhite:
> Makes sense to me. Wondering if things are different in any way for the AdvancedTable or we should remove `updateAriaLabel` from there as well.
> 
> I understand the changelog rationale, and if we haven't released the previous change I would agree it's not required. However, if anything goes wrong because of this change (hypothetically) it will be more difficult for us and our consumers to understand what caused the change, so I'd be tempted to add a brief note in the changelog every time we touch component code, even if it's along the lines of 'Removed unused code'.
> 
> I also noticed that we don't mark isSelected as private here (and we do in the AdvancedTable), so I suspect `@tracked isSelected: boolean;` should likely become `@tracked private _isSelected: boolean`, but can be tackled separately.

That makes sense to me about the changeset- I added one.

.changeset/three-walls-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Table`: Removed unused updateAriaLabel function and event listener
```
alex-ju, line 5: sorry, we [settled on using dash separators](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Component-changes); happy for us to challenge this, but in the meantime we'd have to adhere to the format
```suggestion
`Table` – Removed unused `updateAriaLabel` function and event listener
```

alex-ju (dismissed):

alex-ju (approved):
Glad to see these custom events gone (not that there's anything wrong with them but it makes the code harder to maintain/follow).

PR #2692, opened 2025-02-08
 Website - Add fix for dialog z-index issue (HDS-4377)

KristinLBradley:
### :pushpin: Summary

If merged, this PR adds a fix for a z-index issue which was causing the website side navigation and tabs to appear above the backdrop of open dialogs and above the full screen Code Editor.

NOTE: The website tabs will no longer have a fixed layout while dialogs are open with this fix.

#### Previews (open/expand the components to test):
* `CodeEditor`: https://hds-website-git-hds-4377-web-fullscreen-bug-hashicorp.vercel.app/components/code-editor?tab=code#full-screen-mode
* `Flyout`: https://hds-website-git-hds-4377-web-fullscreen-bug-hashicorp.vercel.app/components/flyout?tab=code#how-to-use-this-component
* `Modal`: https://hds-website-git-hds-4377-web-fullscreen-bug-hashicorp.vercel.app/components/modal?tab=code#how-to-use-this-component

**WITH BUG:**
image

**FIXED:**
image

image

image

### :link: External links

* Jira ticket: [HDS-4377](https://hashicorp.atlassian.net/browse/HDS-4377)

***

### 👀 Component checklist

[HDS-4377]: https://hashicorp.atlassian.net/browse/HDS-4377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/styles/pages/application/sidebar.scss
+++ b/website/app/styles/pages/application/sidebar.scss
@@ -11,6 +11,11 @@ $doc-page-sidebar-transition-duration: 0.25s;
 
 // ⚠️ ---  the sidebar "navigation" is different between the homepage and the other pages --- ⚠️
 
+// if dialog is open, set sidebar z-index to 1
+.doc-page-wrapper:has(dialog[open]) {
+  --doc-z-index-sidebar: 1;
+}
+
 // HOME PAGE
 // This is a special case: the sidebar is horizontal and is visible *only* on small viewports (when opened)
 
--- a/website/app/styles/pages/application/sidebar.scss
+++ b/website/app/styles/pages/application/sidebar.scss
@@ -62,11 +62,13 @@ body.application:not(.index) {
   &.doc-page-wrapper:has(dialog[open]),
   &.doc-page-wrapper:has(.hds-code-editor--is-full-screen) {
     --doc-z-index-sidebar: 1;
+
+    .doc-scroll-to-top {z-index: 1;}
   }
 
-  // if code editor is in full screen mode, set page peader z-index to 1
+  // if code editor is in full screen mode, set page header & scroll to top z-index to 1
   &.doc-page-wrapper:has(.hds-code-editor--is-full-screen) {
-    .doc-page-header {z-index: 1;}
+    .doc-page-header, .doc-scroll-to-top {z-index: 1;}
   }
 
   // Notice: for `display` and animation/transition look below
--- a/website/app/styles/pages/application/sidebar.scss
+++ b/website/app/styles/pages/application/sidebar.scss
@@ -11,11 +11,6 @@ $doc-page-sidebar-transition-duration: 0.25s;
 
 // ⚠️ ---  the sidebar "navigation" is different between the homepage and the other pages --- ⚠️
 
-// if dialog is open, set sidebar z-index to 1
-.doc-page-wrapper:has(dialog[open]) {
-  --doc-z-index-sidebar: 1;
-}
-
 // HOME PAGE
 // This is a special case: the sidebar is horizontal and is visible *only* on small viewports (when opened)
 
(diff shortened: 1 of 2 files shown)


website/app/styles/pages/application/sidebar.scss
```diff
@@ -11,6 +11,11 @@ $doc-page-sidebar-transition-duration: 0.25s;
 
 // ⚠️ ---  the sidebar "navigation" is different between the homepage and the other pages --- ⚠️
 
+// if dialog is open, set sidebar z-index to 1
+.doc-page-wrapper:has(dialog[open]) {
+  --doc-z-index-sidebar: 1;
+}
+
```
KristinLBradley, line 18: Let me know if you think this style should be moved somewhere else or if you think I should override the z-index value directly vs. overriding the token in this way...

website/app/styles/pages/application/tabs.scss
```diff
@@ -36,11 +36,15 @@
...
-  // Make tabs position sticky/fixed if the viewport height is greater than 480px
+// Make tabs position sticky/fixed if the viewport height is greater than 480px and dialog is NOT open
+.doc-page-wrapper:not(:has(dialog[open])) {
   @media (height >= 480px) {
-    position: sticky;
-    top: var(--doc-page-header-height);
+    .doc-page-tabs {
+      position: sticky;
+      top: var(--doc-page-header-height);
+    }
```
KristinLBradley, line 47: I tried every combination of z-indexes on various elements including the code blocks and the only solution I could come up with was this. (To just unstick tabs when a dialog is open.)

alex-ju: I think that might be our only option in the current layout. While not ideal, it does solve the problem for the end user.

website/app/styles/pages/application/sidebar.scss
alex-ju: I think it's safe to move this on L66 or similar, inside `body.application:not(.index)` (as we're not going to need it for the index page),

KristinLBradley: I moved it, thanks

alex-ju (commented):
Looking good! 🙌  One thing that I would try and do is hide the 'scroll to top' button (`doc-scroll-to-top`) as well for examples in full-screen – potentially by adjusting the `z-index` (same technique you used for the sidebar)

KristinLBradley:
> Looking good! 🙌 One thing that I would try and do is hide the 'scroll to top' button (`doc-scroll-to-top`) as well for examples in full-screen – potentially by adjusting the `z-index` (same technique you used for the sidebar)

Pushed up a fix for the "scroll to top" control as well now.

alex-ju (approved):

shleewhite:
Sorry to comment on this, I just saw that it closed - but it kinda looks like the page header is still interactive when the modal/flyout Is open even though it is not. Is that intended?

alex-ju:
> Sorry to comment on this, I just saw that it closed - but it kinda looks like the page header is still interactive when the modal/flyout Is open even though it is not. Is that intended?

Good catch! I somehow missed the navigation. @KristinLBradley, would you be able to follow up on this?

KristinLBradley:
> Sorry to comment on this, I just saw that it closed - but it kinda looks like the page header is still interactive when the modal/flyout Is open even though it is not. Is that intended?

Thanks, I'll look at it.

didoo:
> > Sorry to comment on this, I just saw that it closed - but it kinda looks like the page header is still interactive when the modal/flyout Is open even though it is not. Is that intended?
> 
> Thanks, I'll look at it.
> 
> **PR**: #2699

@KristinLBradley @alex-ju sorry if I looked at this only now (I was busy with other things) but I think there may be a simpler solution to the problem that doesn't require complex CSS selectors with `has` and `not` (that may lead to other unexpected side effects if we were to change things again in the future).

The root cause of the issue is in the fact that the rendered components in the code block/snippet are inside two containers (`.doc-code-block` and `.doc-code-block__code-rendered`) both with `position: relative` and `z-index`. In reality (from what I understand) the reason for having these styles is to allow the "Copy button" to be absolutely positioned.

If we change the structure of the (rendered) code snippet and move the button inside a new container (let's say `.doc-code-block__code-snippet-wrapper`), together with the code itself, then we can apply the `position: relative` to this new container, the button is absolutely positioned in relation to this container (which is even more correct) and we can remove the `position: relative` from the parent `.doc-code-block` container and the `.doc-code-block__code-rendered`, solving in this way the problem of having the modal overlay contained in a parent with `position: relative`.

modified

I did a quick test and it worked, but I haven't fully tested the solution (eg, the responsiveness, if there are special cases that I'm not considering, etc).

I'll leave it to you and Alex decide if it's worth to revert these fixes and use the approach suggested above.

PR #2702, opened 2025-02-12
fix: refocus search button when close the dialog

shleewhite:
### :pushpin: Summary

Algolia doesn't provide an onClose or onCancelSearch callback, so there is no ideal way to refocus the dialog when it closes and the user doesn't execute a search.

I ended up implementing it where any time the modal goes from open to close, refocus the search button - even though it will temporarily focus the search button before the navigation happens if the users clicks a link in the search dialog

Other options considered:
- Refactor so we don't use their "detached" UI anymore and instead build something custom (too much effort)
- Add event listeners for cmd+k, escape, and click button (if press escape anywhere on the page, focuses the search button, adds complexity) 

### :link: External links

Jira ticket: [HDS-4419](https://hashicorp.atlassian.net/browse/HDS-4419)

***

### DIFF
--- a/website/app/components/doc/page/header/algolia-search/index.js
+++ b/website/app/components/doc/page/header/algolia-search/index.js
@@ -247,9 +247,6 @@ export default class DocAlgoliaSearchComponent extends Component {
       // add event listener to open/close the modal via `command-K` hot-key
       if (event.metaKey && event.key === 'k') {
         autocompleteInstance.setIsOpen(!this.isModalOpen);
-        if (!this.isModalOpen) {
-          document.getElementById('search-button').focus();
-        }
       }
     });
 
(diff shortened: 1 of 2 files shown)


website/app/components/doc/page/header/algolia-search/index.js
```diff
@@ -240,11 +243,10 @@ export default class DocAlgoliaSearchComponent extends Component {
       });
     }
 
-    // add event listener to open/close the modal via `command-K` hot-key
     document.addEventListener('keydown', (event) => {
+      // add event listener to open/close the modal via `command-K` hot-key
       if (event.metaKey && event.key === 'k') {
         autocompleteInstance.setIsOpen(!this.isModalOpen);
-        this.isModalOpen = !this.isModalOpen;
```
shleewhite, line 247: Note: I moved this to the onStateChange argument on line 60 because if you used cmd+k after opening the dialog with the search button or closing it with escape, the tracked state wasn't properly updated. Now- no matter how the user opens/closes the dialog the tracked state is always correct.

dchyun (approved):

MelSumner (approved):
I see what you mean by complexity but I think in this case its been handled appropriately. 👍

PR #2703, opened 2025-02-12
feat: enable `embroider-optimized` scenario

aklkv:
### :pushpin: Summary

Enable `embroider-optimized`  and remove `allowedToFail`  flag, this is ultimate goal of this whole project and will take a lot longer to figure out, I do not expect it to be merged fast  (this allows consumers to actually run embroider/webpack and vite with `staticInvokables` turned on

* remap index components so that embroider can correctly resolve them, this produces correct map in package.json
* bump all `@embroider/*` pacakges

### :link: External links

Jira ticket: [HDS-3123](https://hashicorp.atlassian.net/browse/HDS-3123)
Figma file: [if it applies]

***

### 👀 Component checklist

[HDS-3123]: https://hashicorp.atlassian.net/browse/HDS-3123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/rollup.config.mjs
```diff
@@ -27,13 +28,22 @@ const plugins = [
...
+  addon.appReexports(
+    [
+      'components/**/!(*types).js',
+      'helpers/**/*.js',
+      'modifiers/**/*.js',
+      'services/**/!(*types).js',
+      'instance-initializers/**/*.js',
+    ],
+    {
+      mapFilename: (filename) => {
```
aklkv, line 40: This is the main change here, it makes sure that compatibility layer with ember-resolver and embroider mapping is correct, this can be removed if HDS team settles on not using index components format at all for now this works, i have tested it in cloud-ui by turning on full embroider compat mode

.changeset/stale-maps-shake.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+remap index components
```
aklkv: not sure what would be correct entry for change log but I needed it in order to publish rc version to test in cloud-ui

didoo: Is this introducing any changes to the `dist` files? How do these changes impact our consumers?

didoo (approved):
If you have tested it and it works, I have no objections.

Only thing to check is why Percy is failing. Is it because it needs to be rebased?

.changeset/stale-maps-shake.md
aklkv: `dist/` is unchanged the only difference you can see is in the resolution map on the pacakge.json which now correctly maps it to non index paths to index paths

KristinLBradley:
Since this branch was merged into `main` there have been test failures: https://github.com/hashicorp/design-system/actions/runs/17385866427/job/49408595590

Any ideas what the cause would be? I wonder also how tests passed on this branch, allowing it to be merged.

aklkv:
this change only effected embroider-optimized scenario, main is failing on ember-beta, I will take a look in detail what might be happening

aklkv:
I have investigated more, it seems that tests are only failing for modal/flyout etc. anything that uses `DialogPrimitive`, also tests are extremely flaky locally if I run full test suite it fails on all of them, if I run just failed once it passes. I have a slight suspicion that there might be something with test-waiteres. but to underscore it has nothing to do with changes in this PR, it might have coincided with new beta being realized on ember side

PR #2709, opened 2025-02-14
feat: all test scenarios

aklkv:
### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist


.github/workflows/ci-components.yml
```diff
@@ -68,11 +68,11 @@ jobs:
         try-scenario:
           - ember-lts-3.28
           - ember-lts-4.12
-         # - ember-release
+          - ember-release
           - ember-beta
-         # - ember-canary
+          - ember-canary
```
Dhaulagiri, line 73: we had specifically excluded canary due to the noise of any failures there not being particularly useful in the past. is there benefit we should be thinking about with bringing it back?

aklkv: aside from just keeping an eye on what might be coming there not much value, I am happy to retract this change if you thing it doesn't makes sense to enable it

PR #2714, opened 2025-02-19
`Stepper::Nav` and `Stepper::List`: Create components

dchyun:
### :pushpin: Summary

This PR adds the following new components to the HDS components library.
- `Hds::Stepper::Nav`
    - Subcomponents: `Hds::Stepper::Nav::Step` and `Hds::Stepper::Nav::Panel`
-  `Hds::Stepper::List`
    - Subcomponent: `Hds::Stepper::List::Step` 

**Preview pages**
- [Stepper Nav](https://hds-showcase-git-dchyun-stepper-component-hashicorp.vercel.app/components/stepper/nav)
- [Stepper List](https://hds-showcase-git-dchyun-stepper-component-hashicorp.vercel.app/components/stepper/list)

[Requirements Doc](https://docs.google.com/document/d/1itgRQ8qmsMRAld_-GRkixlp25ud-WWk6K3viXaJvHyg/edit?tab=t.efihpzui45kb)

#### `Hds::Stepper::Nav` Component
This component  is a visual indicator of an on-page linear progress, where users can see where they are in a given flow.

The component can be used in two ways. Either using the contextual `<S.Step>` and `<S.Panel>` components, or using the `@steps` argument and passing in an array.

**Using Contextual Components** 
```
// Example
<Hds::Stepper::Nav
  @currentStep={{this.currentStep}}
  @onStepClick={{this.updateCurrentStep}}
  as |S|
>
  <S.Step>
    <:title>Step 1 Title</:title>
    <:description>Description</:description>
  </S.Step>
  <S.Step>
    <:title>Step 2 Title</:title>
  </S.Step>
  <S.Panel>
    Step 1 content
  </S.Panel>
  <S.Panel>
    Step 2 content
  </S.Panel>
</Hds::Stepper::Nav>
```
**Using @steps argument** 
```
// Example
{{#let
  (array
    (hash title='Step 1 Title' description='Description')
    (hash title='Step 2 Title')
  )
  as |steps|
}}

<Hds::Stepper::Nav
  @steps={{steps}}
  @currentStep={{this.currentStep}}
  @onStepChange={{this.updateCurrentStep}}
>
  <:body>
    {{#if (eq this.currentStep 0)}}
      Step 1 content
    {{else}}
      Step 2 content
    {{/if}}
  </:body>
</Hds::Stepper::Nav>
```
`Stepper::Nav`
- `<[S.Step]>`: `Stepper::Nav::Step` as yielded component
- `<[S.Panel]>`: `Stepper::Nav::Panel` as yielded component 
- `<:body>`: Named block used for step content when using `steps` argument
- `steps`: Array of data for steps to be created
- `currentStep`: Base indexed value for step that is currently active
- `isInteractive`: If true, the user may use the stepper to navigation to previous steps
- `titleTag`: Sets the DOM tag used for step title text
- `onStepChange`: Event triggered when the active step is changed through a click on a step

`[S].Step`
- `<:title>`: Contextual block for title text
- `<:description>`: Contextual block for description text

`[S].Panel`
- `yield`: Elements passed as children are yielded as inner content of a <section> HTML element

#### `Hds::Stepper::List` Component
This component is a visual indicator of a linear progress, where users can see where they are in a given flow.

```
// Example
<Hds::Stepper::List as |S|>
  <S.Step @status="complete">
    <:title>Title</:title>
    <:description>Description</:description>
    <:content>
      Step 1 content
    </:content>
  </S.Step>
  <S.Step @status="progress">
    <:title>Title</:title>
    <:description>Description</:description>
    <:content>
      Step 2 content
    </:content>
  </S.Step>
  <S.Step @status="incomplete">
    <:title>Title</:title>
  </S.Step>
</Hds::Stepper::List>
```

`Stepper::List`
- `<[S].Step>`: `Stepper::List::Step` as yielded component
- `titleTag`: Sets the DOM tag used for step title text

`[S].Step`
- `<:title>`: Contextual block for title text
- `<:description>`: Contextual block for description text
- `<:content>`: Contextual block for any additional content
- `status`: Status of the step based on current indicator statuses

#### `Hds::Stepper::Nav`

**Interactive**
Screenshot 2025-03-13 at 3 19 13 PM

**Non-interactive**
Screenshot 2025-03-13 at 3 19 40 PM

#### `Hds::Stepper::List`
Screenshot 2025-03-13 at 3 18 49 PM

### :link: External links

Jira ticket: [HDS-4411](https://hashicorp.atlassian.net/browse/HDS-4411)
Requirements Doc: [Link](https://docs.google.com/document/d/1itgRQ8qmsMRAld_-GRkixlp25ud-WWk6K3viXaJvHyg/edit?tab=t.efihpzui45kb)
Figma file: [Link](https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/branch/whv92tH8FWhxb4kk6Ip3qp/HDS-Components-v2.0?m=auto&node-id=75135-37374&t=8Te4rqjQliS7nHdz-1)

***

### 👀 Component checklist

[HDS-4411]: https://hashicorp.atlassian.net/browse/HDS-4411?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,326 @@
...
+.hds-stepper-nav .hds-stepper-nav__step:last-of-type .hds-stepper-nav__step__progress::after {
+  border-radius: 0 20px 20px 0;
+}
+
+.hds-stepper-nav__step__text {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: 14px 12px 12px 12px;
+  border-radius: 0 0 var(--token-border-radius-small) var(--token-border-radius-small);
```
majedelass, line 85: Instead of small (5), it should be medium (6), the reference from figma:
image

```suggestion
  border-radius: 0 0 var(--token-border-radius-medium) var(--token-border-radius-medium);
```

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,326 @@
...
+.hds-stepper-nav .hds-stepper-nav__step:last-of-type .hds-stepper-nav__step__progress::before,
+.hds-stepper-nav .hds-stepper-nav__step:last-of-type .hds-stepper-nav__step__progress::after {
+  border-radius: 0 20px 20px 0;
+}
+
+.hds-stepper-nav__step__text {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: 14px 12px 12px 12px;
```
majedelass, line 84: I'd adjust to 16px / 8px, that's how the Figma component is structured:
image


```suggestion
  padding: 16px 8px 12px 8px;
```

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,326 @@
...
+  }
+
+  &::after {
+    background-color: var(--token-color-palette-neutral-700);
+  }
+}
+
+.hds-stepper-nav .hds-stepper-nav__step:first-of-type .hds-stepper-nav__step__progress::before,
+.hds-stepper-nav .hds-stepper-nav__step:first-of-type .hds-stepper-nav__step__progress::after {
+  border-radius: 20px 0 0 20px;
```
majedelass, line 72: I'd adjust to what border radius we have now, to follow what is in existence in Figma:
image

```suggestion
  border-radius: var(--token-border-radius-x-small) 0 0 var(--token-border-radius-x-small);
```

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,326 @@
...
+}
+
+.hds-stepper-nav .hds-stepper-nav__step:last-of-type .hds-stepper-nav__step__progress::before,
+.hds-stepper-nav .hds-stepper-nav__step:last-of-type .hds-stepper-nav__step__progress::after {
+  border-radius: 0 20px 20px 0;
```
majedelass, line 77: I'd update to align above here too:

image

```suggestion
  border-radius: 0 var(--token-border-radius-x-small) var(--token-border-radius-x-small) 0;
```

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,326 @@
...
+// Active
+
+.hds-stepper-nav__step--active {
+  .hds-stepper-nav__step__btn {
+    @include hds-focus-ring-with-pseudo-element(
+      $top: -3px,
+      $right: -1px,
+      $bottom: -1px,
+      $left: -1px,
+      $radius: 0 0 6px 6px
```
majedelass, line 121: Does this work here too?

```suggestion
      $radius: 0 0 var(--token-border-radius-medium) var(--token-border-radius-medium)
```

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,326 @@
...
+    );
+  }
+
+  .hds-stepper-nav__step__progress::after {
+    width: calc(50% - 16px); // 1/2 width of icon (12px) + 4px margin
+    border-radius: 0 20px 20px 0;
```
majedelass, line 127: Alignment elsewhere here too

```suggestion
    border-radius: 0 var(--token-border-radius-x-small) var(--token-border-radius-x-small) 0;
```

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,326 @@
...
+.hds-stepper-nav__step__text {
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  padding: 14px 12px 12px 12px;
+  border-radius: 0 0 var(--token-border-radius-small) var(--token-border-radius-small);
+}
+
+.hds-stepper-nav__step__title {
+  color: var(--token-color-foreground-primary);
```
majedelass, line 89: Strong instead of primary, same in Figma:

image

```suggestion
  color: var(--token-color-foreground-strong);
```

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,326 @@
...
+
+.hds-stepper-nav__step--complete .hds-stepper-nav__step__progress::after {
+  width: 100%;
+}
+
+// NAV INTERACTIVE
+
+.hds-stepper-nav__step--nav-interactive {
+  .hds-stepper-nav__step__progress::after {
+    background-color: var(--token-color-palette-blue-300);
```
majedelass, line 141: We made it lighter in Figma:

image

```suggestion
    background-color: var(--token-color-palette-blue-200);
```

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,326 @@
...
+    top: 10px;
+    left: 0;
+    z-index: -1;
+    width: 100%;
+    height: 4px;
+    content: "";
+  }
+
+  &::before {
+    background-color: var(--token-color-palette-alpha-200);
```
majedelass, line 62: We need to layer the alpha 200 with white behind it so its consistent regardless of where this is placed on top of:

```suggestion
    background: linear-gradient(0deg, var(--token-color-palette-alpha-200) 0%, var(--token-color-palette-alpha-200) 100%), --token-color-palette-neutral-100;
```

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,105 @@
...
+        position: absolute;
+        top: 2px;
+        left: 11px;
+        width: 2px;
+        content: "";
+      }
+
+      &::before {
+        height: calc(100% + $hds-stepper-list-step-padding-bottom);
+        background-color: var(--token-color-palette-alpha-200);
```
majedelass, line 41: Same as nav variant, layering alpha with white.

```suggestion
        background: linear-gradient(0deg, var(--token-color-palette-alpha-200) 0%, var(--token-color-palette-alpha-200) 100%), --token-color-palette-neutral-100;
```

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,105 @@
...
+  grid-template-columns: auto 1fr;
+
+  &:not(:last-of-type) {
+    padding-bottom: $hds-stepper-list-step-padding-bottom;
+
+    .hds-stepper-list__step__progress {
+      &::before,
+      &::after {
+        position: absolute;
+        top: 2px;
```
majedelass, line 33: Similar to figma, adjust this to 1px

image


```suggestion
        top: 1px;
```

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,105 @@
...
+
+      &::before {
+        height: calc(100% + $hds-stepper-list-step-padding-bottom);
+        background-color: var(--token-color-palette-alpha-200);
+      }
+
+      &::after {
+        z-index: -1;
+        background-color: var(--token-color-palette-neutral-700);
+        border-radius: 0 0 20px 20px;
```
majedelass, line 47: Update to the appropriate token (3px)

```suggestion
        border-radius: 0 0 var(--token-border-radius-x-small) var(--token-border-radius-x-small);
```

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,105 @@
...
+.hds-stepper-list__step__progress {
+  position: relative;
+}
+
+.hds-stepper-list__step__text {
+  margin-top: 2px;
+}
+
+.hds-stepper-list__step__title {
+  color: var(--token-color-foreground-primary);
```
majedelass, line 62: Change to strong instead of primary.

```suggestion
  color: var(--token-color-foreground-strong);
```

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,105 @@
...
+}
+
+.hds-stepper-list__step__description {
+  margin-top: 4px;
+}
+
+.hds-stepper-list__step__content {
+  margin-top: 12px;
```
majedelass, line 70: Instead of 12, it should be 8, as seen in Figma:

image


```suggestion
  margin-top: 8px;
```

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,105 @@
...
+// Complete
+
+.hds-stepper-list__step {
+  &:not(:last-of-type).hds-stepper-list__step--complete .hds-stepper-list__step__progress::after {
+    height: calc(100% + $hds-stepper-list-step-padding-bottom - 4px);
+  }
+
+  &:not(:last-of-type, :first-of-type).hds-stepper-list__step--complete .hds-stepper-list__step__progress::after {
+    top: -4px;
+    height: calc(100% + $hds-stepper-list-step-padding-bottom);
```
majedelass, line 92: I'd double check if this follow Figma. Can't tell, but can you double check?

majedelass (commented):
I found a couple of items that are related to figma discrepancies, some token usages and other small bits here and there. Let me know if you need to pair on anything!

packages/components/src/styles/components/stepper/nav.scss
dchyun: Yeah it does

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -324,3 +324,30 @@
...
+@media screen and (width <= 550px) {
+  .hds-stepper-nav__list {
+    flex-direction: column;
+  }
+
+  .hds-stepper-nav__step:not(:last-of-type) {
+    margin-bottom: 16px;
+  }
+
+  .hds-stepper-nav__step__progress {
```
KristinLBradley, line 339: It's usually not recommended to have an overly nested naming structure like this when using BEM format naming. You should only go one level deep for names of child components. (i.e. there should be a max of one `__` indicating a child component of a parent block within a name)

So, for example, you could rename this to something like just `.hds-stepper-nav__progress`, if that makes sense, or `.hds-stepper-nav__step-progress` if that's clearer. (or choose another more descriptive name.)

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -324,3 +324,30 @@
...
+    justify-content: start;
+
+    &::before,
+    &::after {
+      content: none;
+    }
+  }
+
+  .hds-stepper-nav__step__text {
+    align-items: start;
```
KristinLBradley, line 351: Here too, I would rename this to not have an overly nested naming structure. (Only use one "__" max in the name.)

.changeset/four-jokes-wait.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Stepper::List` - Added `Stepper::List` and related sub-components
+
+`Stepper::Navigation` - Added `Stepper::Navigaiton` and related sub-components
```
KristinLBradley, line 7:
```suggestion
`Stepper::Navigation` - Added `Stepper::Navigation` component and related sub-components
```

.changeset/four-jokes-wait.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Stepper::List` - Added `Stepper::List` and related sub-components
```
KristinLBradley, line 5:
```suggestion
`Stepper::List` - Added `Stepper::List` component and related sub-components
```

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,111 @@
...
+
+      &::after {
+        z-index: -1;
+        background-color: var(--token-color-palette-neutral-700);
+      }
+    }
+  }
+}
+
+.hds-stepper-list__step__progress {
```
KristinLBradley, line 52: I commented on an individual commit but repeating here for easier reference.

In BEM naming conventions, it's recommended only going one level deep in the naming structure for child components, so class names should only contain one "__" max.

packages/components/src/styles/components/stepper/nav.scss
dchyun: Thanks for catching this. I've aligned to the BEM format. I used names such as `hds-stepper-nav__step-progress` as that aligned with similar elements in other components such as the tabs.

packages/components/src/components/hds/stepper/list/index.ts
```diff
@@ -0,0 +1,73 @@
...
+  willDestroyStep(element: HTMLElement): void {
+    this._stepNodes = this._stepNodes.filter(
+      (node): boolean => node.id !== element.id
+    );
+    this._stepIds = this._stepIds.filter(
+      (stepId): boolean => stepId !== element.id
+    );
+  }
+
+  /**
```
zamoore, line 63: You can remove the jsdoc comments unless the getter would benefit from more description. For this, you can just add a `string` return type.

zamoore, line 63: You can remove the jsdoc comments unless the getter would benefit from more description. For this, you can just add a string return type.

packages/components/src/components/hds/stepper/list/step.hbs
```diff
@@ -0,0 +1,37 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<li
+  class={{this.classNames}}
+  ...attributes
+  id={{this._stepId}}
+  {{this._setUpStep this.didInsertNode this.willDestroyNode}}
```
zamoore, line 9: [question] Instead of calling `this.didInsertNode` and `this.willDestroyNode`, could you call `@didInsertNode` and `@willDestroyNode` and then remove the actions?

dchyun: In the `stepper/list/step.ts` file we are doing a check to make sure what is passed into the `@didInsertNode` and `@willDestroyNodes` args is a function. This was a pattern carried over from the tabs component. I guess that check is a little unnecessary since the parent is the one passing in those args.

packages/components/src/components/hds/stepper/nav/step.hbs
```diff
@@ -0,0 +1,80 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+{{! template-lint-disable require-context-role no-invalid-role }}
+{{! template-lint-disable require-presentational-children }}
+<li class={{this.classNames}} ...attributes role={{if this.isNavInteractive "presentation"}}>
```
dchyun, line 7: [Question] In the cloud UI [stepper form component](https://github.com/hashicorp/cloud-ui/blob/b2f16ad058a7ba06e5002aa656d6e992db36fd28/addons/core/package/src/components/stepper-form/tab.gts#L90) `...attributes`  is set on the `<button` instead of the `<li>`, and they are attaching data analytics attributes this way. Our tabs component has `...attributes` set on the `<li>`, but would it be better to move it and align to cloud ui?

packages/components/src/components/hds/stepper/list/step.hbs
zamoore: Unless you want to render the components independently, I would simplify it to reduce comlpexity.

packages/components/src/components/hds/stepper/nav/step.hbs
```diff
@@ -0,0 +1,80 @@
...
+        <Hds::Stepper::Step::Indicator
+          @text="{{this.stepNumber}}"
+          @status={{this.stepIndicatorStatus}}
+          @isInteractive={{true}}
+          class="hds-stepper-nav__step-indicator"
+        />
+      </div>
+      <div class="hds-stepper-nav__step-text">
+        <Hds::Text::Body class="hds-stepper-nav__step-title" @tag={{this.titleTag}} @size="200" @weight="semibold">
+          <span class="sr-only">{{this.statusSrOnlyText}}</span>
```
shleewhite, line 32: Could this be moved to after the description? I ask because even though the status icon is before the visible text in the tab, the label should always start with the visible text. This makes it easier for voice dictation users to click the button. 

Right now the label is "Complete: title description", I would suggest updating it to "Title description (complete)"

```suggestion
          <span class="sr-only">({{this.statusSrOnlyText}})</span>
```

packages/components/src/components/hds/stepper/nav/step.hbs
```diff
@@ -0,0 +1,80 @@
...
+    </button>
+  {{else}}
+    <div
+      class="hds-stepper-nav__step-content"
+      id={{this._stepId}}
+      aria-current={{if (eq this.status "active") "true" "false"}}
+      {{this._setUpStep this.didInsertNode this.willDestroyNode}}
+    >
+      <div class="hds-stepper-nav__step-progress">
+        <Hds::Stepper::Step::Indicator
```
shleewhite, line 54: I think the step indicator should be `aria-hidden` when it is not interactive.. reasoning being- the ordered list handles announcing which item in the list you are on so the step number being available to screen readers is redundant information.

packages/components/src/components/hds/stepper/nav/panel.hbs
```diff
@@ -0,0 +1,17 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<section
+  class={{this.classNames}}
+  ...attributes
+  role="tabpanel"
```
shleewhite, line 8: The role shouldnt be tabpanel if it is non-interactive. The reason is setting an ARIA role sets expectations for the user on how the UI will work.

showcase/tests/acceptance/components/hds/stepper-test.js
```diff
@@ -12,8 +12,15 @@ module('Acceptance | Component | hds/stepper', function (hooks) {
   setupApplicationTest(hooks);
 
   test('Components/hds/stepper page passes automated a11y checks', async function (assert) {
+    let axeOptions = {
+      rules: {
+        'landmark-unique': {
```
shleewhite, line 17: Was this disabled because sections didn't have unique names?

showcase/tests/integration/components/hds/stepper/list-step-test.js
```diff
@@ -0,0 +1,133 @@
...
+  test('it renders the content when the content contextual component block is used', async function (assert) {
+    await render(
+      hbs`
+          <Hds::Stepper::List::Step>
+            <:content>Test</:content>
+          </Hds::Stepper::List::Step>
+        `
+    );
+    assert.dom('.hds-stepper-list__step-content').containsText('Test');
+  });
```
shleewhite, line 132: Suggestion: would be useful to add tests that the correct screen reader text is added for Stepper list steps and Stepper Nav steps

showcase/tests/integration/components/hds/stepper/nav-panel-test.js
```diff
@@ -0,0 +1,53 @@
...
+    await this.createNavPanel({ currentStep: 1 });
+    assert.dom('.hds-stepper-nav__panel').hasAttribute('hidden');
+    assert.dom('#test-panel-content').doesNotExist();
+  });
+
+  test('it sets the panel content to visible when the @currentStep argument matches the panel index in the @panelIds argument', async function (assert) {
+    await this.createNavPanel({ currentStep: 0 });
+    assert.dom('.hds-stepper-nav__panel').hasNoAttribute('hidden');
+    assert.dom('#test-panel-content').exists();
+  });
```
shleewhite, line 52: Suggestion: add tests that aria-labelledby is always present, also that the correct role is set (depending on if the list is interactive or not)

shleewhite (commented):
Overall works well! Most of my comments are around a11y details.

packages/components/src/components/hds/stepper/list/index.hbs
```diff
@@ -0,0 +1,17 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<ol class="hds-stepper-list" ...attributes>
```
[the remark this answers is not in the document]
KristinLBradley: Here's an example from another component for reference: https://helios.hashicorp.design/components/accordion?tab=code#accessible-name

showcase/app/templates/components/stepper.hbs
KristinLBradley, line 1: I think I would prefer that the `StepperNav` and `StepperList` each have their own Showcase pages and I would also keep the `StepperIndicator` on its own page as it is currently: https://hds-showcase.vercel.app/components/stepper

What do others think though? @hashicorp/hds-engineering

packages/components/src/components/hds/stepper/list/step.ts
```diff
@@ -0,0 +1,156 @@
...
+  };
+  Element: HTMLElement;
+}
+
+export default class HdsStepperListStep extends Component<HdsStepperListStepSignature> {
+  /**
+   * Generate a unique ID for the Step
+   * @return {string}
+   * @param _stepId
+   */
```
KristinLBradley, line 48: We're not using this style anymore. You could replace with a simple one-line comment like:

```
// Generate a unique ID for the Step
```

showcase/app/templates/components/stepper.hbs
```diff
@@ -67,4 +362,114 @@
       <SG.Item />
     {{/each}}
   </Shw::Grid>
+
+  <Shw::Divider />
+
+  <Shw::Text::H2>Examples of nav implementation</Shw::Text::H2>
```
KristinLBradley, line 368: Could you add at least one example of a StepperList implementation as well?

showcase/app/templates/components/stepper.hbs
shleewhite: I agree, it would mirror how the website docs are set up too.

packages/components/src/components/hds/stepper/nav/index.ts
```diff
@@ -0,0 +1,196 @@
...
+    });
+
+    return () => {};
+  });
+
+  /**
+   * @param currentStep
+   * @type {number}
+   * @default 0
+   */
```
KristinLBradley, line 66: You could delete comments like these. We don't use them anymore. If the function of something needs explanation to clarify just add a simple comment. Most of these seem pretty self explanatory though.

Delete:
```
  /**
   * @param currentStep
   * @type {number}
   * @default 0
   */
```

packages/components/src/components/hds/stepper/list/index.ts
```diff
@@ -0,0 +1,62 @@
...
+        Step?: ComponentLike<HdsStepperListStepSignature>;
+      },
+    ];
+  };
+  Element: HTMLElement;
+}
+
+export default class HdsStepperList extends Component<HdsStepperListSignature> {
+  @tracked private _stepIds: HdsStepperListStepIds = [];
+  @tracked private _stepNodes: HTMLElement[] = [];
```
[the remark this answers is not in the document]
KristinLBradley: It looks like maybe these were something needed for the `StepperNav`? But maybe not for the `StepperList`. @dchyun could you please double check?

packages/components/src/components/hds/stepper/nav/step.hbs
KristinLBradley: They can also still access the button from data attributes added on the `li` parent so I agree with Lee.

packages/components/src/components/hds/stepper/nav/step.hbs
```diff
@@ -0,0 +1,80 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+{{! template-lint-disable require-context-role no-invalid-role }}
+{{! template-lint-disable require-presentational-children }}
+<li class={{this.classNames}} ...attributes role={{if this.isNavInteractive "presentation"}}>
+  {{#if this.isNavInteractive}}
+    <button
+      class="hds-stepper-nav__step-content hds-stepper-nav__step-btn"
```
KristinLBradley, line 10: We tend to avoid abbreviating words in names like "btn". I would instead spell it out or rename it.

packages/components/src/components/hds/stepper/list/index.ts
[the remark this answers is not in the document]
dchyun: My bad this has been removed now. It was not needed.

packages/components/src/components/hds/stepper/nav/step.hbs
dchyun: I have been referencing the [w3c step indicator example](https://www.w3.org/WAI/tutorials/forms/multi-page/#using-step-by-step-indicator), and in that example the status sr only text is placed before the title. Is it because this can be a clickable button it's better to have the text after? 

Screenshot 2025-03-06 at 12 57 35 PM

packages/components/src/components/hds/stepper/list/index.hbs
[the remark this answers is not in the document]
dchyun: I can add this as an argument, but is the title text not sufficient to cover this for a step? The title is contained inside the `li`, and is required for each step. Or is the issue that the titles could possibly not be unique?

showcase/tests/acceptance/components/hds/stepper-test.js
dchyun: It's added for the showcase examples where the `Nav::Step` and `Nav::Panel` are shown on their own. Errors were getting thrown because there were elements with `role="tab"` and `role="tabpanel` without an partner element attached to them.

showcase/tests/integration/components/hds/stepper/list-step-test.js
dchyun: There are tests in the `// STATUS` section for the list and the `// STATES` section for the nav testing that the correct sr text gets added.

showcase/app/templates/components/stepper.hbs
dchyun: Updated to match the website and separate out the pages

packages/components/src/components/hds/stepper/list/index.hbs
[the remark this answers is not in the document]
zamoore: I think it's that the list itself needs to have a unique name if I'm reading the initial comment right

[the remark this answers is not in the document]
shleewhite: Yeah thats correct, people using a screen reader can navigate by list so it is important for the list itself to be labelled.

packages/components/src/components/hds/stepper/nav/step.hbs
shleewhite: Ahh I see, their example is slightly different than ours. 

For them, they have visibly hidden text "Current: " _outside_ the "Billing Address" link. This is ok because the visually hidden text isn't part of the link. I also disagree with this implementation because visually, the icon is part of the link but to a screen reader it is outside.

In our use case, the entire step is wrapped in a button because we want a bigger target. So we have the icon + visually hidden text + the visible text inside. This is why I would prefer the hidden text after the hidden text.

packages/components/src/components/hds/stepper/list/index.hbs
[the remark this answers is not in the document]
MelSumner: Here's a screenshot of what that would look like: 
CleanShot 2025-03-07 at 09 27 56

Now, you might be thinking that they won't have that many of this component on the page. BUT they could have other lists on the page. That's why each list needs an AccName.

MelSumner:
I'm getting a false positive (I think) on the Stepper::Navigation showcase page. I've filed an issue with axe-core: https://github.com/dequelabs/axe-core/issues/4722

packages/components/src/components/hds/stepper/nav/step.hbs
shleewhite: Hmm I don't think we need to hide it for both cases. Reason being, when its not interactive, it is an ordered list and the index of each list item is announced and its understood that the order is meaningful.

For interactive ones, it is no longer a list but a set of tabs so the step number information would be useful.

showcase/tests/integration/components/hds/stepper/list-step-test.js
shleewhite: Cool! Sorry, must not have seen them before.

packages/components/src/components/hds/stepper/list/index.hbs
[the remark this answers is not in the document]
dchyun: Thanks @shleewhite @MelSumner this clears it up. Will add this in `@ariaLabel` as a required argument.

MelSumner:
cc @hashicorp/hds-design 
I think this will likely fail color only for a11y. Looking at it with the different color-blindness tools, I just don't see enough of a difference to be able to discern, as a user, which one is interactive and which one isn't. 

Example: 
CleanShot 2025-03-07 at 11 27 22

For clarity, I'm using the [colorblindness emulator extension](https://chromewebstore.google.com/detail/colorblindness-emulator/ffmdflnkahdfgkhneidigopdlokgipki?hl=en) for Chrome.

heatherlarsen:
@MelSumner Can you please be more specific? Which elements are you comparing to one another?

heatherlarsen:
@shleewhite For clarification, are you all comparing the two variants (blue vs black) or between the steps within a variant (check vs number)?

shleewhite:
> @shleewhite For clarification, are you all comparing the two variants (blue vs black) or between the steps within a variant (check vs number)?

Comparing the 2 variants.

shleewhite:
> @shleewhite For clarification, are you all comparing the two variants (blue vs black) or between the steps within a variant (check vs number)?

@heatherlarsen I am comparing the 2 variants.

packages/components/src/styles/components/stepper/list.scss
KristinLBradley: @dchyun Could you please check the suggested changes from @majedelass and apply them if appropriate?

KristinLBradley:
> @heatherlarsen I think my specific concern is that while this may technically meet the requirements for use of color, it doesn't meet the intention of use of color.
> 
> Specifically, the use of checkmarks vs numbers is not a way that users know that something is interactive or not. It's a tiny design detail that I suspect only trained eyes (designers) will notice.
> 
> I suppose if we don't _really expect_ users to know (or intuit) that the stepper could be interactive, then 🤷 But if we are really going for differentiation in the UI/for improved UX, then the interactive version needs to be dramatically and distinctly different.

@MelSumner In the Guidelines, we tell consumers to always pair the Stepper Nav with buttons for navigation: https://github.com/hashicorp/design-system/pull/2727/files#r1988406951

Would that be enough to satisfy accessibility requirements?

shleewhite:
> @MelSumner In the Guidelines, we tell consumers to always pair the Stepper Nav with buttons for navigation: https://github.com/hashicorp/design-system/pull/2727/files#r1988406951
> 
> Would that be enough to satisfy accessibility requirements?

@KristinLBradley  I dont think so, since the non-interactive nav is also always paired with buttons. The issue is being able to distinguish the two variants without color.

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,111 @@
...
+.hds-stepper-list__step-text {
+  margin-top: 2px;
+}
+
+.hds-stepper-list__step-title {
+  color: var(--token-color-foreground-strong);
+}
+
+.hds-stepper-list__step-description {
+  margin-top: 4px;
```
majedelass, line 65: I can't tell if this is applied else where, or if the color is accurate in the showcase, so adding it here:
```suggestion
  margin-top: 4px;
  color: var(--token-color-foreground-primary);
```

packages/components/src/components/hds/stepper/nav/step.hbs
```diff
@@ -0,0 +1,87 @@
...
+  SPDX-License-Identifier: MPL-2.0
+}}
+{{! template-lint-disable require-context-role no-invalid-role }}
+{{! template-lint-disable require-presentational-children }}
+{{#if this.isNavInteractive}}
+  <li class={{this.classNames}} ...attributes role={{if this.isNavInteractive "presentation"}}>
+    <button
+      class="hds-stepper-nav__step-content hds-stepper-nav__step-button"
+      id={{this._stepId}}
+      tabindex={{unless (eq this.status "active") "-1"}}
```
shleewhite, line 12: Woahhh I haven't see `unless` before, TIL

packages/components/src/components/hds/stepper/nav/step.hbs
```diff
@@ -0,0 +1,87 @@
...
+        <Hds::Stepper::Step::Indicator
+          @text="{{this.stepNumber}}"
+          @status={{this.stepIndicatorStatus}}
+          @isInteractive={{true}}
+          class="hds-stepper-nav__step-indicator"
+        />
+      </div>
+      <div class="hds-stepper-nav__step-text">
+        <Hds::Text::Body class="hds-stepper-nav__step-title" @tag={{this.titleTag}} @size="200" @weight="semibold">
+          <span class="sr-only">{{this.statusSrOnlyText}}</span>
```
shleewhite, line 32: @dchyun I saw that the sr-only text got moved after the title for the list variant, but not for the nav one yet.

packages/components/src/components/hds/stepper/list/index.ts
```diff
@@ -0,0 +1,52 @@
...
+  @tracked private _stepIds: HdsStepperListStepIds = [];
+
+  get titleTag(): HdsStepperTitleTags {
+    return this.args.titleTag ?? HdsStepperTitleTagValues.Div;
+  }
+
+  @action
+  didInsertStep(element: HTMLElement): void {
+    // eslint-disable-next-line ember/no-runloop
+    schedule('afterRender', (): void => {
```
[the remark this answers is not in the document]
dchyun: I can update this to pass the id directly, but the `schedule` was added because updating the tracked array with `this._stepIds = [...this._stepIds, id];` causes an error to be thrown. It seems to not like that the tracked variable is both being used in a calculation and reassigned in one go. 

Screenshot 2025-03-13 at 5 43 28 PM

I've tried other ways of updating the array and none seemed to fix the error. The only solutions that worked were either using `schedule` or going back to the `did-insert` modifier. But I could be missing something here.

packages/components/src/components/hds/stepper/nav/index.ts
```diff
@@ -0,0 +1,239 @@
...
+      } else {
+        return currentStep;
+      }
+    } else {
+      return 0;
+    }
+  }
+
+  get isInteractive(): boolean {
+    return this.args.isInteractive != undefined
```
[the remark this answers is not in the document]
dchyun: Ran into an issue doing it this way. With `this.args.isInteractive`, both `false` and `undefined` were treated the same and would return `true`. So when `@isInteractive` was set to `false` this was actually returning `true`. 

That's why I did the check based on `undefined` instead.

packages/components/src/components/hds/stepper/nav/step.hbs
dchyun: This is an odd one. It's been moved correctly locally and looking at the [branch file](https://github.com/hashicorp/design-system/blob/cafa954575f59e450fc74bf6de3e0f2b6c0e7784/packages/components/src/components/hds/stepper/nav/step.hbs#L71) it's been moved, but in the PR and in vercel it looks the old way. 

I'm going to try rebasing to main and see if that clears it up in the PR.

dchyun: Fixed!

packages/components/src/components/hds/stepper/list/index.ts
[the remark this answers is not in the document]
dchyun: @zamoore do you know the proper pattern to follow here for updating a tracked array on load? Or how we can avoid this error without using `schedule`?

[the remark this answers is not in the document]
zamoore: I don't think there's a way around it here without changing the component API. You would need to make the list aware of the steps before rendering. I think this is an acceptable use-case.

packages/components/src/components/hds/stepper/list/step.ts
```diff
@@ -0,0 +1,96 @@
...
+export default class HdsStepperListStep extends Component<HdsStepperListStepSignature> {
+  private _stepId = 'step-' + guidFor(this);
+
+  private _setUpStep = modifier(
+    (
+      element: HTMLElement,
+      [insertCallbackFunction, destoryCallbackFunction]
+    ) => {
+      if (typeof insertCallbackFunction === 'function') {
+        insertCallbackFunction(element);
```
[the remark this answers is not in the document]
zamoore: Unfortunately, the issue isn’t what we pass, but when _stepIds is updated. Since we're reading and modifying it in the same cycle, Ember throws an error.

packages/components/src/components/hds/stepper/nav/index.ts
```diff
@@ -0,0 +1,239 @@
...
+  @tracked private _stepIds: HdsStepperNavStepIds = [];
+  @tracked private _stepNodes: HTMLElement[] = [];
+  @tracked private _panelNodes: HTMLElement[] = [];
+  @tracked private _panelIds: HdsStepperNavPanelIds = [];
+
+  private _element!: HTMLDivElement;
+
+  private _setUpStepperNav = modifier((element: HTMLDivElement) => {
+    // eslint-disable-next-line ember/no-runloop
+    schedule('afterRender', (): void => {
```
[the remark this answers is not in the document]
zamoore: Agreed. I believe this will work fine without scheduling it after render.

packages/components/src/components/hds/stepper/nav/index.ts
```diff
@@ -0,0 +1,239 @@
...
+        currentStepIndex,
+        this._stepIds.length - 1
+      );
+      this.focusStep(prevStepIndex, event);
+    }
+  }
+
+  // Update the step arrays based on how they are ordered in the DOM
+  private updateSteps(): void {
+    const steps = this._element.querySelectorAll(STEP_ELEMENT_SELECTOR);
```
zamoore, line 177: Could you do this with modifiers similar to how it was done in the list? Querying the DOM to determine state often ends up with things being out of state and a need to use the run loop more often.

dchyun: Originally this did mirror what was done in the list. An issue came up when steps were added or removed from the flow after it's first render. 

Before, a new step id was always appended to the end of `this._stepIds`. However if a newly added step wasn't at the end of the flow, but in the middle, then the order of `this._stepIds` would not match the actual order of the steps in the DOM. The order of the ids is how step states are being determined. [Showcase example of use case](https://hds-showcase-git-dchyun-stepper-component-hashicorp.vercel.app/components/stepper/nav#number-of-steps-changing-based-on-input)

The only way I found to correctly order the steps was by looking at the dom. I'm not sure if there's a way in ember to know the order of the conditional component instances, but if so I could look into that.

packages/components/src/components/hds/stepper/nav/index.ts
```diff
@@ -0,0 +1,239 @@
...
+      newStepIndex = (newStepIndex + increment) % this._stepIds.length;
+      isNewStepInteractive = this.isStepInteractive(
+        this._stepNodes[newStepIndex]!
+      );
+    }
+    return newStepIndex;
+  }
+
+  private isStepInteractive(el: HTMLElement): boolean {
+    return !(el.getAttribute('aria-disabled') === 'true');
```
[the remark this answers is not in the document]
dchyun: Good catch, was actually able to refactor and remove this entirely.

packages/components/src/components/hds/stepper/list/step.ts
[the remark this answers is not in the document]
dchyun: I've kept the functions unchanged, and still passing in the element, as changing it to an id won't fix the other issue. But can update it to use ids instead if we want.

shleewhite (dismissed):
Nice! Thanks for making all the fixes.

packages/components/src/components/hds/stepper/nav/index.ts
zamoore: That makes sense. I didn't consider that steps could be dynamic.

zamoore (dismissed):

dchyun:
> The `Stepper::List` black bar isn't rendering and I'm not sure why. Once that is fixed, I'll be able to approve this.

Good catch. Issue has been fixed.

majedelass (approved):
👍 Looking great :shipit:

packages/components/src/components/hds/stepper/list/step.ts
```diff
@@ -0,0 +1,96 @@
...
+import {
+  HdsStepperStatusesValues,
+  HdsStepperTitleTagValues,
+} from '../types.ts';
+import {
+  type HdsStepperStatuses,
+  type HdsStepperTitleTags,
+  type HdsStepperListStepIds,
+  HdsStepperStatusToSrOnlyText,
+} from '../types.ts';
```
KristinLBradley, line 20: @dchyun Should the last one go above? And then you could import the types together without repeating "type".

Like:
```import {
  HdsStepperStatusesValues,
  HdsStepperTitleTagValues,
  HdsStepperStatusToSrOnlyText,
} from '../types.ts';

import type {
  HdsStepperStatuses,
  HdsStepperTitleTags,
  HdsStepperListStepIds,
} from '../types.ts';

File: packages/components/src/components/hds/stepper/nav/index.hbs [hunk 46]
```diff
@@ -0,0 +1,76 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<div
+  class={{this.classNames}}
+  ...attributes
+  {{style --progress-bar-width=this.progressBarWidthStyle}}
```
KristinLBradley, line 8: Perhaps you could set inline styles similarly to how it's done here:

* HBS: https://github.com/hashicorp/design-system/blob/9194be0b1382b1297a772a30d97e5b5c88c8bfec/packages/components/src/components/hds/layout/grid/index.hbs#L10
* TS: https://github.com/hashicorp/design-system/blob/9194be0b1382b1297a772a30d97e5b5c88c8bfec/packages/components/src/components/hds/layout/grid/index.ts#L84

packages/components/src/components/hds/stepper/nav/panel.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+export default class HdsStepperNavPanel extends Component<HdsStepperNavPanelSignature> {
+  private _panelId = 'panel-' + guidFor(this);
+  private _elementId?: string;
+
+  private _setUpPanel = modifier(
+    (
+      element: HTMLElement,
+      [insertCallbackFunction, destoryCallbackFunction]
```
KristinLBradley, line 35:
```suggestion
      [insertCallbackFunction, destroyCallbackFunction]
```

Typo

packages/components/src/components/hds/stepper/nav/panel.ts
```diff
@@ -0,0 +1,87 @@
...
+    ) => {
+      if (typeof insertCallbackFunction === 'function') {
+        insertCallbackFunction(element);
+      }
+
+      return () => {
+        if (typeof destoryCallbackFunction === 'function') {
```
KristinLBradley, line 42:
```suggestion
        if (typeof destroyCallbackFunction === 'function') {
```

packages/components/src/components/hds/stepper/nav/panel.ts
```diff
@@ -0,0 +1,87 @@
...
+          destoryCallbackFunction(element);
```
KristinLBradley, line 43:
```suggestion
          destroyCallbackFunction(element);
```

KristinLBradley: Check for others in case I missed some.

packages/components/src/components/hds/stepper/nav/step.hbs
```diff
@@ -0,0 +1,87 @@
...
+        <Hds::Text::Body class="hds-stepper-nav__step-title" @tag={{this.titleTag}} @size="200" @weight="semibold">
+          {{yield to="title"}}
+          <span class="sr-only">{{this.statusSrOnlyText}}</span>
+        </Hds::Text::Body>
+        <Hds::Text::Body
+          class="hds-stepper-nav__step-description"
+          @tag="div"
+          @size="100"
+          @weight="regular"
+          @color="regular"
```
KristinLBradley, line 40: `regular` isn't one of the possible values for color:  
 https://helios.hashicorp.design/components/text#textbody

packages/components/src/components/hds/stepper/nav/step.hbs
```diff
@@ -0,0 +1,87 @@
...
```
KristinLBradley, line 78: Here too. "regular" isn't one of the color values.

https://helios.hashicorp.design/components/text#textbody

packages/components/src/components/hds/stepper/list/step.hbs
```diff
@@ -0,0 +1,32 @@
...
+    <Hds::Text::Body class="hds-stepper-list__step-title" @tag={{this.titleTag}} @size="200" @weight="semibold">
+      {{yield to="title"}}
+      <span class="sr-only">{{this.statusSrOnlyText}}</span>
+    </Hds::Text::Body>
+    <Hds::Text::Body
+      class="hds-stepper-list__step-description"
+      @tag="div"
+      @size="100"
+      @weight="regular"
+      @color="regular"
```
KristinLBradley, line 24: "regular" isn't one of the color values.

https://helios.hashicorp.design/components/text#textbody

packages/components/src/components/hds/stepper/nav/step.ts
```diff
@@ -0,0 +1,186 @@
...
+import { action } from '@ember/object';
+import { modifier } from 'ember-modifier';
+
+import {
+  HdsStepperTitleTagValues,
+  HdsStepperNavStatusesValues,
+  HdsStepperNavStatusToIndicatorStatus,
+  HdsStepperNavStatusToSrOnlyText,
+} from '../types.ts';
+import type {
```
KristinLBradley, line 17:
```suggestion

import type {
```

packages/components/src/components/hds/stepper/nav/step.ts
```diff
@@ -0,0 +1,186 @@
...
+}
+
+export default class HdsStepperNavStep extends Component<HdsStepperNavStepSignature> {
+  private _stepId = 'step-' + guidFor(this);
+  private _elementId?: string;
+
+  private _setUpStep = modifier(
+    (
+      element: HTMLElement,
+      [insertCallbackFunction, destoryCallbackFunction]
```
KristinLBradley, line 55:
```suggestion
      [insertCallbackFunction, destroyCallbackFunction]
```

KristinLBradley: There's more below so just do a search & replace.

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,113 @@
...
+// STEPPER > LIST
+//
+
+$hds-stepper-list-step-padding-bottom: 16px;
+
+.hds-stepper-list,
+.hds-stepper-list__step {
+  margin: 0;
+  padding: 0;
+  list-style: none;
```
KristinLBradley, line 16:
```suggestion
  list-style-type: none;
```

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// SUB COMPONENT: STEPPER > LIST > STEP
+
+.hds-stepper-list__step {
+  display: grid;
+  grid-column-gap: 8px;
```
KristinLBradley, line 23:
```suggestion
  gap: 8px;
```

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,113 @@
...
+      &::after {
+        position: absolute;
+        top: 1px;
+        left: 11px;
+        z-index: -1;
+        width: 2px;
+        content: "";
+      }
+
+      &::before {
```
KristinLBradley, line 40: It would be helpful to label these pseudo element styles with comments explaining what they are used for

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,113 @@
...
+
+// SUB COMPONENT: STEPPER > LIST > STEP
+
+.hds-stepper-list__step {
+  display: grid;
+  grid-column-gap: 8px;
+  grid-template-columns: auto 1fr;
+
+  &:not(:last-of-type) {
+    padding-bottom: $hds-stepper-list-step-padding-bottom;
```
KristinLBradley, line 27: If you use flexbox & gap, you can remove this.

(Btw if not using flexbox or CSS grid + `gap` for spacing, using `margin` is preferable to `padding` in most cases for setting spacing between elements.)

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,113 @@
...
+      &::after {
+        background-color: var(--token-color-palette-neutral-700);
+      }
+    }
+  }
+}
+
+.hds-stepper-list__step-progress {
+  position: relative;
+  z-index: 0;
```
KristinLBradley, line 54: It may be more explicit to use `isolation: isolate;` instead of `z-index: 0;`.

https://developer.mozilla.org/en-US/docs/Web/CSS/isolation

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,364 @@
...
+  padding: 0;
+  background-color: transparent;
+  border: none;
+}
+
+.hds-stepper-nav__step-progress {
+  position: absolute;
+  top: -14px;
+  left: 0;
+  z-index: 0;
```
KristinLBradley, line 78: Try using `isolation: isolate;` instead.

https://developer.mozilla.org/en-US/docs/Web/CSS/isolation

KristinLBradley: I tried just removing the z-index: 0 though (without replacing it) after adding a box around the stepper nav and wasn't seeing a difference.

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,364 @@
...
+  color: var(--token-color-foreground-faint);
+  text-decoration: none;
+}
+
+.hds-stepper-nav__step-description {
+  margin-top: 4px;
+  color: var(--token-color-foreground-faint);
+}
+
+// STAUSES
```
KristinLBradley, line 102:
```suggestion
// STATUSES
```

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,364 @@
...
+      content: none;
+    }
+  }
+
+  .hds-stepper-nav__list {
+    flex-direction: column;
+  }
+
+  .hds-stepper-nav__step:not(:last-of-type) {
+    margin-bottom: 16px;
```
KristinLBradley, line 352: If you use a flex layout for `.hds-stepper-nav__list` then you can just set `gap: 16px` so you could remove this.

showcase/app/styles/showcase-pages/stepper/nav.scss
```diff
@@ -0,0 +1,16 @@
...
+ */
+
+// STEPPER::NAV
+
+body.components-stepper-nav {
+  .shw-component-stepper-nav-sample-form-btn {
+    margin-top: 12px;
+  }
+
+  .shw-component-stepper-nav-mock-incomplete::after {
```
KristinLBradley, line 13: Could you add a comment explaining what this is for?

showcase/app/templates/components/stepper/nav.hbs
```diff
@@ -0,0 +1,464 @@
...
+    </SG.Item>
+  </Shw::Grid>
+
+  <Shw::Text::H3>Nav::Panel</Shw::Text::H3>
+
+  <Hds::Stepper::Nav::Panel>
+    <Shw::Placeholder @text="Generic content" @height="100" />
+  </Hds::Stepper::Nav::Panel>
+
+  <Shw::Text::H2>Examples implementations</Shw::Text::H2>
```
KristinLBradley, line 261:
```suggestion
  <Shw::Text::H2>Example implementations</Shw::Text::H2>
```

showcase/app/templates/components/stepper/nav.hbs
```diff
@@ -0,0 +1,464 @@
...
+{{page-title "Stepper Component"}}
+
+<Shw::Text::H1>Stepper::Nav</Shw::Text::H1>
+
+<section data-test-percy>
+
+  <Shw::Text::H2>Interactivity</Shw::Text::H2>
+
+  <Shw::Text::Body>Interactive</Shw::Text::Body>
+
```
KristinLBradley, line 15: I would use `<Shw::Grid @columns={{1}} as |SG|>` with SG.Item labels for these. Then you can add whatever spacing seems good between them easily and use a consistent text style for label text such as "Interactive" & "Non-interactive"

KristinLBradley: Also reformat the examples below to use `Shw::Grid` etc.

KristinLBradley (commented):
@dchyun I went through all the latest code except integration tests and added comments & questions.

I'll go through the integration tests Thurs morning.

showcase/tests/integration/components/hds/stepper/list-step-test.js
```diff
@@ -0,0 +1,167 @@
...
+          </Hds::Stepper::List::Step>
+        `
+    );
+    assert.dom('.hds-stepper-list__step-title').hasTagName('div');
+  });
+
+  test('it renders the custom title tag when the @titleTag argument is provided', async function (assert) {
+    await render(
+      hbs`
+          <Hds::Stepper::List::Step @titleTag="h2">
```
KristinLBradley, line 107: Btw, it looks like `@titleTag` was included in the API docs for the Stepper::List but not for the `Stepper::List::Step`: https://hds-website-git-stepper-docs-hashicorp.vercel.app/components/stepper/list?tab=code#component-api

KristinLBradley: It should be documented for both if it's an argument on both of them.

packages/components/src/styles/components/stepper/nav.scss
dchyun: Converted this to `isolation: isolate;`. It's needed when there is a background color on the page level, as without it the completed progress bar gets hidden due to it having `z-index: -1`. The `z-index: -1` is needed to get the line to go behind the step icons.

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,364 @@
...
+  }
+
+  .hds-stepper-nav__step-progress {
+    top: -12px;
```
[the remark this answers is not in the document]
dchyun: When working with @majedelass this was the intended design, as the desktop version also uses negative positioning for the step icons to place them half above the bounding box. @majedelass are we still okay with this approach?

showcase/tests/integration/components/hds/stepper/list-step-test.js
dchyun: The `@titleTag` argument on the parent `Stepper::List` is meant to set the title tag for all step items. This ensures that all of them align with their tags. The `@titleTag` argument on the child `Stepper::List::Step` is used just to take in this argument from the parent. 

Since we do not want to encourage users to mix and match tags across steps, this argument has been hidden from the docs. This follows the [Accordion docs](https://helios.hashicorp.design/components/accordion?tab=code#component-api) where the `titleTag` argument is shown for the parent, but hidden for the accordion item.

packages/components/src/styles/components/stepper/nav.scss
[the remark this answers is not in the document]
KristinLBradley: I'm just worried that this could cause layout issues for consumers when they try to use it together with other content on a page. (I focused on the Mobile layout but it could be an issue for desktop as well.)

showcase/tests/integration/components/hds/stepper/list-step-test.js
KristinLBradley: I see, thanks for explaining.

showcase/tests/integration/components/hds/stepper/nav-panel-test.js
```diff
@@ -0,0 +1,68 @@
...
+
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+// NOTICE
+// Because of how the `panel` subcomponent is built,
+// it's practically impossible to test in isolation, so in our tests
+// in this file it will be wrapped inside its parent comoponent.
```
KristinLBradley, line 14:
```suggestion
// in this file it will be wrapped inside its parent component.
```

packages/components/src/styles/components/stepper/nav.scss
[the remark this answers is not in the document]
KristinLBradley: Actually, since in the Desktop view there's the progress bar line and the icons don't stick out as far above it (plus they are centered) I think the layout works ok for that view since spacing with other content makes sense to set as relative to the progress bar line. For mobile view though, it does still seem potentially problematic to me.

Here you can see that the negative spacing makes the icon look too close to the label text above the Stepper:

image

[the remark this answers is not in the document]
KristinLBradley: This shows more clearly how the negative icon positioning overlaps the margin (orange area) of the label:

image

packages/components/src/styles/components/stepper/list.scss
```diff
@@ -0,0 +1,106 @@
...
+      &::after {
+        background-color: var(--token-color-palette-neutral-700);
+      }
+    }
+  }
+}
+
+.hds-stepper-list__step-progress {
+  position: relative;
+  isolation: isolate;
```
dchyun, line 60: @KristinLBradley here the isolation style

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,364 @@
...
+
+@use "sass:map";
+@use "../../mixins/focus-ring" as *;
+@use "./step-indicator.scss" as *;
+
+$progress-bar-animation-duration: .25s;
+
+.hds-stepper-nav {
+  position: relative;
+  isolation: isolate;
```
dchyun, line 18: @KristinLBradley here is the isolation style

packages/components/src/styles/components/stepper/nav.scss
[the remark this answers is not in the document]
majedelass: We could take one of two approaches here:
Since this is for the Stepper Nav's responsive view we could either keep it as is, and rely on the guidance that spacing is needed between this element and other elements on the page [as seen here](https://hds-website-git-stepper-docs-hashicorp.vercel.app/components/stepper/nav#spacing), or (and maybe this is the safest route), apply a padding in the responsive view at the top (+12px) so it doesn't overlap other content on the page.

Considering that this new in the sense that no other components behave this way, we should probably apply a 12px padding to the container of the Stepper Nav in the responsive view. Thoughts?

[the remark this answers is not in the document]
majedelass: Only thing to consider: If we decide to change this, then the component will appear to be differently aligned in a responsive view than it does in the desktop view. Are we ok with that? The more I think about it, the more I'm inclined to believe we should keep it as is. If y'all have strong opinions about this (ie: the guidance isn't enough for spacing), then we can take Dylan's approach in the responsive view.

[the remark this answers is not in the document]
KristinLBradley: Would others from @hashicorp/hds-design like to add feedback on this as well?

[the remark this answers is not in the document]
KristinLBradley: We may also want to consider using a container query vs. a media query for the responsive behavior (if that wasn't already considered): https://github.com/hashicorp/design-system/pull/2714/files#r2006208620

packages/components/src/styles/components/stepper/nav.scss
```diff
@@ -0,0 +1,364 @@
...
+          }
+        }
+      }
+    }
+  }
+}
+
+// MOBILE
+
+@media screen and (width <= 550px) {
```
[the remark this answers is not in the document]
dchyun: > The total width of the Stepper Nav should match the page content width.

We do provide guidance in the [docs](https://hds-website-git-stepper-docs-hashicorp.vercel.app/components/stepper/nav#resizing-behavior) that the stepper nav should only be used in a full page width context, but not opposed to using container queries if we want to go that way.

packages/components/src/styles/components/stepper/nav.scss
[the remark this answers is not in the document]
majedelass: Yeah we can just remove the absolute positioning from this. I think that's a good path forward.

packages/components/src/styles/components/stepper/nav.scss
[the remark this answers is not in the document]
KristinLBradley: Thanks for linking to the relevant docs.

Pages also usually include the `SideNav` but on some rare occasions (or in some product apps) may not so I was just thinking that using a container query would allow us to be more precise as to when the layout of the Stepper shifted. It's just a suggestion though and we could always look at it again at a later date when/if we focus more on strategies for supporting responsiveness more generally.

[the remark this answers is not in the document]
dchyun: @majedelass do you have a preference here? I tested out container queries and it works just fine. The only difference is if we go with `550px` as the container width, the actual screen size breakpoint at which mobile would occur would be a little wider, based on what padding / other elements a page has.

If we do go with this do we want to try a new number or stick with 550?

dchyun:
@majedelass @KristinLBradley I implemented the mobile design updates we discussed earlier. I tried to match the old design as closely as possible and made a few choices for now outlined below. Let me know if we are okay with these or want changes.

Mobile design decisions (so far) from implementation 
- Padding above and below a step is 8px
- Margin between step indicator and title is 4px
- Margin between two steps is 0
- Border radius for hover / focus state is 0

KristinLBradley:
> @majedelass @KristinLBradley I implemented the mobile design updates we discussed earlier. I tried to match the old design as closely as possible and made a few choices for now outlined below. Let me know if we are okay with these or want changes.
> 
> Mobile design decisions (so far) from implementation
> 
> * Padding above and below a step is 8px
> * Margin between step indicator and title is 4px
> * Margin between two steps is 0
> * Border radius for hover / focus state is 0

@dchyun Thanks for making those updates. The top padding of the first step looks like it's adding too much space above the Stepper now though. Also the padding around the Stepper indents it from other content like the "Interactive" label shown here above it.

What do you think though? @majedelass 

image

image

image

packages/components/src/styles/components/stepper/nav.scss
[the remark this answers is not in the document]
didoo: I would be super-careful about using container queries for this component that works at such high-level in the page. Container queries create a new stacking context (`z-index`, to simplify) and it already "bit" us [time ago](https://github.com/hashicorp/design-system/blob/main/packages/components/src/styles/components/page-header.scss#L11-L13) with the `PageHeader` (and that is a self-contained component) while this one contains other content, which may need to use `z-index` values that are "global" (with the new stacking context would be local).

(I see you're using `isolation: isolate` which also creates a new stacking context, but not at top level, only at navigation level)

majedelass:
> > @majedelass @KristinLBradley I implemented the mobile design updates we discussed earlier. I tried to match the old design as closely as possible and made a few choices for now outlined below. Let me know if we are okay with these or want changes.
> > Mobile design decisions (so far) from implementation
> > 
> > * Padding above and below a step is 8px
> > * Margin between step indicator and title is 4px
> > * Margin between two steps is 0
> > * Border radius for hover / focus state is 0
> 
> @dchyun Thanks for making those updates. The top padding of the first step looks like it's adding too much space above the Stepper now though. Also the padding around the Stepper indents it from other content like the "Interactive" label shown here above it (that could be the intent though but just thought I should ask).
> 
> What do you think though? @majedelass
> 
> image image image

Agreed, maybe remove the 8px padding for the first step? I think once you do that it should be good to go.

packages/components/src/styles/components/stepper/nav.scss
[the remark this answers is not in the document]
dchyun: The `z-index` values used in the component itself are only in relation to the component context (hence the `isolate`), but the stepper body could have any generic content from consumers that may rely on the global stacking context. 

I think for now it's safer to keep with media queries. Maybe we can revisit it if use cases come up that could benefit from container queries. Are you good with that @KristinLBradley?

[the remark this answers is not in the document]
KristinLBradley: It's fine to leave as is. We could always revisit this in the future if we come up with a more defined strategy around where/when to use container queries vs. media queries as a team.

KristinLBradley:
If the design is looking as expected in the Percy snapshot, then you can approve to clear that check.

KristinLBradley (approved):
Thanks for making all the updates from feedback so quickly. Looks good!

PR #2715, opened 2025-02-19
Linting support for hds-code-editor modifier and component

zamoore:
### :pushpin: Summary

If merged, this PR will add support for linting to the hds-code-editor modifier and Hds::CodeEditor component.

- Linting is currently available only for the JSON language.
- Linting can be enabled by setting `language` to `json` & `isLintingEnabled` to `true`. 
- The editor accepts a new `onLint` argument which accepts a function that is called when linting runs and receives an array of CodeMirror [Diagnostic](https://codemirror.net/docs/ref/#lint.Diagnostic) objects.

Screenshot 2025-03-07 at 2 51 00 PM

### :link: External links

Jira ticket: [HDS-4361](https://hashicorp.atlassian.net/browse/HDS-4361)
Figma file: [if it applies]

***

### 👀 Component checklist

[HDS-4361]: https://hashicorp.atlassian.net/browse/HDS-4361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/modifiers/hds-code-editor/themes/hds-dark-theme.ts
```diff
@@ -65,6 +74,150 @@ const hdsDark = EditorView.theme(
...
+    '.cm-lint-marker-error': {
+      width: '24px',
+      height: '24px',
+      padding: '6px',
+      content: `url('data:image/svg+xml,<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><g id="x-diamond-fill"><path id="Path" fill-rule="evenodd" clip-rule="evenodd" d="M9.56912 0.752289C8.69045 -0.126408 7.26581 -0.126415 6.38713 0.752272L0.730298 6.40911C-0.148388 7.28779 -0.148379 8.71243 0.730316 9.5911L6.38715 15.2478C7.26583 16.1265 8.69044 16.1265 9.56911 15.2478L15.2258 9.59109C16.1045 8.71241 16.1045 7.2878 15.2258 6.40912L9.56912 0.752289ZM5.21967 5.21955C5.51256 4.92665 5.98744 4.92665 6.28033 5.21955L8 6.93922L9.71967 5.21955C10.0126 4.92665 10.4874 4.92665 10.7803 5.21955C11.0732 5.51244 11.0732 5.98731 10.7803 6.28021L9.06066 7.99988L10.7803 9.71955C11.0732 10.0124 11.0732 10.4873 10.7803 10.7802C10.4874 11.0731 10.0126 11.0731 9.71967 10.7802L8 9.06054L6.28033 10.7802C5.98744 11.0731 5.51256 11.0731 5.21967 10.7802C4.92678 10.4873 4.92678 10.0124 5.21967 9.71955L6.93934 7.99988L5.21967 6.28021C4.92678 5.98731 4.92678 5.51244 5.21967 5.21955Z" fill="${encodeURIComponent(HDS_CODE_EDITOR_COLOR_FOREGROUND_CRITICAL)}"/></g></svg>');`,
```
dchyun, line 102:
```suggestion
      content: `url('data:image/svg+xml,<svg width="16" height="16" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path fill-rule="evenodd" clip-rule="evenodd" d="M9.56912 0.752289C8.69045 -0.126408 7.26581 -0.126415 6.38713 0.752272L0.730298 6.40911C-0.148388 7.28779 -0.148379 8.71243 0.730316 9.5911L6.38715 15.2478C7.26583 16.1265 8.69044 16.1265 9.56911 15.2478L15.2258 9.59109C16.1045 8.71241 16.1045 7.2878 15.2258 6.40912L9.56912 0.752289ZM5.21967 5.21955C5.51256 4.92665 5.98744 4.92665 6.28033 5.21955L8 6.93922L9.71967 5.21955C10.0126 4.92665 10.4874 4.92665 10.7803 5.21955C11.0732 5.51244 11.0732 5.98731 10.7803 6.28021L9.06066 7.99988L10.7803 9.71955C11.0732 10.0124 11.0732 10.4873 10.7803 10.7802C10.4874 11.0731 10.0126 11.0731 9.71967 10.7802L8 9.06054L6.28033 10.7802C5.98744 11.0731 5.51256 11.0731 5.21967 10.7802C4.92678 10.4873 4.92678 10.0124 5.21967 9.71955L6.93934 7.99988L5.21967 6.28021C4.92678 5.98731 4.92678 5.51244 5.21967 5.21955Z" fill="${encodeURIComponent(HDS_CODE_EDITOR_COLOR_FOREGROUND_CRITICAL)}"/></svg>');`,
```
[Nit] These ids and groups are not needed. In the past, I've ran into a11y issues using ids in inline svgs. It should be fine for a url svg, but can remove just to clean up a little.

packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
```diff
@@ -0,0 +1,148 @@
...
+    });
+
+    onLint?.(diagnostics);
+
+    return diagnostics;
+  });
+
+  return [
+    jsonLinter,
+    lintGutter(),
```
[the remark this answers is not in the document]
zamoore: When you say gutter, do you mean the linting gutter (the bar to the side of the line numbers) or the drawer that you can open up with the list of errors?

showcase/package.json
```diff
@@ -122,5 +123,7 @@
   },
   "ember": {
     "edition": "octane"
+  },
+  "dependencies": {
```
shleewhite, line 127: Nit: should probably remove this since its empty.

packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
```diff
@@ -0,0 +1,148 @@
...
+      message = HdsCodeEditorJsonLintingError.ValueExpected;
+    }
+  }
+
+  return message;
+}
+
+// this renders the error message for both the tooltip and the drawer item
+export function renderErrorMessage(message: string): HTMLElement {
+  const wrapper = document.createElement('div');
```
shleewhite, line 75: Issue: the tooltips are currently not accessible and I'm not seeing an easy way to improve it in [their docs](https://codemirror.net/docs/ref/#lint)...

The issues are:
- the trigger for the tooltip is not a button, so you can't navigate to them with a keyboard.
- the tooltip also only shows up on hover, it should also appear on focus
- there is no markup connecting the button to the tooltip, it should have aria-controls and aria-describedby set to the id of the tooltip

One issue that we can fix in this PR is to add alt text for each icon. Currently there's no text, its just the icon. The text could be like "Syntax error on line {line number}" or something like that.

For the other issues, it may be best to just mark the component as not compliant if you use linting and file an issue with CodeMirror.

packages/components/src/components/hds/code-editor/index.hbs
```diff
@@ -56,11 +56,13 @@
       ariaLabel=@ariaLabel
       ariaLabelledBy=this.ariaLabelledBy
       hasLineWrapping=@hasLineWrapping
+      isLintingEnabled=@isLintingEnabled
```
shleewhite, line 59: Issue: there's not a way for users to know what the keyboard shortcut is to open the drawer. At a minimum, we should add some screen reader only text between the header and the editor what the shortcut is and what it does.

Maybe @LilithJames-HDS could weigh in, but it would be nice if the instructions were available to sighted users too.

packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
shleewhite: Ah one more thing: the icon is currently showing up in the a11y tree as an image. This can be fixed by adding `aria-hidden="true"` to the element with the class `cm-lint-marker-error`

Screenshot 2025-03-14 at 10 46 05 AM

packages/components/src/components/hds/code-editor/index.hbs
shleewhite: Thinking about it more, this is especially important because the drawer is more accessible than the tooltips - so we want to make it super easy to know how to open it.

packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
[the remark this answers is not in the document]
zamoore: We addressed the first issue by ensuring the linting panel displays below the editor.

For the second issue, I'm not sure there's a solution. There's not an event we can hook into when the linting panel closes. I'm open to some other suggestions on how we can go about solving this.

packages/components/src/components/hds/code-editor/index.hbs
alex-ju: Agree the shortcut should be surfaced to screen reader user – I remember this being [discussed during the kick-off as well](https://docs.google.com/document/d/1u27M2Qc8-uWQoZ9D0nz_k-UO_PUma8S13rlQR4kAWvU/edit?tab=t.efihpzui45kb#bookmark=id.5mwmbqde5n7r)

packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
alex-ju: Acknowledging the current limitations in CodeMirror6 around customizing the tooltip in a way that makes it discoverable by assistive tech, I wonder if we could lean more on the panel to provide an equivalent experience (highlighting linting issues) rather than trying to make the inline tooltip discoverable.

If each error in the panel mentions what line it refers to (e.g. "Line 2: Key must be double quoted" or "Key must be double quoted on line 2") it would surface the same information, without cluttering the editor and/or interrupting the editing flow for an assistive tech user. Would this be feasible @zamoore?

zamoore: I can do this from a technical perspective. @shleewhite What do you think?

shleewhite: Yeah, that works for me! We should add a note to the a11y tab tho that the tooltips are not conformant

packages/components/src/modifiers/hds-code-editor/themes/hds-dark-theme.ts
```diff
@@ -65,6 +74,141 @@ const hdsDark = EditorView.theme(
...
+    },
+
+    // linter diagnostics panel
+    '.cm-panel.cm-panel-lint ul': {
+      backgroundColor: 'var(--token-color-palette-neutral-500)',
+      maxHeight: '80px',
```
dchyun, line 108:
```suggestion
      maxHeight: '85px',
```
[Suggestion] A nit-picky design question but would it be possible to slightly increase this? With the current height one error is fully visible, and the second just peaks out, but not enough to really see that it's there. This isn't as much an issue when scrolling through the errors, but was noticeable when using the keyboard.

**Current**
Screenshot 2025-03-20 at 11 44 34 AM

**With 85px**
Screenshot 2025-03-20 at 12 02 16 PM

packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
dchyun: Just a note. If we're relying on the panel as the effective alternative, we'll have to fix the focus getting lost issue there or make a more general statement about the a11y limitations for linting.

packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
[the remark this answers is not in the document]
zamoore: Alright, I think I was able to solve issue 2 with a MutationObserver. Focus should now return to the editor when the linting panel closes.

packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
zamoore: I was able to fix the focus issue

packages/components/src/components/hds/code-editor/index.hbs
zamoore: I was able to prepend a `p` tag above the modified editor element with instructions. The id for this element is added to the `aria-describedby` attribute (in addition to the yielded Description component id) if linting is enabled.

packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
zamoore: I was also able to add the line number to the errors in the panel

packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -239,15 +256,47 @@ export default class HdsCodeEditorModifier extends Modifier<HdsCodeEditorSignatu
...
+      ?.setAttribute('aria-describedby', ariaDescribedByArray.join(' '));
+  }
+
+  private _createLintingDescriptionElement(): HTMLParagraphElement {
+    const element = document.createElement('p');
+
+    element.id = `lint-panel-instructions-${this.element.id}`;
+    element.classList.add('sr-only');
+    element.textContent =
+      'Press `Ctrl-Shift-m` (`Cmd-Shift-m` on macOS) to open the linting panel';
```
shleewhite, line 295:
```suggestion
      'Press `Ctrl-Shift-m` (`Cmd-Shift-m` on macOS) while focus is on the textbox to open the linting panel';
```

Suggestion: Just to be slightly more specific.

packages/components/src/components/hds/code-editor/index.hbs
shleewhite: Looks great!

packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
shleewhite: I'm still seeing the issue where focus gets lost when i close the panel in the showcase :/

packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -412,6 +505,30 @@ export default class HdsCodeEditorModifier extends Modifier<HdsCodeEditorSignatu
...
+  private _setupEditorMutationObserver() {
+    this.mutationObserver = new MutationObserver((mutations) => {
+      mutations.forEach((mutation) => {
+        mutation.removedNodes.forEach((node) => {
+          if (!(node instanceof HTMLElement)) {
+            return;
+          }
+
+          const removedNodeContainsLintPanel =
+            node.querySelector('.cm-panel-lint') !== null;
```
dchyun, line 517:
```suggestion
            node.querySelector('.cm-panels) !== null;
```
[Suggestion] Maybe this needs to be the parent class, since that's the outermost element getting remove? 🤷 
Screenshot 2025-03-20 at 5 30 11 PM

packages/components/src/modifiers/hds-code-editor/themes/hds-dark-theme.ts
zamoore: Ran it by lilith and implemented this

packages/components/src/modifiers/hds-code-editor.ts
zamoore: Just missed a function call 🤦

packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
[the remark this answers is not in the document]
dchyun: Focus looks like it's returning now! Even returns to the line you were previously on ✨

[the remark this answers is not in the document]
zamoore: Fixed now!

packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
dchyun, line 1: [Question] One thing I noticed for the panel is that it remains opens, and you can move back to the editor with it still open. You could then make changes and fix issues, and the panel would update it's errors to match.

@shleewhite do we need to add any kind of `aria-live="polite` or other indication, since the content in the panel is updating dynamically, potentially while it is open? Or is it fine as is?

dchyun (commented):
Looking great! Thanks for making all the changes. Just a couple things I spotted.

showcase/tests/integration/components/hds/code-editor/index-test.js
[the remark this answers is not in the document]
zamoore: Good call. Done.

packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
zamoore: That makes sense. Do you think we should make any changes to what is set up currently?

dchyun: Yeah I agree it probably could be more distracting than helpful. And the most common workflow is most likely what you mention where the panel is opened after typing and closed before typing again.

shleewhite: @zamoore no, you don't need to change anything.

dchyun (approved):

PR #2684, opened 2025-02-04
Code editor additional languages

zamoore:
### :pushpin: Summary

If merged, this PR adds syntax highlighting support for JavaScript and Rego to the `hds-code-editor` modifier.

### :hammer_and_wrench: Detailed description

- Implements custom Rego StreamParser
- Uses official JS parser for javascript

### :link: External links

Jira tickets:
[HDS-4452](https://hashicorp.atlassian.net/browse/HDS-4452)
[HDS-4469](https://hashicorp.atlassian.net/browse/HDS-4469)

***

### 👀 Component checklist

- [ ] Percy was checked for any visual regression
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-4452]: https://hashicorp.atlassian.net/browse/HDS-4452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HDS-4469]: https://hashicorp.atlassian.net/browse/HDS-4469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2717, opened 2025-02-20
Add missing entry point for rego language script

alex-ju:
Follow-up on #2684.

This line gets generated at build and makes the file available to consumers.

***


shleewhite (approved):
I noticed this pop up in the masked input pr too, thanks for addressing!

PR #2725, opened 2025-02-26
`Time` - Update visual style when tooltip is present (HDS-4550)

KristinLBradley:
### :pushpin: Summary

If merged, this PR:
* updates the visual style of the `Time` component to display a dotted underline when `hasTooltip` is true matching the visual style of the `RichTooltip` in order to add a visual indication of interactivity.
* fixes a Safari bug in the original `RichTooltip` styles which caused the dotted underline not to display.

**Showcase:** https://hds-showcase-git-hds-4550-update-time-visual-design-hashicorp.vercel.app/components/time

**_Chrome screenshot:_**
visual difference with tooltip vs. without

### :link: External links

* Jira ticket: [HDS-4550](https://hashicorp.atlassian.net/browse/HDS-4550)

***

### 👀 Component checklist

- ~~[ ] Percy was checked for any visual regression~~ _- Percy snapshot not included due to dynamic nature of component which triggers an infinite rendering invalidation error_

[HDS-4550]: https://hashicorp.atlassian.net/browse/HDS-4550?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/styles/components/time.scss
+++ b/packages/components/src/styles/components/time.scss
@@ -14,5 +14,7 @@
 
 // Note: The time-wrapper is only present when time is wrapped by a tooltip button
 .hds-time-wrapper {
+  // stylelint-disable-next-line property-no-vendor-prefix
+  -webkit-text-decoration: underline dotted;
   text-decoration: underline dotted;
 }
--- a/packages/components/src/styles/components/time.scss
+++ b/packages/components/src/styles/components/time.scss
@@ -14,6 +14,7 @@
 
 // Note: The time-wrapper is only present when time is wrapped by a tooltip button
 .hds-time-wrapper .hds-time {
+  // decoration style is specified separately to fix a bug in Safari causing it not to render
   text-decoration: underline;
   text-decoration-style: dotted;
 }
--- a/packages/components/src/styles/components/time.scss
+++ b/packages/components/src/styles/components/time.scss
@@ -14,7 +14,6 @@
 
 // Note: The time-wrapper is only present when time is wrapped by a tooltip button
 .hds-time-wrapper .hds-time {
-  // stylelint-disable-next-line property-no-vendor-prefix
-  -webkit-text-decoration: underline dotted; // Safari
-  text-decoration: underline dotted;
+  text-decoration: underline;
+  text-decoration-style: dotted;
 }
--- a/packages/components/src/styles/components/time.scss
+++ b/packages/components/src/styles/components/time.scss
@@ -11,3 +11,8 @@
   display: inline-flex; // remove excess space between time range elements
   gap: 2px; // add a bit of space around en dash so it doesn't touch some number characters
 }
+
+// Note: The time-wrapper is only present when time is wrapped by a tooltip button
+.hds-time-wrapper {
+  text-decoration: underline dotted;
+}
--- a/showcase/tests/acceptance/percy-test.js
+++ b/showcase/tests/acceptance/percy-test.js
@@ -169,8 +169,9 @@ module('Acceptance | Percy test', function (hooks) {
     await visit('/components/text');
     await percySnapshot('Text');
 
-    await visit('/components/time');
-    await percySnapshot('Time');
+    // Note: The dynamic nature of the Time component triggers an infinite rendering invalidation error so we are skipping this component for now.
+    // await visit('/components/time');
+    // await percySnapshot('Time');
 
     await visit('/components/toast');
     await percySnapshot('Toast');
--- a/showcase/tests/acceptance/percy-test.js
+++ b/showcase/tests/acceptance/percy-test.js
@@ -169,6 +169,9 @@ module('Acceptance | Percy test', function (hooks) {
     await visit('/components/text');
     await percySnapshot('Text');
 
+    await visit('/components/time');
+    await percySnapshot('Time');
+
     await visit('/components/toast');
     await percySnapshot('Toast');
 
--- a/.changeset/tidy-eels-doubt.md
+++ b/.changeset/tidy-eels-doubt.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`Time` - Updated visual style to display a dotted underline when the `hasTooltip` argument is true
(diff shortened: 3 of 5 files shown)


showcase/tests/acceptance/percy-test.js
```diff
@@ -169,6 +169,9 @@ module('Acceptance | Percy test', function (hooks) {
     await visit('/components/text');
     await percySnapshot('Text');
 
+    await visit('/components/time');
+    await percySnapshot('Time');
```
KristinLBradley, line 173: We made a decision to not include the Time component in these tests originally as it didn't really include defined visual styles. As this PR adds a more defined visual style now I wanted to add it but it seems to have wiped out the other snapshots. Is there something I missed or did wrong?

alex-ju: I believe this to be a case where Percy failed, rather than something to do with the changes in this PR. Unfortunately, there's no way to retrigger Percy manually, so we'd either have to trust it does what we expect it to do, or push a new commit/rebase to force a new trigger.

KristinLBradley: I can try rebasing once there's something new in `main`.

KristinLBradley: Oh I just remembered another issue with visual testing of Time is that the times change which messes up the visual tests. Actually this could be causing the issue. I'll try scoping the test to a limited section and add a comment.

alex-ju: Ah, I see you're already doing it in this PR

KristinLBradley: I've tried to narrow the parts tested but it still triggers this error:
`Error: infinite rendering invalidation detected`

I think that unfortunately the Time component just can't be easily tested with Percy. Unless there's a way to force it to render once and then stop updating. I just don't have a good idea on a practical way to do it currently.

packages/components/src/styles/components/time.scss
```diff
@@ -10,4 +10,9 @@
 .hds-time--range {
   display: inline-flex; // remove excess space between time range elements
   gap: 2px; // add a bit of space around en dash so it doesn't touch some number characters
+}
+
+// Note: The time-wrapper is only present when time is wrapped by a tooltip button
+.hds-time-wrapper {
```
didoo, line 16: [nit] if you want to be even more specific
```suggestion
.hds-time-wrapper .hds-time {
```

shleewhite (dismissed):
Thanks for updating this!

packages/components/src/styles/components/time.scss
```diff
@@ -10,4 +10,9 @@
 .hds-time--range {
   display: inline-flex; // remove excess space between time range elements
   gap: 2px; // add a bit of space around en dash so it doesn't touch some number characters
+}
+
+// Note: The time-wrapper is only present when time is wrapped by a tooltip button
+.hds-time-wrapper .hds-time {
+  text-decoration: underline dotted;
```
MelSumner, line 17: recommendation: add `text-underline-offset` with a little bit of a value to make it easier to read. When reviewing the showcase, it felt like it really needed that.

didoo: should we do it in a separate ticket, so we can update also the `RichTooltip` decoration too? https://hds-showcase.vercel.app/components/rich-tooltip#interactions

KristinLBradley: I would want to update the style for both the Time component with tooltip and the `RichTooltip` at the same time as @didoo suggested.

If we change the placement then I think we also need more weigh in from @hashicorp/hds-design

KristinLBradley: Btw, in Chrome it already looks to me like it has a bit of space closer to the "After" pic above.

image

heatherlarsen: +1, Chrome looks like the screenshot Kristin shared. @MelSumner What browser are you using?

MelSumner: CleanShot 2025-02-28 at 14 08 40@2x

didoo: here how they look across different browsers (for me):

**chrome**
chrome

**safari**
safari

**edge**
edge

**firefox**
firefox

@MelSumner could it be some setting or CSS override in your browser/environment?

@KristinLBradley in Safari/Edge the underline was not visible, I suspect your declaration is not valid, better double check.

KristinLBradley: Thanks, I thought I had copied the CSS from the Rich Tooltip but I will double check.

KristinLBradley: @didoo so this is the line of CSS I used to add the underline which I copied from the RichTooltip: `text-decoration: underline dotted;` which it looks like you had added to the RicheTooltip originally.

I checked out the Showcase page in Edge and I am seeing the text decoration.

**Edge:**
image

However I am not seeing the underline/text decoration in Safari so I will investigate further. (I think you may have switched the labels on your screenshots accidentally.)

KristinLBradley: Looks like Safari requires a -web-kit- prefix:
`-webkit-text-decoration: underline dotted;`

heatherlarsen: Thanks, y'all, for investigating. If we add a slight offset (2-3px), I think that would be slightly more readable.

KristinLBradley: Should I use this PR to add `-webkit-text-decoration: underline dotted;` to the RichTooltip styles as well? Or open a separate PR?

heatherlarsen: A separate PR works for me so we can address them at the same time.

KristinLBradley: Thanks everyone! I'm going ahead and fixing the Safari bug for both the `Time` and `RichTooltip` components in this PR. I will create a new ticket/PR for exploring visual style updates however.

heatherlarsen (approved):
Looks great! Thanks for this fix 🙂

didoo (approved):
👍

packages/components/src/styles/components/time.scss
KristinLBradley: PR of visual updates to dotted line spacing: 
https://github.com/hashicorp/design-system/pull/2745

PR #2726, opened 2025-02-27
`hds-tooltip` modifier -  Fix CSS import (HDS-4542)

KristinLBradley:
### :pushpin: Summary

If merged, this PR removes the tippy.js CSS import from the tooltip modifier, imports the styles into the tooltip SCSS file for the component and adds the tippy file path to rollup.

### :link: External links

* Jira ticket: [HDS-4542](https://hashicorp.atlassian.net/browse/HDS-4542)

***

### 👀 Component checklist

[HDS-4542]: https://hashicorp.atlassian.net/browse/HDS-4542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/styles/components/tooltip.scss
+++ b/packages/components/src/styles/components/tooltip.scss
@@ -8,7 +8,6 @@
 //
 
 @use "../mixins/focus-ring" as *;
-@use '../../../node_modules/tippy.js/dist/svg-arrow.css';
 
 .hds-tooltip-button {
   @include hds-focus-ring-with-pseudo-element(
@@ -86,6 +85,55 @@
 
   // works with Tippy's custom SVG arrow variation:
   .tippy-svg-arrow {
+    width: 16px;
+    height: 16px;
+    text-align: initial;
     fill: var(--token-tooltip-color-surface-primary);
+
+    &, 
+    & > svg {
+      position: absolute;
+    }
+  }
+
+  // Arrow positioning styles taken from tippy.js (tippy.js/dist/svg-arrow.css)
+  &[data-placement^="top"] > .tippy-svg-arrow {
+    bottom: 0;
+  }
+
+  &[data-placement^="top"] > .tippy-svg-arrow::after,
+  &[data-placement^="top"] > .tippy-svg-arrow > svg {
+    top: 16px;
+    transform: rotate(180deg);
+  }
+
+  &[data-placement^="bottom"] > .tippy-svg-arrow {
+    top: 0;
+  }
+
+  &[data-placement^="bottom"] > .tippy-svg-arrow > svg {
+    bottom: 16px;
+  }
+
+  &[data-placement^="left"] > .tippy-svg-arrow {
+    right: 0;
+  }
+
+  &[data-placement^="left"] > .tippy-svg-arrow::after,
+  &[data-placement^="left"] > .tippy-svg-arrow > svg {
+    top: calc(50% - 3px);
+    left: 11px;
+    transform: rotate(90deg);
+  }
+
+  &[data-placement^="right"] > .tippy-svg-arrow {
+    left: 0;
+  }
+
+  &[data-placement^="right"] > .tippy-svg-arrow::after,
+  &[data-placement^="right"] > .tippy-svg-arrow > svg {
+    top: calc(50% - 3px);
+    right: 11px;
+    transform: rotate(-90deg);
   }
 }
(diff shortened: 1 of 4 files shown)


KristinLBradley:
Would this be a "patch" change? @alex-ju

packages/components/src/styles/components/tooltip.scss
```diff
@@ -8,6 +8,7 @@
 //
 
 @use "../mixins/focus-ring" as *;
+@use '../../../node_modules/tippy.js/dist/svg-arrow.css';
```
aklkv, line 11:
```suggestion
@use 'svg-arrow.css';
```

KristinLBradley: This is what I had initially but it was causing an error. I'll test again locally, thanks.

KristinLBradley: If I don't include the full path I get an error that the `MIME type ('text/html') is not a supported stylesheet MIME type` so I think it's 404ing.

alex-ju:
> Would this be a "patch" change? @alex-ju

patch seems appropriate. I'll have a look at the CSS path later today as well

packages/components/src/styles/components/tooltip.scss
alex-ju: double-check me on this, but `@use` is only available for `scss` files and in our case we aim to bring in a `css` file, meaning `@import` is still the way to go

```suggestion
@import 'svg-arrow.css';
```

KristinLBradley: @alex-ju  Ohh thanks for catching the `@import` vs `@use`. I will experiment locally.

aklkv: I think my original suggestion stands, I will double check but here is how I did it for core in cloud-ui and it just works (we need it for our old tooltip implementation) https://github.com/hashicorp/cloud-ui/blob/a1819ffd21e1ff18c0541b41b59d27c13167ed86/addons/core/package/src/styles/core/index.scss#L2

KristinLBradley: @alex-ju It doesn't seem to have to do with the order. I'm not seeing the placement styles at all it seems when I inspect.

Missing example style:
```
.tippy-box[data-placement^=right] > .tippy-svg-arrow {
    left: 0;
}
```

KristinLBradley: @aklkv I'd be happy to follow your original suggestion but as I said for some reason it's not working and results in an error. I'll try to investigate more.

aklkv: ah I think I see what is the issue might be, you are trying  to import it from deeply nested stylesheet, usual practice in sass is to have vendor imports in the root stylesheet but I will double check

aklkv: actually I figured it out, because `showcase` consumes uncompiled sass files it tries to figure out where to import `svg-arrow.css` but it doesn't know the paths. this presents interesting question if we move forward with this change outside consumer will need to switch to use compiled variant `.css`

aklkv: here is a working example https://github.com/hashicorp/design-system/pull/2734
but as I said we would need to communicate to all consumer to use compiled `.css` files

aklkv: I have been contemplating this a bit more and here is where I am at:
- The goal of this PR is to fix `cloud-ui` CSP restriction for local dev, which is okayish if it's not fixed, preferable but not critical.
- If we were to do it correctly, we would need to have more in-depth revision on what is exposed for consumers to use. From what I have seen, a lot use `.scss` still and having path-dependent imports is simply not going to work, so the location of node modules can differ from project to project.
- We can either mandate to consumers to only use `.css` or keep it as is on the main branch for now and revisit this issue at a later point. I do not see the reason why end users should need to use main .scss files if we keep mixins etc. still available and main CSS just in .css (in order for this to work correctly, we would need to communicate in the next release that, preferably major, that main stylesheet is only available in CSS format).

KristinLBradley: @aklkv Thanks for outlining those options. I'm thinking at this point perhaps it would be simpler to just copy the arrow positioning styles over to ours. 

@alex-ju What do you think?

KristinLBradley: I went ahead and just copied the styles over directly.

alex-ju (approved):

aklkv (approved):
🚀

PR #2729, opened 2025-02-27
feat(website): modernize dynamic-template component

aklkv:
### :pushpin: Summary

Modernize website dynamic-template component

### :link: External links

Based on: https://github.com/empress/field-guide/pull/53/files#diff-a2cab16146e173caa300a356bfa92eaa89e5228d1cc1179fc1b16d1da298ab70

Context: https://hashicorp.slack.com/archives/C089K6P5TFF/p1740635217880569

***

### 👀 Component checklist

### DIFF
--- a/website/app/components/dynamic-template.js
+++ b/website/app/components/dynamic-template.js
@@ -1,21 +1,14 @@
-/* eslint-disable ember/no-classic-components, ember/no-classic-classes, prettier/prettier, no-console */
-
-import Component from '@ember/component';
-import GlimmerComponent from '@glimmer/component';
-import layout from './dynamic-template';
-
-import { computed } from '@ember/object';
+import { setComponentTemplate } from '@ember/component';
 import { getOwner } from '@ember/application';
 import { compileTemplate } from '@ember/template-compilation';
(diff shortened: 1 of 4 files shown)


website/app/components/dynamic-template.js
```diff
@@ -1,61 +1,68 @@
...
 
 let templateOwnerMap = new Map();
-let templateId = 0;
 
-export default Component.extend({
-  tagName: '',
-  layout,
-  init() {
-    this._super(...arguments);
+export default class DynamicTemplate extends Component {
```
didoo, line 9: [nit] to keep the implementation in sync ;)

```suggestion
export default class DynamicTemplateComponent extends Component {
```

didoo (approved):
👍

website/app/components/dynamic-template.js
aklkv: Actually this is now antipattern, because in the future you would be importing things like you name them eg.g. `<HdsAlert>` backing class has to be named `HdsAlert` not `HdsAlertComponent` otherwise it wold look strange

PR #2733, opened 2025-02-28
feat(components): exclude `app-header` components

aklkv:
### :pushpin: Summary

Since this components were excluded from `appReexport` it actually confuses `embroider/vite` and breaks the build. We can either exclude them completely including disabling some parts of showcase or we keep them in the `appReexport`

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist

### DIFF
--- a/packages/components/rollup.config.mjs
+++ b/packages/components/rollup.config.mjs
@@ -14,15 +14,22 @@ const addon = new Addon({
   destDir: 'dist',
 });
 
+const exclude = ['components/hds/app-header', 'components/hds/app-side-nav'];
+
 const plugins = [
   // These are the modules that users should be able to import from your
   // addon. Anything not listed here may get optimized away.
-  addon.publicEntrypoints([
-    '**/*.{js,ts}',
-    'index.js',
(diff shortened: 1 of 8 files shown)


alex-ju (commented):
The change makes sense to me. However, we'd need to work around [a mock component](https://github.com/hashicorp/design-system/blob/27a4bff1c9f46b4e8c389f14d1d73a278188646c/showcase/app/components/mock/app/header/app-header.gts#L10) that uses these exports to emulate a full-page layout. The `Mock::App::Header` component itself is not being used right now, so it can be commented out, but the imported type implications may cascade.

alex-ju (approved):
Looking good! We'll still need to keep MockApp (without the AppHeader bits) otherwise [the last example on this page](https://hds-showcase-9rjvrebka-hashicorp.vercel.app/layouts/app-frame) [will not render](https://hds-showcase-ivxitng2p-hashicorp.vercel.app/layouts/app-frame). I pushed a commit for that – hope you don't mind – but please check.

didoo (approved):

PR #2738, opened 2025-03-04
Fix Vercel alias creation when running from Actions

alex-ju:
### :pushpin: Summary

In #2658 we introduced a GitHub workflow to automate the creation of a Vercel alias for new releases. At the time we tested the result of this workflow by triggering it in the PR, but not by running the GitHub Action.

It turns out we need a different logic to extract the branch name depending on where we run the action.

***


MelSumner (approved):
👍

dchyun (approved):

PR #2732, opened 2025-02-27
Release candidates

alex-ju:
### :pushpin: Summary

This PR introduces a GitHub Action workflow to support us with publishing release candidates.

#### How it works

The GitHub action will be dispatched automatically when a PR gets the `release-candidate` label.

If the current version is `4.16.0` the release candidate would be `4.16.1-rc-20250227130741` or `4.17.0-rc-20250227130741` (depending on the changes).

A comment will be added to the PR when the GitHub action completes, with instructions on how to use the release candidate when testing in products. Note that we default to yarn instructions as most products are currently using yarn.

#### Credits and docs

This approach is inspired by how https://github.com/hashicorp/web-platform-packages manages snapshot releases and has been adapted to our current setup and needs.

[Changesets' snapshot release documentation](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md)

#### Follow-up

[Our current guidance](https://hashicorp.atlassian.net/wiki/x/24Fx0Q) will be updated to include this option for testing.

### :link: External links

Jira ticket: [HDS-4391](https://hashicorp.atlassian.net/browse/HDS-4391)

***

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-4391]: https://hashicorp.atlassian.net/browse/HDS-4391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2741, opened 2025-03-05
Attempt to fix the changeset version step for RC

alex-ju:
### :pushpin: Summary

When [testing the Release Candidate workflow](https://github.com/hashicorp/design-system/pull/2689#event-16596793734) we discovered that [the changeset version command did not run as expected](https://github.com/hashicorp/design-system/actions/runs/13673237611/job/38227910905?pr=2689#step:7:1).

We suspect this is because the `version` and `publish` inputs are not being executed as separate commands. Instead, they are being passed as arguments to the `changesets/action`, which is probably not the intended behavior.

Attempting to fix this we separate the version and publish steps into two distinct steps in our workflow.

Follow-up on #2732.

***

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,9 +3,9 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "release-candidate-packages": "changeset publish --tag rc",
+    "release-candidate-packages": "pnpm changeset publish --tag rc",
     "release-packages": "pnpm changeset publish",
-    "version-candidate-packages": "pnpm changeset version --snapshot rc",
+    "version-candidate-packages": "pnpm changeset version --snapshot rc && pnpm install --lockfile-only",
     "version-packages": "pnpm changeset version && pnpm install --lockfile-only && pnpm -F website generate-changelog-markdown-files"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -3,9 +3,9 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "release-candidate-packages": "changeset publish --tag rc",
+    "release-candidate-packages": "pnpm changeset publish --tag rc",
     "release-packages": "pnpm changeset publish",
-    "version-candidate-packages": "pnpm changeset version --snapshot rc",
+    "version-candidate-packages": "pnpm changeset version --snapshot rc && pnpm install --lockfile-only",
```
alex-ju, line 8: not required, but for symmetry with the `version-packages` script below

package.json
```diff
@@ -3,9 +3,9 @@
   "version": "0.0.0",
   "private": true,
   "scripts": {
-    "release-candidate-packages": "changeset publish --tag rc",
+    "release-candidate-packages": "pnpm changeset publish --tag rc",
```
alex-ju, line 6: not required, but for symmetry with the `release-packages` script below

dchyun (approved):

PR #2746, opened 2025-03-05
Showcase - Add min-height to `Shw::Label`

didoo:
### :pushpin: Summary

Simple PR to add a `min-height` to the `Shw::Label` in the showcase, so it doesn't collapse when empty.

***

### 👀 Component checklist

### DIFF
--- a/showcase/app/styles/showcase-components/label.scss
+++ b/showcase/app/styles/showcase-components/label.scss
@@ -9,6 +9,7 @@
 
 .shw-label {
   @include shw-font-family("rubik");
+  min-height: 0.96rem; // needed to avoid vertical collapsing when the label is empty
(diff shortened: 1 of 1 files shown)


showcase/app/styles/showcase-components/label.scss
```diff
@@ -9,6 +9,7 @@
 
 .shw-label {
   @include shw-font-family("rubik");
+  min-height: 0.96rem; // needed to avoid vertical collapsing when the label is empty
```
KristinLBradley, line 12: Why so specific?

didoo:
I'm moving the PR to draft because there are too many failing visual regression tests in Percy, and I want to think a bit more about potential alternative solutions.

PR #2751, opened 2025-03-06
`Flex` - Component implementation

didoo:
### :pushpin: Summary

TODO

TODO

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-4635
Project plan: [HDS Project Plan - Flex/Grid components](https://docs.google.com/document/d/1SIW_vOCEYkvK6U5XkRVSVwGyhh6hR3XY5o4RdaH2g-Q)

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/components/hds/layout/flex/index.hbs
+++ b/packages/components/src/components/hds/layout/flex/index.hbs
@@ -1,0 +1,8 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+{{! ADD YOUR TEMPLATE CONTENT HERE }}
+<div class={{this.classNames}} ...attributes>
+  {{yield}}
+</div>
--- a/packages/components/src/components/hds/layout/flex/index.hbs
+++ b/packages/components/src/components/hds/layout/flex/index.hbs
@@ -2,7 +2,8 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-{{! ADD YOUR TEMPLATE CONTENT HERE }}
-<div class={{this.classNames}} ...attributes>
-  {{yield}}
-</div>
+{{#if (eq this.componentTag "div")}}
+  <div class={{this.classNames}} ...attributes>{{yield}}</div>
+{{else}}
+  {{#let (element this.componentTag) as |Tag|}}<Tag class={{this.classNames}} ...attributes>{{yield}}</Tag>{{/let}}
+{{/if}}
--- a/packages/components/src/components/hds/layout/flex/index.hbs
+++ b/packages/components/src/components/hds/layout/flex/index.hbs
@@ -3,7 +3,7 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 {{#if (eq this.componentTag "div")}}
-  <div class={{this.classNames}} ...attributes>{{yield}}</div>
+  <div class={{this.classNames}} ...attributes>{{yield (hash Item=(component "hds/layout/flex/item"))}}</div>
 {{else}}
-  {{#let (element this.componentTag) as |Tag|}}<Tag class={{this.classNames}} ...attributes>{{yield}}</Tag>{{/let}}
+  {{#let (element this.componentTag) as |Tag|}}<Tag class={{this.classNames}} ...attributes>{{yield (hash Item=(component "hds/layout/flex/item"))}}</Tag>{{/let}}
 {{/if}}
--- a/packages/components/src/components/hds/layout/flex/item.ts
+++ b/packages/components/src/components/hds/layout/flex/item.ts
@@ -1,0 +1,81 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+
+// A list of all existing tag names in the HTMLElementTagNameMap interface
+type AvailableTagNames = keyof HTMLElementTagNameMap;
+// A union of all types in the HTMLElementTagNameMap interface
+type AvailableElements = HTMLElementTagNameMap[keyof HTMLElementTagNameMap];
+
+export interface HdsLayoutFlexItemSignature {
+  Args: {
+    tag?: AvailableTagNames;
+    basis?: string;
+    grow?: boolean | number | string;
+    shrink?: boolean | number | string;
+  };
+  Blocks: {
+    default: [];
+  };
+  Element: AvailableElements;
+}
+
+export default class HdsLayoutFlexItem extends Component<HdsLayoutFlexItemSignature> {
+  get componentTag(): AvailableTagNames {
+    return this.args.tag ?? 'div';
+  }
+
+  get inlineStyles(): Record<string, unknown> {
+    const inlineStyles: {
+      'flex-basis'?: string;
+      'flex-grow'?: string;
+      'flex-shrink'?: string;
+    } = {};
+
+    // we handle all cases of `basis` values via inline styles
+    if (this.args.basis) {
+      inlineStyles['flex-basis'] = this.args.basis;
+    }
+
+    // we handle non-standard cases of `grow` values via inline styles
+    if (typeof this.args.grow === 'number' && this.args.grow > 1) {
+      // the `{{style}}` modifier accepts only strings
+      inlineStyles['flex-grow'] = this.args.grow.toString();
+    } else if (typeof this.args.grow === 'string') {
+      inlineStyles['flex-grow'] = this.args.grow;
+    }
+
+    // we handle non-standard cases of `shrink` values via inline styles
+    if (typeof this.args.shrink === 'number' && this.args.shrink > 1) {
+      // the `{{style}}` modifier accepts only strings
+      inlineStyles['flex-shrink'] = this.args.shrink.toString();
+    } else if (typeof this.args.shrink === 'string') {
+      inlineStyles['flex-shrink'] = this.args.shrink;
+    }
+
+    return inlineStyles;
+  }
+
+  get classNames() {
+    const classes = ['hds-layout-flex-item'];
+
+    // add a class based on the @grow argument (if set to `0/1` or `true/false`)
+    if (this.args.grow === 0 || this.args.grow === false) {
+      classes.push('hds-layout-flex-item--grow-0');
+    } else if (this.args.grow === 1 || this.args.grow === true) {
+      classes.push('hds-layout-flex-item--grow-1');
+    }
+
+    // add a class based on the @shrink argument (if set to `0/1` or `true/false`)
+    if (this.args.shrink === 0 || this.args.shrink === false) {
+      classes.push('hds-layout-flex-item--shrink-0');
+    } else if (this.args.shrink === 1 || this.args.shrink === true) {
+      classes.push('hds-layout-flex-item--shrink-1');
+    }
+
+    return classes.join(' ');
+  }
+}
--- a/packages/components/src/components/hds/layout/flex/types.ts
+++ b/packages/components/src/components/hds/layout/flex/types.ts
@@ -9,3 +9,24 @@ export enum HdsLayoutFlexDirectionValues {
 }
 
 export type HdsLayoutFlexDirections = `${HdsLayoutFlexDirectionValues}`;
+
+export enum HdsLayoutFlexJustifyValues {
+  Start = 'start',
+  Center = 'center',
+  End = 'end',
+  SpaceBetween = 'space-between',
+  SpaceAround = 'space-around',
+  SpaceEvenly = 'space-evenly',
+}
+
+export type HdsLayoutFlexJustifys = `${HdsLayoutFlexJustifyValues}`;
+
+export enum HdsLayoutFlexAlignValues {
+  Start = 'start',
+  Center = 'center',
+  End = 'end',
+  Stretch = 'stretch',
+}
+
+export type HdsLayoutFlexAligns = `${HdsLayoutFlexAlignValues}`;
+
--- a/packages/components/src/components/hds/layout/flex/types.ts
+++ b/packages/components/src/components/hds/layout/flex/types.ts
@@ -1,0 +1,11 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+export enum HdsLayoutFlexDirectionValues {
+  Row = 'row',
+  Column = 'column',
+}
+
+export type HdsLayoutFlexDirections = `${HdsLayoutFlexDirectionValues}`;
--- a/packages/components/src/styles/components/layout/flex.scss
+++ b/packages/components/src/styles/components/layout/flex.scss
@@ -26,3 +26,9 @@
 .hds-layout-flex--direction-column {
   flex-direction: column;
 }
+
+// wrap
+
+.hds-layout-flex--has-wrapping {
+  flex-wrap: wrap;
+}
--- a/packages/components/src/styles/components/layout/flex.scss
+++ b/packages/components/src/styles/components/layout/flex.scss
@@ -80,3 +80,22 @@
 .hds-layout-flex--align-items-stretch {
   align-items: stretch;
 }
+
+
+// FLEX ITEM
+
+.hds-layout-flex-item--grow-0 {
+  flex-grow: 0;
+}
+
+.hds-layout-flex-item--grow-1 {
+  flex-grow: 1;
+}
+
+.hds-layout-flex-item--shrink-0 {
+  flex-shrink: 0;
+}
+
+.hds-layout-flex-item--shrink-1 {
+  flex-shrink: 1;
+}
--- a/packages/components/src/styles/components/layout/flex.scss
+++ b/packages/components/src/styles/components/layout/flex.scss
@@ -32,3 +32,51 @@
 .hds-layout-flex--has-wrapping {
   flex-wrap: wrap;
 }
+
+// justify
+
+.hds-layout-flex--justify-content-start {
+  justify-content: start;
+}
+
+.hds-layout-flex--justify-content-center {
+  justify-content: center;
+}
+
+.hds-layout-flex--justify-content-end {
+  justify-content: end;
+}
+
+.hds-layout-flex--justify-content-space-between {
+  justify-content: space-between;
+}
+
+.hds-layout-flex--justify-content-space-around {
+  justify-content: space-around;
+}
+
+.hds-layout-flex--justify-content-space-evenly {
+  justify-content: space-evenly;
+}
+
+.hds-layout-flex--justify-content-stretch {
+  justify-content: stretch;
+}
+
+// align
+
+.hds-layout-flex--align-items-start {
+  align-items: start;
+}
+
+.hds-layout-flex--align-items-center {
+  align-items: center;
+}
+
+.hds-layout-flex--align-items-end {
+  align-items: end;
+}
+
+.hds-layout-flex--align-items-stretch {
+  align-items: stretch;
+}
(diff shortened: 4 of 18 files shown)


packages/components/src/components/hds/layout/flex/index.hbs
```diff
@@ -0,0 +1,11 @@
...
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+{{#if (eq this.componentTag "div")}}
+  <div class={{this.classNames}} ...attributes>{{yield (hash Item=(component "hds/layout/flex/item"))}}</div>
+{{else}}
+  {{#let (element this.componentTag) as |Tag|}}<Tag class={{this.classNames}} ...attributes>{{yield
+        (hash Item=(component "hds/layout/flex/item"))
+      }}</Tag>{{/let}}
+{{/if}}
```
KristinLBradley, line 11: @didoo Since the default tag is already a "div" (specified in the JS), does it need to be specified in the template here also using the "if/else" statements? That seems duplicative unless there's a reason I'm not aware of.

didoo: I have noticed that when the the tag is generated dynamically, Ember adds a `ember-view` class to it, which is usually the one used for "components", while for the `<div>` default one it doesn't. 

I suspect that when it's dynamic, it's more javascript intense, is some kind of component, while in the other case it's just an HTML div. But I am making an educated guess, I'm not sure.

I have noticed this pattern here: https://github.com/prysmex/ember-eui/blob/a6ebc29d0c2b501472514d9dfff334a884a2e85b/packages/core/src/components/eui-flex-grid.gts#L43-L70

didoo: @KristinLBradley let me know if you think is better to write a comment about it in the code (for future reference)

packages/components/src/components/hds/layout/flex/item.ts
```diff
@@ -0,0 +1,90 @@
...
+      'flex-grow'?: string;
+      'flex-shrink'?: string;
+    } = {};
+
+    // we handle all non-zero cases of `basis` values via inline styles
+    if (typeof this.args.basis === 'string') {
+      inlineStyles['flex-basis'] = this.args.basis;
+    }
+
+    // we handle non-standard cases of `grow` values via inline styles
```
didoo, line 42: @KristinLBradley let me know what you think of this double approach, partially based on CSS class names for the most common use cases, and on inline styles for the non-standard cases

packages/components/src/components/hds/layout/flex/types.ts
```diff
@@ -0,0 +1,49 @@
+/**
```
didoo, line 1: @KristinLBradley have a look at the values that I've associated to each flex prop, and let me know if you think we're missing/we should add others

showcase/app/templates/layouts/flex.hbs
```diff
@@ -0,0 +1,495 @@
...
+          <SG.Item
+            @label={{if (eq jc 0) (concat "align=" align) "​"}}
+            class="shw-layout-flex-example-outline-flex-container shw-layout-flex-example-tint-flex-items"
+          >
+            <div {{style width="120px" height="120px"}}>
+              <Hds::Layout::Flex
+                @direction={{direction}}
+                @justify={{justify}}
+                @align={{align}}
+                {{style width="100%" height="100%"}}
```
didoo, line 97: @KristinLBradley this is one of those cases I mentioned (but there are others in this file) where having `@width` and `@height` arguments directly on the `Flex` component would come in handy

packages/components/src/components/hds/layout/flex/index.hbs
KristinLBradley: Could be helpful to add a brief comment to explain

packages/components/src/components/hds/layout/flex/item.ts
```diff
@@ -0,0 +1,90 @@
...
+import type { AvailableTagNames, AvailableElements } from './types.ts';
+
+export interface HdsLayoutFlexItemSignature {
+  Args: {
+    tag?: AvailableTagNames;
+    basis?: string | 0;
+    grow?: boolean | number | string;
+    shrink?: boolean | number | string;
+    // TODO final name TBD
+    enableCollapseBelowContentSize?: boolean;
```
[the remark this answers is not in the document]
KristinLBradley: What are the possible use cases & do they seem common?

This seems like a confusing feature that perhaps could be left up to consumers to implement something for.

packages/components/src/components/hds/layout/flex/item.ts
KristinLBradley: Being able to pass "false/true" for the values seems to be helpful & easier to understand/use for consumers so I think that's a worthwhile approach to allow for either.

packages/components/src/components/hds/layout/flex/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  'Thirtytwo' = '32',
+  'Fortyeight' = '48',
+}
+
+export type HdsLayoutFlexGaps = `${HdsLayoutFlexGapValues}`;
+
+// A list of all existing tag names in the HTMLElementTagNameMap interface
+export type AvailableTagNames = keyof HTMLElementTagNameMap;
+// A union of all types in the HTMLElementTagNameMap interface
+export type AvailableElements = HTMLElementTagNameMap[keyof HTMLElementTagNameMap];
```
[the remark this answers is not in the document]
KristinLBradley: I think it's fine for the use case.

(Saying "justifys", for example, may be weird in English but here I think it's ok and probably clearer than trying to come up with another name like "justifications" or something...)

KristinLBradley (commented):
I think this is looking good overall. I replied to a couple comments.

Still reviewing the features more in detail...

packages/components/src/components/hds/layout/flex/index.hbs
didoo: @DingoEatingFuzz @aklkv before adding the comment in the code and write something which is not true, can you confirm my hunch that makes sense to add this `if` condition and why is that? thanks 🙏

packages/components/src/components/hds/layout/flex/item.ts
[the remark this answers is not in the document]
didoo: Examples of where it's used in our HDS components:
- https://github.com/hashicorp/design-system/blob/main/packages/components/src/styles/components/page-header.scss#L46
- https://github.com/hashicorp/design-system/blob/main/packages/components/src/styles/components/app-footer.scss#L36
- https://github.com/hashicorp/design-system/blob/main/packages/components/src/styles/components/side-nav/content.scss#L153

packages/components/src/components/hds/layout/flex/index.hbs
didoo: 🙇‍♂️

packages/components/src/components/hds/layout/flex/item.ts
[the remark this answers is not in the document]
KristinLBradley: Thanks, I'm more wondering if there are common use cases for _not_ wanting to set it to true.

packages/components/src/components/hds/layout/flex/types.ts
KristinLBradley: I think you've included a good range of options that seem helpful and not overly complex for consumers to understand and make use of.

showcase/app/templates/layouts/flex.hbs
KristinLBradley: I think that width and height properties could be helpful on layout components like these.

(I don't think that we would need to add them to other HDS components as you mentioned since the other components either aren't focused around layout or have more standardized dimensions.)

didoo:
> There's a few errors from the linter. Would you mind running a fix on them? (I'm getting the errors in my branches that point to this one too.)

@KristinLBradley fixed now. Sorry 🙏

packages/components/src/components/hds/layout/flex/index.hbs
didoo: Added a comment about this.

packages/components/src/components/hds/layout/flex/types.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+export enum HdsLayoutFlexDirectionValues {
```
dchyun, line 6: [Question] Would it be useful to add `row-reverse` and `column-reverse` as options here?

packages/components/src/components/hds/layout/flex/item.ts
```diff
@@ -0,0 +1,90 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+
+import type { AvailableTagNames, AvailableElements } from './types.ts';
+
+export interface HdsLayoutFlexItemSignature {
+  Args: {
```
dchyun, line 11: [Question] Would it be useful to add in an `@order` argument for the `order` css property?

showcase/app/templates/layouts/flex.hbs
```diff
@@ -0,0 +1,525 @@
...
+      </Hds::Layout::Flex>
+    </SF.Item>
+    <SF.Item
+      @label="one item at the end"
+      class="shw-layout-flex-example-outline-flex-container shw-layout-flex-example-tint-flex-items"
+    >
+      <Hds::Layout::Flex @gap="16" @tag="section">
+        <Hds::Text::Body @size="200" @tag="p" {{style padding="4px 8px"}}>I am on the left side</Hds::Text::Body>
+        <Hds::Text::Body @size="200" @tag="p" {{style padding="4px 8px"}}>I am on the left side too</Hds::Text::Body>
+        <Hds::Text::Body @size="200" @tag="p" {{style padding="4px 8px"}} {{style margin-left="auto"}}>I on the right
```
dchyun, line 518: [Suggestion] It could be useful to have an argument on the `Flex::Item` that could cover this, if this is a common use case. If we look into adding a `@margin` arg in the future, this could be covered there. Or could have some separate arg for setting auto-margins similar to how [bootstrap](https://getbootstrap.com/docs/5.3/utilities/flex/#auto-margins) handles it through classes.

packages/components/src/components/hds/layout/flex/types.ts
```diff
@@ -0,0 +1,49 @@
...
+  SpaceEvenly = 'space-evenly',
+}
+
+export type HdsLayoutFlexJustifys = `${HdsLayoutFlexJustifyValues}`;
+
+export enum HdsLayoutFlexAlignValues {
+  Start = 'start',
+  Center = 'center',
+  End = 'end',
+  Stretch = 'stretch',
```
dchyun, line 28:
```suggestion
  Stretch = 'stretch',
  Baseline = 'baseline',
```
[Question] Would it be useful to add `baseline` as an option? I don't see it being used in cloud ui at all, but looking at the audit there are ~20 instances of it in atlas.

dchyun (commented):
This is looking really great, and covers a lot of use cases. I added a couple suggestions that would cover some additional flex box features. If we are trying to keep the component less complex for consumers then may not need the features I'm suggesting.

packages/components/src/components/hds/layout/flex/types.ts
dchyun: Disregard this comment. Just saw the note in the layout docs draft about not using reverse. 😅

packages/components/src/components/hds/layout/flex/item.ts
didoo: Changing the order of elements in the DOM can have (negative) accessibility implications, as @MelSumner told me many times :) so we don't expose it.

showcase/app/templates/layouts/flex.hbs
didoo: We've opened a discussion about this in the slack channel https://hashicorp.slack.com/archives/C08HCQZTL13/p1741966996285359 but didn't have much traction, so with @KristinLBradley we've decided to discuss it with the HDS Engineers as follow-up incremental change.

packages/components/src/components/hds/layout/flex/types.ts
didoo: The `baseline` alignment is very tricky to understand, so better to keep things simple

packages/components/src/components/hds/layout/flex/index.hbs
```diff
@@ -0,0 +1,15 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+{{!
+  Dynamically generating an HTML tag in Ember creates a dynamic component class (with the corresponding tagName), while rendering
+  a plain HTML elements requires less computing cycles for Ember (you will notice it doesn't add the `ember-view` class to it).
```
KristinLBradley, line 7:
```suggestion
  a plain HTML element requires less computing cycles for Ember (you will notice it doesn't add the `ember-view` class to it).
```

packages/components/src/styles/components/layout/flex.scss
```diff
@@ -0,0 +1,140 @@
...
+//
+// LAYOUT > FLEX
+//
+
+.hds-layout-flex {
+  display: flex;
+
+  // Note: The gap style is defined here to avoid setting it repeatedly for the gap size variants defined below.
+  // For the gap size variants, we override the value of the gap custom properties to set the desired sizes.
+  // These variables can be used by the consumers as escape hatch if they need to set non-standard gap values (but adds a bit of friction to it)
```
KristinLBradley, line 15:
```suggestion
  // These variables can be used by the consumers as an escape hatch if they need to set non-standard gap values (but adds a bit of friction to it)
```

packages/components/src/styles/components/layout/flex.scss
```diff
@@ -0,0 +1,140 @@
...
+  align-items: end;
+}
+
+.hds-layout-flex--align-items-stretch {
+  align-items: stretch;
+}
+
+// gap
+
+// set just row gap size
```
KristinLBradley, line 93:
```suggestion
// set row gap size
```

packages/components/src/styles/components/layout/flex.scss
```diff
@@ -0,0 +1,140 @@
...
+.hds-layout-flex--row-gap-4 { --hds-layout-flex-row-gap: 4px; }
+.hds-layout-flex--row-gap-8 { --hds-layout-flex-row-gap: 8px; }
+.hds-layout-flex--row-gap-12 { --hds-layout-flex-row-gap: 12px; }
+.hds-layout-flex--row-gap-16 { --hds-layout-flex-row-gap: 16px; }
+.hds-layout-flex--row-gap-24 { --hds-layout-flex-row-gap: 24px; }
+.hds-layout-flex--row-gap-32 { --hds-layout-flex-row-gap: 32px; }
+.hds-layout-flex--row-gap-48 { --hds-layout-flex-row-gap: 48px; }
+
+// set just column gap size
```
KristinLBradley, line 102:
```suggestion
// set column gap size
```

KristinLBradley (commented):
I think this is looking quite good overall. What do you think about adding `@width` and `@height` options though?

Or perhaps those could be considered together with break points & responsiveness in a phase 2 release? (I guess we already discussed doing these as a follow-on, sorry.)

packages/components/src/styles/components/layout/flex.scss
didoo: I am going to remove the comment at all, it's redundant at this point

packages/components/src/styles/components/layout/flex.scss
didoo: I am going to remove the comment at all, it's redundant at this point

packages/components/src/components/hds/layout/flex/item.ts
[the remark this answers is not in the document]
KristinLBradley: Any more thoughts around this one?

KristinLBradley (approved):

showcase/app/templates/layouts/flex.hbs
```diff
@@ -0,0 +1,525 @@
...
+        </Hds::Layout::Flex>
+      </div>
+    </SG.Item>
+    <SG.Item
+      @label="first flex item has basis=0px"
+      class="shw-layout-flex-example-outline-flex-container shw-layout-flex-example-tint-flex-items"
+    >
+      <div {{style max-width="600px"}}>
+        <Hds::Layout::Flex as |HLF|>
+          <HLF.Item @basis="0px">
```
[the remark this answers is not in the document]
didoo: It's to show that it can't be solved using `basis: 0px`, it's necessary to use `min-width: 0` to have the flex item not overstretch beyond its expected size

showcase/tests/integration/components/hds/layout/flex/index-test.js
```diff
@@ -0,0 +1,168 @@
...
+
+  // JUSTIFY / ALIGN / WRAP / IS-INLINE
+
+  test('it should render the element without specific classes if no @justify/@align/@wrap/@isInline are declared', async function (assert) {
+    await render(hbs`<Hds::Layout::Flex id="test-layout-flex" />`);
+    assert
+      .dom('#test-layout-flex')
+      .doesNotHaveClass(/hds-layout-flex--justify-content-/)
+      .doesNotHaveClass(/hds-layout-flex--align-items-/)
+      .doesNotHaveClass('.hds-layout-flex--has-wrapping')
```
dchyun, line 77:
```suggestion
      .doesNotHaveClass('hds-layout-flex--has-wrapping')
```

showcase/tests/integration/components/hds/layout/flex/index-test.js
```diff
@@ -0,0 +1,168 @@
...
+      .doesNotHaveClass('.hds-layout-flex--is-inline');
```
dchyun, line 78:
```suggestion
      .doesNotHaveClass('hds-layout-flex--is-inline');
```

showcase/tests/integration/components/hds/layout/flex/item-test.js
```diff
@@ -0,0 +1,177 @@
...
+  // ENABLE COLLAPSE BELOW CONTENT SIZE
+
+  test('it should render the element without specific classes if no @enableCollapseBelowContentSize is declared', async function (assert) {
+    await render(
+      hbs`<Hds::Layout::Flex as |LF|><LF.Item id="test-layout-flex-item" /></Hds::Layout::Flex>`
+    );
+    assert
+      .dom('#test-layout-flex-item')
+      .doesNotHaveClass(
+        '.hds-layout-flex-item--enable-collapse-below-content-size'
```
dchyun, line 166:
```suggestion
        'hds-layout-flex-item--enable-collapse-below-content-size'
```

dchyun (commented):
Looks all good to me. Just found a couple small things in the test suite.

showcase/tests/integration/components/hds/layout/flex/index-test.js
KristinLBradley: Thanks for catching these

KristinLBradley (approved):

dchyun (approved):

packages/components/src/components/hds/layout/flex/item.ts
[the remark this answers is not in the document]
shleewhite: Has this conversation been resolved?

shleewhite (approved):

packages/components/src/components/hds/layout/flex/item.ts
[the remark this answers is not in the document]
KristinLBradley: Leaving the name as is. We couldn't think of a better one when we discussed. Thanks for checking.

PR #2754, opened 2025-03-07
docs: add a title example to icon component page

shleewhite:
### :pushpin: Summary

I noticed that there is no example for adding an accessible name to an icon on the icon component page. I think it would be useful to surface this information more visibly to consumers.

Screenshot 2025-03-07 at 12 02 02 PM

***


MelSumner (changes requested):
This is on purpose. Icons have `aria-hidden="true"` on them by default. 

An icon is not supposed to be used alone in our design system. It should be used in an another context and that context is supposed to provide the accessible name.

PR #2757, opened 2025-03-09
chore(components): update `eslint` & `prettier` setup

aklkv:
### :pushpin: Summary

It addresses two things
- aligns to blueprint eslint/prettier
- as a result of above addon is fully ready to start writing gts files whenever team is ready. This changes are coming to app it self as part of the next LTS as well but thanks to pnpm dependency isolation seems to work great we can gradually upgrade workspaces one by one 

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist


alex-ju (commented):
The proposed changes in terms of setup look good overall, aligning us with the ember blueprint.

The main reason I'm not in a position to approve yet, is the `prettier` bump – 3.5 is more opinionated on Sass to a point where a single change (the 'Break before breaking comma separated values', poorly implemented in my opinion) results in things like inline comments being misplaced, messing up the readability of the code. And it doesn't seem to provide a way to disable this rule via config.

I'm going to highlight a few such changes that I'm not fond of:

Screenshot 2025-03-12 at 10 35 19
Screenshot 2025-03-12 at 10 37 50
Screenshot 2025-03-12 at 10 38 48

I appreciate prettier 3.5 also comes with support for TypeScript configuration file, which is great, so I see the tradeoffs we'll have to make.

I will bring this PR to engineers attention during our sync today and aim to come up with a resolution.

alex-ju (approved):
Tried my best to preserve those inline comments in CSS to no avail, so I re-positioned them manually in [3bca98a](https://github.com/hashicorp/design-system/pull/2757/commits/3bca98a080939fec1b0dd182bbb3af7f9d0ce049) in a way that hopefully makes sense to everyone. Comments are now either before the line or in line, never after the line.

didoo:
> The proposed changes in terms of setup look good overall, aligning us with the ember blueprint.
> 
> The main reason I'm not in a position to approve yet, is the `prettier` bump – 3.5 is more opinionated on Sass to a point where a single change (the 'Break before breaking comma separated values', poorly implemented in my opinion) results in things like inline comments being misplaced, messing up the readability of the code. And it doesn't seem to provide a way to disable this rule via config.

Can we find a way (eg. open a PR/request/issue upstream) to avoid messing up with the Sass format, in particular the comments? For a CSS developer comments deliver a lot of meaning to the code (and their location too). The alternative for me is to have a different automation for Sass files that is less opinionated/restrictive/destructive.

aklkv:
I can prettier ignore all scss files for now

packages/components/src/components/hds/code-block/index.ts
```diff
@@ -90,6 +90,7 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
     );
 
     if (Prism?.plugins?.['NormalizeWhitespace']) {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-return, @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
```
shleewhite, line 93: Is there a ticket already for us to address these eslint-disables later?

packages/components/src/modifiers/hds-advanced-table-cell/dom-management.ts
```diff
@@ -35,9 +34,7 @@ export const updateTabbableChildren = (
 };
 
 export const didInsertGridCell = (
-  cell:
-    | HdsAdvancedTableThSignature['Element']
-    | HdsAdvancedTableTdSignature['Element']
+  cell: HdsAdvancedTableThSignature['Element']
```
shleewhite, line 37: Was this changed because the ThSignature and TdSignature are both HTMLDivElement?

packages/components/src/components/hds/code-block/index.ts
alex-ju: We have [an entire epic](https://hashicorp.atlassian.net/browse/HDS-4656) – can break these new additions into more actionable tasks

alex-ju (dismissed):
This has been stalling for a little while. I'm happy to get it in with the latest prettier config for Sass and decide later if we want to turn it off and handle it somehow else.

I appreciate this needs a new rebase – happy to re-approve after that.

alex-ju (approved):

packages/components/src/modifiers/hds-advanced-table-cell/dom-management.ts
alex-ju: missed this one! @aklkv was this intentional? otherwise we can raise a follow-up to bring it back

aklkv: yeah it was implying identical value
Screenshot 2025-04-02 at 4 35 17 PM

shleewhite: Figured, that should be fine. Those types will always be the same.

PR #2761, opened 2025-03-11
AdvancedTable: set checkbox column width to `min-content`

shleewhite:
### :pushpin: Summary

If merged, this PR would fix an issue the explorer table had when converting to AdvancedTable. Basically, because every column in explorer has a set pixel width, if the user hides many columns then the checkbox column grows to fill the available space.

This PR also adds an example to the showcase that mimics the explorer use case.

**Before**
Screenshot 2025-03-11 at 4 07 27 PM

**After**
Screenshot 2025-03-11 at 4 18 38 PM

***

### 👀 Component checklist

### DIFF
--- a/.changeset/purple-poems-battle.md
+++ b/.changeset/purple-poems-battle.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+AdvancedTable - always set the select checkbox column width to `50px` so it does not grow when the AdvancedTable narrower than the container
--- a/.changeset/purple-poems-battle.md
+++ b/.changeset/purple-poems-battle.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-AdvancedTable - always set the select checkbox column width to `min-content` so it does not grow when the AdvancedTable is narrower than the container
+`AdvancedTable` - always set the select checkbox column width to `min-content` so it does not grow when the `AdvancedTable` is narrower than the container
--- a/.changeset/purple-poems-battle.md
+++ b/.changeset/purple-poems-battle.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-AdvancedTable - always set the select checkbox column width to `50px` so it does not grow when the AdvancedTable narrower than the container
+AdvancedTable - always set the select checkbox column width to `min-content` so it does not grow when the AdvancedTable narrower than the container
--- a/.changeset/purple-poems-battle.md
+++ b/.changeset/purple-poems-battle.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-AdvancedTable - always set the select checkbox column width to `min-content` so it does not grow when the AdvancedTable narrower than the container
+AdvancedTable - always set the select checkbox column width to `min-content` so it does not grow when the AdvancedTable is narrower than the container
--- a/.changeset/purple-poems-battle.md
+++ b/.changeset/purple-poems-battle.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-`AdvancedTable` - always set the select checkbox column width to `min-content` so it does not grow when the `AdvancedTable` is narrower than the container
+`AdvancedTable` - Always set the select checkbox column width to `min-content` so it does not grow when the `AdvancedTable` is narrower than the container
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -246,11 +246,11 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
 
     // if there is no custom column widths, each column is the same width and they take up the available space (except the select checkbox)
     if (!this.columnWidths) {
-      return `${isSelectable ? '50px ' : ''}repeat(${columns.length}, 1fr)`;
+      return `${isSelectable ? 'min-content ' : ''}repeat(${columns.length}, 1fr)`;
     }
 
-    // if there is a select checkbox, the first column is '50px' width to hug the checkbox content
-    let style = isSelectable ? '50px ' : '';
+    // if there is a select checkbox, the first column is 'min-content' width to hug the checkbox content
+    let style = isSelectable ? 'min-content ' : '';
 
     // check the custom column widths, if the current column has a custom width use the custom width. otherwise take the available space.
     for (let i = 0; i < this.columnWidths.length; i++) {
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -249,7 +249,7 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
       return `${isSelectable ? 'min-content ' : ''}repeat(${columns.length}, 1fr)`;
     }
 
-    // if there is a select checkbox, the first column is 'min-content' width to hug the checkbox content
+    // if there is a select checkbox, the first column has a 'min-content' width to hug the checkbox content
     let style = isSelectable ? 'min-content ' : '';
 
     // check the custom column widths, if the current column has a custom width use the custom width. otherwise take the available space.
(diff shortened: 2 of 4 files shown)


packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -246,11 +246,11 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
 
     // if there is no custom column widths, each column is the same width and they take up the available space (except the select checkbox)
     if (!this.columnWidths) {
-      return `${isSelectable ? 'auto ' : ''}repeat(${columns.length}, 1fr)`;
+      return `${isSelectable ? '50px ' : ''}repeat(${columns.length}, 1fr)`;
```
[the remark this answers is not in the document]
shleewhite: nice! that works, ill switch it

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -246,11 +246,11 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
     // if there is no custom column widths, each column is the same width and they take up the available space (except the select checkbox)
     if (!this.columnWidths) {
-      return `${isSelectable ? 'auto ' : ''}repeat(${columns.length}, 1fr)`;
+      return `${isSelectable ? 'min-content ' : ''}repeat(${columns.length}, 1fr)`;
     }
 
-    // if there is a select checkbox, the first column is 'auto' width to hug the checkbox content
-    let style = isSelectable ? 'auto' : '';
+    // if there is a select checkbox, the first column is 'min-content' width to hug the checkbox content
+    let style = isSelectable ? 'min-content ' : '';
```
jorytindall, line 253: Technically in Figma this is element is 48x48px, which referencing the showcase the dimensions are 49x48.

shleewhite: @jorytindall does that include the border?

jorytindall: @shleewhite good point, Figma doesn't visually represent the DOM in the same way, so no, that doesn't include the border; technically the border is "inside" of the element, which is a conceptual difference.

didoo: +1 on the border not being an issue due to how Figma treats the border (plus, in a browser the border is always outside)

.changeset/purple-poems-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+AdvancedTable - always set the select checkbox column width to `min-content` so it does not grow when the AdvancedTable is narrower than the container
```
didoo, line 5: [issue?] I think there is a specific syntax to use here for an automation tool that @alex-ju has built

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -246,11 +246,11 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
 
     // if there is no custom column widths, each column is the same width and they take up the available space (except the select checkbox)
     if (!this.columnWidths) {
-      return `${isSelectable ? 'auto ' : ''}repeat(${columns.length}, 1fr)`;
+      return `${isSelectable ? 'min-content ' : ''}repeat(${columns.length}, 1fr)`;
     }
 
-    // if there is a select checkbox, the first column is 'auto' width to hug the checkbox content
-    let style = isSelectable ? 'auto' : '';
+    // if there is a select checkbox, the first column is 'min-content' width to hug the checkbox content
```
didoo, line 252:
```suggestion
    // if there is a select checkbox, the first column has a 'min-content' width to hug the checkbox content
```

packages/components/src/components/hds/advanced-table/index.ts
didoo, line 1: [suggestion] For your consideration: while the code works correctly as is, you may want to consider a possible simplification like the one below:

```
// if there is a select checkbox, the first column is 'min-content' width to hug the checkbox content
let style = isSelectable ? 'min-content ' : '';

if (this.columnWidths) {
  // check the custom column widths, if the current column has a custom width use the custom width. otherwise take the available space.
  for (let i = 0; i < this.columnWidths.length; i++) {
    style += ` ${this.columnWidths[i] ? this.columnWidths[i] : '1fr'}`;
  }
} else {
  // if there is no custom column widths, each column is the same width and they take up the available space (except the select checkbox)
  style += `repeat(${columns.length}, 1fr)`;
}

return style;
```

(I haven't tested it in a browser, please do if you use it)

showcase/app/templates/components/advanced-table.hbs
```diff
@@ -930,6 +930,32 @@
 
   <Shw::Divider @level={{2}} />
 
+  <Shw::Text::H3>Wide container with a narrow AdvancedTable</Shw::Text::H3>
+
+  <div class="shw-component-advanced-table-narrow-wrapper">
+    <Hds::AdvancedTable
+      @isSelectable={{true}}
+      @onSelectionChange={{this.onMultiSelectUsersSelectionChange__demo3}}
+      @model={{this.model.userData}}
```
didoo, line 939: [suggestion] since this table is used only to demo the case of a narrow table with a select column, you could use a simpler data model here (see for example the "Large AdvancedTable with overflowing container") so that we don't show 39 rows, in a page which is already extremely long

showcase/app/templates/components/advanced-table.hbs
```diff
@@ -930,6 +930,32 @@
...
+      @columns={{array (hash key="id" label="ID" width="150px") (hash key="name" label="Name" width="150px")}}
+      @hasStickyHeader={{true}}
+      @isStriped={{true}}
```
didoo, line 942: [suggestion] to simplify things, you could remove these two (they're not strictly needed, for this example)
```suggestion
```

didoo (approved):
left a few non-blocking suggestions

.changeset/purple-poems-battle.md
alex-ju:
```suggestion
`AdvancedTable` - always set the select checkbox column width to `min-content` so it does not grow when the `AdvancedTable` is narrower than the container
```

packages/components/src/components/hds/advanced-table/index.ts
shleewhite: I can look into it, but this seems like a logical simplification.

shleewhite: Updated!

jorytindall (approved):
🚀

.changeset/purple-poems-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`AdvancedTable` - always set the select checkbox column width to `min-content` so it does not grow when the `AdvancedTable` is narrower than the container
```
dchyun, line 5:
```suggestion
`AdvancedTable` - Always set the select checkbox column width to `min-content` so it does not grow when the `AdvancedTable` is narrower than the container
```

dchyun (approved):
🚢

dchyun (approved):

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -244,17 +244,17 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
+    let style = isSelectable ? 'min-content ' : '';
 
-    // check the custom column widths, if the current column has a custom width use the custom width. otherwise take the available space.
-    for (let i = 0; i < this.columnWidths.length; i++) {
-      style += ` ${this.columnWidths[i] ? this.columnWidths[i] : '1fr'}`;
+    if (!this.columnWidths) {
+      // if there are no custom column widths, each column is the same width and they take up the available space
+      style += `repeat(${columns.length}, 1fr)`;
+    } else {
+      // check the custom column widths, if the current column has a custom width use the custom width. otherwise take the available space.
```
KristinLBradley, line 254:
```suggestion
      // check the custom column widths. if the current column has a custom width, use the custom width. otherwise, take the available space.
```

KristinLBradley (approved):
Looks good!

PR #2764, opened 2025-03-12
`Layout::Flex` - documentation

didoo:
_**Note**: `Grid` docs have been merged in to be released together with `Flex` docs_

----

### :pushpin: Summary

This is the documentation counterpart of the `Flex` component implementation done in https://github.com/hashicorp/design-system/pull/2751

👉 👉 👉 **Preview**: https://hds-website-git-flex-component-documentation-hashicorp.vercel.app/layouts/flex

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-4646

***

### 👀 Component checklist


website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,75 @@
...
+    HTML tag to be used to render the flexbox element.
+  </C.Property>
+  <C.Property @name="direction" @type="enum" @values={{array "row" "column"}} @default="row">
+    The value of the CSS `flex-direction` property, which sets how the flex items (children) are placed in the flex container, defining the [main axis](https://developer.mozilla.org/en-US/docs/Glossary/Main_Axis) and the direction (for a technical explanation: [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction)).
+    <br/><br/>
+    <em>Notice: we don't expose the "reverse" directions because they come with intrinsic accessibility limitations that we prefer our consumers to avoid.</em>
+  </C.Property>
+  <C.Property @name="justify" @type="enum" @values={{array "start" "center" "end" "space-between" "space-around" "space-evenly"}}>
+    The value of the CSS `justift-content` property, which defines how the space is distributed between and around content items along the [main axis](https://developer.mozilla.org/en-US/docs/Glossary/Main_Axis) of the flex container (for a technical explanation: [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content)).
```
KristinLBradley, line 15:
```suggestion
    The value of the CSS `justify-content` property, which defines how the space is distributed between and around content items along the [main axis](https://developer.mozilla.org/en-US/docs/Glossary/Main_Axis) of the flex container (for a technical explanation: [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content)).
```

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,75 @@
...
+    <br/><br/>
+    <em>Notice: we expose only a subset of the values allowed for this property, covering only the most common use cases.</em>
+    <br/><br/>
+    <em>Tip: when the `@direction` is `row` this argument controls the horizontal alignment; when it's `column` it controls the vertical alignment.</em>
+  </C.Property>
+  <C.Property @name="align" @type="enum" @values={{array "start" "center" "end" "stretch"}}>
+    The value of the CSS `align-item` property, which controls the alignment of the flex items on the [cross axis](https://developer.mozilla.org/en-US/docs/Glossary/Cross_Axis) (for a technical explanation: [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items)).
```
KristinLBradley, line 22:
```suggestion
    The value of the CSS `align-items` property, which controls the alignment of the flex items on the [cross axis](https://developer.mozilla.org/en-US/docs/Glossary/Cross_Axis) (for a technical explanation: [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items)).
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,541 @@
...
+### Tag
+
+To specify which HTML tag to use to render the flex container and/or item(s), use the `@tag` argument:
+
+```handlebars{data-execute=false}
+<Hds::Layout::Flex @tag="ul" as |LF|>
+  <li>{{! some content here }}</li>
+  <LF.Item @tag="li">
+    {{! some other content here }}
+  </HLF.Item>
```
KristinLBradley, line 33:
```suggestion
  </LF.Item>
```

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,78 @@
...
+    <br/><br/>
+    <em>Notice: we expose only a subset of the values allowed for this property, covering only the most common use cases.</em>
+    <br/><br/>
+    <em>Tip: when the `@direction` is `row` this argument controls the vertical alignment; when it's `column` it controls the horizontal alignment.</em>
+  </C.Property>
+  <C.Property @name="wrap" @type="boolean" @default="false">
+    If a `@wrap` parameter is provided, the flex items can wrap onto multiple lines when there is not enough space in the container to fit them all in a single line.
+  </C.Property>
+  <C.Property @name="gap" @type="enum|[enum,enum]" @values={{array "4" "8" "12" "16" "24" "32" "48"}}>
+    Use the `@gap` argument to control the spacing between flex items. To differentiate the vertical and horizontal spacing between items when they wrap on multiple rows, provide an array of two values , where the first value refers to the vertical gap between "rows" of items (`row-gap` in CSS), the second one to the horizontal spacing between "columns" of items (`column-gap` in CSS).
```
KristinLBradley, line 32:
```suggestion
    Use the `@gap` argument to control the spacing between flex items. To differentiate the vertical and horizontal spacing between items when they wrap on multiple rows, provide an array of two values, where the first value refers to the vertical gap between "rows" of items (`row-gap` in CSS), and the second one to the horizontal spacing between "columns" of items (`column-gap` in CSS).
```

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,78 @@
...
+#### [LF].Item
+
+The `Layout::Flex::Item` component, yielded as contextual component, to be used as child of the `flexbox` element to control its `basis/grow/shrink` values (and other properties).
+
+<Doc::ComponentApi as |C|>
+  <C.Property @name="tag" @type="string" @default="div">
+    HTML tag to be used to render the flex item element.
+  </C.Property>
+  <C.Property @name="basis" @type="string|0">
+    The value (size) of the CSS `flex-basis` property, which sets the initial main size the flex item (for a technical explanation: [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis)).
```
KristinLBradley, line 53:
```suggestion
    The value (size) of the CSS `flex-basis` property, which sets the initial main size of the flex item (for a technical explanation: [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis)).
```

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,78 @@
...
+    <br/><br/>
+    <em>Notice: when the value is set to `0` a CSS class is used to apply the `flex-basis` property; in all the other cases an inline `style` declaration is used, to accomodate for all the possible values that this CSS property can have.</em>
```
KristinLBradley, line 55:
```suggestion
    <em>Notice: when the value is set to `0`, a CSS class is used to apply the `flex-basis` property. In all the other cases, an inline `style` declaration is used to accommodate all the possible values that this CSS property can have.</em>
```

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,78 @@
...
+  </C.Property>
+  <C.Property @name="grow" @type="boolean|number|string">
+    The value (size or keyword) of the CSS `flex-grow` property, which sets the flex grow factor of the flex item (for a technical explanation:  [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow)).
+    <br/><br/>
+    <em>Notice: when the value is set to `true/false` or `0/1` a CSS class is used to apply the `flex-grow` property; in all the other cases an inline `style` declaration is used, to accomodate for all the possible values that this CSS property can have.</em>
```
KristinLBradley, line 60:
```suggestion
    <em>Notice: when the value is set to `true/false` or `0/1` a CSS class is used to apply the `flex-grow` property. In all other cases, an inline `style` declaration is used, to accommodate all the possible values that this CSS property can have.</em>
```

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,78 @@
...
+  </C.Property>
+  <C.Property @name="shrink" @type="boolean|number|string">
+    The value (size or keyword) of the CSS `flex-shrink` property, which sets sets the flex shrink factor of the flex item (for a technical explanation: [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink)).
+    <br/><br/>
+    <em>Notice: when the value is set to `true/false` or `0/1` a CSS class is used to apply the `flex-shrink` property; in all the other cases an inline `style` declaration is used, to accomodate for all the possible values that this CSS property can have.</em>
```
KristinLBradley, line 65:
```suggestion
    <em>Notice: when the value is set to `true/false` or `0/1` a CSS class is used to apply the `flex-shrink` property. In all other cases, an inline `style` declaration is used, to accommodate all the possible values that this CSS property can have.</em>
```

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,78 @@
...
+  </C.Property>
+  <C.Property @name="enableCollapseBelowContentSize" @type="boolean">
+    When this special argument is set to `true` it applies a `min-width: 0` to the element, allowing the flex item to shrink below its content's intrinsic minimum width (e.g. used with elliptized text)
```
KristinLBradley, line 68:
```suggestion
    When this special argument is set to `true`, it applies `min-width: 0` to the element, allowing the flex item to shrink below its content's intrinsic minimum width (e.g., used with elliptized text)
```

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,78 @@
...
+    <br/><br/>
+    <em>Notice: this may have accessibility implications, be considerate in how this special argument is used and how this impacts the content of the flex item element.</em>
```
KristinLBradley, line 70:
```suggestion
    <em>Notice: this may have accessibility implications; be considerate in how this special argument is used and how this impacts the content of the flex item element.</em>
```

KristinLBradley (commented):
Made a few minor editing suggestions but otherwise looking good!

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,562 @@
...
+<br/>
+A good and quick introduction to flexbox can be found in [**this MDN guide**](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout/Basic_concepts_of_flexbox).
+
+!!!
+
+The `Layout::Flex` and optional `Layout::Flex::Item` components provide a way to quickly build out flexbox-based layouts of components or elements without needing to write a lot of custom CSS code or understand all the intricacies of CSS flexbox styles.
+
+### Basic usage
+
+The simplest way to implementat a flexbox layout is by using the `Layout::Flex` component to wrap some content:
```
KristinLBradley, line 15:
```suggestion
The simplest way to implement a flexbox layout is by using the `Layout::Flex` component to wrap some content:
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,562 @@
...
+
+**Important**
+
+The **pre-defined value(s)** passed to the `@gap` argument **must be string(s)**, not numbers!
+
+!!!
+
+📚 Refer to the ["Recipes"](#flexbox-layout-recipes) and [Patterns](#common-layout-patterns) sections below for practical examples of how the Flexbox `gap` property can be used to implement specific layouts.
+
+#### Non-standard gap values
```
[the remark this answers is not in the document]
didoo: In some cases this will be a necessity, expecially if they need to migrate some existing layouts, so we have to give them a (well documented) way to do it.

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,562 @@
...
+</Hds::Layout::Flex>
+```
+
+How these three properties impact on the (relative) size of a flex item is not trivial, we suggest to look at the introduction to flexbox guide linked at the top of this page.
+
+📚 Refer to the ["Recipes"](#flexbox-layout-recipes) and [Patterns](#common-layout-patterns) sections below for practical examples of how the Flexbox `flex-basis`/`flex-grow`/`flex-shrink` properties can be used to implement specific layouts.
+
+---
+
+## Flexbox layout "recipes"
```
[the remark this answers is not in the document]
didoo: I hear your point. Let's see what others think and then we can make a decision.

website/docs/layouts/flex/partials/code/how-to-use.md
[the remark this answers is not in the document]
dchyun: Makes sense. @KristinLBradley thanks for adding a section like this in the grid docs.

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+<Hds::Layout::Flex>
+  <div>{{! some content here }}</div>
+  <div>{{! some other content here }}</div>
+  <div>{{! more content here }}</div>
+</Hds::Layout::Flex>
+```
+
+In this way, every direct child element of the **flexbox container** will be treated, by the browser's layout engine, as a **flex item** (for details about what this means, refer to the guide linked at the top of this page).
+
+There are cases in which is necessary to wrap one or more child elements in a specific `Layout::Flex::Item` (eg. to apply the `@basis/@grow/@shrink` arguments, see below for details):
```
KristinLBradley, line 27:
```suggestion
There are cases in which it is necessary to wrap one or more child elements in a specific `Layout::Flex::Item` (eg. to apply the `@basis/@grow/@shrink` arguments, see below for details):
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+  <LF.Item @grow={{false}}>
+    {{! some other content here }}
+  </LF.Item>
+  <div>{{! more content here }}</div>
+</Hds::Layout::Flex>
+```
+
+!!! Info
+
+Note: there is no strict need to use the `Layout::Flex::Item` subcomponent as a direct child of `Layout::Flex`; use it only when necessary to tweak flex styles of individual child item via the `@basis/@grow/@shrink` (to avoid rendering an extra Ember component).
```
KristinLBradley, line 41:
```suggestion
Note: there is no strict need to use the `Layout::Flex::Item` subcomponent as a direct child of `Layout::Flex`; use it only when necessary to tweak flex styles of an individual child item via the `@basis/@grow/@shrink` (to avoid rendering an extra Ember component).
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,562 @@
...
+<Hds::Layout::Flex @tag="ul" as |LF|>
+  <li>{{! some content here }}</li>
+  <LF.Item @tag="li">
+    {{! some other content here }}
+  </LF.Item>
+  <li>{{! more content here }}</li>
+</Hds::Layout::Flex>
+```
+
+!!! Insight
```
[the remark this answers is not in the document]
KristinLBradley: Added, thanks!

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+
+While by default the component renders a `<div>`, we invite consumers to consider which semantic HTML tag is the correct one for the context in which the text is used to meet WCAG Success Criterion [1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html) as the visual experience should match what is presented to the user with assistive technology.
```
KristinLBradley, line 61:
```suggestion
While, by default, the component renders a `<div>`, we invite consumers to consider which semantic HTML tag is the correct one for the context in which the text is used to meet WCAG Success Criterion [1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html) as the visual experience should match what is presented to the user with assistive technology.
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+
+!!!
+
+### Direction
+
+To specify which in which [direction](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction) the flex items are laid out, use the `@direction` argument:
```
KristinLBradley, line 67:
```suggestion
To specify in which [direction](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction) the flex items are laid out, use the `@direction` argument:
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+    multiple flex items here, with a non-standard gap between them
+    by assigning a custom value for `--hds-layout-flex-column-gap`
+    declared in the `.doc-flex-demo-custom-flex-column-gap` local class
+  }}
+</Hds::Layout::Flex>
+```
+
+In this case we're overwriting only the "column" gap value via the custom CSS class.
+
+If the flex items are wrapping on multiple lines, you have to overwrite both the "row" and "column" gap values:
```
KristinLBradley, line 162:
```suggestion
If the flex items are wrapping to multiple lines, you have to overwrite both the "row" and "column" gap values:
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+<Hds::Layout::Flex as |LF|>
+  ...
+  <LF.Item @basis="200px" @grow={{1}} @shrink={{0}}>
+    {{! the flex item content here }}
+  </HLF.Item>
+  ...
+</Hds::Layout::Flex>
+```
+
+How these three properties impact on the (relative) size of a flex item is not trivial, we suggest to look at the introduction to flexbox guide linked at the top of this page.
```
KristinLBradley, line 200:
```suggestion
How these three properties impact the (relative) size of a flex item is not trivial, we suggest consulting the MDN flexbox guide linked at the top of this page.
```

website/docs/layouts/flex/partials/code/how-to-use.md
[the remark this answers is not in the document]
KristinLBradley: To me, "patterns" and "recipes" seem pretty much synonymous used in this context.

Maybe you could experiment with taking some of the simpler examples from the "recipes" and instead using them as in-context examples though and see how that works?

For the Grid component though, it's a simpler component and I wanted to keep the documentation shorter and more focused. So I don't think the Flex documentation necessarily needs to follow the exact same format. Also there are a lot more common use cases the Flex component applies to vs. the Grid.

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+
+---
+
+## Flexbox layout "recipes"
+
+Below some examples of common generic layouts that can be achieved using CSS `flexbox`, and therefore the `Layout::Flex` component.
+
+### Alignment
+
+Using `flexbox`/`Layout::Flex` is one of the simplest way to control the alignment of lists of items.
```
KristinLBradley, line 212:
```suggestion
Using `flexbox`/`Layout::Flex` is one of the simplest ways to control the alignment of lists of items.
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+  <Doc::placeholder @width="auto" @height="40px" @background="#d2f4ff">Other content</Doc::placeholder>
+  <Doc::placeholder class="doc-flex-margin-left-auto" @width="auto" @height="40px" @background="#fff8d2">Content flushed on the right</Doc::placeholder>
+</Hds::Layout::Flex>
+```
+
+In this case:
+
+- by applying a `margin-left: auto` to the last flex item, the flexbox layout engine pushes it completely to the right (because implicitly it has `margin-right: 0`)
+
+The same technique can also be applied when there are multiple elements that need to be aligned on the right:
```
KristinLBradley, line 247:
```suggestion
The same technique can also be applied when there are multiple elements that need to be aligned to the right:
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+  <Doc::placeholder @width="auto" @height="40px" @background="#d2f4ff">Some content</Doc::placeholder>
+  <Doc::placeholder @width="auto" @height="40px" @background="#d2f4ff">Other content</Doc::placeholder>
+  <Doc::placeholder class="doc-flex-margin-left-auto" @width="auto" @height="40px" @background="#fff8d2">More content</Doc::placeholder>
+  <Doc::placeholder @width="auto" @height="40px" @background="#e4c5f3">Extra content</Doc::placeholder>
+</Hds::Layout::Flex>
+```
+
+In this case:
+
+- the `margin-left: auto` is applied only to the first item that is flushed to the right; all the other items are automatically pushed to the right by the first flushed item
```
KristinLBradley, line 260:
```suggestion
- the `margin-left: auto` is applied only to the first item that is flushed to the right; all the other items are automatically pushed to the right by the first flushed right item
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+
+#### Vertical centering
+
+For a long time, this has been one classic problem in CSS: how to vertically center two or more items? The more reliable solution is to use flexbox's alignment:
```
KristinLBradley, line 264:
```suggestion
A classic CSS problem: how to vertically center two or more items? A reliable solution is to use flexbox alignment:
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+<Hds::Layout::Flex @align="center" @gap="8">
+  <Doc::placeholder @width="100px" @height="100px" @background="#d2f4ff">Some content</Doc::placeholder>
+  <Doc::placeholder @width="200px" @height="40px" @background="#fff8d2">Other content</Doc::placeholder>
+  <Doc::placeholder @width="300" @height="20px" @background="#e4c5f3">More content</Doc::placeholder>
+</Hds::Layout::Flex>
+```
+
+In this case:
+
+- the `@align="center"` applied to the flex container forces the flex items to vertically align their centers along what is called the "cross axis"
```
KristinLBradley, line 276:
```suggestion
- the `@align="center"` applied to the flex container forces the flex items to vertically align their centers along the "cross axis"
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+
+#### Both horizontal and vertical centering
+
+This has also been another classic CSS problem: how to center _both_ horizontally and vertically some content, inside a larger container? Again, the solution relies in using flexbox's alignment:
```
KristinLBradley, line 280:
```suggestion
Another classic CSS problem: how to center content _both_ horizontally and vertically inside a larger container? Again, the solution is to use flexbox alignment:
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+
+```handlebars
+<Hds::Layout::Flex @justify="center" @align="center" class="doc-flex-fixed-height-container doc-flex-outlined-container">
+  <Doc::placeholder @width="240px" @height="180px" @background="#d2f4ff">Some content</Doc::placeholder>
+</Hds::Layout::Flex>
+```
+
+In this case:
+
+- the `@justify="center"` and `@align="stretch"` applied to the flex container force the flex item to be centered both horizontally and vertically (_note: we have applied an outline and a fixed height to the flex container to demonstrate the behavior_)
```
KristinLBradley, line 290:
```suggestion
- the `@justify="center"` and `@align="stretch"` arguments applied to the flex container force the flex item to be centered both horizontally and vertically (_note: we have applied an outline and a fixed height to the flex container to demonstrate the behavior_)
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+  <LF.Item @grow={{true}}>
+    <Doc::placeholder @height="40px" @background="#e5ffd2">This content occupies as much space as possible</Doc::placeholder>
+  </LF.Item>
+  <Doc::placeholder @width="auto" @height="40px" @background="#d2f4ff">Extra content</Doc::placeholder>
+</Hds::Layout::Flex>
+```
+
+In this case:
+
+- we have applied `@grow=\{{true}}` to the wrapping `Flex::Item`, forcing it to use all the available extra space inside the flexbox container
```
KristinLBradley, line 324:
```suggestion
- we have applied `@grow=\{{true}}` to the wrapping `Flex::Item`, forcing it to use all the available space inside the flexbox container
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+  <Hds::Badge @text="Stretched badge" @color="critical" />
+  <LF.Item>
+    <Hds::Badge @text="Non-stretched badge" @color="success" />
+  </LF.Item>
+</Hds::Layout::Flex>
+```
+
+### Wrapping
+
+One of the useful functionalities of the flexbox layout model is that it allows the items to wrap along multiple rows when there is not enough space in the container to fit them all in a single line:
```
KristinLBradley, line 391:
```suggestion
One useful function of the flexbox layout model is that it allows items to wrap along multiple rows when there is not enough space in the container to fit them all in a single line:
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+    </LF.Item>
+  </Hds::Layout::Flex>
+  <Doc::placeholder @height="50px" @background="#fff8d2">Bottom content</Doc::placeholder>
+</Hds::Layout::Flex>
+```
+
+In this case:
+
+- the outer container is a column-based flexbox layout, with three child items
+- the second item is a row-based flexbox layout, used to create two side-by side content blocks; of these two the first has a fixed width, while the second fills the remaining area thanks to the `@grow=\{{true}}` attribute
```
KristinLBradley, line 427:
```suggestion
- the second item is a row-based flexbox layout, used to create two side-by-side content blocks; of these two the first has a fixed width, while the second fills the remaining area thanks to the `@grow=\{{true}}` attribute
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+
+!!! Insight
+
+Depending on the complexity of the design you need to implement, you may want to consider using a CSS `grid` layout, instead of a `flex` layout.
```
KristinLBradley, line 431:
```suggestion
Depending on the complexity and type of design you need to implement, you may want to consider using a CSS `grid` layout, instead of a `flex` layout.
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+
+## Common layout patterns
+
+Below some more realistic examples of common layouts that can be achieved using the `Layout::Flex` component, in composition with other HDS components.
+
+!!! Warning
+
+**Important**
+
+The examples below are meant to show how one _could_ use the `Layout::Flex` component to implement certain common/standard UI patterns. They're **not** meant to be taken literally as they are and be used in production code. Also, some of these patterns are already implemented directly in HDS components that are ready to use.
```
KristinLBradley, line 445:
```suggestion
The examples below are meant to show how one _could_ use the `Layout::Flex` component to implement certain common/standard UI patterns. They're **not** meant to be taken literally as they are and be used in production code. Also, some of these patterns may already implemented directly in HDS components that are ready to use.
```

KristinLBradley: Up to you if you want to use "are" or "may" here. (I used "may" in this warning notice in the `Grid` documentation but the grid layout has fewer uses...)

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+
+!!!
+
+### Media + Text/Content
+
+This is a classic layout, where a "media" element (it could be an image, an icon, an avatar, etc) is laid out next to a block of text (but could be more complex content). The two elements are visually centered along their horizontal axis.
```
KristinLBradley, line 451: Just a note: The [classic "Media object"](https://www.stubbornella.org/2010/06/25/the-media-object-saves-hundreds-of-lines-of-code/) referred to a media element laid out next to associated text with both aligned to the top.

I've sometimes seen Media objects with a centered horizontal axis layout referred to as "Flag objects" (really just a variation of the Media Object).

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+
+
+```handlebars
+<Hds::Layout::Flex @align="center" @gap="8">
+  <img class="doc-flex-media-avatar" src="/assets/images/avatar.png" alt="portrait of a cat wearing coat and tie" />
+  <Hds::Text::Body @size="200" @tag="p">Lorem ipsum dolor sit amet</Hds::Text::Body>
+</Hds::Layout::Flex>
+
+<br/>
```
KristinLBradley, line 460: Super nit: it bothers me to see `<br/>` tags used as spacers like this... I'd personally avoid including those in code examples that others may copy.

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+    <Hds::Card::Container @level="mid" @hasBorder={{true}} class="doc-flex-cards-item">
+      <Hds::Text::Body @tag="p" @size="200">This is also some generic content inside a card, but the text is longer and makes the card grow vertically</Hds::Text::Body>
+    </Hds::Card::Container>
+  </LF.Item>
+</Hds::Layout::Flex>
+```
+
+!!! Warning
+
+A responsive implementation of this UI pattern is a bit more complex, and requires to use `min-max/max-width` values for the cards, and allow wrapping of the flex item elements. It may also require changing direction of the flex container at a certain breakpoint.
```
KristinLBradley, line 505:
```suggestion
A responsive implementation of this UI pattern is a bit more complex and requires usage of `min-max/max-width` values for the cards and allowing wrapping of the flex item elements. It may also require changing the direction of the flex container at a certain breakpoint.
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+  <Hds::ApplicationState as |A|>
+    <A.Header @title="Empty state title text" @icon="alert-circle" />
+    <A.Body @text="The item you were looking for was not found." />
+  </Hds::ApplicationState>
+</Hds::Layout::Flex>
+```
+
+### Right aligned actions
+
+Using the `margin-left: auto` trick described above, one can achieve the following layout, where one of the flex items is flushed on the right:
```
KristinLBradley, line 544:
```suggestion
Using the `margin-left: auto` trick described above, one can achieve the following layout, where one of the flex items is flushed to the right:
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+  <Hds::Dropdown class="doc-flex-margin-left-auto" as |D|>
+    <D.ToggleButton @text="Main actions" />
+    <D.Interactive @route="components">Action One</D.Interactive>
+    <D.Interactive @route="components">Action Two</D.Interactive>
+    <D.Interactive @route="components">Action Three</D.Interactive>
+  </Hds::Dropdown>
+</Hds::Layout::Flex>
+```
+
+If needed, multiple items can be flushed on the right:
```
KristinLBradley, line 559:
```suggestion
If needed, multiple items can be flushed to the right:
```

KristinLBradley (commented):
I went through the current text in detail and added a number of small suggested fixes/edits.

didoo:
> I went through the current text in detail and added a number of small suggested fixes/edits.

@KristinLBradley you're a 🌟

website/docs/layouts/flex/partials/code/how-to-use.md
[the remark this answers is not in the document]
didoo: Tomorrow I'll try to "merge" the content/sections in a different way.

website/docs/layouts/flex/partials/code/how-to-use.md
didoo: 👏

website/docs/layouts/flex/partials/code/how-to-use.md
didoo: I'll update the description:
```suggestion
This is a classic layout, where a "media" element (it could be an image, an icon, an avatar, etc) is laid out next to a block of text (but could be more complex content). The two elements are either visually centered along their horizontal axis, or aligned to their top.
```

website/docs/layouts/flex/partials/code/how-to-use.md
didoo: the alternative is to have `doc-something` classes just for this, which is also not great, no? if you have strong feelings I can change it though.

website/docs/layouts/flex/partials/code/how-to-use.md
KristinLBradley: I just deleted one word... :)

website/docs/layouts/flex/partials/code/how-to-use.md
KristinLBradley: It just really bugs me to model something in an example that I wouldn't recommend people use but it's probably just me

didoo: Fixed in https://github.com/hashicorp/design-system/pull/2764/commits/e89a70e4460f05b13b27a5d87616195d33b1173b

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+  </Hds::Card::Container>
+  <Hds::Card::Container @level="mid" @hasBorder={{true}} class="doc-flex-cards-item">
+    <Hds::Text::Body @tag="p" @size="200">This is also some generic content inside a card, but the text is longer and makes the card grow vertically</Hds::Text::Body>
+  </Hds::Card::Container>
+</Hds::Layout::Flex>
+```
+
+### Empty state
+
+Using the horizontal/vertical alignment offered by flexbox it's easy to implement a standard empty state in a page (as consumer you will just have to take care of the height of the container in relation to the available vertical space in the page):
```
shleewhite, line 527:
```suggestion
Using the horizontal/vertical alignment offered by flexbox, it is also possible to implement a standard empty state in a page. As consumer you will just have to take care of the height of the container in relation to the available vertical space in the page.
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+  <Hds::ApplicationState as |A|>
+    <A.Header @title="Empty state title text" @icon="alert-circle" />
+    <A.Body @text="The item you were looking for was not found." />
+  </Hds::ApplicationState>
+</Hds::Layout::Flex>
+```
+
+### Right aligned actions
+
+Using the `margin-left: auto` trick described above, one can achieve the following layout, where one of the flex items is flushed to the right:
```
shleewhite, line 544:
```suggestion
Using the `margin-left: auto` trick described above, one can achieve the following layout, where one of the flex items is flushed to the right.
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+  <Hds::Dropdown class="doc-flex-margin-left-auto" as |D|>
+    <D.ToggleButton @text="Main actions" />
+    <D.Interactive @route="components">Action One</D.Interactive>
+    <D.Interactive @route="components">Action Two</D.Interactive>
+    <D.Interactive @route="components">Action Three</D.Interactive>
+  </Hds::Dropdown>
+</Hds::Layout::Flex>
+```
+
+If needed, multiple items can be flushed to the right:
```
shleewhite, line 559:
```suggestion
If needed, multiple items can be flushed to the right.
```

shleewhite: This example seems repetitive with the right flushed example in "recipes"... might be worth it to consolidate them and only show the "real world example".

website/docs/layouts/flex/index.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+
+export default class Index extends Component {
+  yourSidebarBooleanFlag = false;
```
shleewhite, line 9: What is this for?

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,78 @@
...
+### Layout::Flex
+
+<Doc::ComponentApi as |C|>
+  <C.Property @name="tag" @type="HTMLElementTagName" @default="div">
+    A valid HTML tag name to be used to render the flexbox element.
+  </C.Property>
+  <C.Property @name="direction" @type="enum" @values={{array "row" "column"}} @default="row">
+    The value of the CSS `flex-direction` property, which sets how the flex items (children) are placed in the flex container, defining the [main axis](https://developer.mozilla.org/en-US/docs/Glossary/Main_Axis) and the direction (for a technical explanation: [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction)).
+    <br/><br/>
+    <em>Notice: we don't expose the "reverse" directions because they come with intrinsic accessibility limitations that we prefer our consumers to avoid.</em>
```
shleewhite, line 12:
```suggestion
    <em>Note: we don't expose the "reverse" directions because they come with intrinsic accessibility limitations.</em>
```

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,78 @@
...
+  </C.Property>
+  <C.Property @name="justify" @type="enum" @values={{array "start" "center" "end" "space-between" "space-around" "space-evenly"}}>
+    The value of the CSS `justify-content` property, which defines how the space is distributed between and around content items along the [main axis](https://developer.mozilla.org/en-US/docs/Glossary/Main_Axis) of the flex container (for a technical explanation: [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content)).
+    <br/><br/>
+    <em>Notice: we expose only a subset of the values allowed for this property, covering only the most common use cases.</em>
+    <br/><br/>
+    <em>Tip: when the `@direction` is `row` this argument controls the horizontal alignment; when it's `column` it controls the vertical alignment.</em>
```
shleewhite, line 19: Suggestion: it might be useful to move this up to the examples that use justify/align.

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,78 @@
...
+The `Layout::Flex::Item` component, yielded as contextual component, to be used as child of the `flexbox` element to control its `basis/grow/shrink` values (and other properties).
+
+<Doc::ComponentApi as |C|>
+  <C.Property @name="tag" @type="string" @default="div">
+    HTML tag to be used to render the flex item element.
+  </C.Property>
+  <C.Property @name="basis" @type="string|0">
+    The value (size) of the CSS `flex-basis` property, which sets the initial main size of the flex item (for a technical explanation: [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis)).
+    <br/><br/>
+    <em>Notice: when the value is set to `0`, a CSS class is used to apply the `flex-basis` property. In all other cases, an inline `style` declaration is used to accommodate all the possible values that this CSS property can have.</em>
```
shleewhite, line 55:
```suggestion
    <em>Note: when the value is set to `0`, a CSS class is used to apply the `flex-basis` property. In all other cases, an inline `style` declaration is used to accommodate all the possible values that this CSS property can have.</em>
```

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,78 @@
...
+  </C.Property>
+  <C.Property @name="grow" @type="boolean|number|string">
+    The value (size or keyword) of the CSS `flex-grow` property, which sets the flex grow factor of the flex item (for a technical explanation:  [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-grow)).
+    <br/><br/>
+    <em>Notice: when the value is set to `true/false` or `0/1` a CSS class is used to apply the `flex-grow` property. In all other cases, an inline `style` declaration is used, to accommodate all the possible values that this CSS property can have.</em>
```
shleewhite, line 60:
```suggestion
    <em>Note: when the value is set to `true/false` or `0/1` a CSS class is used to apply the `flex-grow` property. In all other cases, an inline `style` declaration is used, to accommodate all the possible values that this CSS property can have.</em>
```

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,78 @@
...
+  </C.Property>
+  <C.Property @name="shrink" @type="boolean|number|string">
+    The value (size or keyword) of the CSS `flex-shrink` property, which sets sets the flex shrink factor of the flex item (for a technical explanation: [see MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-shrink)).
+    <br/><br/>
+    <em>Notice: when the value is set to `true/false` or `0/1` a CSS class is used to apply the `flex-shrink` property. In all other cases, an inline `style` declaration is used, to accommodate all the possible values that this CSS property can have.</em>
```
shleewhite, line 65:
```suggestion
    <em>Note: when the value is set to `true/false` or `0/1` a CSS class is used to apply the `flex-shrink` property. In all other cases, an inline `style` declaration is used, to accommodate all the possible values that this CSS property can have.</em>
```

website/docs/layouts/flex/partials/code/component-api.md
```diff
@@ -0,0 +1,78 @@
...
+  </C.Property>
+  <C.Property @name="enableCollapseBelowContentSize" @type="boolean">
+    When this special argument is set to `true`, it applies `min-width: 0` to the element, allowing the flex item to shrink below its content's intrinsic minimum width (e.g., used with text truncated with ellipses)
+    <br/><br/>
+    <em>Notice: this may have accessibility implications; be considerate in how this special argument is used and how this impacts the content of the flex item element.</em>
```
shleewhite, line 70:
```suggestion
    <em>Note: this may have accessibility implications; be considerate in how this special argument is used and how this impacts the content of the flex item element.</em>
```

Suggestion: Could we be more explicit about what the issue is that they should look for?

shleewhite (commented):
Nice! Overall, super helpful docs. All of my comments are just suggestions.

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+  <LF.Item @grow={{false}}>
+    {{! some other content here }}
+  </LF.Item>
+  <div>{{! more content here }}</div>
+</Hds::Layout::Flex>
+```
+
+!!! Info
+
+Note: there is no strict need to use the `Layout::Flex::Item` subcomponent as a direct child of `Layout::Flex`; use it only when necessary to tweak flex styles of an individual child item via the `@basis/@grow/@shrink` (to avoid rendering an extra Ember component).
```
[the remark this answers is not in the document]
didoo: Nice!

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+!!!
+
+📚 Refer to the ["Recipes"](#flexbox-layout-recipes) and [Patterns](#common-layout-patterns) sections below for practical examples of how the Flexbox `gap` property can be used to implement specific layouts.
+
+#### Non-standard gap values
+
+If you absolutely have to use non-standard spacing value(s) for the flex `gap`, you can use the internal `--hds-layout-flex-row-gap` and `--hds-layout-flex-column-gap` CSS variables and pass custom values to them (e.g., via a local CSS variable or an inline style):
+
+```handlebars{data-execute=false}
+<Hds::Layout::Flex class="doc-flex-demo-custom-flex-column-gap">
```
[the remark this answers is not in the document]
didoo: I'm going to add a comment with the CSS declaration, I think we have done it elsewhere

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+  {{!
+    multiple flex items appearing on multiple rows
+    with a vertical gap of 10px and an horizontal one of 0.625rem
+  }}
+</Hds::Layout::Flex>
+```
+
+### isInline
+
+To change the default display from `flex` to `inline-flex`, set `@isInline` to `true`:
```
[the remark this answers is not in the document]
didoo: It's pretty standard in our documentation to use a colon at the end of the paragraph that introduces/explains a code snippet:
https://helios.hashicorp.design/components/button?tab=code#add-an-icon

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+  <Doc::placeholder @width="auto" @height="40px" @background="#d2f4ff">Some content</Doc::placeholder>
+  <Doc::placeholder @width="auto" @height="40px" @background="#d2f4ff">Other content</Doc::placeholder>
+  <Doc::placeholder @width="auto" @height="40px" @background="#d2f4ff">More content</Doc::placeholder>
+  <Doc::placeholder @width="200px" @height="40px" @background="#d2f4ff">Some content that is wider</Doc::placeholder>
+</Hds::Layout::Flex>
+```
+
+In this case:
+
+- the `@justify="space-between"` applied to the flex container forces the flex layout model to distribute the space evenly between the items; the space between the items remains the same, even if their size is different
```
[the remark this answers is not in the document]
didoo: For consistency, I prefer to keep a list even if it's a single item

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+```handlebars
+<Hds::Layout::Flex @gap="16" class="doc-flex-outlined-container" as |LF|>
+  <Doc::placeholder @width="auto" @height="80px" @background="#d2f4ff">Some content that is taller than the sibling item</Doc::placeholder>
+  <button type="button" class="doc-flex-margin-left-auto">A simple button</button>
+</Hds::Layout::Flex>
+```
+
+This behavior is due to the default behavior of the flexbox layout model, and depends on the CSS `display` of those items.
+
+To avoid this, one has to choose a non-stretching alignment for the items:
```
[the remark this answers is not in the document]
didoo: Yeah, makes sense (I added the second one later)

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+```
+
+In this case:
+
+- the outer container is a column-based flexbox layout, with three child items
+- the second item is a row-based flexbox layout, used to create two side-by-side content blocks; of these two the first has a fixed width, while the second fills the remaining area thanks to the `@grow=\{{true}}` attribute
+
+!!! Insight
+
+Depending on the complexity and type of design you need to implement, you may want to consider using a CSS `grid` layout, instead of a `flex` layout.
```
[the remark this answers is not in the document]
didoo: Yes, since the two branches are independent (not one based off the other) we have a plan for a follow-up ticket once they're both merged, to bi-directionally link the two documentation pages (also as related components)

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+
+## Common layout patterns
+
+Below are some more realistic examples of common layouts that can be achieved using the `Layout::Flex` component, in composition with other HDS components.
+
+!!! Warning
+
+**Important**
+
+The examples below are meant to show how one _could_ use the `Layout::Flex` component to implement certain common/standard UI patterns. They're **not** meant to be taken literally as they are and be used in production code. Also, some of these patterns may already implemented directly in HDS components that are ready to use.
```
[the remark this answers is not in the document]
didoo: We use flex in a lot of places! We can't list them all here :)

website/docs/layouts/flex/partials/code/how-to-use.md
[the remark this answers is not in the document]
shleewhite: Oops! sorry, I thought our standard was to do a period.

website/docs/layouts/flex/partials/code/how-to-use.md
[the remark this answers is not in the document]
shleewhite: Hahah I meant more specifically which of the common UI patterns you demonstrate below are already in HDS components. Is it the card layouts? Otherwise I'd just remove that sentence.

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+
+```handlebars{data-execute=false}
+<Hds::Layout::Flex>
+  <div>{{! some content here }}</div>
+  <div>{{! some other content here }}</div>
+  <div>{{! more content here }}</div>
+</Hds::Layout::Flex>
+```
+
+In this way, every direct child element of the **flexbox container** will be treated, by the browser's layout engine, as a **flex item** (for details about what this means, refer to the guide linked at the top of this page).
```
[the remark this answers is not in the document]
didoo: I prefer to have a single place where we update the link if necessary. Does it make sense?

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,573 @@
...
+    <Hds::Text::Display @size="400" @tag="h5">Lorem ipsum dolor</Hds::Text::Display>
+    <Hds::Text::Body @size="200" @tag="p">Lorem ipsum dolor sit amet consectetur adipiscing elit</Hds::Text::Body>
+    <Hds::Link::Standalone @icon="external-link" @iconPosition="trailing" @text="Documentation" @href="#" />
+  </Hds::Layout::Flex>
+</Hds::Layout::Flex>
+```
+
+### Group of cards
+
+Using the automatic layout offered by flexbox, it's extremely easy to create a group of identically sized, evenly spaced cards:
```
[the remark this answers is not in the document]
didoo:
```suggestion
Using the automatic layout offered by flexbox, you can create a group of identically sized, evenly spaced cards:
```

website/docs/layouts/flex/partials/code/how-to-use.md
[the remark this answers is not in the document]
shleewhite: ¯\_(ツ)_/¯ thats totally fair. I just think its easier to give the users the link in context if you're going to mention it, plus its also linked several times in the API section.

This is def a personal preference thing - feel free to not apply.

website/docs/layouts/flex/partials/code/how-to-use.md
dchyun: On the punctuation suggestion, In this doc, I saw all of the text that is followed by a code example uses a `:` instead of a `.`.  Do we usually use periods or colons in this situation?

KristinLBradley: @dchyun Good call out. Checking a few existing examples, it looks like we normally end these introductory sentences for code samples with a period "." vs. a colon ":"

website/docs/layouts/flex/partials/code/how-to-use.md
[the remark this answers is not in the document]
didoo: Changed as you suggested.

website/docs/layouts/flex/index.js
didoo: I thought it was generated but the blueprint, but on a second look you're right, it comes from me copying the controller from the `AppFrame` docs, where it's actually used. I'll remove entirely the file, is not needed.

website/docs/layouts/flex/partials/code/how-to-use.md
didoo: So which one? I'm confused (eg. see here: https://helios.hashicorp.design/components/button?tab=code)

website/docs/layouts/flex/partials/version-history/version-history.md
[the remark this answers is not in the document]
alex-ju: the automation script does create the file if it doesn't exist, but it doesn't import it in the main `index.md` (we can upgrade the script to handle this as well) so I'd say it's generally safer to create it with the component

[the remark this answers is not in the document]
alex-ju: raised #2784 to add the `version-history` partial to our blueprint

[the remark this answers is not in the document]
dchyun: Ah thanks for catching that. @KristinLBradley I may have given you some bad advice to delete this file on the `Flex`. 😅  Can you add it back in?

website/docs/layouts/flex/partials/code/how-to-use.md
dchyun: Seems like _mostly_ we use periods. Such as in the [badge](https://helios.hashicorp.design/components/badge?tab=code). Colons generally are used in the guidelines before a list. 

Maybe the guidance can be if what you write before the code block is a complete sentence, use a period. If it is not, use a colon. 

> To specify in which [direction](https://developer.mozilla.org/en-US/docs/Web/CSS/flex-direction) the flex items are laid out, use the `@direction` argument.

> Use the `@direction` argument to specify which direction the flex items are laid out in. For example:

From button guidelines
Screenshot 2025-03-21 at 9 25 21 AM

didoo: I've removed as many `:` as I could.

KristinLBradley: In either case it's not too big a deal to mix & match the two.

website/docs/layouts/flex/partials/code/how-to-use.md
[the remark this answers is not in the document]
didoo: @KristinLBradley @dchyun @shleewhite I have re-worked the "How to use" section, merging/moving content per your suggestions. See if now works for you: https://hds-website-git-flex-component-documentation-hashicorp.vercel.app/layouts/flex

website/docs/layouts/flex/partials/code/how-to-use.md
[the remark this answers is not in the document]
didoo: With the reorganization of the content done in https://github.com/hashicorp/design-system/pull/2764#discussion_r2007852375 I think now it's much better

website/docs/layouts/flex/partials/code/how-to-use.md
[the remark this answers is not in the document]
dchyun: This is looking great! Thanks for reorganizing. I like how the recipe examples were integrated into each section, and how the new simple examples are placed before them. I'll take some time to comb through it in more detail, but works great for me.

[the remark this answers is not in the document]
shleewhite: Yup! 100% agree with Dylan, the new structure is great.

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,569 @@
...
+</Hds::Layout::Flex>
+```
+
+In this case, the `@justify="center"` and `@align="stretch"` arguments applied to the flex container force the flex item to be centered both horizontally and vertically
+
+_Note: we have applied an outline and a fixed height to the flex container to demonstrate the behavior_
+
+### Wrapping
+
+One useful function of the flexbox layout model is that it allows items to wrap along multiple rows when there is not enough space in the container to fit them all in a single line
```
dchyun, line 264:
```suggestion
One useful function of the flexbox layout model is that it allows items to wrap along multiple rows when there is not enough space in the container to fit them all in a single line.
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,569 @@
...
+  <Doc::placeholder @width="300px" @height="40px" @background="#fff8d2">Extra content</Doc::placeholder>
+  <Doc::placeholder @width="240px" @height="40px" @background="#f3d9c5">Even more content</Doc::placeholder>
+</Hds::Layout::Flex>
+```
+In this case the `@wrap=\{{true}}` applied to the flex container allows the flex items to wrap on multiple lines. The `@gap=\{{array "8" "24"}}` applied to the flex container defines the gap between the items along the two "directions" (the first value is the space between "rows", the second between "columns").
+
+
+### Sizing
+
+In flexbox it's possible to control the relative sizing of flex items within a flex container using the `basis`, `grow`, and `shrink` properties.
```
dchyun, line 288:
```suggestion
In flexbox, it's possible to control the relative sizing of flex items within a flex container using the `basis`, `grow`, and `shrink` properties.
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,569 @@
...
+<Hds::Layout::Flex as |LF|>
+  ...
+  <LF.Item @basis="200px" @grow={{1}} @shrink={{0}}>
+    {{! the flex item content here }}
+  </HLF.Item>
+  ...
+</Hds::Layout::Flex>
+```
+
+How these three properties impact the (relative) size of a flex item is not trivial, we suggest consulting [the MDN guide](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout/Basic_concepts_of_flexbox).
```
dchyun, line 302:
```suggestion
How these three properties impact the (relative) size of a flex item is not trivial. We suggest consulting [the MDN guide](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_flexible_box_layout/Basic_concepts_of_flexbox).
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,569 @@
...
+  </Hds::Card::Container>
+  <Hds::Card::Container @level="mid" @hasBorder={{true}} class="doc-flex-cards-item">
+    <Hds::Text::Body @tag="p" @size="200">This is also some generic content inside a card, but the text is longer and makes the card grow vertically</Hds::Text::Body>
+  </Hds::Card::Container>
+</Hds::Layout::Flex>
+```
+
+### Empty state
+
+Using the horizontal/vertical alignment offered by flexbox, it is also possible to implement a standard empty state in a page. As consumer you will just have to take care of the height of the container in relation to the available vertical space in the page.
```
dchyun, line 523:
```suggestion
Using the horizontal/vertical alignment offered by flexbox, it is also possible to implement a standard empty state in a page. As consumer, you will just have to take care of the height of the container in relation to the available vertical space in the page.
```

dchyun (commented):
Looks great! New example are really helpful and organized. Just a few small things I caught.

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,569 @@
...
+    with a vertical gap of 10px and an horizontal one of 0.625rem
+  }}
+</Hds::Layout::Flex>
+```
+
+### Alignment
+
+Using `flexbox` is one of the simplest ways to control how lists of items are aligned and spaced, and it's done via a set of `justify-*` and `align-*` properties.
+
+In the `Layout::Flex` component this is achieved use the `@justify` and `align` arguments.
```
KristinLBradley, line 178:
```suggestion
In the `Layout::Flex` component this is achieved using the `@justify` and `align` arguments.
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,569 @@
...
+
+```handlebars
+<Hds::Layout::Flex @justify="center" @align="center" class="doc-flex-fixed-height-container doc-flex-outlined-container">
+  <Doc::placeholder @width="240px" @height="180px" @background="#d2f4ff">Some content</Doc::placeholder>
+</Hds::Layout::Flex>
+```
+
+In this case, the `@justify="center"` and `@align="stretch"` arguments applied to the flex container force the flex item to be centered both horizontally and vertically
+
+_Note: we have applied an outline and a fixed height to the flex container to demonstrate the behavior_
```
KristinLBradley, line 260:
```suggestion
_Note: We have applied an outline and a fixed height to the flex container to demonstrate the behavior._
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,569 @@
...
+  </LF.Item>
+  <Doc::placeholder @width="auto" @height="40px" @background="#d2f4ff">Extra content</Doc::placeholder>
+</Hds::Layout::Flex>
+```
+
+In this case we have applied `@grow=\{{true}}` to the wrapping `Flex::Item`, forcing it to use all the available space inside the flexbox container.
+
+#### Prevent collapsing of content
+
+In the example below, you can see (look at the button element) how, if one of the flex items has an intrinsic size larger than the flex container, it may squeeze the rest of the flex items:
```
KristinLBradley, line 334:
```suggestion
In the example below, you can see how, if one of the flex items has an intrinsic size larger than the flex container, it may squeeze the rest of the flex items (specifically the button element in this case).
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,569 @@
...
+```handlebars
+<Hds::Layout::Flex @gap="16" class="doc-flex-outlined-container" as |LF|>
+  <Doc::placeholder @width="100%" @height="auto" @background="#d2f4ff">Some content that wants to occupy all the available space</Doc::placeholder>
+  <LF.Item @shrink={{false}}>
+    <button type="button">A simple button</button>
+  </LF.Item>
+</Hds::Layout::Flex>
+```
+
+In this case we have used a wrapping `Flex::Item` and applied `@shrink=\{{false}}` to it, to prevent it from shrinking.
```
KristinLBradley, line 354:
```suggestion
In this case, we have used a wrapping `Flex::Item` and applied `@shrink=\{{false}}` to it, to prevent it from shrinking.
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,569 @@
...
+To avoid this, one can choose a non-stretching alignment for the items.
+
+```handlebars
+<Hds::Layout::Flex @align="start" @gap="16" class="doc-flex-outlined-container" as |LF|>
+  <Doc::placeholder @width="auto" @height="80px" @background="#d2f4ff">Some content that is taller than the sibling item</Doc::placeholder>
+  <button type="button" class="doc-flex-margin-left-auto">A simple button</button>
+</Hds::Layout::Flex>
+```
+
+In this case the `@align="start"` applied to the flex container forces the flex items to vertically align at the start (top) along what is called the "cross axis".
```
KristinLBradley, line 378:
```suggestion
In this case, the `@align="start"` applied to the flex container forces the flex items to vertically align at the start (top) along what is called the "cross axis".
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,569 @@
...
+
+An alternative solution is to wrap the element that shouldn't stretch with a flex `Item`, like in this example:
```
KristinLBradley, line 380:
```suggestion
An alternative solution is to wrap the element that shouldn't stretch with a flex `Item` like in this example:
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,569 @@
...
+    <Doc::placeholder @width="150px" @height="auto" @background="#e4c5f3">Side content</Doc::placeholder>
+    <LF.Item @grow={{true}}>
+      <Doc::placeholder @width="auto" @height="250px" @background="#d2f4ff">Main content</Doc::placeholder>
+    </LF.Item>
+  </Hds::Layout::Flex>
+  <Doc::placeholder @height="50px" @background="#fff8d2">Bottom content</Doc::placeholder>
+</Hds::Layout::Flex>
+```
+
+In this case the outer container is a column-based flexbox layout, with three child items while the second item is a row-based flexbox layout, used to create two side-by-side content blocks. Of these two, the first has a fixed width, while the second fills the remaining area thanks to the `@grow=\{{true}}` attribute.
```
KristinLBradley, line 423:
```suggestion
In this case, the outer container is a column-based flexbox layout, with three child items while the second item is a row-based flexbox layout, used to create two side-by-side content blocks. Of these two, the first has a fixed width, while the second fills the remaining area thanks to the `@grow=\{{true}}` attribute.
```

website/docs/layouts/flex/partials/code/how-to-use.md
```diff
@@ -0,0 +1,569 @@
...
+
+Depending on the complexity and type of design you need to implement, you may want to consider using a CSS `grid` layout, instead of a `flex` layout.
+
+!!!
+
+---
+
+## Common layout patterns
+
+Below are some examples of common layouts that can be achieved using the `Layout::Flex` component, in composition with other HDS components.
```
KristinLBradley, line 435:
```suggestion
Below are some examples of common layouts that can be achieved using the `Layout::Flex` component, in combination with other HDS components.
```

website/docs/layouts/flex/partials/code/how-to-use.md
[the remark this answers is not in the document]
KristinLBradley: Agree. This looks good. Resolving this comment thread.

KristinLBradley (dismissed):

website/docs/layouts/flex/index.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Layout::Flex
+description: A component used to implement layouts based on CSS flex
+caption: A layout-only component for the organization of content based on the CSS Flexbox model
```
dchyun, line 4:
```suggestion
caption: A layout component for the organization of content based on the CSS Flexbox model.
```
[Nit] Align to the AppFrame caption, which uses "layout" not "layout-only". Other component captions also have periods at the end.

website/docs/layouts/flex/index.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Layout::Flex
+description: A component used to implement layouts based on CSS flex.
+caption: A layout component for the organization of content based on the CSS Flexbox model.
```
KristinLBradley, line 4:
```suggestion
caption: A layout-only component for the organization of content based on the CSS Flexbox model.
```

dchyun (dismissed):

shleewhite (approved):
My comments are just suggestions to match the wording in the other component API notes.

website/docs/layouts/flex/index.md
```diff
@@ -0,0 +1,20 @@
+---
+title: Layout::Flex
+description: A component used to implement layouts based on CSS flex.
+caption: A layout-only component for the organization of content based on the CSS Flexbox model.
```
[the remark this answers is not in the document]
KristinLBradley: Hi yes, I commented on the other PR I think. In thinking about it, I thought emphasizing that these are layout-only components actually made sense. They aren't quite the same as the `AppFrame` component even though they are grouped together with it.

KristinLBradley:
Merged in latest `main` to this branch & fixed conflicts. Going to merge the Flex & Grid docs together next so they can be cross-linked & then go out together.

dchyun:
This is something I caught when working on #2791. Can you add an acceptance test for the new page? I noticed there are no acceptance tests for the layouts pages, so we will need to make a `/layouts` folder as well.

shleewhite:
> This is something I caught when working on #2791. Can you add an acceptance test for the new page? I noticed there are no acceptance tests for the layouts pages, so we will need to make a `/layouts` folder as well.

I don't think adding acceptance tests for Flex/Grid is useful because there are no elements with roles or styles we need to check for contrast since its just layout (but we should probably add acceptance tests for AppFrame since it has elements that aren't divs)

dchyun:
> > This is something I caught when working on #2791. Can you add an acceptance test for the new page? I noticed there are no acceptance tests for the layouts pages, so we will need to make a `/layouts` folder as well.
> 
> I don't think adding acceptance tests for Flex/Grid is useful because there are no elements with roles or styles we need to check for contrast since its just layout (but we should probably add acceptance tests for AppFrame since it has elements that aren't divs)
> 
> cc @dchyun

That makes sense to me to not add the a11y checks. Should we still add the acceptance tests for making sure visiting the url works? I'll add the appframe test in a separate pr.

KristinLBradley:
> > > This is something I caught when working on #2791. Can you add an acceptance test for the new page? I noticed there are no acceptance tests for the layouts pages, so we will need to make a `/layouts` folder as well.
> > 
> > 
> > I don't think adding acceptance tests for Flex/Grid is useful because there are no elements with roles or styles we need to check for contrast since its just layout (but we should probably add acceptance tests for AppFrame since it has elements that aren't divs)
> > cc @dchyun
> 
> That makes sense to me to not add the a11y checks. Should we still add the acceptance tests for making sure visiting the url works? I'll add the appframe test in a separate pr. CC: @shleewhite @KristinLBradley

@dchyun Do you mean the Percy tests? They're already included in the component PRs that were merged to `main`: https://github.com/hashicorp/design-system/blob/751b6a8d15130a05c2c606f0de2119f4652ec112/showcase/tests/acceptance/percy-test.js

dchyun:
> > > > This is something I caught when working on #2791. Can you add an acceptance test for the new page? I noticed there are no acceptance tests for the layouts pages, so we will need to make a `/layouts` folder as well.
> > > 
> > > 
> > > I don't think adding acceptance tests for Flex/Grid is useful because there are no elements with roles or styles we need to check for contrast since its just layout (but we should probably add acceptance tests for AppFrame since it has elements that aren't divs)
> > > cc @dchyun
> > 
> > 
> > That makes sense to me to not add the a11y checks. Should we still add the acceptance tests for making sure visiting the url works? I'll add the appframe test in a separate pr. CC: @shleewhite @KristinLBradley
> 
> @dchyun Do you mean the Percy tests? They're already included in the component PRs that were merged to `main`: https://github.com/hashicorp/design-system/blob/751b6a8d15130a05c2c606f0de2119f4652ec112/showcase/tests/acceptance/percy-test.js

No not percy. For each website page we seem to have an acceptance test that makes sure the url is visited correctly. 
```
test('visiting /components/alert', async function (assert) {
  await visit('/components/alert');

  assert.strictEqual(currentURL(), '/components/alert');
});
```

KristinLBradley:
> > > > > This is something I caught when working on #2791. Can you add an acceptance test for the new page? I noticed there are no acceptance tests for the layouts pages, so we will need to make a `/layouts` folder as well.
> > > > 
> > > > 
> > > > I don't think adding acceptance tests for Flex/Grid is useful because there are no elements with roles or styles we need to check for contrast since its just layout (but we should probably add acceptance tests for AppFrame since it has elements that aren't divs)
> > > > cc @dchyun
> > > 
> > > 
> > > That makes sense to me to not add the a11y checks. Should we still add the acceptance tests for making sure visiting the url works? I'll add the appframe test in a separate pr. CC: @shleewhite @KristinLBradley
> > 
> > 
> > @dchyun Do you mean the Percy tests? They're already included in the component PRs that were merged to `main`: https://github.com/hashicorp/design-system/blob/751b6a8d15130a05c2c606f0de2119f4652ec112/showcase/tests/acceptance/percy-test.js
> 
> No not percy. For each website page we seem to have an acceptance test that makes sure the url is visited correctly.
> 
> ```
> test('visiting /components/alert', async function (assert) {
>   await visit('/components/alert');
> 
>   assert.strictEqual(currentURL(), '/components/alert');
> });
> ```

@dchyun Added basic acceptance tests now, thanks.

KristinLBradley:
@dchyun @shleewhite The `Grid` docs have been merged into this PR so that they can be released together. Could you please re-review/approve?

dchyun (approved):

PR #2768, opened 2025-03-13
AdvancedTable - Declarative Template Updates

zamoore:
### :pushpin: Summary

If merged, this PR changes the DOM updates from an imperative model to a declarative one.

For instance, instead of deriving state from the DOM, this new approach derives DOM state from the data model.

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/components/hds/advanced-table/index.hbs
+++ b/packages/components/src/components/hds/advanced-table/index.hbs
@@ -74,7 +74,11 @@
           {{yield
             (hash
               Tr=(component
-                "hds/advanced-table/tr" isParentRow=T.isExpandable depth=T.depth displayRow=T.shouldDisplayChildRows
+                "hds/advanced-table/tr"
+                isLastRow=(eq this._tableModel.lastVisibleRow.id T.data.id)
+                isParentRow=T.isExpandable
+                depth=T.depth
+                displayRow=T.shouldDisplayChildRows
               )
               Th=(component
                 "hds/advanced-table/th"
@@ -100,6 +104,7 @@
             Tr=(component
               "hds/advanced-table/tr"
               selectionScope="row"
+              isLastRow=(eq this._tableModel.lastVisibleRow.id record.id)
               isSelectable=this.isSelectable
               onSelectionChange=this.onSelectionRowChange
               didInsert=this.didInsertRowCheckbox
--- a/packages/components/src/components/hds/advanced-table/index.hbs
+++ b/packages/components/src/components/hds/advanced-table/index.hbs
@@ -67,7 +67,6 @@
       {{#if this.hasNestedRows}}
         <Hds::AdvancedTable::ExpandableTrGroup
           @record={{record}}
-          @childrenKey={{this.childrenKey}}
           @rowIndex={{index}}
           @didInsertExpandButton={{this.didInsertExpandButton}}
           @willDestroyExpandButton={{this.willDestroyExpandButton}}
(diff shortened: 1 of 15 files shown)


packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -149,39 +168,6 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
-    }
-
-    for (const obj of model) {
-      if (this.childrenKey in obj) {
-        hasNestedRows = true;
-        break;
-      }
-    }
-
-    if (hasNestedRows) {
```
zamoore, line 175: I moved this assertion into the constructor since value is derived from the model.

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -91,20 +90,42 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
-  @tracked private _expandAllButtonState?: boolean | 'mixed' = undefined;
 
   private _selectableRows: HdsAdvancedTableSelectableRow[] = [];
-  private _expandableRows: HTMLButtonElement[] = [];
   private _captionId = 'caption-' + guidFor(this);
   private _intersectionObserver: IntersectionObserver | undefined = undefined;
   private _element!: HTMLDivElement;
+  private _tableModel!: HdsAdvancedTableTableModel;
+
+  constructor(owner: unknown, args: HdsAdvancedTableSignature['Args']) {
```
shleewhite, line 100: Alexey updated all the owners to use
```js
import type Owner from '@ember/owner';
```

packages/components/src/components/hds/advanced-table/index.hbs
```diff
@@ -42,50 +45,50 @@
             @tooltip={{column.tooltip}}
             @isVisuallyHidden={{column.isVisuallyHidden}}
             @isExpandable={{column.isExpandable}}
-            @onClickToggle={{this.onExpandAllClick}}
-            @isExpanded={{this._expandAllButtonState}}
-            @hasExpandAllButton={{this.hasNestedRows}}
-            @didInsertExpandButton={{this.didInsertExpandAllButton}}
-            @willDestroyExpandButton={{this.willDestroyExpandAllButton}}
-          >{{column.label}}</Hds::AdvancedTable::Th>
+            @onClickToggle={{this._tableModel.toggleAll}}
```
shleewhite, line 48: Suggestion: the type for this arg optionally accepts a newValue, since thats not needed anymore we could remove it.

shleewhite (commented):
Looks great! had a couple small comments, but this seems to be working great and is so much more straightforward.

packages/components/src/components/hds/advanced-table/models/row.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  @action
+  collapseAll() {
+    this.isOpen = false;
+    this.children.forEach((child) => child.collapseAll());
+  }
+
+  @action
+  onClickToggle() {
+    this.isOpen = !this.isOpen;
```
shleewhite, line 65: Issue: it looks like the tests did correctly find one bug - if start with a fully collapsed table and then manually expand each row, the expand all button doesn't update to `aria-expanded="true"`, it currently stays "mixed"

zamoore: Ah good point. I think I found a fix that addressed this.

PR #2787, opened 2025-03-24
`Card` - Add tag argument (HDS-4688)

KristinLBradley:
### :pushpin: Summary

If merged, this PR adds a `tag` argument to the `Card` component with accompanying Showcase example and tests.

**Showcase**: https://hds-showcase-git-hds-4688-add-card-tag-arg-hashicorp.vercel.app/components/card#tag

**See related PR for docs**: https://github.com/hashicorp/design-system/pull/2805

* Adds `@tag` argument to `Card` allowing choice of `div` (the default) or `li` tags.
* Adds example to Showcase comparing default `div` tag to using an `li` tag.
* Adds integration tests for `@tag` argument

### :link: External links

* PR for tag arg documentation: https://github.com/hashicorp/design-system/pull/2805
* Jira ticket: [HDS-4688](https://hashicorp.atlassian.net/browse/HDS-4688)

***

### 👀 Component checklist

[HDS-4688]: https://hashicorp.atlassian.net/browse/HDS-4688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/card/container.hbs
+++ b/packages/components/src/components/hds/card/container.hbs
@@ -10,6 +10,6 @@
   <div class={{this.classNames}} ...attributes>{{yield}}</div>
 {{else}}
   {{#let (element this.componentTag) as |Tag|}}
-    <Tag class={{this.classNames}} ...attributes>{{yield}}</Tag>
+    <Tag class={{this.classNames}} role={{if (eq this.componentTag "li") "listitem"}} ...attributes>{{yield}}</Tag>
   {{/let}}
 {{/if}}
--- a/packages/components/src/components/hds/card/container.hbs
+++ b/packages/components/src/components/hds/card/container.hbs
@@ -2,6 +2,14 @@
   Copyright (c) HashiCorp, Inc.
   SPDX-License-Identifier: MPL-2.0
 }}
-<div class={{this.classNames}} ...attributes>
-  {{yield}}
-</div>
+{{!
+  Dynamically generating an HTML tag in Ember creates a dynamic component class (with the corresponding tagName), while rendering
+  a plain HTML element requires less computing cycles for Ember (you will notice it doesn't add the `ember-view` class to it).
+}}
+{{#if (eq this.componentTag "div")}}
+  <div class={{this.classNames}} ...attributes>{{yield}}</div>
+{{else}}
+  {{#let (element this.componentTag) as |Tag|}}
+    <Tag class={{this.classNames}} ...attributes>{{yield}}</Tag>
+  {{/let}}
+{{/if}}
--- a/packages/components/src/components/hds/card/container.ts
+++ b/packages/components/src/components/hds/card/container.ts
@@ -5,20 +5,25 @@
 
 import Component from '@glimmer/component';
 import { assert } from '@ember/debug';
+
 import {
   HdsCardBackgroundValues,
   HdsCardLevelValues,
   HdsCardOverflowValues,
+  HdsCardTagValues,
 } from './types.ts';
+
 import type {
   HdsCardBackground,
   HdsCardLevel,
   HdsCardOverflow,
+  HdsCardTag,
 } from './types.ts';
 
 export const DEFAULT_LEVEL = HdsCardLevelValues.Base;
 export const DEFAULT_BACKGROUND = HdsCardBackgroundValues.NeutralPrimary;
 export const DEFAULT_OVERFLOW = HdsCardOverflowValues.Visible;
+export const DEFAULT_TAG = HdsCardTagValues.Div;
 export const AVAILABLE_LEVELS: string[] = Object.values(HdsCardLevelValues);
 export const AVAILABLE_BACKGROUNDS: string[] = Object.values(
   HdsCardBackgroundValues
@@ -26,6 +31,7 @@ export const AVAILABLE_BACKGROUNDS: string[] = Object.values(
 export const AVAILABLE_OVERFLOWS: string[] = Object.values(
   HdsCardOverflowValues
 );
+export const AVAILABLE_TAGS: string[] = Object.values(HdsCardTagValues);
 
 export interface HdsCardContainerSignature {
   Args: {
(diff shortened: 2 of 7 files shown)


MelSumner:
A couple of notes: 

1. Make sure to add a note that if the consumer uses the `li` tag, they need to wrap it in a `ul` element. 
2. We'll probably want to make sure to do that on the showcase as well, so a11y tests won't fail
3. Please make sure to run the linter and/or a11y tests locally, just to catch whether or not the `ul` is working as intended. 
4. The a11y section of the docs may need to be updated to `conditionally conformant` since the consumer is responsible for wrapping the cards in a `ul` element.

showcase/app/styles/showcase-pages/card.scss
```diff
@@ -25,4 +25,10 @@ body.components-card {
       border-radius: 8px;
     }
   }
+
+  .shw-component-card-list {
+    margin: 0;
+    padding: 0;
+    list-style-type: none;
```
dchyun, line 32: [Suggestion] Do you think it would be useful to add some styles to the component so `list-style-type: none;` get applied automatically to the card if someone is using the `li` tag?

KristinLBradley: We only provide the Card component currently, not any type of list-like container for the Cards so styling along those lines should be left to the consumer. I can include related guidance in the documentation however.

dchyun: Adding that style on the card, not a container, would still have the desired affect of removing the list item styling. But this is really just a small thing. I'm fine with it as is.

dchyun: Had no idea that could happen. Looks like [safari removes all list roles](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type#accessibility) if `list-type` is `none`, and this is intentional behavior.

I think it's probably easiest to let developers handle this styling themselves then.

showcase/app/templates/components/card.hbs
```diff
@@ -92,4 +94,24 @@
...
+      <SG.Item @label="Card using default div tag">
+        <Hds::Card::Container @level="mid" @hasBorder={{true}}>
+          <Shw::Placeholder @text="div" @width="200" @height="200" @background="transparent" />
+        </Hds::Card::Container>
+      </SG.Item>
+
+      <SG.Item @label="Card using list item tag">
+        <ul class="shw-component-card-list">
+          <Hds::Card::Container @level="mid" @hasBorder={{true}} @tag="li">
+            <Shw::Placeholder @text="div" @width="200" @height="200" @background="transparent" />
```
dchyun, line 110:
```suggestion
            <Shw::Placeholder @text="li" @width="200" @height="200" @background="transparent" />
```

dchyun (approved):

showcase/app/styles/showcase-pages/card.scss
KristinLBradley: I'd prefer to leave all related styling to consumers to handle, but if there's any way you think I could add to/improve guidance that would be great!

Docs PR: https://github.com/hashicorp/design-system/pull/2805

showcase/app/templates/components/card.hbs
KristinLBradley: Oops, forgot to update the text. Thanks!

dchyun (approved):

showcase/app/styles/showcase-pages/card.scss
shleewhite: Yah know.. thinking about it more, maybe we just add the explicit role just in case anyway. If people use this tag, they will likely add `list-style-type` so we could set them up for success?

packages/components/src/components/hds/card/container.hbs
```diff
@@ -2,6 +2,14 @@
...
\ No newline at end of file
+{{!
+  Dynamically generating an HTML tag in Ember creates a dynamic component class (with the corresponding tagName), while rendering
+  a plain HTML element requires less computing cycles for Ember (you will notice it doesn't add the `ember-view` class to it).
+}}
+{{#if (eq this.componentTag "div")}}
+  <div class={{this.classNames}} ...attributes>{{yield}}</div>
+{{else}}
+  {{#let (element this.componentTag) as |Tag|}}
+    <Tag class={{this.classNames}} ...attributes>{{yield}}</Tag>
```
KristinLBradley, line 13: @shleewhite So the way this is done now, even though only 2 choices are currently permitted (`div` or `li`) I'm not hard coding the optional tag choice of `li`. So if I add `role="listitem"` to it if consumers set the tag to `li` I'll have to refactor this or add some additional logic. I can look to do that though or else we could mention the issue in the documentation. What do you think?

showcase/app/styles/showcase-pages/card.scss
KristinLBradley: So adding it would require some refactoring though. See my comment here: https://github.com/hashicorp/design-system/pull/2787/files#r2025156511

KristinLBradley: Let me see...

packages/components/src/components/hds/card/container.hbs
KristinLBradley: Ok actually it's not difficult to add. I just need to update the related test as well. Will push up when done.

dchyun (approved):

packages/components/src/components/hds/card/container.ts
```diff
@@ -129,14 +110,7 @@ export default class HdsCardContainer extends Component<HdsCardContainerSignatur
...
 
-  /**
-   * Sets the level for the card
-   * Accepted values: visible, hidden
-   *
-   * @param overflow
-   * @type {HdsCardOverflow}
-   * @default 'visible'
-   */
+  // Sets the level for the card
```
didoo, line 113: @KristinLBradley this comment was not correct (it's actually setting the overflow). Maybe we could remove the comments that are redundant, and leave only the ones that are not obvoius?

PR #2788, opened 2025-03-25
`SideNav` - Conditionally set `aria-labelledby`

dchyun:
### :pushpin: Summary

If merged, this PR would set the `aria-labelledby` attribute in the `SideNav` toggle button, only if the `@ariaLabel` argument is not provided.

Currently in the `SideNav` component's toggle button, there is a hardcoded `aria-labelledby` value of `hds-side-nav-header` present. The `@ariaLabel` argument sets the `aria-label` on the toggle button, but when this argument is set, the existing `aria-labelledby` is not removed. This leads to both `aria-label` and `aria-labelledby` being present on the toggle button. Since `aria-labelledby` overrides `aria-label` for screenreaders, the provided `aria-label` value is never announced.

To correct this behavior, the functionality has been changed so the hardcoded `aria-labelledby` is only set when the `@ariaLabel` argument is not used. 

**Before**

Screenshot 2025-03-25 at 12 55 43 PM

**After**

Screenshot 2025-03-25 at 12 54 37 PM

### :link: External links

Jira ticket: [HDS-4549](https://hashicorp.atlassian.net/browse/HDS-4549)

***

### 👀 Component checklist

[HDS-4549]: https://hashicorp.atlassian.net/browse/HDS-4549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/dry-buses-eat.md
+++ b/.changeset/dry-buses-eat.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`SideNav` - Conditionally set `aria-labelledby` attribute for toggle button based on if `@ariaLabel` argument is provided.
--- a/packages/components/src/components/hds/side-nav/index.hbs
+++ b/packages/components/src/components/hds/side-nav/index.hbs
@@ -31,7 +31,7 @@
       {{! template-lint-enable no-invalid-interactive}}
       <Hds::SideNav::ToggleButton
         aria-label={{this.ariaLabel}}
-        aria-labelledby="hds-side-nav-header"
+        aria-labelledby={{unless this.ariaLabel "hds-side-nav-header"}}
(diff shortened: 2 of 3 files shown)


dchyun:
> Please add documentation so that users understand what's happening and how to use this component correctly. Thank you!

We have marked the `@ariaLabel` argument as [deprecated](https://helios.hashicorp.design/components/side-nav?tab=code#component-api). So, these changes I believe are only needed for teams that are still relying on that eventually deprecated argument. The docs already provide guidance on the preferred approach, using the `SideNav::Header` instead of `@ariaLabel`. I can add a note into the `@ariaLabel` component API anyway if we think that is needed.

MelSumner (approved):

shleewhite (approved):

PR #2792, opened 2025-03-26
[HDS-4345] - Fix double loading of flight icons in engines

RobbieTheWagner:
### :pushpin: Summary

This PR fixes an issue where flight icons would be loaded every time an engine was accessed.

This required updating the `load-sprite` instance initializer to check if the parent app had the icons loaded yet or not.

### :link: External links

Jira ticket: [HDS-4345](https://hashicorp.atlassian.net/browse/HDS-4345)

***

### 👀 Component checklist

[HDS-4345]: https://hashicorp.atlassian.net/browse/HDS-4345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/instance-initializers/load-sprite.ts
+++ b/packages/components/src/instance-initializers/load-sprite.ts
@@ -4,17 +4,27 @@
  */
 
 import config from 'ember-get-config';
-import type ApplicationInstance from '@ember/application/instance';
+import ApplicationInstance from '@ember/application/instance';
+import EngineInstance from '@ember/engine/instance';
+
+type AppInstanceWithSpriteFlag = ApplicationInstance & {
+  __flightIconsSpriteLoaded?: boolean;
+};
+
+type EngineInstanceWithSpriteFlag = EngineInstance & {
+  __flightIconsSpriteLoaded?: boolean;
+};
 
 export async function initialize(
-  appInstance: ApplicationInstance & {
-    __flightIconsSpriteLoaded?: boolean;
-  }
+  instance: AppInstanceWithSpriteFlag | EngineInstanceWithSpriteFlag
 ) {
+  const parentApp = getRootAppInstance(instance);
+
   if (
     config?.emberFlightIcons?.lazyEmbed &&
     // we use this flag to avoid loading the sprite multiple times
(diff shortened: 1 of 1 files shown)


alex-ju (commented):
Congrats on your first PR in this repo!

I saw that you applied the `release-candidate` label, presumably to validate this solution - were the results of testing this in products (e.g. `cloud-ui`, `atlas`) satisfactory?

How do you feel about setting a global variable (attached to `window`, similar to our previous approach where we attached it to the Ember instance) instead of querying the DOM?

zamoore (approved):
I like Alex's suggestion, but I'm also fine with this as is. Thanks!

RobbieTheWagner:
> Congrats on your first PR in this repo!

Thank you!

> I saw that you applied the `release-candidate` label, presumably to validate this solution - were the results of testing this in products (e.g. `cloud-ui`, `atlas`) satisfactory?

Yes, I tested this in cloud-ui, where the bug could be reproduced by switching between engines. I switched between engines and there was only one `.flight-sprite-container` in the DOM, which indicates it is working as expected and all icons were showing up in the app.

> How do you feel about setting a global variable (attached to `window`, similar to our previous approach where we attached it to the Ember instance) instead of querying the DOM?

I think querying the DOM is a little cleaner, since we essentially want to insert this into the DOM when it does not already exist in the DOM, so I think actually checking if it exists already seems perhaps a little more accurate than storing a flag, but I don't feel super strongly.

shleewhite (approved):

RobbieTheWagner:
I spoke to @alex-ju via Slack and he said we could go ahead and merge this.

alex-ju:
@RobbieTheWagner it looks like this PR is missing a changelog entry. you can add one with `pnpm changeset` (a patch on `@hashicorp/design-system-components`, we use [past tense for the changelog](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices))

PR #2794, opened 2025-03-26
Remove RSVP.all usage

zamoore:
### :pushpin: Summary

If merged, this PR removes a usage of RSVP.all in favor of Promise.all

### DIFF
--- a/packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
+++ b/packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
@@ -1,5 +1,3 @@
-import RSVP from 'rsvp';
-
 import type { Diagnostic as DiagnosticType } from '@codemirror/lint';
 import type { HdsCodeEditorSignature } from '../../hds-code-editor';
 import type { Extension, Text } from '@codemirror/state';
@@ -100,7 +98,7 @@ export default async function jsonLinter(
     { EditorView, keymap },
     { syntaxTree },
     { linter, lintGutter, lintKeymap },
-  ] = await RSVP.all([
+  ] = await Promise.all([
     import('@codemirror/view'),
     import('@codemirror/language'),
     import('@codemirror/lint'),


packages/components/src/modifiers/hds-code-editor/linters/json-linter.ts
```diff
@@ -100,7 +98,7 @@ export default async function jsonLinter(
     { EditorView, keymap },
     { syntaxTree },
     { linter, lintGutter, lintKeymap },
-  ] = await RSVP.all([
+  ] = await Promise.all([
```
KristinLBradley, line 101: Is this change mainly to remove the need for an import? Or are there other advantages?

zamoore: We don't have RSVP as a listed dependency. We shouldn't technically be able to import it and may not be able to in the future with yarn changes.

aklkv:
@zamoore just wanted to double check if with this change everything still works as expected as RSVP.all and Promise.all inner workings is a little different. Otherwise looks good

zamoore:
> @zamoore just wanted to double check if with this change everything still works as expected as RSVP.all and Promise.all inner workings is a little different. Otherwise I would be interested to hear if @RobbieTheWagner has any comments about this import technique, it's been a whle since I have done this type of imports but if you refer to sprite loader https://github.com/hashicorp/design-system/blob/main/packages/components/src/instance-initializers/load-sprite.ts#L19 it looks to do it deferently but more like something I remember I had to do to make dynamic imports work

@aklkv It does still work as expected. And yes - totally happy to get some other opinions.

RobbieTheWagner:
According to https://www.npmjs.com/package/ember-auto-import `Promise.all` should work here. However, we could also leverage `importSync` from https://www.npmjs.com/package/@embroider/macros if we wanted. I am okay with `Promise.all` personally.

RobbieTheWagner (approved):

shleewhite (approved):

PR #2808, opened 2025-04-02
AdvancedTable - @model ember-data support

zamoore:
### :pushpin: Summary

If merged, this PR allows for the usage of ember-data model instances as model data in the Advanced Table component.

We use duck typing to avoid bringing in ember-data as a direct dependency.

```
export const isEmberDataModel = (
  item: HdsAdvancedTableModelItem
): item is import('@ember-data/model').default => {
  return (
    'constructor' in item &&
    typeof item.constructor === 'function' &&
    'modelName' in item.constructor &&
    typeof item.constructor.modelName === 'string' &&
    'serialize' in item &&
    typeof item.serialize === 'function'
  );
};
```

If the data is an ember-data model, we can extract its attributes during serialization into an `HdsAdvancedTableTableModel`.

```
if (isEmberDataModel(row)) {
  const attrs: Record<string, unknown> = {};
  const modelClass = row.constructor as typeof Model;

  modelClass.eachAttribute((key: string) => {
    attrs[key] = (row as Model).get(key);
  });

  attrs['id'] = row.id;

  rowData = attrs;
}
```

### :link: External links

Jira ticket: [HDS-4749](https://hashicorp.atlassian.net/browse/HDS-4749)

***

### 👀 Component checklist

[HDS-4749]: https://hashicorp.atlassian.net/browse/HDS-4749?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zamoore:
Not implementing this for now as we have varied ember-data support across all of our downstream apps. Consumers will need to convert ember-data records to plain js objects.

PR #2812, opened 2025-04-04
Adding a voice and tone page

heatherlarsen:
### :pushpin: Summary

If merged, this PR will add a "Voice and tone" page adapted from the Marketing team's guidelines: https://hashicorp.atlassian.net/wiki/spaces/CCAS/pages/3377922084/Our+tone

👉🏻 Preview: https://hds-website-git-hl-docs-tone-voice-hashicorp.vercel.app/content/voice-tone

### ✔️ TODO

- [x] Create illustration for card
- [ ] ~Figure out how to hide the sidecar~ (This is being addressed in #2815) 

### 📷 Screenshots
Since the Sidecar is still broken (#2815), they won't display appropriately in the Vercel deployment. This is what it looks like locally at various screen sizes:
Screenshot 2025-04-17 at 10 55 22 AM

Screenshot 2025-04-17 at 10 55 39 AM

Screenshot 2025-04-17 at 10 55 57 AM

### :link: External links

Jira ticket: [HDS-4291](https://hashicorp.atlassian.net/browse/HDS-4291)

***

### 👀 Component checklist

[HDS-4291]: https://hashicorp.atlassian.net/browse/HDS-4291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


KristinLBradley (approved):

website/docs/content/voice-tone/index.md
```diff
@@ -0,0 +1,39 @@
...
+previewImage: assets/illustrations/content/voice-and-tone.jpg
+layout:
+  sidecar: false
+---
+
+**The HashiCorp voice** reflects our personality and is **what** we say to our users. Our voice stays consistent across all of our content.
+
+**Our tone** is how we speak and say things to our users. Tone sets the mood of your writing and can change depending on the type of content you’re creating.
+
+<Doc::Layout @spacing="12px">
```
didoo, line 14: [idea] general consideration: since this is a "special" (non-standard) page, we could also consider having a special custom content layout / visual treatment for it, think of the "principles" page for example (in which case, we can implement it per your design specs); if you instead want to keep it simple, as you've implemented it it's OK but I would probably use a CSS class instead of `Doc::Layout` so we can have a more responsive layout at small screens (stacked content, instead of side by side)
screenshot_4851

heatherlarsen: This is a great suggestion. I've given this a try and am curious to see what you think of the new approach.

KristinLBradley: The layout doesn't seem to be working quite right. The Do & Dont blocks stack at almost all browser window widths. They line up side by side only within a small range of widths. Maybe that's the intention though?

It seems to me like there should/could be a greater range of views in which they display side by side.

heatherlarsen: Unfortunately, it's because the sidecar is still displaying in the Vercel deployment. When I test it locally, the sidecar isn't there (which is the intention, but there's a bug being looked at in #2815), so they're side by side more often. I'll drop some screenshots in the description, but feel free to test locally as well.

KristinLBradley: I see, thanks for explaining

KristinLBradley (approved):
Approving so this can be merged.

I think the content is good but the responsive layout of the Do & Dont could still use refining. I don't think that's a blocker though & could be worked on later.

jorytindall (approved):
I agree with @KristinLBradley , the responsive layout still needs a bit of work, but I think the content is good. It's a combination of the viewport width changing, plus the sidecar being hidden that is causing things to jump around a bit. Let me know if you want  to pair on it, might be a slightly different way to approach it.
CleanShot 2025-04-18 at 15 57 08

heatherlarsen:
@jorytindall, thanks for taking a look! This is a byproduct of the broken sidecar. Once the sidecar is fixed, it won't jump back and forth like this. Here's what it looks like without the sidecar on local:
https://github.com/user-attachments/assets/e3366da3-d477-49fc-b0db-e5e2a3e21811

PR #2810, opened 2025-04-03
chore(showcase): `eslint` & `prettier`

aklkv:
### :pushpin: Summary

* add latest eslint and ~~prettier~~ setup
* bump `pnpm`

### :link: External links

Jira ticket: [HDS-XXX](https://hashicorp.atlassian.net/browse/HDS-XXX)
Figma file: [if it applies]

***

### 👀 Component checklist

- [ ] Percy was checked for any visual regression
- [ ] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

PR #2361, opened 2024-08-25
Compressed Image Assets

hashibot-hds:
Images automagically compressed by [Calibre](https://calibreapp.com)'s [image-actions](https://github.com/marketplace/actions/image-actions) ✨

Compression reduced images by <strong>25.6%</strong>, saving <strong>67.44 KB</strong>.

| Filename | Before | After | Improvement |
| -------- | ------ | ----- | ----------- |
| <code>website/public/assets/foundations/color/colors-border-examples.png</code> | 21.56 KB | 16.28 KB | -24.5% |
| <code>website/public/assets/foundations/color/colors-foreground-examples.png</code> | 38.08 KB | 27.75 KB | -27.1% |
| <code>website/public/assets/foundations/color/colors-semantic-tokens-accessibility-examples.png</code> | 58.16 KB | 42.11 KB | -27.6% |
| <code>website/public/assets/foundations/color/colors-status-examples.png</code> | 29.31 KB | 21.13 KB | -27.9% |
| <code>website/public/assets/foundations/color/colors-surface-examples.png</code> | 27.59 KB | 20.54 KB | -25.6% |
| <code>website/public/assets/foundations/color/colors-what-are-semantic-tokens.png</code> | 89.14 KB | 68.59 KB | -23.1% |

820 images did not require optimisation.

PR #2815, opened 2025-04-10
Fix sidecar in website

alex-ju:
### :pushpin: Summary

Fix the sidecar in `website` by ensuring the `layout` object exists. While the fix doesn't demystify the root cause it does seem to fix the immediate issue. I suspect a mix of race and `frontmatter` object override.

To test the fix run `ember s --environment=production` to get prember to serve static assets. `/about/principles` is a good page to check the sidecar doesn't render.

### :link: External links

Jira ticket: [HDS-4770](https://hashicorp.atlassian.net/browse/HDS-4770)

***

### 👀 Component checklist

[HDS-4770]: https://hashicorp.atlassian.net/browse/HDS-4770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/routes/show.js
+++ b/website/app/routes/show.js
@@ -106,10 +106,10 @@ export default class ShowRoute extends Route {
           }
         });
 
-        // check if there are special needs for the page layout
-        const hasCover = frontmatter?.layout?.cover ?? true;
         // TODO! probably we should also check if we have TOC data for the sidecar
-        const hasSidecar = frontmatter?.layout?.sidecar ?? true;
+        const layout = frontmatter.layout || {}; // fallback to an empty object if layout is undefined
+        const hasCover = layout.cover ?? true;
+        const hasSidecar = layout.sidecar ?? true;
         const showContentId = `show-content-${res.data.id
           .replace(/\/index$/, '')
           .replaceAll('/', '-')}`;


shleewhite (approved):

didoo:
@alex-ju I'm looking at this to see if I remember something more of what I did and why

didoo:
@alex-ju something very weird is happening here.

Compare the source code:
modified

With how it's compiled:
modified

I need to investigate more...

didoo:
@alex-ju I went back to a previous commit (https://github.com/hashicorp/design-system/pull/2361 and since then the logic has not been changed); then I looked at how the code would be compiled, and from what I see it's compiled as one would expect:
modified

So it must be something not related to our code but to the compilation itself. Could it be something introduced with the migration to pnpm? @aklkv can you think of something that could change so drastically the code compilation?

alex-ju:
> @alex-ju I went back to a previous commit (#2361 and since then the logic has not been changed); then I looked at how the code would be compiled, and from what I see it's compiled as one would expect

> So it must be something not related to our code but to the compilation itself. Could it be something introduced with the migration to pnpm? @aklkv can you think of something that could change so drastically the code compilation?

The compiled code is the same, from what I've seen, but the computed value for frontmatter is different – I saw the same results in my debugging with the ones you shared above. This makes me think that the issue has to do with various async tasks taking longer or shorter than before (potentially after the pnpm migration), but couldn't get to the bottom of it.

aklkv:
I finally had time to poke around, I was able to reproduce in current deployed version but if I check my #2810 website preview it seems to be just fine. I wanted to make sure I was not missing anything

aklkv:
@didoo you hunch is correct this type of check:

```
const hasCover = frontmatter?.layout?.cover ?? true;
```

there was a version of babel transpiler < 7.8, if left side is false, expression would return right hand value.

I think correct version here and also what happened in my branch somewhere I bumped babel and it just works now. Gotta love transpilers

aklkv:
I’ve bumped `babel` to the latest version across all packages as a precautionary measure.

didoo:
@alex-ju @aklkv the website issue with the sidenav has been fixed by #2884 so we can close this PR (and the associated ticket): https://hds-website-git-chore-babel-hashicorp.vercel.app/about/principles

PR #2818, opened 2025-04-14
fix: update files changelog automation missed

shleewhite:
### :pushpin: Summary

If merged, this PR would add the changelog entry back in that the version packages PR missed: https://github.com/hashicorp/design-system/pull/2817

### DIFF
--- a/.changeset/good-jars-attend.md
+++ b/.changeset/good-jars-attend.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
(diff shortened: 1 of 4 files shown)


KristinLBradley (approved):

alex-ju (approved):

alex-ju:
for context, we believe this had to do with the fact that we bumped to 4.18.1 while 2817 was not rebased – we'll see if it all goes back to normal after this 🤞

PR #2789, opened 2025-03-25
Showcase: add app frame example with generic advanced table

shleewhite:
### :pushpin: Summary

If merged, this PR adds a new example to the AppFrame page in the showcase to show how an AdvancedTable works in the context of a page.

Also adds a link to the full page example from the Advanced Table showcase page.

**[PREVIEW](https://hds-showcase-git-hds-4700-advanced-table-showcase-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-advanced-table)**

### :hammer_and_wrench: Detailed description

**Note:** I had to add `overflow: auto` to the main element in Mock::App component to prevent the Advanced Table from overflowing the page container.

### :camera_flash: Screenshots
Screenshot 2025-03-27 at 2 25 29 PM

### :link: External links

Jira ticket: [HDS-4582](https://hashicorp.atlassian.net/browse/HDS-4582)

***

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-4582]: https://hashicorp.atlassian.net/browse/HDS-4582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2822, opened 2025-04-14
AdvancedTable - AppFrame followup

didoo:
### :pushpin: Summary

Looking at https://github.com/hashicorp/design-system/pull/2789 I noticed a couple of possible improvements.

In this PR I have:
- moved the `iFrame` with the AdvancedTable AppFrame demo in the “AdvancedTable” showcase page
    - reason: is more consistent with the other "demo" sections, it provide an in-page example via the iFrame, and a link to the full page via the "external link" icon in the frame
- moved the  “overflow wrapping logic” from the AppFrame main container to a local wrapper of the GenericAdvancedTable demo component
    - reason: we're not touching the `main` container, making the example more similar to what consumers would have in production

⚠️ Notice: @shleewhite this made me think if the "problem" of having a wrapper with `overflow: auto` (or `min-width: 0`) would be a problem that consumer would have too. In which case, we should consider having the `hds-advanced-table__container` as `display: grid` or something similar, so this would work out of the box instead of having to taking care of the overflow themselves. @KristinLBradley what do you think?

Preview:
- [AdvancedTable showcase page demo](https://hds-showcase-git-advanced-table-app-frame-followup-hashicorp.vercel.app/components/advanced-table#demo)
- [AppFrame standalone demo](https://hds-showcase-git-advanced-table-app-frame-followup-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-advanced-table)

***

### 👀 Component checklist

### DIFF
--- a/showcase/app/styles/mock-components/app.scss
+++ b/showcase/app/styles/mock-components/app.scss
@@ -19,6 +19,7 @@
 }
 
 .mock-app-main-generic-advanced-table-wrapper {
+  // TODO remove after https://hashicorp.atlassian.net/browse/HDS-4751 is done 
   display: grid;
   height: 600px;
   overflow: auto;
--- a/showcase/app/styles/mock-components/app.scss
+++ b/showcase/app/styles/mock-components/app.scss
@@ -18,6 +18,8 @@
   margin-bottom: 12px;
 }
 
-.mock-app-main-generic-advanced-table {
+.mock-app-main-generic-advanced-table-wrapper {
+  display: grid;
   height: 600px;
+  overflow: auto;
 }
--- a/showcase/app/templates/components/advanced-table.hbs
+++ b/showcase/app/templates/components/advanced-table.hbs
@@ -8,18 +8,6 @@
 <Shw::Text::H1>AdvancedTable</Shw::Text::H1>
 <section data-test-percy>
 
-  <Shw::Text::Body>
-    <a
-      class="shw-frame__open-link"
-      href="/layouts/app-frame/frameless/demo-full-app-frame-with-advanced-table"
-      target="_blank"
-      rel="noopener noreferrer"
-    >
-      See an Advanced Table in the context of a full App Frame
-      <span class="sr-only">open the frame in a new window</span>
-    </a>
-  </Shw::Text::Body>
-
   <Shw::Text::H2>Basic AdvancedTable</Shw::Text::H2>
   <Shw::Text::H3>Keyboard interaction</Shw::Text::H3>
 
@@ -1082,6 +1070,17 @@
       </B.Tr>
     </:body>
   </Hds::AdvancedTable>
+
+  <Shw::Divider />
+
+  <Shw::Text::H2>Demo</Shw::Text::H2>
+
+  <Shw::Frame
+    @id="demo-full-app-frame-with-advanced-table"
+    @src="/layouts/app-frame/frameless/demo-full-app-frame-with-advanced-table"
+    @height="780"
+    @label="Advanced Table in the context of a full App Frame"
+  />
 
   <Shw::Divider />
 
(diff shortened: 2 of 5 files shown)


shleewhite:
@didoo agree about the overflow: auto thing.. this came up because the scroll bar was not showing up in the "right" context for the horizontal scrolling.

Another thing, we have [a ticket](https://hashicorp.atlassian.net/browse/HDS-4751) open to investigate the best way to set the height for the table. In the showcase, I had to cap the table at 600px to make sure the thead would stick which isnt ideal.

showcase/app/templates/components/advanced-table.hbs
```diff
@@ -8,18 +8,6 @@
 <Shw::Text::H1>AdvancedTable</Shw::Text::H1>
 <section data-test-percy>
 
-  <Shw::Text::Body>
-    <a
```
shleewhite, line 12: I think it would be useful to either leave the link at the top of the page or move the demo up higher. Designers mainly use the demo for their testing.

shleewhite (approved):
Changes look good, I just left the one suggestion.

showcase/app/styles/mock-components/app.scss
```diff
@@ -18,6 +18,8 @@
   margin-bottom: 12px;
 }
 
-.mock-app-main-generic-advanced-table {
+.mock-app-main-generic-advanced-table-wrapper {
+  display: grid;
```
shleewhite, line 22: testing locally, it looks like adding display grid is what makes the shadow appear correctly on initial page load.

KristinLBradley: Nice!

KristinLBradley (approved):

showcase/app/styles/mock-components/app.scss
didoo: @shleewhite is this override still needed? or you changed things on the AdvancedTable side and now it's not necessary anymore?

showcase/app/templates/components/advanced-table.hbs
```diff
@@ -8,18 +8,6 @@
 <Shw::Text::H1>AdvancedTable</Shw::Text::H1>
 <section data-test-percy>
```
didoo, line 10:
```suggestion

  <Shw::Text::Body>
    <a
      class="shw-frame__open-link"
      href="/layouts/app-frame/frameless/demo-full-app-frame-with-advanced-table"
      target="_blank"
      rel="noopener noreferrer"
    >
      See an Advanced Table in the context of a full App Frame
      <span class="sr-only">open the frame in a new window</span>
    </a>
  </Shw::Text::Body>
```

showcase/app/styles/mock-components/app.scss
shleewhite: It is still needed, it has not been updated in the component yet. Side note: when I was converting Explorer to Advanced Table, adding `display: grid` made the grid too tall sometimes and there would be a big gap between the thead and tbody. I ended up adding `max-height: max-content` to resolve the issue.

shleewhite (dismissed):

showcase/app/styles/mock-components/app.scss
didoo: > It is still needed, it has not been updated in the component yet

@shleewhite OK, so when it's updated we will have to remember to remove this extra wrapper and class. Do you have a PR or Jira ticket I can reference? so I can add a comment to avoid we forget

didoo:
@shleewhite I've rebased my branch on `hds-4777/revive-enterprise-nav` (#2837) to avoid conflicts in your PR (which is more urgent/important)

showcase/app/styles/mock-components/app.scss
didoo, line 22:
```suggestion
  // TODO remove after https://hashicorp.atlassian.net/browse/HDS-4751 is done 
  display: grid;
```

shleewhite (approved):

shleewhite (approved):

didoo:
@LilithJames-HDS after your branch has been merged to `main` I had to rebase; can you re-approve?

PR #2826, opened 2025-04-16
`CodeBlock` -  height toggle feature (HDS-4745)

KristinLBradley:
### :pushpin: Summary

If merged, this PR adds a height toggle control feature for expanding the height of the `CodeBlock` which appears when a `maxHeight` value is set that exceeds by the content height.

**Showcase**: https://hds-showcase-git-hds-4745-codeblock-height-toggle-hashicorp.vercel.app/components/code-block#height-limit

**Before**
Screenshot 2025-05-01 at 4 39 15 PM

**After**
Screenshot 2025-05-01 at 4 38 53 PM

### :link: External links

- Jira ticket: [HDS-4745](https://hashicorp.atlassian.net/browse/HDS-4745)
- [Figma file](https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/branch/OVPXOsakJpswhGdYMgoUye/HDS-Components-v2.0?node-id=78307-1949&p=f&t=VVgPncQTWbw3hv26-0)

***

### 👀 Component checklist

[HDS-4745]: https://hashicorp.atlassian.net/browse/HDS-4745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/styles/components/code-block/index.scss
```diff
@@ -125,13 +130,49 @@ $hds-code-block-code-padding: 16px;
...
+    position: absolute;
+    right: 0;
+    bottom: 0;
+    left: 0;
+  }
+
+  .hds-code-block__code {
+    // gradient to add contrast for button & indicate overflow
+    &::after {
+      position: sticky; // prevent gradient from scrolling together with content
```
KristinLBradley, line 151: This took awhile to figure out. Implementing the gradient backdrop in this way prevents it from overlapping the scrollbar and also keeps it stuck at the bottom of the container so that it won't scroll along with the other content.

packages/components/src/components/hds/code-block/index.hbs
```diff
@@ -30,4 +30,27 @@
       />
     {{/if}}
   </div>
+  {{#if this.showFooter}}
+    <div class="hds-code-block__overlay-footer">
```
KristinLBradley, line 34: The name of this container could maybe be rethought. I called it a "footer" based on an earlier design in which it looked more like a footer. In the current implemented design, that's no longer really the case though...

KristinLBradley:
Implemented designs are at a point where @hashicorp/hds-design feedback would be welcome.

(You can take a look at how the current behavior is looking ect. to inform decisions for finalizing designs.)

packages/components/src/components/hds/code-block/index.hbs
dchyun: Confirmed with design that we are still good with the "Footer" name

didoo:
A few thing that I've noticed in the showcase of this PR, and I don't see in the "production" showcase:

Highlight rectangles not in the right place
screenshot_4932

Something strange with the line numbers
screenshot_4933

Small offset of the highlight rectangles
screenshot_4934

dchyun:
> A few thing that I've noticed in the showcase of this PR, and I don't see in the "production" showcase:

@didoo Thanks for catching those. The line highlight issues have been fixed. For the line number issue inside the tabs, I do see this issue myself in production. The issue seems to be that the `setPrismCode` function does not get run when the code block becomes visible. 

I can address this issue in a separate pr, since it's not related to anything here, to not hold up Kristin's feature. It will require heavier structural changes. I tested adding an IntersectionObserver to trigger the update and that seems to fix it.

packages/components/src/styles/components/code-block/index.scss
dchyun: I had to revert back to the original padding styles, as the changes were causing issues with how prism calculates the line highlighting. 

I attacked the gradient styles to the footer element itself. Adding extra padding to the bottom of the code block when expanded worked to replicate the placement of the footer at the bottom of the code.

shleewhite (commented):
Suggestion: It would be nice to make the shadow not overlap the scrollbar. Advanced Table does it for the scroll indicators.. since youre already getting dimensions of the code block I think its worth it.

Screenshot 2025-05-05 at 4 07 02 PM

dchyun:
> > A few thing that I've noticed in the showcase of this PR, and I don't see in the "production" showcase:
> 
> @didoo Thanks for catching those. The line highlight issues have been fixed. For the line number issue inside the tabs, I do see this issue myself in production. The issue seems to be that the `setPrismCode` function does not get run when the code block becomes visible.
> 
> I can address this issue in a separate pr, since it's not related to anything here, to not hold up Kristin's feature. It will require heavier structural changes. I tested adding an IntersectionObserver to trigger the update and that seems to fix it.

I have a PR #2856 to fix the above issues, and branched off of here to avoid any conflicts. Don't think those fixes are needed to merge this feature, but if they look good can get them all together.

andgen404 (dismissed):
🎉

KristinLBradley:
> Suggestion: It would be nice to make the shadow not overlap the scrollbar. Advanced Table does it for the scroll indicators.. since youre already getting dimensions of the code block I think its worth it.
> 
> Screenshot 2025-05-05 at 4 07 02 PM

Could you please share what browser this is in? Before, I thought I had this working (a fix to prevent shadow overlap)

shleewhite:
> > Suggestion: It would be nice to make the shadow not overlap the scrollbar. Advanced Table does it for the scroll indicators.. since youre already getting dimensions of the code block I think its worth it.
> > Screenshot 2025-05-05 at 4 07 02 PM
> 
> @shleewhite Could you please share what browser this is in? Before, I thought I had this working (a fix to prevent shadow overlap)


@KristinLBradley it was chrome with scrollbars set to always visible as a macOS setting.

shleewhite (dismissed):

dchyun:
> > Suggestion: It would be nice to make the shadow not overlap the scrollbar. Advanced Table does it for the scroll indicators.. since youre already getting dimensions of the code block I think its worth it.
> > Screenshot 2025-05-05 at 4 07 02 PM
> 
> @shleewhite Could you please share what browser this is in? Before, I thought I had this working (a fix to prevent shadow overlap)

@KristinLBradley I originally had changed your structure to have the gradient attached to the button element, which caused this issue. I reverted back to your approach of using a pseudo element.

KristinLBradley:
> > > Suggestion: It would be nice to make the shadow not overlap the scrollbar. Advanced Table does it for the scroll indicators.. since youre already getting dimensions of the code block I think its worth it.
> > > Screenshot 2025-05-05 at 4 07 02 PM
> > 
> > 
> > @shleewhite Could you please share what browser this is in? Before, I thought I had this working (a fix to prevent shadow overlap)
> 
> @KristinLBradley I originally had changed your structure to have the gradient attached to the button element, which caused this issue. I reverted back to your approach of using a pseudo element.

@dchyun Thanks, was there a reason to change it before?

Btw, I'm now seeing unnecessary horizontal scroll bars in some Showcase examples that weren't there before.

image

packages/components/src/components/hds/code-block/index.ts
```diff
@@ -206,6 +201,15 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
       classes.push('line-numbers');
     }
 
+    if (this.showFooter) {
+      classes.push('hds-code-block--has-overlay-footer');
+    }
+
+    // Add a class if if the overlay footer is shown and the height is expanded
+    if (this.showFooter && this._isExpanded) {
```
KristinLBradley, line 209: Just wondered if this works if you remove `this.showFooter`? (i.e., is it necessary?)

packages/components/src/styles/components/code-block/index.scss
```diff
@@ -125,13 +126,63 @@ $hds-code-block-code-padding: 16px;
...
+    width: calc(100% + #{$hds-code-block-code-padding * 2});
+    height: $hds-code-block-code-footer-height;
+    background: linear-gradient(360deg, #0d0e12 37.07%, rgba(13, 14, 18, 25%) 100%);
+    transform: translateX(-$hds-code-block-code-padding);
+    content: "";
+    pointer-events: none;
+  }
+}
+
+.hds-code-block--has-overlay-footer-expanded {
```
KristinLBradley, line 161: The naming of this class name seems confusing. To me, it sounds like the footer itself is expanded somehow.

How about changing it to just ".hds-code-block--is-expanded"?

packages/components/src/styles/components/code-block/index.scss
```diff
@@ -125,13 +126,63 @@ $hds-code-block-code-padding: 16px;
...
+    // gradient element
+    position: sticky; // prevent gradient from scrolling together with content
+    bottom: -$hds-code-block-code-padding;
+    display: block;
+    width: calc(100% + #{$hds-code-block-code-padding * 2});
+    height: $hds-code-block-code-footer-height;
+    background: linear-gradient(360deg, #0d0e12 37.07%, rgba(13, 14, 18, 25%) 100%);
+    transform: translateX(-$hds-code-block-code-padding);
+    content: "";
+    pointer-events: none;
```
KristinLBradley, line 157: I determined that the unnecessary horizontal scrolling I was seeing was triggered by this pseudo element.

To resolve, try deleting the width and the transform styles here.

image

Then, on the `.hds-code-block__code` parent, set the right side padding to "0". I don't think this right side padding is ever actually necessary but double check if you can see a visual difference without it. (I couldn't.)

image

KristinLBradley: I think you can also add `margin-bottom: -16px;` to the ::after pseudo element to remove the excess space below it once you scroll to the bottom of the code and its position is no longer "stuck".

dchyun: This fix worked great! Thanks for discovering that.

packages/components/src/components/hds/code-block/index.ts
dchyun: I was able to remove it

dchyun:
> @dchyun Thanks, was there a reason to change it before?

@KristinLBradley the original approach was causing issues with the line highlighting that Prism.js needed to apply.

KristinLBradley:
> > @dchyun Thanks, was there a reason to change it before?
> 
> @KristinLBradley the original approach was causing issues with the line highlighting that Prism.js needed to apply.

Thanks, I think that was due to how I had restructured how the padding was applied?

andgen404 (approved):

KristinLBradley:
@dchyun Thanks for adding the fixes for the horizontal overflow issues. It's looking much better.

I am actually noticing now though that removing the right side padding does make a difference when you scroll all the way to the right of code that has horizontal overflow.
I'll try experimenting a bit to see if I can come up with any possible solutions.

KristinLBradley:
So Dylan and I are starting to uncover a few issues related to the gradient overlay used with the overflowing code content...

If you scroll horizontally, the gradient no longer covers the sides of the code content. If you try to compensate by stretching the width of the gradient overlay to be wider that 100% then it introduces unnecessary horizontal scroll bars when not needed.

These issues are related to styles adding padding around the code content. Restructuring this padding can fix the positioning issues with the gradient overlay but it then throws off how the the JS calculates positioning for line highlighting.

I'm not sure of the best way forward to solve these issues.

We could possibly decide that we don't mind the unnecessary horizontal scrollbar or else decide that we don't mind that the gradient doesn't cover the full width of the code in the case the user scrolls horizontally when needed.

Or else we could come up with a different design solution such as Andrew's original design which added a bottom footer to contain the height toggle button vs. having the button overlay the code content. Or else we could find another way to increase contrast with the overlaid button and code content such as by using a dropshadow behind the button or something else instead of the current gradient.

_**Issue 1:** Unnecessary horizontal scrollbar (if gradient is stretched to fix issue #2)_
image

_**Issue 2:** Gradient doesn't cover both sides if user scrolls horizontally_
image

majedelass:
I noticed this visual issue: If you scroll down when the height isn't toggled to auto:
image

Shouldn't the line touch all the way at the bottom if the line number is active? Because when you toggle the auto. height as on, it does the following (and should be consistent):
image

shleewhite:
> So Dylan and I are starting to uncover a few issues related to the gradient overlay used with the overflowing code content...
> 
> If you scroll horizontally, the gradient no longer covers the sides of the code content. If you try to compensate by stretching the width of the gradient overlay to be wider that 100% then it introduces unnecessary horizontal scroll bars when not needed.
> 
> These issues are related to styles adding padding around the code content. Restructuring this padding can fix the positioning issues with the gradient overlay but it then throws off how the the JS calculates positioning for line highlighting.
> 
> I'm not sure of the best way forward to solve these issues.
> 
> We could possibly decide that we don't mind the unnecessary horizontal scrollbar or else decide that we don't mind that the gradient doesn't cover the full width of the code in the case the user scrolls horizontally when needed.
> 
> Or else we could come up with a different design solution such as Andrew's original design which added a bottom footer to contain the height toggle button vs. having the button overlay the code content. Or else we could find another way to increase contrast with the overlaid button and code content such as by using a dropshadow behind the button or something else instead of the current gradient.
> 
> _**Issue 1:** Unnecessary horizontal scrollbar (if gradient is stretched to fix issue #2)_ image
> 
> _**Issue 2:** Gradient doesn't cover both sides if user scrolls horizontally_ image

@KristinLBradley  Not to throw another option in the ring - but the way the advanced table shadows work is that they are a sibling of the grid and then the height/width are set appropriately and its positioned absolutely on top of the grid. that way if you scroll it always looks correct.

dchyun:
> @KristinLBradley Not to throw another option in the ring - but the way the advanced table shadows work is that they are a sibling of the grid and then the height/width are set appropriately and its positioned absolutely on top of the grid. that way if you scroll it always looks correct.

@shleewhite The issue is that the gradient needs to be a child of the overflowed container, and have `position: sticky`. This is so it can avoid overlapping the line numbers, but move with the content as a user scrolls horizontally. 

We tried placing the gradient as a sibling, and using `z-index` to have the line numbers appear above, but then part of the code is exposed when scrolling horizontally. 

Screenshot 2025-05-08 at 11 40 44 AM

KristinLBradley:
@dchyun Just noticed an issue with line-wrapping:

image

dchyun:
> @dchyun Just noticed an issue with line-wrapping not working properly:
@KristinLBradley thanks for catching that. I added the `display: block` when line wrapping is enabled.

KristinLBradley:
> I noticed this visual issue: If you scroll down when the height isn't toggled to auto: image
> 
> Shouldn't the line touch all the way at the bottom if the line number is active? Because when you toggle the auto height as active, it does the following (and should be consistent): image

@majedelass There needs to be extra space at the bottom so that the "Show more code" button won't overlap the last few lines of code when you scroll to the bottom. The numbering does not extend to the bottom because the extra space at the bottom is just that (space), it's not extra lines of actual code content.

packages/components/src/styles/components/code-block/index.scss
```diff
@@ -262,3 +294,60 @@ $hds-code-block-code-padding: 16px;
...
+
+.hds-code-block--has-overlay-footer {
+  .hds-code-block__code::after {
+    // gradient element
+    position: sticky; // prevent gradient from scrolling together with content
+    bottom: -$hds-code-block-code-padding;
+    left: 0;
+    display: block;
+    height: $hds-code-block-code-footer-height;
+    margin: 0 0 -#{$hds-code-block-code-padding} -1000px;
```
shleewhite, line 324: Might be nice to add a comment here to remember why we set these values specifically.

shleewhite (approved):
Tested more and it seems like the gradient is working well now!

packages/components/src/styles/components/code-block/index.scss
didoo: Probably also link to the slack thread(s), just in case (there's a lot of info on what we tried, what worked/didn't work, etc)

shleewhite (approved):

andgen404 (approved):
🎉

PR #2832, opened 2025-04-19
fix(components): `LinkToExternal`

aklkv:
### :pushpin: Summary

Today we are referencing `<LinkToExternal/>` component and in the many cases it doesn't exists unless consumer has `ember-engines` as dependancy. In the `embroider/vite` world all components are evaluated at the build time and needs to resolve correct dependancy. If we were to run `hds` embroider/vite today it would result in an error about unresolved `LinkToExternal` component. In order to make this behaviour more controlled we are encapsulating requirements for `ember-engines` to be explicitly required only in situation when we pass `@isRouteExternal` as `true` in which case we do our check for required dependancy resigne component instance or through an error.

Shipping `ember-engines` is not an option for a couple of reasons:

* it has very interesting v1 era requirements which are impossible to satisfy in addon structure (nor should we)
* it is consumer land dependancy and we should leverage this tacktics of not locking users to  our version and letting consumer provide own copy only in situation when it's needed. I can see other optimizations in similar fashion but this is the most pressing one at the moment

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist


aklkv:
superseded by #2838

PR #2835, opened 2025-04-21
`Tag` - Prevent inheritance overrides for font styles

dchyun:
### :pushpin: Summary

This PR fixes an issue with the `Tag` where when the text is at a specific character length, the overflow tooltip is infinitely added and removed.

This issue was originally observed in Atlas. Where when applying filters to the organization run page, the tooltip button that is added when a tag's text goes beyond a max width, was being infinitely added and removed causing the tag to flash back and forth.

This issue is caused by the fact that the font of the tag's text is changing from when it is not overflowing and using a `span` vs. when it is overflowing and using a `button` element from the `Hds::TooltipButton`. 

In atlas there is a different default font used for `button` elements, that does not align to the `HDS` default font. This was causing the following cycle.
1. The text is rendered for a span and the text is set to overflow
2. The overflow causes the component to update to a tooltip button
3. The tooltip button has a narrower font, which causes the overflow to be removed 

Setting `font: inherit;` insures that all fonts are standardized when both the `button` element and `span` element is used.

Steps to reproduce in Atlas
1. Run atlas locally, and  go to the `/settings/runs` page.
5. Click on "add filters"
6. Add the filter Status -> Running -> Cost estimating
7. Observe the tag issue

### External Link

- [Atlas RC Test PR](https://github.com/hashicorp/atlas/pull/22922)

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/styles/components/tag.scss
+++ b/packages/components/src/styles/components/tag.scss
@@ -46,6 +46,7 @@ $hds-tag-border-radius: 50px;
   max-width: 166px; // account for excess horizontal padding of text in non-dismissible variant
   padding: 3px 10px 5px 10px;
   border-radius: inherit;
+  font: inherit;
 }
 
 .hds-tag__text-container {
--- a/packages/components/src/styles/components/tag.scss
+++ b/packages/components/src/styles/components/tag.scss
@@ -36,7 +36,17 @@ $hds-tag-border-radius: 50px;
 .hds-tag__dismiss-icon {
   width: 12px;
   height: 12px;
+}
+
+.hds-tag__dismiss-icon,
+.hds-tag__text {
   color: var(--token-color-foreground-primary);
+}
+
+.hds-tag__dismiss,
+.hds-tag__text,
+.hds-tag__link {
+  background-color: var(--token-color-surface-interactive);
 }
 
 .hds-tag__text,
@@ -44,6 +54,7 @@ $hds-tag-border-radius: 50px;
   flex: 1 0 0;
   max-width: 166px; // account for excess horizontal padding of text in non-dismissible variant
   padding: 3px 10px 5px 10px;
+  background-color: var(--token-color-surface-interactive);
   border-radius: inherit;
 }
 
@@ -69,7 +80,6 @@ $hds-tag-border-radius: 50px;
 
 .hds-tag__dismiss,
 .hds-tag__link {
-  background-color: var(--token-color-surface-interactive);
   cursor: pointer;
 
   &:hover,
--- a/packages/components/src/styles/components/tag.scss
+++ b/packages/components/src/styles/components/tag.scss
@@ -49,6 +49,10 @@ $hds-tag-border-radius: 50px;
   border-radius: inherit;
 }
 
+.hds-tag__text {
+  color: var(--token-color-foreground-primary);
+}
+
 .hds-tag__text-container {
   display: -webkit-box;
   overflow: hidden;
--- a/packages/components/src/styles/components/tag.scss
+++ b/packages/components/src/styles/components/tag.scss
@@ -45,8 +45,8 @@ $hds-tag-border-radius: 50px;
   flex: 1 0 0;
   max-width: 166px; // account for excess horizontal padding of text in non-dismissible variant
   padding: 3px 10px 5px 10px;
-  border-radius: inherit;
   font: inherit;
+  border-radius: inherit;
 }
 
 .hds-tag__text-container {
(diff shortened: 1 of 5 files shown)


didoo:
@dchyun maybe it would be worth to add a specific test/example to the showcase, to make sure there are no regressions in the future and this use case is correctly covered? (see for example how it's done in the [Tooltip](https://hds-showcase.vercel.app/components/tooltip#inheritance) or [RichTooltip](https://hds-showcase.vercel.app/components/rich-tooltip#inheritance))

dchyun:
> @dchyun maybe it would be worth to add a specific test/example to the showcase, to make sure there are no regressions in the future and this use case is correctly covered? (see for example how it's done in the [Tooltip](https://hds-showcase.vercel.app/components/tooltip#inheritance) or [RichTooltip](https://hds-showcase.vercel.app/components/rich-tooltip#inheritance))

@didoo Thanks for the suggestion. Added an inheritance section into the showcase page.

showcase/app/templates/components/tag.hbs
```diff
@@ -144,4 +144,68 @@
...
+      </div>
+    </SF.Item>
+  </Shw::Flex>
+
+  <Shw::Flex @label="Applied to the tag" as |SF|>
+    <SF.Item>
+      <Hds::Tag
+        @text="My text tag"
+        @onDismiss={{this.noop}}
+        class="hds-typography-display-400 hds-foreground-warning-on-surface"
```
didoo, line 183: [issue] There is something not working with these overrides (try to comment your fixes, and you will see that the color changes, but not the font).  What I suggest to do is to use a custom local class (see how it's done in the other examples linked before) so you can use the same class for all these tests, and try to override all these values: you will see how many are currently still leaking:
```
.xyz {
  color: #911ced;
  font-weight: 900;
  font-size: 18px;
  font-family: Verdana, Geneva, Tahoma, sans-serif;
  line-height: 2rem;
  background-color: #f9f2ff;
}
```

screenshot_4891

For me there is a deeper problem, which is that the `Hds::Text::Body` should be at `hds-tag__text-container` not at the top level of the `Tag` component, so it controls the style of the _actual_ text in the component, not rely on inheritance. 

If you want I can jump on a call and we can pair of this, if yuo're unsure of what I mean or want to discuss more.

didoo (commented):
Left a few comments, but I think I _may_ have found the root problem.

showcase/app/templates/components/tag.hbs
dchyun: Thanks for this deep dive into it. You got it spot on I think. I updated the structure to use the `Hds::Text::Body` at the `hds-tag__text-container` level, instead of on the top. It seems to prevent all of the inheritance issues.

One open question I have is how much to lock down the inheritance. Any number of possible styles could be introduced through the classes attached to the tag. For example, the background color is set on the outer `hds-tag`, and if a class is added through `...attributes` to this then it will override it. For now I've made sure the font styles remain unaffected, but that's it. 

Screenshot 2025-04-22 at 3 04 24 PM

packages/components/src/styles/components/tag.scss
```diff
@@ -37,20 +36,32 @@ $hds-tag-border-radius: 50px;
...
+
+.hds-tag__dismiss-icon,
+.hds-tag__text {
   color: var(--token-color-foreground-primary);
 }
 
+.hds-tag__dismiss,
+.hds-tag__text,
+.hds-tag__link {
+  background-color: var(--token-color-surface-interactive);
```
didoo, line 49: [todo] this has already been applied at line 57, so you can move the declaration under the `.hds-tag__dismiss` and remove this whole block.

packages/components/src/styles/components/tag.scss
```diff
@@ -37,20 +36,32 @@ $hds-tag-border-radius: 50px;
 .hds-tag__dismiss-icon {
   width: 12px;
   height: 12px;
+}
+
+.hds-tag__dismiss-icon,
+.hds-tag__text {
```
didoo, line 42: [for consideration] in theory you could duplicate the `color` declaration, move it under `.hds-tag__dismiss-icon` and `.hds-tag__text, .hds-tag__link`, and remove this block too. It's not 100% a "clean" alternative, because we're applying a `color` also to the `.hds-tag__link`, but it gets overwritten below (with higher specificity) so I'll leave it to you to decide what to do, keep it as is or duplicate/move the declaration.

didoo (commented):
Left a couple of comment, everything is minor.

packages/components/src/styles/components/tag.scss
dchyun: I've left it as is for now, Mainly for organization just because I think it's a little cleaner to show the dismiss icon and text are grouped to be the same color, and have all link colors handled in the link color section. But don't have a strong opinion either way if we want to go with the duplicate way.

didoo (approved):
LGTM

PR #2838, opened 2025-04-22
fix(components): `LinkToExternal` & make `ember-engines` optional

aklkv:
### :pushpin: Summary

Today we are referencing `<LinkToExternal/>` component and in the many cases it doesn't exists unless consumer has `ember-engines` as dependancy. In the `embroider/vite` world all components are evaluated at the build time and needs to resolve correct dependancy. If we were to run `hds` embroider/vite today it would result in an error about unresolved `LinkToExternal` component. In order to make this behaviour more controlled we are encapsulating requirements for `ember-engines` to be explicitly required only in situation when we pass `@isRouteExternal` as `true` in which case we do our check for required dependancy resigne component instance or through an error.

Shipping `ember-engines` is not an option for a couple of reasons:

* it has very interesting v1 era requirements which are impossible to satisfy in addon structure (nor should we)
* it is consumer land dependancy and we should leverage this tacktics of not locking users to  our version and letting consumer provide own copy only in situation when it's needed. I can see other optimizations in similar fashion but this is the most pressing one at the moment

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -120,7 +120,13 @@
     "webpack": "^5.97.1"
   },
   "peerDependencies": {
-    "ember-source": "^3.28.0 || ^4.0.0 || ^5.3.0"
+    "ember-source": "^3.28.0 || ^4.0.0 || ^5.3.0",
+    "ember-engines": ">= 0.11.0"
+  },
+  "peerDependenciesMeta": {
+    "ember-engines": {
+      "optional": true
+    }
   },
   "ember": {
     "edition": "octane"
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -351,6 +351,7 @@
       "./helpers/hds-format-relative.js": "./dist/_app_/helpers/hds-format-relative.js",
       "./helpers/hds-link-to-models.js": "./dist/_app_/helpers/hds-link-to-models.js",
       "./helpers/hds-link-to-query.js": "./dist/_app_/helpers/hds-link-to-query.js",
+      "./helpers/hds-resolve-link-to-component.js": "./dist/_app_/helpers/hds-resolve-link-to-component.js",
       "./instance-initializers/load-sprite.js": "./dist/_app_/instance-initializers/load-sprite.js",
       "./modifiers/hds-advanced-table-cell.js": "./dist/_app_/modifiers/hds-advanced-table-cell.js",
       "./modifiers/hds-advanced-table-cell/dom-management.js": "./dist/_app_/modifiers/hds-advanced-table-cell/dom-management.js",
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -351,7 +351,6 @@
       "./helpers/hds-format-relative.js": "./dist/_app_/helpers/hds-format-relative.js",
       "./helpers/hds-link-to-models.js": "./dist/_app_/helpers/hds-link-to-models.js",
       "./helpers/hds-link-to-query.js": "./dist/_app_/helpers/hds-link-to-query.js",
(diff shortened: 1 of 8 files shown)


aklkv:
tested on:

* `cloud-ui` - `embroider/webpack` + `ember-engines`
* `terraform-enterprise` - `embroider/vite` no `ember-engines`

packages/components/src/components/hds/breadcrumb/item.ts
```diff
@@ -27,6 +35,25 @@ export interface HdsBreadcrumbItemSignature {
...
+
+    if (this.args.isRouteExternal) {
+      if (macroCondition(dependencySatisfies('ember-engines', '*'))) {
+        // @ts-expect-error: shape is unknown
+        this.linkToComponent = importSync(
+          'ember-engines/components/link-to-external-component.js'
+        ).default as LinkTo;
+      } else {
+        assert(
+          `@isRouteExternal is only available when using the "ember-engines" addon. Please install it to use this feature.`,
```
didoo, line 51:
```suggestion
          `@isRouteExternal for "Hds::Breadcrumb::Item" is only available when using the "ember-engines" addon. Please install it to use this feature.`,
```

packages/components/src/components/hds/breadcrumb/item.ts
```diff
@@ -27,6 +35,25 @@ export interface HdsBreadcrumbItemSignature {
 }
 
 export default class HdsBreadcrumbItem extends Component<HdsBreadcrumbItemSignature> {
+  linkToComponent = LinkTo;
+
+  constructor(owner: Owner, args: HdsBreadcrumbItemSignature['Args']) {
+    super(owner, args);
+
+    if (this.args.isRouteExternal) {
```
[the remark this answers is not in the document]
aklkv: I had similar though, but I have decided to use it in js world

packages/components/src/helpers/hds-resolve-link-to-component.ts
```diff
@@ -0,0 +1,33 @@
+import { LinkTo } from '@ember/routing';
```
RobbieTheWagner, line 1:
```suggestion
import { LinkTo } from '@ember/routing';
import { helper } from '@ember/component/helper';
```

packages/components/src/helpers/hds-resolve-link-to-component.ts
```diff
@@ -0,0 +1,33 @@
...
+    } else {
+      assert(
+        `@isRouteExternal is only available when using the "ember-engines" addon. Please install it to use this feature.`,
+        false
+      );
+    }
+  }
+
+  return LinkTo;
+}
```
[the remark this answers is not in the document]
aklkv: moved to utils

RobbieTheWagner (approved):

packages/components/package.json
```diff
@@ -120,7 +120,13 @@
     "webpack": "^5.97.1"
   },
   "peerDependencies": {
-    "ember-source": "^3.28.0 || ^4.0.0 || ^5.3.0"
+    "ember-source": "^3.28.0 || ^4.0.0 || ^5.3.0",
+    "ember-engines": ">= 0.11.0"
+  },
+  "peerDependenciesMeta": {
```
zamoore, line 126: Just to be safe, did you try to install on a codebase without ember-engines?

zamoore (approved):

packages/components/package.json
didoo: https://github.com/hashicorp/design-system/pull/2838#issuecomment-2822648599

PR #2515, opened 2024-10-24
`Time` - component (HDS-3945)

KristinLBradley:
### :pushpin: Summary

If merged, this PR will add the `Time` component and service to HDS.

* **Showcase:** https://hds-showcase-git-hds-3945-date-time-display-hashicorp.vercel.app/components/time

#### Related:

* **Documentation branch:** https://github.com/hashicorp/design-system/pull/2557

#### Product tests:

* Atlas/Terraform: https://github.com/hashicorp/atlas/pull/21709
* HCP/Cloud-ui: https://github.com/hashicorp/cloud-ui/pull/11273

### :link: External links

* Jira ticket: [HDS-3945](https://hashicorp.atlassian.net/browse/HDS-3945)

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-3945]: https://hashicorp.atlassian.net/browse/HDS-3945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2839, opened 2025-04-23
Use luxon 3.x

RobbieTheWagner:
### :pushpin: Summary

This PR drops the 2.x version of luxon from dependencies.

I think the way it was defined consumers would always get 2.x, which has security vulnerabilities. If we do need to support 2.x, we could also make it a peerDep instead of dropping it, but I am not sure if there is a reason we need 2.x support or not.

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist

### DIFF
--- a/.changeset/chilled-phones-give.md
+++ b/.changeset/chilled-phones-give.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+Use luxon 3.x
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -70,7 +70,7 @@
     "ember-stargate": "^0.4.3",
     "ember-style-modifier": "^4.4.0",
     "ember-truth-helpers": "^4.0.3",
-    "luxon": "^2.3.2 || ^3.4.2",
+    "luxon": "^3.4.2",
     "prismjs": "^1.30.0",
     "sass": "^1.83.0",
     "tabbable": "^6.2.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -184,7 +184,7 @@ importers:
         specifier: ^4.0.3
         version: 4.0.3(ember-source@5.12.0(@glimmer/component@1.1.2(@babel/core@7.26.10))(@glint/template@1.5.2)(rsvp@4.8.5))
       luxon:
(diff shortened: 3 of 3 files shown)


didoo:
@RobbieTheWagner for reference: https://github.com/hashicorp/design-system/pull/2515#discussion_r1854452246 and https://github.com/hashicorp/design-system/pull/2515#discussion_r1866022247

/cc @alex-ju

RobbieTheWagner:
@didoo I took a look at those comments and atlas does not have luxon installed at all, so I am confused. I think the correct paths forward here are either just use 3.x like this PR proposes or we could make it a peerDep and allow whatever versions, but currently if anyone installs HDS and they do not specify their own version of luxon, this is forcing them to use 2.x which has vulnerabilities.

didoo:
> @didoo I took a look at those comments and atlas does not have luxon installed at all, so I am confused. I think the correct paths forward here are either just use 3.x like this PR proposes or we could make it a peerDep and allow whatever versions, but currently if anyone installs HDS and they do not specify their own version of luxon, this is forcing them to use 2.x which has vulnerabilities.

@alex-ju and @KristinLBradley worked on this and I'm sure have more context; would be a problem to wait for when they're back from PTO next week?

RobbieTheWagner:
> @alex-ju and @KristinLBradley worked on this and I'm sure have more context; would be a problem to wait for when they're back from PTO next week?

@didoo I think waiting should be fine, but I am not sure. I don't know what our requirements are for a timeline for having all the vulnerabilities resolved in atlas though. If the main reason we were trying to support 2.x was for atlas, I don't think we need it anymore, and in the meantime we are forcing all consumers of HDS to install the old 2.x version with the vulnerabilities, since this is not a peerDep.

zamoore (approved):

MelSumner (approved):

RobbieTheWagner:
@didoo do you still think we should wait to merge this or are the approvals from @zamoore and @MelSumner sufficient to merge?

didoo:
> @didoo do you still think we should wait to merge this or are the approvals from @zamoore and @MelSumner sufficient to merge?

@RobbieTheWagner we discussed this today in our HDS Engineering sync, and if @MelSumner and @zamoore approve, it's OK to merge. 👍

PR #2840, opened 2025-04-23
Update ember-stargate to 0.5.0

RobbieTheWagner:
### :pushpin: Summary

This PR updates ember-stargate to 0.5.0.

This allows updating ember-stargate and ember-resources in consuming apps and gets us on newer versions to allow us to continue iterating toward latest. While I was making this change, I noticed we had some webpack peerDeps missing, so I also installed those.

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -67,7 +67,7 @@
     "ember-get-config": "^2.1.1",
     "ember-modifier": "^4.2.0",
     "ember-power-select": "^8.6.2",
-    "ember-stargate": "^0.4.3",
+    "ember-stargate": "^0.5.0",
     "ember-style-modifier": "^4.4.0",
     "ember-truth-helpers": "^4.0.3",
     "luxon": "^2.3.2 || ^3.4.2",
@@ -116,7 +116,8 @@
     "stylelint-config-rational-order": "^0.1.2",
     "stylelint-config-standard-scss": "^14.0.0",
     "typescript-eslint": "^8.29.0",
-    "typescript": "^5.8.2"
+    "typescript": "^5.8.2",
+    "webpack": "^5.97.1"
   },
   "peerDependencies": {
     "ember-source": "^3.28.0 || ^4.0.0 || ^5.3.0"
--- a/packages/ember-flight-icons/package.json
+++ b/packages/ember-flight-icons/package.json
@@ -65,7 +65,8 @@
     "prettier": "^3.4.2",
     "rollup": "^4.28.1",
     "rollup-plugin-copy": "^3.5.0",
-    "typescript": "^5.7.2"
+    "typescript": "^5.7.2",
+    "webpack": "^5.97.1"
   },
   "ember": {
     "edition": "octane"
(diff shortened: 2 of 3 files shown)


didoo:
A couple of questions:
- what will happen to consumers that are using older versions of this package? https://github.com/search?q=%22ember-stargate%22+user%3Ahashicorp+path%3A**%2Fpackage.json&type=code&ref=advsearch
- do you know if this is still a problem, in the current configuration? https://github.com/hashicorp/design-system/blob/main/packages/components/unpublished-development-types/global.d.ts (was added in this PR: https://github.com/hashicorp/design-system/pull/2102)

RobbieTheWagner:
@didoo 

> * what will happen to consumers that are using older versions of this package? https://github.com/search?q=%22ember-stargate%22+user%3Ahashicorp+path%3A**%2Fpackage.json&type=code&ref=advsearch

A lot of these seem to be on much older versions like 0.2.0, so are already not in sync with the version we have here. This is another one we could fix with a peerDep and allow apps to install their own versions, if that is what we want. The only breaking change in ember-stargate 0.5.0 is it requires Ember 3.24+, so anything on older Ember than that would have problems, but it looks like HDS only supports Ember 3.28+ already, is that correct?

> * do you know if this is still a problem, in the current configuration? https://github.com/hashicorp/design-system/blob/main/packages/components/unpublished-development-types/global.d.ts (was added in this PR: [`Sidenav` - Convert to TypeScript #2102](https://github.com/hashicorp/design-system/pull/2102))

The 3.28 tests seem to be passing, so I do not think it is a problem.

packages/components/unpublished-development-types/global.d.ts
```diff
@@ -26,7 +26,6 @@ declare module '@glint/environment-ember-loose/registry' {
       EmbroiderUtilRegistry /*, other addon registries */ {
     // local entries
     LinkToExternal: typeof LinkTo;
-    // we have to use `ember-stargate` version `0.4.3` because version `0.5.0` causes a break in the tests for `ember-lts-3.28`
```
MelSumner, line 29: This seems like an important reason to not make this change RN

didoo:
@zamoore can you have a look also at this PR? thanks

packages/components/unpublished-development-types/global.d.ts
RobbieTheWagner: I installed this in nomad and all tests are still passing there as well https://github.com/hashicorp/nomad/pull/25750

I would propose we remove the comment completely, as I don't think it provides any relevant details at this time, but if there is a new comment you would like there, I would be happy to add it, just let me know!

didoo: @MelSumner if the tests are passing, there's no need anymore of the comment (I was the one that added it, and at that time the tests were not passing).

didoo (approved):

MelSumner (approved):
My concerns have been addressed, thank you! :shipit:

PR #2846, opened 2025-04-29
`Modal`: fix issue with click outside to dismiss

shleewhite:
### :pushpin: Summary

If merged this PR would update Modal to not use `ember-focus-trap` to dismiss the Modal on click outside and instead use our own click event.

It also fixes an issue where the scroll bar would get added back if try to close the modal when `isDismissDisabled=true`.

This is necessary because the `ember-focus-trap`'s `clickOutSideDeactivates` only ever runs once. If a consumer conditionally disables the ability to dismiss the modal, the user may not be able to click outside to deactivate it.

Steps to reproduce:
1. Navigate to: https://hds-showcase.vercel.app/components/modal#demo
2. Click "Open non-dismissable modal"
3. Click outside the modal after it opens, notice it doesn't close and you are able to scroll the page behind the modal.
4. Click "reset isDismissDisabled"
5. Click outside the modal, notice it doesnt close.
6. Click "confirm"/"Cancel"/"close", notice the modal does close. 

### :link: External links

Jira ticket: [HDS-3972](https://hashicorp.atlassian.net/browse/HDS-3972)

***

### 👀 Component checklist

[HDS-3972]: https://hashicorp.atlassian.net/browse/HDS-3972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/khaki-swans-type.md
+++ b/.changeset/khaki-swans-type.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Modal` - Fixed issue where conditionall disabling ability to dismiss the `Modal` breaks click outside to dismiss.
--- a/.changeset/khaki-swans-type.md
+++ b/.changeset/khaki-swans-type.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-`Modal` - Fixed issue where conditionall disabling ability to dismiss the `Modal` breaks click outside to dismiss.
+`Modal` - Fixed issue where conditionally disabling ability to dismiss the `Modal` breaks click outside to dismiss.
--- a/packages/components/src/components/hds/modal/index.hbs
+++ b/packages/components/src/components/hds/modal/index.hbs
@@ -9,7 +9,7 @@
   {{did-insert this.didInsert}}
   {{will-destroy this.willDestroyNode}}
   {{! @glint-expect-error - https://github.com/josemarluedke/ember-focus-trap/issues/86 }}
-  {{focus-trap isActive=this._isOpen focusTrapOptions=(hash onDeactivate=this.onDismiss clickOutsideDeactivates=false)}}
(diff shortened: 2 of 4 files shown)


didoo:
It's quite a complex PR with potential side-effects. Maybe @RobbieTheWagner could review it too?

zamoore:
Have you considered moving the logic out of the modal and into a global `on-click-outside` modifier? Might be a useful modifier to ship with HDS.

RobbieTheWagner:
> Have you considered moving the logic out of the modal and into a global `on-click-outside` modifier? Might be a useful modifier to ship with HDS.

I like this idea. FWIW, there is a well maintained addon for this, if we want to just use that https://github.com/lifeart/ember-click-outside-modifier

.changeset/khaki-swans-type.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Modal` - Fixed issue where conditionall disabling ability to dismiss the `Modal` breaks click outside to dismiss.
```
MelSumner, line 5:
```suggestion
`Modal` - Fixed issue where conditionally disabling ability to dismiss the `Modal` breaks click outside to dismiss.
```

shleewhite:
> > Have you considered moving the logic out of the modal and into a global `on-click-outside` modifier? Might be a useful modifier to ship with HDS.
> 
> I like this idea. FWIW, there is a well maintained addon for this, if we want to just use that https://github.com/lifeart/ember-click-outside-modifier

@RobbieTheWagner @zamoore This is definitely an interesting idea... I think for now because there really is only one use so far I'm leaning towards leaving the one off implementation. We usually rely on `ember-focus-trap` for this functionality, so I don't want to add another dependency.

dchyun (dismissed):
LGTM! Only suggestion, should we add a test case for the document click in the existing `isDIsmissDisabled` test?

dchyun (approved):

PR #2848, opened 2025-05-02
Foundations / Viewport breakpoints - Implementation

didoo:
### :pushpin: Summary

This PR introduces standard viewport breakpoints in HDS and makes them available to consumers via Sass (keys-values map and mixins) and JavaScript (keys-values map).

For details, see the memo document: https://docs.google.com/document/d/1gG-c9xUul3mwHvGUq7Z2NIWiM_-artjnjNCdGdrtLmI/

In this PR I have:
- added Sass file with breakpoint definitions/values and helper mixins under `foundations/breakpoints`
- added TypeScript with breakpoint definitions/values under `foundations/breakpoints`
- added a showcase page for `foundations/breakpoints`
- added percy test for breakpoints (frameless) page

 👉 👉 👉 **Preview**: https://hds-showcase-git-foundations-viewport-breakpoints-hashicorp.vercel.app/foundations/breakpoints

🧪 **Tested in both Cloud UI and Atlas**, by
- importing the Sass  mixin as `@use 'mixins/breakpoints' as *;` 
- importing the JS object as `import { hdsBreakpointValues } from '@hashicorp/design-system-components/utils/hds-breakpoints';`

ℹ️  Note: we have a separate PR to use the new breakpoints (and their helpers) in the existing HDS components: https://github.com/hashicorp/design-system/pull/2859  

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-4807

***

### 👀 Component checklist


showcase/app/styles/mock-components/demo/breakpoints/ruler.scss
```diff
@@ -0,0 +1,83 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+@use "sass:map";
+// from @hashicorp/design-system-components package
+// @use "mixins/breakpoints" as *;
```
didoo, line 8: @aklkv @RobbieTheWagner can you help me understand why this is not working? 

screenshot_4936

The same approach is used in Cloud UI ([see here](https://github.com/hashicorp/cloud-ui/blob/main/addons/core/package/src/styles/core/_overrides.scss#L11-L12)) so it should work here too. The only difference I can think of is that Cloud UI is using rollup while the showcase doesn't.

In any case, any idea how to get this working?

packages/components/rollup.config.mjs
```diff
@@ -28,6 +28,7 @@ const plugins = [
   // "app" tree. Things in here should also be in publicEntrypoints above, but
   // not everything in publicEntrypoints necessarily needs to go here.
   addon.appReexports([
+    'foundations/**/!(*types).js',
```
aklkv, line 31: why?

packages/components/src/foundations/breakpoints.ts
```diff
@@ -0,0 +1,12 @@
+// todo: understand if we can generate this list reading the correspondin Sass file
+// see: https://sergiocarracedo.es/2020/07/17/sharing-variables-between-scss-and-typescript/
+
+export const hdsBreakpointsValues = {
```
aklkv, line 4: in order to follow ember conventions this would be in utils folder

aklkv: additionally why we need this file? not everything needs to be a component or helper we can just use css

showcase/app/styles/_temp-hds-breakpoints.scss
```diff
@@ -0,0 +1,98 @@
+/**
```
aklkv, line 1: why are we not just consuming this from components package?

packages/components/src/foundations/breakpoints.ts
aklkv: also something like this should be possible as well https://medium.com/@miroslav_petrov/how-to-share-variables-between-scss-and-js-7b4225830abe

showcase/app/styles/mock-components/demo/breakpoints/ruler.scss
aklkv: this should work, maybe you have not rebuild components package?

aklkv: I have confirmed it just works

didoo: @aklkv I still see the same error, even after pulling in your comit:
screenshot_4939

aklkv: since test are passing makes me think that there is something wrong with you local setup, try deleting dist folder in components package and re build it

aklkv: nvm, I think I got it reproduced locally, I will check what is up

didoo: I suspect is something related to caching:
screenshot_4940

This is after I moved the brekpoints file from `foundations/` to `utils`, and yes I have deleted `dist`, and re-run `pnpm build` and/or `pnpm start` under the `packages/components`

aklkv: `pnpm i` the issues

didoo:
```
npx npkill -D -x -y
pnpm i
pnpm start
```

same error
screenshot_4941

aklkv: yes, `pnpm` sync dependancies on `pnpm install`, where as before it was symlink it. I am thinking maybe we should disable this behaviour. @RobbieTheWagner any thoughts

aklkv: I have confirmed that for existing files this strategy works fine they are being correctly linked and changes are propagated correctly but in cases where you rename/remove the file you need to reestablish the link by `pnpm i`. and this is what you are seeing after moving the file our as well

didoo: @RobbieTheWagner can you help me on this? I am kind of stuck.

aklkv: is there other issues you are facing after changing file location and re running `pnpm i`?

packages/components/rollup.config.mjs
didoo: why not? :)

aklkv:
I am having hard time reproducing what you have, but this no longer seems to be related to original issues and if I were to guess what is happening is that you are starting the showcase server to fast, because showcase has start command that triggers components build it might be interfering with already running components command. we can either remove that build command from showcase start or you can start showcase a little slower. I would like to move this conversation out of this PR as well as it's no longer related and it might be useful to here if anyone else in @hashicorp/hds-engineering has similar issues

packages/components/src/foundations/breakpoints.ts
didoo: > not everything needs to be a component or helper we can just use css

It will be needed in our JS implementation ([for the AppSideNav collapsing behaviour](https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/app-side-nav/index.ts#L36-L38)) and potentially/likely also consumers may need to access some of these values in JS if they need to implement logic related to media-queries

aklkv: I would lean towards keeping this manual and it changes a lot on regular bases we can look into automatic creation of this map from sass

packages/components/src/utils/hds-breakpoints.ts
```diff
@@ -0,0 +1,12 @@
+// todo: understand if we can generate this list reading the correspondin Sass file
+// see: https://sergiocarracedo.es/2020/07/17/sharing-variables-between-scss-and-typescript/
+
+export const hdsBreakpointsValues = {
```
aklkv, line 4: can we just do this: 

```
export const hdsBreakpointsNames = Object.keys({
  sm: { value: 480, px: '480px', rem: '30rem' },
  md: { value: 768, px: '768px', rem: '48rem' },
  lg: { value: 1088, px: '1088px', rem: '68rem' },
  xl: { value: 1440, px: '1440px', rem: '90rem' },
  xxl: { value: 1920, px: ' 1920px', rem: '120rem' },
});
```
are you planning to use `hdsBreakpointsValues` anywhere you might mutate the value?

packages/components/src/foundations/breakpoints.ts
didoo: > also something like this should be possible as well if you really need this file https://medium.com/@miroslav_petrov/how-to-share-variables-between-scss-and-js-7b4225830abe

yes, see my comment at line 2 in this file.

in any case I tried, and I didn't manage to get it working; it could not find the file (I used different paths, local/relative or the package path), and also TS was complaining about not using a mixin
modified

So I don't think is worth to spend too much time on this, it's a nice to have but low priority at the moment

packages/components/src/utils/hds-breakpoints.ts
KristinLBradley, line 4: Nit: For these kinds of variable names, how about renaming to remove the "s" from "Breakpoints"?

i.e., change to "hdsBreakpointValues"

When a word is used as an adjective like this it's usual to use the singular form instead of plural.

(Any other opinions on variable naming? @hashicorp/hds-engineering @hashicorp/hds-design )

didoo: Yeah, makes sense. Actually, I can even get rid of `hdsBreakpointsNames` since it's easily computed, so I am going to leave only the main map and use the name `hdsBreakpoints` (bonus, it's also consistent with the Sass map)

Changed in fe577fa3a85f58e711110d5f90eb0ae829c829a5

packages/components/src/utils/hds-breakpoints.ts
```diff
@@ -0,0 +1,10 @@
+// todo: understand if we can generate this list reading the correspondin Sass file
```
zamoore, line 1:
```suggestion
// todo: understand if we can generate this list reading the corresponding Sass file
```

showcase/app/components/mock/demo/breakpoints-ruler.gts
```diff
@@ -0,0 +1,23 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { hdsBreakpoints } from '@hashicorp/design-system-components/utils/hds-breakpoints';
+
+import ShwLabel from '../../shw/label';
+
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface MockDemoBreakpointsRulerSignature {}
```
zamoore, line 11: Question: Should this be removed until necessary?

zamoore (approved):
I really like your examples. The demo page is very clear and looks great.

dchyun (approved):
That showcase page is beautiful. 🎆

showcase/app/components/mock/demo/breakpoints-ruler.gts
didoo: Yes, nice catch (was a leftover)

dchyun (approved):

PR #2849, opened 2025-05-02
AdvancedTable - Column resizing

zamoore:
### :pushpin: Summary

If merged, this PR will allow users to resize AdvancedTable columns.

- Adds a new `column` model alongside the existing `table` and `row` models.
- Adds the `@hasResizableColumns` (boolean) to the index component API. This controls whether columns are resizable
- Add a `@onColumnResize` (function) argument to the index component. This function is called with column resize details when a column is resized.

#### Added components
- `Hds::AdvancedTable::ThContextMenu`: Right now just includes an option for resetting the column width to its initial value
Screenshot 2025-05-29 at 3 15 24 PM
- `Hds::AdvancedTable::ThResizeHandle`: A div that acts as a border handle for click-and-drag resizing on resizable `AdvancedTable` columns.
Screenshot 2025-05-29 at 3 14 40 PM

#### Added dependencies
- `ember-math-helpers`

toggle

### :link: External links

Jira ticket: [HDS-4585](https://hashicorp.atlassian.net/browse/HDS-4585)
Figma file: [Figma](https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/branch/5bRIc2dYHHYqRKwqratmFT/HDS-Components-v2.0?node-id=2-7&p=f&m=dev)

***

### 👀 Component checklist

[HDS-4585]: https://hashicorp.atlassian.net/browse/HDS-4585?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/components/hds/advanced-table/th-resize-handle.hbs
```diff
@@ -0,0 +1,15 @@
+<div
+  class="hds-advanced-table__th-resize-handle {{if this.resizing 'hds-advanced-table__th-resize-handle--resizing'}}"
+  draggable="false"
+  role="separator"
```
shleewhite, line 4: Issue: Ok this is super interesting, testing with VoiceOver: with role="separator", it doesn't announce the aria-label. It just calls it a "vertical splitter". It also announces the values as a percentage of the max.

When you set `role="slider"` (the implicit role for the input type="range"), it announces the aria-label and the values as the actual current value.

We should change this to `role='slider'`.

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -156,6 +156,7 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
   private _selectAllCheckbox?: HdsFormCheckboxBaseSignature['Element'] =
     undefined;
   @tracked private _isSelectAllCheckboxSelected?: boolean = undefined;
+  @tracked tableHeight = 0;
```
shleewhite, line 159: Suggestion: should this be private?

packages/components/src/components/hds/advanced-table/td.ts
```diff
@@ -22,6 +22,7 @@ export interface HdsAdvancedTableTdSignature {
   Args: {
     align?: HdsAdvancedTableHorizontalAlignment;
     rowspan?: number;
+    columnKey?: string;
```
shleewhite, line 25: What is this used for?

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -107,6 +109,48 @@ $hds-advanced-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px differenc
...
+
+      &::after {
+        position: absolute;
+        top: 0;
+        right: 0;
+        bottom: 0;
+        z-index: 2;
+        width: 1px;
+        background-color: $hds-advanced-table-border-color;
+        transform: translateX(-11.5px);
```
shleewhite, line 131: Could you add some comments on why these transform values specifically? (`-11.5px`, `12.5px`, `-10.5px`)?

packages/components/src/components/hds/advanced-table/th.ts
```diff
@@ -80,7 +81,7 @@ export default class HdsAdvancedTableTh extends Component<HdsAdvancedTableThSign
   }
 
   get align(): HdsAdvancedTableHorizontalAlignment {
-    const { align = DEFAULT_ALIGN } = this.args;
+    const { align = DEFAULT_ALIGN } = this.args.column ?? {};
```
shleewhite, line 84: when would column not be defined? the arg is required for the AdvancedTable, so wouldnt be the column model always exist?

showcase/app/templates/components/advanced-table.hbs
```diff
@@ -537,6 +537,34 @@
 
   <Shw::Divider @level={{2}} />
 
+  <Shw::Text::H2>Resizable columns</Shw::Text::H2>
```
shleewhite, line 540: Issue: could you add examples of the ThResizeHandle to the bottom of the showcase, like where the other base element examples are?

I was also thinking it would be good to have a sortable resize example... just since its 2 different Th components.

packages/components/src/components/hds/advanced-table/th.hbs
```diff
@@ -23,36 +23,48 @@
...
-            @onToggle={{@onClickToggle}}
-            @isExpanded={{@isExpanded}}
-            @isExpandAll={{@hasExpandAllButton}}
-            {{this._manageExpandButton}}
-          />
-        {{/if}}
-        <span id={{this._labelId}} class="hds-typography-body-200 hds-font-weight-semibold">{{yield}}</span>
-        <Hds::AdvancedTable::ThButtonTooltip @tooltip={{@tooltip}} @labelId={{this._labelId}} />
-      </div>
+  <Hds::Layout::Flex @justify="space-between" @align="center">
```
shleewhite, line 26: Should these layout changes happen for th-sort.hbs too?

packages/components/src/styles/components/advanced-table.scss
zamoore: Good call

packages/components/src/components/hds/advanced-table/th.ts
zamoore: We're not passing it to th components in the table body

packages/components/src/components/hds/advanced-table/th-resize-handle.hbs
zamoore: Nice! Done. Thanks for checking this out in voiceover

packages/components/src/components/hds/advanced-table/td.ts
zamoore: Good call. Nothing yet. That came over from the reordering feature by mistake.

showcase/app/templates/components/advanced-table.hbs
zamoore: Yes, makes sense to me

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -176,13 +178,28 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
     this._tableModel = new HdsAdvancedTableTableModel({
       model,
       childrenKey,
+      columns,
+      onColumnResize,
     });
 
+    if (this._tableModel.hasResizableColumns) {
+      assert(
+        'Cannot have a sticky first column if there are resizable columns.',
```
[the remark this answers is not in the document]
shleewhite: Oh I guess the scrolling indicators will also need to be updated when the column resizes
Screenshot 2025-05-30 at 4 52 36 PM

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -202,12 +255,15 @@ $hds-advanced-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px differenc
...
+  flex-shrink: 1;
   gap: 8px;
   align-items: center;
   justify-content: flex-start;
+  min-width: 0;
 }
 
-.hds-advanced-table__th-button {
+.hds-advanced-table__th-button,
+.hds-advanced-table__th-context-menu .hds-dropdown-toggle-icon {
```
jorytindall, line 266: Focusing on the context menu the focus ring is quite a bit thicker than other focusable elements.

Tooltip:
Screenshot 2025-05-30 at 4 00 20 PM

Context menu:
Screenshot 2025-05-30 at 4 00 36 PM

jorytindall: I'll note that this is only happening when I use a keyboard, when I use a mouse the width is what I would expect.

jorytindall: Maybe unrelated but another focus ring issue; when using a mouse, the internal focus ring border-radius is too sharp so it's mis-matched with the external. It also looks like there's a small gap in between the inner and outer ring.

Screenshot 2025-05-30 at 4 06 27 PM

jorytindall: There's also a slight mismatch in the dimensions.

Screenshot 2025-05-30 at 4 08 41 PM

packages/components/src/components/hds/advanced-table/th-context-menu.ts
```diff
@@ -0,0 +1,60 @@
...
+  get _options(): HdsAdvancedTableThContextMenuOption[] {
+    const { column } = this.args;
+
+    const options: HdsAdvancedTableThContextMenuOption[] = [];
+
+    if (column.isResizable) {
+      options.push({
+        key: 'reset-column-width',
+        label: 'Reset column width',
+        icon: 'reload',
```
jorytindall, line 47: [nit] Could you change this to the `rotate-ccw` icon instead of `reload`. The former kind of gives the impression of "rewind" or "go back" from the direction.

jorytindall (commented):
This is looking great! I left some pretty small nit-picky comments, let me know of questions around those.

I'm not sure where to leave this comment, but when sorting the table with resizable columns, the border radius/border style seems to get clipped a bit on the bottom right/left.
CleanShot 2025-05-30 at 16 14 16@2x

The AppFrame example used to have resizable enabled and I was noticing some visual bugs with the sticky column (I think @shleewhite caught them as well), but doesn't look like that argument is enabled anymore. Any chance I could test that a bit more?

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -107,6 +109,57 @@ $hds-advanced-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px differenc
...
+        width: 100%;
+        min-width: 30px;
+        overflow: hidden;
+        white-space: nowrap;
+        text-overflow: ellipsis;
+      }
+    }
+
+    .hds-advanced-table__th-resize-handle {
+      @include hds-focus-ring-with-pseudo-element($position: absolute);
```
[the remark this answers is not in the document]
zamoore: I played around with this a bit and I don't think we'll be able to get around this unfortunately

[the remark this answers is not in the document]
didoo: @jorytindall @zamoore what if the focus ring just touches the edges of the table (so remains within the overflow area)? I think it's better than being cut off:
screenshot_5011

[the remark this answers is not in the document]
jorytindall: That's a good compromise to me if it works!

packages/components/src/styles/components/advanced-table.scss
shleewhite: @zamoore I still see the issue with the thicker focus ring on the context menu.

zamoore: Should be resolved now

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -202,9 +253,46 @@ $hds-advanced-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px differenc
...
+  height: 24px;
+  margin: -2px 0;
+  padding: 0;
+  color: var(--token-color-foreground-faint);
+  background-color: transparent;
+  border: 1px solid transparent;
+  border-radius: 3px;
+
+  &:hover,
+  &.mock-hover {
```
dchyun, line 282:
```suggestion
  &.mock-hover,
  &:focus,
  &.mock-focus {
```
[Question] Should the focus state also have these hover state styles? Right now the focus state does not have the white background or black icon.
Screenshot 2025-06-05 at 1 21 40 PM

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -202,9 +253,46 @@ $hds-advanced-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px differenc
...
+    color: var(--token-color-foreground-primary);
+    background-color: var(--token-color-surface-interactive);
+    border-color: var(--token-color-border-strong);
+    box-shadow: var(--token-elevation-low-box-shadow);
+  }
+
+  &:focus,
+  &.mock-focus {
+    border-color: transparent;
```
dchyun: [Question] When the menu is open, the hover state for the button does not have a border. Should it match the hover state for when it is closed? This is noticeable when you click to open the menu and the border gets removed from the button.
Screenshot 2025-06-05 at 1 21 06 PM

showcase/app/templates/components/advanced-table.hbs
dchyun, line 1: [Suggestion] Can you add a section at the bottom for the `ThContextMenu` and show the different states as is done for the other buttons and icons?

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -107,6 +110,54 @@ $hds-advanced-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px differenc
         z-index: 1;
         isolation: isolate;
       }
+
+      &.hds-advanced-table__th--is-resizable .hds-advanced-table__th-content-text {
+        display: block;
+        width: 100%;
+        min-width: 30px;
+        overflow: hidden;
```
dchyun: [Suggestion] Not sure if this is doable, but could we apply these styles limiting the header text to one line to _all_
 column headers if any of the columns are resizable? In one example the last column is wrapping to two lines when the column before it is increased even though it doesn't need to. The table will overflow anyway so the column doesn't need to go to two lines. 
 
 https://github.com/user-attachments/assets/2aba6f5c-2a21-4fd2-bdbb-b68f8b6b2ab0

packages/components/src/components/hds/advanced-table/th-context-menu.ts
```diff
@@ -0,0 +1,60 @@
...
+  Args: {
+    column: HdsAdvancedTableColumn;
+  };
+  Element: HdsDropdownSignature['Element'];
+}
+
+export default class HdsAdvancedTableThContextMenu extends Component<HdsAdvancedTableThContextMenuSignature> {
+  originalColumnWidth?: string = this.args.column.width;
+
+  get classNames() {
```
dchyun: [Nit] since it's only one class could just add it in the template

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -202,9 +253,46 @@ $hds-advanced-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px differenc
...
   display: flex;
+  flex-shrink: 1;
   gap: 8px;
   align-items: center;
   justify-content: flex-start;
+  min-width: 0;
+}
+
+.hds-advanced-table__th-context-menu .hds-dropdown-toggle-icon {
+  width: 24px;
```
dchyun: [Suggestion] Might be worth creating a variable for this width and using it for all the buttons (tooltip, sort, expand)

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -202,9 +253,46 @@ $hds-advanced-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px differenc
...
+  height: 24px;
+  margin: -2px 0;
+  padding: 0;
+  color: var(--token-color-foreground-faint);
+  background-color: transparent;
+  border: 1px solid transparent;
+  border-radius: 3px;
```
dchyun:
```suggestion
  border-radius: var(--token-border-radius-x-small);
```

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -202,9 +253,46 @@ $hds-advanced-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px differenc
...
+  &.mock-focus {
+    border-color: transparent;
+
+    &::before {
+      top: 0;
+      right: 0;
+      bottom: 0;
+      left: 0;
+      border: none;
+      border-radius: 3px;
```
dchyun:
```suggestion
      border-radius: var(--token-border-radius-x-small);
```

packages/components/src/styles/components/advanced-table.scss
jorytindall: The sizing and things look good now, but when the menu is open, the focus ring isn't present until I press something on my keyboard. Here's a screen grab trying to express this (using my arrow keys)
CleanShot 2025-06-06 at 15 59 53

packages/components/src/styles/components/advanced-table.scss
jorytindall: Good question, I'm not overly opinionated, but I do think that leaving out the white background and the change in icon color on focus is fine.

packages/components/src/styles/components/advanced-table.scss
jorytindall: I dropped a follow-up comment about this on this [thread](https://github.com/hashicorp/design-system/pull/2849/files#r2116842541). When the menu is open, the focus ring should persist (at least that's how our other dropdown components function.

jorytindall (commented):
Followed up on a few threads, I'm seeing another bug with the focus state (that @dchyun caught as well), and I'm still seeing a discrepancy in the border radius on the resizable column sorting table when a column is sorted.

packages/components/src/components/hds/advanced-table/index.ts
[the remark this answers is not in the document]
zamoore: I talked with Jory and we're going to make this a follow-up bit of work.

packages/components/src/styles/components/advanced-table.scss
zamoore: Going to leave it for now as it matches the other buttons.

packages/components/src/components/hds/advanced-table/th-context-menu.ts
zamoore: Fair point

packages/components/src/styles/components/advanced-table.scss
zamoore: Good call

packages/components/src/styles/components/advanced-table.scss
zamoore: I went ahead and implemented this as suggested

packages/components/src/styles/components/advanced-table.scss
zamoore: Fixed in latest push

packages/components/src/components/hds/advanced-table/th-resize-handle.ts
```diff
@@ -0,0 +1,236 @@
...
+    startX: number;
+    startColumnPxWidth: number;
+    startNextColumnPxWidth?: number;
+  } | null = null;
+
+  private handleElement!: HdsAdvancedTableThResizeHandleSignature['Element'];
+  private boundResize: (event: PointerEvent) => void;
+  private boundStopResize: () => void;
+
+  private registerHandleElement = modifier(
```
dchyun:
```suggestion
  private _registerHandleElement = modifier(
```
[Suggestion] Since this is private should it use the `_` prefix?

packages/components/src/components/hds/advanced-table/models/column.ts
```diff
@@ -0,0 +1,144 @@
...
+function pxToNumber(pxString: string): number {
+  return parseFloat(pxString.slice(0, -2));
+}
+
+export default class HdsAdvancedTableColumn {
+  @tracked label: string = '';
+  @tracked align?: HdsAdvancedTableHorizontalAlignment = 'left';
+  @tracked isExpandable?: boolean = false;
+  @tracked isReorderable?: boolean = false;
+  @tracked isResizable?: boolean = false;
```
dchyun: [Suggestion] Running into some bugs related to this being set at the col level. Might be better to have it set at the table level and have all cols either be resizable or not. Also the last col could always not be resizable. 

In doing some testing playing around with the showcase, I found a couple UI bugs for a couple instances

When the last column is resizable:
- The table no longer overflows when expanding the second to last column
- Using the reset column width button on the last col can cause the table width to go below or above its initial width in certain situations

https://github.com/user-attachments/assets/19f750b9-2399-48ec-af61-69f7a28427e6

In _extreme_ situations things like this can occur
Screenshot 2025-06-11 at 1 41 19 PM

 

When there is a mixture of resizable columns and non-resizable columns, if the non-resizable columns are non-consecutive:
- Because both columns have a width of `1fr` they both grow and shrink at the same time
- This can give the appearance sometimes that a column you are not interacting with is shrinking 
  - Ex: The 2nd and 4th cols are not resizable, you use the resize handler in between col 1 and 2. Col 2 and col 4 shrink at the same time
 

https://github.com/user-attachments/assets/b7306131-f5ae-4710-a991-a0bfce1c8e4b

zamoore: We we're going to handle a lot of these scenarios through guidance on how to use the component, but I agree that just setting this at the table level would make more sense. I'm going to explore that option.

jorytindall:
Confirming that everything is looking resolved from a visual perspective 👍

showcase/app/components/mock/app/main/generic-advanced-table.gts
```diff
@@ -33,19 +33,22 @@ const SAMPLE_COLUMNS = [
     label: 'Project name',
     key: 'project-name',
     isSortable: true,
-    width: 'max-content',
+    isResizable: true,
```
dchyun: [Nit] Think these edits can be removed from this file now that the resizing is moved up to the table level.

showcase/app/templates/components/advanced-table.hbs
```diff
@@ -1945,4 +2003,59 @@
...
   </Shw::Grid>
 
+  <Shw::Divider @level={{2}} />
+
+  <Shw::Text::H3>ThContextMenu</Shw::Text::H3>
+
+  <Shw::Grid @label="Expand all interactive states + Expand state" @columns={{4}} as |SG|>
+    {{#each @model.STATES as |state|}}
+      <SG.Item @label={{capitalize state}}>
+        <Hds::Dropdown::Toggle::Icon
```
dchyun: [Suggestion] Could you use the `Hds::AdvancedTable::ThContextMenu` component here? Using the dropdown instead is causing the styles displayed to be different from those of the actual context menu

Screenshot 2025-06-13 at 2 29 49 PM

Screenshot 2025-06-13 at 2 29 59 PM

showcase/tests/integration/components/hds/advanced-table/index-test.js
```diff
@@ -126,6 +160,27 @@ const setNestedTableData = (context) => {
...
+const setResizableColumnsTableData = (context) => {
+  context.set('model', [
+    { id: '1', col1: 'A', col2: 'B' },
+    { id: '2', col1: 'C', col2: 'D' },
+  ]);
+  context.set('columns', [
+    {
+      key: 'col1',
+      label: 'Col 1',
+      isResizable: true,
```
dchyun: [Nit] Think the tests need to be updated to move the column resizing to the table level. This can be removed.

showcase/tests/integration/components/hds/advanced-table/index-test.js
```diff
@@ -175,6 +230,17 @@ const hbsNestedAdvancedTable = hbs`<Hds::AdvancedTable
   </:body>
 </Hds::AdvancedTable>`;
 
+const hbsResizableColumnsAdvancedTable = hbs`<Hds::AdvancedTable
+  @model={{this.model}} @columns={{this.columns}} id="resize-test-table"
```
dchyun:
```suggestion
  @model={{this.model}} @columns={{this.columns}} id="resize-test-table" @hasResizableColumns={{true}}
```
[Issue] Now that the resizing has been moved up to the table level it needs to be set here for the tests.

showcase/app/templates/components/advanced-table.hbs
zamoore: I can't unfortunately. I can't simulate the state of the toggle element from the parent component. I did fix the wrapper class so it renders correctly, though.

packages/components/src/components/hds/advanced-table/th-context-menu.hbs
```diff
@@ -0,0 +1,17 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+<Hds::Dropdown class="hds-advanced-table__th-context-menu" ...attributes as |D|>
```
jorytindall: [question] Should `@enableCollisionDetection` be set to true by default on this? Would that work properly within the scrollable content of the table?

showcase/app/templates/components/advanced-table.hbs
```diff
@@ -6,8 +6,9 @@
 {{page-title "AdvancedTable Component"}}
 
 <Shw::Text::H1>AdvancedTable</Shw::Text::H1>
-<section data-test-percy>
 
+<section data-test-percy>
+  import type { ActionState } from '@ember/-internals/glimmer/lib/modifiers/action';
```
dchyun:
```suggestion
```
[Question] Was this meant to go somewhere else?

showcase/app/templates/components/form/base-elements.hbs
```diff
@@ -8,7 +8,6 @@
 <Shw::Text::H1>Form / Base elements</Shw::Text::H1>
 
 <section data-test-percy>
-
```
dchyun: [Nit] Did you mean to update this form file?

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -200,43 +213,12 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
-    let hasCustomColumnWidth = false;
-
-    for (let i = 0; i < columns.length; i++) {
-      const column = columns[i];
-
-      if (column?.['width']) {
-        widths[i] = column.width;
-        if (!hasCustomColumnWidth) hasCustomColumnWidth = true;
-      }
+      assert(
```
dchyun, line 242: [Suggestion] Can you add a test for this assertion?

packages/components/src/components/hds/advanced-table/th-resize-handle.ts
dchyun: @zamoore think this got lost in other comments but bringing this back up.

.changeset/chilly-cheetahs-rescue.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+`AdvancedTable` - Added `isResizable` option to column object. When `true`, allows the column to be resized with both a click-and-drag and a keyboard interface.
```
dchyun: [Issue] Needs to be updated to reflect new argument structure.

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -109,6 +113,47 @@ $hds-advanced-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px differenc
       }
     }
 
+    .hds-advanced-table__th-resize-handle {
```
shleewhite, line 116: Suggestion: it could be nice to add a scroll margin to this so as it scrolls youre more likely to see the whole resize handle. Currently, it is always right on the edge

```
scroll-margin-right: 24px;
scroll-margin-left: 24px;
```

Screenshot 2025-06-19 at 4 12 25 PM

packages/components/src/components/hds/advanced-table/th-context-menu.hbs
```diff
@@ -0,0 +1,17 @@
...
+  <D.ToggleIcon @icon="more-vertical" @text="Context menu" @hasChevron={{false}} @size="small" />
```
shleewhite: Suggestion: We need to make the name of this menu unique for each column... it would be best to follow the pattern for what we do with the tooltip.

alos, even though we've been calling it context menu internally, I think "Additional actions" might mean more to people outside the team. 

```suggestion
  <D.ToggleIcon @icon="more-vertical" @text="Additional actions for {column label}" @hasChevron={{false}} @size="small" />
```

packages/components/src/components/hds/advanced-table/th-resize-handle.hbs
```diff
@@ -0,0 +1,17 @@
+{{! template-lint-disable no-pointer-down-event-binding }}
+<div
+  class={{this.classNames}}
+  draggable="false"
+  role="slider"
+  aria-orientation="vertical"
```
shleewhite: Even though the slider is super tall, the movement is still horizontal so horizontal is. a more appropriate value here.

```suggestion
  aria-orientation="horizontal"
```

packages/components/src/components/hds/advanced-table/th-resize-handle.ts
```diff
@@ -0,0 +1,260 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { modifier } from 'ember-modifier';
+
+import type HdsAdvancedTableColumn from './models/column';
```
shleewhite: Nit: I think we are supposed to use actual paths for imports now
```suggestion
import type HdsAdvancedTableColumn from './models/column.ts';
```

packages/components/src/components/hds/advanced-table/th-resize-handle.ts
```diff
@@ -0,0 +1,260 @@
...
+import type { HdsAdvancedTableSignature } from '.';
```
shleewhite: See comment above

```suggestion
import type { HdsAdvancedTableSignature } from './index.ts';
```

showcase/app/controllers/components/advanced-table.js
```diff
@@ -358,6 +358,44 @@ export default class ComponentsTableController extends Controller {
...
+    {
+      key: 'artist',
+      label: 'Artist',
+      tooltip: 'More information.',
+    },
+    {
+      key: 'album',
+      label: 'Album',
+      tooltip: 'More information.',
+      width: '350px',
```
shleewhite, line 372: Question: is the requirement that at least 1 column has an explicit pixel width?

showcase/app/components/mock/app/main/generic-advanced-table.gts
```diff
@@ -33,19 +33,19 @@ const SAMPLE_COLUMNS = [
     label: 'Project name',
     key: 'project-name',
     isSortable: true,
-    width: 'max-content',
+    width: '300px',
```
shleewhite: Question: is there a reason why only 3 columns have an explicit width now? Also, just in general going forward, will it be a problem for columns to have a width like 'max-width' set?

showcase/app/templates/components/advanced-table.hbs
```diff
@@ -543,6 +543,64 @@
 
   <Shw::Divider @level={{2}} />
 
+  <Shw::Text::H2>Resizable columns</Shw::Text::H2>
```
shleewhite, line 546: Suggestion: could you add an example where there is resizing + the cell content truncates? just to show what a consumer could do.

showcase/tests/integration/components/hds/advanced-table/index-test.js
```diff
@@ -306,6 +371,34 @@ module('Integration | Component | hds/advanced-table/index', function (hooks) {
     });
   });
 
+  test('it throws an assertion if @hasResizableColumns and has nested rows', async function (assert) {
```
shleewhite, line 374: If there are requirements for how many columns need pixel widths set in the columns argument, could you add assertions for that and a test?

packages/components/src/components/hds/advanced-table/models/table.ts
```diff
@@ -6,16 +6,25 @@
 import HdsAdvancedTableRow from './row.ts';
```
shleewhite, line 6: Suggestion: could you add a test or showcase example where the columns argument changes? Just to check that we dont have the same issue as the rows model.

packages/components/src/components/hds/advanced-table/th-sort.hbs
```diff
@@ -22,11 +22,48 @@
...
   ...attributes
 >
-  <div class="hds-advanced-table__th-content">
-    <span id={{this._labelId}} class="hds-typography-body-200 hds-font-weight-semibold">{{yield}}</span>
-    {{#if @tooltip}}
-      <Hds::AdvancedTable::ThButtonTooltip @tooltip={{@tooltip}} @labelId={{this._labelId}} />
-    {{/if}}
-    <Hds::AdvancedTable::ThButtonSort @sortOrder={{@sortOrder}} @onClick={{@onClickSort}} @labelId={{this._labelId}} />
-  </div>
+  <Hds::Layout::Flex @justify="space-between" @align="center" @gap="8">
```
shleewhite, line 25: Praise: love that youre using the flex component!

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -120,20 +120,22 @@ export interface HdsAdvancedTableSignature {
...
     childrenKey?: string;
     maxHeight?: string;
+    onColumnResize?: (columnKey: string, newWidth?: string) => void;
+    onSelectionChange?: (
+      selection: HdsAdvancedTableOnSelectionChangeSignature
+    ) => void;
+    onSort?: (sortBy: string, sortOrder: HdsAdvancedTableThSortOrder) => void;
   };
   Blocks: {
     body?: [
```
shleewhite, line 141: Nit: can you update the ComponentLikes to WithBoundArgs? I know we pass a lot of data to the children components.

showcase/app/controllers/components/advanced-table.js
zamoore: Not anymore. None of the columns need an explicit width defined.

showcase/tests/integration/components/hds/advanced-table/index-test.js
zamoore: See comment above

packages/components/src/styles/components/advanced-table.scss
zamoore: TIL!

packages/components/src/components/hds/advanced-table/th-resize-handle.hbs
zamoore: That is a great point

showcase/app/components/mock/app/main/generic-advanced-table.gts
zamoore: No, I can revert. It should work. Columns should still respect the max width set in CSS.

packages/components/src/components/hds/advanced-table/models/table.ts
zamoore: Definitely

showcase/app/templates/components/advanced-table.hbs
zamoore: Done

showcase/app/controllers/components/advanced-table.js
shleewhite: Nice! That simplifies things.

shleewhite (approved):
Nice!! Looking slick.

PR #2856, opened 2025-05-06
CodeBlock - Fix line wrapping and highlighting rendering

dchyun:
### :pushpin: Summary

[Preview](https://hds-showcase-git-dchyun-code-block-tabs-render-fix-hashicorp.vercel.app/components/code-block#demo)

If merged, this PR would fix rendering issues in the `CodeBlock` with the line wrapping and line highlighting. The primary issues found and resolved are.

- Line numbers are not displayed correctly when line wrapping is enabled
- Line numbers are not displayed correctly when the number of lines is updated dynamically
- Line numbers and line highlighting are not displayed correctly when line wrapping is enabled, and the code block is hidden from view initially (ex: inside a tab)
- Line highlighting is not displayed correctly when line numbers are not present

The `CodeBlock` was also converted from using `did-insert` and `did-update` to custom modifiers.

Originally in #2826 there were several issues observed in the `CodeBlock`. It was observed that when the code block was [used inside the tabs](https://hds-showcase.vercel.app/components/code-block#demo) the line numbers and line highlighting of the code blocks inside tabs initially hidden were not rendering correctly.

After additional investigation several issues with the Prism.js library were contributing to this problem.

#### Line numbers issues when line wrapping is enabled
First, line numbers are not being rendered correctly in the following situation

- `hasLineWrapping` is `true`
- no lines are highlighted
- there is only one `CodeBlock` on the page

The reason this issue has not been caught before is that when line highlighting is true, we trigger a `resize` event on the window to re-trigger Prism.js's line highlighting rendering. When there are other code blocks on a page that have line highlighting, this event was being listened to by code blocks without line highlighting as well. It would then re-trigger a Prism update, which would fix the line number issues.

If a page has no other code blocks with line highlighting to trigger this `resize` event, then the issues with the line numbers are not fixed.

#### Line number and highlight issues inside tabs

When the code block is initially hidden, such as being inside an inactive tab, the line numbers and highlighting are not rendered correctly when it is made visible, if `hasLineWrapping` is `true`.

When the code block is hidden on load, the `resize` event to re-trigger the prism.js updates does not correctly apply the styles to the hidden content. To solve this issue, we are now re-triggering Prism.js when the code block becomes visible using an `IntersectionObserver`.

#### Line highlighting with no line numbers

When [line highlighting is set](https://hds-showcase.vercel.app/components/code-block#highlight-lines) but `hasLineNumbers` is `false` the highlights start one line earlier than they should. This is due to Prism js calculating the line numbers based on the `pre` element when line numbers are present, but calculating it based on the `code` element when they are not ([source code link](https://github.com/PrismJS/prism/blob/v2/src/plugins/line-highlight/prism-line-highlight.ts#L92)).

#### Line number & wrapping issues

Note: This was only possible to show locally by only displaying one code block on the showcase page

**Before**
Screenshot 2025-05-05 at 6 50 51 PM

**After**
Screenshot 2025-05-05 at 6 50 22 PM

### Tabs issue

**Before**
Screenshot 2025-05-06 at 8 20 11 AM

**After**
Screenshot 2025-05-06 at 8 27 27 AM

#### Line highlighting with no line numbers

**Before**
Screenshot 2025-05-16 at 11 00 29 AM

**After**
Screenshot 2025-05-16 at 11 00 38 AM

### :link: External links

Jira ticket: [HDS-4875](https://hashicorp.atlassian.net/browse/HDS-4875)

***

### 👀 Component checklist

[HDS-4875]: https://hashicorp.atlassian.net/browse/HDS-4875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/selfish-apes-breathe.md
+++ b/.changeset/selfish-apes-breathe.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-`CodeBlock` - Fixed issues with line numbers when line wrapping is present and when the number of lines changes dynamically. Also fixed line highlighting when Code Block is hidden from view initially such as when used inside a Tabs component.
+`CodeBlock` - Fixed issues with line numbers when line wrapping is present and when the number of lines changes dynamically; line highlighting when the Code Block is hidden from view initially such as when used inside a Tabs component; and line highlighting when hasLineNumbers is false.
--- a/.changeset/selfish-apes-breathe.md
+++ b/.changeset/selfish-apes-breathe.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`CodeBlock` - Fixed issues with line numbers when line wrapping is present, and line highlighting when hidden from view initially such as inside a tabs component
--- a/.changeset/selfish-apes-breathe.md
+++ b/.changeset/selfish-apes-breathe.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-`CodeBlock` - Fixed issues with line numbers when line wrapping is present, and line highlighting when hidden from view initially such as inside a tabs component
+`CodeBlock` - Fixed issues with line numbers when line wrapping is present and when the number of lines changes dynamically. Also fixed line highlighting when Code Block is hidden from view initially such as when used inside a Tabs component.
--- a/packages/components/src/components/hds/code-block/index.hbs
+++ b/packages/components/src/components/hds/code-block/index.hbs
@@ -3,7 +3,7 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-<div class={{this.classNames}} ...attributes {{this._setUpObserver}}>
+<div class={{this.classNames}} ...attributes {{this._setUpCodeObserver}}>
   <div class="hds-code-block__header">
     {{~yield (hash Title=(component "hds/code-block/title"))~}}
     {{~yield (hash Description=(component "hds/code-block/description"))~}}
--- a/packages/components/src/components/hds/code-block/index.hbs
+++ b/packages/components/src/components/hds/code-block/index.hbs
@@ -3,7 +3,7 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-<div class={{this.classNames}} ...attributes>
+<div class={{this.classNames}} ...attributes {{this._setUpObserver}}>
   <div class="hds-code-block__header">
     {{~yield (hash Title=(component "hds/code-block/title"))~}}
     {{~yield (hash Description=(component "hds/code-block/description"))~}}
@@ -17,7 +17,7 @@
       data-start={{@lineNumberStart}}
       id={{this._preCodeId}}
       tabindex="0"
-    ><code {{did-insert this.setPrismCode}} {{did-update this.setPrismCode this.code @language}}>
+    ><code {{this._setUpCodeBlockCode}}>
         {{~this._prismCode~}}
       </code></pre>
 
--- a/packages/components/src/components/hds/code-block/index.ts
+++ b/packages/components/src/components/hds/code-block/index.ts
@@ -81,7 +81,9 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
 
   // If a code block is hidden from view, and made visible after load, the Prism code needs to be re-run
   private _setUpObserver = modifier((element: HTMLElement) => {
-    this._preCodeElement = element.querySelector('.hds-code-block__code') as HTMLPreElement;
+    this._preCodeElement = element.querySelector(
+      '.hds-code-block__code'
+    ) as HTMLPreElement;
     this._observer = new ResizeObserver((entries) => {
       entries.forEach((entry) => {
         if (entry.contentBoxSize) {
@@ -98,6 +100,7 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
   });
 
   private _setUpCodeBlockCode = modifier((element: HTMLElement) => {
+    this._isExpanded = false; // reset expanded state on updates
     this.setPrismCode(element);
     return () => {};
   });
@@ -169,6 +172,14 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
         } else {
           // eslint-disable-next-line @typescript-eslint/no-base-to-string
           this._prismCode = htmlSafe(Prism.util.encode(code).toString());
+        }
+
+        // Existing line numbers must be removed in order to be updated correctly
+        if (element.querySelector('.line-numbers-rows')) {
+          const lineNumbers = element.querySelector(
+            '.line-numbers-rows'
+          ) as HTMLElement;
+          element.removeChild(lineNumbers);
         }
 
         // Force prism-line-numbers plugin initialization, required for Prism.highlight usage
--- a/packages/components/src/components/hds/code-block/index.ts
+++ b/packages/components/src/components/hds/code-block/index.ts
@@ -100,7 +100,9 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
   });
 
   private _setUpCodeBlockCode = modifier((element: HTMLElement) => {
-    this._isExpanded = false; // reset expanded state on updates
+    if (this.showFooter) {
+      this._isExpanded = false; // reset expanded state on updates
+    }
     this.setPrismCode(element);
     return () => {};
   });
@@ -175,10 +177,10 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
         }
 
         // Existing line numbers must be removed in order to be updated correctly
-        if (element.querySelector('.line-numbers-rows')) {
-          const lineNumbers = element.querySelector(
-            '.line-numbers-rows'
-          ) as HTMLElement;
+        const lineNumbers = element.querySelector(
+          '.line-numbers-rows'
+        ) as HTMLElement;
+        if (lineNumbers) {
           element.removeChild(lineNumbers);
         }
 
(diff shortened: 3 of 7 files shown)


packages/components/src/components/hds/code-block/index.ts
```diff
@@ -75,6 +76,31 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
 
   // Generates a unique ID for the code content
   private _preCodeId = 'pre-code-' + guidFor(this);
+  private _preCodeElement!: HTMLPreElement;
+  private _observer!: IntersectionObserver;
+
+  // If a code block is hidden from view, and made visible after load, the Prism code needs to be re-run
+  private _setUpObserver = modifier((element: HTMLElement) => {
```
dchyun, line 83: I added an `IntersectionObserver` in order to detect when a code block becomes visible, to fix the issue when it is used inside tabs. It has the downside of also being ran anytime the code block is scrolled into view when it doesn't need to be ran. Wasn't sure of another way to detect for visibility changes.

packages/components/src/components/hds/code-block/index.ts
```diff
@@ -152,24 +178,26 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
...
-          // eslint-disable-next-line ember/no-runloop
-          schedule('afterRender', (): void => {
+          // Get the actual height & the content height of the preCodeElement
+          this._codeContentHeight = this._preCodeElement?.scrollHeight ?? 0;
+          this._codeContainerHeight = this._preCodeElement?.clientHeight ?? 0;
+
+          // we need to re-trigger the line numbers generation as late as possible to account for any line wrapping styles that are applied
+          if (this.args.hasLineWrapping && Prism?.plugins?.['lineNumbers']) {
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
+            Prism.plugins['lineNumbers'].resize(this._preCodeElement);
```
dchyun, line 190: We can't use the same logic as the line highlight and piggy back on triggering a `resize` event because the Prism.js line number plugin does [a check](https://github.com/PrismJS/prism/blob/44429cac5aa9e2ae56d8adab20195edb1496fd30/src/plugins/line-numbers/prism-line-numbers.ts#L206) to see if the viewport width actually changes on a `resize` event. It aborts any updates if the viewport is unchanged.

.changeset/selfish-apes-breathe.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`CodeBlock` - Fixed issues with line numbers when line wrapping is present, and line highlighting when hidden from view initially such as inside a tabs component
```
KristinLBradley, line 5:
```suggestion
`CodeBlock` - Fixed issues with line numbers when line wrapping is present. Also fixed line highlighting when Code Block is hidden from view initially such as when used inside a Tabs component.
```

packages/components/src/components/hds/code-block/index.ts
```diff
@@ -75,6 +76,31 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
 
   // Generates a unique ID for the code content
   private _preCodeId = 'pre-code-' + guidFor(this);
+  private _preCodeElement!: HTMLPreElement;
+  private _observer!: IntersectionObserver;
+
+  // If a code block is hidden from view, and made visible after load, the Prism code needs to be re-run
+  private _setUpObserver = modifier((element: HTMLElement) => {
+    this._preCodeElement = element.querySelector('pre') as HTMLPreElement;
+    const codeBlock = element.querySelector('code') as HTMLElement;
```
KristinLBradley, line 85: Would we want to reference the CodeBlock's id and/or reference class names to grab the `pre` and `code` elements here? (to better scope this if it makes a difference)

packages/components/src/components/hds/code-block/index.ts
didoo: CodePilot suggested this, which is not a bad idea (code adapted to our use case): 
```
    this._observer3 = new ResizeObserver((entries) => {
      entries.forEach((entry) => {
        if (entry.contentRect.width > 0 && entry.contentRect.height > 0) {
          console.log('_setUpObserver update invoked via ResizeObserver and element size changed', element, entry.contentRect);
          this.setPrismCode(codeBlock);
        }
      });
    });
    this._observer3.observe(element);
```
It suggested also to use `MutationObserver`, but that doesn't work.

The cool thing about the `ResizeObserver` is that it doesn't fire every time the element intersect the viewport (like with the `IntersectionObserver`) and _technically_ is more connected with what we're trying to do (eg. via the artificial triggering of the `resize` event)

Just an idea (but please explore it, test it and validate it: I just did a quick test)

packages/components/src/components/hds/code-block/index.ts
```diff
@@ -75,6 +76,31 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
...
+    this._observer = new IntersectionObserver((entries) => {
+      entries.forEach((entry) => {
+        if (entry.isIntersecting) {
+          this.setPrismCode(codeBlock);
```
didoo, line 89: [question] here you're re-running the entire `setPrismCode`. would it make sense to try to re-run only a part of the `setPrismCode` logic, on second runs?

didoo (commented):
@dchyun I did a quick first pass, but happy to do a second more in depth pass if/when you want

packages/components/src/components/hds/code-block/index.ts
dchyun: Great idea. Switching to a `ResizeObserver` worked, and now it doesn't fire every time a code block goes into view.

packages/components/src/components/hds/code-block/index.ts
dchyun: Scoped the pre element to the class instead. The code block element was not needed after a refactor.

packages/components/src/components/hds/code-block/index.hbs
```diff
@@ -3,7 +3,7 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
-<div class={{this.classNames}} ...attributes>
+<div class={{this.classNames}} ...attributes {{this._setUpObserver}}>
```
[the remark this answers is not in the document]
dchyun: Wasn't sure what to go with but went with `_setUpCodeObserver`. Open to other suggestions if we want to be more specific.

packages/components/src/components/hds/code-block/index.hbs
```diff
@@ -17,7 +17,7 @@
       data-start={{@lineNumberStart}}
       id={{this._preCodeId}}
       tabindex="0"
-    ><code {{did-insert this.setPrismCode}} {{did-update this.setPrismCode this.code @language}}>
+    ><code {{this._setUpCodeBlockCode}}>
```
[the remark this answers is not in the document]
dchyun: Did some testing and this will run both on update and insert. In testing, saw that there were issues with the line numbers not getting updated after an update both on this branch and in prod. Found a fix, and added a [showcase example](https://hds-showcase-1kafkdixo-hashicorp.vercel.app/components/code-block#demo) for this situation.

packages/components/src/components/hds/code-block/index.ts
```diff
@@ -75,6 +76,31 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
 
   // Generates a unique ID for the code content
   private _preCodeId = 'pre-code-' + guidFor(this);
+  private _preCodeElement!: HTMLPreElement;
+  private _observer!: IntersectionObserver;
+
+  // If a code block is hidden from view, and made visible after load, the Prism code needs to be re-run
+  private _setUpObserver = modifier((element: HTMLElement) => {
+    this._preCodeElement = element.querySelector('pre') as HTMLPreElement;
```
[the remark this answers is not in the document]
dchyun: Scoped it to the `hds-code-block__code` class instead of the element tag. Used the class name just because it felt cleaner than doing a `document.getElementById` call. The code block element reference wasn't needed anymore after refactoring.

packages/components/src/components/hds/code-block/index.ts
dchyun: Refactored to only run the code for updating the line number and line highlighting plugins in the observer.

didoo (approved):
Left a couple of comments/suggestions but nothing that is a blocker.

packages/components/src/components/hds/code-block/index.ts
```diff
@@ -75,6 +76,34 @@ export default class HdsCodeBlock extends Component<HdsCodeBlockSignature> {
...
+    });
+    this._observer.observe(element);
+
+    return () => {
+      this._observer.disconnect();
+    };
+  });
+
+  private _setUpCodeBlockCode = modifier((element: HTMLElement) => {
+    this._isExpanded = false; // reset expanded state on updates
```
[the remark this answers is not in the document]
dchyun: Was able to run this check only when `showFooter` is true and it worked.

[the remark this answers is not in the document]
dchyun: After pushing this change saw that this was causing test timeouts. Calling the showFooter function, which accesses tracked variables, in the set up modifier seems to trigger an infinite loop of the modifier being called. 

Tried a few other checks, such as doing a check on `_isExpanded`, and none seemed to fix the issue, so reverted back to the "harder" reset. But there could be something I'm missing.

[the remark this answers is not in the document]
didoo: Everything I tried too, was causing an infinite loop, because of the tracked variables. So it's not worth to spend too much time on this, it is an edge case. Thanks for trying anyway :)

didoo (approved):
Great PR! 👏👏👏

PR #2865, opened 2025-05-09
`AdvancedTable` - add `@maxHeight` argument and fix the container styles

shleewhite:
### :pushpin: Summary

If merged, this PR would: 
* add a `maxHeight` argument. If it is set, `hasStickyHeader` is automatically set to true.
* add a check that if `hasStickyHeader` is true, the consumer has to set  `@maxHeight`
* update the showcase

**[Showcase preview](https://hds-showcase-git-hds-4751-advanced-table-height-hashicorp.vercel.app/components/advanced-table)**

### :link: External links

Jira ticket: [HDS-4751](https://hashicorp.atlassian.net/browse/HDS-4751)

***

### 👀 Component checklist

[HDS-4751]: https://hashicorp.atlassian.net/browse/HDS-4751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/rotten-geese-swim.md
+++ b/.changeset/rotten-geese-swim.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`AdvancedTable` - Added `@maxHeight` argument which automatically adds a sticky header when set and sets the max height of the AdvancedTable. Also updated the container styles to constrain the Advanced Table width to the parent's width.
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -132,6 +132,7 @@ export interface HdsAdvancedTableSignature {
     hasStickyHeader?: boolean;
     hasStickyFirstColumn?: boolean;
     childrenKey?: string;
+    maxHeight?: string;
   };
   Blocks: {
     body?: [
@@ -311,6 +312,19 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     return density;
   }
 
+  get hasStickyHeader(): boolean {
+    if (this.args.maxHeight && this.args.hasStickyHeader !== false) {
+      return true;
+    } else if (this.args.hasStickyHeader && !this.args.maxHeight) {
+      assert(
+        'Must set @maxHeight to use @hasStickyHeader.',
+        false
+      );
+    }
+
+    return false
+  }
+
   get valign(): HdsAdvancedTableVerticalAlignment {
     const { valign = DEFAULT_VALIGN } = this.args;
 
@@ -371,7 +385,7 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
   get theadClassNames(): string {
     const classes = ['hds-advanced-table__thead'];
 
-    if (this.args.hasStickyHeader) {
+    if (this.hasStickyHeader) {
       classes.push('hds-advanced-table__thead--sticky');
     }
 
@@ -411,12 +425,12 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
 
       // sticky header
       if (element.scrollTop > 0) {
-        if (this.args.hasStickyHeader) {
+        if (this.hasStickyHeader) {
           this.isStickyHeaderPinned = true;
         }
         this.showScrollIndicatorTop = true;
       } else {
-        if (this.args.hasStickyHeader) {
+        if (this.hasStickyHeader) {
           this.isStickyHeaderPinned = false;
         }
         this.showScrollIndicatorTop = false;
@@ -440,7 +454,7 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
       this.scrollIndicatorDimensions = getScrollIndicatorDimensions(
         element,
         this._theadElement,
-        hasStickyHeader,
+        this.hasStickyHeader,
         hasStickyFirstColumn
       );
 
@@ -453,7 +467,6 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     };
 
     const {
-      hasStickyHeader = false,
       hasStickyFirstColumn = false,
       isSelectable = false,
     } = this.args;
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -316,13 +316,10 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     if (this.args.maxHeight && this.args.hasStickyHeader !== false) {
       return true;
     } else if (this.args.hasStickyHeader && !this.args.maxHeight) {
-      assert(
-        'Must set @maxHeight to use @hasStickyHeader.',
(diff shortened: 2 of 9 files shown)


.changeset/rotten-geese-swim.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`AdvancedTable` - Added `@maxHeight` argument which automatically adds a sticky header when set and sets the max height of the AdvancedTable. Also updated the container styles to constrain the Advanced Table width to the parent's width.
```
KristinLBradley, line 5: I would put the added features on separate lines as done here: https://helios.hashicorp.design/components/code-editor?tab=version%20history

Added `@maxHeight` argument which automatically adds a sticky header when set and sets the max height of the `AdvancedTable`.

Updated the container styles to constrain the `AdvancedTable` width to the parent's width.

KristinLBradley (dismissed):
Looks good! I just added a suggestion for the changelog format

.changeset/rotten-geese-swim.md
shleewhite: I think were not really supposed to do the multiple lines anymore.. the automation doesn't know to look for line breaks

website/app/styles/pages/components/advanced-table.scss
didoo, line 1: Dylan suggested in a different PR to have the documentation updates go in a separate PR, so they're not published until the code changes are actually released, which makes sense, so I am going to do the same :)

showcase/app/templates/components/advanced-table.hbs
```diff
@@ -174,7 +174,7 @@
         (hash key="email" label="Email")
         (hash key="role" label="Role")
       }}
-      @hasStickyHeader={{true}}
+      @maxHeight="400px"
```
didoo, line 177: [suggestion] here we're testing the condition where only the `@maxHeight` is declared, which implicitly sets the `@hasStickyHeader` to `true`; maybe we should add also a use case where  `@maxHeight` is declared, but `@hasStickyHeader` is explicitly set to `false`?

showcase/app/templates/components/advanced-table.hbs
```diff
@@ -250,7 +250,7 @@
 
   <Shw::Text::H2>Sticky header and sticky column</Shw::Text::H2>
 
-  <div class="shw-component-advanced-table-fixed-dimensions-wrapper">
+  <div class="shw-component-advanced-table-fixed-width-wrapper">
```
didoo, line 253: [issue] I think the new class name has not been updated in Sass

didoo: [todo] also check the `shw-component-advanced-table-fixed-height-wrapper` class which is used in the template but there's no counterpart in Sass (which means you can remove the `<div class="shw-component-advanced-table-fixed-height-wrapper">` wrapper.

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -311,6 +312,16 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     return density;
   }
 
+  get hasStickyHeader(): boolean {
```
didoo, line 315: [question] this logic allows the condition where the table has a max height but the header is not sticky; is it intentional (I imagine so, given the integration tests below)? and is it accessible (I also imagine so, but I wanted to double check)

showcase/app/styles/mock-components/app.scss
```diff
@@ -17,10 +17,3 @@
   gap: 12px;
   margin-bottom: 12px;
 }
-
-.mock-app-main-generic-advanced-table-wrapper {
```
didoo, line 21: [praise] nice :)

showcase/app/styles/showcase-pages/advanced-table.scss
```diff
@@ -22,17 +22,13 @@ body.components-advanced-table {
...
       align-items: center;
-  
+
       .hds-icon {
         flex: none;
         margin-right: 5px;
       }
     }
 
-  .shw-component-advanced-table-fixed-height-wrapper {
```
didoo, line 32: [todo] see comment above, about wrapping `<div>` that can be removed

website/docs/components/table/advanced-table/partials/code/component-api.md
```diff
@@ -92,8 +92,11 @@ The Advanced Table component itself is where most of the options will be applied
     Determines if even-numbered rows will have a different background color from odd-numbered rows.<br><br>
     **Important**: Advanced Table does **not** support setting `isStriped` to true when there are nested rows.
   </C.Property>
+  <C.Property @name="maxHeight" @type="string">
+    Sets the maximum height of the Advanced Table. If the `@maxHeight` is set, there will automatically be a sticky header. To turn off the sticky header and still have a max height, set `@hasStickyHeader` to false.
+  </C.Property>
   <C.Property @name="hasStickyHeader" @type="boolean" @default="false">
-    Determines if the Advanced Table has a sticky header.
+    Determines if the Advanced Table has a sticky header. If set to `true`, must be used with the `@maxHeight` argument.
```
didoo, line 99: [consideration] I would remove `@default="false"` from the property (because it may be misleading, and technically is not true, depending on another argument) and I would try to describe the two conditions in the description of the property.

didoo (commented):
Left a few comments, but nothing major. I am pre-approving (even if I know it will be reset once you push new commits)

showcase/app/templates/components/advanced-table.hbs
shleewhite: True! I have a test for this but not a showcase example. Will update.

website/app/styles/pages/components/advanced-table.scss
shleewhite: For sure! Here is the new PR: https://github.com/hashicorp/design-system/pull/2872

website/docs/components/table/advanced-table/partials/code/component-api.md
shleewhite: I fixed this in the docs PR: https://github.com/hashicorp/design-system/pull/2872

packages/components/src/components/hds/advanced-table/index.ts
shleewhite: Yup! We just wanted to give people an escape hatch to not have the sticky header + set a max height.

showcase/app/styles/showcase-pages/advanced-table.scss
shleewhite: Removed!

showcase/app/templates/components/advanced-table.hbs
shleewhite: I ended up deleting the `shw-component-advanced-table-fixed-dimensions-wrapper` class and using the `shw-component-advanced-table-fixed-width-wrapper` class instead in combination with the `maxHeight` + `hasStickyHeader={{false}}` a few times in the showcase. It should be good now.

didoo (approved):
👍

PR #2867, opened 2025-05-10
fix(components): `LinkToExternal` retake

aklkv:
### :pushpin: Summary

Retake on https://github.com/hashicorp/design-system/pull/2838 with dynamic imports approach which was suggested in https://github.com/embroider-build/embroider/issues/2471#issuecomment-2863435183 which seems to be most resilient solution

One thing we loose with this approach is build time code elimination. In our case is not a big deal as we still throw an error to users if they use `@isRouteExternal` without `ember-engines` installed

This seems to work now in:

* cloud-ui - https://github.com/hashicorp/cloud-ui/pull/12217
* atlas - https://github.com/hashicorp/atlas/pull/23282

All CI errors I see are unrelated this change

[suggestion] @hashicorp/hds-engineering I keep seeing people confusing `@isRouteExternal` and `@isHrefExternal` where `@href` passed in, I think we can add assertion if `@isRouteExternal` require `@route` etc.

[side-note]: I had to battle with runloop a little to make Percy happy

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist

### DIFF
--- a/.changeset/tall-rings-act.md
+++ b/.changeset/tall-rings-act.md
@@ -1,0 +1,5 @@
+---
+'@hashicorp/design-system-components': patch
+---
+
+Introduce the `hds-resolve-link-to-component` utility to correctly resolve the LinkTo component when `@isRouteExternal` is set on `HdsBreadcrumbItem` or `HdsInteractive`. Consumers are now required to install `ember-engines` when `@isRouteExternal` is `true`.
--- a/packages/components/src/components/hds/breadcrumb/item.hbs
+++ b/packages/components/src/components/hds/breadcrumb/item.hbs
@@ -13,7 +13,7 @@
       <span class="hds-breadcrumb__text">{{@text}}</span>
     </div>
   {{else}}
-    {{~#if this.linkToComponent~}}
+    {{#if this.linkToComponent}}
       <this.linkToComponent
         class="hds-breadcrumb__link"
         @current-when={{@current-when}}
@@ -29,6 +29,6 @@
         {{/if}}
         <span class="hds-breadcrumb__text">{{@text}}</span>
       </this.linkToComponent>
-    {{~/if~}}
+    {{/if}}
   {{/if}}
 </li>
--- a/packages/components/src/components/hds/breadcrumb/item.hbs
+++ b/packages/components/src/components/hds/breadcrumb/item.hbs
@@ -13,38 +13,20 @@
       <span class="hds-breadcrumb__text">{{@text}}</span>
     </div>
   {{else}}
-    {{#if @isRouteExternal}}
-      <LinkToExternal
-        class="hds-breadcrumb__link"
-        @current-when={{@current-when}}
-        @models={{hds-link-to-models @model @models}}
-        @query={{hds-link-to-query @query}}
-        @replace={{@replace}}
-        @route={{@route}}
-      >
-        {{#if @icon}}
-          <div class="hds-breadcrumb__icon">
-            <Hds::Icon @name={{@icon}} @size="16" @stretched={{true}} />
-          </div>
-        {{/if}}
-        <span class="hds-breadcrumb__text">{{@text}}</span>
-      </LinkToExternal>
-    {{else}}
-      <LinkTo
-        class="hds-breadcrumb__link"
-        @current-when={{@current-when}}
-        @models={{hds-link-to-models @model @models}}
-        @query={{hds-link-to-query @query}}
-        @replace={{@replace}}
-        @route={{@route}}
-      >
-        {{#if @icon}}
-          <div class="hds-breadcrumb__icon">
-            <Hds::Icon @name={{@icon}} @size="16" @stretched={{true}} />
-          </div>
-        {{/if}}
-        <span class="hds-breadcrumb__text">{{@text}}</span>
-      </LinkTo>
-    {{/if}}
+    <this.linkToComponent
+      class="hds-breadcrumb__link"
+      @current-when={{@current-when}}
+      @models={{hds-link-to-models @model @models}}
+      @query={{hds-link-to-query @query}}
+      @replace={{@replace}}
+      @route={{@route}}
+    >
+      {{#if @icon}}
+        <div class="hds-breadcrumb__icon">
+          <Hds::Icon @name={{@icon}} @size="16" @stretched={{true}} />
+        </div>
+      {{/if}}
+      <span class="hds-breadcrumb__text">{{@text}}</span>
+    </this.linkToComponent>
   {{/if}}
 </li>
--- a/packages/components/src/components/hds/breadcrumb/item.hbs
+++ b/packages/components/src/components/hds/breadcrumb/item.hbs
@@ -13,20 +13,22 @@
       <span class="hds-breadcrumb__text">{{@text}}</span>
     </div>
   {{else}}
-    <this.linkToComponent
-      class="hds-breadcrumb__link"
-      @current-when={{@current-when}}
-      @models={{hds-link-to-models @model @models}}
-      @query={{hds-link-to-query @query}}
-      @replace={{@replace}}
-      @route={{@route}}
-    >
-      {{#if @icon}}
-        <div class="hds-breadcrumb__icon">
-          <Hds::Icon @name={{@icon}} @size="16" @stretched={{true}} />
-        </div>
-      {{/if}}
-      <span class="hds-breadcrumb__text">{{@text}}</span>
-    </this.linkToComponent>
+    {{~#if this.linkToComponent~}}
+      <this.linkToComponent
+        class="hds-breadcrumb__link"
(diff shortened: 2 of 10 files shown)


showcase/app/controllers/components.js
```diff
@@ -5,7 +5,7 @@
 
 import Controller from '@ember/controller';
 import { inject as service } from '@ember/service';
-import { scheduleOnce } from '@ember/runloop';
+import { later } from '@ember/runloop';
```
didoo, line 8: [question] why this change fixed the issue?

context: `scheduleOnce` is used in a lot of places in the showcase, for the replacement of mock data attributes to emulate states; why is not failing in other cases, and why it was not failing before? what is the change in your code that caused the failure of percy tests, the dynamic import? in which case, could this cause some side effect on the consumers that we are not thinking of?

didoo (commented):
added a couple of comments

showcase/app/controllers/components.js
aklkv: well there is a multiple issues at play here:

* you are using controller which is already a bad idea for render based listeners and DOM manipulation as it's a singleton
* we a lying with `eslint-disable-next-line ember/no-runloop` if you look up what the reasoning for this rule is https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-runloop.md it would be obvious that is not aware of rendering async events like having ability to rerun when `tracked` property changes in our case we are swapping out actual component all together
* related to above since you are using `scheduleOnce` it only run once on the first paint but since we use `tracked` property to swap out component it doesn't rerun again

as for consumers with all my testing I have not seen anything failing in linked PRs

packages/components/src/components/hds/breadcrumb/item.hbs
```diff
@@ -13,7 +13,7 @@
       <span class="hds-breadcrumb__text">{{@text}}</span>
     </div>
   {{else}}
-    {{~#if this.linkToComponent~}}
+    {{#if this.linkToComponent}}
```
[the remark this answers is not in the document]
didoo: Ahhh! 👍

showcase/app/controllers/components.js
aklkv: I was able to side step percy issue by initializing to an actual value instead of `null` which I think stops unwanted rerenders

aklkv:
@didoo yes I think it's is ready for final review. as for failing tests they are not related to my changes, as you can see here https://github.com/hashicorp/cloud-ui/pull/12240 this is purely bumping to lates hds and has identical failed test. I am not sure if 4.19 was tested before the release

.changeset/tall-rings-act.md
```diff
@@ -0,0 +1,5 @@
+---
+'@hashicorp/design-system-components': patch
+---
+
+Introduce the `hds-resolve-link-to-component` utility to correctly resolve the LinkTo component when `@isRouteExternal` is set on `HdsBreadcrumbItem` or `HdsInteractive`. Consumers are now required to install `ember-engines` when `@isRouteExternal` is `true`.
```
didoo, line 5: for @alex-ju: does this works with the automated script for release notes generation?

packages/components/unpublished-development-types/global.d.ts
```diff
@@ -25,7 +24,6 @@ declare module '@glint/environment-ember-loose/registry' {
       RenderModifiersRegistry,
       EmbroiderUtilRegistry /*, other addon registries */ {
     // local entries
-    LinkToExternal: typeof LinkTo;
```
didoo, line 28: [praise] Great that you remembered to remove it :)

packages/components/src/components/hds/breadcrumb/item.ts
```diff
@@ -27,6 +33,20 @@ export interface HdsBreadcrumbItemSignature {
 }
 
 export default class HdsBreadcrumbItem extends Component<HdsBreadcrumbItemSignature> {
+  @tracked linkToExternal: LinkTo | null = null;
+
+  constructor(owner: Owner, args: HdsBreadcrumbItemSignature['Args']) {
+    super(owner, args);
+
+    void this.setupComponent();
```
didoo, line 41: [question] just a thought, but if we move the check on `this.args.isRouteExternal` here, some computational cycles can be saved? (_I think_) 

(even if we lose the abstraction in the check that now is shared in the helper)

didoo (dismissed):
Neat PR 🙌. Left a comment, but not a blocker

packages/components/src/components/hds/breadcrumb/item.ts
aklkv: I was contemplating the same thing, and since this would be applicable to very small amunt of cases we could really save on some rerenders, I will modify this 👍🏻

packages/components/src/components/hds/breadcrumb/item.ts
```diff
@@ -27,6 +33,23 @@ export interface HdsBreadcrumbItemSignature {
 }
 
 export default class HdsBreadcrumbItem extends Component<HdsBreadcrumbItemSignature> {
+  @tracked linkToExternal: LinkTo | null = null;
+
+  constructor(owner: Owner, args: HdsBreadcrumbItemSignature['Args']) {
+    super(owner, args);
+
+    // if want to make sure that we do not try to resolve the component if it's not needed
+    if (args.isRouteExternal) {
```
aklkv, line 42: @didoo I add this safeguard to we do not run it if do not need to however as you can see I did not modify unility function as I still want it to be safe contained, even though we might delete it in the future all together once ember-engines aren't a thing

packages/components/src/components/hds/breadcrumb/item.ts
```diff
@@ -27,6 +33,23 @@ export interface HdsBreadcrumbItemSignature {
 }
 
 export default class HdsBreadcrumbItem extends Component<HdsBreadcrumbItemSignature> {
+  @tracked linkToExternal: LinkTo | null = null;
+
+  constructor(owner: Owner, args: HdsBreadcrumbItemSignature['Args']) {
+    super(owner, args);
+
+    // if want to make sure that we do not try to resolve the component if it's not needed
```
didoo, line 41:
```suggestion
    // want to make sure that we do not try to resolve the component if it's not needed
```

packages/components/src/components/hds/interactive/index.ts
```diff
@@ -27,6 +33,22 @@ export interface HdsInteractiveSignature {
 }
 
 export default class HdsInteractive extends Component<HdsInteractiveSignature> {
+  @tracked linkToExternal: LinkTo | null = null;
+
+  constructor(owner: Owner, args: HdsInteractiveSignature['Args']) {
+    super(owner, args);
+
+    // if want to make sure that we do not try to resolve the component if it's not needed
```
didoo, line 41:
```suggestion
    // want to make sure that we do not try to resolve the component if it's not needed
```

packages/components/src/components/hds/breadcrumb/item.ts
didoo: > I did not modify unility function as I still want it to be safe contained,
makes sense

didoo (approved):

packages/components/src/components/hds/interactive/index.ts
```diff
@@ -27,6 +33,22 @@ export interface HdsInteractiveSignature {
...
 
 export default class HdsInteractive extends Component<HdsInteractiveSignature> {
+  @tracked linkToExternal: LinkTo | null = null;
+
+  constructor(owner: Owner, args: HdsInteractiveSignature['Args']) {
+    super(owner, args);
+
+    // we want to avoid resolving the component if it's not needed
+    if (args.isRouteExternal) {
+      void this.setupComponent();
```
didoo, line 43:
```suggestion
      void this.resolveLinkToExternal();
```

packages/components/src/components/hds/interactive/index.ts
```diff
@@ -27,6 +33,22 @@ export interface HdsInteractiveSignature {
...
+    }
+  }
+
+  async setupComponent() {
```
didoo, line 47:
```suggestion
  async resolveLinkToExternal() {
```

didoo (dismissed):
[non-blocker] Nitpicking, I know, but I think we should rename a couple of things, so they're semantically more meaningful/correct:

`setupComponent()` → `resolveLinkToExternal()`
`hdsResolveLinkToComponent`  → `hdsResolveLinkToExternal`

That said, not a blocker, but if you rename them I'm here to approve straightaway, don't worry

didoo (dismissed):
🙌

didoo (approved):

didoo:
@aklkv the fix has been released in [version 4.20](https://helios.hashicorp.design/whats-new/release-notes#4200)

thanks for the contribution 🙏

PR #2869, opened 2025-05-12
`AppSideNav`: fix scrolling issue and being able to focus hidden links issue

shleewhite:
### :pushpin: Summary

If merged, this PR would do some cleanup of the AppSideNav component
* only lock scrolling if the user is on mobile (when on desktop, scrolling shouldnt be prevented because the user can still interact with the rest of the page)
* reset ability to scroll when close the AppSideNav with the escape key
* prevent the user from tabbing to the body of the AppSideNav when it is minimized

There are also updates to tests to check for the correct behavior  (+ 1 bonus fix for a Modal test):
* add test to check the user cannot tab to the content when the AppSideNav is minimized
* add checks to existing tests that `overflow: hidden` is set appropriately
* add test that when a user resizes from mobile -> desktop, ability to scroll is returned
* re-enable test for Modal because I figured out how to target the `body` element in tests

**🧪 Tested in Atlas:**
* [Slack thread with context](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1745609577404549)
* [deployment](https://atlas-inspirational-prairie-334208.prod.previews.hashicorp.services/app/hashicorp_v2/projects/prj-7tgHyPpDmBebEkYN0)

### :link: External links

Jira ticket: [HDS-4827](https://hashicorp.atlassian.net/browse/HDS-4827)
Jira ticket: [HDS-4858](https://hashicorp.atlassian.net/browse/HDS-4858)

***

### 👀 Component checklist

[HDS-4827]: https://hashicorp.atlassian.net/browse/HDS-4827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[HDS-4858]: https://hashicorp.atlassian.net/browse/HDS-4858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/honest-experts-battle.md
+++ b/.changeset/honest-experts-battle.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-`AppSideNav` - fixed bug where scrolling was blocked when the `AppSideNav` was expanded on desktop views. Also fixed bug where able to focus links that are visually hidden.
+`AppSideNav` - Fixed bug where scrolling was blocked when the `AppSideNav` was expanded on desktop views. Also fixed bug which allowed user to focus links that were visually hidden.
--- a/.changeset/honest-experts-battle.md
+++ b/.changeset/honest-experts-battle.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-`AppSideNav` - fixed bug where scrolling was blocked when the `AppSideNav` was expanded on desktop views.
+`AppSideNav` - fixed bug where scrolling was blocked when the `AppSideNav` was expanded on desktop views. Also fixed bug where able to focus links that are visually hidden.
--- a/.changeset/honest-experts-battle.md
+++ b/.changeset/honest-experts-battle.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`AppSideNav` - fixed bug where scrolling was blocked when the `AppSideNav` was expanded on desktop views.
--- a/packages/components/src/components/hds/app-side-nav/index.hbs
+++ b/packages/components/src/components/hds/app-side-nav/index.hbs
@@ -27,7 +27,7 @@
       />
     {{/if}}
 
-    <div class="hds-app-side-nav__wrapper-body hds-app-side-nav--hide-when-minimized">
+    <div class="hds-app-side-nav__wrapper-body" {{did-insert this.didInsertNavWrapperBody}}>
       {{~yield~}}
     </div>
   </div>
--- a/packages/components/src/components/hds/app-side-nav/index.hbs
+++ b/packages/components/src/components/hds/app-side-nav/index.hbs
@@ -10,7 +10,7 @@
   {{on "transitionend" (fn this.setTransition "end")}}
   {{! @glint-expect-error - https://github.com/josemarluedke/ember-focus-trap/issues/86 }}
   {{focus-trap isActive=this.shouldTrapFocus}}
-  {{did-insert this.didInsert}}
+  {{this._setUpBodyElement}}
 >
   <h2 class="sr-only" id="hds-app-side-nav-header">Application local navigation</h2>
 
@@ -27,7 +27,7 @@
       />
     {{/if}}
 
-    <div class="hds-app-side-nav__wrapper-body" {{did-insert this.didInsertNavWrapperBody}}>
+    <div class="hds-app-side-nav__wrapper-body" {{this._setUpNavWrapperBody}}>
       {{~yield~}}
     </div>
   </div>
--- a/packages/components/src/components/hds/app-side-nav/index.hbs
+++ b/packages/components/src/components/hds/app-side-nav/index.hbs
@@ -27,7 +27,7 @@
       />
     {{/if}}
 
-    <div class="hds-app-side-nav__wrapper-body hds-app-side-nav-hide-when-minimized">
+    <div class="hds-app-side-nav__wrapper-body hds-app-side-nav--hide-when-minimized">
       {{~yield~}}
     </div>
   </div>
--- a/packages/components/src/components/hds/app-side-nav/index.hbs
+++ b/packages/components/src/components/hds/app-side-nav/index.hbs
@@ -27,7 +27,7 @@
       />
     {{/if}}
 
-    <div class="hds-app-side-nav__wrapper-body">
+    <div class="hds-app-side-nav__wrapper-body hds-app-side-nav-hide-when-minimized">
       {{~yield~}}
     </div>
   </div>
--- a/packages/components/src/components/hds/app-side-nav/index.ts
+++ b/packages/components/src/components/hds/app-side-nav/index.ts
@@ -188,7 +188,7 @@ export default class HdsAppSideNav extends Component<HdsAppSideNavSignature> {
   @action
   didInsert(element: HTMLElement): void {
     this._containersToHide = element.querySelectorAll(
-      '.hds-app-side-nav-hide-when-minimized'
+      '.hds-app-side-nav--hide-when-minimized'
     );
     this._body = document.body;
     // Store the initial `overflow` value of `<body>` so we can reset to it
--- a/packages/components/src/components/hds/app-side-nav/index.ts
+++ b/packages/components/src/components/hds/app-side-nav/index.ts
@@ -175,10 +175,12 @@ export default class HdsAppSideNav extends Component<HdsAppSideNavSignature> {
       onToggleMinimizedStatus(this._isMinimized);
     }
 
-    if (this._isMinimized) {
-      this.unlockBodyScroll();
-    } else {
-      this.lockBodyScroll();
+    if (!this._isDesktop) {
+      if (this._isMinimized) {
+        this.unlockBodyScroll();
+      } else {
+        this.lockBodyScroll();
+      }
     }
   }
 
@@ -214,6 +216,11 @@ export default class HdsAppSideNav extends Component<HdsAppSideNavSignature> {
     this._isMinimized = !this._isDesktop;
 
     this.synchronizeInert();
+
+    if (this._isDesktop) {
+      // make sure scrolling is enabled if the user resizes the window from mobile to desktop
+      this.unlockBodyScroll();
+    }
 
     const { onDesktopViewportChange } = this.args;
 
--- a/packages/components/src/components/hds/app-side-nav/index.ts
+++ b/packages/components/src/components/hds/app-side-nav/index.ts
@@ -34,7 +34,7 @@ export default class HdsAppSideNav extends Component<HdsAppSideNavSignature> {
   private _body!: HTMLElement;
   private _bodyInitialOverflowValue = '';
   private _desktopMQ: MediaQueryList;
-  private _containersToHide!: NodeListOf<Element>;
+  private _navWrapperBody!: HTMLElement;
 
   // we use the `lg` breakpoint for `desktop` viewports, but consumers can override its value
   private _desktopMQVal = this.args.breakpoint ?? hdsBreakpoints['lg'].px;
@@ -123,13 +123,11 @@ export default class HdsAppSideNav extends Component<HdsAppSideNavSignature> {
   }
 
   synchronizeInert(): void {
-    this._containersToHide?.forEach((element): void => {
-      if (this._isMinimized) {
-        element.setAttribute('inert', '');
-      } else {
-        element.removeAttribute('inert');
-      }
-    });
+    if (this._isMinimized) {
+      this._navWrapperBody?.setAttribute('inert', '');
+    } else {
+      this._navWrapperBody?.removeAttribute('inert');
+    }
   }
 
   lockBodyScroll(): void {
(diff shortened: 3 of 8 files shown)


packages/components/src/components/hds/app-side-nav/index.hbs
```diff
@@ -27,7 +27,7 @@
       />
     {{/if}}
 
-    <div class="hds-app-side-nav__wrapper-body">
+    <div class="hds-app-side-nav__wrapper-body hds-app-side-nav-hide-when-minimized">
```
shleewhite, line 30: Note: I add to add this class because this is what the component is checking to decide what to add `inert` to. Previously, this was only getting added to the target for the portal ([link](https://github.com/hashicorp/design-system/blob/589e32fadd810ea7dabacb31aef99dc42d76ebd5/packages/components/src/components/hds/app-side-nav/portal/target.hbs#L7)).

Now it gets added when the consumer isn't using Portals too.

.changeset/honest-experts-battle.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`AppSideNav` - fixed bug where scrolling was blocked when the `AppSideNav` was expanded on desktop views. Also fixed bug where able to focus links that are visually hidden.
```
KristinLBradley, line 5:
```suggestion
`AppSideNav` - Fixed bug where scrolling was blocked when the `AppSideNav` was expanded on desktop views. Also fixed bug which allowed user to focus links that were visually hidden.
```

KristinLBradley: Some of the wording above seemed a little awkward so I edited to try to clarify. (Feel free to tweak further.)

packages/components/src/components/hds/app-side-nav/index.hbs
KristinLBradley: This seems a little confusing since the 2nd class is always present. Could you just use the existing `hds-app-side-nav__wrapper-body` class name instead?

KristinLBradley: (Also BEM format would be "hds-app-side-nav--hide-when-minimized")

.changeset/honest-experts-battle.md
shleewhite: This is definitely better!

packages/components/src/components/hds/app-side-nav/index.hbs
shleewhite: Yeah.. I think this comes from SideNav. I can update it to be BEM format.

When the AppSideNav is inserted, it looks for elements with this class name [to decide where to put inert](https://github.com/hashicorp/design-system/blob/130b66858cdf06df6cf7c707e4ac3890dfe7ad99/packages/components/src/components/hds/app-side-nav/index.ts#L184).

I didn't want to change how it works for Portals, which is why I kept using the existing class.

website/docs/components/app-side-nav/partials/code/how-to-use.md
```diff
@@ -206,15 +206,13 @@ The App Side Nav component **automatically**:
...
+One possible way to do it is to use the **`hds-app-side-nav--hide-when-minimized` class**. This is a special class that can be applied to a DOM element so that it **automatically** fades in/out when the App Side Nav changes its “minimization” state.
 
 More specifically:
 
 - `minimized → maximized` transition: the content appears with a fade-in effect, when the width animation is already completed (the width is maximized)
 - `maximized → minimized` transition: the content disappears at once with no transition, before the width animation starts
 
-Another option is to use the **`isMinimized` parameter**, which is useful in those cases where the content is so custom/specialized that it can’t just be faded in/out but needs to have a different kind of transition (eg. remain visible but change layout or respond to the width of the container). This value is passed down by the `<:header/body/footer>` named blocks as parameters, and can be used to build custom logic on the consumers’ side.
```
shleewhite, line 216: The component doesn't return named blocks or the `isMinimized` state to the consumer, so this is not true.

packages/components/src/components/hds/app-side-nav/index.hbs
KristinLBradley: It still seems a bit confusing to add an extra class that's always there instead of just using the existing class. Would be curious to know what other engineers think though. @hashicorp/hds-engineering 

Maybe you could at least add a comment to explain why the extra class is there?

website/docs/components/app-side-nav/partials/code/how-to-use.md
```diff
@@ -206,15 +206,13 @@ The App Side Nav component **automatically**:
 
 Any other content in the App Side Nav needs to be **explicitly handled** by the consumers (in this way they have full control of the content they add, and they can customize the transition as they want/need).
 
-One possible way to do it is to use the **`hds-app-side-nav-hide-when-minimized` class**. This is a special class that can be applied to a DOM element so that it **automatically** fades in/out when the App Side Nav changes its “minimization” state.
+One possible way to do it is to use the **`hds-app-side-nav--hide-when-minimized` class**. This is a special class that can be applied to a DOM element so that it **automatically** fades in/out when the App Side Nav changes its “minimization” state.
```
KristinLBradley, line 209: But this class is always present now, isn't it?

KristinLBradley (commented):
Looking good overall but I had a few questions still (see comments)

packages/components/src/components/hds/app-side-nav/index.hbs
didoo: I'll have a look tomorrow

packages/components/src/components/hds/app-side-nav/index.hbs
```diff
@@ -27,7 +27,7 @@
       />
     {{/if}}
 
-    <div class="hds-app-side-nav__wrapper-body">
+    <div class="hds-app-side-nav__wrapper-body hds-app-side-nav--hide-when-minimized">
```
didoo, line 30: [thought] in the old `SideNav` we exposed three named blocks (`header/body/footer`) in which consumers were able to add extra content, without being hidden when minimized (unless they explicitly set this class on the element); I need a better look to familiarize with the API of the new `AppSideNav`, so bear with me while I try to understand how this updated logic works

website/docs/components/app-side-nav/partials/code/how-to-use.md
shleewhite: Yes... Ok I have looked into it more and I believe we can safely simplify this logic. I will push up a refactor shortly.

packages/components/src/components/hds/app-side-nav/index.hbs
```diff
@@ -27,7 +27,7 @@
       />
     {{/if}}
 
-    <div class="hds-app-side-nav__wrapper-body">
+    <div class="hds-app-side-nav__wrapper-body" {{did-insert this.didInsertNavWrapperBody}}>
```
didoo, line 30: [suggestion] my understanding is that we're moving away from `ember-render-modifiers` in favour of custom modifiers (see Dylan's PR https://github.com/hashicorp/design-system/pull/2856/files for example); maybe something we can do here as well?

didoo: or is it because you didn't want to have to do also the migration of the `did-insert` applied to the root element of this component? in which case makes sense, and I have no objection to keep it "old school" for now :)

packages/components/src/components/hds/app-side-nav/index.ts
```diff
@@ -213,6 +216,11 @@ export default class HdsAppSideNav extends Component<HdsAppSideNavSignature> {
 
     this.synchronizeInert();
 
+    if (this._isDesktop) {
+      // make sure scrolling is enabled if the user resizes the window from mobile to desktop
```
didoo, line 222: [praise] 👏 👏 👏

packages/components/src/components/hds/app-side-nav/index.ts
```diff
@@ -123,13 +123,11 @@ export default class HdsAppSideNav extends Component<HdsAppSideNavSignature> {
...
 
   synchronizeInert(): void {
-    this._containersToHide?.forEach((element): void => {
-      if (this._isMinimized) {
-        element.setAttribute('inert', '');
-      } else {
-        element.removeAttribute('inert');
-      }
-    });
+    if (this._isMinimized) {
```
didoo, line 126: [question/suggestion?] MDN [says](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert) that "The inert [...] attribute is a Boolean attribute". Does it mean this could be converted to `this._navWrapperBody?.setAttribute('inert', this._isMinimized);`?  (in case, test that it works, I didn't)

showcase/tests/integration/components/hds/modal/index-test.js
```diff
@@ -249,7 +248,7 @@ module('Integration | Component | hds/modal/index', function (hooks) {
     await click('#test-interactive');
     assert.true(this.showModal);
     await click('button.hds-modal__dismiss');
-    assert.dom('body', 'body').isFocused();
+    assert.dom('body', document).isFocused();
```
didoo, line 251: 😮

showcase/tests/integration/components/hds/app-side-nav/index-test.js
```diff
@@ -254,29 +257,81 @@ module('Integration | Component | hds/app-side-nav/index', function (hooks) {
...
+    );
+
+    assert.dom('body', document).doesNotHaveStyle('overflow');
+  });
+
+  test('when collapsed, the content in the AppSideNav is not focusable', async function (assert) {
+    await render(hbs`<Hds::AppSideNav
+    id='test-app-side-nav'
+  @isCollapsible={{true}}
+  @onDesktopViewportChange={{this.onDesktopViewportChange}}
```
didoo, line 324: [question] could it be that `this.onDesktopViewportChange` is `undefined` for this test, and you have to add its definition like in the other tests above?

showcase/tests/integration/components/hds/app-side-nav/index-test.js
```diff
@@ -254,29 +257,81 @@ module('Integration | Component | hds/app-side-nav/index', function (hooks) {
...
+>
+  <span id='test-app-side-nav-body' />
```
didoo, line 326: [question] what is this `span#test-app-side-nav-body` used for? 🤔

(I see other places where is not used in the tests assertions/declarations; maybe there's a reason for that? otherwise maybe we can remove them?)

showcase/tests/integration/components/hds/app-side-nav/index-test.js
```diff
@@ -254,29 +257,81 @@ module('Integration | Component | hds/app-side-nav/index', function (hooks) {
...
+</Hds::AppSideNav><button id='button-2'>Click</button>`);
```
didoo, line 327:
```suggestion
</Hds::AppSideNav><button id='button-outside'>Click</button>`);
```

showcase/tests/integration/components/hds/app-side-nav/index-test.js
```diff
@@ -254,29 +257,81 @@ module('Integration | Component | hds/app-side-nav/index', function (hooks) {
...
+    assert.deepEqual(
+      calls[2],
+      [true],
+      'resizing to desktop triggers a true event'
+    );
+
+    assert.dom('body', document).doesNotHaveStyle('overflow');
+  });
+
+  test('when collapsed, the content in the AppSideNav is not focusable', async function (assert) {
```
didoo, line 320: [question] I don't understand this test: how do you know if the content in the AppSideNav is focusable or not, if there's nothing focusable (like a button) inside it? 🤔

didoo (commented):
@shleewhite left a bunch of comments/questions, but overall the PR is so much better and clean with your latest changes! 🙌

showcase/tests/integration/components/hds/app-side-nav/index-test.js
shleewhite: oh my gosh, that is right. I will update.

showcase/tests/integration/components/hds/app-side-nav/index-test.js
shleewhite: I can remove them... They were initially used in tests along with `<span id='hds-app-side-nav-hide-when-minimized' />` to show that inert was only added to the correct elements.

showcase/tests/integration/components/hds/app-side-nav/index-test.js
shleewhite: Ah, I can remove it. It is unneeded for this test.

packages/components/src/components/hds/app-side-nav/index.ts
shleewhite: Tested in chrome and it treated `inert="false"` as inert still, will leave it as is for now.

packages/components/src/components/hds/app-side-nav/index.hbs
shleewhite: Hahah I can update to use the new modifier pattern! Might as well since I'm touching the file.

packages/components/src/components/hds/app-side-nav/index.ts
didoo: Ohhh, intersting! 👍

didoo (approved):
I haven't tested the changes in a production environment (I imagine you have done) but the code review is OK for me. Well, not just OK: this is a **great** PR. 👏 👏 👏

KristinLBradley (approved):
Nice work!

website/docs/components/app-side-nav/partials/code/how-to-use.md
shleewhite: Just to close out the comment. We ended up refactoring the component and removing this class entirely. The whole content of the AppSideNav body is animated in/out so consumers dont need this escape hatch.

PR #2871, opened 2025-05-14
Website `v4.20.0` Release 

shleewhite:
### :pushpin: Summary

If merged, this PR would add new content for the `v4.20.0` release of Helios.
- CodeBlock height toggle docs


didoo:
Closing this PR because the HashiCorp GitHub org seems to have seems to have org-wide branch protection rules for any `release/* branches` that we can’t disable. So the only workaround is to open a new PR (https://github.com/hashicorp/design-system/pull/2926)

PR #2877, opened 2025-05-14
`Time` - fix type error where `TooltipButton` text could be undefined

shleewhite:
### :pushpin: Summary

If merged, this PR would resolve a type error where the `TooltipButton` text could be undefined. Opted to just return `''` if the date is invalid or if it cannot be converted into a UTC ISO format because other functions like `rangeIsoUtcString` and the `hds-format-time` helper return an empty string if it is invalid.

**Other options considered**
* use the JavaScript date methods as a fallback: not used because the case where this happens most likely is when it is an invalid date, so this would not work
* update the `TooltipButton` to let the `text` argument be optional: not used because this would have much larger consequences and we'd rather ensure tooltips always have text.

### :link: External links

Jira ticket: [HDS-4517](https://hashicorp.atlassian.net/browse/HDS-4517)

***

### 👀 Component checklist


KristinLBradley (approved):
Looks good!

PR #2878, opened 2025-05-14
`ember-intl` integration

zamoore:
### :pushpin: Summary

If merged, this PR enables `ember-intl` for the components package and translates some of the bare strings. It also sets up linting rules preventing the usage of bare strings.

https://hashicorp.atlassian.net/browse/HDS-4856

Jira ticket: [HDS-4856](https://hashicorp.atlassian.net/browse/HDS-4856)

***

### 👀 Component checklist


packages/components/translations/en-us.yaml
```diff
@@ -0,0 +1,36 @@
+hds:
+  components:
```
aklkv, line 2: [idea]: it's common practice to break translation files per component bases, you check `cloud-ui ` for examples. it makes this ever growing files much cleaner

zamoore: I agree with you for most codebases. I've had to deal with some rather large translation files at past jobs. I think in our case, just given how text-lite our components are as a whole, we would mostly end up with single-line files. My intuition is colocating the translations for a project like this might make more sense, but I'm curious to get input from the rest of the team as well.

showcase/app/routes/application.js
```diff
@@ -4,5 +4,12 @@
  */
 
 import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
 
-export default class ApplicationRoute extends Route {}
+export default class ApplicationRoute extends Route {
+  @service intl;
+
+  beforeModel() {
```
zamoore, line 12: This sets up the initial locale that the showcase uses for translation.

packages/components/src/components/hds/advanced-table/th-selectable.hbs
```diff
@@ -21,7 +21,9 @@
       {{on "change" this.onSelectionChange}}
     />
     {{#if this.isSortable}}
-      <span id={{this._labelId}} class="sr-only">selection state</span>
+      <span id={{this._labelId}} class="sr-only">
+        {{t "hds.components.shared.selection_state"}}
```
shleewhite, line 25: Suggestion: this might be a hot take, but I think this string can be removed in AdvancedTable and Table. Its not adding any values.

packages/components/translations/en-us.yaml
```diff
@@ -0,0 +1,48 @@
...
+    shared:
+      loading: "Loading"
+      menu: "Menu"
+      optional: "Optional"
+      required: "Required"
+      selection_state: "Selection state"
+      sort_by: "Sort by"
+      toggle: "Toggle"
+      tooltip_prefix: "More information for"
+    advanced_table:
```
aklkv, line 12: This stood out to me as most project in HC and wider communities use `-` in the key names, as well as you do not need quotes

zamoore: I can switch to kebab-case. YAML typically follows snake-case conventions, so this was mostly habit.

packages/components/src/components/hds/advanced-table/th-selectable.hbs
zamoore: Makes sense to me.

packages/components/translations/en-us.yaml
didoo: Personally, I am for "separation of concerns" :) 

(as in, even if a file has only a single line, it's OK)

packages/components/translations/en-us.yaml
```diff
@@ -0,0 +1,44 @@
+hds:
+  components:
+    shared:
```
didoo, line 3: maybe? I have mixed feelings... 

```suggestion
    common:
```

packages/components/translations/en-us.yaml
```diff
@@ -0,0 +1,44 @@
...
+      loading: Loading
+      menu: Menu
+      optional: Optional
+      required: Required
+      sort-by: Sort by
+      toggle: Toggle
+      tooltip-prefix: More information for
+    advanced-table:
+      th-button-expand:
+        aria-description: Toggle the visibility of the related rows.
```
dchyun, line 13:
```suggestion
        aria-description: Toggle the visibility of the related rows
```
[Nit] Probably don't need punctuation

packages/components/translations/en-us.yaml
```diff
@@ -0,0 +1,44 @@
...
+    app-footer:
+      copyright:
+        copyright-text: "&copy; {year} HashiCorp"
```
dchyun, line 16: [Question] More just for my own curiosity, but what happens if we have a word we don't want translated? Would we just add a disable linting line by the text in the template file? Just thinking our company name is something we would probably want to prevent from being changed.

packages/components/src/ember-template-lint-plugin-local/rules/no-bare-strings-in-attributes.mjs
```diff
@@ -0,0 +1,47 @@
+import { Rule } from 'ember-template-lint';
+
+export default class NoBareStringsInAttributes extends Rule {
```
didoo, line 3: [question] do you know if there is a way to have tests for this, in some way?

showcase/translations/en-us.yaml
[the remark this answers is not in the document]
zamoore: You have to have at least 1 translation in the translation file or there are issues with the build. This is just a jumping off point for when the translation work is done for the showcase app proper.

packages/components/translations/en-us.yaml
KristinLBradley: I haven't worked with yaml files a whole lot so maybe this is a silly question, but could we possibly include quotes around all the strings for consistency?

packages/components/src/ember-template-lint-plugin-local/index.mjs
```diff
@@ -0,0 +1,8 @@
+import NoBareStringsInAttributes from './rules/no-bare-strings-in-attributes.mjs';
```
aklkv, line 1: non of this files should be inside of src folder as they have nothing todo with what we ship to consumers, ideally we would create new package that has shared configs but at the minimum we should move this out of src folder

packages/components/translations/en-us.yaml
aklkv: or remove them everywhere

showcase/app/routes/application.js
```diff
@@ -4,5 +4,12 @@
  */
 
 import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
```
aklkv, line 7: since this is a net new line of code it worth writing it in a new way:

```
import { service } from '@ember/service';
```

showcase/config/ember-intl.js
```diff
@@ -0,0 +1,93 @@
+module.exports = function (/* environment */) {
```
alex-ju, line 1: Placing this configuration in the showcase app means we expect all our consumers to have `ember-intl` configured (or requires them to add a configuration, which would make this a breaking change).

Have we considered having a setup in the add-on, potentially via an initializer, with a fallback locale?

packages/components/src/ember-template-lint-plugin-local/index.mjs
zamoore: Good catch, this was meant to be a stop-gap during development. I was getting typescript errors when trying to import from outside of src. May need to ping you to help clear those.

packages/components/translations/en-us.yaml
zamoore: I'm fine with `common`

packages/components/translations/en-us.yaml
zamoore: 👍

showcase/app/routes/application.js
zamoore: That's going to be a difficult habit to break.

packages/components/src/ember-template-lint-plugin-local/index.mjs
aklkv: ideally this would be it's own package that exports this common configs, we have similar thing in cloud-ui we call it `common` and it's own workspace

packages/components/package.json
```diff
@@ -97,6 +97,7 @@
     "babel-plugin-ember-template-compilation": "^2.4.1",
     "concurrently": "^9.1.2",
     "ember-basic-dropdown": "^8.4.0",
+    "ember-intl": "^6.5.3",
```
aklkv, line 100: why this particular version, seems pretty low and given that is optional peerDep

packages/components/translations/en-us.yaml
zamoore: @aklkv You need the quotes for translations with dynamic segments. I get build errors without them

showcase/config/ember-intl.js
zamoore: We have a wrapper helper now that addresses this

packages/components/package.json
zamoore: I had it at the latest versions, but there were compatibility issues with the version that is used in cloud-ui, 6.5.3.

We can upgrade to 7 when cloud-ui is up-to-date.

packages/components/translations/en-us.yaml
aklkv: oh yes my bad 👍🏻

packages/components/src/components/hds/application-state/header.hbs
```diff
@@ -5,7 +5,7 @@
 <div class="hds-application-state__header" ...attributes>
   {{#if @errorCode}}
     <Hds::Text::Body class="hds-application-state__error-code" @tag="div" @size="100" @weight="medium" @color="faint">
-      ERROR
+      {{t "hds.components.application-state.header.error"}}
```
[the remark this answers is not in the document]
zamoore: Agreed

packages/components/src/ember-template-lint-plugin-local/index.mjs
zamoore: Agreed. I'm moving it out of this PR to set it up as its own workspace.

aklkv: Here is coincidental inspiration https://github.com/ember-intl/ember-intl/tree/main/configs

didoo (commented):
@zamoore I have a few general, high-level questions, after a first look:

- has this been tested across different products/consumers? (I'm asking because I don't see an "📦 RC Packages Published" entry in the history
- the fact that we have a `default` fallback means that if, by chance someone moves one of the yaml files in a different folder, or uses an incorrect file/folder name/path, the app will not break or report an error, and this may go unnoticed?
    - don't hate me, but could this be a smell that it's better to have a single file for the translations?
    - also, having a default, means that now we have to keep in sync the default and the `en-us` translation too, not sure is desirable (I'm sure it will lead to inconsistencies in the long term, and at that point which one is correct?)
- if I wear the hat of a consumer, what if I need to override one specific translation of a text in an HDS component? would it be possible for me to add somewhere a `yaml` file in my codebase, and override that specific value? or I should override the entire file/path with another yaml? or it's not possible at all?
- similarly, as consumer, would it be possible to add a new language/translation set for the HDS labels, without asking HDS to add it to their codebase, but adding it to my codebase
- is there a way to test/demo these functionalities in the showcase (potentially in a dedicated page, with a language switch)? In other word, use the showcase as a way to test what a consumer could do?
- I know technically is not necessary, but would make sense to namespace the translations, so that they look like `{{hds-t "hds.components.something.etc"}}` (in case they can be overwritten by the consumers)

packages/components/package.json
aklkv: interesting 🤔 but we are going to upgrade pretty soon. most likely in the next couple of weeks

zamoore:
@didoo 

> has this been tested across different products/consumers? (I'm asking because I don't see an "📦 RC Packages Published" entry in the history

Yep, I tested it locally in apps with and without `ember-intl` installed

> the fact that we have a default fallback means that if, by chance someone moves one of the yaml files in a different folder, or uses an incorrect file/folder name/path, the app will not break or report an error, and this may go unnoticed?

There is a setting in in `ember-intl`, [errorOnMissingTranslations](https://ember-intl.github.io/ember-intl/docs/advanced/configuration#erroronmissingtranslations), that we could potentially enable to prevent this from happening.

> is there a way to test/demo these functionalities in the showcase (potentially in a dedicated page, with a language switch)? In other word, use the showcase as a way to test what a consumer could do?

Yes, we can do this. Do you have some thoughts on what the content of the page should be?

> I know technically is not necessary, but would make sense to namespace the translations, so that they look like {{hds-t "hds.components.something.etc"}} (in case they can be overwritten by the consumers)

Yeah, we can do that. We just need to add another level to the directory structure.

I know I missed a couple of your questions, but I want to do some testing before I answer. I'll get back to you tomorrow.

packages/components/src/components/hds/pagination/info/index.hbs
```diff
@@ -1,11 +1,22 @@
 {{!
-  Copyright (c) HashiCorp, Inc.
+  Copyright (c) HashiCorp, Inc.`
```
dchyun:
```suggestion
  Copyright (c) HashiCorp, Inc.
```
[Nit] Think this was added by mistake.

dchyun (commented):
One general question. How will text set inside js/ts files be handled? In the table resize PR, the context menu does this. I also found a couple instances in our modifiers of it such as in the code editor modifier.

```
element.textContent =
      'Press `Ctrl-Shift-m` (`Cmd-Shift-m` on macOS) while focus is on the textbox to open the linting panel';
```

aklkv:
@zamoore I lost our conversation but cloud-ui is now on the latest `ember-intl` and version v8 in the works wich would turn it into v2 addon which some significant changes https://github.com/ember-intl/ember-intl/pull/1975

zamoore:
> @zamoore I lost our conversation but cloud-ui is now on the latest `ember-intl` and version v8 in the works wich would turn it into v2 addon which some significant changes [ember-intl/ember-intl#1975](https://github.com/ember-intl/ember-intl/pull/1975)

Nice. I went ahead and bumped up to the latest release. Thanks.

zamoore:
> One general question. How will text set inside js/ts files be handled? In the table resize PR, the context menu does this. I also found a couple instances in our modifiers of it such as in the code editor modifier.
> 
> ```
> element.textContent =
>       'Press `Ctrl-Shift-m` (`Cmd-Shift-m` on macOS) while focus is on the textbox to open the linting panel';
> ```

You're right. You need to do that with the intl service, but the ember-intl intl service doesn't support the fallback translation value. I added a new hds-t service that wraps the ember-intl one. I also updated the helper to use our wrapper service rather than the library version.

packages/components/src/ember-template-lint-plugin-local/rules/no-bare-strings-in-attributes.mjs
zamoore: The plugin has been removed for now.

aklkv:
@zamoore @didoo what is the reasoning behind instroducing consumer usable `hds-t` helper and `hds-intl` service?

.changeset/early-years-jog.md
```diff
@@ -0,0 +1,6 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+- Added `ember-intl` as a dependency
+- Added translations for untranslated strings in `components`
```
didoo: [question] For future reference: should we add here the list of components that have been updated too? so their "Version history" tab gets populated too?

packages/components/src/components/hds/advanced-table/th-selectable.hbs
```diff
@@ -21,7 +21,6 @@
       {{on "change" this.onSelectionChange}}
     />
     {{#if this.isSortable}}
-      <span id={{this._labelId}} class="sr-only">selection state</span>
```
didoo, line 24: [question] what happened to this? here we're removing a DOM element, but I don't see it added somewhere else

packages/components/src/components/hds/form/indicator/index.hbs
```diff
@@ -3,8 +3,16 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 {{#if @isOptional}}
-  <Hds::Text::Body class={{this.classNames}} tag="span" @size="100" @weight="regular">(Optional)</Hds::Text::Body>
+  <Hds::Text::Body class={{this.classNames}} tag="span" @size="100" @weight="regular">
+    ({{hds-t "components.common.optional" default="Optional"}})
```
didoo: [todo] I don't think this should live in common. Despite the _word_ being common,  it's _use_ here, in this context, is very specific to the `form` controls / fields

packages/components/src/components/hds/form/indicator/index.hbs
```diff
@@ -3,8 +3,16 @@
...
+  </Hds::Text::Body>
 {{/if}}
 {{#if @isRequired}}
-  &nbsp;<Hds::Badge aria-hidden="true" class={{this.classNames}} @size="small" @color="neutral" @text="Required" />
+  &nbsp;<Hds::Badge
+    aria-hidden="true"
+    class={{this.classNames}}
+    @size="small"
+    @color="neutral"
+    @text={{hds-t "components.common.required" default="Required"}}
```
didoo: [todo] I don't think this should live in common. Despite the _word_ being common,  it's _use_ here, in this context, is very specific to the `form` controls / fields

packages/components/src/components/hds/pagination/info/index.hbs
```diff
@@ -3,9 +3,20 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 <Hds::Text::Body class="hds-pagination-info" @tag="div" @size="100" @weight="medium" ...attributes>
-  {{@itemsRangeStart}}–{{@itemsRangeEnd}}
   {{#if this.showTotalItems}}
-    of
-    {{@totalItems}}
+    {{hds-t
+      "components.pagination.info.range-of-total"
+      default=(concat @itemsRangeStart "–" @itemsRangeEnd " of " @totalItems)
```
didoo: [thought] here we're "duplicating the logic", just to have a `default` argument (not great)

packages/components/src/components/hds/pagination/info/index.hbs
```diff
@@ -3,9 +3,20 @@
...
+      start=@itemsRangeStart
+      end=@itemsRangeEnd
+      total=@totalItems
+    }}
+  {{else}}
+    {{hds-t
+      "components.pagination.info.range"
+      default=(concat @itemsRangeStart "–" @itemsRangeEnd)
```
didoo: [thought] here we're "duplicating the logic", just to have a `default` argument (not great)

packages/components/src/components/hds/side-nav/base.hbs
```diff
@@ -4,7 +4,9 @@
 }}
 {{! IMPORTANT: we need to add "squishies" here (~) because otherwise the whitespace added by Ember causes the empty element to still have visible padding - See https://handlebarsjs.com/guide/expressions.html#whitespace-control }}
 <div class="hds-side-nav" ...attributes>
-  <h2 class="sr-only" id="hds-side-nav-header">Application local navigation</h2>
+  <h2 class="sr-only" id="hds-side-nav-header">
+    {{hds-t "components.app-side-nav.heading" default="Application local navigation"}}
```
didoo: [suggestion] in other cases you're calling this `screen-reader-label`, maybe we could use a similar naming pattern for all the labels/text that are used in a similar way?

packages/components/src/components/hds/table/th-button-tooltip.hbs
```diff
@@ -10,6 +10,8 @@
   aria-labelledby="{{this._prefixLabelId}} {{@labelId}}"
   ...attributes
 >
-  <span id={{this._prefixLabelId}} class="hds-table__th-button-aria-label-hidden-segment">More information for</span>
+  <span id={{this._prefixLabelId}} class="hds-table__th-button-aria-label-hidden-segment">
+    {{hds-t "components.common.tooltip-prefix" default="More information for"}}
```
didoo: [question] is this really "common"? to me it seems specific to the table/th-button/tooltip

packages/components/src/components/hds/table/th-selectable.hbs
```diff
@@ -20,7 +20,6 @@
       {{on "change" this.onSelectionChange}}
     />
     {{#if this.isSortable}}
-      <span id={{this._labelId}} class="sr-only">selection state</span>
```
didoo, line 23: [question] same as above, where did this go? what am I missing? 🤔

packages/components/src/helpers/hds-t.ts
```diff
@@ -0,0 +1,47 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Helper from '@ember/component/helper';
+import { service } from '@ember/service';
+import { assert } from '@ember/debug';
+import { isPresent } from '@ember/utils';
+
+import type HdsIntlService from '../services/hds-intl';
+interface HdsTHelperSignature {
```
didoo, line 14: [nit]
```suggestion

interface HdsTHelperSignature {
```

packages/components/src/helpers/hds-t.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export default class HdsTHelper extends Helper<HdsTHelperSignature> {
+  @service hdsIntl!: HdsIntlService;
+
+  compute(
+    positional: HdsTHelperSignature['Args']['Positional'],
+    named: HdsTHelperSignature['Args']['Named']
+  ): HdsTHelperSignature['Return'] {
+    const key = positional[0];
+    const { default: defaultString, ...restNamed } = named;
```
didoo: [question] why do you need to destructure `default: defaultString`? I don't see it used anywhere, (and then you pass it back as `default: defaultString` anyway)

packages/components/src/helpers/hds-t.ts
```diff
@@ -0,0 +1,47 @@
...
+      'Hds::T helper requires a key as the first positional argument',
+      typeof key === 'string' && isPresent(key)
+    );
+
+    return this.hdsIntl.t(key, {
+      default: defaultString,
+      htmlSafe: named.htmlSafe,
+      locale: named.locale,
+      ...restNamed,
+    });
```
didoo: [question] probably I'm missing something, but why can't this be just this?
```suggestion
    return this.hdsIntl.t(key, named);
```

packages/components/src/helpers/hds-t.ts
```diff
@@ -0,0 +1,47 @@
...
+import { isPresent } from '@ember/utils';
+
+import type HdsIntlService from '../services/hds-intl';
+interface HdsTHelperSignature {
+  Args: {
+    Positional: string[];
+    Named: {
+      default: string;
+      htmlSafe?: boolean | undefined;
+      locale?: string | undefined;
```
didoo: [question] should these three come from the `HdsIntlTOptions` of the service?

packages/components/src/helpers/hds-t.ts
didoo: [probably a leftover when you refactored the code to have a service too]

packages/components/src/services/hds-intl.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    return undefined;
+  }
+
+  t(key: string, options: HdsIntlTOptions): string {
+    const { default: defaultString, ...restOptions } = options;
+
+    assert(
+      'Hds::T helper requires a key as the first positional argument',
```
didoo: [nit] this is not an helper (it's a service) and it's called `HdsIntlService`
```suggestion
      'HdsIntlService requires a key as the first positional argument for the `t()` method',
```

showcase/app/controllers/internationalization/translation.js
```diff
@@ -0,0 +1,28 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Controller from '@ember/controller';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { service } from '@ember/service';
+
+export default class InternationalizationTranslationController extends Controller {
+  @service intl;
```
didoo, line 32: [suggestion] maybe we could use the showcase to test here the usage of the `hds-t` service (not the helper, which is tested in the template), even if technically consumers would not use it this way, but we can validate that the service would work as expected

showcase/app/formats.js
```diff
@@ -0,0 +1,29 @@
...
+  },
+  number: {
+    compact: {
+      notation: 'compact',
+    },
+    EUR: {
+      style: 'currency',
+      currency: 'EUR',
+    },
+    USD: {
```
didoo, line 24: [question] should we add Canadian currencies here too?

showcase/app/templates/components/accordion.hbs
```diff
@@ -5,7 +5,9 @@
 
 {{page-title "Accordion Component"}}
 
-<Shw::Text::H1>Accordion</Shw::Text::H1>
+<Shw::Text::H1>
```
didoo: [suggestion] instead of changing the title of the `Accordion` page, maybe we should change it in the   `internationalization/translation.hbs` page?

showcase/app/templates/internationalization/translation.hbs
didoo, line 1: [note] see comment here: https://hashicorp.slack.com/archives/C025N5V4PFZ/p1750151978595369?thread_ts=1750082493.598369&cid=C025N5V4PFZ

showcase/tests/unit/services/hds-intl-test.js
```diff
@@ -0,0 +1,95 @@
...
+import { setupTest } from 'showcase/tests/helpers';
+
+const defaultString = 'Default text';
+
+module('Unit | Service | hds-intl', function (hooks) {
+  setupTest(hooks);
+
+  hooks.beforeEach(function () {
+    this.intl = this.owner.lookup('service:intl');
+    this.service = this.owner.lookup('service:hds-intl');
```
didoo: [nit] maybe? (I know we're testing the `hds-intl` _service_, so this is the _service_, but I don't know, it felt strange when reading the code
```suggestion
    this.hdsIntl = this.owner.lookup('service:hds-intl');
```

showcase/tests/unit/services/hds-intl-test.js
```diff
@@ -0,0 +1,95 @@
...
+      result,
+      defaultString,
+      'returns default for non-existent key',
+    );
+  });
+
+  test('it returns the translated string if key exists in translations', function (assert) {
+    const testKey = 'greeting';
+    this.intl.addTranslations('en-us', {
+      greeting: 'Hello from Real Intl!',
```
didoo: [suggestion] assign the string to a `const`, so you can use it later for the comparison

showcase/tests/unit/services/hds-intl-test.js
didoo, line 1: [question] should we test the assertion that is thrown if `key` is missing, too?

showcase/tests/integration/helpers/hds-t-test.js
```diff
@@ -0,0 +1,176 @@
...
+
+module('Integration | Helper | hds-t', function (hooks) {
+  setupRenderingTest(hooks);
+
+  hooks.beforeEach(function () {
+    this.set('defaultString', defaultString);
+  });
+
+  module('invalid key values', function () {
+    test('it throws an error if key is undefined', async function (assert) {
```
didoo: [suggestion] since these tests are all very similar, maybe we could abstract them in some sort of `for/foreach` loop?

showcase/tests/integration/helpers/hds-t-test.js
```diff
@@ -0,0 +1,176 @@
...
+        hbs`{{hds-t this.translationKey default=this.defaultString}}`,
+      );
+
+      assert.throws(function () {
+        throw new Error(errorMessage);
+      });
+    });
+  });
+
+  module('ember-intl service is available', function (hooks) {
```
didoo, line 67: [question] should we also test when ember-intl service is not available?

showcase/tests/integration/helpers/hds-t-test.js
```diff
@@ -0,0 +1,176 @@
...
+      await render(hbs`{{hds-t this.key default=this.defaultString}}`);
+
+      assert
+        .dom()
+        .hasText(defaultString, 'returns default for non-existent key');
+    });
+
+    test('it returns translated string if locale is present and key exists', async function (assert) {
+      this.intl.addTranslations('en-us', {
+        greeting: 'Hello from Real Intl!',
```
didoo, line 74: [suggestion] store the string in a `const` so can be reused in the comparison below

showcase/tests/integration/helpers/hds-t-test.js
```diff
@@ -0,0 +1,176 @@
...
+
+      await render(
+        hbs`{{hds-t this.key name=this.nameParam age=this.ageParam default=this.defaultString}}`,
+      );
+
+      assert.dom().hasText('Goodbye Tester, aged 30!');
+    });
+
+    test('it returns default string if key is valid but not found in translations', async function (assert) {
+      const testKey = 'untranslated.key';
```
didoo, line 103: [question] what is the difference with the `'key.that.does.not.exist'` case?

didoo (commented):
Left a bunch of comments, questions, and suggestions.

There are also a few [open questions](https://github.com/hashicorp/design-system/pull/2878/#pullrequestreview-2881919800) too:
- if I wear the hat of a consumer, what if I need to override one specific translation of a text in an HDS component? would it be possible for me to add somewhere a `yaml` file in my codebase, and override that specific value? or I should override the entire file/path with another yaml? or it's not possible at all?
- similarly, as consumer, would it be possible to add a new language/translation set for the HDS labels, without asking HDS to add it to their codebase, but adding it to my codebase
   - as in: suppose HDS provides translations for the HDS components for `en-us` and `en-uk`, but as consumer I also want to add `fr-fr` and fr-ca`, would it be possible?
- I know technically is not necessary, but would make sense to namespace the translations, so that they look like `{{hds-t "hds.components.something.etc"}}` (in case they can be overwritten by the consumers, or if for some reason they want to use them in their codebase directly?)

finally, I _think_ we need to translate/update also the strings that are declared in JS/TS files, right? (per @dchyun [comment here](https://github.com/hashicorp/design-system/pull/2878/#pullrequestreview-2901687171))

packages/components/src/components/hds/table/th-selectable.hbs
zamoore: Lee indicated that this was unnecessary and could be removed.

showcase/tests/integration/helpers/hds-t-test.js
zamoore: I have not found a way where testing this is reliable. Unregistering the intl service does not work.

packages/components/src/components/hds/advanced-table/th-selectable.hbs
zamoore: Lee indicated that this was unnecessary and could be removed.

packages/components/src/components/hds/form/indicator/index.hbs
zamoore: What would make its usage in a form distinct from its usage elsewhere?

packages/components/src/components/hds/form/indicator/index.hbs
zamoore: What would make its usage in a form distinct from its usage elsewhere?

didoo: For me, in this specific case, it's not the _word_ "Required", that's being translated. For me, it is the **text** that appears next to a **label** to indicate that a **field** needs to be filled in to proceed with the **form** submission. The fact that it is "Required" is coincidental. In principle, for [reasons], in the future we may decide that we use the word "Mandatory" instead, and in that case we should update the `form.field.required` (pseudoname) not the common/general `required` word.

packages/components/src/components/hds/form/indicator/index.hbs
didoo: https://github.com/hashicorp/design-system/pull/2878#discussion_r2152332333

showcase/tests/integration/helpers/hds-t-test.js
didoo: what about this?
```
  module('ember-intl service is not available', function () {
    test('it returns default string', async function (assert) {
      const testKey = 'key.that.does.not.exist';

      this.set('key', testKey);

      await render(hbs`{{hds-t this.key default=this.defaultString}}`);

      assert
        .dom()
        .hasText(defaultString, 'returns default when `this.intl` is undefined');
    });
  });
```

zamoore: The intl service would still be available in this case unless I'm missing something.

zamoore:
> @zamoore @didoo what is the reasoning behind instroducing consumer usable `hds-t` helper and `hds-intl` service?

These don't need to be consumer usable, but we do need these wrapper classes to handle installation in codebases that don't have `ember-intl` configured.

Is there a way to prevent them from being added to the build?

packages/components/src/components/hds/form/indicator/index.hbs
zamoore: Makes sense. I'm moving this up to `components.form.common.required_field_indicator`

packages/components/src/components/hds/form/indicator/index.hbs
zamoore: That makes sense to me. I created a common translation file at the form level for this kind of thing.

packages/components/src/components/hds/pagination/info/index.hbs
zamoore: I was able to condense this down to a single complex translation that should be cleaner.

packages/components/src/components/hds/pagination/info/index.hbs
zamoore: I was able to condense this down to a single complex translation that should be cleaner.

packages/components/src/components/hds/side-nav/base.hbs
zamoore: That makes sense to me. Thanks for the suggestion.

showcase/tests/integration/helpers/hds-t-test.js
zamoore: Answering here for posterity. @didoo and I discussed in a Zoom call. This doesn't appear to be currently possible,

packages/components/src/helpers/hds-t.ts
zamoore: We need to remove the `default` value from the name properties object first. If you spread it with `restNamed`, it no longer contains the `default` value.

We probably could remove the htmlSafe and locale options though as those will be added appropriately.

packages/components/src/helpers/hds-t.ts
zamoore: `default` is a reserved keyword, but your're correct that we don't need to do this destructuring here anymore after going with the service.

packages/components/src/components/hds/table/th-button-tooltip.hbs
zamoore: It's common in that it's shared between the table and the advanced-table components. IMO, shared can be "generic" but it can also be used for translations that appear in multiple components.

packages/components/src/helpers/hds-t.ts
zamoore: Yes, good call.

showcase/app/formats.js
zamoore: We shouldn't need to for the showcase. This will be up to consuming apps to configure.

aklkv:
> > @zamoore @didoo what is the reasoning behind instroducing consumer usable `hds-t` helper and `hds-intl` service?
> 
> These don't need to be consumer usable, but we do need these wrapper classes to handle installation in codebases that don't have `ember-intl` configured.
> 
> Is there a way to prevent them from being added to the build?

definitely you can add it to https://github.com/hashicorp/design-system/blob/main/packages/components/rollup.config.mjs#L30 as exception

showcase/tests/integration/helpers/hds-t-test.js
zamoore: That makes sense to me

showcase/tests/integration/helpers/hds-t-test.js
zamoore: There isn't one. Good call.

showcase/app/controllers/internationalization/translation.js
zamoore: I went ahead and added examples with the service and the helper for both `ember-intl` and our custom wrapper to show the differences and how the `default` arg works.

zamoore:
> > > @zamoore @didoo what is the reasoning behind instroducing consumer usable `hds-t` helper and `hds-intl` service?
> > 
> > 
> > These don't need to be consumer usable, but we do need these wrapper classes to handle installation in codebases that don't have `ember-intl` configured.
> > Is there a way to prevent them from being added to the build?
> 
> definitely you can add it to https://github.com/hashicorp/design-system/blob/main/packages/components/rollup.config.mjs#L30 as exception

Perfect, thanks

zamoore:
> Left a bunch of comments, questions, and suggestions.
> 
> There are also a few [open questions](https://github.com/hashicorp/design-system/pull/2878/#pullrequestreview-2881919800) too:
> 
> * if I wear the hat of a consumer, what if I need to override one specific translation of a text in an HDS component? would it be possible for me to add somewhere a `yaml` file in my codebase, and override that specific value? or I should override the entire file/path with another yaml? or it's not possible at all?
> * similarly, as consumer, would it be possible to add a new language/translation set for the HDS labels, without asking HDS to add it to their codebase, but adding it to my codebase

I know we discussed this on a call, but I want to respond that both of these are possible for posterity.
   
> * I know technically is not necessary, but would make sense to namespace the translations, so that they look like `{{hds-t "hds.components.something.etc"}}` (in case they can be overwritten by the consumers, or if for some reason they want to use them in their codebase directly?)

Yes, agreed.

> finally, I _think_ we need to translate/update also the strings that are declared in JS/TS files, right? (per @dchyun [comment here](https://github.com/hashicorp/design-system/pull/2878/#pullrequestreview-2901687171))

Yes, we need to track down all of the bare strings in each component

.changeset/early-years-jog.md
zamoore: I went ahead and did this

showcase/.template-lintrc.js
aklkv, line 1: why was this needed, it's going to go back to cjs in the new blueprint?

packages/components/package.json
```diff
@@ -134,9 +139,11 @@
     "type": "addon",
     "version": 2,
     "app-js": {
+      "./components.js": "./dist/_app_/components.js",
```
didoo: [question] is this related to your PR? 🤔

packages/components/package.json
```diff
@@ -134,9 +139,11 @@
...
       "./components/hds/accordion/index.js": "./dist/_app_/components/hds/accordion/index.js",
       "./components/hds/accordion/item/button.js": "./dist/_app_/components/hds/accordion/item/button.js",
       "./components/hds/accordion/item/index.js": "./dist/_app_/components/hds/accordion/item/index.js",
+      "./components/hds/accordion/types.js": "./dist/_app_/components/hds/accordion/types.js",
```
didoo: [question] is this related to your PR? 🤔

packages/components/package.json
```diff
@@ -365,7 +411,298 @@
...
+      "./services/hds-intl.js": "./dist/_app_/services/hds-intl.js",
+      "./services/hds-time-types.js": "./dist/_app_/services/hds-time-types.js",
+      "./services/hds-time.js": "./dist/_app_/services/hds-time.js",
+      "./styles/@hashicorp/design-system-components.scss": "./dist/_app_/styles/@hashicorp/design-system-components.scss",
+      "./template-registry.js": "./dist/_app_/template-registry.js",
+      "./utils/hds-aria-described-by.js": "./dist/_app_/utils/hds-aria-described-by.js",
+      "./utils/hds-breakpoints.js": "./dist/_app_/utils/hds-breakpoints.js",
+      "./utils/hds-get-element-id.js": "./dist/_app_/utils/hds-get-element-id.js",
+      "./utils/hds-resolve-link-to-external.js": "./dist/_app_/utils/hds-resolve-link-to-external.js",
+      "./components.js.map": "./dist/_app_/components.js.map",
```
didoo: [question] is this related to your PR? 🤔

packages/components/rollup.config.mjs
```diff
@@ -33,6 +33,8 @@ const plugins = [
     'modifiers/**/*.js',
     'services/**/!(*types).js',
     'instance-initializers/**/*.js',
+    '!helpers/hds-t.js',
```
didoo: [issue?] it's likely that this is what caused all the strange entries added to `package.json` file

packages/components/src/components/hds/form/indicator/index.hbs
```diff
@@ -3,8 +3,16 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 {{#if @isOptional}}
-  <Hds::Text::Body class={{this.classNames}} tag="span" @size="100" @weight="regular">(Optional)</Hds::Text::Body>
+  <Hds::Text::Body class={{this.classNames}} tag="span" @size="100" @weight="regular">
+    ({{hds-t "ccomponents.form.common.optional_field_indicator" default="Optional"}})
```
didoo: [praise] I love the naming here! so much better! 👏

packages/components/src/components/hds/pagination/info/index.hbs
```diff
@@ -3,9 +3,16 @@
...
+    "hds.components.pagination.info"
+    default=(if
+      @totalItems
+      (concat @itemsRangeStart "–" @itemsRangeEnd " of " @totalItems)
+      (concat @itemsRangeStart "–" @itemsRangeEnd)
+    )
+    start=@itemsRangeStart
+    end=@itemsRangeEnd
+    total=@totalItems
+    hasTotal=@totalItems
```
didoo: [question] why rename the variables?

packages/components/src/components/hds/table/th-selectable.hbs
didoo, line 23: [thought/nit] _technically_, this should also be added to the changelog (same for the other places where you have removed the `sr-only` span elements)

packages/components/src/helpers/hds-t.ts
```diff
@@ -0,0 +1,38 @@
...
+    named: HdsTHelperSignature['Args']['Named']
+  ): HdsTHelperSignature['Return'] {
+    const key = positional[0];
+
+    assert(
+      'Hds::T helper requires a key as the first positional argument',
+      typeof key === 'string' && isPresent(key)
+    );
+
+    return this.hdsIntl.t(key, named);
```
didoo, line 36: [praise] nice!

showcase/app/controllers/internationalization/translation.js
```diff
@@ -0,0 +1,71 @@
...
+  },
+};
+
+export default class InternationalizationTranslationController extends Controller {
+  @service intl;
+  @service hdsIntl;
+
+  @tracked lang = null;
+
+  get serviceTranslations() {
```
didoo: [suggestion] to make more explicit/evident what it is/what it refers to
```suggestion
  get hdsIntlTranslations() {
```

showcase/app/controllers/internationalization/translation.js
```diff
@@ -0,0 +1,71 @@
...
+    return Object.entries(SERVICE_TRANSLATIONS).reduce(
```
didoo, line 38: [todo] probably better to store this in a local variable instead of computing it every time (same below)

showcase/app/templates/components/advanced-table.hbs
```diff
@@ -408,11 +408,11 @@
         <B.Td @align="right">
           <Hds::Dropdown @isInline={{true}} as |dd|>
             <dd.ToggleIcon @icon="more-horizontal" @text="Overflow Options" @hasChevron={{false}} @size="small" />
-            <dd.Interactive @route="components.table">Create</dd.Interactive>
-            <dd.Interactive @route="components.table">Read</dd.Interactive>
-            <dd.Interactive @route="components.table">Update</dd.Interactive>
+            <dd.Interactive @route="hds.components.table">Create</dd.Interactive>
```
didoo: [blocker] ops! find&replace went too far :)

didoo: (and in many other files too ;) )

didoo: screenshot_5078

showcase/app/templates/internationalization/translation.hbs
didoo, line 1: [praise] massive praise for how you are testing the different use cases (helpers/services) now! 👏 👏 👏

didoo: That said, we need to add also testing for overriding of `hds` translations too (as we discussed yesterday)

showcase/app/controllers/internationalization/translation.js
didoo: I LOVE how you did it! 🙌

showcase/app/templates/components/advanced-table.hbs
zamoore: This is why you don't code late at night

showcase/.template-lintrc.js
zamoore: Bad merge. Already fixed.

packages/components/rollup.config.mjs
zamoore: Yes it is. I had the config wrong.

packages/components/src/components/hds/pagination/info/index.hbs
zamoore: It was for brevity in the translation string, but not necessary.

zamoore:
> > > @zamoore @didoo what is the reasoning behind instroducing consumer usable `hds-t` helper and `hds-intl` service?
> > 
> > 
> > These don't need to be consumer usable, but we do need these wrapper classes to handle installation in codebases that don't have `ember-intl` configured.
> > Is there a way to prevent them from being added to the build?
> 
> definitely you can add it to https://github.com/hashicorp/design-system/blob/main/packages/components/rollup.config.mjs#L30 as exception

@aklkv Unfortunately, doing this also excludes it from use in the showcase app, which we need.

showcase/app/templates/internationalization/translation.hbs
zamoore: I don't think we'll be able to test overriding the transactions without actually overriding them unfortunately.

didoo: But we can "override" (technically, assign) them for languages that don't exists in HDS, right?

zamoore: True. I took care of that.

.changeset/early-years-jog.md
```diff
@@ -0,0 +1,56 @@
...
+- Added `hds-intl` service for translations with default fallback values
+- Added `hds-t` helper which uses the `hds-intl` service to provide translations in templates
+
+`AdvancedTable::ThButtonExpand` - Translated template strings
+
+`AdvancedTable::ThButtonSort` - Translated template strings
+
+`AdvancedTable::ThButtonTooltip` - Translated template strings
+
+`AdvancedTable::ThSelectable`
```
dchyun:
```suggestion
`AdvancedTable::ThSelectable` - Translated template strings. Removed extraneous screen-reader-only text.
```
[Issue] With the current script format, multi-line changes won't get picked up.

.changeset/early-years-jog.md
```diff
@@ -0,0 +1,56 @@
...
+
+`Pagination:Nav::Number` - Translated template strings
+
+`SideNav::Base` - Translated template strings
+
+`Table::ThButtonSort` - Translated template strings
+
+`Table::ThButtonTooltip` - Translated template strings
+
+`Table::ThSelectable`
```
dchyun:
```suggestion
`Table::ThSelectable` - Translated template strings. Removed extraneous screen-reader-only text.
```
[Issue] With the current script format, multi-line changes won't get picked up.

packages/components/src/components/hds/dismiss-button/index.ts
```diff
@@ -13,12 +16,14 @@ export interface HdsDismissButtonSignature {
...
-  /**
-   * @param ariaLabel
-   * @type {string}
-   * @default 'Dismiss'
-   */
+  @service hdsIntl!: HdsIntlService;
+
   get ariaLabel(): string {
-    return this.args.ariaLabel ?? 'Dismiss';
+    return (
```
dchyun, line 22: [Question] Are we looking to address all the strings inside js/ts in this PR as well? If so I think there are some more that are still untranslated.

Just looking for other aria-label strings the following components have untranslated values

- AdvancedTable - THSelectable
- AppFooter
- AppFooter - LegalLinks
- Breadcrumbs
- Breadcrumbs - Truncation
- DismissButton
- Pagination - Compact
- Pagination - Numbered
- Table - THSelectable
- Tag

packages/components/translations/hds/components/form/masked-input/base/en-us.yaml
dchyun, line 1: [Question] Seems like there's a mix of some files wrapping the translations in quotation marks and others don't. Is there a reason for that? It looks like it only needs to be done when passing in some value, so may not be needed in this file.

showcase/translations/en-us.yaml
dchyun, line 1: [Nit] Same question as my previous one, but is there a reason these files have quotation marks around the translations? We could standardize the formatting to what we are using in the components package.

website/docs/components/table/advanced-table/partials/code/how-to-use.md
```diff
@@ -693,9 +726,9 @@ Here’s an Advanced Table implementation that uses an array hash with strings f
 <Hds::AdvancedTable
   @model={{this.model.myDemoData}}
   @columns={{array
-      (hash key="artist" label=(t "components.table.headers.artist") isSortable=true)
-      (hash key="album" label=(t "components.table.headers.album") isSortable=true)
-      (hash key="year" label=(t "components.table.headers.year") isSortable=true)
+      (hash key="artist" label=(t "hds.components.table.headers.artist") isSortable=true)
```
dchyun: [Suggestion] Can we separate out the website changes to a separate PR? I think if we include them here when this merges the website will update, but these translations will not work there since the package won't have been published yet.

packages/components/src/components/hds/dismiss-button/index.ts
zamoore: I don't think we necessarily need to have everything translated in this PR. This was initially more to set up the internationalization infrastructure, and then I got a little carried away. We can create more followup issues for ensuring the whole project is correctly translated.

packages/components/translations/hds/components/form/masked-input/base/en-us.yaml
zamoore: Yeah, you need them for strings with variables. These don't strictly need the quotes and can be changed.

website/docs/components/table/advanced-table/partials/code/how-to-use.md
zamoore: Those were a mistake. Backed out those changes.

.changeset/early-years-jog.md
```diff
@@ -0,0 +1,56 @@
...
+
+`AdvancedTable::ThButtonExpand` - Translated template strings
+
+`AdvancedTable::ThButtonSort` - Translated template strings
+
+`AdvancedTable::ThButtonTooltip` - Translated template strings
+
+`AdvancedTable::ThSelectable` - Translated template strings. Removed extraneous screen-reader-only text.
+
+- Translated template strings
```
dchyun: [Chore] Still need to remove these entries

.changeset/early-years-jog.md
```diff
@@ -0,0 +1,56 @@
...
+
+`SideNav::Base` - Translated template strings
+
+`Table::ThButtonSort` - Translated template strings
+
+`Table::ThButtonTooltip` - Translated template strings
+
+`Table::ThSelectable` - Translated template strings. Removed extraneous screen-reader-only text.
+
+- Translated template strings
```
dchyun: [Chore] still need to remove these entries

.changeset/early-years-jog.md
dchyun, line 1: [Issue] I should've caught this the first time, but the script won't pick up something if it has the subcomponent added in the name. For example, you would need to say `AdvancedTable` instead of `AdvancedTable::ThButtonExpand`. 

I don't think we usually list out the subcomponent names unless it's for a specific reason, so we could probably consolidate some of these entries and just say like 
```
`AdvancedTable` - Translated template strings. Removed extraneous screen-reader-only text in `ThSelectable`. 
```
Below is a version of the changeset I tested and seemed to work with the script. 
```
`AdvancedTable:` - Translated template strings. Removed extraneous screen-reader-only text.

`AppFooter` - Translated template strings

`AppSideNav` - Translated template strings

`ApplicationState` - Translated template strings

`CodeEditor` - Translated template strings

`DismissButton` - Translated template strings

`Form::Indicator` - Translated template strings

`Form::MaskedInput` - Translated template strings

`Form::SuperSelect` - Translated template strings

`Pagination` - Translated template strings

`SideNav` - Translated template strings

`Table` - Translated template strings. Removed extraneous screen-reader-only text in `ThSelectable`

`Time` - Translated template strings
```

packages/components/translations/hds/components/form/masked-input/base/en-us.yaml
dchyun, line 1: [Nit] Can you remove the quotes in this file?

packages/components/translations/hds/modifiers/hds-code-editor/en-us.yaml
```diff
@@ -0,0 +1 @@
+lint-panel-description: "Press `Ctrl-Shift-m` (`Cmd-Shift-m` on macOS) while focus is on the textbox to open the linting panel"
```
dchyun:
```suggestion
lint-panel-description: Press `Ctrl-Shift-m` (`Cmd-Shift-m` on macOS) while focus is on the textbox to open the linting panel
```
[Nit] Remove quotation marks

.changeset/early-years-jog.md
zamoore: Ah, ok. Good to know. Thanks.

packages/components/src/components/hds/dismiss-button/index.ts
```diff
@@ -13,12 +16,14 @@ export interface HdsDismissButtonSignature {
 }
 
 export default class HdsDismissButton extends Component<HdsDismissButtonSignature> {
-  /**
-   * @param ariaLabel
-   * @type {string}
-   * @default 'Dismiss'
-   */
+  @service hdsIntl!: HdsIntlService;
```
didoo, line 19: [question] for my knowledge: why do we need the `!` here, in this case? isn't something we control, and so should the type be known?

packages/components/src/components/hds/dismiss-button/index.ts
```diff
@@ -13,12 +16,14 @@ export interface HdsDismissButtonSignature {
...
+
   get ariaLabel(): string {
-    return this.args.ariaLabel ?? 'Dismiss';
+    return (
+      this.args.ariaLabel ??
+      this.hdsIntl.t('hds.components.dismiss-button.aria-label', {
+        default: 'Dismiss',
```
didoo, line 25: [thought] you told me (rightly so) that `default` is a reserved word in JS; do you see any cases in which having this parameter called `default` could become problematic (now or in the future)?

packages/components/src/components/hds/table/index.ts
```diff
@@ -134,7 +138,23 @@ export default class HdsTable extends Component<HdsTableSignature> {
       return this.args.sortedMessageText;
     } else if (this.sortBy && this.sortOrder) {
       // we should allow the user to define a custom value here (e.g., for i18n) - tracked with HDS-965
-      return `Sorted by ${this.sortBy} ${this.sortOrder}ending`;
+      const translatedSortOrder = {
```
didoo, line 141: wow, what a change!

packages/components/src/components/hds/pagination/info/index.hbs
```diff
@@ -3,9 +3,16 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 <Hds::Text::Body class="hds-pagination-info" @tag="div" @size="100" @weight="medium" ...attributes>
-  {{@itemsRangeStart}}–{{@itemsRangeEnd}}
-  {{#if this.showTotalItems}}
-    of
-    {{@totalItems}}
-  {{/if}}
+  {{hds-t
```
didoo: [thought] would make sense to have a getter here, at this point, and do the translation on the JS side, instead of having this convoluted "logic" here?

packages/components/translations/hds/components/pagination/info/en-us.yaml
```diff
@@ -0,0 +1 @@
+page_range: "{itemsRangeStart}–{itemsRangeEnd}{showTotalItems, select, true{ of {totalItems}} false{} other{}}"
```
didoo: [thought] instead of having conditional logic here, what if we have two different string definitions, and we do the switch elsewhere (the template, or even better the backing class, see comment above)

showcase/translations/en-us.yaml
```diff
@@ -0,0 +1,24 @@
...
+          secondary: Secondary Action
+        body: The text is displayed in the selected language.
+        language: Language
+        language-options:
+          none: None
+          en: 🇺🇸 English (United States)
+          fr: 🇫🇷 French (France)
+          es: 🇪🇸 Spanish (Spain)
+        title: Selected language
+hds:
```
didoo, line 17: [thought] should we add a comment here (and in the other files with the overrides) explaining why such translations are added here, in this context?

didoo (approved):
For me is a go! 👍
I've left only a few questions/thoughts (nothing major or blocker). If you decide to act on them let me know so I can re-approve.

Lovely PR, very well done! 🙌

.changeset/early-years-jog.md
```diff
@@ -0,0 +1,32 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+- Added `ember-intl` as a dependency
+- Added translations for untranslated strings in `components`
```
dchyun:
```suggestion
```
[Nit] Sorry another one I missed 😅. This can probably be removed since it's mentioned below.

.changeset/early-years-jog.md
```diff
@@ -0,0 +1,32 @@
...
+- Added `hds-intl` service for translations with default fallback values
```
dchyun:
```suggestion
- Added `hdsIntl` service for translations with default fallback values
```
[Question] It might not make a difference, but what would the correct name be to add in the changeset? The name of the file `hds-intl` or how people would call it, `hdsIntl`?

dchyun (dismissed):
This looks great! Très bien! 🇫🇷  Only have a few non-blocking comments

aklkv:
Given that it's pretty hard to follow is going on here and most importantly why we need this I would like to request to have RFC before this goes out

zamoore:
> Given that it's pretty hard to follow is going on here and most importantly why we need this I would like to request to have RFC before this goes out

Hey @aklkv . We need this as there is a broader organizational goal to internationalize our applications. Notably for french language requirements as we conduct business in Quebec. Are there some specific questions I can answer for you to help you figure out what's going on?

packages/components/src/components/hds/dismiss-button/index.ts
zamoore: We do that because the value isn't set in the initializer, so TS sees this as a non-optional field without a value. However, we know that the service will be injected, so we can silence the TS complaint by assuring it that the value will be set at run time.

packages/components/src/components/hds/dismiss-button/index.ts
zamoore: I don't believe it could cause any sort of issue if used the way its being used here.

packages/components/src/components/hds/table/index.ts
zamoore: Yes! Sometimes adding translations can be a bit of a pain

packages/components/src/components/hds/dismiss-button/index.ts
didoo: OK, thanks for the clarification 🙏

packages/components/src/components/hds/pagination/info/index.hbs
zamoore: I like that. Implemented.

packages/components/translations/hds/components/pagination/info/en-us.yaml
zamoore: I like that. Implemented.

showcase/translations/en-us.yaml
zamoore: Good call. done

.changeset/early-years-jog.md
zamoore: that makes sense to me

showcase/app/templates/internationalization/translation.hbs
```diff
@@ -0,0 +1,162 @@
...
+            <F.Button
+              @color="secondary"
+              @text={{t "showcase.pages.internationalization.translation.actions.secondary"}}
+            />
+          </A.Footer>
+        </Hds::ApplicationState>
+      {{/if}}
+    </SG.Item>
+    <SG.Item as |SGI|>
+      <SGI.Label>Using the <code>hds-intl</code> service"</SGI.Label>
```
dchyun: [Issue] Looks like both service examples are not having their translations applied when using the dropdown. It looks like the translation paths are correct in the controller so not sure what the issue is.
Screenshot 2025-06-20 at 4 00 09 PM

showcase/app/templates/internationalization/translation.hbs
```diff
@@ -0,0 +1,162 @@
...
+            @text={{hds-t
+              "showcase.pages.internationalization.translation.actions.secondary"
+              default="Secondary action (Fallback translation)"
+            }}
+          />
+        </A.Footer>
+      </Hds::ApplicationState>
+    </SG.Item>
+    <SG.Item as |SGI|>
+      <SGI.Label>Using the <code>ember-intl</code> <code>t</code> helper"</SGI.Label>
```
dchyun:
```suggestion
      <SGI.Label>Using the <code>ember-intl</code> <code>t</code> helper</SGI.Label>
```
[Nit] Trailing quotation mark. Note: this is on all the labels here.

aklkv (changes requested):
@zamoore I am blocking for now, I will reach out with my questions/concerns to @alex-ju on Monday as I believe this is not correct forum to having this discussions

alex-ju:
> @zamoore I am blocking for now, I will reach out with my questions/concerns to @alex-ju on Monday as I believe this is not correct forum to having this discussions

@aklkv since this is an open-source project and we're working in the open, this repository is actually the best forum to raise any questions or concerns. that way, the full context is visible to @hashicorp/hds-engineering, and others who may have similar questions can benefit from the discussion. looking forward to continuing the conversation here.

showcase/app/templates/page-components/accordion.hbs
```diff
@@ -5,7 +5,9 @@
 
 {{page-title "Accordion Component"}}
 
-<Shw::Text::H1>Accordion</Shw::Text::H1>
+<Shw::Text::H1>
+  Accordion
+</Shw::Text::H1>
```
didoo:
```suggestion
<Shw::Text::H1>Accordion</Shw::Text::H1>
```

.changeset/early-years-jog.md
```diff
@@ -0,0 +1,31 @@
...
+"@hashicorp/design-system-components": minor
+---
+
+- Added `ember-intl` as a dependency
+- Added `hdsIntl` service for translations with default fallback values
+- Added `hds-t` helper which uses the `hds-intl` service to provide translations in templates
+
+`AdvancedTable` - Translated template strings. Removed extraneous screen-reader-only text.
+
+`AppFooter` - Translated template strings
```
didoo, line 17: [todo] these entries need to use [the new changelog format](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Component-formatting-for-npm-packages)

```suggestion
<!-- START components/app-footer -->
`AppFooter` - Translated template strings
<!-- END -->
```

packages/components/src/components/hds/pagination/info/index.ts
```diff
@@ -17,7 +21,38 @@ export interface HdsPaginationInfoSignature {
...
+        totalItems,
+      };
+      defaultValue = `${defaultValue} of ${totalItems}`;
+    }
+
+    return this.hdsIntl.t(translationKey, {
+      ...interpolationProps,
+      default: defaultValue,
+    });
+  }
```
didoo: [suggestion] I was a little confused by the implementation, so I tried to refactor/simplify it, and this seems to work as well (and is much simpler):
```suggestion
  get translatedItemsRange(): string {
    const { itemsRangeStart, itemsRangeEnd, totalItems } = this.args;

    let translationKey;
    let defaultValue;

    if (this.showTotalItems) {
      translationKey = 'hds.components.pagination.info.page_range_with_total';
      defaultValue = `${defaultValue} of ${totalItems}`;
    } else {
      translationKey = 'hds.components.pagination.info.page_range'
      defaultValue = `${itemsRangeStart}–${itemsRangeEnd}`;
    }

    return this.hdsIntl.t(translationKey, {
      default: defaultValue,
      itemsRangeStart,
      itemsRangeEnd,
      totalItems,
    });
  }
```

(but please double check!)

showcase/app/templates/index.hbs
```diff
@@ -371,5 +371,14 @@
         </LinkTo>
       </li>
     </ol>
+
+    <Shw::Text::H2>Internationalization</Shw::Text::H2>
```
didoo: [suggestion] consider moving this below the `Foundations` block

showcase/app/templates/internationalization/translation.hbs
```diff
@@ -0,0 +1,162 @@
+{{!
+  Copyright (c) HashiCorp, Inc.
+  SPDX-License-Identifier: MPL-2.0
+}}
+
+{{page-title "Translation"}}
```
didoo, line 6: [todo] we should add this page to the Percy tests (see: `showcase/tests/acceptance/percy-test.js`) and probably we should test:
- the english language
- another language
- the fallback translation

(see `form/base-elements` entry for how to execute actions between screenshots, pretty standard)

showcase/app/templates/internationalization/translation.hbs
didoo: I think this is still an issue, right? could it be that the `hdsIntlTranslations` and `emberIntlTranslations` should be tracked objects or getters?

didoo (commented):
@zamoore re-reviewed and left some comments/suggestions

packages/components/src/components/hds/app-side-nav/index.hbs
```diff
@@ -12,7 +12,9 @@
   {{focus-trap isActive=this.shouldTrapFocus}}
   {{this._setUpBodyElement}}
 >
-  <h2 class="sr-only" id="hds-app-side-nav-header">Application local navigation</h2>
+  <h2 class="sr-only" id="hds-app-side-nav-header">
+    {{hds-t "hds.components.app-side-nav.heading" default="Application local navigation"}}
```
dchyun:
```suggestion
    {{hds-t "hds.components.app-side-nav.screen-reader-label" default="Application local navigation"}}
```
[Issue] I think the correct name here is `screen-reader-label` looking at the translation file.

packages/components/src/components/hds/pagination/nav/number.hbs
```diff
@@ -12,6 +12,8 @@
   ...attributes
   aria-current={{if @isSelected "page" null}}
 >
-  <Hds::Text::Body @tag="span" @size="100" @weight="medium"><span class="sr-only">page
-    </span>{{this.page}}</Hds::Text::Body>
+  <Hds::Text::Body @tag="span" @size="100" @weight="medium">
+    <span class="sr-only">{{hds-t "hds.components.pagination.nav.number.screen-reader-label" default="page"}}</span>
```
dchyun:
```suggestion
    <span class="sr-only">{{hds-t "hds.components.pagination.nav.number.screen-reader-label" default="Page"}}</span>
```
[Nit] The translation value is capitalized, so could align to that.

packages/components/src/components/hds/table/th-button-sort.hbs
```diff
@@ -9,7 +9,11 @@
   aria-labelledby="{{this._prefixLabelId}} {{@labelId}} {{this._suffixLabelId}}"
   ...attributes
 >
-  <span id={{this._prefixLabelId}} class="hds-table__th-button-aria-label-hidden-segment">Sort by</span>
-  <span id={{this._suffixLabelId}} class="hds-table__th-button-aria-label-hidden-segment">{{this.sortOrderLabel}}</span>
+  <span id={{this._prefixLabelId}} class="hds-table__th-button-aria-label-hidden-segment">
+    {{hds-t "hds.components.common.sort-by" default="Sort by"}}
+  </span>
+  <span id={{this._suffixLabelId}} class="hds-table__th-button-aria-label-hidden-segment">
+    {{this.sortOrderLabel}}
```
dchyun, line 16: [Question] Is this potentially an untranslated string? The value is returning the typed values directly without translation it appears.
```
  get sortOrderLabel(): HdsTableThSortOrderLabels {
    return this.args.sortOrder === HdsTableThSortOrderValues.Asc
      ? HdsTableThSortOrderLabelValues.Desc
      : HdsTableThSortOrderLabelValues.Asc;
  }
```

showcase/app/templates/internationalization/translation.hbs
dchyun: Yes still seeing this as an issue.

showcase/app/templates/internationalization/translation.hbs
```diff
@@ -0,0 +1,162 @@
...
+            <F.Button @color="secondary" @text={{this.emberIntlTranslations.secondaryAction}} />
+          </A.Footer>
+        </Hds::ApplicationState>
+      {{/if}}
+    </SG.Item>
+  </Shw::Grid>
+
+  <Shw::Divider @level={{2}} />
+
+  <Shw::Text::H4 @tag="h3">Overriding of HDS translations</Shw::Text::H4>
```
dchyun, line 153: [Question] Is this example meant to test a user overriding HDS translations with their own translation file? If so, is there somewhere that custom translation file is?

showcase/tests/integration/helpers/hds-t-test.js
dchyun, line 1: [Suggestion] If we're supporting users overriding HDS provided translations, like the example in the showcase, is that something we could test here?

showcase/app/templates/internationalization/translation.hbs
zamoore: Yes, exactly. You can find those files at showcase/translations

showcase/tests/integration/helpers/hds-t-test.js
zamoore: Possibly. I can look more into that.

showcase/app/templates/internationalization/translation.hbs
zamoore: Added the Percy test. Still need to get the additional suggested changes in

packages/components/src/components/hds/pagination/info/index.ts
zamoore: I went with something similar, but a version that won't pass `totalItems` if it's not necessary. Let me know what you think.

showcase/app/templates/internationalization/translation.hbs
zamoore: Should be resolved now.

showcase/app/templates/internationalization/translation.hbs
zamoore: Would we take multiple snapshots for this page after changing the locale?

showcase/tests/integration/helpers/hds-t-test.js
zamoore: Thinking about this further, I don't think this is something we can test. Adding an override translation in the test wouldn't be any functionally different than adding the initial translation, so it wouldn't really test the functionality in a meaningful way.

Also, this isn't functionality were explicitly adding, just something that `ember-intl` supports

showcase/app/templates/internationalization/translation.hbs
zamoore: I think I figured it out. Pushed up changes

.changeset/early-years-jog.md
```diff
@@ -0,0 +1,79 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+- Added `ember-intl` as a dependency
+- Added `hdsIntl` service for translations with default fallback values
+- Added `hds-t` helper which uses the `hds-intl` service to provide translations in templates
```
didoo: Unless we want to expose them to the consumers (which I don't think, right?) better to be explicit about the fact that these are meant for internal usage only
```suggestion
- Added `hdsIntl` service for translations in HDS components with default fallback values
- Added `hds-t` helper which uses the `hds-intl` service to provide translations in HDS templates
```

packages/components/src/components/hds/advanced-table/th-resize-handle.hbs
```diff
@@ -13,7 +13,11 @@
   aria-valuemin={{@column.pxMinWidth}}
   aria-valuemax={{@column.pxMaxWidth}}
   tabindex="0"
-  aria-label="Resize {{@column.label}} column"
+  aria-label={{hds-t
+    "hds.components.advanced-table.th-resize-handle.aria-label"
+    label=@column.label
```
didoo: [suggestion] consider calling this `columnLabel` so in the translation file it will be `aria-label: "Resize {columnLabel} column"` instead of `aria-label: "Resize {label} column"` (less clear what the variable refers to)

showcase/tests/acceptance/percy-test.js
```diff
@@ -31,6 +31,14 @@ module('Acceptance | Percy test', function (hooks) {
     await visit('/foundations/focus-ring');
     await percySnapshot('FocusRing');
 
+    // Take snapshots for English, Spanish, and None (Fallback) translations
```
didoo, line 34: [praise] 👏

showcase/app/templates/internationalization/translation.hbs
```diff
@@ -0,0 +1,162 @@
...
+{{page-title "Translation"}}
+
+<Shw::Text::H1>Translation</Shw::Text::H1>
+
+<section data-test-percy>
+  <Shw::Text::H2>Demo</Shw::Text::H2>
+  <Shw::Text::Body>
+    This page demonstrates the use of the
+    <code>hds-t</code>
+    helper for internationalizing text. Selecting a language from the dropdown below will set the global locale used by
```
didoo: [suggestion] for completeness (note: linter will complain, so check the correct format)
```suggestion
    helper and <code>hds-intl</code> service for internationalizing text. Selecting a language from the dropdown below will set the global locale used by
```

didoo (dismissed):
Left minor suggestions

PR #2879, opened 2025-05-15
`CodeBlock` - aria argument and screen-reader improvements

dchyun:
### :pushpin: Summary

[PREVIEW](https://hds-showcase-git-dchyun-code-block-a11y-improvements-hashicorp.vercel.app/components/code-block)

If merged, this PR would improve the accessibility of the `CodeBlock` by implementing the following features

- New arguments for `@ariaLabel`, `@ariaLabelledBy`, and `@ariaDescribedBy`
- Attaching the title and description, if present, to the code block element via `aria-labelledby` and `aria-describedby`
- Adding screen-reader only text to indicate line highlighting

It would also update the `hds-clipboard` modifier to support not copying `sr-only` text.

#### Accessible names

The `@ariaLabel`, `@ariaLabelledBy`, and `@ariaDescribedBy` arguments have been added to the `CodeBlock` to align its approach to the `CodeEditor`. 

Also, the setting of these aria attributes has been aligned to the `CodeEditor` in the following ways.

- If a title is present, `aria-labelledby` will be set to the ID of the title element
- If a description is present, `aria-describedby`, will be set to the ID of the description element
- If a custom value is set on the `@ariaLabel` argument, and a title is present, the title id will not be set in `aria-labelledby`

The docs (#2896) will be updated to match the `CodeEditor` and say that `@ariaLabel` or `@ariaLabelledby` must be used if a title is not present. Existing implementation of the `CodeBlock` should be updated to add these arguments, but this will not cause breaking changes for anyone. 

#### Line highlighting for screen readers

Currently, there is no indication to screen reader users when line highlighting is used. Due to the fact that line numbers are also inaccessible to the screen reader (which can not be fixed since line numbers come from Prism), we can not add sr only text such as "Line 2 and 4 highlighed"

Instead the approach taken is to edit the code generated by prism to add `sr-only` text before and after line highlight starts indicating when highlighting starts, and when it ends. To prevent the `sr-only` text from being copied to the clipboard with the `CopyButton`, the `hds-clipboard` modifier has been updated to ignore `sr-only` text

The logic for determining where to add the sr only text mirrors the Prism plugin's code for determining highlights.

#### Aria attributes for title and description

**Before**
Screenshot 2025-05-15 at 3 43 39 PM

**After**
Screenshot 2025-05-15 at 3 44 09 PM

#### Line highlighting screen-reader only copy

Screenshot 2025-05-15 at 3 47 33 PM

### :link: External links

Jira ticket: [HDS-3349](https://hashicorp.atlassian.net/browse/HDS-3349)

***

### 👀 Component checklist

[HDS-3349]: https://hashicorp.atlassian.net/browse/HDS-3349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/modifiers/hds-clipboard.ts
+++ b/packages/components/src/modifiers/hds-clipboard.ts
@@ -88,7 +88,7 @@ export const getTextToCopyFromTargetElement = (
       textToCopy = targetElement.value;
     } else {
       // Hide any screen reader only text from the innerText calculation
-      let srOnlyTexts = targetElement.querySelectorAll('.sr-only');
+      const srOnlyTexts = targetElement.querySelectorAll('.sr-only');
       srOnlyTexts.forEach((el: Element) => {
         el.setAttribute('style', 'display: none;');
       });
--- a/packages/components/src/modifiers/hds-clipboard.ts
+++ b/packages/components/src/modifiers/hds-clipboard.ts
@@ -87,8 +87,19 @@ export const getTextToCopyFromTargetElement = (
     ) {
       textToCopy = targetElement.value;
     } else {
+      // Hide any screen reader only text from the innerText calculation
+      let srOnlyTexts = targetElement.querySelectorAll('.sr-only');
+      srOnlyTexts.forEach((el: Element) => {
+        el.setAttribute('style', 'display: none;');
+      });
+
       // simplest approach
       textToCopy = targetElement.innerText;
+
+      // Restore visibility of screen reader only text
+      srOnlyTexts.forEach((el: Element) => {
+        el.removeAttribute('style');
+      });
 
       // approach based on text selection (left for backup just in case)
       // var selection = window.getSelection();
--- a/.changeset/strong-bananas-drum.md
+++ b/.changeset/strong-bananas-drum.md
@@ -3,3 +3,5 @@
 ---
 
 `CodeBlock` - Added arguments `@ariaLabel`, `@ariaLabelledBy`, and `@ariaDescribedBy`. Added screen-reader only copy for highlighted lines.
+
+`hds-clipboard` - Prevent screen-reader only text (text with the `sr-only` class) from being copied to the clipboard.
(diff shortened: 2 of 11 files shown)


dchyun:
@MelSumner @shleewhite I've been working on the a11y improvements originally raised in [HDS-3349](https://hashicorp.atlassian.net/browse/HDS-3349). 

For the accessible name and description I've aligned the Code Block to the Code Editor and attached the title + description to the code element if they are present + added some aria arguments. 

My main open question is related to making line highlighting visible to the SR. 

One approach would be to add some sr text saying which lines are highlighted. However, the code block line numbers come from Prism, and are not accessible to the sr. The code is read as one long block of text. 

Screenshot 2025-05-15 at 3 56 32 PM

The approach I've taken currently is to add `sr-only` text stating when a highlight starts, and when it ends in the block of text.

Screenshot 2025-05-15 at 3 47 33 PM


This involves directly editing the code that prism js gives us, but I wanted to see if this would be a good approach to take, or if there are any other ideas we have for announcing line highlights to the screen reader.

shleewhite (commented):
the approach looks good! looking at the showcase though, seems like the "highlight start/end" strings are visible right now.

MelSumner:
this seems like a great approach!

packages/components/src/modifiers/hds-clipboard.ts
```diff
@@ -87,9 +87,20 @@ export const getTextToCopyFromTargetElement = (
     ) {
       textToCopy = targetElement.value;
     } else {
+      // Hide any screen reader only text from the innerText calculation
+      const srOnlyTexts = targetElement.querySelectorAll('.sr-only');
```
shleewhite, line 91: Is there a case for this outside of the CodeBlock component?

shleewhite (approved):
Looks good to me!

packages/components/src/modifiers/hds-clipboard.ts
dchyun: There isn't right now, but it could be useful in general. I don't think there's a use case for having sr only text and wanting it copied, which is what would happen currently.

zamoore (approved):

PR #2881, opened 2025-05-16
Remove `ember-flight-icons`

shleewhite:
### :pushpin: Summary

If merged, this PR would fully remove the `ember-flight-icons` package from our repository.

### :link: External links

Jira ticket: [HDS-3410](https://hashicorp.atlassian.net/browse/HDS-3410)

***

### 👀 Component checklist

~~A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))~~

[HDS-3410]: https://hashicorp.atlassian.net/browse/HDS-3410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/addon-main.cjs
+++ b/packages/components/addon-main.cjs
@@ -9,9 +9,11 @@ const flightIconSprite = require('@hashicorp/flight-icons/svg-sprite/svg-sprite-
 module.exports = {
   ...addonV1Shim(__dirname),
   contentFor(type, config) {
+    const legacyLazyEmbed = config?.emberFlightIcons?.lazyEmbed;
+
     if (
-      !config.flightIconSpriteLazyEmbed &&
-      !config.emberFlightIcons?.lazyEmbed &&
+      !config.flightIconsSpriteLazyEmbed &&
+      !legacyLazyEmbed &&
       !config.__flightIconsSpriteLoaded &&
       type === 'body-footer'
     ) {
--- a/packages/components/src/instance-initializers/load-sprite.ts
+++ b/packages/components/src/instance-initializers/load-sprite.ts
@@ -10,7 +10,7 @@ export async function initialize() {
   const legacyLazyEmbed = config?.emberFlightIcons?.lazyEmbed;
 
   // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-  if (config?.flightIconSpriteLazyEmbed || legacyLazyEmbed) {
+  if (config?.flightIconsSpriteLazyEmbed || legacyLazyEmbed) {
     const { default: svgSprite } = await import(
       '@hashicorp/flight-icons/svg-sprite/svg-sprite-module'
     );
--- a/showcase/config/environment.js
+++ b/showcase/config/environment.js
@@ -19,7 +19,7 @@ module.exports = function (environment) {
       },
     },
     // enabled here to aid in testing this feature within our component library
-    flightIconSpriteLazyEmbed: true,
+    flightIconsSpriteLazyEmbed: true,
 
     APP: {
       // Here you can pass flags/options to your application instance


packages/components/src/instance-initializers/load-sprite.ts
```diff
@@ -6,6 +6,7 @@
 import config from 'ember-get-config';
 
 export async function initialize() {
+  // SHOULD THIS BE REMOVED? OR SHOULD WE LOOK FOR A DIFF CONFIG KEY?
   // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
   if (config?.emberFlightIcons?.lazyEmbed) {
```
shleewhite, line 11: Should this config be renamed? it is used in the showcase, atlas, and cloud-ui

shleewhite: Also, if I remove it from the showcase - icons dont load in percy.

shleewhite:
@hashicorp/hds-engineering should there be a changelog for this change? the package is gone so it wont really do anything.

packages/components/src/instance-initializers/load-sprite.ts
aklkv: This would be breaking change if you change this key, as it's being set on consumer side

didoo:
> @hashicorp/hds-engineering should there be a changelog for this change? the package is gone so it wont really do anything.

Yes, technically we're still changing our code in relation to potential consumers (even outside HashiCorp). @alex-ju thoughts?

alex-ju:
> > @hashicorp/hds-engineering should there be a changelog for this change? the package is gone so it wont really do anything.
> 
> Yes, technically we're still changing our code in relation to potential consumers (even outside HashiCorp). @alex-ju thoughts?

This PR doesn't introduce any changes for our consumers (not for the `ember-flight-icons` consumers, nor for the `design-system-components` consumers), hence no changelog entry – we've already marked it as deprecated a while back.

As a follow-up (or in this PR, however you'd like) we should remove `ember-flight-icons` from the main readme file and remove/update any references in `website`.

packages/components/src/instance-initializers/load-sprite.ts
shleewhite: Good idea! I added a new config option called `flightIconsSpriteLazyEmbed` and I can open a ticket to deprecate the old config in v5.0.0.

shleewhite: Ticket: https://hashicorp.atlassian.net/browse/HDS-4883?atlOrigin=eyJpIjoiZDM0NDM3M2JlNDFmNDFjMDg3NGQyOWU5NTdmM2U4NmEiLCJwIjoiaiJ9

aklkv (dismissed):
🎉

packages/components/addon-main.cjs
```diff
@@ -10,6 +10,7 @@ module.exports = {
   ...addonV1Shim(__dirname),
   contentFor(type, config) {
     if (
+      !config.flightIconSpriteLazyEmbed &&
       !config.emberFlightIcons?.lazyEmbed &&
```
didoo, line 14: [suggestion] should we add a comment here about the two similarly named config parameters, and the fact that one is the new (correct) one and the other is the old and deprecated one?

didoo: [question] also, would it male sense to define a `const legacyLazyEmbed` like below? (the variable's name makes it clear that it's a legacy thing)

didoo (commented):
Left a few commits, nothing blocking.

I have a question (also for @alex-ju): now that we've introduced a new config parameter (that consumers should migrate to) should we add a changeset to the PR to explicitly declare this change in our changelog?

packages/components/src/instance-initializers/load-sprite.ts
```diff
@@ -6,8 +6,11 @@
 import config from 'ember-get-config';
 
 export async function initialize() {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-assignment
+  const legacyLazyEmbed = config?.emberFlightIcons?.lazyEmbed;
```
alex-ju, line 10: I'm not sure I follow why we need to make changes to `load-sprite.ts`

didoo (approved):
👍

aklkv (approved):

PR #2896, opened 2025-05-22
[Website] `CodeBlock` a11y improvements docs

dchyun:
### :pushpin: Summary

If merged, this PR would update the `CodeBlock` docs to include new aria arguments from #2879.

The `ariaDescribedBy` argument was also added to the `CodeEditor` docs, which was already present in the component through its extension of the `hds-code-editor` modifier. 

### :link: External links

Jira ticket: [HDS-3349](https://hashicorp.atlassian.net/browse/HDS-3349)

***

### 👀 Component checklist


zamoore (approved):
LGTM

website/docs/components/code-block/partials/code/component-api.md
```diff
@@ -11,6 +11,15 @@ This component uses [prism.js](https://prismjs.com/) under the hood.
   <C.Property @name="<[CB].Description>" @type="yielded component">
     `ContentBlock::Description` yielded as contextual component (see below).
   </C.Property>
+  <C.Property @name="ariaLabel" @type="string">
+    Accepts a localized string. The `ariaLabel` value is applied to the code block pre element.
```
didoo, line 15: [question] why are we saying "localized" here? we don't mention localization in any other parts o the APIs (plus it works even if it's not localized)

website/docs/components/code-block/partials/code/how-to-use.md
```diff
@@ -1,6 +1,6 @@
 ## How to use this component
 
-The basic invocation requires a `@value` argument. The component encodes this argument before displaying it.
+The basic invocation requires a `@value` argument. The component encodes this argument before displaying it. To use this component, you must either include the yielded `Title` component, provide an `@ariaLabel`, or specify an `@ariaLabelledBy`.
```
didoo: [question] Why the added text is not on its own paragraph? It seems that it's related to `@value`, but it's not, right?

website/docs/components/code-block/partials/code/component-api.md
dchyun: I was matching the language used for the aria attributes in the component api of the [Code Editor docs](https://github.com/hashicorp/design-system/pull/2896)

website/docs/components/code-block/partials/code/how-to-use.md
dchyun: I was trying to align to the code editor docs, but can see how this seems confusing. Moved the aria args text block to a separate paragraph.

website/docs/components/code-block/partials/code/component-api.md
didoo: OK, I'll ask the team what they think of this.

PR #2901, opened 2025-05-23
chore: ember v6.4

aklkv:
### :pushpin: Summary

Update `showcase`, `website` and `components` to new LTS ember 6.4

* update `ember-source` & `ember-cli` to 6.4 and run blueprint

I needed to bump some minor or patch deps version to satisfy peerDeps 
* bump `@embroider/*`
* bump `@ember/render-modifiers` - tho we should move away for this
* bump `@nullvoxpopuli/ember-composable-helpers`
* bump `ember-modifier`
* bump `ember-power-select` & `ember-basic-dropdown`
* bump `ember-concurrency`

[side-note]: I did not upgrade `@glimmer/components` to v2 as per blueprint because it drop support for ember 3.28

### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist


packages/components/package.json
```diff
@@ -45,28 +45,28 @@
     "@codemirror/lint": "^6.8.4",
     "@codemirror/state": "^6.5.0",
     "@codemirror/view": "^6.36.2",
-    "@ember/render-modifiers": "^2.1.0",
+    "@ember/render-modifiers": "^3.0.0",
```
alex-ju: This major version of `render-modifiers` drops support for ember < 4.12.

If we need this change for the v6.4 upgrade we'd have to keep this PR open and batch it with the rest of major changes for the next breaking release of our component library (v5).

showcase/config/targets.js
```diff
@@ -6,9 +6,9 @@
 'use strict';
 
 const browsers = [
-  'last 2 Chrome versions',
-  'last 2 Firefox versions',
-  'last 2 Safari versions',
+  'last 1 Chrome versions',
+  'last 1 Firefox versions',
+  'last 1 Safari versions',
```
alex-ju, line 11: last 2 versions is still our current support

alex-ju (commented):
Overall looks good. If I understand correctly this PR should make it into the next major (left a comment below).

showcase/config/targets.js
aklkv: I can revert that if this is HDS stance or browser support, but this is what came from blueprint and I think there is an RFC to back that up, but you tell me

packages/components/package.json
aklkv: good question, let me see if it was just technical change or real one

aklkv: by the sound of it it's just mainly technical https://github.com/emberjs/ember-render-modifiers/pull/81#issue-2754514160 , but I think we can keep it on 2.1 but it will show some pnpm warnings, also if you can added to the team backlog and rules to stop using this package all together and move to https://github.com/ember-modifier/ember-modifier, here is explainer why https://www.youtube.com/watch?v=zwewg2xmpU8

alex-ju: Let's try 2.1 and see if it works. We have an epic for [migrating away from deprecated dependencies](https://hashicorp.atlassian.net/browse/HDS-4356) covering `ember-modifier` as well

showcase/config/targets.js
alex-ju: Let's leave it to 1. For a moment I thought this change was in `components`.

alex-ju (approved):

PR #2902, opened 2025-05-23
wip: fixes

shleewhite:
### :link: External links

Figma file: [if it applies]

***

### 👀 Component checklist

### DIFF
--- a/packages/components/src/components/hds/modal/index.ts
+++ b/packages/components/src/components/hds/modal/index.ts
@@ -154,14 +154,6 @@ export default class HdsModal extends Component<HdsModalSignature> {
           );
         }
       }
-
-      // Return focus to a specific element (if provided)
-      if (this.args.returnFocusTo) {
-        const initiator = document.getElementById(this.args.returnFocusTo);
-        if (initiator) {
-          initiator.focus();
-        }
-      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/components/src/components/hds/modal/index.ts
```diff
@@ -69,7 +69,7 @@ export default class HdsModal extends Component<HdsModalSignature> {
     super(owner, args);
 
     registerDestructor(this, (): void => {
-      document.removeEventListener('click', this._clickHandler);
+      document.removeEventListener('click', this._clickHandler, true);
```
didoo, line 72: No way! 🤯

didoo: [question]: since the listener is added as:
```
document.addEventListener('click', this._clickHandler, {
  capture: true,
  passive: false,
});
```
should also be removed in the same way (just in case)?
```
document.removeEventListener('click', this._clickHandler, {
  capture: true,
  passive: false,
});
```

didoo (commented):
Left a comment, but ready to approve once it's clarified that detail.

packages/components/src/components/hds/modal/index.ts
shleewhite: Ok this is a good question, looking at mdn: they only list capture as an available option ([docs](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#options)), but they have an example that sets passive: false (https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#matching_event_listeners_for_removal).

But ultimately they recommend matching the values when you add / remove the listener so I'll update.

didoo (approved):
👍

PR #2919, opened 2025-05-30
`AdvancedTable` bug - Handle `model` updates

zamoore:
### :pushpin: Summary

There was a bug introduced when we migrated some of the internal state out of the AdvancedTable component and into a model class instance.

This fix allows updates to `@model` to pass through to the internal table model.

- Add a modifier that tracks changes to the `@model` argument
- Add an `updateModel` method to the `table` model class to allow the value to be updated via the modifier

***

<details>

</details>

### DIFF
--- a/packages/components/src/components/hds/advanced-table/index.hbs
+++ b/packages/components/src/components/hds/advanced-table/index.hbs
@@ -5,6 +5,7 @@
 <div
   class="hds-advanced-table__container
     {{(if this.isStickyHeaderPinned 'hds-advanced-table__container--header-is-pinned')}}"
+  {{this._didUpdateModel}}
   ...attributes
 >
   {{! Caption }}
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -157,7 +157,6 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
   private _selectAllCheckbox?: HdsFormCheckboxBaseSignature['Element'] =
     undefined;
   @tracked private _isSelectAllCheckboxSelected?: boolean = undefined;
-  @tracked private _lastModel!: HdsAdvancedTableModel;
 
   private _selectableRows: HdsAdvancedTableSelectableRow[] = [];
   private _captionId = 'caption-' + guidFor(this);
@@ -179,8 +178,6 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     super(owner, args);
 
     const { model, childrenKey, columns, hasStickyFirstColumn } = args;
-
-    this._lastModel = model;
 
     this._tableModel = new HdsAdvancedTableTableModel({
       model,
@@ -398,12 +395,6 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
   }
 
   private _didUpdateModel = modifier(() => {
-    if (this.args.model === this._lastModel) {
-      return;
-    }
-
-    this._lastModel = this.args.model;
-
     this._tableModel.updateModel(this.args.model);
   });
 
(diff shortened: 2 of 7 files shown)


shleewhite (approved):

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -393,6 +394,10 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     return classes.join(' ');
   }
 
+  private _didUpdateModel = modifier(() => {
```
didoo, line 398: [suggestion] what do you think of adding a comment here, to explain what this code does and why it's needed (for future maintainers)?

showcase/tests/integration/components/hds/advanced-table/index-test.js
```diff
@@ -498,6 +498,64 @@ module('Integration | Component | hds/advanced-table/index', function (hooks) {
...
+      <B.Td>{{B.data.key}}</B.Td>
+      <B.Td>{{B.data.name}}</B.Td>
+      <B.Td>{{B.data.description}}</B.Td>
+    </B.Tr>
+  </:body>
+</Hds::AdvancedTable>`);
+
+    assert
+      .dom('#data-advanced-test-table .hds-advanced-table__tr')
+      .exists({ count: 4 }); // header + 1 row
```
didoo: [question/suggestion] maybe?

```suggestion
      .exists({ count: 4 }); // header + 3 rows
```

packages/components/src/components/hds/advanced-table/index.hbs
```diff
@@ -5,6 +5,7 @@
 <div
   class="hds-advanced-table__container
     {{(if this.isStickyHeaderPinned 'hds-advanced-table__container--header-is-pinned')}}"
+  {{this._didUpdateModel}}
```
KristinLBradley, line 8: Comment: It looks strange to me to use a variable starting with an underscore (indicating it's private) within a template but I guess we've done this elsewhere. 

Is there a particular convention around when it's ok to use a private variable within the template?

showcase/tests/integration/components/hds/advanced-table/index-test.js
KristinLBradley: Should the comment say: `// header + 3 rows` here?

KristinLBradley (commented):
Looks good but I had a question about a comment on one of the tests that doesn't seem correct (about the row count)

showcase/tests/integration/components/hds/advanced-table/index-test.js
```diff
@@ -498,6 +498,64 @@ module('Integration | Component | hds/advanced-table/index', function (hooks) {
...
+        '#data-advanced-test-table .hds-advanced-table__tr:first-of-type .hds-advanced-table__td:nth-of-type(2n)',
+      )
+      .hasText('Test 1');
+    assert
+      .dom(
+        '#data-advanced-test-table .hds-advanced-table__tr:last-of-type .hds-advanced-table__td:last-of-type',
+      )
+      .hasText('Test 3 description');
+
+    this.set('model', [
```
didoo: [suggestion] here you're essentially completely redefining the whole `model` so it's automatic that `@tracked` works; what if, before this test/assertion, you add another test/assertion in which you update a single value of the `model` object (let's say the `description` of the third item (`Test 3 description`) and make sure that the test passes also in this case?

packages/components/src/components/hds/advanced-table/index.hbs
zamoore: I asked about this and it seems that anything that isn't exposed to the user should be marked as private.

packages/components/src/components/hds/advanced-table/index.ts
zamoore: Great idea

showcase/tests/integration/components/hds/advanced-table/index-test.js
zamoore: How this behaves when model member objects change would be dependent on how the model was set up. If the model is set up with `@deepTrack`, it would trigger the modifier to rerun. However, changes to a member of an array of plain JS objects would not trigger Ember to reevaluate.

Because this behavior is determined outside of the component, I don't think it's something we need to test for directly.

I did verify that the modifier is retriggered if the model attribute is deeply tracked.

PR #2936, opened 2025-06-06
`Stepper::Nav` Update signature to use `WithBoundArgs`

dchyun:
### :pushpin: Summary

If merged this PR would update the `Stepper::Nav` signature to use `WithBoundArgs` instead of `ComponentLike`. This is to avoid any errors thrown in consumer repos by having required arguments in the `Step` and `Panel` contextual components. These updates are similar to those made in the `CodeBlock` in #2925.

<details>

</details>

### DIFF
--- a/.changeset/nervous-eggs-kick.md
+++ b/.changeset/nervous-eggs-kick.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Stepper::Nav` - Updated signature to use `WithBoundArgs` instead of `ComponentLike` for contextual components to resolve linting issues


packages/components/src/components/hds/stepper/nav/index.ts
```diff
@@ -37,8 +36,11 @@ export interface HdsStepperNavSignature {
     body?: [];
     default: [
       {
-        Step?: ComponentLike<HdsStepperNavStepSignature>;
-        Panel?: ComponentLike<HdsStepperNavPanelSignature>;
+        Step?: WithBoundArgs<typeof HdsStepperNavStepComponent, 'currentStep'>;
+        Panel?: WithBoundArgs<
+          typeof HdsStepperNavPanelComponent,
+          'currentStep'
```
didoo: AFAIK we have to add all the arguments that are passed in the `yield` declaration: https://github.com/hashicorp/design-system/blob/8066b466aec268b932c6636b17f42da9e1fd7da6/packages/components/src/components/hds/stepper/nav/index.hbs#L29-L37

(if unsure, maybe check with @aklkv, he's the one who told me this thing)

dchyun: Made the update. [The docs](https://typed-ember.gitbook.io/glint/using-glint/glint-types#withboundargs-and-withboundpositionals) say any argument passed in through the yield should be listed there so made the update you suggested.

showcase/app/templates/components/stepper/nav.hbs
```diff
@@ -257,9 +257,11 @@
...
   <Shw::Text::H3>Nav::Panel</Shw::Text::H3>
 
-  <Hds::Stepper::Nav::Panel>
-    <Shw::Placeholder @text="Generic content" @height="100" />
-  </Hds::Stepper::Nav::Panel>
+  <Hds::Stepper::Nav @currentStep={{0}} @isInteractive={{false}} @ariaLabel="Nav panel" as |S|>
+    <S.Panel>
+      <Shw::Placeholder @text="Generic content" @height="100" />
+    </S.Panel>
+  </Hds::Stepper::Nav>
```
didoo, line 264: [question] would this have worked too? the reason I'm asking is that when we tend to showcase the components in isolation, we try to avoid as much as possible wrapping them in their expected parents.

```suggestion
  <Hds::Stepper::Nav::Panel @currentStep={{0}}>
    <Shw::Placeholder @text="Generic content" @height="100" />
  </Hds::Stepper::Nav::Panel>```

Comment (dchyun) [reply showcase/app/templates/components/stepper/nav.hbs:264]: Yeah I tried out a few different things on this one. This doesn't work because the panel isn't visible. The panel does a check to see if the step number matches the index of its id in the `panelIds`. But when using it standalone there's no good way to set the `panelIds` since this is usually set by the nav, and the id is generated by the component. 

It works when `currentStep` is undefined but then that throws the lint error. 

The current solution also isn't perfect because it's using the stepper nav with out a step. Another option would be to remove this section as the panel is really just a yield visually anyway.

Comment (didoo) [reply showcase/app/templates/components/stepper/nav.hbs:264]: can you maybe force its visibility via CSS using the dedicated showcase CSS file specific to this page? (something we do also in other pages)?

Comment (dchyun) [reply showcase/app/templates/components/stepper/nav.hbs:264]: The panel visibility isn't just css driven. The template doesn't render anything inside the panel also if it's not visible. So overriding the visibility alone won't fix it. 

```
<section
  class="hds-stepper-nav__panel"
  ...attributes
  role={{if this.isNavInteractive "tabpanel"}}
  id={{this._panelId}}
  hidden={{not this.isVisible}}
  aria-labelledby={{this.coupledStepId}}
  {{this._setUpPanel this.didInsertNode this.willDestroyNode}}
>
  {{#if this.isVisible}}
    {{yield}}
  {{/if}}
</section>
```

didoo: Ahhh! OK then, I'll approve then. And if you prefer to remove it, push the change and I'll re-approve.

didoo (commented):
I think we need a changeset

dchyun:
> I think we need a changeset

Wasn't sure if we needed a changeset or not since the code block didn't have one, but think it's better to be transparent. Have added one in.

didoo:
> since the code block didn't have one

Oh, we missed that! :( 

Do you mind opening a PR just to add that changeset for the codeblock? I can approve it immediately

didoo (approved):
👍 (thank you)

PR #2948, opened 2025-06-12
`SuperSelect::Multiple` - Fix grouped options with show selected display

dchyun:
### :pushpin: Summary

If merged, this PR would enhance the display of `OptionGroup` options in the `SuperSelect::Multiple`, when only selected options are shown. 

[PREVIEW](https://hds-showcase-git-dchyun-super-select-groups-sh-2a94c6-hashicorp.vercel.app/components/form/super-select#complex-content-1)

Currently in the `OptionGroup` of the `Super::Select::Multiple`, when the "show selected" button is used to only show selected options, if no options are selected the group and sub-group titles are shown with no options.

This differs from the pattern used on super selects that do not contain groups. In those instances there is nothing shown in the menu besides the `noMatchesMessage`.

Screenshot 2025-06-12 at 2 36 02 PM

In this PR we have made it so when "show selected" is active, the group or sub-group title of an option is only visible when an option in that group is selected. If a group / sub-group contains no selection options then their titles are not shown.

#### "Show selected" with no options selected

**Before**
Screenshot 2025-06-12 at 2 17 47 PM

**After**
Screenshot 2025-06-12 at 2 19 11 PM

#### "Show selected" with some options selected

**Before**
Screenshot 2025-06-12 at 2 18 57 PM

**After**
Screenshot 2025-06-12 at 2 19 24 PM

### :link: External links

Jira ticket: [HDS-4775](https://hashicorp.atlassian.net/browse/HDS-4775)

***

<details>

</details>

[HDS-4775]: https://hashicorp.atlassian.net/browse/HDS-4775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/giant-shoes-yell.md
+++ b/.changeset/giant-shoes-yell.md
@@ -2,4 +2,4 @@
 "@hashicorp/design-system-components": patch
 ---
 
-`SuperSelect::Multiple` - Enhanced display of grouped options when only selected options are shown
+`Form::SuperSelect` - Enhanced display of grouped options when only selected options are shown in the `SuperSelect::Multple`


packages/components/src/styles/components/form/super-select.scss
```diff
@@ -366,6 +366,21 @@ $hds-super-select-item-height: 36px;
...
+  }
+
+  // Hiding all dividers and showing one when a selected groupd is preceeded by another selected group
+  // Note: The :first-of-type selector was not used, as it doesn't work when appended to the end of a :has selector
+  .hds-form-super-select__option-group::before {
+    display: none;
+  }
+
+  .hds-form-super-select__option-group:has(.ember-power-select-option[aria-selected="true"])
+    ~ .hds-form-super-select__option-group:has(.ember-power-select-option[aria-selected="true"])::before {
```
dchyun, line 381: [Note] Open to suggestions if there's a simpler selector that could work here 😅 The issue is we need to remove the separator line from the first group shown in the selected groups list. 

I tried using the selector `.hds-form-super-select__option-group:has(.ember-power-select-option[aria-selected="true"]):first-of-type` but it only selects the first group overall, not the first selected group. Instead I'm removing all the separators, and then showing ones that come after a selected group.

shleewhite: I've fought with `:first-of-type` so many times - it doesn't work the way I expect. I think it is supposed to be used for the first element of that type and doesn't work to select the first element with a certain class.

What you have seems like the most straightforward way to achieve what youre trying to do.

shleewhite (dismissed):

packages/components/src/styles/components/form/super-select.scss
```diff
@@ -366,6 +366,20 @@ $hds-super-select-item-height: 36px;
       display: none;
     }
   }
+
+  .hds-form-super-select__option-group:not(:has(.ember-power-select-option[aria-selected="true"])) {
```
KristinLBradley, line 370: For complex selectors like these, could you please add comments explaining the logic of what they are intended to do?

For example:
`// Don't display option group that doesn't contain a selected option`

packages/components/src/styles/components/form/super-select.scss
KristinLBradley: Should include an explanation of what the pseudo element actually represents though

KristinLBradley: So maybe this:
`// Display a separator before an option-group containing a selected option that is the sibling of an option group also containing a selected option`

KristinLBradley: Sorry, I initially missed the code comment above. Going through all the logic still...

KristinLBradley: So far, I can't think of a good way to simplify the selectors for these unless you could use JS to somehow add a class name to option groups without any selected options or something but that could be even more complex than this...

KristinLBradley (dismissed):
I agree with @shleewhite I can't think of a better way to select these using CSS than what you have

packages/components/src/styles/components/form/super-select.scss
dchyun: Yeah the alternative was adding a class to the selected groups via js, but was trying to first see if this could be accomplished with only CSS. Which it can, it just involves this selector. Decided that was still preferable to using js, but up for discussion.

shleewhite (dismissed):

jorytindall (dismissed):
Interesting, this is a nice improvement! We don't really have the concept of grouped options in Figma (at least from a layout perspective), so I don't think any changes are needed in design tooling.

jorytindall (approved):

KristinLBradley (approved):

PR #2953, opened 2025-06-16
[Website] Changelog component docs script format enhancements

dchyun:
### :pushpin: Summary

If merged, this PR would update the script that generate component changelog entries to a new format, which would support the following enhancements in writing changelog entries.

- Support for multi-line changes for one component
- Remove the requirement that component names in the changelog match exactly to what the script expects

#### New format
In the new format, changelog component entries must contain comments above and below each component entry that tell the script which portions to use in the appropriate component changelog files. 
```
<!-- START components/name -->
`ComponentName` - Change description
<!-- END -->
```
The path used inside the comments directs the script to what website docs file to attach the entry. That path must match exactly. For example a change to the `LinkStandalone` would use the path `components/links/standalone`.

When the script is ran, these comments will be removed following parsing.

### Testing

> [!NOTE]
> The testing commit has now been removed, but changes were previously approved. 

There is a temporary commit included in this PR that allows testing of the script locally. To test this change do the following:

1. Cd into `./website`
2. Run `pnpm generate-component-changelog-entries`
3. Review that the appropriate component changelog files have been updated, and the comments have been removed from the `CHANGELOG.md` and website release notes page. 
4. Do not commit any changes made

#### Changeset file format

**Before**
```
`ComponentName` - This is a list of changes. Change one. Change two.
```

**After**
```
<!-- START components/name -->
`ComponentName` - This is a list of changes
- Change one
- Change two
<!-- END -->
```

### :link: External links

Jira ticket: [HDS-4951](https://hashicorp.atlassian.net/browse/HDS-4951)

***

<details>

</details>

[HDS-4951]: https://hashicorp.atlassian.net/browse/HDS-4951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/lib/changelog/generate-component-changelog-entries.mjs
+++ b/website/lib/changelog/generate-component-changelog-entries.mjs
@@ -5,6 +5,7 @@
 
 import fs from 'fs';
 import matter from 'gray-matter';
+import { glob } from 'glob';
 
 const readVersionFromPackageJson = (filePath) => {
   try {
@@ -17,26 +18,18 @@ const readVersionFromPackageJson = (filePath) => {
   }
 };
 
-const getComponentPaths = (baseDir) => {
-  const components = [];
-  try {
-    const folders = fs.readdirSync(baseDir, { withFileTypes: true });
-    folders.forEach((folder) => {
-      if (folder.isDirectory()) {
-        const componentPath = `${baseDir}/${folder.name}`;
-        const partialsPath = `${componentPath}/partials`;
-        if (fs.existsSync(partialsPath)) {
-          components.push(componentPath.replace('./docs/', ''));
-        } else {
-          // For component docs that are nested, we need to recursively get their paths
-          components.push(...getComponentPaths(componentPath));
-        }
-      }
-    });
-  } catch (err) {
-    console.error(`Error reading components from ${baseDir}:`, err);
-  }
-  return components;
+const getComponentPaths = async () => {
+  const componentPaths = await glob(
+    './docs/{components,layouts,overrides,utilities}/**/partials/',
+    { onlyDirectories: true },
+  );
+  const regex = new RegExp(
+    `(components|layouts|overrides|utilities)\/((?!\/partials).)*`,
+  );
+  const cleanedComponentPaths = componentPaths.map((path) => {
+    return path.match(regex)[0];
+  });
+  return cleanedComponentPaths;
 };
 
 const readChangelogContent = (filePath) => {
@@ -245,16 +238,7 @@ const version = readVersionFromPackageJson(
 );
 
 // Determine component paths
-const componentPaths = getComponentPaths('./docs/components');
-const layoutComponentPaths = getComponentPaths('./docs/layouts');
-const overrideComponentPaths = getComponentPaths('./docs/overrides');
-const utilityComponentPaths = getComponentPaths('./docs/utilities');
-const allComponentPaths = [
-  ...componentPaths,
-  ...layoutComponentPaths,
-  ...overrideComponentPaths,
-  ...utilityComponentPaths,
-];
+const allComponentPaths = await getComponentPaths();
 
 // Read the main changelog entry for components
 const changelogContent = readChangelogContent(
--- a/website/lib/changelog/generate-component-changelog-entries.mjs
+++ b/website/lib/changelog/generate-component-changelog-entries.mjs
@@ -55,7 +55,7 @@ const extractVersion = (changelogContent, version) => {
   return match ? match[0] : null;
 };
 
-const extractComponentChangelogEntries = (components, lastVersionContent) => {
+const extractComponentChangelogEntries = (lastVersionContent) => {
   const componentChangelogEntries = {};
 
   // Regular expression to match and capture a block of text delimited by two HTML comments used as START/END markers:
@@ -265,7 +265,6 @@ const currentVersionContent = extractVersion(changelogContent, version);
 
 // Extracts changelog entries for each components
 const componentChangelogEntries = extractComponentChangelogEntries(
-  allComponentPaths,
   currentVersionContent,
 );
 
--- a/website/lib/changelog/generate-component-changelog-entries.mjs
+++ b/website/lib/changelog/generate-component-changelog-entries.mjs
@@ -18,7 +18,7 @@ const readVersionFromPackageJson = (filePath) => {
 };
 
 const getComponentPaths = (baseDir) => {
-  const components = {};
+  const components = [];
   try {
     const folders = fs.readdirSync(baseDir, { withFileTypes: true });
     folders.forEach((folder) => {
@@ -26,48 +26,10 @@ const getComponentPaths = (baseDir) => {
         const componentPath = `${baseDir}/${folder.name}`;
         const partialsPath = `${componentPath}/partials`;
         if (fs.existsSync(partialsPath)) {
-          // we have some special cases where intermediate namespacing is used to group components:
-          if (baseDir.endsWith('/copy')) {
-            components[`copy-${folder.name}`] = componentPath;
-          } else if (baseDir.endsWith('/form')) {
-            if (folder.name === 'primitives') {
-              const primitiveNames = [
-                'character-count',
-                'error',
-                'field',
-                'fieldset',
-                'helper-text',
-                'indicator',
-                'label',
-                'legend',
-              ];
-              primitiveNames.forEach((componentName) => {
-                components[`form-${componentName}`] = componentPath;
-              });
-            } else if (folder.name === 'layout') {
-              components[`form`] = componentPath;
-            } else {
-              components[`form-${folder.name}`] = componentPath;
-            }
-          } else if (baseDir.endsWith('/layouts')) {
-            if (folder.name === 'app-frame') {
-              components[`${folder.name}`] = componentPath;
-            } else {
-              components[`layout-${folder.name}`] = componentPath;
-            }
-          } else if (baseDir.endsWith('/link')) {
-            components[`link-${folder.name}`] = componentPath;
-          } else if (baseDir.endsWith('/stepper')) {
-            // The components/stepper/indicator page contains both the Stepper::Step::Indicator and Stepper::Task::Indicator
-            if (folder.name === 'indicator') {
-              components[`stepper-step-${folder.name}`] = componentPath;
-              components[`stepper-task-${folder.name}`] = componentPath;
-            } else {
-              components[`stepper-${folder.name}`] = componentPath;
-            }
-          } else {
-            components[folder.name] = componentPath;
-          }
+          components.push(componentPath.replace('./docs/', ''));
+        } else {
+          // For component docs that are nested, we need to recursively get their paths
+          components.push(...getComponentPaths(componentPath));
         }
       }
     });
@@ -93,41 +55,26 @@ const extractVersion = (changelogContent, version) => {
   return match ? match[0] : null;
 };
 
-const convertComponentNameFormat = (componentName) => {
-  const twoLevelComponentNames = ['copy', 'form', 'layout', 'link', 'stepper'];
-  const threeLevelComponentNames = [
-    'stepper-step-indicator',
-    'stepper-task-indicator',
-  ];
-  if (
-    twoLevelComponentNames.includes(componentName.split('-')[0]) &&
-    componentName != 'form'
-  ) {
-    let words = componentName
-      .split('-')
-      .map((word) => word.charAt(0).toUpperCase() + word.slice(1));
-    if (threeLevelComponentNames.includes(componentName)) {
-      return words.join('::');
-    } else {
-      return words[0] + '::' + words.slice(1).join('');
-    }
-  } else {
-    return componentName
-      .split('-')
-      .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
-      .join('');
-  }
-};
-
 const extractComponentChangelogEntries = (components, lastVersionContent) => {
   const componentChangelogEntries = {};
 
-  Object.keys(components).forEach((componentName) => {
-    const formattedComponentName = convertComponentNameFormat(componentName);
-    const regex = new RegExp(`^\`${formattedComponentName}\` - .*$`, 'gm');
+  components.forEach((componentName) => {
+    const regex = new RegExp(
+      `^(<!-- START ${componentName})((.|\n)*?)(<!-- END ${componentName} -->)$`,
+      'gm',
+    );
     const matches = lastVersionContent.match(regex);
     if (matches) {
-      componentChangelogEntries[componentName] = matches;
+      const cleanedMatches = [];
+      matches.forEach((match) => {
+        // Remove the start and end comments to get the changelog entry
+        const cleanMatch = match
+          .replace(`<!-- START ${componentName} -->`, '')
+          .replace(`<!-- END ${componentName} -->`, '')
+          .trim();
+        cleanedMatches.push(cleanMatch);
+      });
+      componentChangelogEntries[componentName] = cleanedMatches;
     }
   });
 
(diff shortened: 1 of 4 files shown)


packages/components/CHANGELOG.md
```diff
@@ -6,6 +6,24 @@
 
 ### Patch Changes
 
+<!-- START components/card -->
```
dchyun:
```suggestion
<!-- START components/card -->

```
[Note] There is a linting error that gets thrown for this format and is only fixed by adding a new line after this comment? Any way to disable this, or another workaround?

alex-ju: I imagine prettier is the culprit here. the radical approach would be to ignore all *.md files; a more granular one would be to disabled only the specific rule that requires such new lines – let me know if you need support with any of the two

dchyun: I tried messing around with the [Prettier options](https://prettier.io/docs/options) but it seems (intentionally) they are fairly limited, and none fixed the problem. Messing around with [their playground](https://prettier.io/playground/) it seems this is a linting rule there is no way around.
Screenshot 2025-07-03 at 2 38 44 PM

So I've added the changelog file to the prettierignore for now. If you can think of any other workarounds please let me know.

heatherlarsen (dismissed):
@dchyun Love this change! 👏🏻

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -301,8 +259,12 @@ updateComponentVersionHistory(componentChangelogEntries, version);
...
 if (isNotPatchVersion(version)) {
   // Clean previous front matter status for all components
-  cleanComponentFrontMatter(allComponentsPath, version);
+  cleanComponentFrontMatter(allComponentPaths, version);
 }
 
 // Update front matter for each updated component
 updateComponentFrontMatter(componentChangelogEntries, version);
+
+// // Clean the changelog entries for components
```
didoo:
```suggestion
// Clean the changelog entries for components
```

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -153,9 +100,9 @@ const updateComponentVersionHistory = (componentChangelogEntries, version) => {
       .map((entry) => {
         // If the component is a form primitive, we want to keep the component name in the description
         if (
-          allComponentsPath[componentName] ===
+          allComponentPaths[componentName] ===
```
didoo: [issue/blocker] maybe I'm missing something, but `allComponentPaths` is not anymore an object, but an array, so `allComponentPaths[componentName]` returns `undefined`

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -180,7 +131,7 @@ const updateComponentVersionHistory = (componentChangelogEntries, version) => {
 
 const updateComponentFrontMatter = (componentChangelogEntries, version) => {
   Object.keys(componentChangelogEntries).forEach((componentName) => {
-    const indexPath = `${allComponentsPath[componentName]}/index.md`;
+    const indexPath = `${allComponentPaths[componentName]}/index.md`;
```
didoo: [issue/blocker] same as above `allComponentPaths` is not an object now, but an array

didoo: note: I am not even sure you need `allComponentPaths` at all anymore: you have the component path directly from the regex/`START` itself, no?

didoo (changes requested):
There is an issue with `allComponentPaths` being used as an object, where now is an array (and potentially is not needed anymore). Also, a few general observations:

**Non-existing path** (eg. for a typo)
This (wrong) case is not handled (I would expect an error to be thrown, so there is visibility that something was not correct in the format used:
```
<!-- START foo/bar -->
`Card` - Added something to the component
<!-- END foo/bar -->
```

**Opening/Closing tags not matching** (eg. for a typo)
Same as above, I think this should be handled too
```
<!-- START components/card -->
`Card` - Added something to the component
<!-- END cmnponents/car -->
```
(this leads me to a question: could the regex be this `^(<!-- START \[${componentName}\])((.|\n)*?)(<!-- END -->)$` and be run against this (simplified) format?
```
<!-- START [components/tag] -->
`Tag` - Test change
<!-- END -->
```
(I've added the `[]` for better readability; it's an idea, for your consideration)

website/lib/changelog/generate-component-changelog-entries.mjs
dchyun: Thanks for catching this. This was a mistake I made when resolving the conflicts between the changes made for the layout components and this branch. I was able to remove all instances of `allComponentPaths` inside this function as you suggested.

dchyun:
@didoo I took your suggestion and updated the format slightly to no longer have the path in the "END" statement. 
```
<!-- START components/tag -->
`Tag` - Test change
<!-- END -->
```
I also added in support for notifying the user of an incorrect component path. It will output a message in the console when the script is run with the name of the incorrect path.
```
No path found for changelog entry: components/missing-path
```
> could the regex be this `^(<!-- START \[${componentName}\])((.|\n)*?)(<!-- END -->)$`
> (I've added the [] for better readability; it's an idea, for your consideration)

I tried adding this in with the regex you suggested, and ran into some issues with the regex parser no longer automatically escaping the `/` characters inside the `componentName`. I've kept it without the `[]` for now, but don't have a strong opinion and could look into it further if we want to adopt that.

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -93,85 +55,91 @@ const extractVersion = (changelogContent, version) => {
...
-    const versionHistoryPath = `${allComponentsPath[componentName]}/partials/version-history/version-history.md`;
+    const versionHistoryPath = `./docs/${componentName}/partials/version-history/version-history.md`;
     let versionHistoryContent = '';
 
     if (fs.existsSync(versionHistoryPath)) {
       versionHistoryContent = fs.readFileSync(versionHistoryPath, 'utf8');
     } else {
       fs.mkdirSync(
-        `${allComponentsPath[componentName]}/partials/version-history`,
+        `./docs/${componentName}/partials/version-history/version-history`,
```
didoo: [issue] this creates an unnecessary extra folder 
```suggestion
        `./docs/${componentName}/partials/version-history`,
```

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -93,85 +55,91 @@ const extractVersion = (changelogContent, version) => {
...
-    }
-  } else {
-    return componentName
-      .split('-')
-      .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
-      .join('');
-  }
-};
-
 const extractComponentChangelogEntries = (components, lastVersionContent) => {
```
didoo: I hope you don't mind if I suggest directly the whole implementation in code, but is much easier that trying to explain things in abstract.

I think this function can be simplified, by changing the logic: instead of looping on all the existing components, we can loop over the entries in the changelog, using the regex to extract the path (what you call here `componentName`, even if technically is the component path) and the "clean" content. 

See the code below:

```
const extractComponentChangelogEntries = (lastVersionContent) => {
  const componentChangelogEntries = {};

  // Regular expression to match and capture a block of text delimited by two HTML comments used as START/END markers:
  // <!-- START ... -->
  //   ... (any content, including newlines)
  // <!-- END -->
  //
  // We capture the following groups:
  // - $1 = The full opening comment line (e.g., <!-- START ... -->\n)
  // - $2 = The component path identifier declared in the "START" marker
  // - $3 = The content between the start and end comments (non-greedy)
  // - $4 = The full closing comment line (\n<!-- END -->)
  //
  // Flags:
  // - 'g': global search (find all matches)
  // - 'm': multiline mode (^ and $ match start/end of line)
  //
  // Explanation/demo: https://regex101.com/r/IDnaLU/1
  //
  const regex = new RegExp(
    `^(<!-- START (.*) -->\n)((.|\n)*?)(\n<!-- END -->)$`,
    'gm',
  );

  // we use `matchAll` to get an iterator of *all* regex matches in the `lastVersionContent`
  // this allows to access directly to the capture groups and extract only what we need (see below)
  const matchesIterator = lastVersionContent.matchAll(regex);
  // we then convert the iterator to an array for easier processing with array methods
  const matches = Array.from(matchesIterator);

  if (matches.length > 0) {
    matches.forEach((match) => {
      const [_fullString, _startMarker, componentPath, cleanUpdate, _endMarker] = match;
      // we want to make sure there are no typos or errors in the path declared in the `START` marker
      if (fs.existsSync(`./docs/${componentPath}`)) {
        (componentChangelogEntries[componentPath] ??= []).push(cleanUpdate);
      } else {
        // TODO discuss with Dylan if throwing causes problems with the CI flow
        // throw new Error(`No path found for changelog entry: ${componentPath}`);
      }
    });
  } else {
    console.warn(`No changelog entries found in last version content: ${lastVersionContent}`);
  }

  return componentChangelogEntries;
};
```

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -93,85 +55,91 @@ const extractVersion = (changelogContent, version) => {
...
   const componentChangelogEntries = {};
 
-  Object.keys(components).forEach((componentName) => {
-    const formattedComponentName = convertComponentNameFormat(componentName);
-    const regex = new RegExp(`^\`${formattedComponentName}\` - .*$`, 'gm');
+  components.forEach((componentName) => {
```
didoo: consider renaming this from `componentName` to `componentPath` (more correct, _I think_)

didoo (commented):
Left a couple of suggestions and a possible improvement/simplification

website/lib/changelog/generate-component-changelog-entries.mjs
dchyun: Yeah the old name was `componentName` but it makes more sense to call it a path now with the changes.

website/lib/changelog/generate-component-changelog-entries.mjs
dchyun: Thanks for the suggestion and the detailed demo code. It worked great in implementing your suggestion! This is a much simpler approach. 

For incorrect names throwing an error should be fine with the CI. This only get run manually by the release engineer once the release is ready to go out. The one situation that could happen is an error thrown for a new component if the docs for that component aren't in main yet. But merging docs should happen before running this so it's a good check to make sure that's been done anyway.

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -262,25 +246,15 @@ const version = readVersionFromPackageJson(
...
-const copyComponentPaths = getComponentPaths('./docs/components/copy');
-const formComponentPaths = getComponentPaths('./docs/components/form');
-const linkComponentPaths = getComponentPaths('./docs/components/link');
-const stepperComponentPaths = getComponentPaths('./docs/components/stepper');
-const tableComponentPaths = getComponentPaths('./docs/components/table');
 const layoutComponentPaths = getComponentPaths('./docs/layouts');
 const overrideComponentPaths = getComponentPaths('./docs/overrides');
 const utilityComponentPaths = getComponentPaths('./docs/utilities');
-const allComponentsPath = {
+const allComponentPaths = [
```
didoo: [idea] potentially, all this logic could be simplified if we use a package like `glob` in which case we could find all the paths containing a `partial` folder with just this:
```
await glob('./doc/{components,layouts,overrides,utilities}/**/partials/', { onlyDirectories: true });
```
and once we have the list, we could use a regex to remove the `.doc/` and `/partials/` parts from the path.

[for context: I haven't tried it in code, it's just in idea if you want to simplify even more the code]

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -93,94 +55,99 @@ const extractVersion = (changelogContent, version) => {
...
-    }
-  } else {
-    return componentName
-      .split('-')
-      .map((word) => word.charAt(0).toUpperCase() + word.slice(1))
-      .join('');
-  }
-};
-
 const extractComponentChangelogEntries = (components, lastVersionContent) => {
```
didoo: [todo] you don't need anymore to pass the `components` variable here

didoo (approved):
Left a couple of non-blocking comments

website/lib/changelog/generate-component-changelog-entries.mjs
dchyun: Great idea. I added `glob` and cut down on the logic for the `getComponentPaths` function.

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -17,64 +18,18 @@ const readVersionFromPackageJson = (filePath) => {
...
-            components[folder.name] = componentPath;
-          }
-        }
-      }
-    });
-  } catch (err) {
-    console.error(`Error reading components from ${baseDir}:`, err);
-  }
-  return components;
+const getComponentPaths = async () => {
```
didoo, line 21: nice!

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -215,8 +175,8 @@ const updateComponentFrontMatter = (componentChangelogEntries, version) => {
 };
 
 const cleanComponentFrontMatter = (components, version) => {
```
didoo:
```suggestion
const cleanComponentFrontMatter = (allComponentPaths, version) => {
```

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -215,8 +175,8 @@ const updateComponentFrontMatter = (componentChangelogEntries, version) => {
...
-  Object.keys(components).forEach((componentName) => {
-    const indexPath = `${components[componentName]}/index.md`;
+  Object.keys(components).forEach((componentPath) => {
+    const indexPath = `${components[componentPath]}/index.md`;
```
didoo:
```suggestion
  Object.keys(allComponentPaths).forEach((componentPath) => {
    const indexPath = `${componentPath}/index.md`;
```

didoo (commented):
Left a couple of suggestion (one of them I am not sure how the code would work otherwise 🤔, or maybe I'm missing something or I'm confused)

website/lib/changelog/generate-component-changelog-entries.mjs
dchyun: Renamed it to `componentPaths` to keep the function parameter name more generic and different from the `allComponentPaths` variable defined below.

website/lib/changelog/generate-component-changelog-entries.mjs
dchyun: good catch, the issue is I'm still using `Object.keys` even though I changed `allComponentPaths` to an array. I updated this to reflect that.

didoo (approved):

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -214,9 +174,9 @@ const updateComponentFrontMatter = (componentChangelogEntries, version) => {
   });
 };
 
-const cleanComponentFrontMatter = (components, version) => {
-  Object.keys(components).forEach((componentName) => {
-    const indexPath = `${components[componentName]}/index.md`;
+const cleanComponentFrontMatter = (componentPaths, version) => {
+  componentPaths.forEach((componentPath) => {
+    const indexPath = `${componentPath}/index.md`;
```
didoo: [todo/blocker] for the `cleanComponentFrontMatter` function to work, the path needs to be updated or the `if (fs.existsSync(indexPath)) {` will fail

```suggestion
    const indexPath = `./docs/${componentPath}/index.md`;
```

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -93,94 +48,99 @@ const extractVersion = (changelogContent, version) => {
...
       fs.writeFileSync(versionHistoryPath, newHeading, 'utf8');
     }
   });
 };
 
 const updateComponentFrontMatter = (componentChangelogEntries, version) => {
-  Object.keys(componentChangelogEntries).forEach((componentName) => {
-    const indexPath = `${allComponentsPath[componentName]}/index.md`;
+  Object.keys(componentChangelogEntries).forEach((componentPath) => {
+    const indexPath = `${componentPath}/index.md`;
```
didoo: [todo/blocker] for the `updateComponentFrontMatter` function to work, the path needs to be updated or the `if (fs.existsSync(indexPath)) {` will fail

```suggestion
    const indexPath = `./docs/${componentPath}/index.md`;
```

didoo (approved):

PR #2961, opened 2025-06-19
[🧪] Testing possible fix for `Modal` issues

didoo:
### :pushpin: Summary

This is a branch that is used to test a possible fix for the issue of a `Modal` being closed not by using the `F.close` method, but by simply removing it from the DOM. 

With the current logic, there's no cleanup (the `willDestroyNode` is invoked but the cleanup is done in the `onDismiss` function).

This PR does multiple things:

- adds detailed logging to the `Modal` backing class, so we can see the order or execution and which functions are called (and their nested invocation)
- adds a couple of examples to the showcase page, to demonstrate the bug
- does a large refactoring, so the naming is more meaningful and the order of executions easier to follow
- removes the `onDeactivate=this.onManualDismiss` from the `focus-trap` options, because it was invoking the "manual" dismiss even with a non-manual closure, or otherwise it caused the onDismiss to be called twice (see below)

The idea is to:
- see that all the HDS tests are passing (we will need to add more for this fix, to test the case that we didn't consider)
- smoke test the solution/fix across different products/consumers
- if everything works, prepare a proper PR and from there on follow the standard process (testing, review, approval, release)

Difference in Atlas, between when the modal is closed via dismiss vs submit
screenshot_5104

Video of the fix working:
https://github.com/user-attachments/assets/f81e2e6f-7806-4731-b518-91e29eaabcb8

Reason why the `onDeactivate=this.onManualDismiss` has been removed from the `focus-trap` options (see red areas):
screenshot_5103

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-4975

***

<details>

</details>

### DIFF
--- a/packages/components/src/components/hds/modal/index.hbs
+++ b/packages/components/src/components/hds/modal/index.hbs
@@ -9,7 +9,7 @@
   {{did-insert this.didInsert}}
   {{will-destroy this.willDestroyNode}}
   {{! @glint-expect-error - https://github.com/josemarluedke/ember-focus-trap/issues/86 }}
-  {{focus-trap isActive=this._isOpen focusTrapOptions=(hash onDeactivate=this.onManualDismiss)}}
+  {{focus-trap isActive=this._isOpen}}
 >
   <:header>
     {{yield


packages/components/src/components/hds/modal/index.ts
```diff
@@ -63,15 +61,7 @@ export default class HdsModal extends Component<HdsModalSignature> {
   private _element!: HTMLDialogElement;
   private _body!: HTMLElement;
   private _bodyInitialOverflowValue = '';
-  private _clickHandler!: (event: MouseEvent) => void;
-
-  constructor(owner: Owner, args: HdsModalSignature['Args']) {
-    super(owner, args);
-
-    registerDestructor(this, (): void => {
-      document.removeEventListener('click', this._clickHandler, true);
```
meirish, line 72: I don't follow all of the changes here, but why not do the style cleanup here and remove the use of `will-destroy` render modifier?
(the destructor will run when the component is cleaned up / unrendered)

didoo:
This can be closed now. We may use it as reference for the refactoring of the `Modal/Flyout` logic.

PR #2962, opened 2025-06-19
Fix Modal destructor when modal is removed while open

alex-ju:
### :pushpin: Summary

Fix destructor when <dialog> is open and the component is destroyed.

There are instances where the Modal component is destroyed without being properly closed (using the `close()` function via top right button, click outside or the `.close` function available in footer).

To avoid such instances we emulate the `.close` event if the `<dialog>` is removed from the DOM while open, so it behaves similar to a natural dismissal.

### :link: External links

Jira ticket: [HDS-4975](https://hashicorp.atlassian.net/browse/HDS-4975)

***

### :eyes: Component checklist

- [ ] Percy was checked for any visual regression
- [ ] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

<details>

<summary>:clipboard: PCI review checklist</summary>

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.
- [ ] If applicable, I've worked with GRC to document the impact of any changes to security controls.
  Examples of changes to controls include access controls, encryption, logging, etc.
- [ ] If applicable, I've worked with GRC to ensure compliance due to a significant change to the in-scope PCI environment.
  Examples include changes to operating systems, ports, protocols, services, cryptography-related components, PII processing code, etc.

</details>

PR #2964, opened 2025-06-20
`Modal` - Revert fixes #2846 and #2902

didoo:
### :pushpin: Summary

This PR reverts the fixes made in #2846 and #2902 that introduced an unexpected side-effect.

When the Modal was not closed via the "manual" dismiss (click outside, click dismiss button, esc key) or the `F.Close` yielded method, the code that cleans up the `overflow: hidden` from the `<body>` element is run inside the `onDismiss` function. When the modal is simply destroyed and removed from the DOM (eg. on submit) the `onDismiss` is not run, just the `willDestroyNode` method, which leaves the overflow applied to the body, making the page unscrollable. 

We've decided that, even if we have other fixes to the bug (see #2961 and #2962), this is the safest and quickest route. After the release, we will discuss and decide how to reintroduce the original/intended changes (for better accessibility) in a safe and controlled way (that would likely include a refactoring of the existing code to make it easier to follow the logic flow and debug if necessary)

You can see examples of this bug in the linked Jira ticket.

In this PR I have:
- reverted #2846 
- reverted #2902 (but left the integration test that was added)

The updates to the showcase and the integration tests are done in a different PR (https://github.com/hashicorp/design-system/pull/2966) to keep this one very simple and straightforward.

Modal:
https://github.com/user-attachments/assets/4a681e64-9f50-4a59-9001-b7cab3525f4c

Flyout:
https://github.com/user-attachments/assets/3c34dfe5-f503-4bf2-9cd6-694ccc06e692

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-4975

Related PRs:
- https://github.com/hashicorp/design-system/pull/2961
- https://github.com/hashicorp/design-system/pull/2965
- https://github.com/hashicorp/design-system/pull/2965

***

<details>

</details>

### DIFF
--- a/.changeset/tidy-tigers-care.md
+++ b/.changeset/tidy-tigers-care.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Modal` - Reverted changes introduced in [#2846](https://github.com/hashicorp/design-system/pull/2846) and [#2902](https://github.com/hashicorp/design-system/pull/2902) related to the click behaviour outside the modal when dismissing is disabled


alex-ju:
> the original/intended changes (for better accessibility)

looking at #2846, it doesn't just improve accessibility, it fixes a functional issue with click outside the Modal not working properly when `isDismissDisabled` is set to `true` 🤔 

so by simply reverting it we're bringing back a functional issue – @shleewhite may be able to keep me honest here – we could make the call of moving forward, but want to make sure it's a conscious decision

didoo:
> > the original/intended changes (for better accessibility)
> 
> looking at #2846, it doesn't just improve accessibility, it fixes a functional issue with click outside the Modal not working properly when `isDismissDisabled` is set to `true` 🤔
> 
> so by simply reverting it we're bringing back a functional issue – @shleewhite may be able to keep me honest here – we could make the call of moving forward, but want to make sure it's a conscious decision

Per discussion amongst us (@alex-ju @shleewhite @didoo): we’ve decided that it would be safer/quicker to use the PR where we revert the changes introduced with the two previous fixes, because 1) we know that the code worked before and 2) the only issue, the original one that was being fixed, was reported by us and was a very specific edge case (click outside when the modal is not dismissible - see [HDS-3972](https://hashicorp.atlassian.net/browse/HDS-3972))

We’ve also decided to split in two the PRs, so we have a very simple PR for reverting things, and one for the showcase updates and the integration tests refactoring (see https://github.com/hashicorp/design-system/pull/2966).

[HDS-3972]: https://hashicorp.atlassian.net/browse/HDS-3972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

dchyun:
This doesn't need to be a blocker for this PR, but for the examples deactivating on destroy and deactivating on form submit the focus is not getting reset back to the button after the modal is removed. In the on close example the focus is being reset.

This is because `willDestroyNode` is removing the `close` event listener before it gets called in the `onDismiss` with `this._element.close()`. 

Using the `returnFocusTo` argument makes the focus return correctly. It may be the case that if consumers use the modal in this way then they have to use `returnFocusTo`. 

Overall the fix looks to be working great though. Didn't see any issues with scrolling being disabled.

alex-ju:
> This doesn't need to be a blocker for this PR, but for the examples deactivating on destroy and deactivating on form submit the focus is not getting reset back to the button after the modal is removed. In the on close example the focus is being reset.
> 
> This is because `willDestroyNode` is removing the `close` event listener before it gets called in the `onDismiss` with `this._element.close()`.
> 
> Using the `returnFocusTo` argument makes the focus return correctly. It may be the case that if consumers use the modal in this way then they have to use `returnFocusTo`.
> 
> Overall the fix looks to be working great though. Didn't see any issues with scrolling being disabled.

I noticed that too @dchyun. The intention here is to provide a clean revert addressing the issue at hand, rather than a comprehensive fix – for now – and review the component for improvements in the coming days. We have a few potential ways to improve the component and clarify some of the actions.

dchyun (approved):

shleewhite (approved):

PR #2898, opened 2025-05-22
Form layout—Components (HDS-4885)

KristinLBradley:
### :pushpin: Summary

If merged, this PR adds `Form` layout components to HDS.

👉 **Showcase**: https://hds-showcase-git-kristin-hds-4885-form-layout-d00037-hashicorp.vercel.app/components/form/layout

#### Related:

* **Documentation PR**: https://github.com/hashicorp/design-system/pull/2947 (DRAFT)

#### Changes:

* Reorganized Showcase links on home page to create new "section" for Form-related components: https://hds-showcase-git-kristin-hds-4885-form-layout-d00037-hashicorp.vercel.app/
  - Please let me know your opinions on this change.
* Created `Form` component with `@tag` prop.
* Created `FormHeader` component with `@hasMaxWidth` prop
* Created `FormHeaderTitle` component with `@tag` & `@size` props
* Created `FormHeaderDescription` component.
* Created `FormSection` component with `@hasMaxWidth` & `@hasBorder` props
* Created `FormSectionHeader` component
* Created `FormSectionHeaderTitle` component with `@tag` & `@size` props
* Created `FormSectionHeaderDescription` component.
* Created `FormSectionFieldGroup` component based off `Layout::Flex` component which exposes `@direction` prop & includes responsive behavior to stack Fields at screen width below 768px
* Created `FormSeparator` component with `@hasMaxWidth` prop
* Created [Showcase page for Form layout components](https://hds-showcase-git-kristin-hds-4885-form-layout-d00037-hashicorp.vercel.app/components/form)

##### Component structure in use:

* `Form`
  * `FormHeader`
    * `FormHeaderTitle`
    * `FormHeaderDescription`
  * `FormSection`
    * `FormSectionHeader`
      * `FormSectionHeaderTitle`
      * `FormSectionHeaderDescription`
    * Form Fields such as `TextInput::Field` (pre-existing)
    * `FieldGroup` (lay out related groups of inputs in a row or stacked)
      * Form Fields such as `TextInput::Field` (pre-existing)
  * `FormSeparator` (extra visual separation between Sections)
  * `ButtonSet` (pre-existing, can be wrapped in Section to control max-width)

##### How widths work:

For Form Sections & similar:

* If `@isFullWidth="false"` (the default) `FormSection` will expand to a max of 700px
* If `@isFullWidth="true"` `FormSection` & similar will expand to 100% of available width
[...]

PR #2969, opened 2025-06-23
[Website] Support form layout components in changelog script

dchyun:
**Note:** PR has been updated to be based off of main. This means it can not be tested against the new form layout folders as those are created in a separate docs PR. It was tested off of them previously and approved. 

### :pushpin: Summary

This PR update the `generate-component-changelog-entries-script` to support the two new top level form layout components `Form` and `Form::Separator`.

The script needed to be updated as the component documentation lives under `components/form/layout`, but the script is based around a components documentation path matching its name. Ex: `Tabs` maps to `components/tabs`. 

With the updated script now changelog entries for both `Form` and `Form::Separator` will get added to the `components/form/layout` docs page. 

### Testing

There is a temporary commit included in this PR that allows testing of the script locally. To test this change do the following:

1. Cd into `./website`
2. Run `pnpm generate-component-changelog-entries`
3. Do not commit any changes made

### :link: External links

Figma file: [if it applies]

***

<details>

</details>

### DIFF
--- a/.changeset/afraid-hornets-lie.md
+++ b/.changeset/afraid-hornets-lie.md
@@ -3,5 +3,3 @@
 ---
 
 `Form` - Added `Form` component and related sub-components for form layout
-
-`Form::Separator` - Added `Form::Separator` component


didoo (dismissed):
Just glanced at the code, let's merge this to unblock @KristinLBradley

dchyun:
> Just glanced at the code, let's merge this to unblock @KristinLBradley

@didoo this is not a blocker for @KristinLBradley as it only affect the script that will get run when the release goes out. She can still merge the component or docs to main before this goes in. This just needs to go in before the release.

That said, will get this cleaned up and ready to merge today.

website/lib/changelog/generate-component-changelog-entries.mjs
```diff
@@ -44,6 +44,9 @@ const getComponentPaths = (baseDir) => {
               primitiveNames.forEach((componentName) => {
                 components[`form-${componentName}`] = componentPath;
               });
+            } else if (folder.name === 'layout') {
+              components[`form`] = componentPath;
+              components[`form-separator`] = componentPath;
```
didoo: [question] why only the `form-separator` is indicated here? should also all the other new `form-[something]` be added?

dchyun: In [discussing this](https://github.com/hashicorp/design-system/pull/2898/commits/3d7327082f0fcba0a1dbecf47f338da9d649afa9#r2159651651) with @KristinLBradley for the original changeset, I mentioned that usually for changelog entries we only mention the main component names in the changelog entries. Ex: If we made a change to the `AccordionItem` the entry would still lead with `Accordion`. 

For the form layout components from discussion with Kristin and looking at the example only the `Form` and `Form::Separator` are those top level components intended to be used on their own. The `FormTitle` or `FormDescription` are not used on their own, but inside the `Form`.  So if a future update was made to the `FormTitle` the entry would be something like 

> `Form` - Updated `FormTitle`

This is my long-winded way of saying that I was just trying to follow how we mention components in the changeset today. Happy to accommodate the other form sub components here if we want. 

[Final layout changset](https://github.com/hashicorp/design-system/blob/4e0a6d32c3a200005e316eba387ad1ea2bb0774e/.changeset/afraid-hornets-lie.md)

didoo: > only the Form and Form::Separator are those top level components intended to be used on their own

@KristinLBradley afaik the `Form::Separator` is meant to be used inside the `Form`, right? so @dchyun should actually remove it then from the code as distinct entity.

KristinLBradley: Yes, sorry, I had a comment/question related to that but lost it and forgot to revisit the question.

The `Form::Separator` is intended only to be used within the `Form` so the separate entry for it should be deleted. @dchyun

dchyun: Thanks for letting me know @KristinLBradley. I updated the script to remove the form separator reference. Also I updated your changeset to remove the mention of the separator.

.changeset/afraid-hornets-lie.md
```diff
@@ -3,5 +3,3 @@
 ---
 
 `Form` - Added `Form` component and related sub-components for form layout
-
-`Form::Separator` - Added `Form::Separator` component
```
didoo, line 7: Actually this should remain, as test case: in the future we may want to change just one of the sub-components for layout (I would also add a row with `Form::HelperText` to make sure it's associated with the form base elements, and `Form::TextInput::Base` to make sure it's associated with the form text input

dchyun: If we make a change to a subcomponent of `Form` that could just be mentioned in this way
```
`Form` - Updated `Form::Separator` 
```

For any change we want picked up by the script and added to the component version history page I have to account for it in the script. That's why I originally had the separator mentioned in it. If we want all of those names accounted for the changeset should be updated to include them, and I will update the script to account for them.

dchyun: Also one note, this is not applicable if we adopt the new naming structure of #2953. With the new structure you could do something like 
```
<!-- START form/layouts -->
`Form::Separator` - Change description
<!-- END form/layouts -->
```

didoo: Yeah, it depends if we want to have this as a temporary fix, just for Kristin's PR (which is likely) so I'm maybe just overthinking.

didoo (approved):

PR #2979, opened 2025-06-27
[HDS] Match minimized-sidebar width from enterprise nav changes

philrenaud:
### :pushpin: Summary
A [recent change to HCP / Cloud-UI](https://github.com/hashicorp/cloud-ui/pull/12433) introduced a new sidebar nav style, the minimized width of which is shorter than it previously was.

This PR updates the minimized sidebar width, which if unchanged causes the following misalignment on pages with sticky footers, among potentially other places:

image

<details>

</details>

### DIFF
--- a/packages/components/src/styles/components/side-nav/vars.scss
+++ b/packages/components/src/styles/components/side-nav/vars.scss
@@ -19,7 +19,7 @@
   // breakpoint
   --hds-app-desktop-breakpoint: #{map.get($hds-breakpoints, "lg")}; // this is used only to read its value via JS and set the `hds-side-nav--is-desktop` class
   // widths
-  --hds-app-sidenav-width-minimized: 48px;
+  --hds-app-sidenav-width-minimized: 16px;
   --hds-app-sidenav-width-expanded: 280px;
   --hds-app-sidenav-width-fixed: var(--hds-app-sidenav-width-expanded);
   // animation


packages/components/src/styles/components/side-nav/vars.scss
```diff
@@ -19,7 +19,7 @@
   // breakpoint
   --hds-app-desktop-breakpoint: #{map.get($hds-breakpoints, "lg")}; // this is used only to read its value via JS and set the `hds-side-nav--is-desktop` class
   // widths
-  --hds-app-sidenav-width-minimized: 48px;
+  --hds-app-sidenav-width-minimized: 16px;
```
KristinLBradley, line 22: It may take some time for all product teams to update from the old `SideNav` to the new `AppSideNav`. Changing this value here will mess up the layout in the old SideNav as it will be too narrow for the HomeLink logo to fit.

Would it be possible to update the related value for your sticky footer and other components instead?

didoo: There are two CSS variables in play here (with very similar names, it took me a while to notice the difference):

https://github.com/hashicorp/design-system/blob/main/packages/components/src/styles/components/side-nav/vars.scss#L22
```
--hds-app-sidenav-width-minimized: 48px;
```

https://github.com/hashicorp/design-system/blob/main/packages/components/src/styles/components/app-side-nav/vars.scss#L16
```
--hds-app-side-nav-width-minimized: 16px;
```

The first one (the "old" one) is used also in the Flyout: https://github.com/hashicorp/design-system/blob/main/packages/components/src/styles/components/flyout.scss#L10

So I imagine the "solution" would not be to override the old one, but use the new one for the sticky footer (@philrenaud do you remember we were discussing this, for the sticky footer implementation in HDS+? and I think we came up also with a solution to this, to make it agnostic; but that's another story).

didoo:
@philrenaud is this still relevant? otherwise we are going to close it

philrenaud:
I still see this in the latest: 
image

I can, per @KristinLBradley 's suggestion, simply add an `.infragraph` wrapper around my own element, but it seems like we should consider that it's a good time to make the change more widely.

philrenaud:
Will implement the local fix as suggested in https://github.com/hashicorp/design-system/pull/2979#discussion_r2174747195 thanks

PR #2991, opened 2025-07-04
`KeyValueInput` - add support for `identityKey` in the `#each` loop

didoo:
### :pushpin: Summary

While looking at examples of existing implementation of Key/Value components in consumer's codebases, I have noticed that in a couple of cases the `{{#each}}` loop had an extra `key`, while other cases didn't have it.

With `key`:

- https://github.com/hashicorp/cloud-ui/blob/main/engines/vault/addon/components/form/inputs/key-value-pairs/index.hbs#L3
- https://github.com/hashicorp/cloud-ui/blob/main/engines/waypoint/addon/components/key-value-input.hbs#L14

Without `key`:

- https://github.com/hashicorp/atlas/blob/main/frontend/atlas/app/components/v2/key-value-list-input/key-value-input.hbs#L15
- https://github.com/hashicorp/vault/blob/main/ui/lib/core/addon/components/kv-object-editor.hbs#L20
- https://github.com/hashicorp/boundary-ui/blob/main/ui/admin/app/components/form/field/list-wrapper/key-value/index.hbs#L29

For context, we already have something similar in the `Table` and `AdvancedTable` components.

With this PR I want to collect feedback if it's worth to add it or not.

In this PR I have:
- added support for `identityKey` in the `#each` loop of the `KeyValueInput` template (same implementation as the existing one for `Table` and `AdvancedTable` loops.

**TODO** (if we decide to proceed)
- add changelog
- add ticket and PR for the documentation update
- create Jira ticket specific for this task (child of [HDS-5071](https://hashicorp.atlassian.net/browse/HDS-5071))

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-5071 (overarching task)

***

<details>

</details>

### DIFF
--- a/packages/components/src/components/hds/form/key-value-inputs/index.hbs
+++ b/packages/components/src/components/hds/form/key-value-inputs/index.hbs
@@ -66,7 +66,7 @@
     </div>
   {{/if}}
 
-  {{#each @data as |item index|}}
+  {{#each @data key=this.identityKey as |item index|}}
     <div class="hds-form-key-value-inputs__row {{if (eq index 0) 'hds-form-key-value-inputs__row--first'}}">
       {{yield
         (hash
--- a/packages/components/src/components/hds/form/key-value-inputs/index.ts
+++ b/packages/components/src/components/hds/form/key-value-inputs/index.ts
@@ -43,6 +43,7 @@ export interface HdsFormKeyValueInputsSignature {
     extraAriaDescribedBy?: string;
     isOptional?: boolean;
     isRequired?: boolean;
+    identityKey?: string;
   };
   Blocks: {
     header?: [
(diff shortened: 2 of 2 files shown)


didoo:
@RobbieTheWagner @dstaley @aklkv I don't have enough knowledge of the internals of Ember to understand if it's worth giving consumers an option to provide an "identity key" for the `{{#each}}` loop of this component (I've seen different implementations, some with the `key` and some without it. Can you share your thought? That would be really appreciated 🙏

cc @shleewhite @dchyun @zamoore

meirish:
@didoo the official docs on this (though kinda buried), are pretty good: https://api.emberjs.com/ember/release/classes/Ember.Templates.helpers/methods/each?anchor=each 

It serves the same purpose as [in React](https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key), though the specifics are a bit different. Mostly it's not needed, but in cases where the data may be the same and the default `@identity` does change (for example, when dealing with objects or re-setting a whole array), specifying the key allows devs to help the framework do less work.

RobbieTheWagner:
@didoo I echo what @meirish said. I think allowing folks to specify keys on essentially all `each` would be a good idea, but I do not know if this component specifically will benefit from it. If the lists can be long, it is possible it would save Ember some work.

Thanks for sharing you point of view and experience 🙏

didoo:
> If the lists can be long, it is possible it would save Ember some work.

It's unlikely that will be long.

@meirish @RobbieTheWagner from what I read, this would not add much to the component. For now I am closing this PR, if in the future we will see the need for this, we can always open a new one.

PR #2993, opened 2025-07-07
`Web docs` - Remove mention of filled badge in tables

majedelass:
### :pushpin: Summary

If merged, this PR will remove the reference of a fill badge in a table. 

We are removing this reference because of A11Y requirements of color contrast. Not all filled in badges will work with striping, while outlining will.

### :link: External links

Jira ticket: [HDS-5075](https://hashicorp.atlassian.net/browse/HDS-5075)

***

[HDS-5075]: https://hashicorp.atlassian.net/browse/HDS-5075?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/docs/components/badge/partials/guidelines/guidelines.md
+++ b/website/docs/components/badge/partials/guidelines/guidelines.md
@@ -61,8 +61,6 @@ Use filled Badges when displaying many on a single page or for subtle callouts.
 <Hds::Badge @color="critical" @text="Critical filled" />
 
 For example:
-
-- when listing statuses in a table
 - for successful or passive actions
 
 ### Inverted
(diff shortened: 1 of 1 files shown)


jorytindall:
For consistency I might consider actually moving this point below the Type / Outlined section, since there are example usages for the other types. Since they are all now single bullet points, I might also consider removing the bullets and just formatting these as sentences

majedelass:
> For consistency I might consider actually moving this point below the Type / Outlined section, since there are example usages for the other types. Since they are all now single bullet points, I might also consider removing the bullets and just formatting these as sentences

This was the original intent until I noticed the below list only having one item so I was ok to keep it as is.

zamoore (approved):

PR #2996, opened 2025-07-07
CardContainer: fix component name in template registry

shleewhite:
### :pushpin: Summary

If merged, this PR would fix the name for the Card Container component in the template registry. 

I noticed that there is a discrepancy where in template registry, we export `hds/card` and `Hds::Card`, but in [our docs](https://helios.hashicorp.design/components/card?tab=code) and in [the showcase template](https://github.com/hashicorp/design-system/blob/c605a1805d64943a6ce916611fbce3f2b740af32/showcase/app/templates/components/card.hbs#L19), we exclusively refer to it as `Hds::Card::Container`.

Changing this would be **a breaking change** because it would change the name of a component externally. Options for moving forward:
1. Wait to fix this until the 5.0 release and add glint ignores to the Card template file to supress current errors. 
2. Just rename it `Hds::Card` and update the showcase/website to match
3. Export it as both names as a fix for now, circle back to remove the incorrect name in the 5.0 release _(Lee's preference)_

I found this while I was upgrading the routes files to typescript ([example test failure](https://github.com/hashicorp/design-system/actions/runs/16124069042/job/45496690922))
Screenshot 2025-07-07 at 2 17 33 PM

***

<details>

</details>

### DIFF
--- a/packages/components/src/template-registry.ts
+++ b/packages/components/src/template-registry.ts
@@ -426,8 +426,8 @@ export default interface HdsComponentsRegistry {
   'hds/button-set': typeof HdsButtonSetComponent;
 
   // Card
-  'Hds::Card': typeof HdsCardContainerComponent;
-  'hds/card': typeof HdsCardContainerComponent;
+  'Hds::Card::Container': typeof HdsCardContainerComponent;
+  'hds/card/container': typeof HdsCardContainerComponent;
 
   // Code Block
   'Hds::CodeBlock': typeof HdsCodeBlockComponent;
--- a/packages/components/src/template-registry.ts
+++ b/packages/components/src/template-registry.ts
@@ -426,6 +426,9 @@ export default interface HdsComponentsRegistry {
   'hds/button-set': typeof HdsButtonSetComponent;
 
   // Card
+  // NOTE: this will be removed in 5.0.0, use 'Hds::Card::Container' instead
+  'Hds::Card': typeof HdsCardContainerComponent;
+  'hds/card': typeof HdsCardContainerComponent;
   'Hds::Card::Container': typeof HdsCardContainerComponent;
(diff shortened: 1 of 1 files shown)


packages/components/src/template-registry.ts
```diff
@@ -426,8 +426,11 @@ export default interface HdsComponentsRegistry {
   'hds/button-set': typeof HdsButtonSetComponent;
 
   // Card
+  // NOTE: this will be removed in 5.0.0, use 'Hds::Card::Container' instead
```
alex-ju:
```suggestion
  // NOTE: `Hds::Card` and `hds/card` will be removed in 5.0.0, use 'Hds::Card::Container' instead
```
could you add a quick task under [the major release epic](https://hashicorp.atlassian.net/browse/HDS-4670?atlOrigin=eyJpIjoiN2QwMGUxY2QyN2QwNDAwNzlmMDYwNjMyNDJhMGFhNGIiLCJwIjoiaiJ9)?

aklkv:
Just FYI, no one uses hds template registry aside from showcase

shleewhite:
> Just FYI, no one uses hds template registry aside from showcase

@aklkv ahh! thats good to know. To clarify, the template-registry is just for internal use and then `components.ts` is to export the components/types externally?

If we can just fix this immediately that is preferable.

KristinLBradley:
Option 3 seems like a prudent choice to me.

"3. Export it as both names as a fix for now, circle back to remove the incorrect name in the 5.0 release"

packages/components/src/template-registry.ts
shleewhite: @alex-ju I added the task!

alex-ju (approved):

PR #2997, opened 2025-07-07
Zamoore/advanced table/column reordering 2

zamoore:
In-progress; Do not review/merge.

### :pushpin: Summary

If merged, this PR will add support for column reordering to the AdvancedTable component.

### :link: External links

Jira ticket: [HDS-4588](https://hashicorp.atlassian.net/browse/HDS-4588)
Figma file: [if it applies]

***

<details>

</details>

[HDS-4588]: https://hashicorp.atlassian.net/browse/HDS-4588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -29,6 +29,7 @@ import type {
   HdsAdvancedTableVerticalAlignment,
   HdsAdvancedTableModel,
   HdsAdvancedTableExpandState,
+  HdsAdvancedTableColumnReorderCallback,
 } from './types.ts';
 import type HdsAdvancedTableColumnType from './models/column.ts';
 import type { HdsFormCheckboxBaseSignature } from '../form/checkbox/base.ts';
@@ -115,6 +116,7 @@ export interface HdsAdvancedTableSignature {
     align?: HdsAdvancedTableHorizontalAlignment;
     caption?: string;
     columns: HdsAdvancedTableColumn[];
+    columnOrder?: string[];
     density?: HdsAdvancedTableDensities;
     identityKey?: string;
     isSelectable?: boolean;
@@ -126,11 +128,13 @@ export interface HdsAdvancedTableSignature {
     sortedMessageText?: string;
     sortOrder?: HdsAdvancedTableThSortOrder;
     valign?: HdsAdvancedTableVerticalAlignment;
+    hasReorderableColumns?: boolean;
     hasResizableColumns?: boolean;
     hasStickyHeader?: boolean;
     hasStickyFirstColumn?: boolean;
     childrenKey?: string;
     maxHeight?: string;
+    onColumnReorder?: HdsAdvancedTableColumnReorderCallback;
     onColumnResize?: (columnKey: string, newWidth?: string) => void;
     onSelectionChange?: (
       selection: HdsAdvancedTableOnSelectionChangeSignature
@@ -205,6 +209,7 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     const {
       model,
       columns,
+      columnOrder,
       childrenKey,
       hasResizableColumns,
       sortBy,
@@ -216,6 +221,7 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     this._tableModel = new HdsAdvancedTableTableModel({
       model,
       columns,
+      columnOrder,
       childrenKey,
       hasResizableColumns,
       sortBy,
--- a/packages/components/src/components/hds/advanced-table/models/column.ts
+++ b/packages/components/src/components/hds/advanced-table/models/column.ts
@@ -8,6 +8,7 @@ import { action } from '@ember/object';
 
 import type HdsAdvancedTableModel from './table.ts';
 import type {
+  HdsAdvancedTableCell,
   HdsAdvancedTableHorizontalAlignment,
   HdsAdvancedTableColumn as HdsAdvancedTableColumnType,
 } from '../types';
@@ -31,7 +32,6 @@ export default class HdsAdvancedTableColumn {
   @tracked label: string = '';
   @tracked align?: HdsAdvancedTableHorizontalAlignment = 'left';
   @tracked isExpandable?: boolean = false;
-  @tracked isReorderable?: boolean = false;
   @tracked isSortable?: boolean = false;
   @tracked isVisuallyHidden?: boolean = false;
   @tracked key?: string = undefined;
@@ -42,9 +42,18 @@ export default class HdsAdvancedTableColumn {
   @tracked originalWidth?: string = undefined; // used to restore the width when resetting
   @tracked imposedWidthDelta: number = 0; // used to track the width change imposed by the previous column
 
+  @tracked isBeingDragged: boolean = false;
   @tracked sortingFunction?: (a: unknown, b: unknown) => number = undefined;
 
   table: HdsAdvancedTableModel;
+
+  get cells(): HdsAdvancedTableCell[] {
+    return this.table.flattenedVisibleRows.map((row) => {
+      const cell = row.cells.find((cell) => cell.columnKey === this.key);
+
+      return cell!;
+    });
+  }
 
   get pxWidth(): number | undefined {
     if (isPxSize(this.width)) {
@@ -65,6 +74,44 @@ export default class HdsAdvancedTableColumn {
     if (isPxSize(this.maxWidth)) {
       return pxToNumber(this.maxWidth!);
     }
+  }
+
+  get index(): number {
+    const { orderedColumns } = this.table;
+
+    if (orderedColumns.length === 0) {
+      return -1;
+    }
+
+    return orderedColumns.findIndex((column) => column.key === this.key);
+  }
+
+  get isFirst(): boolean {
+    return this.index === 0;
+  }
+
+  get isLast(): boolean {
+    return this.index === this.table.orderedColumns.length - 1;
+  }
+
+  get siblings(): {
+    previous?: HdsAdvancedTableColumn;
+    next?: HdsAdvancedTableColumn;
+  } {
+    const { index, table } = this;
+    const { orderedColumns } = table;
+
+    if (index === -1) {
+      return {};
+    }
+
+    return {
+      previous: index > 0 ? orderedColumns[index - 1] : undefined,
+      next:
+        index < orderedColumns.length - 1
+          ? orderedColumns[index + 1]
+          : undefined,
+    };
   }
 
   constructor(args: {
--- a/packages/components/src/components/hds/advanced-table/models/row.ts
+++ b/packages/components/src/components/hds/advanced-table/models/row.ts
@@ -7,13 +7,14 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { guidFor } from '@ember/object/internals';
 
-import type { HdsAdvancedTableColumn } from '../types';
+import type { HdsAdvancedTableColumn, HdsAdvancedTableCell } from '../types';
 
 interface HdsAdvancedTableRowArgs {
   [key: string]: unknown;
   columns: HdsAdvancedTableColumn[];
   id?: string;
   childrenKey?: string;
+  columnOrder?: string[];
 }
 
 export default class HdsAdvancedTableRow {
@@ -23,6 +24,8 @@ export default class HdsAdvancedTableRow {
   [key: string]: unknown;
 
   @tracked isOpen: boolean = false;
+  @tracked cells: HdsAdvancedTableCell[] = [];
+  @tracked columnOrder: string[] = [];
 
   children: HdsAdvancedTableRow[] = [];
   childrenKey: string;
@@ -35,7 +38,33 @@ export default class HdsAdvancedTableRow {
     return this.isOpen && this.hasChildren;
   }
 
+  get orderedCells(): HdsAdvancedTableCell[] {
+    return this.columnOrder.reduce((acc, key) => {
+      const cell = this.cells.find((cell) => cell.columnKey === key);
+
+      if (cell) {
+        acc.push(cell);
+      }
+
+      return acc;
+    }, [] as HdsAdvancedTableCell[]);
+  }
+
   constructor(args: HdsAdvancedTableRowArgs) {
+    const { columns } = args;
+
+    this.cells = columns.map((column) => {
+      const cell = args[column.key ?? ''];
+
+      return {
+        columnKey: column.key ?? '',
+        content: cell,
+      };
+    });
+
+    this.columnOrder =
+      args.columnOrder ?? this.cells.map((cell) => cell.columnKey);
+
     // set row data
     Object.assign(this, args);
 
@@ -48,6 +77,15 @@ export default class HdsAdvancedTableRow {
         // eslint-disable-next-line @typescript-eslint/no-unsafe-argument
         (child) => new HdsAdvancedTableRow(child)
       );
+    }
+  }
+
+  @action
+  updateColumnOrder(columnOrder: string[]) {
+    this.columnOrder = columnOrder;
+
+    for (const child of this.children) {
+      child.updateColumnOrder(columnOrder);
     }
   }
 
--- a/packages/components/src/components/hds/advanced-table/models/table.ts
+++ b/packages/components/src/components/hds/advanced-table/models/table.ts
@@ -12,6 +12,8 @@ import { HdsAdvancedTableThSortOrderValues } from '../types.ts';
 import type { HdsAdvancedTableSignature } from '../index.ts';
 import type {
   HdsAdvancedTableExpandState,
+  HdsAdvancedTableCell,
+  HdsAdvancedTableColumnReorderCallback,
   HdsAdvancedTableSortingFunction,
 } from '../types';
 
@@ -19,11 +21,14 @@ type HdsAdvancedTableTableArgs = Pick<
   HdsAdvancedTableSignature['Args'],
   | 'model'
   | 'columns'
+  | 'columnOrder'
   | 'childrenKey'
   | 'hasResizableColumns'
   | 'sortBy'
   | 'sortOrder'
   | 'onSort'
+  | 'onColumnReorder'
+  | 'onColumnResize'
 >;
 
 function getVisibleRows(rows: HdsAdvancedTableRow[]): HdsAdvancedTableRow[] {
@@ -44,6 +49,8 @@ function getChildrenCount(rows: HdsAdvancedTableRow[]): number {
 
 export default class HdsAdvancedTableTableModel {
   @tracked columns: HdsAdvancedTableColumn[] = [];
+  @tracked columnOrder: string[] = [];
+  @tracked reorderDraggedColumn: HdsAdvancedTableColumn | null = null;
   @tracked rows: HdsAdvancedTableRow[] = [];
   @tracked sortBy: HdsAdvancedTableTableArgs['sortBy'] = undefined;
   @tracked sortOrder: HdsAdvancedTableTableArgs['sortOrder'] =
@@ -51,16 +58,19 @@ export default class HdsAdvancedTableTableModel {
 
   childrenKey?: HdsAdvancedTableTableArgs['childrenKey'];
   hasResizableColumns?: HdsAdvancedTableTableArgs['hasResizableColumns'];
+  onColumnReorder?: HdsAdvancedTableColumnReorderCallback;
   onSort?: HdsAdvancedTableSignature['Args']['onSort'];
 
   constructor(args: HdsAdvancedTableTableArgs) {
     const {
       model,
       columns,
+      columnOrder,
       childrenKey,
       hasResizableColumns,
       sortBy,
       sortOrder,
+      onColumnReorder,
       onSort,
     } = args;
 
@@ -69,6 +79,45 @@ export default class HdsAdvancedTableTableModel {
     this.onSort = onSort;
 
     this.setupData({ model, columns, sortBy, sortOrder });
+
+    this.columnOrder =
+      columnOrder ??
+      this.columns.map((column) => {
+        // todo: make this work without column keys correctly
+        return column.key ?? '';
+      });
+
+    this.onColumnReorder = onColumnReorder;
+  }
+
+  get hasColumnBeingDragged(): boolean {
+    return this.reorderDraggedColumn !== null;
+  }
+
+  get reorderDraggedColumnCells(): HdsAdvancedTableCell[] {
+    if (this.reorderDraggedColumn === null) {
+      return [];
+    }
+
+    const { key } = this.reorderDraggedColumn;
+
+    return this.flattenedVisibleRows.map((row) => {
+      const cell = row.cells.find((cell) => cell.columnKey === key);
+
+      return cell!;
+    });
+  }
+
+  get orderedColumns(): HdsAdvancedTableColumn[] {
+    return this.columnOrder.map((key) => {
+      const column = this.columns.find((column) => column.key === key);
+
+      if (!column) {
+        throw new Error(`Column with key ${key} not found`);
+      }
+
+      return column;
+    });
   }
 
   get sortCriteria(): string | HdsAdvancedTableSortingFunction<unknown> {
@@ -223,6 +272,49 @@ export default class HdsAdvancedTableTableModel {
       this.collapseAll();
     } else {
       this.openAll();
+    }
+  }
+
+  @action
+  reorderColumn(targetColumn: HdsAdvancedTableColumn, side: 'left' | 'right') {
+    const sourceColumn = this.reorderDraggedColumn;
+
+    if (sourceColumn == null || sourceColumn === targetColumn) {
+      return;
+    }
+
+    const oldIndex = this.orderedColumns.indexOf(sourceColumn);
+    const newIndex = this.orderedColumns.indexOf(targetColumn);
+
+    if (oldIndex !== -1 && newIndex !== -1) {
+      const updated = [...this.columnOrder];
+
+      updated.splice(oldIndex, 1); // Remove from old position
+
+      // Calculate the insertion index based on the side
+      // If dropping to the right of the target, insert after the target
+      // If dropping to the left of the target, insert before the target
+      // Adjust for the shift in indices caused by removing the source column
+      const adjustedIndex =
+        side === 'right'
+          ? newIndex > oldIndex
+            ? newIndex
+            : newIndex + 1
+          : newIndex > oldIndex
+            ? newIndex - 1
+            : newIndex;
+
+      updated.splice(adjustedIndex, 0, sourceColumn.key as string); // Insert at new position
+
+      this.columnOrder = updated;
+
+      for (const row of this.rows) {
+        row.updateColumnOrder(updated);
+      }
+
+      sourceColumn.isBeingDragged = false;
+
+      this.onColumnReorder?.(updated);
     }
   }
 }
--- a/packages/components/src/components/hds/advanced-table/th-reorder-drop-target.hbs
+++ b/packages/components/src/components/hds/advanced-table/th-reorder-drop-target.hbs
@@ -1,0 +1,12 @@
+<div
+  class={{this.classNames}}
+  aria-hidden="true"
+  role="presentation"
+  {{style height=(if @tableHeight (concat (sub @tableHeight 3.5) "px"))}}
+  {{this._registerElement}}
+  {{on "dragover" this.handleDragOver}}
+  {{on "dragenter" this.handleDragEnter}}
+  {{on "dragleave" this.handleDragLeave}}
+  {{on "drop" this.handleDrop}}
+  ...attributes
+/>
--- a/packages/components/src/components/hds/advanced-table/th-reorder-drop-target.ts
+++ b/packages/components/src/components/hds/advanced-table/th-reorder-drop-target.ts
@@ -1,0 +1,149 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
+import { modifier } from 'ember-modifier';
+
+import type HdsAdvancedTableColumn from './models/column.ts';
+
+export interface HdsAdvancedTableThReorderDropTargetSignature {
+  Args: {
+    column: HdsAdvancedTableColumn;
+    tableHeight?: number;
+    onReorderDrop?: (
+      column: HdsAdvancedTableColumn,
+      side: 'left' | 'right'
+    ) => void;
+  };
+  Blocks: {
+    default?: [];
+  };
+  Element: HTMLDivElement;
+}
+
+export default class HdsAdvancedTableThReorderDropTarget extends Component<HdsAdvancedTableThReorderDropTargetSignature> {
+  @tracked private _dragSide: 'left' | 'right' | null = null;
+  @tracked private _isDraggingOver = false;
+  @tracked private _dragCount = 0;
+
+  private _element!: HdsAdvancedTableThReorderDropTargetSignature['Element'];
+
+  get classNames(): string {
+    const { column } = this.args;
+
+    const classes = ['hds-advanced-table__th-reorder-drop-target'];
+
+    if (column.isFirst) {
+      classes.push('hds-advanced-table__th-reorder-drop-target--is-first');
+    } else if (column.isLast) {
+      classes.push('hds-advanced-table__th-reorder-drop-target--is-last');
+    }
+
+    if (column.isBeingDragged) {
+      classes.push(
+        'hds-advanced-table__th-reorder-drop-target--is-being-dragged'
+      );
+    } else if (this._isDraggingOver && this._dragSide !== null) {
+      classes.push(
+        ...[
+          'hds-advanced-table__th-reorder-drop-target--is-dragging-over',
+          `hds-advanced-table__th-reorder-drop-target--is-dragging-over--${this._dragSide}`,
+        ]
+      );
+    }
+
+    return classes.join(' ');
+  }
+
+  private _registerElement = modifier(
+    (element: HdsAdvancedTableThReorderDropTargetSignature['Element']) => {
+      this._element = element;
+    }
+  );
+
+  private _resetDragState(): void {
+    this._dragCount = 0;
+    this._isDraggingOver = false;
+    this._dragSide = null;
+  }
+
+  // determines whether the drag event is occurring on the left or right side of the element
+  private _getDragSide(event: DragEvent): 'left' | 'right' {
+    const rect = this._element.getBoundingClientRect();
+    const mouseX = event.clientX;
+    const elementMiddleX = rect.left + rect.width / 2;
+
+    return mouseX < elementMiddleX ? 'left' : 'right';
+  }
+
+  @action
+  handleDragOver(event: DragEvent): void {
+    event.preventDefault();
+
+    const { column } = this.args;
+    const { reorderDraggedColumn } = column.table;
+
+    if (column.isBeingDragged || reorderDraggedColumn === null) {
+      return;
+    }
+
+    const { next, previous } = reorderDraggedColumn.siblings;
+    const dragSide = this._getDragSide(event);
+
+    if (
+      (column === previous && dragSide === 'left') ||
+      (column === next && dragSide === 'right') ||
+      (column !== previous && column !== next)
+    ) {
+      this._dragSide = dragSide;
+    }
+  }
+
+  @action
+  handleDragEnter(event: DragEvent): void {
+    event.preventDefault();
+
+    this._dragCount = this._dragCount + 1;
+
+    if (this._dragCount === 1) {
+      this._isDraggingOver = true;
+    }
+  }
+
+  @action
+  handleDragLeave(event: DragEvent): void {
+    event.preventDefault();
+
+    this._dragCount = this._dragCount - 1;
+
+    // ensure count doesn't go negative and reset isDraggingOver when appropriate
+    if (this._dragCount <= 0) {
+      this._resetDragState();
+    }
+  }
+
+  @action
+  handleDrop(event: DragEvent): void {
+    event.preventDefault();
+
+    const { column, onReorderDrop } = this.args;
+    const { _dragSide } = this;
+
+    // reset drag state completely when an item is dropped
+    this._resetDragState();
+
+    if (
+      column === undefined ||
+      _dragSide === null ||
+      typeof onReorderDrop !== 'function'
+    ) {
+      return;
+    }
+
+    onReorderDrop(column, _dragSide);
+  }
+}
--- a/packages/components/src/components/hds/advanced-table/th-reorder-handle.hbs
+++ b/packages/components/src/components/hds/advanced-table/th-reorder-handle.hbs
@@ -1,0 +1,8 @@
+{{! template-lint-disable no-pointer-down-event-binding }}
+<div
+  draggable="true"
+  class={{this.classNames}}
+  {{on "dragstart" this.handleDragStart}}
+  {{on "dragend" this.handleDragEnd}}
+  ...attributes
+/>
--- a/packages/components/src/components/hds/advanced-table/th-reorder-handle.ts
+++ b/packages/components/src/components/hds/advanced-table/th-reorder-handle.ts
@@ -1,0 +1,83 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+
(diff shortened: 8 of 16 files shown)


packages/components/src/styles/components/advanced-table.scss
```diff
@@ -583,3 +695,12 @@ $hds-advanced-table-button-size: 24px; // the size of the buttons and dropdown t
...
   background: linear-gradient(to bottom, rgba(59, 61, 69, 0%) 0%, rgba(59, 61, 69, 8%) 100%);
 }
+
+.hds-advanced-table__th-reorder-drag-preview {
+  position: absolute;
+  top: -9999px;
+  left: -9999px;
+  background-color: $hds-advanced-table-drag-preview-background-color;
+  border: 5px solid var(--token-color-focus-action-external);
+  border-radius: var(--token-border-radius-medium);
```
jorytindall, line 745: @zamoore is possible to add an elevation token here? Specifically `--token-elevation-mid-box-shadow`?

zamoore: Sure thing

packages/components/src/components/hds/advanced-table/types.ts
```diff
@@ -112,4 +112,18 @@ export interface HdsAdvancedTableOnSelectionChangeSignature {
   }[];
 }
 
-export type HdsAdvancedTableModel = Array<Record<string, unknown>>;
+export type HdsAdvancedTableModel = object[];
```
zamoore: I was getting ts errors with specifically-typed arrays like `FolkMusic` in the showcase. Making this more generic seems to work, but I want input on this choice for sure.

packages/components/src/components/hds/advanced-table/th-sort.ts
```diff
@@ -122,9 +140,32 @@ export default class HdsAdvancedTableThSort extends Component<HdsAdvancedTableTh
...
+    if (
+      column === undefined ||
+      column.key === undefined ||
+      typeof onReorderDragStart !== 'function'
+    ) {
+      return;
+    }
+
+    // Set the local state that shows this column is being dragged
+    column.isBeingDragged = true;
```
shleewhite: Nit: why is this being set in the context of the th-sort component? When the drag operation finishes, column.isBeingDragged gets set to false in the th-reorderhandle component. Might be nice to do both in the same file so you can see both happen easier.

packages/components/src/components/hds/advanced-table/th-reorder-handle.hbs
```diff
@@ -0,0 +1,19 @@
...
+  role="button"
+  draggable="true"
+  class={{this.classNames}}
+  tabindex="0"
+  aria-label={{hds-t
+    "hds.components.advanced-table.th-reorder-handle.aria-label"
+    columnLabel=@column.label
+    default=(concat "Reorder " @column.label " column")
+  }}
+  {{this._registerElement}}
```
shleewhite, line 16: Suggestion: add instructions for how a user can expect to reorder columns 

(would also need to be translated)

(....also I think our linter doesnt like aria-description, but it is valid aria so you can just add a lint ignore if needed)

```suggestion
  aria-description="Use left and right arrow keys while focused on this button to move the column to a new position."
  {{this._registerElement}}
```

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -270,6 +341,50 @@ $hds-advanced-table-button-size: 24px; // the size of the buttons and dropdown t
...
+.hds-advanced-table__th-reorder-handle {
+  position: absolute;
+  bottom: 0;
+  left: 50%;
+  z-index: 2;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 24px;
+  height: 16px;
```
dchyun: [Issue] I think this may be a [WCAG violation](https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html) as the minimum clickable area for items should be 24px by 24px. Maybe we can extend the clickable area while keeping the icon the same size.

packages/components/src/components/hds/advanced-table/th-reorder-handle.hbs
```diff
@@ -0,0 +1,19 @@
...
+  {{on "dragstart" this.handleDragStart}}
+  {{on "dragend" this.handleDragEnd}}
+  {{on "keydown" this.handleKeydown}}
+  ...attributes
+>
+  <Hds::Icon @name="move-horizontal" />
```
dchyun: [Note][Non-blocker] Just want to call out that the icons used for reordering and the cursor icon for resizing are quite similar. It was a little noticeable for me when hovering over the resize handle and my cursor would have the same icon as the button in the middle for reordering. There's a low chance someone could think the middle icon is for resizing. 

**Reordering**
Screenshot 2025-08-28 at 2 06 53 PM

**Resizing cursor**
https___dev-to-uploads s3 amazonaws com_uploads_articles_ln95u172m59cehotguqb

packages/components/src/components/hds/advanced-table/th-reorder-drop-target.hbs
```diff
@@ -0,0 +1,10 @@
+<div
+  class={{this.classNames}}
+  aria-hidden="true"
+  role="presentation"
```
shleewhite: Nit: this is redundant, divs have role presentation by default
```suggestion
```

packages/components/src/components/hds/advanced-table/th-context-menu.ts
```diff
@@ -124,10 +208,29 @@ export default class HdsAdvancedTableThContextMenu extends Component<HdsAdvanced
...
+    // eslint-disable-next-line ember/no-runloop
+    scheduleOnce(
+      'afterRender',
+      this,
+      this.args.column.focusReorderHandle.bind(this)
+    );
+  }
+
+  @action
+  private moveColumnToPosition(
```
dchyun: [Issue] When using the "move column to start" and "move column to end" actions with the keyboard, the focus is lost after moving in certain situations.

- Moving any column the end
- Moving any column to the start that is not the column adjacent to the start column

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -270,6 +341,50 @@ $hds-advanced-table-button-size: 24px; // the size of the buttons and dropdown t
...
+    height: 12px;
+  }
+
+  &:hover,
+  &.mock-hover {
+    background-color: var(--token-color-surface-interactive-hover);
+    cursor: grab;
+  }
+
+  &:active,
```
dchyun, line 399: [Nit] Small thing, but after completing a drag operation, the reorder handle is visible in the column that is in the old position of the moved column.

Example: I move "Artist" from 3rd to 4th. Album is now column 3. When the drag ends my mouse is over col 4 "Artist", but in col 3 "Album" the handle is visible.

Once I move my mouse it goes away so I think this has something to do with the active state.

showcase/app/templates/page-components/advanced-table.hbs
```diff
@@ -621,6 +621,90 @@
 
   <Shw::Divider @level={{2}} />
 
+  <Shw::Text::H2>Reorderable columns</Shw::Text::H2>
```
dchyun, line 624: [Question] Do we support reordering with selectable columns or with sticky columns / header? If so, would be nice to have an example of that.

packages/components/src/components/hds/advanced-table/th-context-menu.ts
dchyun: There is also an issue with focus being lost when using the start and end actions when the table has horizontal overflow. We may have to do something similar to what was done for the cell keyboard nav to keep things in view after a drop operation.
```
newCell.scrollIntoView({
  behavior: 'smooth',
  block: 'center',
  inline: 'center',
});
```
It would also be helpful to add an example to the showcase for column overflow. I was mainly testing this in mobile.

packages/components/src/components/hds/advanced-table/th.hbs
```diff
@@ -66,26 +66,48 @@
     {{/if}}
 
     {{#if @column}}
+      {{#if (and @hasReorderableColumns (not @isStickyColumn))}}
```
dchyun, line 80: [Suggestion] This could be moved below the context menu in the DOM order to make the focus order more logical.
In the current order the focus kind of goes in an out of the header when it jump down to the reorder handle before going back up to the context menu.

This jumping is more apparent in the search where it can go tooltip, sort, reorder, context menu.

packages/components/src/components/hds/advanced-table/th-reorder-handle.hbs
jorytindall: We did discuss this amongst designers a bit, but decided based on the context that it wasn't problematic. Though I'll encourage @hashicorp/hds-design to chime in with differing opinions.

packages/components/src/components/hds/advanced-table/th.ts
```diff
@@ -20,8 +20,9 @@ import type {
   HdsAdvancedTableScope,
   HdsAdvancedTableExpandState,
 } from './types.ts';
-import type { HdsAdvancedTableSignature } from './index.ts';
+import type { HdsAdvancedTableThReorderHandleSignature } from './th-reorder-handle.ts';
```
shleewhite, line 24: Just a thought, out of scope for this PR: but it would be nice to see if theres some way we can consolidate th-sort and th components. a lot of logic is duplicated across both components.

showcase/app/templates/page-components/advanced-table.hbs
zamoore: We do. I can add some examples.

packages/components/src/components/hds/advanced-table/th.hbs
zamoore: That's a good idea. I think that makes sense.

packages/components/src/styles/components/advanced-table.scss
zamoore: You're totally right. Done.

packages/components/src/components/hds/advanced-table/th.hbs
zamoore: Done. Let me know what you think.

packages/components/src/components/hds/advanced-table/models/table.ts
```diff
@@ -6,24 +6,39 @@
...
 
 import type { HdsAdvancedTableSignature } from '../index.ts';
 import type {
   HdsAdvancedTableExpandState,
+  HdsAdvancedTableCell,
+  HdsAdvancedTableColumnReorderCallback,
   HdsAdvancedTableSortingFunction,
 } from '../types';
 
+enum HdsAdvancedTableColumnReorderSideValues {
```
shleewhite: Suggestion: this should be moved to the types file to be more inline with how other types are defined.

shleewhite (commented):
Drag and drop looking great! Something got funky with the  focus state for the drag handle.

Image

Also, could you prevent default when you press space on the drag handle? If you don't. the browser scrolls the page which I think is aggressive in this situation.

The other big thing is we need a screen reader message that the column has moved. When you sort a column, a message gets added to the table's caption, we could do a similar thing here where when you move the column it announces its new position.

packages/components/src/components/hds/advanced-table/th-context-menu.ts
```diff
@@ -42,77 +44,159 @@ export default class HdsAdvancedTableThContextMenu extends Component<HdsAdvanced
...
+        if (index > 0) {
+          return [...options, { key: 'separator' }, ...group];
+        }
+        return [...options, ...group];
+      },
+      []
+    );
   }
 
   @action
```
dchyun:
```suggestion
```
[Suggestion] Since all of these context menu action functions are private now I think we can remove the `@action`. We could also update the names to use the `_privateFunction` syntax.

packages/components/src/components/hds/advanced-table/th-reorder-drop-target.ts
```diff
@@ -0,0 +1,151 @@
...
+
+import type HdsAdvancedTableColumn from './models/column.ts';
+
+export interface HdsAdvancedTableThReorderDropTargetSignature {
+  Args: {
+    column: HdsAdvancedTableColumn;
+    tableHeight?: number;
+    onReorderDrop?: (
+      column: HdsAdvancedTableColumn,
+      side: 'left' | 'right'
```
dchyun:
```suggestion
      side: HdsAdvancedTableColumnReorderSide
```
[Suggestion] Could we use the `HdsAdvancedTableColumnReorderSide` if we move it to the types file?

packages/components/src/components/hds/advanced-table/th-reorder-drop-target.ts
```diff
@@ -0,0 +1,151 @@
...
+    ) => void;
+  };
+  Blocks: {
+    default?: [];
+  };
+  Element: HTMLDivElement;
+}
+
+export default class HdsAdvancedTableThReorderDropTarget extends Component<HdsAdvancedTableThReorderDropTargetSignature> {
+  @tracked private _dragSide: 'left' | 'right' | null = null;
```
dchyun:
```suggestion
  @tracked private _dragSide: HdsAdvancedTableColumnReorderSide | null = null;
```

packages/components/src/components/hds/advanced-table/th-reorder-drop-target.ts
```diff
@@ -0,0 +1,151 @@
...
+  private _element!: HdsAdvancedTableThReorderDropTargetSignature['Element'];
+
+  private _registerElement = modifier(
+    (element: HdsAdvancedTableThReorderDropTargetSignature['Element']) => {
+      this._element = element;
+    }
+  );
+
+  // determines whether the drag event is occurring on the left or right side of the element
+  private _getDragSide(event: DragEvent): 'left' | 'right' {
```
dchyun:
```suggestion
  private _getDragSide(event: DragEvent): HdsAdvancedTableColumnReorderSide {
```

packages/components/src/components/hds/advanced-table/th-sort.ts
```diff
@@ -48,6 +52,12 @@ export interface HdsAdvancedTableThSortSignature {
     isStickyColumnPinned?: boolean;
     onColumnResize?: HdsAdvancedTableSignature['Args']['onColumnResize'];
     onPinFirstColumn?: () => void;
+    onReorderDragEnd?: () => void;
+    onReorderDragStart?: (column: HdsAdvancedTableColumn) => void;
+    onReorderDrop?: (
+      column: HdsAdvancedTableColumn,
+      side: 'left' | 'right'
```
dchyun:
```suggestion
      side: HdsAdvancedTableColumnReorderSide
```

packages/components/src/components/hds/advanced-table/th.ts
```diff
@@ -51,6 +54,12 @@ export interface HdsAdvancedTableThSignature {
     onClickToggle?: () => void;
     onColumnResize?: HdsAdvancedTableSignature['Args']['onColumnResize'];
     onPinFirstColumn?: () => void;
+    onReorderDragEnd?: () => void;
+    onReorderDragStart?: (column: HdsAdvancedTableColumn) => void;
+    onReorderDrop?: (
+      column: HdsAdvancedTableColumn,
+      side: 'left' | 'right'
```
dchyun:
```suggestion
      side: HdsAdvancedTableColumnReorderSide
```

showcase/app/components/mock/app/main/generic-advanced-table.gts
dchyun, line 1: [Issue] Testing reordering with pinned cols through this example. When I swap a new column to be first and pin it, the header and selectable col are pinned, but the content of the pinned col is not. 

Screenshot 2025-08-29 at 10 29 28 AM

showcase/app/styles/showcase-pages/advanced-table.scss
```diff
@@ -83,4 +83,10 @@ body.page-components-advanced-table {
     white-space: nowrap;
     text-overflow: ellipsis;
   }
+
+  .shw-component-advanced-table-reorder-handle-container
+    .hds-advanced-table__th-reorder-handle {
+    visibility: visible;
```
dchyun, line 93:
```suggestion
      visibility: visible;
```
[Nit] Fix style indentation

showcase/app/styles/showcase-pages/advanced-table.scss
```diff
@@ -83,4 +83,10 @@ body.page-components-advanced-table {
...
+    opacity: 1;
```
dchyun, line 94:
```suggestion
      opacity: 1;
```
[Nit] Fix style indentation

packages/components/src/components/hds/advanced-table/th.ts
zamoore: I've been thinking the same thing.

showcase/app/components/mock/app/main/generic-advanced-table.gts
dchyun, line 1: [Issue] The context menu of the pinned column still contains all of the actions for moving the column, even though the reorder handle is not provided. These options should be removed if the pinned column is intended to be prevented from moving. 
Screenshot 2025-08-29 at 10 38 38 AM

showcase/tests/integration/components/hds/advanced-table/index-test.js
```diff
@@ -282,6 +380,523 @@ const hbsResizableColumnsAdvancedTable = hbs`<div style="width: 1000px;">
...
+      this.set('columnOrder', ['year', 'album', 'artist']);
+      columnOrder = await getColumnOrder(this.columns);
+      assert.deepEqual(
+        columnOrder,
+        ['year', 'album', 'artist'],
+        'The column order is updated correctly',
+      );
+    });
+
+    module('with @hasStickyFirstColumn enabled', function () {
```
dchyun: [Suggestion] Can add a test in here that the pinned column does not have the reorder handle or context menu actions for reordering present.

dchyun:
> Something got funky with the focus state for the drag handle.

I think this is due to the change made for my earlier comment around increasing the size of the clickable area on the reorder handle.

packages/components/src/styles/components/advanced-table.scss
dchyun: Think the fix is causing some of the focus ring issues @shleewhite mentions.

packages/components/src/components/hds/advanced-table/th.hbs
dchyun: Looks great thanks for doing that.

packages/components/src/components/hds/advanced-table/th-sort.ts
zamoore: That's a good point. Fixed.

packages/components/src/components/hds/advanced-table/th-reorder-handle.hbs
zamoore: Makes sense

packages/components/src/components/hds/advanced-table/th-context-menu.ts
zamoore: Good call

showcase/app/templates/page-components/advanced-table.hbs
```diff
@@ -621,6 +621,174 @@
...
+            </B.Td>
+          {{/if}}
+        {{/each}}
+      </B.Tr>
+    </:body>
+  </Hds::AdvancedTable>
+
+  <Shw::Text::H4 @tag="h3">Reorderable columns with sticky first column</Shw::Text::H4>
+
+  <div class="shw-component-advanced-table-fixed-width-wrapper">
```
dchyun: [Suggestion] Can you make the width of this example wider? It's not quite possible to test out reordering columns with the first column pinned with how narrow it is right now.

showcase/app/templates/page-components/advanced-table.hbs
dchyun: Thanks for adding the sticky examples. Can you add one for selectable columns as well?

showcase/app/components/mock/app/main/generic-advanced-table.gts
dchyun: Still seeing this issue with the context menu present in the latest example.

dchyun, line 1: [Issue] When a column is pinned, and the table is scrolled, if you reorder a column and drag it over the pinned column it detected the drop target behind the pinned column. The drop target is only shown when hovering over the table body, not the header. 

This may be a z-index issue. The header cell has `z-index: 7` and the body cells have `z-index: 3`. Maybe the body cells z-index needs to be higher.


https://github.com/user-attachments/assets/5d492b88-a74d-4f4a-a284-6c546ec15b22

dchyun, line 1: [Question] When the first column is pinnable, clicking "move column to start" on another column in the table moves that column to 2nd not 1st. The pinnable column is still first even if it is not pinned at the time. Is this intentional?

dchyun (commented):
Thanks for addressing all my code suggestions. Most of my remaining feedback is around reordering + pinnable columns.

showcase/app/components/mock/app/main/generic-advanced-table.gts
zamoore: Should be addressed now

showcase/app/templates/page-components/advanced-table.hbs
zamoore: Yep, done.

showcase/app/templates/page-components/advanced-table.hbs
zamoore: Done

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -270,6 +341,65 @@ $hds-advanced-table-button-size: 24px; // the size of the buttons and dropdown t
...
+    cursor: grabbing;
+
+    .hds-advanced-table__th-reorder-handle__inner {
+      background-color: var(--token-color-surface-interactive-hover);
+    }
+  }
+
+  &:focus .hds-advanced-table__th-reorder-handle__inner,
+  &.mock-focus .hds-advanced-table__th-reorder-handle__inner {
+    @include hds-focus-ring-with-pseudo-element-focus-always-visible($position: absolute);
```
shleewhite: This is causing an issue where the focus ring is now never visible. This is because the inner part of the reorder handle is never focused, only the outer part is.

you could either 
1. move the focus styles back to the parent
2. make the inner part of the reorder handle the target. This would involve moving all the semantic parts (draggable, aria-label, aria-description, role, tabindex) to the inner part

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -111,6 +112,76 @@ $hds-advanced-table-button-size: 24px; // the size of the buttons and dropdown t
...
+    }
+
+    // dragging over left
+    .hds-advanced-table__th-reorder-drop-target--is-dragging-over--left::after {
+      left: 0;
+      transform: translateX(-3px);
+    }
+
+    // dragging over left and is first
+    .hds-advanced-table__th-reorder-drop-target--is-dragging-over--left.hds-advanced-table__th-reorder-drop-target--is-first::after {
```
dchyun, line 170: [Issue] This selector shouldn't be applied if there is a selectable column. Border radius isn't needed in that situation.

Screenshot 2025-09-08 at 10 21 43 AM

packages/components/src/components/hds/advanced-table/th-context-menu.ts
dchyun: Just resurfacing this. Still seeing focus getting lost in this situation.

packages/components/src/components/hds/advanced-table/th-context-menu.ts
```diff
@@ -42,77 +43,157 @@ export default class HdsAdvancedTableThContextMenu extends Component<HdsAdvanced
...
 
-    if (isStickyColumn !== undefined && column.isFirst) {
-      const translatedPinLabel = this.hdsIntl.t(
-        'hds.advanced-table.th-context-menu.pin',
-        { default: 'Pin column' }
-      );
-      const translatedUnpinLabel = this.hdsIntl.t(
-        'hds.advanced-table.th-context-menu.unpin',
-        { default: 'Unpin column' }
+    if (!column.isLast) {
```
dchyun, line 117: [Issue] When column pinning is allowed, but the first column is not pinned, the context menu does not show the options for "move column" or "move column to end". The reorder handle is shown and is usable.

showcase/app/components/mock/app/main/generic-advanced-table.gts
dchyun: Looked into this issue more. The issue is after a new column (a `td`) is moved to the first position, the sticky column class is being added to the old first column (a `th`) and not to the new one. There are a few sub-issues probably related to this.

- On the `th` `@isStickyColumn` is wrongly getting set to `true` when its not in the first position
- The `td` component doesn't take in any sticky arguments from the parent, and has no logic for applying sticky classes that currently exist in the `th`. When it's moved into the first position it doesn't have any way of showing it is sticky.

zamoore: Decision was made to not support reordering sticky columns or unpinning sticky columns in tables with column reordering.

zamoore: I'm unable to reproduce this one. It could be that something has changed, but I'm curious if you're still seeing it, which browser you're seeing it in

packages/components/src/styles/components/advanced-table.scss
zamoore: Alright, I think I managed to figure out a solution by adjusting the positioning of the ring. It's a little dirty because the focus ring is still applied to the outer element, but it works.

packages/components/src/styles/components/advanced-table.scss
zamoore: Should be addressed now

packages/components/src/components/hds/advanced-table/th-context-menu.ts
zamoore: This should now be correct based on our decision to not allow moving the first column if isStickyFirstColumn is true. Let me know if I'm missing something else here.

dchyun: Looks all good to me!

packages/components/src/components/hds/advanced-table/th-context-menu.ts
dchyun, line 1: [Issue] In Safari, the context menu items are unreachable via keyboard. When opening the context menu with `enter`, the expected behavior is clicking `tab` focuses each menu item. Instead what is happening is on `tab` the context menu is closing and the next focusable element is focused. 

This issue is observed in all context menu functionality, not just what was introduced here and can be addressed in a separate ticket if need be.

showcase/app/components/mock/app/main/generic-advanced-table.gts
dchyun: I tested again and was able to reproduce this issue in Chrome and Safari. There is a related issue I see too when moving columns with the arrow keys. When moving a column behind the sticky column the reorder handle is visible on top of the table cells, but not the header.


https://github.com/user-attachments/assets/df83c151-3782-4c4a-a458-36615a911563

packages/components/src/styles/components/advanced-table.scss
zamoore: Good catch. Done.

packages/components/src/components/hds/advanced-table/th-context-menu.ts
zamoore: This is a tough one to crack. scrollIntoView is doing more vertical scrolling than anything.

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -511,10 +506,83 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
+      assert(
+        `Cannot have resizable columns if there are nested rows.`,
+        !hasResizableColumns
+      );
+    }
+
+    if (hasReorderableColumns) {
+      assert(
+        'Cannot have both reorderable columns and a sticky first column.',
+        !hasStickyFirstColumn
```
dchyun:
```suggestion
        !(hasStickyFirstColumn === undefined)
```
[Issue] If `hasStickyFirstColumn` is `false` the context menu item for pinning and upinning the column is still shown. So a user could end up setting the column to pinned. The menu item is only not shown when it is `undefined`.

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -511,10 +506,83 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
+      );
+
+      assert(
+        `Cannot have sortable columns if there are nested rows. Sortable columns are ${sortableColumnLabels.toString()}`,
+        sortableColumns.length === 0
+      );
+
+      assert(
+        'Cannot have a sticky first column if there are nested rows.',
+        !hasStickyFirstColumn
```
dchyun:
```suggestion
       !(hasStickyFirstColumn === undefined)
```
[todo] I know this is old code, but this should be updated to reflect the hasStickyFirstColumn behavior mentioned previously as well.

packages/components/src/components/hds/advanced-table/models/table.ts
```diff
@@ -246,4 +306,133 @@ export default class HdsAdvancedTableTableModel {
...
+
+    table.moveColumnToTarget(column, targetColumn, side);
+  }
+
+  @action
+  moveColumnToTerminalPosition(
+    column: HdsAdvancedTableColumn,
+    position: 'start' | 'end'
+  ): void {
+    const firstNonStickyColumn = this.orderedColumns.find(
```
dchyun: [Question] Since we've removed support for reorderable columns + sticky columns can this be simplified to remove checks for sticky columns?

packages/components/src/components/hds/advanced-table/models/column.ts
```diff
@@ -93,19 +112,23 @@ export default class HdsAdvancedTableColumn {
...
 
-    return columns.findIndex((column) => column.key === this.key);
+    return orderedColumns.findIndex((column) => column.key === this.key);
   }
 
   get isFirst(): boolean {
     return this.index === 0;
   }
 
+  get isFirstNonSticky(): boolean {
```
dchyun: [Question] Can this be removed now that we're not supporting reorderable columns + sticky columns?

packages/components/src/components/hds/advanced-table/models/table.ts
```diff
@@ -52,24 +69,67 @@ export default class HdsAdvancedTableTableModel {
 
   childrenKey?: HdsAdvancedTableTableArgs['childrenKey'];
   hasResizableColumns?: HdsAdvancedTableTableArgs['hasResizableColumns'];
+  hasStickyFirstColumn?: HdsAdvancedTableTableArgs['hasStickyFirstColumn'];
```
dchyun: [Question] Can this also be removed? It's only being used in the `stepColumn` functions I believe which is related to reordering.

packages/components/src/components/hds/advanced-table/th-context-menu.ts
dchyun: Created a ticket for this issue. It is an existing issue and not introduced here. https://hashicorp.atlassian.net/browse/HDS-5483

packages/components/src/components/hds/advanced-table/th-context-menu.ts
```diff
@@ -42,77 +45,168 @@ export default class HdsAdvancedTableThContextMenu extends Component<HdsAdvanced
...
+    let reorderOptions: HdsAdvancedTableThContextMenuOption[] = [
+      {
+        key: 'reorder-column',
+        label: translatedMoveColumnLabel,
+        icon: 'move-horizontal',
+        action: () => this._moveColumn(),
+      },
+    ];
+
+    if (!column.isFirst && !hasStickyFirstColumn) {
```
dchyun:
```suggestion
    if (!column.isFirst) {
```
[Todo] Could remove this check since reordering and sticky col aren't supported

packages/components/src/components/hds/advanced-table/th-reorder-handle.ts
```diff
@@ -0,0 +1,125 @@
...
+  return dragPreviewElement;
+}
+
+export default class HdsAdvancedTableThReorderHandle extends Component<HdsAdvancedTableThReorderHandleSignature> {
+  private _registerElement = modifier((element: HTMLDivElement) => {
+    this.args.column.reorderHandleElement = element;
+  });
+
+  get classNames(): string {
+    const classes = ['hds-advanced-table__th-reorder-handle'];
```
dchyun: [Nit] Could add the class directly in the template since there's only one class.

packages/components/src/components/hds/advanced-table/th-sort.hbs
```diff
@@ -44,21 +44,31 @@
...
-            @onColumnResize={{@onColumnResize}}
-            @onPinFirstColumn={{@onPinFirstColumn}}
+            @tableHeight={{@tableHeight}}
+            @onReorderDragStart={{this.handleDragStart}}
+            @onReorderDragEnd={{@onReorderDragEnd}}
+            {{this._registerReorderHandleElement}}
           />
         {{/if}}
 
         {{#if (and @hasResizableColumns (not @column.isLast))}}
```
dchyun:
```suggestion
        {{#if (and @hasResizableColumns (not @column.isLast) (not @column.table.hasColumnBeingDragged))}}
```
[todo] The `th` template adds in this check for being dragged, so I assume the same check needs to be added for `th-sort`

packages/components/translations/hds/components/advanced-table/th-reorder-handle/en-us.yaml
```diff
@@ -0,0 +1,2 @@
+aria-description: "Use left and right arrow keys while focused on this button to move the column to a new position."
```
dchyun:
```suggestion
aria-description: Use left and right arrow keys while focused on this button to move the column to a new position.
```
[Nit] remove quotation marks

packages/components/translations/hds/components/advanced-table/th-reorder-handle/en-us.yaml
```diff
@@ -0,0 +1,2 @@
...
+aria-label: "Reorder {columnLabel} column"
```
dchyun, line 2:
```suggestion
aria-label: Reorder {columnLabel} column
```
[Nit] remove quotation marks

showcase/tests/integration/components/hds/advanced-table/index-test.js
```diff
@@ -282,6 +380,422 @@ const hbsResizableColumnsAdvancedTable = hbs`<div style="width: 1000px;">
 module('Integration | Component | hds/advanced-table/index', function (hooks) {
   setupRenderingTest(hooks);
 
+  module('column reordering', function (hooks) {
+    hooks.beforeEach(function () {
+      setReorderableColumnsTableData(this);
+    });
+
+    test('it renders reorder handles when reordering is enabled', async function (assert) {
```
dchyun, line 388: [Suggestion] Could add a check somewhere in here that the reorder drop handle is not shown for selectable columns.

packages/components/src/components/hds/advanced-table/th-context-menu.ts
zamoore: Alright, I think this one is solved now

showcase/app/components/mock/app/main/generic-advanced-table.gts
zamoore: We've removed support for reordering and sticky columns now, so this should no longer be an issue.

packages/components/src/components/hds/advanced-table/th-context-menu.ts
dchyun: After further testing, there were certain settings in my Safari that were causing the issue. Working as expected now.

packages/components/translations/hds/components/advanced-table/th-reorder-handle/en-us.yaml
zamoore: This one needs them due to the dynamic segment

packages/components/src/components/hds/advanced-table/th-sort.hbs
zamoore: Good call

showcase/tests/integration/components/hds/advanced-table/index-test.js
zamoore: Good call. Added

packages/components/src/components/hds/advanced-table/th-sort.hbs
```diff
@@ -44,21 +44,30 @@
...
+          @hasReorderableColumns={{@hasReorderableColumns}}
+          @hasResizableColumns={{@hasResizableColumns}}
+          @isStickyColumn={{@isStickyColumn}}
+          @reorderHandleElement={{this._reorderHandleElement}}
+          @resizeHandleElement={{this._resizeHandleElement}}
+          @onColumnResize={{@onColumnResize}}
+          @onPinFirstColumn={{@onPinFirstColumn}}
+        />
+
+        {{#if (and @hasReorderableColumns (not @isStickyColumn) (not @column.table.hasColumnBeingDragged))}}
```
dchyun:
```suggestion
        {{#if (and @hasReorderableColumns (not @isStickyColumn))}}
```
[Issue] In your update to my last comment you added this check into the reorder handle block when it should be in resize handle block. This is currently causing the reordering on the sortable cols to not work.

packages/components/src/components/hds/advanced-table/th-sort.hbs
dchyun: It looks like this update got added into the wrong place. It needs to be on the resize handle if block, not the reorder handle.

packages/components/src/styles/components/advanced-table.scss
zamoore: I've played around with a few ways to try to blur this programmatically, but haven't had luck with getting it to work. I can add this as a potential improvement down the line.

.changeset/wet-files-joke.md
```diff
@@ -0,0 +1,11 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+<!-- START components/table/advanced-table -->
+
+`AdvancedTable` - Added support for column reordering.
```
dchyun, line 6: [Question] is `reorderedMessageText` a new argument as well?

.changeset/wet-files-joke.md
```diff
@@ -0,0 +1,12 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+<!-- START components/table/advanced-table -->
+
```
shleewhite:
```suggestion
```

Will this extra return cause issues / awkwardness with the automations?

packages/components/src/components/hds/advanced-table/models/table.ts
```diff
@@ -246,4 +302,130 @@ export default class HdsAdvancedTableTableModel {
...
+    if (oldIndex !== -1 && newIndex !== -1) {
+      const updated = [...this.columnOrder];
+
+      updated.splice(oldIndex, 1); // Remove from old position
+
+      // Calculate the insertion index based on the side
+      // If dropping to the right of the target, insert after the target
+      // If dropping to the left of the target, insert before the target
+      // Adjust for the shift in indices caused by removing the source column
+      const adjustedIndex =
```
shleewhite, line 382: Suggestion: this is totally a style thing that you are more than welcome to ignore, but I find these nested ternaries harder than necessary to understand. I'd just opt for an if statement here.

packages/components/src/components/hds/advanced-table/models/row.ts
```diff
@@ -35,7 +38,33 @@ export default class HdsAdvancedTableRow {
     return this.isOpen && this.hasChildren;
   }
 
+  get orderedCells(): HdsAdvancedTableCell[] {
+    return this.columnOrder.map((key) => {
+      const cell = this.cells.find((cell) => cell.columnKey === key);
+
+      if (cell === undefined) {
+        throw new Error(`Column with key ${key} not found`);
```
shleewhite:
```suggestion
        throw new Error(`Cell in the column with key ${key} not found for the row.`);
```

just to make it a little clearer what is going on.

.changeset/wet-files-joke.md
dchyun: It won't but still not a bad idea to remove it

zamoore:
> ok I hate to do this but I found a teeeeeeeny styling bug
> 
> the fantom border is still showing up on the reorder handle if you have a sticky header that is scrolled a bit down. it goes away when the header is not pinned.
> 
> Image

@shleewhite 

Alright, this one was a little tricky. This was the same kind of stacking context issue we were seeing with the pinnable column. I made a change to where the scroll shadow is rendered and updated the styles/js a bit. Let me know what you think of this approach.

dchyun (approved):
🚢 🎊

PR #2999, opened 2025-07-08
Showcase/ts batch 1

shleewhite:
### :pushpin: Summary

If merged, this PR would convert the following pages in the showcase to TypeScript:
* AppHeader
* AppSideNav
* Copy::Button

### :link: External links

Figma file: [if it applies]

***

<details>

</details>


showcase/app/templates/components/app-header/index.hbs
```diff
@@ -451,25 +451,13 @@ SPDX-License-Identifier: MPL-2.0
   <Shw::Flex as |SF|>
     <SF.Item>
       <div class="hds-app-header">
-        <Hds::AppHeader::HomeLink
-          @icon="hashicorp"
-          @text="HashiCorp"
-          @isIconOnly={{false}}
-          @href="#"
-          @title="Admin UI"
```
shleewhite: Note: `@title` is no longer an argument on Hds::AppHeader::HomeLink

showcase/app/templates/components/app-header/index.hbs
```diff
@@ -35,7 +35,7 @@ SPDX-License-Identifier: MPL-2.0
     </SF.Item>
   </Shw::Flex>
 
-  <Shw::Divider @level="2" />
+  <Shw::Divider @level={{2}} />
```
shleewhite: Note: Shw::Divider expects either undefined or the number 2 for the level arg.

showcase/app/templates/components/app-header/index.hbs
```diff
@@ -547,13 +535,13 @@ SPDX-License-Identifier: MPL-2.0
...
 
   <Shw::Text::H3>Dropdowns within AppHeader</Shw::Text::H3>
 
   <Shw::Text::H4>States</Shw::Text::H4>
 
-  <Shw::Grid @columns="6" class="shw-component-dropdown-states-matrix" as |SF|>
+  <Shw::Grid @columns={{6}} class="shw-component-dropdown-states-matrix" as |SF|>
```
shleewhite: Note: columns arg expects a number, not a string.

showcase/app/templates/components/app-side-nav/index.hbs
```diff
@@ -84,7 +84,7 @@
         <Nav.Link @text="Access control (IAM)" @icon="users" @href="#" @hasSubItems={{true}} />
         <Nav.Link @text="Billing" @icon="credit-card" @href="#" @hasSubItems={{true}} />
         <Nav.Link @text="Settings" @icon="settings" @href="#" @hasSubItems={{true}} />
-        <Nav.Link @href="#" @isHrefExternal="true" @icon="guide" @text="Documentation" />
+        <Nav.Link @href="#" @isHrefExternal={{true}} @icon="guide" @text="Documentation" />
```
shleewhite: Note: isHrefExternal expects a boolean, not a string.

showcase/app/routes/components/app-header/index.ts
```diff
@@ -8,15 +8,20 @@ import Route from '@ember/routing/route';
...
+
 export default class ComponentsAppHeaderRoute extends Route {
   model() {
     // these are used only for presentation purpose in the showcase
     const TOGGLE_STATES = ['default', 'hover', 'active', 'focus', 'disabled'];
     const ITEM_STATES = ['default', 'hover', 'active', 'focus'];
     return {
-      TOGGLE_BUTTON_COLORS,
+      TOGGLE_BUTTON_COLORS:
+        TOGGLE_BUTTON_COLORS as HdsDropdownToggleButtonColors[],
```
shleewhite: Note: `TOGGLE_BUTTON_COLORS` is exported from the dropdown component as `string[]`.  When we iterate over the array in the template to show all the options, glint thinks the type doesn't match the type for the argument because it is too generic.

The workaround is to re-cast it as the correct type - but something we could look into in the future is updating what is exported from the component so we don't have to do this.

(this kind of change will happen a lot across all the showcase pages)

showcase/app/controllers/components/copy/button.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  routeDidChange() {
+    if (this.router.currentRoute?.name === 'components.copy.button') {
+      // eslint-disable-next-line ember/no-runloop
+      scheduleOnce('afterRender', this, this.replaceMockCopyStatus.bind(this));
+    }
+  }
+
+  replaceMockCopyStatus() {
```
shleewhite: Note: I moved this function into the controller because it needed to have access to `this.model`. this was working before, but when converted to TS it was undefined.

showcase/app/controllers/components/copy/button.ts
```diff
@@ -0,0 +1,86 @@
...
+      }
+    });
+  }
+
+  get bigIntNumber() {
+    const bigIntNumber = BigInt(12345678910);
+    return bigIntNumber;
+  }
+
+  get targetNodeElement() {
```
shleewhite: Note: updated this function because the `targetToCopy` argument on copy button expects either HTMLElement or undefined.

document.querySelector returns Element or null.

showcase/app/controllers/components/copy/button.ts
```diff
@@ -0,0 +1,86 @@
...
+import { inject as service } from '@ember/service';
+import { scheduleOnce } from '@ember/runloop';
+
+import type Owner from '@ember/owner';
+import type { Registry as Services } from '@ember/service';
+
+import type { ComponentsCopyButtonModelSignature } from 'showcase/routes/components/copy/button';
+
+export default class CopyButtonController extends Controller {
+  declare model: ComponentsCopyButtonModelSignature;
```
shleewhite: Note: this is what tells the typescript compiler that the model exists.

showcase/app/controllers/components/copy/button.js
```diff
@@ -1,80 +0,0 @@
...
-  get bigIntNumber() {
-    let bigIntNumber = BigInt(12345678910);
-    return bigIntNumber;
-  }
-
-  get targetNodeElement() {
-    return document.querySelector('#test-target-node-element');
-  }
-
-  get targetMultipleNodeElements() {
```
shleewhite: Note: I removed this getter from the typescript version because:
1. It was never used in the showcase template
2. Hds::Copy::Button doesn't support copying multiple nodes like this

showcase/app/templates/components/app-side-nav/index.hbs
```diff
@@ -45,9 +45,9 @@
   </Shw::Flex>
 
   <Hds::AppSideNav::Portal @targetName="sidenav-portal-demo-1" @ariaLabel="Primary on portal demo 1" as |Nav|>
-    <Nav.extraBefore><Shw::Placeholder @height="72px" @text="extraBefore" @background="#f3d9c5" /></Nav.extraBefore>
+    <Nav.ExtraBefore><Shw::Placeholder @height="72px" @text="extraBefore" @background="#f3d9c5" /></Nav.ExtraBefore>
```
shleewhite: Note: this causes a small change in Percy because now the blocks are rendered (they weren't before)

showcase/app/templates/components/copy/button.hbs
```diff
@@ -188,13 +188,13 @@
...
     </SF.Item>
-  </Shw::Flex>
+  </Shw::Grid>
 
   <Shw::Text::H4>HDS components</Shw::Text::H4>
 
   <Shw::Grid @columns={{3}} @gap="2rem" as |SG|>
     <SG.Item @label="Input component" class="shw-component-copy-button-input-component">
-      <Hds::Form::TextInput::Field @value="036140285924" @name="test-input" @id="test-input" as |F|>
+      <Hds::Form::TextInput::Field @value="036140285924" name="test-input" @id="test-input" as |F|>
```
shleewhite: Note: there is no name argument on Field components, need to pass it as an HTML attribute instead.

showcase/app/templates/components/copy/button.hbs
```diff
@@ -167,7 +167,7 @@
 
   <Shw::Text::H4>Target types</Shw::Text::H4>
 
-  <Shw::Grid @columns={{3}} @gap="2rem" as |SF|>
+  <Shw::Flex @gap="2rem" as |SF|>
```
shleewhite: Note: columns isnt an argument on Flex (its only on Grid), opted to change it back to Flex and remove the columns argument to avoid changes in Percy.

showcase/app/controllers/components/copy/button.ts
```diff
@@ -0,0 +1,89 @@
...
+  @service router!: Services['router'];
```
aklkv:
```suggestion
  @service declare router: RouterService;
```

showcase/app/controllers/components/copy/button.ts
```diff
@@ -0,0 +1,89 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+import { inject as service } from '@ember/service';
```
aklkv:
```suggestion
import { service } from '@ember/service';
```

showcase/app/controllers/components/copy/snippet.ts
```diff
@@ -0,0 +1,49 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Controller from '@ember/controller';
+import { inject as service } from '@ember/service';
```
aklkv:
```suggestion
import { service } from '@ember/service';
```

aklkv: you might as well just create separate PR to find and replace this imports at least for showcase and website for now

showcase/app/controllers/components/copy/snippet.ts
```diff
@@ -0,0 +1,49 @@
...
+import { scheduleOnce } from '@ember/runloop';
+
+import type Owner from '@ember/owner';
+import type { Registry as Services } from '@ember/service';
+
+import type { ComponentsCopySnippetModelSignature } from 'showcase/routes/components/copy/snippet';
+
+export default class CopySnippetController extends Controller {
+  declare model: ComponentsCopySnippetModelSignature;
+  @service router!: Services['router'];
```
aklkv:
```suggestion
  @service declare router: RouterService;
```

showcase/app/controllers/components/copy/button.ts
alex-ju: For context, Alexey is referring to the following deprecation https://deprecations.emberjs.com/id/importing-inject-from-ember-service/. However, we're still supporting 3.28 at the moment, so we'd need to operate such changes closer to the major release – and can be done across the library with a simple replace.

showcase/app/templates/page-components/pagination/index.hbs
```diff
@@ -244,22 +244,22 @@
         <SF.Item @label={{capitalize state}} mock-state-value={{state}} mock-state-selector="button,a">
           <Shw::Flex as |SF|>
             <SF.Item>
-              <Hds::Pagination::Nav::Number @page="1" />
+              <Hds::Pagination::Nav::Number @page={{1}} @isSelected={{false}} @onClick={{this.noop}} />
```
aklkv, line 247: [nit]: why not just create reusable `noop` helper, I noticed it's copy pasted all over the place

showcase/app/templates/page-components/pagination/index.hbs
```diff
@@ -292,6 +292,7 @@
 
   <div class="shw-component-pagination-table-demo" id="demo1-numbered-with-events">
     <Hds::Table
+      {{! @glint-expect-error }}
```
aklkv: ideally we should include comment why is this needed

showcase/app/templates/page-components/pagination/index.hbs
shleewhite: It might be better to wait to do that until we convert the files to `gts`. I dont see it being super until it can be directly imported into the file where the template is.

showcase/app/controllers/page-components/copy/button.ts
```diff
@@ -0,0 +1,89 @@
...
+  @action
+  deactivateModal() {
+    this.isModalActive = false;
+  }
+
+  // DEBUG
+  // uncomment these if you need to debug the `onSuccess/onError` callback methods
+
+  // @action
+  // onSuccess({ trigger, text, target }) {
```
dchyun:
```suggestion
  // onSuccess({ trigger, text, target }: { trigger: EventTarget; text: string; target: HTMLElement }) {
```
[Nit] If these lines to get commented out, some type issues will get thrown

showcase/app/controllers/page-components/copy/button.ts
```diff
@@ -0,0 +1,89 @@
...
+  //   console.log('onSuccess invoked in the controller', trigger, text, target);
+  // }
+
+  // @action
+  // onError({ trigger, text, target }) {
```
dchyun:
```suggestion
  // onError({ trigger, text, target }: { trigger: EventTarget; text: string; target: HTMLElement }) {
```
[Nit] If these lines to get commented out, some type issues will get thrown

showcase/app/controllers/page-components/pagination/index.ts
```diff
@@ -8,22 +8,39 @@ import { action } from '@ember/object';
...
 
+import type RouterService from '@ember/routing/router-service';
+
+import type { HdsPaginationDirections } from '@hashicorp/design-system-components/components/hds/pagination/types';
+import type { HdsTableThSortOrder } from '@hashicorp/design-system-components/components/hds/table/types';
+
+import type { PageComponentsPaginationModel } from 'showcase/routes/page-components/pagination/index';
+
 // uncomment this to override the `atob/btoa` functions for debugging
 // const atob = (s) => s;
```
dchyun:
```suggestion
// const atob = (s: string) => s;
```
[Nit] if this does get commented out it will cause type errors

showcase/app/controllers/page-components/pagination/index.ts
```diff
@@ -8,22 +8,39 @@ import { action } from '@ember/object';
...
 // const btoa = (s) => s;
```
dchyun:
```suggestion
// const btoa = (s: string) => s;
```
[Nit] if this does get commented out it will cause type errors

showcase/app/controllers/page-components/pagination/index.ts
```diff
@@ -284,10 +258,11 @@ export default class PageComponentsPaginationController extends Controller {
     const { newPrevCursor, newNextCursor } = this.newPrevNextCursors_demo4;
     const currPrevCursor = this.prevCursor_demo4;
     const currNextCursor = this.nextCursor_demo4;
-    return (page, pageSize) => {
+    return (page: HdsPaginationDirections, pageSize: number) => {
       // for the "compact" pagination when the user changes the page size and the `onPageSizeChange` function is invoked
       // the callback function returns a `null` value for the `page` argument so the consumer can decide how to handle the cursors acordingly
       if (page === null) {
+        console.log('page is null');
```
dchyun:
```suggestion
```
[Question] Can this be removed?

showcase/app/controllers/page-components/pagination/index.ts
```diff
@@ -306,17 +281,23 @@ export default class PageComponentsPaginationController extends Controller {
...
 
   get isDisabledNext_demo4() {
-    let { newNextCursor } = this.newPrevNextCursors_demo4;
+    const { newNextCursor } = this.newPrevNextCursors_demo4;
     return newNextCursor === null;
   }
 
-  get paginatedData_demo4() {
-    let token;
+  //   get paginatedData_demo2() {
```
dchyun: [todo] Remove commented code

showcase/app/controllers/page-components/pagination/index.ts
```diff
@@ -340,12 +321,18 @@ export default class PageComponentsPaginationController extends Controller {
       end = cursorIndex + pageSize;
     }
 
+    console.log('paginatedData_demo4 invoked with:');
```
dchyun: [todo] remove these console logs

showcase/app/routes/page-components/pagination/index.ts
```diff
@@ -0,0 +1,26 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Route from '@ember/routing/route';
+import type { ModelFrom } from 'showcase/utils/ModelFromRoute';
+
+interface User {
```
dchyun, line 9: [Suggestion] Since these mock api files are used across multiple component maybe we should define a types file for them somewhere.

shleewhite: Yeah, I think that might be a good follow up ticket.. @aklkv suggested actually making a reusable fixture that imports the typed user data. So I think if we do that, we could have the type in that file.

showcase/app/routes/page-components/app-side-nav/index.ts
```diff
@@ -0,0 +1,12 @@
+/**
```
didoo, line 1: I see that `ComponentsAppSideNavModel` (similar for `PageComponentsAppHeaderModel` above) is never used, but is declared anyway. Does it mean that in the future for every component we have to add this block of code (for consistency, I imagine?). In that case, do you think is worth to update our blueprint script to account for this new code pattern?

showcase/app/controllers/page-components/copy/button.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  routeDidChange() {
+    if (this.router.currentRoute?.name === 'page-components.copy.button') {
+      // eslint-disable-next-line ember/no-runloop
+      scheduleOnce('afterRender', this, this.replaceMockCopyStatus.bind(this));
+    }
+  }
+
+  replaceMockCopyStatus() {
```
didoo, line 34: [question] in the `showcase/app/controllers/page-components/pagination/index.ts` the functions `getCursorParts` and `getNewPrevNextCursors` have been left outside of the main class, while here `replaceMockCopyStatus` has been moved inside. What is the reason/difference between the two approaches? in which case one should be used vs the other?

showcase/app/controllers/page-components/pagination/index.ts
```diff
@@ -84,74 +102,30 @@ export default class PageComponentsPaginationController extends Controller {
...
+  noop() {}
 
   get demoRouteName() {
-    // eg. 'page-components.pagination';
-    return this.router.currentRouteName;
+    // eg. 'components.pagination';
+    const routeName = this.router.currentRouteName;
+
+    if (!routeName) return '';
+    return routeName;
```
didoo: [suggestion] would this work?
```suggestion
    return routeName ?? '';
```

didoo (dismissed):
LGTM!

showcase/app/routes/page-components/app-side-nav/index.ts
shleewhite: We don't have to add it if its not used... I just thought it would be good for consistency. I can definitely update the blueprint script after this is merged (itd be easier to review that change just on its own)

showcase/app/controllers/page-components/copy/button.ts
shleewhite: I moved it into the main class in this case because it was using `this`, the pagination functions don't need to access  it. For typing it was just a lot simpler to move the function into the main class.

showcase/app/controllers/page-components/pagination/index.ts
shleewhite: yes!

showcase/app/controllers/page-components/copy/button.ts
didoo: makes sense, thanks for the clarification

didoo (approved):

PR #3000, opened 2025-07-08
`RadioCard` - Fix hover styles shown when disabled

dchyun:
### :pushpin: Summary

If merged, this PR would fix an issue in the `RadioCard` where hover state styles were visible when `@disabed` was `true`.

It was [observed in atlas](https://hashicorp.slack.com/archives/C7KTUHNUS/p1751998251683399) that `RadioCard` components were still having hover styles applied even when disabled. This same observation was **not** observed in the showcase. 

When `:hover` is applied to a disabled card, the selector for the hover styles `.hds-form-radio-card:hover` takes precedence over the selector for the disabled styles `.hds-form-radio-card--disabled`.

In the showcase the styles are trigged by the `mock-` class with the selector `.hds-form-radio-card.mock-disabled`. This takes precedence over the hover selector. This is what is causing the mismatch between the showcase and the atlas example.

To solve this issue I have increased the specificity of the selectors for the disabled state. The showcase has also been updated to no longer use the `mock-` selector for the disabled state, since the `@disabled` argument is available. 

**Before**

https://github.com/user-attachments/assets/b23a454f-b96c-4b52-ad73-e272fc95102d

**After**

https://github.com/user-attachments/assets/2f853bb1-468f-4a81-8bcd-bfc3cbb8c936

### :link: External links

Jira ticket: [HDS-5084](https://hashicorp.atlassian.net/browse/HDS-5084)

***

<details>

</details>

[HDS-5084]: https://hashicorp.atlassian.net/browse/HDS-5084?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/styles/components/form/radio-card.scss
+++ b/packages/components/src/styles/components/form/radio-card.scss
@@ -76,6 +76,7 @@
     }
   }
 
+  // note: using full class name so it overrides the :hover styles
   &.hds-form-radio-card--disabled {
     background-color: var(--token-color-surface-interactive-disabled);
     border-color: var(--token-color-border-primary);
--- a/packages/components/src/styles/components/form/radio-card.scss
+++ b/packages/components/src/styles/components/form/radio-card.scss
@@ -76,8 +76,7 @@
     }
   }
 
-  &--disabled,
-  &.mock-disabled {
+  &.hds-form-radio-card--disabled {
     background-color: var(--token-color-surface-interactive-disabled);
     border-color: var(--token-color-border-primary);
     box-shadow: none;
--- a/packages/components/src/styles/components/form/radio-card.scss
+++ b/packages/components/src/styles/components/form/radio-card.scss
@@ -61,8 +61,7 @@
     box-shadow: 0 0 0 3px var(--token-color-focus-action-external);
   }
 
-  &--checked,
-  &.mock-checked {
+  &.hds-form-radio-card--checked {
     border-color: var(--token-color-palette-blue-300);
 
     .hds-form-radio-card__control-wrapper {
@@ -90,6 +89,11 @@
 
     .hds-form-radio-card__content {
       opacity: 0.5;
+    }
+
+    &:hover,
+    &.mock-hover {
+      border-color: var(--token-color-border-primary);
     }
   }
 }
--- a/.changeset/curly-dogs-greet.md
+++ b/.changeset/curly-dogs-greet.md
@@ -1,0 +1,7 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+<!-- START components/form/radio-card -->
+`Form::RadioCard` - Fixed styling bug where hover styles were visible when disabled
+<!-- END -->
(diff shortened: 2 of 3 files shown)


shleewhite (approved):
Nice! Such a quick fix!

packages/components/src/styles/components/form/radio-card.scss
```diff
@@ -76,8 +76,7 @@
     }
   }
 
-  &--disabled,
-  &.mock-disabled {
+  &.hds-form-radio-card--disabled {
```
shleewhite, line 79:
```suggestion
  // note: using full class name so it overrides the :hover styles
  &.hds-form-radio-card--disabled {
```

packages/components/src/styles/components/form/radio-card.scss
```diff
@@ -76,8 +76,8 @@
     }
   }
 
-  &--disabled,
```
didoo, line 79: [note] yes, this is an anti-pattern for us (HDS) so that's why it caused the bug (unfortunately, it went undetected). I have noticed that there is a similar one for `&--checked` in this file; should we change that one too, wdyt?

didoo (dismissed):
Left only a note. Nice PR! (and thanks for the quick fix) 🙌

packages/components/src/styles/components/form/radio-card.scss
dchyun: Yeah I went back and forth on that because changing that selector caused some checked styles to leak into the disabled state for the `border-color` on hover. Just pushed an update to make this change. I added in some styles into the disabled state to account for it.

Also unrelated, but removed the `mock-checked` selector as it was unused.

didoo: Nice commit!

didoo (approved):
nice PR! 🙌

zamoore:
@dchyun Looks good. Happy to approve if you want to resolve the conflict.

annawanggg:
hey! in the after video, it seems like the not allowed icon on the cursor might not be showing up when hovering on the disabled card-- perhaps this has been fixed already, but wanted to bring it up in case!

dchyun:
> hey! in the after video, it seems like the not allowed icon on the cursor might not be showing up when hovering on the disabled card-- perhaps this has been fixed already, but wanted to bring it up in case!

@annawanggg ooh great catch! Not sure why the video did that. I do see the icon appearing in the [showcase preview](https://hds-showcase-git-dchyun-radio-card-disabled-selector-hashicorp.vercel.app/components/form/radio-card). Would you mind taking a look on your side to make sure you see it there as well?

annawanggg:
> > hey! in the after video, it seems like the not allowed icon on the cursor might not be showing up when hovering on the disabled card-- perhaps this has been fixed already, but wanted to bring it up in case!
> 
> @annawanggg ooh great catch! Not sure why the video did that. I do see the icon appearing in the [showcase preview](https://hds-showcase-git-dchyun-radio-card-disabled-selector-hashicorp.vercel.app/components/form/radio-card). Would you mind taking a look on your side to make sure you see it there as well?

@dchyun sorry this is late, I was out this morning! but yes it looks good to me there :)

PR #3013, opened 2025-07-15
`Web docs` Spacing docs migration

majedelass:
### :pushpin: Summary

If merged, this PR will migrate all mentions of "Spacing" from the Guidelines tab to the Specifications tab.

The specific docs that this PR affects:

- Breadcrumbs
- Card
- Copy Button
- Form Layout
- Standalone Link
- Pagination
- Reveal
- RichTooltip
- Separator
- Stepper Nav
- Tabs
- Tag
- Toast
- Tooltip

### :link: External links

Jira ticket: [HDS-4715](https://hashicorp.atlassian.net/browse/HDS-4715)

***


website/docs/components/copy/button/partials/specifications/spacing.md
```diff
@@ -0,0 +1,5 @@
+## Spacing
+
+When wrapping an input control and CopyButton in a container, set the gap between each element to 8px.
```
jorytindall: I'd argue that this is different from the other specifications and that this not purely informational because it's the responsibility of the consumer to do this.

website/docs/components/link/standalone/partials/specifications/spacing.md
```diff
@@ -0,0 +1,20 @@
+## Spacing
```
jorytindall: This also seems like guidance to me.

website/docs/components/reveal/partials/specifications/spacing.md
```diff
@@ -0,0 +1,16 @@
+## Spacing
```
jorytindall: This also seems like guidance, especially when the content says "we recommend"

jorytindall (approved):
I won't block on it (and I'm not sure I remember when this conversation was had), but I'm not sure I agree with all of these (I've called some of them out, but not all). There _are_ spacing specifications that are not determined by consumers and are not the responsibility of them to implement; e.g., Breadcrumb, Tooltip/RichTooltip. But there are many of these where it _is_ guidance because a consumer has to implement it and we use the language "we recommend". If we determine it's better to have all of the spacing-related content in the same tab, then that's fine, but I don't think the premise that this is purely informational is necessarily true in all cases.

heatherlarsen:
I agree with what @jorytindall noted above. The overall intention of this update is to move Spacing documentation where predefined spacing is present in the component, rather than when spacing is documented as a guideline that consumers need to implement themselves. There may be situations when both are needed, as well.

majedelass:
@heatherlarsen @jorytindall all fair. Originally I was thinking: should all spacing be moved so that a consumer knows that ALL spacing is in a specific space, regardless if it's a consumer implementation or not, but this makes sense. 

The way I see it (in alignment with what folks are saying) is that it should follow how "States" is represented. It's not a consumer needed implementation, but they should be generally aware of this.

Thanks y'all ill update.

didoo (approved):
👍

PR #3016, opened 2025-07-15
Showcase - Convert Foundations, Layouts, Utilities pages to TS

dchyun:
### :pushpin: Summary

If merged, this PR would convert the Foundations, Layouts, and Overrides showcase pages to TS.

Summary of pages converted

- Foundations
  - Elevation
  - Focus ring
  - Typography
- Layouts
  - App Frame
  - Flex
  - Grid
- Overrides
  - Power Select

### :link: External links

***

<details>

</details>


showcase/app/controllers/page-foundations/typography.js
dchyun, line 1: [Note] All of the functionality of this file could be handled in a route file alone, so deleted this file and move its functionality to the route.

showcase/app/controllers/page-foundations/typography.js
```diff
@@ -1,30 +0,0 @@
...
-} from '../../routes/page-foundations/typography';
-
-export default class PageFoundationsTypographyController extends Controller {
-  get families() {
-    return [...FONT_FAMILIES];
-  }
-  get weights() {
-    return [...FONT_WEIGHTS];
-  }
-  get styles() {
```
dchyun, line 24: [Note] This variable, and the imports that make it up were unused in the template file so it was removed.

showcase/app/templates/page-layouts/flex.hbs
```diff
@@ -236,7 +234,7 @@
       <Hds::Layout::Grid @gap="48" @align="center" as |LG|>
         <Shw::Placeholder @text="item #1" @height="48" />
         <Shw::Placeholder @text="item #2" @height="48" />
-        <LG.Item @basis="25%" @grow="0" @shrink="0">
+        <LG.Item>
```
dchyun, line 237: [Note] The `basis`, `grow` and `shrink` arguments are not present in the `Layout::Grid` `Item` subcomponent.

showcase/app/templates/page-layouts/grid.hbs
```diff
@@ -199,7 +197,7 @@
       <Hds::Layout::Grid @gap="16" @align="center" as |LG|>
         <Shw::Placeholder @text="item #1" @height="48" />
         <Shw::Placeholder @text="item #2" @height="48" />
-        <LG.Item @columnWidth="25%">
+        <LG.Item>
```
dchyun, line 200: [Note] The `columnWidth` argument is not present on the `Layout::Grid` `Item` subcomponent

showcase/app/templates/page-layouts/grid.hbs
```diff
@@ -424,10 +422,10 @@
 <section data-test-percy>
   <Shw::Text::H2>Examples</Shw::Text::H2>
 
-  <Shw::Grid @columns="1" @gap="2rem" as |SG|>
+  <Shw::Grid @columns={{1}} @gap="2rem" as |SG|>
     <SG.Item @label="3 column layout Card layout using columnWidth=33.33%">
       <Hds::Layout::Grid @columnWidth="33.33%" @gap="32">
-        <Hds::Card::Container @level="mid" @hasBorder={{true}} {{style padding="24px"}}>
+        <Hds::Card::Container @level={{@model.HdsCardLevelValues.Mid}} @hasBorder={{true}} {{style padding="24px"}}>
```
dchyun, line 428: [Note] Type errors were thrown when leaving this value as a string. Instead the card level values had to be added to the model and used here. Similar type errors are not thrown for other component values used. 

Ex: When the `Badge` is used below, there is no type error thrown when `@color="success"` is set. 

This is due to a difference in how the types for these two arrguments are defined. 

`HdsCardLevel` is defined using the union type
```
export type HdsCardLevel =
  | HdsCardLevelValues.Base
  | HdsCardLevelValues.Mid
  | HdsCardLevelValues.High;
``` 
`HdsBadgeColors` is defined using template literals
```
export type HdsBadgeColors = `${HdsBadgeColorValues}`;
```

The template literal method allows you to use raw strings directly, meanwhile the union method means you have to use the exact enum values. 

Ideally we should have some alignment on how these types are set. That could be a future alignment effort. My preference would be to use the template literal method so we can avoid having to do things like this in the template files.

showcase/types/global.d.ts
```diff
@@ -24,7 +25,8 @@ declare module '@glint/environment-ember-loose/registry' {
       PageTitle,
       RenderModifiersRegistry,
       ShowcaseTemplateRegistry,
-      EmberComposableHelpersRegistry {
+      EmberComposableHelpersRegistry,
+      PowerSelectRegistry {
```
dchyun, line 29: [Note] Without adding this to the registry template errors were thrown in the Power Select saying `PowerSelect` was not defined.

showcase/app/routes/page-foundations.ts
dchyun, line 1: [Note] The diff can make this a little unclear, but the `FocusRing` route was removed, as it was not needed. A base level `Foundations` route file was separately added as it was missing previously. All other top level routes (components, overrides, utilities, etc) have route files.

showcase/app/controllers/page-foundations/typography.js
didoo: Not sure why I implemented things in this way, tbh 🤷‍♂️🤦‍♂️
https://github.com/hashicorp/design-system/commit/3ff5ddfaf81e8ec636cbabd7865db1ea188eeec9

showcase/app/routes/page-layouts/flex.js
didoo, line 1: [question] @dchyun @shleewhite I see that in some cases we do something like this https://github.com/hashicorp/design-system/blob/main/showcase/app/routes/page-components/app-header/index.ts#L11-L14 or in the file https://github.com/hashicorp/design-system/pull/3016/files#diff-0e413b1ec5714e72a22f78b4ae66461ac358dc7e9cedfb25373dfabee45354a1 in this PR, and in others like this file we don't. When is it necessary and when it's not? It's necessary only when we import some constants from the `@hashicorp/design-system-components` package? It doesn't seem to be the case, since this one is not https://github.com/hashicorp/design-system/pull/3016/files#diff-83fe2fa87a9aadbe63b8ab325f6004af06e3296389636fcd1878123c28861a13. 

Sorry, but I struggle to see the pattern here (and the "rule" to follow)

showcase/app/templates/page-layouts/flex.hbs
didoo: Yeah, probably remnants of a copy&paste from a similar example in the layout flex page.

didoo (commented):
left a few comments

showcase/app/templates/page-layouts/grid.hbs
shleewhite: Totally agree - would you mind adding at ticket in the backlog where we can note all the components that have issues like this so we can fix them all at once?

showcase/app/routes/page-foundations.ts
shleewhite: Makes sense! Thank you for clarifying.

showcase/app/routes/page-layouts/flex.js
shleewhite: The rule is, if there is a controller file you absolutely have to export the model type. 

Basically if there is a controller file you have to declare that the model exists in the controller for the types to work. Glint can implicitly understand that the model is related to the template if there is no controller, but if there is a controller file that implicit understanding doesnt work anymore and you have to declare that the model exists and what its type is in the controller.

shleewhite: I have been exporting the model type all the time now, but that is why you may see it in some routes and not others.

I am hoping that when I spike on a way to break up the template files,  we wont need the controllers anymore and then this confusion will not happen anymore.

dchyun: For all of the route files I'm converting so far I've been adding in the model type, whether they have controllers or not. In this PR the flex route does have the model type defined.

showcase/app/templates/page-layouts/grid.hbs
dchyun: Created a ticket for this in the epic. https://hashicorp.atlassian.net/browse/HDS-5154

showcase/app/routes/page-layouts/flex.js
didoo: OK, let's review this "standard" later then.

showcase/app/templates/page-layouts/flex.hbs
```diff
@@ -557,9 +555,8 @@
     <SF.Item
       @label="centered content"
       class="shw-layout-flex-example-outline-flex-container shw-layout-flex-example-tint-flex-items"
-      @tag="section"
     >
-      <Hds::Layout::Flex @justify="center" @align="center" {{style width="300px" height="200px"}}>
+      <Hds::Layout::Flex @justify="center" @align="center" @tag="section" {{style width="300px" height="200px"}}>
```
didoo: @dchyun something strange happened here: now the Percy test is failing, but if I look at the [showcase preview of your branch](https://hds-showcase-git-dchyun-showcase-overrides-ts-8b2975-hashicorp.vercel.app/layouts/flex#examples), is correct. Maybe worth to remove the `@tag="section"` and see if it still fails?

screenshot_5175
screenshot_5174

dchyun: Yeah I noticed the same thing which was weird. I tried rerunning the percy generation yesterday and it still gave the diff, even though the showcase preview looks fine. I made your suggested fix. Will see if that helps.

dchyun: Percy seems to be happy now!

didoo (approved):
LGTM!

PR #3028, opened 2025-07-16
Update @ember/string to 4.x

RobbieTheWagner:
### :pushpin: Summary

If merged, this PR would update @ember/string to 4.x, which fixes some issues with vite compatibility.

### :link: External links

Figma file: [if it applies]

***

<details>

</details>

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -46,7 +46,6 @@
     "@codemirror/state": "^6.5.0",
     "@codemirror/view": "^6.36.2",
     "@ember/render-modifiers": "^2.1.0",
-    "@ember/string": "^3.1.1",
     "@ember/test-waiters": "^3.1.0",
     "@embroider/addon-shim": "^1.10.0",
     "@embroider/macros": "^1.18.0",
@@ -81,6 +80,7 @@
     "@babel/eslint-parser": "^7.27.1",
     "@babel/plugin-transform-typescript": "^7.27.1",
     "@babel/runtime": "^7.27.1",
+    "@ember/string": "^4.0.0",
     "@ember/test-helpers": "^4.0.4",
     "@embroider/addon-dev": "^7.1.4",
     "@eslint/js": "^9.23.0",
@@ -120,6 +120,7 @@
     "webpack": "^5.97.1"
   },
   "peerDependencies": {
+    "@ember/string": ">= 3.1.1",
     "ember-engines": ">= 0.11.0"
   },
   "peerDependenciesMeta": {
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -105,9 +105,6 @@ importers:
       '@ember/render-modifiers':
         specifier: ^2.1.0
         version: 2.1.0(@babel/core@7.27.1)(@glint/template@1.5.2)(ember-source@6.4.0(@glimmer/component@1.1.2(@babel/core@7.27.1))(rsvp@4.8.5))
-      '@ember/string':
-        specifier: ^3.1.1
-        version: 3.1.1
       '@ember/test-waiters':
         specifier: ^3.1.0
         version: 3.1.0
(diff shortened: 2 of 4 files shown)


aklkv:
may I suggest converting this to draft while you figure things out with ember-cli-string-helpers  + add a bit of background

showcase/types/global.d.ts
```diff
@@ -27,7 +27,7 @@ declare module '@glint/environment-ember-loose/registry' {
       ShowcaseTemplateRegistry,
       EmberComposableHelpersRegistry,
       PowerSelectRegistry {
-    // TODO: `ember-cli-string-helpers` is deprecated and doesn't provide own types, we need to move to `@ember/string` [HDS-4973](https://hashicorp.atlassian.net/browse/HDS-4973)
+    // TODO: `ember-cli-string-helpers` is no longer deprecated and we should add types to it.
```
aklkv: once I merge my [PR](https://github.com/adopted-ember-addons/ember-cli-string-helpers/pull/352)  we would follow back up with a actual template registry

packages/components/package.json
```diff
@@ -60,6 +59,7 @@
     "codemirror-lang-hcl": "^0.0.0-beta.2",
     "decorator-transforms": "^2.3.0",
     "ember-a11y-refocus": "^4.1.4",
+    "ember-auto-import": "2.10.0",
```
aklkv: we will have to remove this once we convert ember-cli-string-helpers to v2 format

aklkv (dismissed):

aklkv:
Not only that but it seems that percy isn't fond of this change

didoo:
> percy isn't fond of this change

It just needs to be rebased

showcase/types/global.d.ts
RobbieTheWagner: Done

packages/components/package.json
RobbieTheWagner: I think it is required for importing from `@ember/string`. We do not seem to be doing that, but if we wanted to, it might be nice to keep this.

RobbieTheWagner:
@didoo I have rebased and created the changeset. Please let me know if there is anything else I need to do. Thanks!

packages/components/package.json
aklkv: Ut this is rollup build package ember-auto-import shouldn't be needed here at all

packages/components/package.json
```diff
@@ -120,6 +121,7 @@
     "webpack": "^5.97.1"
   },
   "peerDependencies": {
+    "@ember/string": ">= 3.1.1",
```
aklkv: I ma not a big fan of adding this particular peerDep, @didoo I think once work on moving to SFCs we would be able to remove this as it would he much easier to use off the shelf library for string manipulation. For now to solve blocker that @RobbieTheWagner is facing I think it's fine

aklkv:
waiting for minor types fix https://github.com/adopted-ember-addons/ember-cli-string-helpers/pull/379

alex-ju (commented):
A few considerations on this PR:
 - the current description doesn't provide much context around why this change is necessary. adding links to related PRs, slack discussions, or an explanation of the motivation would help reviewers better understand the reasoning behind these changes
 - since changelogs are meant to reflect changes relevant to consumers of our packages (in this case, `@hashicorp/design-system-components`), a more accurate entry for the current changes might be: _Moved `@ember/string` to `peerDependencies` and added `ember-auto-import 2.10.0`_
 - note that `ember-cli-string-helpers` is only introduced in `showcase` and `website`, so it doesn't need to be mentioned in the changelog
 - from what I understand, it's non-standard for a v2 addon to include `ember-auto-import` as a direct dependency especially with Embroider, where this is typically part of the build pipeline. is there a specific reason it needs to be included here?

aklkv:
@alex-ju ember-auto-import will be removed

RobbieTheWagner:
> @alex-ju ember-auto-import will be removed

ember-auto-import has now been removed

packages/components/package.json
RobbieTheWagner: Removed 👍

RobbieTheWagner:
@alex-ju @aklkv 

> note that ember-cli-string-helpers is only introduced in showcase and website, so it doesn't need to be mentioned in the changelog

So I actually think `@ember/string` can just be a devDep then, if we are not shipping `ember-cli-string-helpers` or `@ember/sring` related things to consumers of HDS. 

Therefore, I think this is all internal and should have no impact on consumers, with the exception that we were previously shipping `@ember/string` 3.x to consumers as a dependency and we should not have been as it breaks embroider.

Should the changelog then be something like `Removed unnecessary dependency on @ember/string`?

RobbieTheWagner:
Actually, it looks like we need to ship `@ember/string` for `ember-style-modifier`, so I would propose we use their same `peerDep` strategy. I will make those tweaks.

RobbieTheWagner:
@aklkv @alex-ju @didoo I have added to the PR description a better explanation of what we are doing here and why. I do not have links to documentation as to why or Slack conversations or anything, but hopefully the description here will make sense and suffice. I will update the changelog entry as well.

aklkv (approved):

.changeset/eight-wombats-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+Make `@ember/string` a peerDependency to allow consuming apps to choose to use `3.x` or `4.x`. This unblocks apps that need `4.x` for vite compatibility, while retaining `3.x` support as well for older apps.
```
didoo, line 5: [praise] great explanation of what was changed and why 👏 👏 👏

didoo (approved):
👍

alex-ju (approved):

PR #3026, opened 2025-07-16
`AppSideNav::List:Link` - Fix styles when rendered as `button`

didoo:
### :pushpin: Summary

While working on https://github.com/hashicorp/design-system/pull/3025 I noticed that when the `AppSideNav::List:Link` is rendered as a `<button>`, it inherits some of the styles from the HTML element itself. 

modified

This PR addresses this issue.

### :hammer_and_wrench: Detailed description

In this PR I have:
- applied explicit background color to `.hds-app-side-nav__list-item-link` element
    - this issue didn't happen in the old `.hds-side-nav__list-item-link` element because in that case a `background-color` was specifically assigned
- removed invisible border on `.hds-[app-]side-nav__list-item-link` elements
- updated showcase pages to include visual testing for use case where `.hds-[app-]side-nav__list-item-link` is a button

### :camera_flash: Screenshots

**Before:**
screenshot_5172

**After:**
screenshot_5170

### :link: External links

Jira ticket: 

***

### :eyes: Component checklist

- [x] Percy was checked for any visual regression
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

<details>

<summary>:clipboard: PCI review checklist</summary>

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.
- [ ] If applicable, I've worked with GRC to document the impact of any changes to security controls.
  Examples of changes to controls include access controls, encryption, logging, etc.
- [ ] If applicable, I've worked with GRC to ensure compliance due to a significant change to the in-scope PCI environment.
  Examples include changes to operating systems, ports, protocols, services, cryptography-related components, PII processing code, etc.

</details>

PR #3032, opened 2025-07-18
`AppSideNav::List:Link` - Fix background issue

didoo:
### :pushpin: Summary

This is a follow-up of #3026. 

I've noticed that the focused state of the `AppSideNav::List:Link` items was covered by the following item in the list:

screenshot_5184

The reason is that in #3026 I have applied an explicit background to it, the same color as the backdrop of their container. 

This issue didn't happen in the previous implementation (`SideNav::List:Link`):
screenshot_5185

The reason for this difference is that before the focus was applied via a pseudo-element, so the `z-index` would take care of the stacking, while in the new implementation of the side navigation the focus is applied as box shadow.

This PR fixes the issue.

In this PR I have:
- updated the showcase for the status of `AppSideNav::List:Link` to make the problem visible
- set a transparent background color for `.hds-app-side-nav__list-item-link`

👉 👉 👉 **Preview**: https://hds-showcase-git-sidenav-list-link-fix-backgro-cf8dad-hashicorp.vercel.app/components/app-side-nav#states

**Before:** 
screenshot_5186

**After:** 
screenshot_5187

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-5153

***

<details>

</details>

### DIFF
--- a/.changeset/fuzzy-pumpkins-fix.md
+++ b/.changeset/fuzzy-pumpkins-fix.md
@@ -1,0 +1,7 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+<!-- START components/app-side-nav-->
+`AppSideNav::List::Link` - Applied transparent background to the element to avoid ovelapping with previous item's focus ring
+<!-- END -->
--- a/.changeset/fuzzy-pumpkins-fix.md
+++ b/.changeset/fuzzy-pumpkins-fix.md
@@ -3,5 +3,5 @@
 ---
 
 <!-- START components/app-side-nav-->
-`AppSideNav::List::Link` - Applied transparent background to the element to avoid ovelapping with previous item's focus ring
+`AppSideNav::List::Link` - Applied transparent background to the element to avoid overlapping with previous item's focus ring
 <!-- END -->
--- a/packages/components/src/styles/components/app-side-nav/content.scss
+++ b/packages/components/src/styles/components/app-side-nav/content.scss
@@ -70,7 +70,6 @@
 
 // <li> element
 .hds-app-side-nav__list-item {
-  position: relative; // for the "active" state indicator
   list-style-type: none;
 
   & + & {
@@ -99,6 +98,8 @@
   &:focus,
   &.mock-focus {
     @include hds-focus-ring-with-pseudo-element();
+    // to move the item link above the siblings elements when they have a solid background (eg. active) so that the focus ring is not cut out
+    z-index: 1;
   }
 
   // :Hover
@@ -130,6 +131,7 @@
   // Important: This element does not doing anything when interacted with so should not change color according to state
   &.active,
   &.active:hover:focus {
+    position: relative; // for the "active" indicator bar to be correctly positioned
     background: var(--token-color-surface-strong);
 
     // indicator bar
--- a/packages/components/src/styles/components/app-side-nav/content.scss
+++ b/packages/components/src/styles/components/app-side-nav/content.scss
@@ -89,8 +89,8 @@
     var(--token-app-side-nav-body-list-item-padding-horizontal);
   color: var(--token-app-side-nav-color-foreground-primary);
   text-decoration: none;
-  background: var(--token-app-side-nav-color-surface-primary);
-  // "Link" could render as an HTML button element so this overrides the default border style in that case:
+  // "Link" could render as an HTML button element so these override the default background and border styles in that case
+  background: transparent;
   border-color: transparent;
   border-width: 0;
   border-radius: var(--token-app-side-nav-body-list-item-border-radius);
--- a/showcase/app/components/mock/app/sidebar/app-side-nav.gts
+++ b/showcase/app/components/mock/app/sidebar/app-side-nav.gts
@@ -4,11 +4,15 @@
  */
 
 import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { on } from '@ember/modifier';
+import style from 'ember-style-modifier/modifiers/style';
 
 // HDS components
 import {
   HdsAppSideNav,
   HdsAppSideNavList,
+  HdsFormToggleField,
 } from '@hashicorp/design-system-components/components';
 
 // types
@@ -19,8 +23,7 @@ export interface MockAppSidebarAppSideNavSignature {
   Args: {
     isResponsive?: HdsAppSideNavSignature['Args']['isResponsive'];
     isCollapsible?: HdsAppSideNavSignature['Args']['isCollapsible'];
-    showHeader?: boolean;
-    showFooter?: boolean;
+    showDevToggle?: boolean;
   };
   Element: HdsAppSideNavSignature['Element'];
 }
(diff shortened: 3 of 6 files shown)


heatherlarsen (commented):
Great improvement!

I found a couple of instances where the focus ring is still being cut off:

At the bottom of the list...not in all cases, though 🤔
Screenshot 2025-07-18 at 8 32 14 AM

When focus is visible on the list item above an active list item
Screenshot 2025-07-18 at 8 30 40 AM

didoo:
> I found a couple of instances where the focus ring is still being cut off:
> At the bottom of the list...not in all cases, though

@heatherlarsen yes, this is [a known issue](https://hashicorp.slack.com/archives/C05LKJ8LEC9/p1752839264656499?thread_ts=1752533000.110929&cid=C05LKJ8LEC9). While we know what a possible fix may be for it, this PR is only focusing on the background issue

> When focus is visible on the list item above an active list item

yeah, nice catch! there are so many combinations and permutations that it's hard to test them all! I've added more use cases to the showcase page for the AppSideNav, related to this, and I'll move the PR to draft so that @KristinLBradley can have a proper look at it and see if the fixes to this issue are correct, if there are better alternatives, and expecially if we're covering all the use possible cases (there may be more that I have forgotten).

.changeset/fuzzy-pumpkins-fix.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+<!-- START components/app-side-nav-->
+`AppSideNav::List::Link` - Applied transparent background to the element to avoid ovelapping with previous item's focus ring
```
KristinLBradley:
```suggestion
`AppSideNav::List::Link` - Applied transparent background to the element to avoid overlapping with previous item's focus ring
```

KristinLBradley:
From Heather: 
> I found a couple of instances where the focus ring is still being cut off:
> 
> At the bottom of the list...not in all cases, though 🤔 
> 
> When focus is visible on the list item above an active list item 

@heatherlarsen @didoo 

I've been investigating the remaining focus ring issues. For the first one, when panels are used there is an overflow:hidden on the panel that causes the focus ring to be cut off. This was part of a bug fix with scrolling: [Merged PR](https://github.com/hashicorp/design-system/pull/2487), [Jira ticket](https://hashicorp.atlassian.net/browse/HDS-3892)

The safest way to fix this bug (I think) would be to add a bit of top and bottom padding to the panel container to give the focus ring room. This would alter the layout from the AppSideNav's not using panels though. (It would probably be possible to add some tricky CSS to change the overall visual padding dependent of the AppSideNav when it contains a panel vs. not to keep the visual spacing consistent but it would be a touch messy.)

It may also be possible to delete the overflow:hidden style on the panel but that risks re-introducing bugs with scrolling the AppSideNav content and is difficult to test for to account for all possible interactions related to navigating between links.

For the 2nd issue, with active list item backgrounds overlapping the focus ring, I wasn't able to duplicate it so it seems that this issue may already be fixed in this PR?

didoo:
> It may also be possible to delete the overflow:hidden style on the panel but that risks re-introducing bugs with scrolling the AppSideNav content and is difficult to test for to account for all possible interactions related to navigating between links.

@KristinLBradley yes, that was I was hoping you could do or have more context about: do you remember why this fix was introduced, and how to replicate the issue that made us make the fix, so we can see if it's still applicable?
 
> For the 2nd issue, with active list item backgrounds overlapping the focus ring, I wasn't able to duplicate it so it seems that this issue may already be fixed in this PR?

Yes, I would assume this is fixed with this PR, but I need you to double check the proposed solution, because it involves a bit of refactoring of `position` and `z-index` values, so I may have not considered something.

KristinLBradley:
@didoo I checked out your branch and linked to it within Cloud-ui to test. When running cloud-ui locally I'm seeing a type error (I haven't investigated much yet but just wanted to let you know in case you had context).

Error:
```
error TS2353: Object literal may only specify known properties, and 'ariaLabel' does not exist in type 'NamedArgs<{ href?: string | undefined; isHrefExternal?: boolean | undefined; isRouteExternal?: boolean | undefined; route?: string | undefined; models?: unknown[] | undefined; model?: unknown; query?: Record<string, unknown> | undefined; 'current-when'?: string | ... 1 more ... | undefined; replace?: boolean | undef...'.
core:build:v2: 
core:build:v2: 169                   @ariaLabel={{t "nav.logo-aria-label"}}
```

KristinLBradley:
> > It may also be possible to delete the overflow:hidden style on the panel but that risks re-introducing bugs with scrolling the AppSideNav content and is difficult to test for to account for all possible interactions related to navigating between links.
> 
> @KristinLBradley yes, that was I was hoping you could do or have more context about: do you remember why this fix was introduced, and how to replicate the issue that made us make the fix, so we can see if it's still applicable?
> 
> > For the 2nd issue, with active list item backgrounds overlapping the focus ring, I wasn't able to duplicate it so it seems that this issue may already be fixed in this PR?
> 
> Yes, I would assume this is fixed with this PR, but I need you to double check the proposed solution, because it involves a bit of refactoring of `position` and `z-index` values, so I may have not considered something.

@didoo I checked out this PR branch locally and also removed the `overflow: hidden` style from `.hds-side-nav__content-panel` then linked to the local files and tried testing within both cloud-ui & atlas. I navigated around and tried focusing various links. I wasn't able to reproduce the original issue so as far as I can tell it's not needed. Terraform/atlas has a custom overflow: hidden style on an outer ".sidebar-scrollbar" container however which still cuts off the focus ring on focused links at the top & bottom of the AppSideNav. (I tried also commenting out this custom style and then navigated around & focused the AppSideNav links and wasn't able to detect issues.)

I'll continue testing your other solution for the 2nd issue to see if I encounter issues but so far I'm not seeing any.

packages/components/src/styles/components/app-side-nav/content.scss
```diff
@@ -130,6 +131,7 @@
   // Important: This element does not doing anything when interacted with so should not change color according to state
   &.active,
   &.active:hover:focus {
+    position: relative; // for the "active" indicator bar to be correctly positioned
```
KristinLBradley, line 134: So far, I haven't been able to uncover issues related to this change and the z-index above.

didoo: Nice!

didoo:
@KristinLBradley I have moved the PR out of draft, please review. 

Note: this PR fixes only the background issue; for the overflow issue you should open a distinct PR so we have better traceability of the changes we're introducing. As soon as it's open, I can review it.

KristinLBradley (approved):

KristinLBradley:
> @KristinLBradley I have moved the PR out of draft, please review.
> 
> Note: this PR fixes only the background issue; for the overflow issue you should open a distinct PR so we have better traceability of the changes we're introducing. As soon as it's open, I can review it.

@didoo Should I create a Jira ticket or is it ok to open a PR directly?

PR #3033, opened 2025-07-18
Tag - Fix performance issues with single ResizeObserver 

meirish:
### :pushpin: Summary

This PR changes `Hds::Tag` to use a single `ResizeObserver` for all instances of `Hds:Tag`. This drastically improves performance when rendering the component in large arrays.

:wave: This stemmed from an issue reported in [Slack](https://hashicorp.slack.com/archives/C7KTUHNUS/p1752763225377749) where users noted that when rendering a decently sized array of strings, changing from `Hds::Tag` to `Hds::Badge` drastically reduced the rendering time (60s to 5s - note: this also included all of the network calls, etc.).

Looking at the difference between those two components, the major difference is that the `Hds::Tag` adds a new `ResizeObserver` for each instance of the component. Refactoring the component so that only a single `ResizeObserver` is needed for all instances greatly improves the initial rendering performance.

NOTE: The best way to test this is to drop all but the first commit, and load up the `Tag` page in the showcase app which now renders a large list of tags. This will show the slow loading issue with the current implementation. Then adding in the second commit (or looking at the preview deploy) will show how the refactor improves the performance of the large list.

SECOND NOTE (😂): The first commit should be dropped if / when you want to merge this.

#### Tag showcase page loading time w/ new example

**Before**
Screenshot 2025-07-21 at 9 09 25 AM

**After**
Screenshot 2025-07-21 at 9 08 21 AM

### :link: External links

Figma file: [if it applies]

***

<details>

</details>


packages/components/src/components/hds/tag/index.ts
```diff
@@ -33,23 +34,44 @@ export interface HdsTagSignature {
   Element: HTMLSpanElement;
 }
 
+const overflowed = new TrackedWeakSet<Element>();
```
meirish, line 37: This is the key to getting this working - externalizing the observer is only possible if we can track what elements are overflowed. To do this with ember's tracking, `tracked-built-ins` provides a tracking-entangled `WeakSet`. This means that getters and other computed properties that rely on this WeakSet will recompute when items are added or removed.

meirish:
I think the Percy failure should go away once the showcase app commit is dropped.

showcase/app/routes/page-components/tag.js
```diff
@@ -11,6 +11,9 @@ export default class PageComponentsTagRoute extends Route {
   model() {
     // these are used only for presentation purpose in the showcase
     const STATES = ['default', 'hover', 'active', 'focus'];
-    return { COLORS, TOOLTIP_PLACEMENTS, STATES };
+    const range = Array(1000)
+      .fill(1)
+      .map((n, i) => ({ index: n + i }));
+    return { range, COLORS, TOOLTIP_PLACEMENTS, STATES };
```
didoo: @dchyun we could consider preserving a similar example in the showcase page (with a container with a `overflow: scroll` to avoid filling the whole page? or in a separate mock/frameless page within an iframe?)

didoo:
@dchyun I've added you as reviewer, since you have more context about this feature.

packages/components/src/components/hds/tag/index.ts
```diff
@@ -33,23 +34,44 @@ export interface HdsTagSignature {
...
-        );
-      });
-    });
-    this._observer.observe(element);
+    this._element = element;
+    observer.observe(element);
 
     return () => {
-      this._observer.disconnect();
+      delete this._element;
```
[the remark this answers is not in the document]
dchyun: In our _setUp functions in other components where we set `this._element` we don't explicitly uncouple the variable name so this may not be necessary to add.

packages/components/src/components/hds/tag/index.ts
```diff
@@ -33,23 +34,44 @@ export interface HdsTagSignature {
...
 }
 
+const overflowed = new TrackedWeakSet<Element>();
+
+const observer = new ResizeObserver((entries) => {
+  entries.forEach((entry) => {
+    if (overflowed.has(entry.target)) {
+      return;
+    }
+
```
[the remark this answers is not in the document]
dchyun: The above statement is correct. This needs to be deleted in order for the overflow tooltip to be removed if a tag goes from overflowed to not overflowed. 

Right now if a tag is overflowed, which means it is in `overflowed` it never makes it to the `overflowed.delete(entry.target)` statement below.

showcase/app/routes/page-components/tag.js
dchyun: Yeah I agree. I think this would be a good use case for the frameless iframe.

packages/components/src/components/hds/tag/index.ts
dchyun: Checked the performance times for before and after and this fix correct the issues greatly. Will add some screenshots to the PR description for visibility.

packages/components/src/components/hds/tag/index.ts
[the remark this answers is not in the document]
meirish: I disagree with Copilot here - think delete or setting null / undefined are both accepted ways of clearing a reference. 

@dchyun if you're setting a reference to the element, but never clearing it, that could potentially cause memory leaks.

meirish:
> @meirish thank you for catching this issues, and coming up with this solution. It looks like a great solution to the problem to me. If you would like I can pick this up and address the feedback, and make the showcase changes @didoo mentioned previously.

@dchyun thanks for taking a look! And feel free to pick it up - let me know if I can help or if you have other questions.

packages/components/src/components/hds/tag/index.ts
```diff
@@ -33,23 +34,44 @@ export interface HdsTagSignature {
...
-      });
-    });
-    this._observer.observe(element);
+    this._element = element;
+    observer.observe(element);
 
     return () => {
-      this._observer.disconnect();
+      delete this._element;
+      observer.disconnect();
```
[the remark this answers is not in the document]
meirish: Oops, this is also correct! We definitely don't want `disconnect` here ([docs](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#instance_methods)) - sorry about that!

showcase/app/routes/page-components/tag.js
meirish: `overflow:scroll` with the tooltip might look strange - maybe putting it in an accordion or `<details>` to hide / show it would work better?

dchyun: I've added a new [demo section](https://hds-showcase-81psnn93r-hashicorp.vercel.app/components/tag#demos) to the tag page, and added an iframe with a demo.

meirish:
OK I take back what I said about the overflow being weird, that demo / iframe in the showcase app looks great ❤️

.changeset/hungry-meals-serve.md
didoo, line 1: we've also added `tracked-built-ins` as dependency, should we reflect it also this change also in the changelog?

showcase/app/templates/page-components/tag/frameless/demo-observer-performance.hbs
```diff
@@ -0,0 +1,21 @@
...
+<div {{style padding="24px"}}>
+  <Shw::Text::Body><strong>Note:</strong>
+    This demo is to test the loading performance of the ResizeObserver when many Tags are rendered. There should be
+    little increase in loading times even when many Tags with observers are present.</Shw::Text::Body>
+
+  <Hds::Layout::Grid @columnWidth="150px" @gap="16">
+    {{#each @model.DEMO_RANGE as |i|}}
+      <Hds::Tag
+        @text="{{i.index}} This is a very long text that should go on multiple lines"
+        @onDismiss={{this.noop}}
```
didoo: not needed (unless you want to test both variants)
```suggestion
```

showcase/app/templates/page-components/tag/index.hbs
```diff
@@ -208,4 +208,15 @@
     </SF.Item>
   </Shw::Flex>
 
+  <Shw::Divider @level={{2}} />
+
+  <Shw::Text::H2>Demos</Shw::Text::H2>
```
didoo, line 218: [todo] Percy has problems taking screenshots of content inside iframes
```suggestion
</section>

{{! For some reason, Ember tests don't play well with iframes (URL not found) so we can't take snapshots of these examples in Percy }}
<section>
  <Shw::Text::H2>Demos</Shw::Text::H2>
```

didoo (commented):
left a couple of suggestions

.changeset/hungry-meals-serve.md
dchyun: Yeah it should have been added. Added now.

PR #3037, opened 2025-07-22
`Web docs` Card A11Y updates

majedelass:
### :pushpin: Summary

If merged, this PR will address the issues about the interactive card in our web docs.

Vercel preview 👉 https://hds-website-git-card-interactive-doc-update-hashicorp.vercel.app/components/card

Removing contradictions, adding a little more details to help guide consumers to a more accessible, interactive card behavior.

### :link: External links

Jira ticket: [HDS-5066](https://hashicorp.atlassian.net/browse/HDS-5066)

***

[HDS-5066]: https://hashicorp.atlassian.net/browse/HDS-5066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/docs/components/card/partials/guidelines/guidelines.md
+++ b/website/docs/components/card/partials/guidelines/guidelines.md
@@ -14,7 +14,7 @@ Static cards may contain interactive elements but must not be actionable or inte
 
 ### Interactive cards
 
-Interactive cards are actionable, e.g., they can be hovered over, clicked, or focused on. If a card is interactive, it should not include nested interactive elements within it.
+Interactive cards are actionable, e.g., they can be hovered over, clicked, or focused on. 
 
 !!! Dont
 


website/docs/components/card/partials/guidelines/guidelines.md
```diff
@@ -12,7 +12,23 @@ Static cards may have interactive elements within them but are not actionable or
 
 ### Interactive cards
 
-Interactive cards are actionable (e.g., it can be hovered, clicked, focused, etc.)
+Interactive cards are actionable (e.g., it can be hovered, clicked, focused, etc.). Ensure that if your cards link out to other places, do not have additional linkable elements within.
```
heatherlarsen:
```suggestion
Interactive cards are actionable, e.g., they can be hovered over, clicked, or focused on. If a card is interactive, it should not include additional linkable elements within it.
```

website/docs/components/card/partials/guidelines/guidelines.md
```diff
@@ -12,7 +12,23 @@ Static cards may have interactive elements within them but are not actionable or
...
+
+!!! Dont
+
+Don't nest other interactive elements while the card itself is interactive. This is an accessibility violation called a nested interactive.
```
heatherlarsen:
```suggestion
Don't nest interactive elements within an interactive card. This is because screen reader users cannot access nested interactions properly. 
```

website/docs/components/card/partials/code/how-to-use.md
```diff
@@ -44,9 +44,9 @@ Note: If you choose to use the `Card` as a list item, you must wrap it either in
 
 At the moment, we do not recommend using the Card component as an interactive element, although we may add this feature in the future. Despite this, some products have implemented designs that provide visual feedback to the user interacting with a Card by changing the elevation style (on `:hover` or `:active`).
 
-As a stopgap, we have introduced two specific arguments `@levelHover` and `@levelActive` to allow users to declare the specific "level" they want to use for each of these interactive states.
+There are two specific arguments `@levelHover` and `@levelActive` to allow users to declare the specific "level" they want to use for each of these interactive states.
 
-<!-- TODO: this example doesn't work. should we update it or leave out for launch?
+TODO: this example doesn't work. should we update it or leave out for launch?
```
heatherlarsen:
```suggestion
```

heatherlarsen (changes requested):
This doesn't address the issue posed in the Jira ticket, as a discrepancy still exists between what is in the Guidelines and the Code tabs. 

Furthermore, we should evaluate if interactive cards use a specific style vs static cards. Currently, we indicate that they can use elevated cards for both static and interactive cards. This may be the reason why we're seeing so much variation across the products.

website/app/styles/pages/components/card.scss
```diff
@@ -19,4 +19,33 @@
     margin: 0; // styles are being inherited from doc-markdown list styles
     padding: 24px;
   }
+
+  .doc-card-interactive-demo{
```
didoo: [thought] we can't expect consumers to implement this by themselves, it's quite complex. or we provide a suggested approach, in the code snippet, that includes also the CSS on how to achieve this (in which case, we may opt for a simpler solution for them), or we should implement this functionality ourselves (eg. having `@isInteractive={{true}}` or by implementing it in a similar way as it's been implemented in other components, where if consumers provide a `@route` or `@href`, we implicitly assume that it's interactive, change its visual behaviour, and expose the correct APIs to handle the interactivity)

didoo (commented):
Left a high-level comment

website/app/styles/pages/components/card.scss
majedelass: Correct ~ I think this should be a follow up ticket separate from this task. I can create a task if you'd like for this outside of this epic.

didoo: @hashicorp/hds-engineering thoughts?

majedelass: I can include the CSS itself too. How would you propose I do that?

KristinLBradley: Why would consumers need to add so much custom CSS (if that's the intent)? We should update the actual component as needed to work as we recommend (as Cristiano is saying).

Also, the CSS has missing spaces and needs formatting improvements.

didoo: @KristinLBradley spoke with @majedelass about this, in a call; this is what we have agreed/decided:

- this is an interim solution, to provide _something_ to the consumers, so that the documentation around how to use (or not use) the levels/shadows for interactive and non-interactive cards can be updated in a meaningful way
- I made a change to the code "suggested" for the simple use case (the one that doesn't have interactive elements inside the card, so no need to extra pseudo elements or anything like that); can you please review it and see if that clocks with you too, or if you would suggest something different to our consumers?

I have created an idea in Jira to track possible evolutions/extensions of the `Card::Container`: https://hashicorp.atlassian.net/browse/HDSP-195

/cc @MelSumner can you check that what I'm proposing works from an accessibility standpoint? the assumption is that there are no interactive elements inside the `<a>` wrapper

MelSumner (approved):

website/docs/components/card/partials/code/how-to-use.md
```diff
@@ -54,18 +54,15 @@ There are three arguments: `@level`, `@levelHover` and `@levelActive`. These all
 </div>
 ```
 
-where the `.doc-card-interactive-demo` class is something like this:
+Using SCSS, the `.doc-card-interactive-demo` class would look something like this:
 
 ```css
```
didoo: you're right, this is scss, so this should be `scss`

website/public/assets/patterns/table-multi-select/multi-select-in-context-complex-example.png
didoo, line 1: @majedelass do yo know why these images have been updated?

website/docs/components/card/partials/code/how-to-use.md
majedelass: Fixed!

website/public/assets/patterns/table-multi-select/multi-select-in-context-complex-example.png
majedelass: AH SHOOT I was switching between branches and forgot to not update that :/ how do I revert it?

didoo (approved):

website/docs/components/card/partials/guidelines/guidelines.md
```diff
@@ -8,29 +8,45 @@ Use a card container to help build more consistently styled cards.
...
+![A static card titled "Billing summary" with an info notification and a link.](/assets/components/card/card-static-flat-looking.png)
 
 ### Interactive cards
 
-Interactive cards are actionable (e.g., it can be hovered, clicked, focused, etc.)
+Interactive cards are actionable, e.g., they can be hovered over, clicked, or focused on. If a card is interactive, it should not include additional interactive elements within it.
+
+!!! Dont
+
+Don't nest interactive elements within an interactive card. This is because screen reader users cannot access nested interactions properly.
```
LilithJames-HDS:
```suggestion
Don't nest interactive elements within an interactive card. Screen reader users cannot access nested interactions properly. 
```

LilithJames-HDS (dismissed):
One nitpicky non-blocking suggestion but this looks great to me

website/docs/components/card/partials/guidelines/guidelines.md
```diff
@@ -8,29 +8,45 @@ Use a card container to help build more consistently styled cards.
...
 
-Static cards may have interactive elements within them but are not actionable or interactive on their own.
+Static cards may have interactive elements within them but are not actionable or interactive on their own. Card should not have any elevations applied to it to appear visually flat.
+
+![A static card titled "Billing summary" with an info notification and a link.](/assets/components/card/card-static-flat-looking.png)
 
 ### Interactive cards
 
-Interactive cards are actionable (e.g., it can be hovered, clicked, focused, etc.)
+Interactive cards are actionable, e.g., they can be hovered over, clicked, or focused on. If a card is interactive, it should not include additional interactive elements within it.
```
shleewhite:
```suggestion
Interactive cards are actionable, e.g., they can be hovered over, clicked, and focused on. If a card is interactive, it should not include additional interactive elements within it.
```

In order for the interaction to be accessible, it needs to be all 3. Like they can't just apply hover styles bc that implies that it is only usable with a mouse.

website/docs/components/card/partials/code/how-to-use.md
```diff
@@ -42,26 +42,36 @@ Note: If you choose to use the `Card` as a list item, you must wrap it either in
...
-
-<!-- TODO: this example doesn't work. should we update it or leave out for launch?
-
-!!! Warning
-
-This is only an example and not a recommendation. If you have any doubt about which level to use for the different states, please speak with your product designer or contact the Design Systems Team.
-!!!
-
-In the following example, the Card transitions between these elevations _mid → high → mid_ depending on these iteration states _rest → hover → active_:
+There are three arguments: `@level`, `@levelHover` and `@levelActive`. These allow consumers to declare the specific "level" they want to use for each of these interactive states. Following our guidelines, the example transitions between _mid → high → mid_ elevations on these iteration states _default → hover → active_:
```
jorytindall: [suggestion] I _think_ this is a bit more clean, seems easier to read to me, but it's arguably a bit of a tough concept to write about.

```suggestion
The `@level`, `@levelHover`, and `@levelActive` arguments can be used to declare the specific elevation of the Card for each interactive state. Following the guidelines, this example transitions between _mid → high → mid_ elevation for the corresponding interactive states; _default → hover → active_.
```

LilithJames-HDS (approved):

shleewhite (dismissed):

didoo (dismissed):

website/docs/components/card/partials/code/how-to-use.md
```diff
@@ -42,26 +42,36 @@ Note: If you choose to use the `Card` as a list item, you must wrap it either in
...
-    </div>
+      <a href="#">
+        [Your static content here]
+      </a>
   </Hds::Card::Container>
 </div>
 ```
--->
+
+Using SCSS, the `.doc-card-interactive-demo` class would look something like this:
```
jorytindall, line 57: [suggestion] I _think_ you could just say CSS here because this type of nesting is no longer limited to SCSS and it makes the style declaration a but more generic.

website/docs/components/card/partials/guidelines/guidelines.md
```diff
@@ -8,29 +8,45 @@ Use a card container to help build more consistently styled cards.
 
 ### Static cards
 
-Static cards may have interactive elements within them but are not actionable or interactive on their own.
+Static cards may have interactive elements within them but are not actionable or interactive on their own. Card should not have any elevations applied to it to appear visually flat.
```
jorytindall: I'm a little confused by the second sentence here, what do you mean by that?

website/docs/components/card/partials/guidelines/guidelines.md
```diff
@@ -8,29 +8,45 @@ Use a card container to help build more consistently styled cards.
...
+![A static card titled "Billing summary" with an info notification and a link.](/assets/components/card/card-static-flat-looking.png)
 
 ### Interactive cards
 
-Interactive cards are actionable (e.g., it can be hovered, clicked, focused, etc.)
+Interactive cards are actionable, e.g., they can be hovered over, clicked, and focused on. If a card is interactive, it should not include additional interactive elements within it.
+
+!!! Dont
+
+Don't nest interactive elements within an interactive card. Screen reader users cannot access nested interactions properly.
```
jorytindall: [suggestion] This seems a bit more accurate, but open to other suggestions.

```suggestion
Don't nest interactive elements within an interactive card. Assistive technology and keyboard users cannot access nested interactive elements properly. 
```

website/docs/components/card/partials/guidelines/guidelines.md
```diff
@@ -8,29 +8,45 @@ Use a card container to help build more consistently styled cards.
...
 ## Levels
 
 ![example of card styles for various levels, including: base, mid, and high](/assets/components/card/card-levels.png)
 
 ### Base
 
-Use the `base` level for static cards.
+Use for **static cards**.
 
 ### Mid
```
jorytindall, line 43: I'm not sure if we can reasonably exclude consumers from using other levels outside of interactive cards, we would be making a recommendation that is different from how these are implemented in production. A basic example of this is the cards on the HCP dashboard all use the `mid` elevation, but are not interactive.

Maybe I missed this or misunderstood the purpose of this, but I personally still think it's reasonable for consumers to use cards of varying elevations for static cards. But I'm happy to discuss!

website/docs/components/card/partials/code/how-to-use.md
jorytindall: Ah, I see there's another thread about this with @didoo, not opinionated on which route you choose to go!

website/docs/components/card/partials/guidelines/guidelines.md
majedelass: I think instead of using "elevations" as the word, I should use "shadow" instead.

website/docs/components/card/partials/code/how-to-use.md
majedelass: I'll stick with this just to be more specific (because we are actually using SCSS)

website/docs/components/card/partials/guidelines/guidelines.md
majedelass: Thoughts on this?
```suggestion
Static cards may have interactive elements within them but are not actionable or interactive on their own. They should not have any shadows applied to them to appear non-interactive.
```

website/docs/components/card/partials/guidelines/guidelines.md
majedelass: Mmmmm... I would disagree with you on this. Right now the issue with our cards is that they are so heavily varied in how our static and interactive cards are displayed which is causing an accessibility perceivably issue. How would a user know if a card is a link or not if both static and interactive cards can use either of the two styles? Our current guidelines are non-descriptive which has allowed consumers to mix and match. Doing this clearly articulates when to use _which style_ for _which type of card_ so that perceivably is improved here.

website/docs/components/card/partials/code/how-to-use.md
```diff
@@ -42,26 +42,36 @@ Note: If you choose to use the `Card` as a list item, you must wrap it either in
...
+
+```scss
+.doc-card-interactive-demo {
+  .hds-card__container {
+
```
KristinLBradley:
```suggestion
```

website/docs/components/card/partials/guidelines/guidelines.md
MelSumner:
```suggestion
Static cards may contain interactive elements but must not be actionable or interactive on their own. The Card should not have any shadow (CSS) applied to it, so that can be differentiated from an interactive card.
```

This is as clear as I could think to suggest.

KristinLBradley: They should not have any shadows applied to them to appear non-interactive.
=>
They should not have any shadows applied to them _so that they_ appear non-interactive.

KristinLBradley: I think what you have above is better but I think it needs one more tweak for clarity. @majedelass

website/docs/components/card/partials/guidelines/guidelines.md
```diff
@@ -8,29 +8,45 @@ Use a card container to help build more consistently styled cards.
...
 
-Static cards may have interactive elements within them but are not actionable or interactive on their own.
+Static cards may have interactive elements within them but are not actionable or interactive on their own. Card should not have any elevations applied to it to appear visually flat.
+
+![A static card titled "Billing summary" with an info notification and a link.](/assets/components/card/card-static-flat-looking.png)
 
 ### Interactive cards
 
-Interactive cards are actionable (e.g., it can be hovered, clicked, focused, etc.)
+Interactive cards are actionable, e.g., they can be hovered over, clicked, and focused on. If a card is interactive, it should not include additional interactive elements within it.
```
KristinLBradley:
```suggestion
Interactive cards are actionable, e.g., they can be hovered over, clicked, or focused on. If a card is interactive, it should not include nested interactive elements within it.
```

website/docs/components/card/partials/guidelines/guidelines.md
```diff
@@ -8,29 +8,45 @@ Use a card container to help build more consistently styled cards.
...
+
+Don't nest interactive elements within an interactive card. Assistive technology and keyboard users cannot access nested interactive elements properly. 
+
+![An interactive card with a table with links and buttons within causing a nested interactive.](/assets/components/card/card-dont-nest-interactive.png)
+
+!!!
+
+!!! Do
+
+When the card is interactive, ensure elements within are static elements only.
```
KristinLBradley:
```suggestion
When the card is interactive, only static elements should be contained within it.
```

website/docs/components/card/partials/guidelines/guidelines.md
```diff
@@ -8,29 +8,45 @@ Use a card container to help build more consistently styled cards.
...
 ### Base
 
-Use the `base` level for static cards.
+Use for **static cards**.
 
 ### Mid
 
-- For **static cards** use `mid` as the default level.
-- For **interactive cards** use `mid` as the rested state.
+Use for **interactive cards** as the rested (default) and active (pressed) states.
```
KristinLBradley:
```suggestion
Use for the rested (default) and active (pressed) states of **interactive cards**.
```

website/docs/components/card/partials/guidelines/guidelines.md
```diff
@@ -8,29 +8,45 @@ Use a card container to help build more consistently styled cards.
...
 
 ### High
 
-- For **static cards** use `high` for emphasis, but use sparingly.
-- For **interactive cards** use `high` for the raised state.
+Use for **interactive cards** as the raised (hover) state.
```
KristinLBradley:
```suggestion
Use for the raised (hover) state of **interactive cards**.
```

website/docs/components/card/partials/code/how-to-use.md
KristinLBradley: With CSS you'd need to add `&` in front of all the nested parts, so I guess saying SCSS here is clearer and more accurate.

website/docs/components/card/partials/guidelines/guidelines.md
heatherlarsen: +1 to what @majedelass mentioned. It seems like the underlying reason we see variation is due to personal preference, and because we didn't previously have guidance on this. I would love to see more consistent and intentional usage across the products, and I think this guidance helps us get there.

website/docs/components/card/partials/guidelines/guidelines.md
heatherlarsen: @majedelass It looks like this comment got missed.

KristinLBradley (approved):

website/docs/components/card/partials/guidelines/guidelines.md
```diff
@@ -8,29 +8,45 @@ Use a card container to help build more consistently styled cards.
...
+![A static card titled "Billing summary" with an info notification and a link.](/assets/components/card/card-static-flat-looking.png)
 
 ### Interactive cards
 
-Interactive cards are actionable (e.g., it can be hovered, clicked, focused, etc.)
+Interactive cards are actionable, e.g., they can be hovered over, clicked, or focused on. If a card is interactive, it should not include nested interactive elements within it.
+
+!!! Dont
+
+Don't nest interactive elements within an interactive card. Screen reader users cannot access nested interactions properly.
```
heatherlarsen: It sounds like there's a way to build an interactive card with nested interactive elements in it (according to our previous conversation). Should we clarify that we're providing this recommendation for reasons other than accessibility?

majedelass: Technically the other implementation isn't a nested interaction, it just _looks like one_  (and some implementations have downsides to it too). This statement is still accurate. Maybe we can rephrase it to be a little more specific like:

```suggestion
Don't nest interactive elements within a top level interactive card. Screen reader users cannot access nested interactions properly. 
```

What I'm trying to say is if you're going to add a link (`<a href></a>`) to the top level layer then adding other interactive items within should be avoided.

didoo: To me it seems we're conflating two things:
- the "interactive" look&feel of the card (which technically is not interactive, so it has no relevance in terms of accessibility, apart from appearing something you can interact with)
- the nesting of interactive elements (eg. a `<a>` wrapping a `<button>`) but this can be done also in a non-looking-interactive card

Probably I would provide this kind of guidance:
_If you're using an interactive element (e.g., a `<a>` or `<button>` HTML element) inside the `Card::Container` as content wrapper, don't nest other interactive elements within it: screen reader users cannot access nested interactions properly._

didoo: here's the Copilot version:
_If you're using an interactive element (like an `<a>` or `<button>`) as the main content wrapper inside `Card::Container`, avoid placing other interactive elements inside it. This kind of nesting creates accessibility issues—particularly for screen reader users—because nested interactive elements can't be reliably accessed or operated._

zamoore: I think @didoo has it right. The Copilot version makes the most sense to me.

majedelass: This is REALLY well put Cristiano. +1 to how you broke this down. A lot more elegant than I could :)

majedelass: Updated!

website/docs/components/card/partials/guidelines/guidelines.md
KristinLBradley: Those actions don't happen all at the same time. That's why using "or" made sense to me.

website/docs/components/card/partials/guidelines/guidelines.md
```diff
@@ -8,29 +8,45 @@ Use a card container to help build more consistently styled cards.
...
 
-Static cards may have interactive elements within them but are not actionable or interactive on their own.
+Static cards may contain interactive elements but must not be actionable or interactive on their own. They should not have any shadows applied to them, so that they can be differentiated from an interactive card.
+
+![A static card titled "Billing summary" with an info notification and a link.](/assets/components/card/card-static-flat-looking.png)
 
 ### Interactive cards
 
-Interactive cards are actionable (e.g., it can be hovered, clicked, focused, etc.)
+Interactive cards are actionable, e.g., they can be hovered over, clicked, or focused on. If a card is interactive, it should not include nested interactive elements within it.
```
heatherlarsen: [nit] The below "dont" example does a more thorough job of explaining nested interactivity, so I don't think we need it here. 

```suggestion
Interactive cards are actionable, e.g., they can be hovered over, clicked, or focused on. 
```

heatherlarsen (approved):
Looks great!

heatherlarsen (approved):

didoo (approved):

PR #3038, opened 2025-07-22
SPIKE: break up long showcase template files

shleewhite:
### :pushpin: Summary

This is an exploration for what it could look like if the showcase template files were converted to gts and were broken up into sub-components.

Overall, I tried to keep the visual of the showcase the same but there are a couple differences in layout.

So far, pages converted are:
* copy/button

### :link: External links

Jira ticket: [HDS-5077](https://hashicorp.atlassian.net/browse/HDS-5077)

***

<details>

</details>


showcase/app/components/mock/components/copy/button/copy-button-states.gts
```diff
@@ -0,0 +1,69 @@
...
+import {
+  SIZES,
+  SUCCESS_ICON,
+  ERROR_ICON,
+} from '@hashicorp/design-system-components/components/hds/copy/button/index';
+
+const STATES = ['default', 'hover', 'active', 'focus'];
+
+export default class CopyButtonStates extends Component {
+  replaceCopyStatus = modifier((container: HTMLDivElement) => {
```
shleewhite: I opted to change this from a function that runs when the route changes to a modifier for a couple reasons:
1. if I kept it the old way, but moved it into a new component the function would only run if the route changed. if the page refreshed, the function wouldnt run
2. making it a modifier makes it more inline with how we manipulate the dom in the actual components

shleewhite: For reference, here is how it is currently implemented: https://github.com/hashicorp/design-system/blob/41ab596a0835c407da45137b81ba5b241bee5e33/showcase/app/controllers/page-components/copy/button.ts#L27

showcase/app/templates/page-components/copy/button.gts
```diff
@@ -0,0 +1,310 @@
...
+  string
+> = {
+  'masked-input-base': 'With MaskedInput::Base',
+  'masked-input-base-form-field': 'With Form::Field + MaskedInput::Base',
+  'masked-input-field': 'With MaskedInput::Field',
+};
+
+export default class PageComponentCopyButton extends Component {
+  get bigIntNumber() {
+    const bigIntNumber = BigInt(12345678910);
```
aklkv: you do not really gain anything by having this as a getter as there is nothing to watch and re compute, you can either set the value out side fo the class or just use `BigInt` like [so](https://github.com/emberjs/rfcs/blob/9fdecb2341491e5f5ff73974f4619a63bccddb97/text/1070-default-globals-for-strict-mode.md?plain=1#L424) and many more in that RFC which has been already implemented

showcase/app/templates/page-components/copy/button.gts
```diff
@@ -0,0 +1,310 @@
...
+    return bigIntNumber;
+  }
+
+  get targetNodeElement() {
+    const element = document.querySelector('#test-target-node-element');
```
aklkv: you can probably just do this in the constructor and have `#element` class member

showcase/app/templates/page-components/copy/button.gts
```diff
@@ -0,0 +1,310 @@
...
+      return undefined;
+    }
+
+    return element as HTMLElement;
+  }
+
+  <template>
+    {{pageTitle "Copy::Button Component"}}
+
+    <ShwTextH1>Copy::Button</ShwTextH1>
```
aklkv, line 65: Oh this should be 
```suggestion
    <ShwTextH1>CopyButton</ShwTextH1>
```
not sure why linter did pick this up

showcase/app/components/mock/components/copy/button/copy-button-generic-demo.gts
```diff
@@ -0,0 +1,66 @@
...
+import type { TemplateOnlyComponent } from '@ember/component/template-only';
+
+import {
+  HdsCopyButton,
+  HdsFormTextInputBase,
+} from '@hashicorp/design-system-components/components';
+
+import ShwDivider from 'showcase/components/shw/divider';
+
+const CopyButtonGenericContent: TemplateOnlyComponent = <template>
```
aklkv: still need a signature here at least with `Element` in it

showcase/app/components/mock/components/copy/button/copy-button-in-modal.gts
```diff
@@ -0,0 +1,61 @@
...
+  HdsModal,
+} from '@hashicorp/design-system-components/components';
+
+import CopyButtonGenericContent from 'showcase/components/mock/components/copy/button/copy-button-generic-demo';
+
+export default class CopyButtonInModal extends Component {
+  @tracked isModalOpen = false;
+
+  @action
+  openModal() {
```
aklkv: in SFCs you do not need @action decorator you can use arrow functions

showcase/app/components/mock/components/copy/button/copy-button-states.gts
aklkv: this is the way, but also you might simplify this by moving it into constructor, but I may not understand the full implications here, but contractor runs only once when component is rendered similar to `did-insert`

aklkv (commented):
I did initial walk through, this is great and definitely a way to go to make all route templates lighter  with more focused components. I left a few minor comments 👍🏻 🚀

showcase/app/templates/page-components/copy/button.gts
shleewhite: I tried that out but `#element` didnt exist

shleewhite: I tried that out but `#element` didnt exist

aklkv: yeah you need declare it manually first

showcase/app/templates/page-components/copy/button.gts
```diff
@@ -0,0 +1,309 @@
...
+import ShwOutliner from 'showcase/components/shw/outliner';
+import ShwTextH1 from 'showcase/components/shw/text/h1';
+import ShwTextH2 from 'showcase/components/shw/text/h2';
+import ShwTextH3 from 'showcase/components/shw/text/h3';
+import ShwTextH4 from 'showcase/components/shw/text/h4';
+
+import { HdsCopyButton } from '@hashicorp/design-system-components/components';
+import { SIZES } from '@hashicorp/design-system-components/components/hds/copy/button/index';
+
+import CopyButtonStates from 'showcase/templates/page-components/copy/_button-partials/copy-button-states';
```
aklkv: this folder is not for components, maybe in the future you will be able to do this, but templates/ folder is only for route template, components should be in components folder

shleewhite: Ah ok, good to know.

aklkv: I mean you technically can but it would break ember pattern which might be confusing for some developers

showcase/types/template-registry.ts
```diff
@@ -41,6 +41,14 @@ import MockFormSuperSelectGenericContent from '../app/components/mock/components
 import MockFormSuperSelectSelectedComponentSingle from '../app/components/mock/components/form/super-select/selected-component-single';
 import MockFormSuperSelectSelectedComponentMultiple from '../app/components/mock/components/form/super-select/selected-component-multiple';
 
+// PAGE-COMPONENTS
+import CopyButtonInDropdown from '../app/controllers/page-components/copy/button/copy-button-in-dropdown';
```
aklkv, line 45: when everything is converted over to single file components this declarations would not be needed, even now if your route template and all consuming components are single file you do not need this

showcase/app/components/page-components/advanced-table/advanced-table-dynamic-cell.gts
dchyun, line 1: [Note][Discussion] In my table conversion I didn't include `component-name-` in the mock component names. I think it could keep things simpler and align with how we name other components such as the form mock components.

showcase/app/components/page-components/advanced-table/advanced-table-dynamic-cell.gts
```diff
@@ -0,0 +1,116 @@
...
+  Args: {
+    isSelectable?: boolean;
+    density?: HdsAdvancedTableSignature['Args']['density'];
+    isStriped?: boolean;
+    hasTooltips?: boolean;
+  };
+  Element: HTMLDivElement;
+}
+
+export default class AdvancedTableWithDynamicCell extends Component<AdvancedTableWithDynamicCellSignature> {
```
dchyun, line 34:
```suggestion
export default class MockAdvancedTableWithDynamicCell extends Component<MockAdvancedTableWithDynamicCellSignature> {
```
[Question] Should we include `Mock` in the component name to differentiate it from components coming from the HDS library?

showcase/app/components/page-components/advanced-table/advanced-table-dynamic-cell.gts
dchyun, line 1: [Note][Discussion] Just general notes on the differences in how I broke up the table vs how this was broken up. 

I broke up the showcase sections more strongly, and had less example implementation components. For example, all of the multi-select was broken out into its own folder which contained subcomponents for all the different use cases. This led to a shorter main gts file, and it only contained the template. 

An open question is how exactly a showcase page should be broken up. We could make rules like there should be no logic in the main gts file; the mock components should only be example implementations of the main component and not contain showcase components, etc. But I think it's best to keep it more flexible and say whatever works best on a case by case basis.

showcase/app/routes/page-components/advanced-table.ts
dchyun, line 1: [Praise] This to me highlights one of the biggest wins of this approach is the controller files can be completely removed, and the route is stripped of all of its complexity. 

Being able to import type values and data directly into `gts` files helps a lot.

showcase/app/components/page-components/advanced-table/advanced-table-td-matrix.gts
dchyun, line 1: [Suggestion] This might be overkill, but it could be nice if we have a way to differentiate mock components that are showcase sections and those that are example implementations of the main component. 
```
advanced-table
-- examples
---- selectable
---- sortable
-- sections
---- td-matrix
---- th-matrix
```

dchyun (commented):
There are a few differences between this and my table spike #3098, along with some other general topics I've added comments for. Just calling these things out as items to discuss, not that I feel strongly about one approach or another. 

Overall v much on board with the conversion to GTS and breaking up the components. I think being able to remove all of the complexity from the controller and route files is great.

showcase/app/components/page-components/advanced-table/advanced-table-td-matrix.gts
shleewhite: I like this idea a lot, can help reviewers understand if an example is meant to be interactive or not

PR #2535, opened 2024-10-31
`SideNav` - Prevent hidden panels from triggering unnecessary overflow scrolling

KristinLBradley:
### :pushpin: Summary

If merged, this PR will prevent unwanted scrollbars from being triggered by hidden panels.

### :link: External links

* Original AppSideNav PRed fix: https://github.com/hashicorp/design-system/pull/2487
* Related ticket: [HDS-3892](https://hashicorp.atlassian.net/browse/HDS-3892)

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-3892]: https://hashicorp.atlassian.net/browse/HDS-3892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3045, opened 2025-07-23
 `AppSideNav::Panel` - Fix focus ring cut off issue (HDS-5165)

KristinLBradley:
### :pushpin: Summary

If merged, this PR fixes an issue with the focus ring of top and bottom items being cut off when used inside the `AppSideNav` Panel.

**Preview**: https://hds-showcase-git-kristin-hds-5165-fix-appsiden-ca24d0-hashicorp.vercel.app/components/app-side-nav

**To test**: tab through links within AppSideNav examples using content injected via portal

### :link: External links

* Jira ticket: [HDS-5165](https://hashicorp.atlassian.net/browse/HDS-5165)
* Related PR: https://github.com/hashicorp/design-system/pull/3032

***

<details>

</details>

[HDS-5165]: https://hashicorp.atlassian.net/browse/HDS-5165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/rich-suits-count.md
+++ b/.changeset/rich-suits-count.md
@@ -3,5 +3,5 @@
 ---
 
 <!-- START components/app-side-nav-->
-`AppSideNav::Panel - Fixed issue causing focusring of first and last items within the Panel to be cut off
+`AppSideNav::Panel - Fixed issue causing the focus ring of the first and last items within the Panel to be cut off
 <!-- END -->
--- a/.changeset/rich-suits-count.md
+++ b/.changeset/rich-suits-count.md
@@ -2,4 +2,6 @@
 "@hashicorp/design-system-components": patch
 ---
 
+<!-- START components/app-side-nav-->
 `AppSideNav::Panel - Fixed issue causing focusring of first and last items within the Panel to be cut off
+<!-- END -->
--- a/.changeset/rich-suits-count.md
+++ b/.changeset/rich-suits-count.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`AppSideNav::Panel - Fixed issue causing focusring of first and last items within the Panel to be cut off
--- a/packages/components/src/styles/components/app-side-nav/content.scss
+++ b/packages/components/src/styles/components/app-side-nav/content.scss
@@ -35,14 +35,11 @@
 .hds-app-side-nav__content-panel {
   padding: 0 var(--token-app-side-nav-wrapper-padding-horizontal);
 
-  // the panel itself should not be scrollable,
-  // applying the style in this way prevents focus rings from being cut off
-  &:not(:has(:focus, .mock-focus)) {
-    overflow: hidden;
-  }
-
   &[aria-hidden="true"] {
     max-height: 0; // prevents hidden panels from causing scrolling
+    // prevent unwanted scrolling when navigating from panel with long content to panel with short content
+    // applying only to hidden panels, prevents Item focus rings from being cut off
+    overflow: hidden;
   }
 }
 
--- a/packages/components/src/styles/components/app-side-nav/content.scss
+++ b/packages/components/src/styles/components/app-side-nav/content.scss
@@ -34,7 +34,12 @@
 
 .hds-app-side-nav__content-panel {
   padding: 0 var(--token-app-side-nav-wrapper-padding-horizontal);
-  overflow: hidden; // the panel itself does not need to be scrollable
+
+  // the panel itself should not be scrollable,
+  // applying the style in this way prevents focus rings from being cut off
+  &:not(:has(:focus, .mock-focus)) {
+    overflow: hidden;
+  }
 
   &[aria-hidden="true"] {
     max-height: 0; // prevents hidden panels from causing scrolling


packages/components/src/styles/components/app-side-nav/content.scss
```diff
@@ -34,7 +34,12 @@
 
 .hds-app-side-nav__content-panel {
   padding: 0 var(--token-app-side-nav-wrapper-padding-horizontal);
-  overflow: hidden; // the panel itself does not need to be scrollable
+
+  // the panel itself should not be scrollable,
+  // applying the style in this way prevents focus rings from being cut off
+  &:not(:has(:focus, .mock-focus)) {
+    overflow: hidden;
+  }
```
KristinLBradley: Applying the style in this way should prevent unforeseen issues vs. just deleting it and should allow the Atlas/Terraform team to delete their custom `overflow: hidden` style on `.sidebar-scrollbar `

.changeset/rich-suits-count.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+<!-- START components/app-side-nav-->
+`AppSideNav::Panel - Fixed issue causing focusring of first and last items within the Panel to be cut off
```
didoo:
```suggestion
`AppSideNav::Panel - Fixed issue causing the focus ring of the first and last items within the Panel to be cut off
```

packages/components/src/styles/components/app-side-nav/content.scss
didoo: [thought/suggestion] I was thinking: what if, instead of deleting it, we move it under the `&[aria-hidden="true"]` block?

didoo (commented):
left a comment

packages/components/src/styles/components/app-side-nav/content.scss
KristinLBradley: You mean the custom style in Atlas for `.sidebar-scrollbar`?

didoo: no, the line below (see line 44)

KristinLBradley: So you mean just do this?

```
.hds-app-side-nav__content-panel {
  padding: 0 var(--token-app-side-nav-wrapper-padding-horizontal);

  &[aria-hidden="true"] {
    max-height: 0; // prevents hidden panels from causing scrolling
    overflow: hidden;
  }
}
```

I could do that but it doesn't have the same effect as before. The change I have now keeps the current behavior except when an Item within the AppSideNav is focused.

KristinLBradley: Anyway, please let me know if you think that would be better (as in above code example).

didoo: I think we need to better understand **why** the overflow "fix" was added in the first place, to decide which is the _right™_ solution. Otherwise we're flying blind. I think you have more context/memory of the changes introduced in this PR and why: https://github.com/hashicorp/design-system/pull/2535 (I don't remember, tbh)

didoo: @KristinLBradley OK, I did some digging, here's what I've found:
- the `overflow: hidden` applied to the `content-panel` was initially introduced in #2535 for the `SideNav` and then the "fix" was ported to the `AppSideNav` when the CSS file was cloned.
- the Jira ticket associated to the PR is [HDS-3892](https://hashicorp.atlassian.net/browse/HDS-3892) and refers to [this Slack thread](https://hashicorp.slack.com/archives/C7KTUHNUS/p1727211947471179) 
- In the Slack thread, a video is shared ([link](https://hashicorp.enterprise.slack.com/files/U02BDUX98MV/F07NPB0D4MU/screen_recording_2024-09-24_at_2.02.23___pm.mov)) which shows what is the use case where the "fix" is needed: nested navigational panels (rendered as siblings, in the DOM) where one has a lot of items, so it scrolls vertically, and the other has just a few items.
- If the second panel is visible (and the first is shifted out  of sight) the whole parent container can scroll, because the first panel is taller than the visible area, and even if the `max-height` is set to zero, its child element is tall:

https://github.com/user-attachments/assets/3a4cf93a-2deb-4680-8248-400d9a6f3252

Now, for the selector `&[aria-hidden="true"] {` the comment next to the `max-height: 0` set says `// prevents hidden panels from causing scrolling` which for me is a good indicator that _probably_ also the `overflow: hidden` declaration should go there: https://github.com/hashicorp/design-system/blob/main/packages/components/src/styles/components/side-nav/content.scss#L39-L40

I have tested this idea in the browser, and having the `overflow: hidden` moved from the `.hds-side-nav__content-panel` to the `.hds-side-nav__content-panel[aria-hidden="true"]` seems to work: 

https://github.com/user-attachments/assets/5e8e34d6-3ca9-4fe0-99d7-3e4c5ce9bb0b

but it's tested in the `SideNav`, not the `AppSideNav`. So I'll leave it to you to do proper testing, and come up with a conclusive decision on if this works as we need/want, or has other side effects that we didn't consider.

KristinLBradley: Thank you, I had looked at the old PR but hadn't dug enough into the original details from the Slack thread.

I tested within Cloud-ui with the `AppSideNav` locally. I removed the `overflow: hidden` style and navigated from a list of long links to a list of shorter links within the panel and checked if there was unwanted scrolling (as in the bug) and there was. I then added the `overflow: hidden` style back but scoped to `[aria-hidden=true]` and that fixed the scrolling issue.

I wasn't able to uncover any new unwanted side effects so I think this is a pretty safe change. I'll go ahead and push updated code to this PR. 

Navigated from:
image

To:
image

KristinLBradley:
> 👍
> 
> One request: link (and cross-link) as many resources (PRs, Jiras, Slack messages) as possible, in the Jira and PR links, so that in the future we have to dig again, is much easier to collect info about these fixes and build a history of what happened. thanks

@didoo I merged as is for now to get it into the release. I'll open a follow-on PR to update the comments

didoo:
> @didoo I merged as is for now to get it into the release. I'll open a follow-on PR to update the comments

@KristinLBradley no I meant in this PR description and in the Jira description, not in code (even though, a comment in the code pointing to "something" may be worth)

PR #3054, opened 2025-07-24
`ApplicationState`, `BadgeCount`, `Button`, `Card`: fix exported const types

shleewhite:
### :pushpin: Summary

If merged, this PR would fix the types of exported consts from `ApplicationState`, `BadgeCount`, `Button` and `Card`

### :link: External links

Figma file: [if it applies]

***

<details>

</details>

### DIFF
--- a/packages/components/src/components/hds/button/index.ts
+++ b/packages/components/src/components/hds/button/index.ts
@@ -20,14 +20,14 @@ import type {
 import type { HdsInteractiveSignature } from '../interactive/';
 import type { HdsIconSignature } from '../icon';
 
-export const SIZES: string[] = Object.values(HdsButtonSizeValues);
-export const COLORS: string[] = Object.values(HdsButtonColorValues);
-export const ICONPOSITIONS: string[] = Object.values(
+export const SIZES: HdsButtonSizes[] = Object.values(HdsButtonSizeValues);
+export const COLORS: HdsButtonColors[] = Object.values(HdsButtonColorValues);
+export const ICON_POSITIONS: HdsButtonIconPositions[] = Object.values(
   HdsButtonIconPositionValues
 );
 export const DEFAULT_SIZE = HdsButtonSizeValues.Medium;
 export const DEFAULT_COLOR = HdsButtonColorValues.Primary;
-export const DEFAULT_ICONPOSITION = HdsButtonIconPositionValues.Leading;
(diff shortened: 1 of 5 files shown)


packages/components/src/components/hds/button/index.ts
```diff
@@ -20,14 +20,14 @@ import type {
 import type { HdsInteractiveSignature } from '../interactive/';
 import type { HdsIconSignature } from '../icon';
 
-export const SIZES: string[] = Object.values(HdsButtonSizeValues);
-export const COLORS: string[] = Object.values(HdsButtonColorValues);
-export const ICONPOSITIONS: string[] = Object.values(
+export const SIZES: HdsButtonSizes[] = Object.values(HdsButtonSizeValues);
+export const COLORS: HdsButtonColors[] = Object.values(HdsButtonColorValues);
+export const ICON_POSITIONS: HdsButtonIconPositions[] = Object.values(
```
didoo, line 25: [question] I imagine these are not imported (or not anymore) in the showcase?

shleewhite: Ah, its not! Should I delete the const?

dchyun: Maybe we should make examples for them in the showcase considering they're an option available? It could also be good to keep them like this for consistency's sake too.

dchyun (approved):
lgtm!

packages/components/src/components/hds/button/index.ts
shleewhite: Well.. there is an example of a trailing icon, we just don't use the exported array to display it: https://hds-showcase.vercel.app/components/button#content

didoo:
@shleewhite any idea why Percy is failing? I looked at the diff and it's all messed up (probably an issue with Percy)

packages/components/src/components/hds/card/container.ts
dchyun, line 1: [Issue] The card showcase route still uses the `AVAILABLE_` names so the showcase page currently isn't working. It'll need to be updated just for now, even though it'll get converted to TS soon.

dchyun: @didoo this is why Percy is failing

shleewhite: ah! of course, just pushed a fix.

PR #3063, opened 2025-07-29
`Flyout`, `Icon`, `IconTile`, `Modal`, `RichTooltip`: fix types of exported consts

shleewhite:
### :pushpin: Summary

If merged, this PR would fix the types of exported consts for several components.
### :link: External links

***

<details>

</details>


KristinLBradley (approved):

packages/components/src/modifiers/hds-anchored-position.ts
```diff
@@ -50,6 +50,11 @@ export const PLACEMENTS: Placement[] = [
   'left-end',
 ];
 
+type EnableCollisionDetectionOptions = 'shift' | 'flip' | 'auto';
```
didoo: [question] should we use an approach more similar to what we do for other enums that gets converted in CONSTANT_VALUES? https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/button/types.ts#L6-L11

didoo (dismissed):
Left only one comment/question, not a blocker

packages/components/src/modifiers/hds-anchored-position.ts
shleewhite: ahh, good point. updated to match.

packages/components/src/modifiers/hds-anchored-position.ts
```diff
@@ -34,6 +34,8 @@ import type {
   Middleware,
 } from '@floating-ui/dom';
 
+import type { HdsEnableCollisionDetection } from './hds-code-editor/types.ts';
```
didoo: [blocker] could it be you've saved the type in the wrong file? 🤔

dchyun: Yeah this type should probably just be defied in this modifier file.

packages/components/src/modifiers/hds-anchored-position.ts
```diff
@@ -50,6 +52,9 @@ export const PLACEMENTS: Placement[] = [
   'left-end',
 ];
 
+export const ENABLE_COLLISION_DETECTION_OPTIONS: HdsEnableCollisionDetection[] =
```
dchyun, line 62:
```suggestion
export const ENABLE_COLLISION_DETECTION_OPTIONS: HdsEnableCollisionDetection[] = Object.values(HdsEnableCollisionDetectionOptions)
```
[Suggestion] Could use the existing type values object here.

packages/components/src/modifiers/hds-anchored-position.ts
shleewhite: 🤦🏽‍♀️

PR #3066, opened 2025-07-29
`Table` & `AdvancedTable` - Type fixes for `isSelected` and `selectionKey`

dchyun:
### :pushpin: Summary

If merged, this PR would fix type issues found in the `Table` and `AdvancedTable` for the following arguments.

- `isSelected` in the sub-component `Tr` 
  - Change type from `false` to `boolean`
- `selectionKey` and related args in the sub-component `Tr`
  - Change type from `string` to `string | number`

### :hammer_and_wrench: Detailed description

All of these type issues in the table components were caught while converting their showcase pages to TS.

#### `isSelected`

Currently in the sub-components `Table::Tr` and `AdvancedTable::Tr` the `isSelected` argument set to the type `false`. This means that anytime `isSelected` is set to `true` a type error is thrown. 
Screenshot 2025-07-29 at 11 10 33 AM

The type has now been set to `boolean` and a getter has been added to set `this.isSelected` to `false` by default.

#### `selectionKey`
 
Currently in the sub-components `Table::Tr` and `AdvancedTable::Tr` the `selectionKey` argument type is set to `string`. This causes an issue when a `number` is used in this type. It is common to use a `number` for the `selectionKey` if some data's `id` is the key. 

This is the case in our showcase application where our demo data passes in a row's `id` as the `selectionKey`. When this occurs most of the time, it can get through smoothly, but in some cases errors are thrown. 
Screenshot 2025-07-29 at 3 41 40 PM

Also, the in the Table documentation, [the component API](https://helios.hashicorp.design/components/table/table?tab=code#tabletr) incorrectly states the type for `selectionKey` is `string | number`.

I have updated the type of this argument, and all related arguments to be `string | number`.

### :link: External links

Jira ticket: [HDS-5183](https://hashicorp.atlassian.net/browse/HDS-5183)

***

### :eyes: Component checklist

- [ ] Percy was checked for any visual regression
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

<details>

<summary>:clipboard: PCI review checklist</summary>

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.
- [ ] If applicable, I've worked with GRC to document the impact of any changes to security controls.
[...]

PR #3056, opened 2025-07-24
Showcase: convert AdvancedTable route to TS

shleewhite:
### :pushpin: Summary

if merged, this PR would convert the AdvancedTable showcase route to TS.

### :link: External links

Jira ticket: [HDS-5076](https://hashicorp.atlassian.net/browse/HDS-5076)
Figma file: [if it applies]

***

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

<details>

<summary>:clipboard: PCI review checklist</summary>

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.
- [ ] If applicable, I've worked with GRC to document the impact of any changes to security controls.
  Examples of changes to controls include access controls, encryption, logging, etc.
- [ ] If applicable, I've worked with GRC to ensure compliance due to a significant change to the in-scope PCI environment.
  Examples include changes to operating systems, ports, protocols, services, cryptography-related components, PII processing code, etc.

</details>

PR #3067, opened 2025-07-29
Showcase - Convert Table route to TS

dchyun:
### :pushpin: Summary

If merged, this PR would convert the pages in the showcase for the `Table` component to TS.

[Showcase page](https://hds-showcase-git-dchyun-table-showcase-ts-conversion-hashicorp.vercel.app/components/table)

Note: Percy diff is expected as the `medium` `density` is now being displayed.

### :link: External links

***

<details>

</details>

### DIFF
--- a/showcase/app/controllers/page-components/advanced-table.ts
+++ b/showcase/app/controllers/page-components/advanced-table.ts
@@ -10,6 +10,8 @@ import { deepTracked } from 'ember-deep-tracked';
 import { later } from '@ember/runloop';
 
 import type { PageComponentsAdvancedTableModel } from 'showcase/routes/page-components/advanced-table';
+import type { SelectableItem } from 'showcase/mocks/selectable-item-data';
+import type { User } from 'showcase/mocks/user-data';
 
 import type {
   HdsAdvancedTableOnSelectionChangeSignature,
@@ -25,12 +27,12 @@ const customSortingCriteriaArray = [
   'pending',
 ];
 
-const updateModelWithSelectAllState = <T>(
-  modelData: T[],
+const updateModelWithSelectAllState = (
+  modelData: SelectableItem[] | User[],
   selectAllState: boolean,
 ) => {
   modelData.forEach((modelRow) => {
-    if (modelRow instanceof Object && 'isSelected' in modelRow) {
+    if (modelRow instanceof Object) {
       modelRow.isSelected = selectAllState;
     }
   });
--- a/showcase/app/routes/page-components/advanced-table.ts
+++ b/showcase/app/routes/page-components/advanced-table.ts
@@ -15,7 +15,6 @@ import policies from 'showcase/mocks/policy-data';
 import selectableItems from 'showcase/mocks/selectable-item-data';
 import spanningCells from 'showcase/mocks/spanning-cell-data';
 import users from 'showcase/mocks/user-data';
-import userWithMoreColumns from 'showcase/mocks/user-with-more-columns-data';
 
 export type PageComponentsAdvancedTableModel =
   ModelFrom<PageComponentsAdvancedTableRoute>;
@@ -42,7 +41,6 @@ export default class PageComponentsAdvancedTableRoute extends Route {
       userDataDemo4: clone(
         users.slice(0, 4).map((user) => ({ ...user, isAnimated: false })),
       ),
-      manyColumns: userWithMoreColumns,
       nestedData: policies,
       nestedDataCustom: policies.map((policy) => {
         const { children, ...rest } = policy;


showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,649 @@
...
+
+  // CUSTOM SORTING DEMO #5
+  // Sortable table with model and sorting by selected row
+
+  @action
+  onMultiSelectSelectionChange__demo5({
+    selectionKey,
+    selectionCheckboxElement,
+  }: HdsTableOnSelectionChangeSignature) {
+    if (selectionKey) {
```
dchyun, line 265: [Note] `selectionKey` and `selectionCheckboxElement` are optional, so checks have been added to handle that use case. This applies to all similar selection change functions below.

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,649 @@
...
+  @tracked customSortOrder_demo2 = 'asc';
+  @tracked customSortBy_demo3: string | undefined = undefined;
+  @tracked customSortOrder_demo3 = 'asc';
+
+  // multi-select
+  @tracked multiSelectFilterRows__demo1 = 'all';
+  @tracked multiSelectToggleScope__demo1 = false;
+  @tracked multiSelectToggleDebug__demo1 = false;
+  @deepTracked multiSelectSelectableData__demo1 = [
+    // @ts-expect-error - we need to reevaluate how we get the data for the table demos when we break up the template files into sub components
```
dchyun, line 79: [Note] Without this, an error is flagged that the controller is using `this.model` before it is defined. Similar expect errors are added for other data variables.

showcase/app/templates/page-components/table.hbs
```diff
@@ -260,15 +285,19 @@
   <Hds::Table @onSort={{this.extraOnSortCallback_demo3}}>
     <:head as |H|>
       <H.Tr>
-        <H.ThSort
-          @onClickSort={{fn this.onClickThSort__demo3 "peer-name" H.setSortBy}}
-          @sortOrder={{if (eq "peer-name" H.sortBy) H.sortOrder}}
-        >Peer name</H.ThSort>
+        {{#if H.setSortBy}}
```
dchyun: [Note] `setSortBy` is conditionally yielded from the `:head` so calling `{{fn this.onClickThSort__demo3 "peer-name" H.setSortBy}}` raises an error. I added a check for if it is defined, but maybe there is some better solution here so consumers can avoid doing that.

Screenshot 2025-07-29 at 3 36 29 PM

showcase/package.json
```diff
@@ -56,6 +56,7 @@
     "@glint/template": "^1.5.2",
     "@hashicorp/design-system-components": "workspace:*",
     "@hashicorp/design-system-tokens": "workspace:*",
+    "@nullvoxpopuli/ember-composable-helpers": "^5.2.11",
```
dchyun, line 60: [Note] In the Table template the helper `call` is used, but its package, `ember-composable-helpers`, isn't included in the showcase devDependencies.

showcase/app/templates/page-components/table.hbs
```diff
@@ -855,51 +963,49 @@
...
+                      <H.Th>Dolor</H.Th>
+                    </H.Tr>
+                  </:head>
+                  <:body as |B|>
+                    <B.Tr @selectionKey="row1">
+                      <B.Th>Scope Row</B.Th>
+                      <B.Td>Cell Content</B.Td>
+                      <B.Td>Cell Content</B.Td>
+                    </B.Tr>
+                    <B.Tr @selectionKey="row2" @isSelected={{true}}>
```
dchyun, line 984: [Note] There are type errors thrown on `isSelected` when it is set to `true` due to its type being `false`. This is fixed in #3066

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,649 @@
...
+
+  @action
+  onMultiSelectSelectionChange__demo4({
+    selectedRowsKeys,
+  }: HdsTableOnSelectionChangeSignature) {
+    console.group('Selection Change with Model Arguments');
+    console.log('Selected Row Keys:', selectedRowsKeys);
+    console.groupEnd();
+    this.multiSelectUserData__demo4.forEach((user) => {
+      user.isSelected = selectedRowsKeys.includes(user.id);
```
dchyun: [Note] There are type errors thrown here as `user.id` is a `number`, but `selectionRowKeys` expects a string. `selectionRowKeys` has been updated to support `number` in #3066 .

showcase/app/templates/page-components/table.hbs
```diff
@@ -445,13 +483,20 @@
     }}
   >
     <:body as |B|>
+      {{! @glint-expect-error - will be fixed by https://hashicorp.atlassian.net/browse/HDS-5090}}
       <B.Tr
-        @selectionKey={{B.data.id}}
+        {{! @glint-expect-error - will be fixed by https://hashicorp.atlassian.net/browse/HDS-5090}}
+        @selectionKey="{{B.data.id}}"
```
dchyun, line 491: [Note] `@selectionKey` is typed as a `string`, but the data `id` is a `number`, so converting it to a string here, and in following similar instances.

showcase/app/templates/page-components/table.hbs
```diff
@@ -3,6 +3,7 @@
   SPDX-License-Identifier: MPL-2.0
 }}
 
+{{! @glint-expect-error - file will be broken up in https://hashicorp.atlassian.net/browse/HDS-5077}}
```
dchyun, line 6: [Note] There is an error thrown about how lengthy this page is, this should be able to be removed once the page is broken up into sub templates.

showcase/app/templates/page-components/table.hbs
```diff
@@ -202,6 +222,7 @@
   <Shw::Text::H4>Sortable table with custom <code>sortingFunction</code> declared in the column hash</Shw::Text::H4>
 
   <Hds::Table
+    {{! @glint-expect-error - will be fixed by https://hashicorp.atlassian.net/browse/HDS-5090}}
```
shleewhite, line 230: Something to investigate... Its odd that the tables where the model is music it doesnt need a glint-expect-error but this one does

showcase/app/templates/page-components/table.hbs
```diff
@@ -217,20 +238,24 @@
   >
     <:body as |B|>
       <B.Tr>
+        {{! @glint-expect-error - will be fixed by https://hashicorp.atlassian.net/browse/HDS-5090}}
         <B.Td>{{B.data.peer-name}}</B.Td>
+        {{! @glint-expect-error - will be fixed by https://hashicorp.atlassian.net/browse/HDS-5090}}
         <B.Td>{{B.data.cluster-partition}}</B.Td>
         <B.Td>
-          {{#if (eq B.data.status "failing")}}
+          {{#if (eq (get B.data "status") "failing")}}
```
shleewhite, line 251: Should we use this syntax for all the times we access keys in B.data? Would that make the glint expect error on line 243 irrelevant?

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,649 @@
...
+          return 0;
+        }
+      });
+    }
+    return clonedModelClusters;
+  }
+
+  @action
+  extraOnSortCallback_demo3() {
+    console.log(
```
shleewhite: Suggestion: you could use console.group instead if you want

shleewhite (commented):
Just a couple small comments!

showcase/app/routes/page-components/table.ts
```diff
@@ -0,0 +1,137 @@
...
+import { DENSITIES } from '@hashicorp/design-system-components/components/hds/table/index';
+
+import type { ModelFrom } from 'showcase/utils/ModelFromRoute';
+import type {
+  HdsBadgeTypes,
+  HdsBadgeColors,
+} from '@hashicorp/design-system-components/components/hds/badge/types';
+import type { HdsIconSignature } from '@hashicorp/design-system-components/components/hds/icon/index';
+
+export interface MockDataMusic {
```
didoo: @dchyun @shleewhite I see differences in both the naming, the types of  some fields, but also in the fact that here they are exported and consumed inside the constructor, while in Lee's PR https://github.com/hashicorp/design-system/pull/3056 as far as I can tell they're not used in this way.

do you mind sync and come up with as much alignment as possible (also in other parts of the migrated code)? thanks

showcase/app/templates/page-components/table.hbs
didoo: wild guess: in @dchyun's PR I see this `return this.model.clusters.map((record: MockDataCluster)` while in your PR I see this `return this.model.clusters.map((record)`?

showcase/app/templates/page-components/table.hbs
didoo: How would a consumer would "normally" do? 

Also, I don't understand Glint in this case: the signature of `onClickThSort__demo3` says that `setSortBy` is optional, so the check that `H.setSortBy` is not undefined what is solving, from a type safety perspective? 🤷‍♂️

showcase/app/templates/page-components/table.hbs
didoo: This has now been merged, so rebasing should fix the issue

showcase/app/templates/page-components/table.hbs
shleewhite: Looking at the instances where H.setSortBy is used in product (I can only find it in boundary/atlas), it seems like they don't have to check that it exists but I am not sure the files are typed.

It does seem like if I remove the check, glint doesnt complain anymore because it is optional for this.onClickThSort__demo3.

showcase/app/routes/page-components/table.ts
shleewhite: I am currently working on a ticket to simplify the way we do data across the pagination, table, and advanced table showcase pages - which will put all of this stuff in one place instead of having it copied across the files: https://github.com/hashicorp/design-system/pull/3074

showcase/app/templates/page-components/table.hbs
dchyun: Still unsure how to resolve this one. First, I removed the typing of `record`, and that had no effect. This expect error is also added above `this.model.clusters` in @shleewhite's PR.

I think the main issue is that in the model `clusters` is of the object type `MockDataCluster`, but the `@model` argument expects a `Record`. That's why the original error here was thrown.
Screenshot 2025-08-04 at 11 17 26 AM


In the model, `music` is converted into a `Record`. I tried doing the same thing with the cluster data
```
clusters: clusters.map((cluster) => ({ ...cluster })) as Record<string, MockDataCluster[keyof MockDataCluster]>[],
``` 
The issue is then, in the controller and the template `clusters` it is less strongly typed than when it was typed as `MockDataCluster`. So some other errors get thrown. 
Screenshot 2025-08-04 at 11 14 57 AM

I guess we could add expect errors above all of these lines, in order to remove the one above the `this.model.clusters`.

showcase/app/templates/page-components/table.hbs
dchyun: I tried using the get syntax and a glint error is still thrown. 
Screenshot 2025-08-04 at 10 49 47 AM

showcase/app/templates/page-components/table.hbs
```diff
@@ -322,15 +347,19 @@
   <Hds::Table>
     <:head as |H|>
       <H.Tr>
-        <H.ThSort
-          @onClickSort={{fn H.setSortBy "peer-name"}}
-          @sortOrder={{if (eq "peer-name" H.sortBy) H.sortOrder}}
-        >Peer name</H.ThSort>
+        {{#if H.setSortBy}}
```
dchyun, line 355: [Note] @didoo @shleewhite this is where the check for `H.setSortBy` is needed. In `@onClickSort`, `setSortBy` is called directly and because it is conditionally yielded it's possible it could be undefined.

showcase/app/templates/page-components/table.hbs
dchyun: I was able to remove the check from this spot, as it is correct that it is optional in `this.onClickThSort__demo3`. However, this check is still needed in other examples in the template where `setSortBy` is called directly. Added in a [new comment](https://github.com/hashicorp/design-system/pull/3067/files#r2251830493) for one of those examples.

showcase/app/templates/page-components/table.hbs
shleewhite: Ohh interesting. Lets leave It as is for now and then circle back once we have done HDS-5090. I think the issues related to that could be hiding other type issues.

showcase/app/templates/page-components/table.hbs
shleewhite: Dang, never mind then.

shleewhite (approved):

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,650 @@
...
+  get customSortingFunction_demo2(): <T>(s1: T, s2: T) => number {
+    return (s1, s2) => {
+      // check that s1 and s2 are objects and have the 'status' property
+      if (
+        s1 instanceof Object &&
+        s2 instanceof Object &&
+        'status' in s1 &&
+        'status' in s2 &&
+        typeof s1['status'] === 'string' &&
+        typeof s2['status'] === 'string'
```
didoo, line 142: Ooof!

didoo: [question] at line `153` there is similar code (`clonedModelClusters.sort`) but it's not so "strictly typed" as here. Why in that case is not necessary, but here it is?

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,650 @@
...
+        );
+
+        if (recordToUpdate) {
+          recordToUpdate.isSelected = !recordToUpdate.isSelected;
+        }
+      }
+    }
+  }
+
+  // // GENERIC MULTI-SELECT FUNCTIONALITIES
```
didoo:
```suggestion
  // GENERIC MULTI-SELECT FUNCTIONALITIES
```

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,650 @@
...
+
+  @action
+  onSelectionChangeLogArguments({
+    selectionKey,
+    selectionCheckboxElement,
+    selectableRowsStates,
+    selectedRowsKeys,
+  }: HdsTableOnSelectionChangeSignature) {
+    console.group('Selection Change with Model Arguments');
```
didoo:
```suggestion
    console.group('onSelectionChangeLogArguments invoked with arguments');
```

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,650 @@
...
+    this.multiSelectFilterRows__demo1 = checkbox.value;
+  }
+
+  @action
+  onSelectionChangeWithModel__demo1({
+    selectionKey,
+    selectionCheckboxElement,
+    selectableRowsStates,
+  }: HdsTableOnSelectionChangeSignature) {
+    console.group('Selection Change with Model Arguments');
```
didoo:
```suggestion
    console.group('onSelectionChangeWithModel__demo1 invoked with arguments');
```

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,650 @@
...
+    }
+  }
+
+  @action
+  onSelectionChangeWithoutModel__demo1({
+    selectionKey,
+    selectionCheckboxElement,
+    selectableRowsStates,
+  }: HdsTableOnSelectionChangeSignature) {
+    console.group('Selection Change with Model Arguments');
```
didoo:
```suggestion
    console.group('onSelectionChangeWithoutModel__demo1 invoked with arguments');
```

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,650 @@
...
+    this.multiSelectPaginatedCurrentPageSize_demo2 = pageSize;
+  }
+
+  @action
+  onMultiSelectPaginatedSelectionChange__demo2({
+    selectionKey,
+    selectionCheckboxElement,
+    selectableRowsStates,
+  }: HdsTableOnSelectionChangeSignature) {
+    console.group('Selection Change with Model Arguments');
```
didoo:
```suggestion
    console.group('onMultiSelectPaginatedSelectionChange__demo2 invoked with arguments');
```

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,650 @@
...
+    this.multiSelectUsersCurrentPageSize_demo3 = pageSize;
+  }
+
+  @action
+  onMultiSelectUsersSelectionChange__demo3({
+    selectionKey,
+    selectionCheckboxElement,
+    selectableRowsStates,
+  }: HdsTableOnSelectionChangeSignature) {
+    console.group('Selection Change with Model Arguments');
```
didoo:
```suggestion
    console.group('onMultiSelectUsersSelectionChange__demo3 invoked with arguments');
```

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,650 @@
...
+  }
+
+  // // MULTI-SELECT DEMO #4
+  // // Execute action on selected rows
+
+  @action
+  onMultiSelectSelectionChange__demo4({
+    selectedRowsKeys,
+  }: HdsTableOnSelectionChangeSignature) {
+    console.group('Selection Change with Model Arguments');
```
didoo:
```suggestion
    console.group('onMultiSelectSelectionChange__demo4 invoked with arguments');
```

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,650 @@
...
+          return 0;
+        }
+      });
+    }
+    return clonedModelClusters;
+  }
+
+  @action
+  extraOnSortCallback_demo3() {
+    console.group('extraOnSortCallback called with:');
```
didoo:
```suggestion
    console.group('extraOnSortCallback invoked:');
```

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,650 @@
...
+          this.multiSelectNoModelState__demo1[rowKey] = row.isSelected
+            ? true
+            : false;
+        });
+      }
+    }
+  }
+
+  // // MULTI-SELECT DEMO #2
+  // // Multi-select table with pagination
```
didoo:
```suggestion
  // MULTI-SELECT DEMO #2
  // Multi-select table with pagination
```

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,650 @@
...
+        updateModelWithSelectableRowsStates(
+          this.multiSelectSelectableData__demo2,
+          selectableRowsStates,
+        );
+      }
+    }
+  }
+
+  // // MULTI-SELECT DEMO #3
+  // // Delete selected rows
```
didoo:
```suggestion
  // MULTI-SELECT DEMO #3
  // Delete selected rows
```

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,650 @@
...
+  @action
+  multiSelectDeleteSelectedUsers_demo3() {
+    const newData = this.multiSelectUserData__demo3.filter(
+      (user) => !user.isSelected,
+    );
+    this.multiSelectUserData__demo3 = [...newData];
+  }
+
+  // // MULTI-SELECT DEMO #4
+  // // Execute action on selected rows
```
didoo:
```suggestion
  // MULTI-SELECT DEMO #4
  // Execute action on selected rows
```

showcase/app/controllers/page-components/table.ts
```diff
@@ -0,0 +1,650 @@
...
+        } else {
+          return 0;
+        }
+      }
+      return 0;
+    };
+  }
+
+  @action
+  customOnSort_demo2(_sortBy: string, sortOrder: string) {
```
didoo: [note] [in the AdvancedTable](https://github.com/hashicorp/design-system/blob/main/showcase/app/controllers/page-components/advanced-table.ts) this is `customOnSort(_sortBy: string, sortOrder: HdsAdvancedTableThSortOrder)` (it uses the `HdsAdvancedTableThSortOrder`)

didoo (commented):
Left a few comments. 

Can you open the equivalent files from the `AdvancedTable` side by side, and compare the implementation there and here, to make sure they're as similar as possible. 
It's not nitpicking, but because this consistency will facilitate future code refactoring, debugging, reviews, etc.

showcase/app/controllers/page-components/table.ts
dchyun: The difference between this sorting function and others (like line 184 `get sortedModelClusters__demo3()`) is that this function is used in the `sortingFunction` column hash. 

In the function I can type the parameters as `MockDataCluster` so `s1['status']` is defined, but the problem is the parameters of the `sortingFunction` are typed as `unknown` so there is a type mismatch.

This implementation matches the example from the adv table so thinking to leave it as is for now. 

Screenshot 2025-08-07 at 9 30 01 AM

showcase/app/controllers/page-components/table.ts
dchyun: Was able to align the types here to the adv table implementation

dchyun:
> Can you open the equivalent files from the AdvancedTable side by side, and compare the implementation there and here

@didoo Compared the adv table and table files and aligned in a few places to the adv table. Nothing major, but should be aligned for all similar functions now.

didoo:
@dchyun since https://github.com/hashicorp/design-system/pull/3074 is practically ready to be merged (only a couple of tiny things left to fix) do you think it would make sense to rebase you branch on that PR's branch, and use the data sources (and data types) from that PR? I think it could simplify this PR and improve even more the consistency between Table and AdvancedTable migrated code

dchyun:
> @dchyun since #3074 is practically ready to be merged (only a couple of tiny things left to fix) do you think it would make sense to rebase you branch on that PR's branch, and use the data sources (and data types) from that PR? I think it could simplify this PR and improve even more the consistency between Table and AdvancedTable migrated code

@didoo Yes that was my intention to wait for the data reorg to go in before rebasing and merging this. I agree it should simplify things a lot.

showcase/app/controllers/page-components/advanced-table.ts
```diff
@@ -10,6 +10,8 @@ import { deepTracked } from 'ember-deep-tracked';
 import { later } from '@ember/runloop';
 
 import type { PageComponentsAdvancedTableModel } from 'showcase/routes/page-components/advanced-table';
+import type { SelectableItem } from 'showcase/mocks/selectable-item-data';
```
shleewhite, line 13: 😎 nice

showcase/app/controllers/page-components/advanced-table.ts
```diff
@@ -25,12 +27,12 @@ const customSortingCriteriaArray = [
...
 
-const updateModelWithSelectAllState = <T>(
-  modelData: T[],
+const updateModelWithSelectAllState = (
+  modelData: SelectableItem[] | User[],
   selectAllState: boolean,
 ) => {
   modelData.forEach((modelRow) => {
-    if (modelRow instanceof Object && 'isSelected' in modelRow) {
+    if (modelRow instanceof Object) {
```
dchyun, line 35: [Note] The `userData` data doesn't have `isSelected` set, so this check was breaking the ["delete row" example's](https://hds-showcase.vercel.app/components/advanced-table#delete-selected-rows) "expand scope" toggle, which used the userData and this function. The type has been updated to be more specific to ensure `modelRow.isSelected` is available.

showcase/app/routes/page-components/advanced-table.ts
```diff
@@ -42,7 +41,6 @@ export default class PageComponentsAdvancedTableRoute extends Route {
       userDataDemo4: clone(
         users.slice(0, 4).map((user) => ({ ...user, isAnimated: false })),
       ),
-      manyColumns: userWithMoreColumns,
```
dchyun, line 45: [Note] This data didn't appear to be being used anywhere in the adv table so I've removed it from the model

shleewhite, line 45: Thank you for catching this! Didn't realize its not used in the template.

PR #3070, opened 2025-07-30
Showcase: convert RichTooltip route to TS

shleewhite:
### :pushpin: Summary

If merged, this PR would convert the RichTooltip route to TypeScript.

### :link: External links

***

<details>

</details>

### DIFF
--- a/showcase/app/routes/page-components/rich-tooltip.ts
+++ b/showcase/app/routes/page-components/rich-tooltip.ts
@@ -5,7 +5,18 @@
 
 import Route from '@ember/routing/route';
 
-import { PLACEMENTS } from '@hashicorp/design-system-components/modifiers/hds-anchored-position';
+import { SIZES as TOGGLE_SIZES } from '@hashicorp/design-system-components/components/hds/rich-tooltip/toggle';
+import {
+  PLACEMENTS,
+  ENABLE_COLLISION_DETECTION_OPTIONS,
+} from '@hashicorp/design-system-components/modifiers/hds-anchored-position';
+
+import type { ModelFrom } from 'showcase/utils/ModelFromRoute';
+
+export type PageComponentsRichTooltipModel =
+  ModelFrom<PageComponentsRichTooltipRoute>;
+
+type TextAlignOptions = 'left' | 'center' | 'right';
 
 export default class PageComponentsRichTooltipRoute extends Route {
   model() {
@@ -47,10 +58,15 @@ export default class PageComponentsRichTooltipRoute extends Route {
       },
     ];
 
+    const TEXT_ALIGNMENTS: TextAlignOptions[] = ['left', 'center', 'right'];
+
     return {
       PLACEMENTS,
       STATES,
       DEMO_TABLE_DATA,
+      TOGGLE_SIZES,
+      ENABLE_COLLISION_DETECTION_OPTIONS,
+      TEXT_ALIGNMENTS,
     };
   }
 }
--- a/showcase/app/templates/page-components/rich-tooltip.hbs
+++ b/showcase/app/templates/page-components/rich-tooltip.hbs
@@ -44,41 +44,60 @@
 
   <Shw::Text::H4>Sizes</Shw::Text::H4>
 
-  {{#let (array "inherited" "small" "medium" "large") as |sizes|}}
-    {{#each sizes as |size|}}
-      <Shw::Flex @label="{{capitalize size}} {{if (eq size 'inherited') '(default)'}}" as |SF|>
-        <SF.Item>
-          <Hds::RichTooltip as |RT|>
-            <RT.Toggle @text="Lorem ipsum dolor" @size={{if (eq size "inherited") undefined size}} />
-            <RT.Bubble>
-              <Shw::Placeholder @text="generic content" @height="40" />
-            </RT.Bubble>
-          </Hds::RichTooltip>
-        </SF.Item>
-        <SF.Item>
-          <Hds::RichTooltip as |RT|>
-            <RT.Toggle @text="Lorem ipsum dolor" @size={{if (eq size "inherited") undefined size}} @icon="info" />
-            <RT.Bubble>
-              <Shw::Placeholder @text="generic content" @height="40" />
-            </RT.Bubble>
-          </Hds::RichTooltip>
-        </SF.Item>
-        <SF.Item>
-          <Hds::RichTooltip as |RT|>
-            <RT.Toggle
-              @text="Lorem ipsum dolor"
-              @size={{if (eq size "inherited") undefined size}}
-              @icon="info"
-              @iconPosition="leading"
-            />
-            <RT.Bubble>
-              <Shw::Placeholder @text="generic content" @height="40" />
-            </RT.Bubble>
-          </Hds::RichTooltip>
-        </SF.Item>
-      </Shw::Flex>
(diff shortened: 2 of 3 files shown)


shleewhite:
The Percy changes are expected. Instead of using an array defined in the handlebars file of collision detection options - it uses the type defined in the component which is a slightly different order.

showcase/app/templates/page-components/rich-tooltip.hbs
```diff
@@ -305,22 +324,20 @@
...
-                <RT.Bubble @enableCollisionDetection={{detection}}>
-                  <Shw::Placeholder @text="generic content" @height="30" />
-                </RT.Bubble>
-              </Hds::RichTooltip>
-            </div>
-          </Shw::Autoscrollable>
-        </SF.Item>
-      {{/each}}
-    {{/let}}
+    {{#each @model.ENABLE_COLLISION_DETECTION_OPTIONS as |detection|}}
```
dchyun, line 327: [Note] Currently, the collision detection examples are not working for all showcase pages of components using the `PopoverPrimitive. There is a ticket to look into this https://hashicorp.atlassian.net/browse/HDS-5092.

dchyun (approved):

showcase/app/routes/page-components/rich-tooltip.ts
```diff
@@ -5,7 +5,18 @@
...
+  PLACEMENTS,
+  ENABLE_COLLISION_DETECTION_OPTIONS,
+} from '@hashicorp/design-system-components/modifiers/hds-anchored-position';
+
+import type { ModelFrom } from 'showcase/utils/ModelFromRoute';
+
+export type PageComponentsRichTooltipModel =
+  ModelFrom<PageComponentsRichTooltipRoute>;
+
+type TextAlignOptions = 'left' | 'center' | 'right';
```
didoo, line 19: [question] would this suggestion apply also here? https://github.com/hashicorp/design-system/pull/3063#discussion_r2242198195

showcase/app/templates/page-components/rich-tooltip.hbs
```diff
@@ -1114,12 +1212,20 @@
       </Shw::Outliner>
     </SF.Item>
     <SF.Item @label="Bubble (popover)" class="shw-component-rich-tooltip-bubble-standalone">
-      <Hds::RichTooltip::Bubble @isOpen={{true}}>
+      {{! @glint-expect-error - @setupPrimitivePopper is a required arg, but we are not passing to test the bubble in isolation }}
```
didoo, line 1215: @shleewhite @dchyun in one of the PRs I've seen recently the author added a wrapper to solve this issue (I don't remember anymore where) but I think Lee's approach is better here because, despite being "incorrect/incomplete" from a TS perspective, allows to test the (sub)component in isolation. Can you change that PR too to use the same approach? (or if already merged, open a new one to revert that specific change)

showcase/app/templates/page-components/rich-tooltip.hbs
```diff
@@ -44,41 +44,60 @@
...
+    <SF.Item>
+      <Hds::RichTooltip as |RT|>
+        <RT.Toggle @text="Lorem ipsum dolor" @icon="info" @iconPosition="leading" />
+        <RT.Bubble>
+          <Shw::Placeholder @text="generic content" @height="40" />
+        </RT.Bubble>
+      </Hds::RichTooltip>
+    </SF.Item>
+  </Shw::Flex>
+  {{#each @model.TOGGLE_SIZES as |size|}}
```
didoo, line 73: [thought] since this is going to happen more and more frequently, do you think creating helpers (only for the showcase) that can append/prepend items to an existing array could help? If we were to prepend `inherited` to the `@model.TOGGLE_SIZES` via a `{{#let}}` that defines the `sizes` array, and then leave the code as it was before, would TS complain?

showcase/app/templates/page-components/rich-tooltip.hbs
didoo: should we add a comment in the code above (and below) mentioning this?

showcase/app/templates/page-components/rich-tooltip.hbs
```diff
@@ -963,7 +987,7 @@
             <Hds::Text::Body @size="200" @tag="p">Lorem ipsum dolor sit amet consectetur adipisicing elit. Sed
               perferendis culpa expedita assumenda at
               <Hds::RichTooltip as |RT|>
-                <RT.Toggle @etxt="nisi minus unde fuga iure" @isInline={{true}} @icon="info" @iconPosition="leading" />
+                <RT.Toggle @text="nisi minus unde fuga iure" @isInline={{true}} @icon="info" @iconPosition="leading" />
```
didoo, line 990: LOL

showcase/app/templates/page-components/rich-tooltip.hbs
didoo: OK, it was this one 😂
So the question is: why the two approaches? (eg. line 1080)

didoo (commented):
Left a few comments

showcase/app/routes/page-components/rich-tooltip.ts
shleewhite: I thought about this... but it is really just for the showcase. It is creating a type so there is not an error with the StyleModifier. I'd rather leave it as its own thing since theres no component to associate it with. See where its used here: https://github.com/hashicorp/design-system/blob/5f3dd98b2e594db0503df572ecfea6555df4e09e/showcase/app/templates/page-components/rich-tooltip.hbs#L498

showcase/app/templates/page-components/rich-tooltip.hbs
shleewhite: TS wouldnt complain if it was done correctly (like we check if the size is inherited and set size to undefined) but I honestly haven't seen any other instance where we would need to prepend an item to an array like this.

Could also remove the inherited example because there are a ton of examples of RichTooltips with size undefined in this page.

showcase/app/templates/page-components/rich-tooltip.hbs
shleewhite: I dont think we need to, its an existing bug and doesnt require changing any code on this page once its fixed.

showcase/app/templates/page-components/rich-tooltip.hbs
shleewhite: So its because of the way the PopoverPrimitive works. You can have a PopoverPrimitive without the bubble but you cant have one without the trigger.

The easiest way to fix the typescript errors on 1080 was just to wrap it in the PopoverPrimitive and pass the required args to RichTooltip::Toggle correctly.

However, if I wrap the RichToolTip::Bubble with the PopoverPrimitive, I would have to add a trigger otherwise it wouldn't render. So I decided to just add a glint-expect-error for the modifier it expects the PopoverPrimitive to pass to it.

I can definitely add a comment explaining this within the file so it is clearer.

didoo: OK, clear.

showcase/app/routes/page-components/rich-tooltip.ts
didoo: OK, clear.

PR #3074, opened 2025-07-31
Showcase: create `mocks/` folder  to simplify getting data for showcase pages

shleewhite:
### :pushpin: Summary

If merged, this PR would create a `mocks/` folder that has one file per data type. Each file exports the type of the record and an array of the data.

This PR also updates the existing component pages to use these mocks (AdvancedTable, Table, Pagination)

### :link: External links

Jira ticket: [HDS-5159](https://hashicorp.atlassian.net/browse/HDS-5159)

***

<details>

</details>

[HDS-5159]: https://hashicorp.atlassian.net/browse/HDS-5159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


showcase/app/utils/fetchJson/index.ts
aklkv, line 1: why not fetch-json.ts ? also the pattern is to export default

showcase/app/utils/fetchJson/types.ts
aklkv, line 1: unlike components package, you can put this types into the same file as a function or better yet from organizational standpoint you can do something like:

app/
 types/
  api.ts 

we use this pattern in cloud-ui to reexport api definitions so you can use it like

`import type { MyInterface } from 'showcase/types/api';`

e.g. https://github.com/hashicorp/cloud-ui/blob/main/engines/infragraph/addon/types/api.ts#L1

aklkv:
In general this is great clean up and moving in a good direction 🎉 

[future-idea] I think there is more opportunity to simplify this with usage of https://github.com/tracked-tools/ember-async-data some time in the future this addon can be used to load data per component. 

Carry on 👍🏻

showcase/app/routes/page-components/advanced-table.ts
```diff
@@ -24,149 +28,27 @@ export const clone = <T>(arr: T[]): T[] => {
...
-    const spanningManualData =
-      (await responseSpanningManual.json()) as SpanningEntity[];
-    const nestedData = (await responseNested.json()) as Policy[];
-    const nestedDataCustom =
-      (await responseNestedCustom.json()) as PolicyCustom[];
-
+  model() {
     return {
-      music: music.map((record) => ({ id: record.id, ...record.attributes })),
+      music: folkMusic.map((record) => ({
```
aklkv: you could potantially flattn this in the mock file it self, I understand that original json was copied from some sort of ember tutorial which showed how JSONAPI data would look like but we want flat structure so you don't have to do this manipulation

shleewhite: Ah, I think that would be good to do. Will update.

showcase/app/mocks/folk-data.ts
```diff
@@ -0,0 +1,134 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import type {
+  HdsBadgeColors,
+  HdsBadgeTypes,
+} from '@hashicorp/design-system-components/components/hds/badge/types';
+import type { HdsIconSignature } from '@hashicorp/design-system-components/components/hds/icon/index';
+
+export interface MusicEntity {
```
dchyun: [Suggestion] Maybe this would be a good time to clean up some of this data and remove values that aren't being used in any of the existing examples. From looking at the table and adv table I think we could remove `type` and `color` and simplify `badge-color`. 
```
export interface MusicEntity {
  id: string;
  artist: string;
  album: string;
  year: number;
  quote: string;
  'vinyl-cost': string;
  icon: HdsIconSignature['Args']['name'];
  'badge-type': HdsBadgeTypes;
  'badge-color': HdsBadgeColors;
}
```

dchyun (approved):
One non-blocking comment, but nice to see this clean up!

showcase/app/mocks/folk-data.ts
shleewhite: Ah! Good point. I will simplify.

showcase/app/mocks/cluster-data.ts
didoo, line 1: [thought] how about, instead of calling the folder/files `mock/xyx-data` we use `mock-data/xyz`?

showcase/app/mocks/folk-data.ts
didoo, line 1: [todo] we should standardize the naming
- the file is called `folk-data`
- the interface is called `MusicEntity`
- the exported data array is called `FolkMusic`

let's pick one ;)

showcase/app/mocks/many-columns-data.ts
didoo, line 1: [suggestion] for consistency with the `user-data.ts` file, this file should be called `user-with-more-columns[-data]`

showcase/app/mocks/policy-custom-key-data.ts
didoo, line 1: [todo] rename `policies[-data]`?

showcase/app/mocks/policy-custom-key-data.ts
```diff
@@ -0,0 +1,72 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+export interface PolicyCustomData {
```
didoo:
```suggestion
export interface Policy {
```

showcase/app/mocks/policy-custom-key-data.ts
```diff
@@ -0,0 +1,72 @@
...
+  id: number;
+  name: string;
+  description: string;
+  status: string;
+  isOpen?: boolean;
+  data?: PolicyCustomData[];
+}
+
+const Policies: PolicyCustomData[] = [
```
didoo:
```suggestion
const Policies: Policy[] = [
```

showcase/app/mocks/policy-custom-key-data.ts
didoo: [question] Are we sure we need both `policy-custom-key-data.ts` and `policy-data.ts`? They're 99% the same.

screenshot_5263

showcase/app/mocks/selectable-data.ts
didoo, line 1: [todo] rename `selectable-items[-data]`

showcase/app/mocks/spanning-cell-data.ts
```diff
@@ -0,0 +1,87 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+export interface SpanningEntity {
```
didoo:
```suggestion
export interface SpanningCell {
```

showcase/app/mocks/spanning-cell-data.ts
```diff
@@ -0,0 +1,87 @@
...
+  description?:
+    | {
+        text: string;
+        rowspan?: number;
+      }
+    | string;
+  email?: string;
+}
+
+const SpanningCells: SpanningEntity[] = [
```
didoo:
```suggestion
const SpanningCells: SpanningCell[] = [
```

didoo (commented):
Left a few questions/suggestions.

I have a general question, though: before the data was loaded asynchronously, while now it's embedded in the JS bundle. This means that _technically_ it's re-compiled every time? Do we know (as in: have we checked the data) if this slows down even further the compilation, which is already slow TBH, expecially when developing/updating a component and testing the changes in the browser?

btw @shleewhite with the current implementation, you should update the title of the PR which is not accurate anymoe

aklkv:
> Left a few questions/suggestions.
> 
> I have a general question, though: before the data was loaded asynchronously, while now it's embedded in the JS bundle. This means that _technically_ it's re-compiled every time? Do we know (as in: have we checked the data) if this slows down even further the compilation, which is already slow TBH, expecially when developing/updating a component and testing the changes in the browser?

Indeed, the data is not bundled together when loaded, but if anything, this setup will make test execution fractionally faster.

Looking ahead, my suggestion is to generate test data using [Faker.js](https://fakerjs.dev/), which would significantly reduce the size of these files. You could even use factories to create an infinite variety of data, enabling more robust stress testing of individual components.

As a more ambitious idea, we could leverage [Ember Mirage](https://miragejs.com/)—something we already use in other projects—to serve data via a mock API. However, this introduces a large and complex setup, and it’s unclear whether the benefits justify the added overhead.

At the end of the day, you don’t need to care about full page loads from a component’s perspective. What truly matters is how a component renders under different scenarios—which this setup supports and can be further enhanced by integrating Faker.js.

showcase/app/mocks/cluster-data.ts
shleewhite: I was under the impression I was following the ember conventions by naming it `mocks/`

showcase/app/mocks/policy-custom-key-data.ts
shleewhite: I had both to test having a custom children key for nested AdvancedTable. The default is "children", but custom key uses "data".

didoo: since it's used only in one file, would make sense to have a function in the route file, that imports the standard data array, deep clones it and then changes the key name, instead of duplicating the file?

showcase/app/templates/page-components/advanced-table.hbs
```diff
@@ -59,7 +60,7 @@
         </B.Td>
         <B.Td>
           {{! @glint-expect-error - will be fixed by https://hashicorp.atlassian.net/browse/HDS-5090}}
-          <Hds::Badge @text={{B.data.year}} @type={{B.data.badge-type}} @color={{B.data.badge-color.name}} />
+          <Hds::Badge @text={{B.data.year}} @type="outlined" @color={{B.data.badge-color.name}} />
```
dchyun:
```suggestion
          <Hds::Badge @text={{B.data.year}} @type="outlined" @color={{B.data.badge-color}} />
```
[Issue] With the change in the structure of the music data, any references to badge color need to be updated. This applies both to the adv table and table templates. Currently the badge color is not getting displayed.

showcase/app/mocks/folk-music-data.ts
```diff
@@ -0,0 +1,96 @@
...
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import type { HdsBadgeColors } from '@hashicorp/design-system-components/components/hds/badge/types';
+import type { HdsIconSignature } from '@hashicorp/design-system-components/components/hds/icon/index';
+
+export interface FolkMusicEntity {
+  id: string;
+  type: string;
```
dchyun:
```suggestion
```
[Suggestion][non-blocker] the `type` value isn't used in any examples. It could be removed, unless we potentially want to leverage it in the future with an expanded data set.

showcase/app/mocks/policy-custom-key-data.ts
shleewhite: ah ok I understand, updated!

showcase/app/mocks/folk-music-data.ts
```diff
@@ -0,0 +1,89 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import type { HdsBadgeColors } from '@hashicorp/design-system-components/components/hds/badge/types';
+import type { HdsIconSignature } from '@hashicorp/design-system-components/components/hds/icon/index';
+
+export interface FolkMusicEntity {
```
didoo: `cluster-data` → `Cluster`
`folk-music-data` → `FolkMusic*Entity*`

showcase/app/mocks/cluster-data.ts
```diff
@@ -0,0 +1,79 @@
...
+  'peer-name': string;
+  'cluster-partition': string;
+  status: string;
+  services: {
+    imported: number;
+    exported: number;
+  };
+}
+
+const Clusters: Cluster[] = [
```
didoo: [question] should the const be lowercase? it's not a type
```suggestion
const clusters: Cluster[] = [
```

showcase/app/routes/page-components/advanced-table.ts
```diff
@@ -24,149 +27,30 @@ export const clone = <T>(arr: T[]): T[] => {
...
-      manyColumns,
-      nestedData,
-      nestedDataCustom,
+      manyColumns: userWithManyColumnsData,
+      nestedData: policyData,
+      nestedDataCustom: policyData.map((policy) => {
+        const { children, ...rest } = policy;
+        return {
+          ...rest,
+          data: children,
```
didoo, line 52: [question] do you know if we modify any of the `children` or `data` items? otherwise we have to deep clone it (it's an array of objects)

shleewhite: The example in the showcase seems to be working just fine, the data is never modified/updated after this.

showcase/app/routes/page-components/advanced-table.ts
```diff
@@ -6,14 +6,17 @@
...
 import { DENSITIES } from '@hashicorp/design-system-components/components/hds/advanced-table/index';
-import type {
-  HdsBadgeColors,
-  HdsBadgeTypes,
-} from '@hashicorp/design-system-components/components/hds/badge/types';
-import type { HdsIconSignature } from '@hashicorp/design-system-components/components/hds/icon/index';
 
 import type { ModelFrom } from 'showcase/utils/ModelFromRoute';
 
+import clusterData from 'showcase/mocks/cluster-data';
```
dchyun:
```suggestion
import clusters from 'showcase/mocks/cluster-data';
```
[Issue] It looks like all the import names in the adv table, table, and pagination files are slightly off. Not sure how it's working with the current names.

showcase/app/routes/page-components/advanced-table.ts
```diff
@@ -6,14 +6,17 @@
...
+import folkMusicData from 'showcase/mocks/folk-music-data';
+import userWithManyColumnsData from 'showcase/mocks/user-with-more-columns-data';
```
dchyun:
```suggestion
import usersWithMoreColumns from 'showcase/mocks/user-with-more-columns-data';
```
[Issue] There's a naming mismatch here and with the model below where the data says "more" and here it says "many". Not sure if it'd be easier to update the data file or the imports, but some alignment should happen.

showcase/app/routes/page-components/advanced-table.ts
shleewhite: its working because it is the default export, you can import it with any name. I'll update tho for consistency.

showcase/app/routes/page-components/advanced-table.ts
```diff
@@ -24,149 +27,30 @@ export const clone = <T>(arr: T[]): T[] => {
...
       userDataDemo4: clone(
-        userData.slice(0, 4).map((user) => ({ ...user, isAnimated: false })),
+        users.slice(0, 4).map((user) => ({ ...user, isAnimated: false })),
       ),
-      manyColumns,
-      nestedData,
-      nestedDataCustom,
+      manyColumns: userWithMoreColumns,
+      nestedData: policies,
+      nestedDataCustom: policies.map((policy) => {
```
didoo, line 47: [todo] One of the item, in the previous JSON, had `isOpen: true` so the VRT test would test also an opened nested row, you should add it back in this logic here (see Percy diff)
screenshot_5268

showcase/app/mocks/selectable-items-data.ts
didoo, line 1: [nitpick] the file should be called `selectable-item-data.ts` for consistency with the others (singular)

dchyun (approved):

PR #3075, opened 2025-07-31
`ButtonSet` - Update  docs (HDS-5238)

KristinLBradley:
### :pushpin: Summary

If merged, this PR updates web docs for the `ButtonSet` clarifying usage examples.

👉 **Preview**: https://hds-website-git-kristin-hds-5238-update-button-d71644-hashicorp.vercel.app/components/button-set?tab=code

### :link: External links

* Jira ticket: [HDS-5238](https://hashicorp.atlassian.net/browse/HDS-5238)

***

<details>

</details>

[HDS-5238]: https://hashicorp.atlassian.net/browse/HDS-5238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/docs/components/button-set/partials/code/how-to-use.md
+++ b/website/docs/components/button-set/partials/code/how-to-use.md
@@ -15,7 +15,7 @@ The basic invocation `ButtonSet` should include two or more buttons to be provid
 </Hds::ButtonSet>
 ```
 
-Other button-like components can also be used as children such as the [Dropdown](/components/dropdown).
+Other button-like components, such as the [Dropdown](/components/dropdown), can also be used as children.
 
 ```handlebars
 <Hds::ButtonSet>
@@ -42,31 +42,6 @@ If you want buttons with equal widths, set `@isFullWidth` to `true` on the `Butt
 <Hds::ButtonSet {{style maxWidth="15rem"}}>
   <Hds::Button @text="Save" @isFullWidth={{true}} />
   <Hds::Button @text="Cancel" @color="secondary" @href="https://hashicorp.com" @isFullWidth={{true}} />
-</Hds::ButtonSet>
-```
-
-!!! Warning
-
-**Code alert**
-
-The `Dropdown` does not have an `isFullWidth` option for the outer wrapper. Setting isFullWidth to “true” on the nested ToggleButton does not have the same effect.
-
-!!!
-
-```handlebars
-<Hds::ButtonSet>
-  <Hds::Dropdown as |D|>
-    <D.ToggleButton @isFullWidth={{true}} @color="secondary" @text="Select an option" />
-    <D.Title @text="Title Text" />
-    <D.Description @text="Descriptive text goes here." />
-    <D.Interactive @href="#">Add</D.Interactive>
-    <D.Interactive @href="#">Add More</D.Interactive>
-    <D.Interactive @href="#">Add Another Thing Too</D.Interactive>
-    <D.Separator />
-    <D.Interactive @route="components" @icon="trash" @color="critical">Delete</D.Interactive>
-  </Hds::Dropdown>
-
-  <Hds::Button @isFullWidth={{true}} @text="Submit" type="submit" />
 </Hds::ButtonSet>
 ```
 
--- a/website/docs/components/button-set/partials/code/how-to-use.md
+++ b/website/docs/components/button-set/partials/code/how-to-use.md
@@ -1,6 +1,12 @@
 ## How to use this component
 
-The basic invocation should include two or more buttons to be provided as children:
+!!! Info
+
+For guidance on button organization, grouping of button types, and related information please refer to the [button organization](/patterns/button-organization) pattern documentation.
+
+!!!
+
+The basic invocation `ButtonSet` should include two or more buttons to be provided as children. (If used with only one button there will be no noticeable visual difference.)
 
 ```handlebars
 <Hds::ButtonSet>
@@ -9,7 +15,7 @@ The basic invocation should include two or more buttons to be provided as childr
 </Hds::ButtonSet>
 ```
 
-Other button-like components can also be used as children such as the `Dropdown`.
+Other button-like components can also be used as children such as the [Dropdown](/components/dropdown).
 
 ```handlebars
 <Hds::ButtonSet>
@@ -32,8 +38,6 @@ Other button-like components can also be used as children such as the `Dropdown`
 
 If you want buttons with equal widths, set `@isFullWidth` to `true` on the `Button` components. Since the ButtonSet is full-width (100%) by default, you will likely want to constrain its overall width by adding a `max-width` to the `ButtonSet` container (via an inline style or CSS class).
 
-Note: .
-
 ```handlebars
 <Hds::ButtonSet {{style maxWidth="15rem"}}>
   <Hds::Button @text="Save" @isFullWidth={{true}} />
@@ -41,7 +45,13 @@ Note: .
 </Hds::ButtonSet>
 ```
 
-Issue: The `Dropdown` does not have an `isFullWidth` option for the outer wrapper. Setting isFullWidth to “true” on the nested ToggleButton does not have the same effect.
+!!! Warning
+
+**Code alert**
+
+The `Dropdown` does not have an `isFullWidth` option for the outer wrapper. Setting isFullWidth to “true” on the nested ToggleButton does not have the same effect.
+
+!!!
 
 ```handlebars
 <Hds::ButtonSet>
--- a/website/docs/components/button-set/partials/code/how-to-use.md
+++ b/website/docs/components/button-set/partials/code/how-to-use.md
@@ -6,7 +6,7 @@ For guidance on button organization, grouping of button types, and related infor
 
 !!!
 
-The basic invocation `ButtonSet` should include two or more buttons to be provided as children. (If used with only one button there will be no noticeable visual difference.)
+The basic `ButtonSet` invocation should include two or more buttons to be provided as children. (If used with only one button there will be no noticeable visual difference.)
 
 ```handlebars
 <Hds::ButtonSet>
--- a/website/docs/components/button-set/partials/code/how-to-use.md
+++ b/website/docs/components/button-set/partials/code/how-to-use.md
@@ -1,6 +1,6 @@
 ## How to use this component
 
-The basic invocation requires two or more buttons to be provided as children:
+The basic invocation should include two or more buttons to be provided as children:
 
 ```handlebars
 <Hds::ButtonSet>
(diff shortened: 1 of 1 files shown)


website/docs/components/button-set/partials/code/how-to-use.md
```diff
@@ -9,17 +9,57 @@ The basic invocation requires two or more buttons to be provided as children:
...
 
 ```handlebars
-<Hds::ButtonSet {{style width="15rem"}}>
+<Hds::ButtonSet {{style maxWidth="15rem"}}>
   <Hds::Button @text="Save" @isFullWidth={{true}} />
   <Hds::Button @text="Cancel" @color="secondary" @href="https://hashicorp.com" @isFullWidth={{true}} />
 </Hds::ButtonSet>
 ```
 
+Issue: The `Dropdown` does not have an `isFullWidth` option for the outer wrapper. Setting isFullWidth to “true” on the nested ToggleButton does not have the same effect.
```
KristinLBradley: I included this example temporarily to show an issue with the Dropdown:
image

We may want to add an `isFullWidth` option to the Dropdown itself but curious what others think...

website/docs/components/button-set/partials/code/how-to-use.md
```diff
@@ -1,6 +1,6 @@
 ## How to use this component
 
-The basic invocation requires two or more buttons to be provided as children:
+The basic invocation should include two or more buttons to be provided as children:
```
KristinLBradley: Technically, you could add just one Button although there's not much point.

website/docs/components/button-set/partials/code/how-to-use.md
```diff
@@ -9,17 +9,57 @@ The basic invocation requires two or more buttons to be provided as children:
 </Hds::ButtonSet>
 ```
 
+Other button-like components can also be used as children such as the `Dropdown`.
```
KristinLBradley: I could add other examples if others have ideas.

website/docs/components/button-set/partials/code/how-to-use.md
heatherlarsen: Do we know how often this happens?

heatherlarsen: Maybe this should be a warning banner?

```suggestion
!!! Warning 

**Code alert**

The `Dropdown` does not have an `isFullWidth` option for the outer wrapper. Setting isFullWidth to “true” on the nested ToggleButton does not have the same effect.
!!!
```

KristinLBradley: Yeah, I added this mainly to show the issue so we could discuss a possible solution such as adding `isFullWidth` to the `Dropdown`. That would have to be done separately though so I guess we actually may want to keep this and turn it into a warning. I'll hold off changing it for now until we can discuss more though.

KristinLBradley: I first noticed it when reviewing @melanie's Helios demo PR. In any case, I think it's confusing that the `Dropdown` itself doesn't include an `@isFullWidth` option like the `Button` does.

KristinLBradley: I'm not seeing examples within consumer code bases but it's still a legitimate use case.

https://github.com/search?q=org%3Ahashicorp+ButtonSet%3E&type=code

website/docs/components/button-set/partials/code/how-to-use.md
jorytindall: Agreed, this softer language is better.

website/docs/components/button-set/partials/code/how-to-use.md
jorytindall: I agree, turning it into a warning makes sense given the context.

website/docs/components/button-set/partials/code/how-to-use.md
jorytindall: I can't think of anything else that _should_ go here (Dropdown makes the most sense), but do you think it's worth it (either here, or in the above section) to include something around this type of content or expectations:
- Shouldn't use two primary buttons
- Recommend adhering to a primary + secondary pattern, with the occasional tertiary
- Basically referencing the button organization pattern docs. 
I'm not sure how useful it is, but given that there are scenarios out there like what I listed maybe good to at a minimum cross link between the two docs (since it's on the consumer to implement correctly)

jorytindall (commented):
I think this looks good and adds a nice contextual update, just adding the banner would be my only other recommendation. Great idea adding the additional context!

website/docs/components/button-set/partials/code/how-to-use.md
MelSumner: @heatherlarsen here's what we were seeing: 

CleanShot 2025-08-01 at 09 19 21@2x

CleanShot 2025-08-01 at 09 22 50@2x


What we'd like to be able to ensure: 
CleanShot 2025-08-01 at 09 24 09@2x

website/docs/components/button-set/partials/code/how-to-use.md
```diff
@@ -9,17 +9,57 @@ The basic invocation requires two or more buttons to be provided as children:
...
+  <Hds::Button @text="Submit" type="submit" />
+</Hds::ButtonSet>
+```
+
 ### Equal width buttons
 
-If you want to have buttons with equal width, apply a width (via inline style or CSS class) to the `ButtonSet` container and the argument `@isFullWidth={{true}}` to the `Button` components:
+If you want buttons with equal widths, set `@isFullWidth` to `true` on the `Button` components. To constrain the overall max-width, you can also add a width to the `ButtonSet` container (via an inline style or CSS class).
+
+Note: The ButtonSet is full-width (100%) by default.
```
didoo: [suggestion] how about something like this? I know that technically just setting the `@isFullWidth` on the buttons will make them equal size, but they will also blow up, which is unlikely what a consumer will want; so setting a max-width on the `ButtonSet` is practically inevitable.
```suggestion
If you want buttons with equal widths, set `@isFullWidth` to `true` on the `Button` components. Since the ButtonSet is full-width (100%) by default, you will likely want to constrain its overall width by adding a `max-width` to the `ButtonSet` container (via an inline style or CSS class).

Note: .
```

website/docs/components/button-set/partials/code/how-to-use.md
didoo: [thought] I am not entirely convinced we should frame this as an "issue". It's not a problem per se, it's just how `flexbox` works (`ButtonSet` is a flex container). Technically, if one wants to have all the children of a `flex` with equal size, they should apply a `flex: 1 1 0` to each one of them. Applying `isFullWidth` to all the children of `ButtonSet` is just a workaround. Which makes me wonder if we should really add that property to the `Dropdown` container (there are already different factors playing into the width of the toggle and the list: https://hds-showcase.vercel.app/components/dropdown#width and this extra top-level width could interact with them in unexpected ways)

In @MelSumner case, the "solution" I don't think is to set `isFullWidth` on the button/dropdown, but have a way for the cards to wrap when there is not enough space (eg, by applying a min-width or width to the cards, and letting the flex items to wrap).

What we _could_ do, ideally, is to have an optional child yielded (`ButtonSet::Item` / `BS.Item`) like we do for the `Layout::Flex` and `Layout::Grid` exactly for this reason.

Happy to chat more in depth, if needed.

didoo (commented):
Left a couple of comment, but one is a major point of discussion for me.

website/docs/components/button-set/partials/code/how-to-use.md
KristinLBradley: @jorytindall thanks.

@MelSumner I made this change because I thought it was more accurate, not softer.

website/docs/components/button-set/partials/code/how-to-use.md
KristinLBradley: Could be useful to add a bit of that and/or reference the Button pattern docs. I'll take a look...

website/docs/components/button-set/partials/code/how-to-use.md
KristinLBradley: Thanks, I think the additional explanation is helpful. Will add.

website/docs/components/button-set/partials/code/how-to-use.md
KristinLBradley: @MelSumner Please suggest other language if you don't agree with this update.

website/docs/components/button-set/partials/code/how-to-use.md
KristinLBradley: We could explore updating the ButtonSet to add additional arguments or child components or something. I do think it's confusing that the Dropdown doesn't support the `@isFullWidth` argument though which the `Button` does.

The language in the text here "Issue" was mainly meant to be temporary for us to discuss. Since we won't make actual code changes within this PR though I can edit it to be more appropriate for consumers.

KristinLBradley: I went ahead and made this change locally, thanks.

website/docs/components/button-set/partials/code/how-to-use.md
KristinLBradley: I added an info banner at the top of the page with a link to the pattern docs. Let me know if you think using a banner makes it overly prominent or anything.

image

website/docs/components/button-set/partials/code/how-to-use.md
KristinLBradley: I updated this section to make it into a warning as @heatherlarsen & @jorytindall recommended.

Should I open a ticket to investigate making updates either to the `ButtonSet` and/or `Dropdown` as a follow-on? @didoo

website/docs/components/button-set/partials/code/how-to-use.md
KristinLBradley: I updated this part further to mention that using a single button within a `ButtonSet` doesn't change the visual appearance.

website/docs/components/button-set/partials/code/how-to-use.md
```diff
@@ -9,17 +15,61 @@ The basic invocation requires two or more buttons to be provided as children:
...
 </Hds::ButtonSet>
 ```
 
+!!! Warning
+
+**Code alert**
+
+The `Dropdown` does not have an `isFullWidth` option for the outer wrapper. Setting isFullWidth to “true” on the nested ToggleButton does not have the same effect.
+
+!!!
```
KristinLBradley: @andgen404 Please let me know if this warning banner should be moved or changed in any way.

website/docs/components/button-set/partials/code/how-to-use.md
```diff
@@ -9,17 +15,61 @@ The basic invocation requires two or more buttons to be provided as children:
...
-<Hds::ButtonSet {{style width="15rem"}}>
+<Hds::ButtonSet {{style maxWidth="15rem"}}>
   <Hds::Button @text="Save" @isFullWidth={{true}} />
   <Hds::Button @text="Cancel" @color="secondary" @href="https://hashicorp.com" @isFullWidth={{true}} />
 </Hds::ButtonSet>
 ```
 
+!!! Warning
+
+**Code alert**
```
didoo: [thought] Sorry if I insist, but I still think this doesn't work/make sense as is now. 

[wearing the hat of a consumer] As a consumer of this documentation I see:
- an "Equal width buttons" title
- an explanation on how to achieve equal width buttons inside a ButtonSet
- a code snippet that demonstrates what it's explained above, that I can copy & use

but then I see:

- a warning "Code alert"  banner, that mentions a Dropdown (but I haven't seen the demo of the Dropdown yet)
- a code snippet that renders something that likely I don't want to see in production, and a code snippet that I should not be using (it's not the "correct" one to render equal width items)

there's no explanation or suggestions of how I would achieve that with the Dropdown, what is the "correct" way to implement, so all this part about the Dropdown not working as the button is not actionable for me, I am not sure what to make of it (it tells me what _not_ to do, but not what to do instead)

website/docs/components/button-set/partials/code/how-to-use.md
didoo: > Should I open a ticket to investigate making updates either to the `ButtonSet` and/or `Dropdown` as a follow-on? @didoo

For me, yes.

didoo (approved):
Left a general comment. I am not going to block this, since I see everyone approved, but I think we should give it another thought.

website/docs/components/button-set/partials/code/how-to-use.md
KristinLBradley: I can remove that part. I originally added it as a temporary way to show the issue and discuss within this PR to see what others thought about it. It was something I uncovered when I was looking at the Card layout within the HDS demo app and attempting to help Melanie and Anna solve issues with it.

website/docs/components/button-set/partials/code/how-to-use.md
andgen404: This reads as a code alert, so the banner type works well. It should be below the section title as well, since it's a warning to the consumer.

website/docs/components/button-set/partials/code/how-to-use.md
KristinLBradley: I went ahead and removed this example from the included updates as there are no consumers currently using Dropdowns within ButtonSets that I could uncover when doing a code search. I will open a separate ticket to investigate updating either the Dropdown or ButtonSet if appropriate.

andgen404 (approved):

website/docs/components/button-set/partials/code/how-to-use.md
```diff
@@ -1,6 +1,12 @@
 ## How to use this component
 
-The basic invocation requires two or more buttons to be provided as children:
+!!! Info
+
+For guidance on button organization, grouping of button types, and related information please refer to the [button organization](/patterns/button-organization) pattern documentation.
+
+!!!
+
+The basic invocation `ButtonSet` should include two or more buttons to be provided as children. (If used with only one button there will be no noticeable visual difference.)
```
MelSumner:
```suggestion
The basic `ButtonSet` invocation should include two or more buttons to be provided as children. (If used with only one button there will be no noticeable visual difference.)
```

website/docs/components/button-set/partials/code/how-to-use.md
KristinLBradley: I have since deleted this warning banner.

andgen404 (approved):

website/docs/components/button-set/partials/code/how-to-use.md
KristinLBradley: Created a ticket: https://hashicorp.atlassian.net/browse/HDS-5281

KristinLBradley: Created a follow-on ticket: https://hashicorp.atlassian.net/browse/HDS-5281

PR #3081, opened 2025-08-04
`AdvancedTable` Column resize bug fix

zamoore:
### :pushpin: Summary

If merged, this PR fixes a bug that was introduced by the column resize feature. The bug prevented the columns from expanding to fill available space if the viewport was expanded.

A key requirement of column resizing is that the columns being resized need to have an explicit pixel width in order to expand/contract with the cursor.

The previous implementation incorrectly applied the pixel width of the columns at render to the column configuration object. This prevented the table from filling the available space if the viewport was expanded.

The fix was to convert all column widths to pixels during the resizing process. Columns that were actively involved in resizing will have their new pixel widths "baked" into the column model after resizing is complete, whereas columns that were not involved will revert back to pre-resize values.

For example:

If I have a table with 4 columns and do not set any explicit width values on columns, each column will have an initial width of 1 fractional unit or `1fr`.

```
1fr 1fr 1fr 1fr
```

When I start to resize a column, each column will be assigned a transient width to be used during the resize. If the table is within a `1000px` container, each column will be assigned a width of `250px`.

```
250px 250px 250px 250px
```

If I expand the first column by `10px`, at the end of the resize event, the css grid values will be:

```
260px 240px 1fr 1fr
```

### :link: External links

Jira ticket: [HDS-5171](https://hashicorp.atlassian.net/browse/HDS-5171)

***

<details>

</details>

[HDS-5171]: https://hashicorp.atlassian.net/browse/HDS-5171?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/tender-carrots-fix.md
+++ b/.changeset/tender-carrots-fix.md
@@ -1,0 +1,8 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+<!-- START components/advanced-table -->
+
+- Fixed bug with automatic column resizing and scroll-shadow placement.
+<!-- END -->
--- a/.changeset/tender-carrots-fix.md
+++ b/.changeset/tender-carrots-fix.md
@@ -2,7 +2,6 @@
 "@hashicorp/design-system-components": minor
 ---
 
-<!-- START components/advanced-table -->
-
-- Fixed bug with automatic column resizing and scroll-shadow placement.
+<!-- START components/table/advanced-table -->
+`AdvancedTable` - Fixed bug with automatic column resizing and scroll-shadow placement.
 <!-- END -->
--- a/packages/components/src/components/hds/advanced-table/index.hbs
+++ b/packages/components/src/components/hds/advanced-table/index.hbs
@@ -6,7 +6,6 @@
   class="hds-advanced-table__container
     {{(if this.isStickyHeaderPinned 'hds-advanced-table__container--header-is-pinned')}}"
   {{did-update this.setupTableModelData @columns @model @sortBy @sortOrder}}
-  {{this._registerTableElement}}
   ...attributes
 >
   {{! Caption }}
--- a/packages/components/src/components/hds/advanced-table/index.hbs
+++ b/packages/components/src/components/hds/advanced-table/index.hbs
@@ -24,6 +24,7 @@
       --hds-advanced-table-sticky-column-offset=this.stickyColumnOffset
       max-height=@maxHeight
     }}
+    {{this._registerGridElement}}
     {{this._setUpScrollWrapper}}
   >
     {{! Header }}
--- a/packages/components/src/components/hds/advanced-table/index.hbs
+++ b/packages/components/src/components/hds/advanced-table/index.hbs
@@ -54,7 +54,6 @@
               @tableHeight={{this._tableHeight}}
               @onColumnResize={{@onColumnResize}}
               @onPinFirstColumn={{this._onPinFirstColumn}}
-              {{this._setColumnWidth column}}
             >
               {{column.label}}
             </Hds::AdvancedTable::ThSort>
@@ -74,7 +73,6 @@
               @onClickToggle={{this._tableModel.toggleAll}}
               @onColumnResize={{@onColumnResize}}
               @onPinFirstColumn={{this._onPinFirstColumn}}
-              {{this._setColumnWidth column}}
             >
               {{column.label}}
             </Hds::AdvancedTable::Th>
--- a/packages/components/src/components/hds/advanced-table/index.hbs
+++ b/packages/components/src/components/hds/advanced-table/index.hbs
@@ -55,6 +55,7 @@
               @tableHeight={{this._tableHeight}}
               @onColumnResize={{@onColumnResize}}
               @onPinFirstColumn={{this._onPinFirstColumn}}
+              {{this._registerThElement column}}
             >
               {{column.label}}
             </Hds::AdvancedTable::ThSort>
@@ -74,6 +75,7 @@
               @onClickToggle={{this._tableModel.toggleAll}}
               @onColumnResize={{@onColumnResize}}
               @onPinFirstColumn={{this._onPinFirstColumn}}
+              {{this._registerThElement column}}
             >
               {{column.label}}
             </Hds::AdvancedTable::Th>
--- a/packages/components/src/components/hds/advanced-table/index.hbs
+++ b/packages/components/src/components/hds/advanced-table/index.hbs
@@ -6,6 +6,7 @@
   class="hds-advanced-table__container
     {{(if this.isStickyHeaderPinned 'hds-advanced-table__container--header-is-pinned')}}"
   {{did-update this.setupTableModelData @columns @model @sortBy @sortOrder}}
+  {{this._registerTableElement}}
   ...attributes
 >
   {{! Caption }}
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -386,7 +386,6 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     // if there is a select checkbox, the first column has a 'min-content' width to hug the checkbox content
     let style = isSelectable ? 'min-content ' : '';
 
-    // check the custom column widths, if the current column has a custom width use the custom width. otherwise take the available space.
     for (let i = 0; i < columns.length; i++) {
       style += ` ${columns[i]!.appliedWidth}`;
     }
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -447,6 +447,14 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
   private _setUpScrollWrapper = modifier((element: HTMLDivElement) => {
     this._scrollWrapperElement = element;
 
+    const updateHorizontalScrollIndicators = () => {
+      if (element.clientWidth < element.scrollWidth) {
+        this.showScrollIndicatorRight = true;
+      } else {
+        this.showScrollIndicatorRight = false;
+      }
+    };
+
     this._scrollHandler = () => {
       this._updateScrollIndicators(element);
     };
@@ -482,6 +490,7 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     this._resizeObserver = new ResizeObserver((entries) => {
       entries.forEach(() => {
         updateMeasurements();
+        updateHorizontalScrollIndicators();
       });
     });
 
@@ -490,9 +499,7 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     updateMeasurements();
 
     // on render check if should show right scroll indicator
-    if (element.clientWidth < element.scrollWidth) {
-      this.showScrollIndicatorRight = true;
-    }
+    updateHorizontalScrollIndicators();
 
     // on render check if should show bottom scroll indicator
     if (element.clientHeight < element.scrollHeight) {
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -29,6 +29,7 @@ import type {
   HdsAdvancedTableModel,
   HdsAdvancedTableExpandState,
 } from './types.ts';
+import type HdsAdvancedTableColumnType from './models/column.ts';
 import type { HdsFormCheckboxBaseSignature } from '../form/checkbox/base.ts';
 import type HdsAdvancedTableTd from './td.ts';
 import type HdsAdvancedTableTh from './th.ts';
@@ -432,6 +433,16 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
 
     return classes.join(' ');
   }
+
+  private _registerThElement = modifier(
+    (element: HTMLDivElement, [column]: [HdsAdvancedTableColumnType]) => {
+      if (column === undefined) {
+        return;
+      }
+
+      column.thElement = element;
+    }
+  );
 
   private _setUpScrollWrapper = modifier((element: HTMLDivElement) => {
     this._scrollWrapperElement = element;
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -388,7 +388,7 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
 
     // check the custom column widths, if the current column has a custom width use the custom width. otherwise take the available space.
     for (let i = 0; i < columns.length; i++) {
-      style += ` ${columns[i]!.width}`;
+      style += ` ${columns[i]!.appliedWidth}`;
     }
 
     return style;
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -433,6 +433,10 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     return classes.join(' ');
   }
 
+  private _registerGridElement = modifier((element: HTMLDivElement) => {
+    this._tableModel.gridElement = element;
+  });
+
   private _registerThElement = modifier(
     (element: HTMLDivElement, [column]: [HdsAdvancedTableColumnType]) => {
       if (column === undefined) {
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -451,11 +451,7 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     this._scrollWrapperElement = element;
 
     const updateHorizontalScrollIndicators = () => {
-      if (element.clientWidth < element.scrollWidth) {
-        this.showScrollIndicatorRight = true;
-      } else {
-        this.showScrollIndicatorRight = false;
-      }
+      this.showScrollIndicatorRight = element.clientWidth < element.scrollWidth;
     };
 
     this._scrollHandler = () => {
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -434,6 +434,10 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     return classes.join(' ');
   }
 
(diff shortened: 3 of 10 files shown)


.changeset/tender-carrots-fix.md
```diff
@@ -0,0 +1,8 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+<!-- START components/advanced-table -->
+
```
shleewhite:
```suggestion
```

.changeset/tender-carrots-fix.md
```diff
@@ -0,0 +1,8 @@
...
+- Fixed bug with automatic column resizing and scroll-shadow placement.
```
shleewhite:
```suggestion
`AdvancedTable` - Fixed bug with automatic column resizing and scroll-shadow placement.
```

.changeset/tender-carrots-fix.md
```diff
@@ -0,0 +1,8 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+<!-- START components/advanced-table -->
```
shleewhite:
```suggestion
<!-- START components/table/advanced-table -->
```

we have to use the path for where the page on the website is here

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -352,23 +351,12 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
-
-    if (hasCustomColumnWidths) {
-      // check the custom column widths, if the current column has a custom width use the custom width. otherwise take the available space.
-      for (let i = 0; i < columns.length; i++) {
-        style += ` ${columns[i]!.width ?? DEFAULT_COLUMN_WIDTH}`;
-      }
-    } else {
-      // if there are no custom column widths, each column is the same width and they take up the available space
-      style += `repeat(${columns.length}, ${DEFAULT_COLUMN_WIDTH})`;
+    // check the custom column widths, if the current column has a custom width use the custom width. otherwise take the available space.
```
shleewhite:
```suggestion
```

Is this comment still relevant? Seems like no bc there is always an appliedWidth

packages/components/src/components/hds/advanced-table/th-resize-handle.ts
```diff
@@ -121,7 +138,9 @@ export default class HdsAdvancedTableThResizeHandle extends Component<HdsAdvance
 
   @action
   handleKeydown(event: KeyboardEvent): void {
-    if (event.key !== 'ArrowLeft' && event.key !== 'ArrowRight') {
+    const validKeys = ['ArrowLeft', 'ArrowRight'];
```
shleewhite, line 141: Issue: The resize handle is no longer working with a keyboard

packages/components/src/components/hds/advanced-table/models/column.ts
```diff
@@ -123,31 +149,65 @@ export default class HdsAdvancedTableColumn {
...
+
+    if (thisKey === undefined) {
+      return;
+    }
+
+    table.columns.forEach((otherColumn) => {
+      const debtToCollect = otherColumn.widthDebts[thisKey] ?? 0;
+
+      if (debtToCollect > 0) {
+        // Take the width back from the column that owes us
```
dchyun:
```suggestion
        // If debtToCollect > available width to give, collect debts owed to otherColumn so full debtToCollect can be paid
          // Take the width back from the column that owes us
```
[Issue] This is making me feel like a mob boss accountant, but there is an issue where columns can go below their min-width when others are resized. The issue is that the debt a column can pay out can end up being greater than the actual width they have to give when multiple columns are resized. 

I think what needs to happen is almost some kind of repayment plan where if a column has to pay more than it has available then it needs to go collect its own debts first.

**Example**
- B can only afford to pay 25 without dropping below min-width
- B owes C 50
- A owes B 50
- C requests payment
- Current: B pays C 50, B's width drops below min
- Expected: B collects 25 from A and pays 25 of its own 

**Video examples**

https://github.com/user-attachments/assets/cd6c2aed-2cc4-419c-a82a-44ed993cbc7e


https://github.com/user-attachments/assets/d41fd580-9390-4b76-8caf-77edc18d69b0

packages/components/src/components/hds/advanced-table/th-resize-handle.ts
dchyun: Hmm for me it's working as expected.

packages/components/src/components/hds/advanced-table/index.ts
zamoore: Yeah, you're right

packages/components/src/components/hds/advanced-table/th-resize-handle.ts
zamoore: I'm also seeing it working

packages/components/src/components/hds/advanced-table/models/column.ts
zamoore: I have imagined so many mafia memes while writing this code. This is a good catch though. I'll need to investigate this. Thank you for the thorough review and for the video examples.

packages/components/src/components/hds/advanced-table/th-resize-handle.ts
shleewhite: Hmm ok maybe I was looking at something wrong. I also cant replicate the issue now.

packages/components/src/components/hds/advanced-table/models/column.ts
zamoore: I think the work I just pushed is a solution for this. What a rabbit hole. Your suggestion was the right direction, but you have to do it in a cascade, otherwise you run into the same problem when the column tries to recover enough debt to pay its creditors. If it tries to pay more than it has, you hit the same bug. So you need to:

1) Attempt to collect the debt directly
2) If the debtor cant pay, it needs to recursively recover its own debts and so on and so on.

packages/components/src/components/hds/advanced-table/models/column.ts
```diff
@@ -117,37 +140,182 @@ export default class HdsAdvancedTableColumn {
...
+      const paymentFromSurplus = this._collectFromSurplus(
+        debtor,
+        debtToCollect
+      );
+
+      // Reduce the amount we still need to collect
+      debtToCollect = debtToCollect - paymentFromSurplus;
+
+      // If a shortfall remains, trigger the cascading collection
+      if (debtToCollect > 0) {
```
dchyun: [Issue] In certain situations the column widths can go below their minimums or the total width of the table can exceed its original value.

The new sub-debtor behavior captures when all the debt needed can be collected from the first level sub-debtors, but doesn't account for when even those debts aren't enough. 

When this happens there should be some recursive-like step to go track down a further level of sub-debtors, but right now the logic I think assumes the debt has been collected. 

**Example scenario**

https://github.com/user-attachments/assets/aa6dfe8a-f12c-43a9-b9da-cb282a0090f3

Screenshot 2025-08-19 at 4 34 55 PM

**Proposed solution**
I tested out some crude ways of fixing this. In this spot changing the logic to `while (debtToCollect > 0)` ensures that the full debt is always collected. I added more logic to chase down further levels of debts, and _I think_ it somewhat works. 
```
let debtorToCollectFrom = debtor;
while (debtToCollect > 0) {
  const paymentFromCascade = this._collectFromSubDebtors(
    debtorToCollectFrom,
    debtToCollect
  );

  debtToCollect = debtToCollect - paymentFromCascade;

  let subDebtors = this.table.columns.filter(
    (column) => column.widthDebts[debtorToCollectFrom.key]
  );

  debtorToCollectFrom = subDebtors[0]!;
}
```

https://github.com/user-attachments/assets/0871a579-7dc7-4308-9038-53d5d3612b73

zamoore: I think I came up with a way of handling it recursively that should solve things.

packages/components/src/components/hds/advanced-table/th-resize-handle.ts
```diff
@@ -235,29 +275,56 @@ export default class HdsAdvancedTableThResizeHandle extends Component<HdsAdvance
...
+    amount: number
   ): void {
-    if (nextColumn === undefined) {
+    borrower.widthDebts = {
+      ...borrower.widthDebts,
+      [lenderKey]: (borrower.widthDebts[lenderKey] ?? 0) + amount,
+    };
+  }
+
+  private _setWidthDebts(): void {
```
dchyun, line 317: [Issue] When using the resize handle, the debt collection across columns only adds debt, and does not subtract it when a column is resized in a way that reduces debts. 

Example:
1. On first column, press right arrow to increase width 10px
3. Press left arrow to decrease width 10px
Screenshot 2025-08-20 at 10 31 47 AM

The expectation would be that the debts would be back to 0, but what's happening is both the artist and album columns have a debt of 10, because those debts are never evened out properly. 

The "reset column width" is still working properly despite this issue. One issue is since the debts are never cleared properly, every resize action adds to the list of debts to collect. After many resize operations using the "reset column width" action can cause a stack overflow due to the recursive debt collection function. 
Screenshot 2025-08-20 at 10 09 01 AM

dchyun, line 317: [Issue] When using the resize handle, the width of adjacent columns can be changed as well. This was only noticeable on screen sizes larger than 1440px, or at 90% on a 1440px desktop. 

https://github.com/user-attachments/assets/8089f114-75be-4070-8f43-80f530bc2032

  

This may be related to the other debt collection issue with the resize handler mentioned above, but not sure.

dchyun (commented):
Column resizing is working perfectly when using the mouse. Those debt issues are fixed, but just catching a couple others when using the resize handle with the keyboard.

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -446,23 +433,31 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
...
-      });
+      column.thElement = element;
     }
   );
 
   private _setUpScrollWrapper = modifier((element: HTMLDivElement) => {
     this._scrollWrapperElement = element;
 
+    const updateHorizontalScrollIndicators = () => {
+      if (element.clientWidth < element.scrollWidth) {
```
dchyun:
```suggestion
      this.showScrollIndicatorRight = (element.clientWidth < element.scrollWidth) ? true : false;
```
[Nit] Could simplify this statement

showcase/tests/unit/components/hds/advanced-table/models/column-test.js
```diff
@@ -187,39 +197,58 @@ module('Unit | Component | hds/advanced-table/models/column', function () {
...
+    column.setPxTransientWidth(200);
     assert.strictEqual(
-      column.width,
+      column.transientWidth,
       '200px',
       'sets exact width when within constraints',
     );
   });
 
   test('restoreWidth sets width back to original value', function (assert) {
```
dchyun, line 222: [Suggestion] Might be useful to add in a test for setting a column with a non-px initial value back to its original value, or a test setting a column using the `DEFAULT_WIDTH` back to its original value.

packages/components/src/components/hds/advanced-table/models/table.ts
```diff
@@ -152,6 +153,26 @@ export default class HdsAdvancedTableTableModel {
     }
   }
 
+  setTransientColumnWidths(options: { roundValues?: boolean } = {}): void {
```
dchyun, line 156: [Question] Could we always use rounded values? When this function is called in `startResize` with un-rounded values I think the values eventually passed into the resizing are rounded.

packages/components/src/components/hds/advanced-table/models/column.ts
```diff
@@ -5,13 +5,15 @@
...
 import { action } from '@ember/object';
+import { guidFor } from '@ember/object/internals';
 
 import type HdsAdvancedTableModel from './table.ts';
 import type {
   HdsAdvancedTableHorizontalAlignment,
   HdsAdvancedTableColumn as HdsAdvancedTableColumnType,
 } from '../types';
 
+export const DEFAULT_WIDTH = '1fr'; // default to '1fr' to allow flexible width
```
dchyun, line 16: [Issue] When the table has a horizontal overflow, and columns are set to non-px values, it is possible for the initial width of the column to go below the `DEFAULT_MIN_WIDTH` of 150px.

When I was adjusting some of the columns to cause overflow it is possible for a column to go below the default min width of 150px if the calculated width of a width like `1fr` / `auto` / `min-content` is less than 150px.

Screenshot 2025-08-22 at 2 49 25 PM

This can lead to a jumping effect when resizing the column, because it jumps immediately up to 150px when the resize is triggered. 

https://github.com/user-attachments/assets/c933775c-6ba7-49d5-ad0a-82e07f73a6c2

This is probably not likely to occur often. Also, it is likely an existing bug if a user sets the width to a non-px value. So maybe it doesn't need to be a blocker. The fix could be overriding the initial width provided if that width is less than the min and setting it to the min px value instead.

packages/components/src/components/hds/advanced-table/models/table.ts
zamoore: It leads to a small jump away from the cursor when doing the mouse resize

packages/components/src/components/hds/advanced-table/index.ts
zamoore: You're right. I simplified it a bit further to

```
const updateHorizontalScrollIndicators = () => {
      this.showScrollIndicatorRight = element.clientWidth < element.scrollWidth;
    };
    ```

File: packages/components/src/components/hds/advanced-table/th-resize-handle.ts [hunk 13]
```diff
@@ -131,35 +150,42 @@ export default class HdsAdvancedTableThResizeHandle extends Component<HdsAdvance
...
+      this._setWidthDebts();
+      this._applyTransientWidths();
+      column.table.resetTransientColumnWidths();
+      this._transientDelta = 0;
+
+      this.onColumnResize(column.key, column.width);
+    });
   }
 
   @action
```
dchyun, line 191: [Issue] Not related to any changes made here, but when right-clicking on the resize handle, the resizing is never stopped, and columns are resized after the click is done. Caught this while trying to inspect some things.


https://github.com/user-attachments/assets/9ac5cece-3347-4826-a39b-d8eb856285de

This can be fixed by only triggering the resizing inside the `startResize` action when left clicks occur.
```
// Only respond to left-click (primary) mouse button
if (event.button !== 0) {
  return;
}
```

zamoore: Oh good catch on this. Going to add this in for this PR.

KristinLBradley:
Currently, if I narrow the browser window until a scrollbar appears for the ["Resizeable columns" table examples](https://hds-showcase-git-zamoore-5171-advancedtable-pa-700460-hashicorp.vercel.app/components/advanced-table#resizable-columns) (i.e. until there's overflow) then resize the columns to make them narrower until all content fits, the table width will still be over 100% and have a scrollbar even though the content now fits. If I resize the window to make it wider, then the table width is actually no longer 100% however so the scrollbar goes away once everything fits within the window.

Would we want to make any changes to any of this behavior? For example, resize the total table width to 100% on window resize or after the columns have been resized and the content now fits? (This would add more calculations & complexity to the behavior of course though.)

packages/components/src/components/hds/advanced-table/models/table.ts
```diff
@@ -48,6 +48,7 @@ export default class HdsAdvancedTableTableModel {
   @tracked sortBy: HdsAdvancedTableTableArgs['sortBy'] = undefined;
   @tracked sortOrder: HdsAdvancedTableTableArgs['sortOrder'] =
     HdsAdvancedTableThSortOrderValues.Asc;
+  @tracked gridElement?: HTMLDivElement = undefined;
```
dchyun, line 51: [Question] Is this element needed? I don't see it being used anywhere.

zamoore: I'm going to keep it in there for now. It's useful for reoerdering anyways,

zamoore:
> Currently, if I narrow the browser window until a scrollbar appears for the ["Resizeable columns" table examples](https://hds-showcase-git-zamoore-5171-advancedtable-pa-700460-hashicorp.vercel.app/components/advanced-table#resizable-columns) (i.e. until there's overflow) then resize the columns to make them narrower until all content fits, the table width will still be over 100% and have a scrollbar even though the content now fits. If I resize the window to make it wider, then the table width is actually no longer 100% however so the scrollbar goes away once everything fits within the window.
> 
> Would we want to make any changes to any of this behavior? For example, resize the total table width to 100% on window resize or after the columns have been resized and the content now fits? (This would add more calculations & complexity to the behavior of course though.)

I did notice this as well. I think this is going to warrant some more discussion outside of this work. I'll bring this up when we have our AT sync this week. Thanks for the review.

packages/components/src/components/hds/advanced-table/index.hbs
dchyun, line 1: [Nit][Non-blocker]
Because of the new width rules, the pinnable first column examples in the showcase don't overflow as much as they used to, or don't really overflow at all.

Could add explicit widths to the columns just to showcase the pinning a little better.

```
@columns={{array
  (hash key="id" label="ID" width="150px" isSortable=sortableBool)
  (hash key="name" label="Name" width="150px")
  (hash key="email" label="Email" width="150px")
  (hash key="role" label="Role" width="150px")
}}
```

dchyun (approved):
LGTM! 🥳 One nit for the showcase caught going through the percy diff

packages/components/src/components/hds/advanced-table/models/column.ts
```diff
@@ -117,37 +142,130 @@ export default class HdsAdvancedTableColumn {
...
     this.isSortable = column.isSortable ?? false;
     this.isVisuallyHidden = column.isVisuallyHidden ?? false;
-    this.key = column.key;
+    this.key = column.key ?? guidFor(this);
     this.tooltip = column.tooltip;
     this._setWidthValues(column);
     this.sortingFunction = column.sortingFunction;
   }
 
+  // main collection function
```
didoo, line 151: @dchyun @zamoore just a note: I _think_ I understand the metaphor of the payment, collection, debt, lending, etc but I am not _sure_ it's clear what it means if you're not the person that implemented the code. (plus, when I first looked at the code, I thought it was some spurious code injected by an AI agent)

One thing I would have done is to add a large comment block at the top, explaining what the different terms means (eg. "lender") and how they relate to a column that I may see in the browser and interacting with to debug/review the code.

My 2 cents

PR #3090, opened 2025-08-06
Reverting strings updated by `ember-intl` introduction

zamoore:
### :pushpin: Summary

If merged this, will revert a couple of text-casing updates that were made as part of the internationalization work.

When smoke testing the HDS 4.22.0 release, we came across a couple of instances where consumer application tests were looking for case-sensitive strings that had changed as part of the translation work.

Specifically,

- In the `ApplicationState` component, the text "ERROR" was converted to "Error" and then displayed in all-caps via CSS
- In the `Pagination::Nav::Number` component, the screen-reader-only text "page [number]" was changed to "Page [number]"

While these changes make ultimately make sense, we're going to hold off on introducing any additional adoption friction until the next major release.

### :link: External links

Discussion: https://hashicorp.slack.com/archives/C025N5V4PFZ/p1754424235297369

***

<details>

</details>


packages/components/translations/hds/components/application-state/header/en-us.yaml
```diff
@@ -0,0 +1 @@
+ERROR: ERROR
```
dchyun:
```suggestion
error: ERROR
```
[Nit] I think this can be lowercase since you're calling it with `hds.application-state.header.error`

dchyun (commented):
Looks like there's a linting error, and I have one small comment, but looks all good besides that.

zamoore:
> Looks like there's a linting error, and I have one small comment, but looks all good besides that.

🤦 You're totally right. Thanks!

packages/components/translations/hds/components/application-state/header/en-us.yaml
```diff
@@ -0,0 +1 @@
+error: ERROR
```
dchyun:
```suggestion
error: ERROR

```
[todo] Looks like the linter is complaining about this file not ending in a new line

PR #3093, opened 2025-08-06
`AdvancedTable`: fix styles for right/center aligned cells

shleewhite:
### :pushpin: Summary

If merged, this PR would fix the styles when the user sets `@align` to `"right"` or `"center"`. 

**[PREVIEW](https://hds-showcase-git-hds-5254-advanced-table-align-hashicorp.vercel.app/components/advanced-table#thsort)**

**CURRENT PRODUCTION SHOWCASE**
Screenshot 2025-08-06 at 6 00 58 PM
Screenshot 2025-08-06 at 6 01 04 PM
Screenshot 2025-08-06 at 6 01 59 PM

**AFTER PROPOSED FIX**
Screenshot 2025-08-06 at 6 02 29 PM
Screenshot 2025-08-07 at 8 00 10 AM
Screenshot 2025-08-06 at 6 02 51 PM

### :link: External links

Jira ticket: [HDS-5254](https://hashicorp.atlassian.net/browse/HDS-5254)

***

<details>

</details>

[HDS-5254]: https://hashicorp.atlassian.net/browse/HDS-5254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/advanced-table/th.hbs
+++ b/packages/components/src/components/hds/advanced-table/th.hbs
@@ -23,7 +23,7 @@
   }}
   ...attributes
 >
-  <Hds::Layout::Flex @justify="space-between" @align="center" @gap="8" {{style flex-grow="1"}}>
+  <Hds::Layout::Flex @justify="space-between" @align="center" @gap="8">
     {{#if @column.isVisuallyHidden}}
       <span class="sr-only">{{yield}}</span>
     {{else}}
--- a/packages/components/src/components/hds/advanced-table/th.hbs
+++ b/packages/components/src/components/hds/advanced-table/th.hbs
@@ -28,16 +28,16 @@
       <span class="sr-only">{{yield}}</span>
     {{else}}
       {{#if @tooltip}}
+        {{#if @isExpandable}}
+          <Hds::AdvancedTable::ThButtonExpand
+            @labelId={{this._labelId}}
+            @onToggle={{@onClickToggle}}
+            @isExpanded={{@isExpanded}}
+            @isExpandAll={{@hasExpandAllButton}}
+            {{this._manageExpandButton}}
+          />
+        {{/if}}
         <div class="hds-advanced-table__th-content">
-          {{#if @isExpandable}}
-            <Hds::AdvancedTable::ThButtonExpand
-              @labelId={{this._labelId}}
-              @onToggle={{@onClickToggle}}
-              @isExpanded={{@isExpanded}}
-              @isExpandAll={{@hasExpandAllButton}}
-              {{this._manageExpandButton}}
-            />
-          {{/if}}
           <span
             id={{this._labelId}}
             class="hds-advanced-table__th-content-text hds-typography-body-200 hds-font-weight-semibold"
@@ -47,16 +47,16 @@
           <Hds::AdvancedTable::ThButtonTooltip @tooltip={{@tooltip}} @labelId={{this._labelId}} />
         </div>
       {{else}}
+        {{#if @isExpandable}}
+          <Hds::AdvancedTable::ThButtonExpand
+            @labelId={{this._labelId}}
+            @onToggle={{@onClickToggle}}
+            @isExpanded={{@isExpanded}}
+            @isExpandAll={{@hasExpandAllButton}}
+            {{this._manageExpandButton}}
+          />
+        {{/if}}
         <div class="hds-advanced-table__th-content">
-          {{#if @isExpandable}}
-            <Hds::AdvancedTable::ThButtonExpand
-              @labelId={{this._labelId}}
-              @onToggle={{@onClickToggle}}
-              @isExpanded={{@isExpanded}}
-              @isExpandAll={{@hasExpandAllButton}}
-              {{this._manageExpandButton}}
-            />
-          {{/if}}
           <span
             class="hds-advanced-table__th-content-text hds-typography-body-200 hds-font-weight-semibold"
             id={{this._labelId}}
--- a/packages/components/src/components/hds/advanced-table/th.hbs
+++ b/packages/components/src/components/hds/advanced-table/th.hbs
@@ -23,7 +23,7 @@
   }}
   ...attributes
 >
-  <Hds::Layout::Flex @justify="space-between" @align="center" @gap="8">
+  <Hds::Layout::Flex @justify="space-between" @align="center" @gap="8" {{style flex-grow="1"}}>
     {{#if @column.isVisuallyHidden}}
       <span class="sr-only">{{yield}}</span>
     {{else}}
--- a/packages/components/src/styles/components/advanced-table.scss
+++ b/packages/components/src/styles/components/advanced-table.scss
@@ -165,6 +165,7 @@ $hds-advanced-table-button-size: 24px; // the size of the buttons and dropdown t
       text-align: center;
 
       .hds-advanced-table__th-content {
+        flex-grow: 1;
         justify-content: center;
       }
     }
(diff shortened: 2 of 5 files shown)


packages/components/src/styles/components/advanced-table.scss
```diff
@@ -261,6 +261,7 @@ $hds-advanced-table-button-size: 24px; // the size of the buttons and dropdown t
 // multi-select (isSelectable=true)
 .hds-advanced-table__th-content {
   display: flex;
+  flex-grow: 1;
```
didoo, line 265: For reference: there is something not working when the `Th` is inside the `tbody`, spoke with @shleewhite and she will look into it

screenshot_5267

showcase/app/templates/page-components/advanced-table.hbs
dchyun, line 1: [todo]

On line 414, `(hash key="year" label="Release Year")` needs to be updated to `(hash key="year" label="Release Year" align="right")`. The right align was properly set on the data, but not on the header for this example.
Screenshot 2025-08-08 at 1 16 21 PM

packages/components/src/styles/components/advanced-table.scss
dchyun, line 265: [Issue] When the context menu and tooltip are present, this style is causing an issue where the tooltip is always at the end of the flex box.

**Before**
Screenshot 2025-08-08 at 1 21 52 PM

**After**
Screenshot 2025-08-08 at 1 21 27 PM

This is due to the column text `span` having `width: 100%;`. This issue is not present in the `ThSort` because the text there does not have its width set. Maybe we could remove this `width: 100%;`.

packages/components/src/components/hds/advanced-table/th.hbs
```diff
@@ -23,21 +23,21 @@
   }}
   ...attributes
 >
-  <Hds::Layout::Flex @justify="space-between" @align="center" @gap="8">
+  <Hds::Layout::Flex @justify="space-between" @align="center" @gap="8" {{style flex-grow="1"}}>
```
dchyun:
```suggestion
  <Hds::Layout::Flex @justify="space-between" @align="center" @gap="8">
```
[Question] Looking at the inspect tool this style may not be getting applied. Could it be removed? If it is still needed it may make sense to add it into the `ThSort` as well to align the two. 
Screenshot 2025-08-08 at 1 34 23 PM

showcase/app/templates/page-components/advanced-table.hbs
```diff
@@ -1680,15 +1727,76 @@
...
+        class="hds-advanced-table hds-advanced-table--density-medium"
+        role="grid"
+        {{style gridTemplateColumns="repeat(3, 1fr)"}}
+      >
+        <div class="hds-advanced-table__thead" role="rowgroup">
+          <div class="hds-advanced-table__tr" role="row">
+            <Hds::AdvancedTable::Th
+              @isExpandable={{true}}
+              @tooltip="Here is more information"
+              @column={{get this.sampleTableModel.columns 1}}
```
shleewhite, line 1747: Note: for both ThSort and Th, there is a scrollbar for this example. This is because of the visually hidden resize handler. I think it is ok for now.

packages/components/src/components/hds/advanced-table/th.hbs
shleewhite: Ah I figured out what was going on. The cells in the header are display: block, but cells in the body had display: flex. This was making the content not grow correctly in the body examples so I fixed it at the time with flex-grow.

I ended up refactoring so now the inline style isnt needed anymore.

packages/components/src/styles/components/advanced-table.scss
dchyun: This issue with the tooltip being right aligned when there are resizable columns is still present. It's currently not seen in the Th "alignment with all interactive elements" examples because in those examples the `hds-advanced-table__thead--has-resizable-columns` class has not been added to the mock thead. 

When that class is added it's bringing a selector that sets `width: 100%;` on the label span which is what is causing the issue.

shleewhite: Ahhh 🤦🏽‍♀️ I see now. Thank you for clarifying.

dchyun (approved):
Looking good! All percy diff looks expected.

showcase/app/templates/page-components/advanced-table.hbs
```diff
@@ -1534,20 +1534,47 @@
...
-            <Hds::AdvancedTable::ThSort @tooltip="Here is more information">Left</Hds::AdvancedTable::ThSort>
             <Hds::AdvancedTable::ThSort @align="center">Center</Hds::AdvancedTable::ThSort>
+            <Hds::AdvancedTable::ThSort @align="right">Right</Hds::AdvancedTable::ThSort>
+          </div>
+        </div>
+      </div>
+    </SF.Item>
+  </Shw::Flex>
+
+  <Shw::Flex @label="Alignment with all interactive elements" as |SF|>
```
KristinLBradley, line 1553: Nice! Separating the examples out like this makes it much clearer.

showcase/app/templates/page-components/advanced-table.hbs
KristinLBradley: Just wondering... Does it make sense to have a resize handle on the last cell in a row?

KristinLBradley: When I look at the "Resizable columns with sorting" example, the last column is not resizable (does not have a resize handle on the right-most side of the table) which is what I would expect. Could you possibly update your example to get rid of the resize handle on the last row?

showcase/app/templates/page-components/advanced-table.hbs
```diff
@@ -1680,15 +1707,76 @@
...
+              @tooltip="Here is more information"
+              @column={{get this.sampleTableModel.columns 1}}
+              @hasResizableColumns={{true}}
+            >Center</Hds::AdvancedTable::Th>
+            <Hds::AdvancedTable::Th
+              @align="right"
+              @isExpandable={{true}}
+              @tooltip="Here is more information"
+              @column={{get this.sampleTableModel.columns 1}}
+              @hasResizableColumns={{true}}
```
KristinLBradley, line 1762: I think you could just delete `@hasResizableColumns={{true}}` here since it doesn't seem like the way you would normally use this. It looks like normally you would add this argument to the `Hds::AdvancedTable` parent in normal usage which would not make the last column resizable.

showcase/app/templates/page-components/advanced-table.hbs
shleewhite: Ahh ok I see. I need to include the resize handle for the context menu to show up, so I will just add another column that doesnt have one.

showcase/app/templates/page-components/advanced-table.hbs
dchyun, line 1: [todo][non-blocker] I can make this update after this merges, but one small unrealted fix is in the `Td` examples for the inline dropdown the route needs to be updated to `page-components.advanced-table`. Didn't realize I'd been routed to the table page after clicking it and was quite confused for a min 😅.
```
<D.Interactive @route="page-components.table">Dropdown</D.Interactive>
```

showcase/app/templates/page-components/advanced-table.hbs
dchyun: As Lee's comment in the other thread said this is needed to test the alignment with the context menu. This could be updated once the context menu is used for other features, but for now I think adding the non-resizable last column works.

dchyun (approved):

showcase/app/templates/page-components/advanced-table.hbs
shleewhite: hahhaha way ahead of you, I already fixed it in the spike branch.

PR #3094, opened 2025-08-06
Showcase: fix examples of visually hidden header cells in AdvancedTable page

shleewhite:
### :pushpin: Summary

If merged, this PR would fix the examples of visually hidden header cells in the Advanced Header showcase page. This is because the TH component no longer uses `@isVisuallyHidden` to hide cell content - it uses the `columns` model.

In a future PR, we should deprecate the `@isVisuallyHidden` argument because it doesn't do anything. This probably not have any affect on consumers, but we could do it in `5.0.0` to be safe.

**CURRENT SHOWCASE**
Screenshot 2025-08-06 at 6 30 14 PM

**AFTER FIX**
Screenshot 2025-08-06 at 6 30 33 PM

***

<details>

</details>

### DIFF
--- a/showcase/app/controllers/page-components/advanced-table.ts
+++ b/showcase/app/controllers/page-components/advanced-table.ts
@@ -8,6 +8,7 @@ import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { deepTracked } from 'ember-deep-tracked';
 import { later } from '@ember/runloop';
+import type Owner from '@ember/owner';
 
 import type { PageComponentsAdvancedTableModel } from 'showcase/routes/page-components/advanced-table';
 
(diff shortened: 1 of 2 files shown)


showcase/app/controllers/page-components/advanced-table.ts
```diff
@@ -85,15 +86,30 @@ export default class PageComponentsAdvancedTableController extends Controller {
   // @ts-expect-error - we need to reevaluate how we get the data for the table demos when we break up the template files into sub components
   @deepTracked multiSelectUserData__demo4 = [...this.model.userDataDemo4];
   @tracked focusableElementsVisible = false;
-  @deepTracked thResizeHandleTable = new HdsAdvancedTableModel({
-    // @ts-expect-error - this will be fixed in HDS-5090
-    model: this.model.userDataShort,
```
shleewhite, line 90: Upon further testing, this way of creating an AdvancedTableModel outside of the component wasn't actually working as expected. When I actually logged what the first column, it was undefined.

 I refactored to match how it is used in the actual component (ie. in a constructor) and didn't use the userDataShort anymore because it is not defined yet in the constructor. Now it works as expected.

showcase/app/templates/page-components/advanced-table.hbs
```diff
@@ -2258,7 +2258,7 @@
                   </div>
                   <Hds::AdvancedTable::ThResizeHandle
```
dchyun, line 2259: [Note] This is unrelated, but I noticed the active state mock selector was not working for the resize handle. In the adv table styles `.mock-active::after` just needs to be added to the two selectors that include `.hds-advanced-table__th-resize-handle--resizing::after`. I can make this fix in a separate PR or if you want you can throw it in with this.

dchyun (approved):
nice catch, all percy diffs look expected to me.

showcase/app/templates/page-components/advanced-table.hbs
shleewhite: ahh shoot. could you do that in a separate pr? I think that would be better.

dchyun: Created a PR #3102

zamoore (approved):

PR #3099, opened 2025-08-07
`CodeEditor` - React to `@value` changes

zamoore:
### :pushpin: Summary

Before, if the value for `value` was updated externally, the component/modifier would not update to the newly provided value. This fixes that.

- Update Codemirror editor value when the `value` argument on the `hds-code-editor` modifier changes.
- Use `ember-concurrency` to make sure that tests wait for the updates correctly

### :link: External links

Jira ticket: [HDS-5451](https://hashicorp.atlassian.net/browse/HDS-5451)

***

<details>

</details>

[HDS-5451]: https://hashicorp.atlassian.net/browse/HDS-5451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/code-editor/index.ts
+++ b/packages/components/src/components/hds/code-editor/index.ts
@@ -14,9 +14,9 @@ import { StateEffect } from '@codemirror/state';
 import HdsCodeEditorDescription from './description.ts';
 import HdsCodeEditorTitle from './title.ts';
 
+import type Owner from '@ember/owner';
 import type { ViewUpdate } from '@codemirror/view';
 import type { WithBoundArgs } from '@glint/template';
-import type Owner from '@ember/owner';
 import type { ComponentLike } from '@glint/template';
 import type { HdsCodeEditorSignature as HdsCodeEditorModifierSignature } from '../../../modifiers/hds-code-editor.ts';
 import type { HdsCodeEditorDescriptionSignature } from './description';
@@ -154,14 +154,17 @@ export default class HdsCodeEditor extends Component<HdsCodeEditorSignature> {
   @action
   onSetup(editorView: EditorView): void {
     this._isSetupComplete = true;
-    const valUpdateExtension = EditorView.updateListener.of(
+
+    const valueUpdateExtension = EditorView.updateListener.of(
       (update: ViewUpdate) => {
         this._value = update.state.doc.toString();
       }
     );
+
     editorView.dispatch({
-      effects: StateEffect.appendConfig.of([valUpdateExtension]),
+      effects: StateEffect.appendConfig.of([valueUpdateExtension]),
     });
+
     this.args.onSetup?.(editorView);
   }
 }
--- a/packages/components/src/components/hds/code-editor/index.ts
+++ b/packages/components/src/components/hds/code-editor/index.ts
@@ -8,10 +8,13 @@ import { tracked } from '@glimmer/tracking';
 import { action } from '@ember/object';
 import { modifier } from 'ember-modifier';
 import { guidFor } from '@ember/object/internals';
+import { EditorView } from '@codemirror/view';
+import { StateEffect } from '@codemirror/state';
 
 import HdsCodeEditorDescription from './description.ts';
 import HdsCodeEditorTitle from './title.ts';
 
+import type { ViewUpdate } from '@codemirror/view';
 import type { WithBoundArgs } from '@glint/template';
 import type Owner from '@ember/owner';
 import type { ComponentLike } from '@glint/template';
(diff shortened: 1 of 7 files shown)


packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -195,15 +202,12 @@ export default class HdsCodeEditorModifier extends Modifier<HdsCodeEditorSignatu
...
-    // if the editor already exists, update the line wrapping
+    // if the editor already exists, update its state
     if (this.editor) {
-      this.editor.dispatch({
-        effects: this.lineWrappingCompartment.reconfigure(
-          hasLineWrapping ? EditorView.lineWrapping : []
-        ),
-      });
+      // eslint-disable-next-line ember/no-runloop
+      next(this, this._updateEditorState, { value, hasLineWrapping });
```
meirish: I don't think the runloop should be necessary here - you should only set `value` once. The issue you were seeing is from using the same function with the external button and the `onInput` function below (I'll comment down there to make it clearer).

showcase/app/templates/page-components/code-editor.hbs
```diff
@@ -172,4 +172,22 @@
       />
     </SF.Item>
   </Shw::Flex>
+
+  <Shw::Flex @direction="column" as |SF|>
+    <SF.Item @label="Update value externally">
+      <Hds::Button
+        @text="Change value"
+        class="hds-code-editor-controls"
+        {{on "click" (fn this.setExternalUpdateDemo "New value")}}
```
meirish, line 181: So this click handler and the `onInput` function passed to the modifer below are the same (they both point to `this.setExternalUpdateDemo`). This causes an issue though when you click the button that I think goes like this:

1. button click sets `externalUpdateDemo` value
2. that passes to the modifier as the `value` named argument
3. this triggers a call into CodeMirror to set its value (which was added to in this PR 🎉 )
4. this triggers the `EditorView.updateListener` call
5. which calls the `onInput` function, also passed to the modifier as a named arg
6. which... sets `externalUpdateDemo` again

In this case if you wanted to use the same function for `onInput` and external setting of a value passed to `value`, then that function should be debounced, not doing the debounce in the modifier itself.

packages/components/src/components/hds/code-editor/index.ts
```diff
@@ -153,6 +154,14 @@ export default class HdsCodeEditor extends Component<HdsCodeEditorSignature> {
...
   onSetup(editorView: EditorView): void {
     this._isSetupComplete = true;
+    const valUpdateExtension = EditorView.updateListener.of(
+      (update: ViewUpdate) => {
+        this._value = update.state.doc.toString();
+      }
+    );
+    editorView.dispatch({
+      effects: StateEffect.appendConfig.of([valUpdateExtension]),
+    });
```
meirish, line 166: This is the part I don't really love...it feels bad to pull CodeMirror specific code into the component when most of it is encapsulated in the modifier. But this makes it so that the internal `_value` is always up-to-date with what CodeMirror is rendering. The only place this is used in the component is for the copy button functionality. I tried pointing the copy button at the DOM element representing the editor, but because of the extra divs CodeMIrror uses, the value always included extra new lines (`textContent` matched, but `CopyButton` uses `innerText`). 

Extra updates here aren't an issue because it's only changing the internal value, not also calling the `onInput` like the modifier does.

showcase/tests/integration/components/hds/code-editor/index-test.js
```diff
@@ -224,15 +234,16 @@ module('Integration | Component | hds/code-editor/index', function (hooks) {
...
-      changes: {
-        from: this.editorView.state.selection.main.from,
-        insert: 'Additional text. ',
-      },
-    });
+    const newValue = `Additional text.
+
+
+
+      Test Code`;
```
meirish, line 240: Added this to verify that new line handling worked properly with copying.

showcase/tests/integration/components/hds/code-editor/index-test.js
```diff
@@ -318,12 +329,9 @@ module('Integration | Component | hds/code-editor/index', function (hooks) {
...
 
-    this.editorView.dispatch({
-      changes: {
-        from: this.editorView.state.selection.main.from,
-        insert: 'Test string',
-      },
-    });
+    // simulate user input
+    await fillIn('.cm-content', 'Test string');
+    await triggerEvent('.cm-content', 'input');
```
meirish, line 333: `onInput` will now require user interaction - is this a breaking change? I don't think it really is from the consumer standpoint, but it could mean that teams would have to update their tests.

packages/components/src/components/hds/code-editor/index.ts
zamoore: It feels a little bad, but I'm not too too worried about it. The component can't exist independently of CodeMirror even though its not directly coupled.

showcase/tests/integration/components/hds/code-editor/index-test.js
zamoore: It will almost certainly break tests, but it's only currently used in Vault.

didoo:
@zamoore do you think this PR could be closed?

alex-ju:
> @zamoore do you think this PR could be closed?

I asked [the same question](https://hashicorp.atlassian.net/browse/HDS-5451) a while back. I'm going to go ahead in close it for now. Zack can pull the branch and continue the work anytime.

PR #3104, opened 2025-08-11
AdvancedTable - First column pinning

dchyun:
### :pushpin: Summary

If merged, this PR would support allowing the user to pin and unpin the first column of the `AdvancedTable`. It also has a few other alterations and fixes for the Advanced table for translating text in the context menu, and fixing a CSS issues in the scroll indicator. 

[Preview page](https://hds-showcase-git-dchyun-adv-table-pin-col-hashicorp.vercel.app/components/advanced-table#pinnable-first-column)

#### Column pinning

If the existing `@hasStickyFirstColumn` argument is `true` currently the first column is sticky/pinned, but the user can does not have the ability to change that, or to pin a column that started as unpinned. 

With the changes in this PR, `@hasStickyFirstColumn` would have the following interactions:

- if `true`, the first column is sticky and the user can toggle its pinned state with the context menu
- if `false`, the first column is not sticky and the user can toggle its pinned state with the context menu
- if `undefined`, the first column is not sticky and there is no menu option to pin it

Additionally as part of the filtering implementation all text in the `ThContextMenu` has been converted to use translations.

#### Design considerations

With the changes proposed here, a user would be able to change a first column set as sticky to not sticky with the context menu. It would not be possible to have a sticky column that the user can not change the state of. 

#### Scroll indicator CSS issue

While implementing the column pinning feature, it was noticed that when the column was sticky there was a slight gap present between the left scroll indicator, and the border of the sticky cells. 

Screenshot 2025-08-14 at 1 25 45 PM

This is an existing issue caused when the width of the first column is set to a fixed px width. When the first column width is set to `auto` this issue is not seen. 

The cause of this issue is that when the first column is set to `auto` its width grows slightly when sticky as the border widths of it are increased. However, the offset calculated for the scroll indicator uses the smaller pre-scrolled width of the cell in its calculations. This allows the offset of the scroll indicator to be slightly less than the cell width and be hidden appropriately. 

**Not sticky**
Screenshot 2025-08-18 at 11 41 29 AM
**Sticky**
Screenshot 2025-08-18 at 11 41 37 AM

When the width of a column is set to a fixed px value, the width of the cell _does not_ increase when sticky. When this occurs the scroll indicator is not positioned correctly since it is not hidden slightly under the cell. 

To fix this issue, an update has been made to reduce the left offset of the scroll indicator by 1px when the first column has a px width. 

Screenshot 2025-08-11 at 9 49 21 AM

### :link: External links

Jira ticket: [HDS-5237](https://hashicorp.atlassian.net/browse/HDS-5237)

***

<details>

</details>

[HDS-5237]: https://hashicorp.atlassian.net/browse/HDS-5237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/advanced-table/index.ts
+++ b/packages/components/src/components/hds/advanced-table/index.ts
@@ -60,7 +60,8 @@ const getScrollIndicatorDimensions = (
   theadElement: HTMLDivElement,
   hasStickyHeader: boolean,
   hasStickyFirstColumn: boolean,
-  hasFirstColumnPxWidth: boolean
+  hasFirstColumnPxWidth: boolean,
+  isStickyColumnPinned: boolean
 ) => {
   const horizontalScrollBarHeight =
     scrollWrapper.offsetHeight - scrollWrapper.clientHeight;
@@ -85,6 +86,11 @@ const getScrollIndicatorDimensions = (
     if (stickyColumnHeaders.length > 1 || hasFirstColumnPxWidth) {
       leftOffset -= 1;
     }
+
+    // offsets the left: -1px position if the sticky column is already pinned when the scroll indicator is calculated
+    if (isStickyColumnPinned) {
+      leftOffset -= 1;
+    }
   }
 
   return {
@@ -99,7 +105,8 @@ const getScrollIndicatorDimensions = (
 
 const getStickyColumnLeftOffset = (
   theadElement: HTMLDivElement,
-  hasRowSelection: boolean
+  hasRowSelection: boolean,
+  isStickyColumnPinned: boolean
 ) => {
   // if there is no select checkbox column, the sticky column is all the way to the left
   if (!hasRowSelection) return '0px';
@@ -108,7 +115,14 @@ const getStickyColumnLeftOffset = (
     '.hds-advanced-table__th--is-selectable'
   ) as HTMLElement;
 
-  return `${selectableCell?.offsetWidth}px`;
+  let leftOffset = selectableCell?.offsetWidth ?? 0;
+
+  // if the sticky column is pinned when the offset is calculated, we need to account for the increased width of the border
+  if (isStickyColumnPinned && leftOffset > 0) {
+    leftOffset -= 2;
+  }
+
+  return `${leftOffset}px`;
 };
 
 export interface HdsAdvancedTableSignature {
@@ -190,6 +204,7 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
   private _scrollHandler!: (event: Event) => void;
   private _resizeObserver!: ResizeObserver;
   private _theadElement!: HTMLDivElement;
+  private _scrollWrapperElement!: HTMLDivElement;
 
   @tracked scrollIndicatorDimensions = DEFAULT_SCROLL_DIMENSIONS;
   @tracked isStickyColumnPinned = false;
@@ -446,55 +461,10 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
   );
 
   private _setUpScrollWrapper = modifier((element: HTMLDivElement) => {
+    this._scrollWrapperElement = element;
+
     this._scrollHandler = () => {
-      // 6px as a buffer so the shadow doesn't appear over the border radius on the edge of the table
-      const SCROLL_BUFFER = 6;
-
-      // left scroll indicator and sticky column styles
-      if (element.scrollLeft > SCROLL_BUFFER && !this.showScrollIndicatorLeft) {
-        if (this.hasStickyFirstColumn) {
-          this.isStickyColumnPinned = true;
-        }
-        this.showScrollIndicatorLeft = true;
-      } else if (element.scrollLeft === 0 && this.showScrollIndicatorLeft) {
-        this.isStickyColumnPinned = false;
-        this.showScrollIndicatorLeft = false;
-      }
-
-      // the right edge is how far the user can scroll, which is the full width of the table - the visible section of the table (also subtract the buffer)
-      const rightEdge =
-        element.scrollWidth - element.clientWidth - SCROLL_BUFFER;
-
-      // right scroll indicator
-      if (element.scrollLeft < rightEdge) {
-        this.showScrollIndicatorRight = true;
-      } else {
-        this.showScrollIndicatorRight = false;
-      }
-
-      // sticky header
-      if (element.scrollTop > 0) {
-        if (this.hasStickyHeader) {
-          this.isStickyHeaderPinned = true;
-        }
-        this.showScrollIndicatorTop = true;
-      } else {
-        if (this.hasStickyHeader) {
-          this.isStickyHeaderPinned = false;
-        }
-        this.showScrollIndicatorTop = false;
-      }
-
-      // the bottom edge is how far the user can scroll, which is the full height of the table - the visible section of the table (also subtract the buffer)
-      const bottomEdge =
-        element.scrollHeight - element.clientHeight - SCROLL_BUFFER;
-
-      // bottom scroll indicator
-      if (element.scrollTop < bottomEdge) {
-        this.showScrollIndicatorBottom = true;
-      } else {
-        this.showScrollIndicatorBottom = false;
-      }
+      this._updateScrollIndicators(element);
     };
 
     element.addEventListener('scroll', this._scrollHandler);
@@ -510,13 +480,15 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
         this._theadElement,
         this.hasStickyHeader,
         this.hasStickyFirstColumn ? true : false,
-        hasFirstColumnPxWidth
+        hasFirstColumnPxWidth,
+        this.isStickyColumnPinned
       );
 
       if (this.hasStickyFirstColumn) {
         this.stickyColumnOffset = getStickyColumnLeftOffset(
           this._theadElement,
-          isSelectable
+          isSelectable,
+          this.isStickyColumnPinned
         );
       }
     };
@@ -659,8 +631,62 @@ export default class HdsAdvancedTable extends Component<HdsAdvancedTableSignatur
     }
   }
 
+  private _updateScrollIndicators(element: HTMLElement): void {
+    // 6px as a buffer so the shadow doesn't appear over the border radius on the edge of the table
+    const SCROLL_BUFFER = 6;
+
+    // left scroll indicator and sticky column styles
+    if (element.scrollLeft > SCROLL_BUFFER) {
+      if (this.hasStickyFirstColumn) {
+        this.isStickyColumnPinned = true;
+      }
+      if (!this.showScrollIndicatorLeft) {
+        this.showScrollIndicatorLeft = true;
+      }
+    } else if (element.scrollLeft === 0 && this.showScrollIndicatorLeft) {
+      this.isStickyColumnPinned = false;
+      this.showScrollIndicatorLeft = false;
+    }
+
+    // the right edge is how far the user can scroll, which is the full width of the table - the visible section of the table (also subtract the buffer)
+    const rightEdge = element.scrollWidth - element.clientWidth - SCROLL_BUFFER;
+
+    // right scroll indicator
+    if (element.scrollLeft < rightEdge) {
+      this.showScrollIndicatorRight = true;
+    } else {
+      this.showScrollIndicatorRight = false;
+    }
+
+    // sticky header
+    if (element.scrollTop > 0) {
+      if (this.hasStickyHeader) {
+        this.isStickyHeaderPinned = true;
+      }
+      this.showScrollIndicatorTop = true;
+    } else {
+      if (this.hasStickyHeader) {
+        this.isStickyHeaderPinned = false;
+      }
+      this.showScrollIndicatorTop = false;
+    }
+
+    // the bottom edge is how far the user can scroll, which is the full height of the table - the visible section of the table (also subtract the buffer)
+    const bottomEdge =
+      element.scrollHeight - element.clientHeight - SCROLL_BUFFER;
+
+    // bottom scroll indicator
+    if (element.scrollTop < bottomEdge) {
+      this.showScrollIndicatorBottom = true;
+    } else {
+      this.showScrollIndicatorBottom = false;
+    }
+  }
+
   private _onPinFirstColumn = (): void => {
     this.hasPinnedFirstColumn = this.hasPinnedFirstColumn ? false : true;
+    // we need to retrigger the scroll indicator updates if the pinned state is changed when the table is already scrolled
+    this._updateScrollIndicators(this._scrollWrapperElement);
   };
 
   private _isStickyColumn = (


packages/components/src/components/hds/advanced-table/index.hbs
```diff
@@ -135,13 +141,13 @@
                 didInsert=this.didInsertRowCheckbox
                 willDestroy=this.willDestroyRowCheckbox
                 selectionAriaLabelSuffix=@selectionAriaLabelSuffix
-                hasStickyColumn=@hasStickyFirstColumn
+                hasStickyColumn=this.hasStickyFirstColumn
```
shleewhite, line 140: Nit: can you add this to the list of bound args in the type?

packages/components/src/components/hds/advanced-table/index.hbs
```diff
@@ -135,13 +141,13 @@
...
                 isStickyColumnPinned=this.isStickyColumnPinned
               )
               Th=(component
                 "hds/advanced-table/th"
                 scope="row"
-                isStickyColumn=@hasStickyFirstColumn
+                isStickyColumn=this.hasStickyFirstColumn
```
shleewhite, line 146: Nit: same as above, this needs to be added to the list of bound args.

packages/components/src/components/hds/advanced-table/th-context-menu.ts
```diff
@@ -65,6 +73,18 @@ export default class HdsAdvancedTableThContextMenu extends Component<HdsAdvanced
       ];
     }
 
+    if (hasPinnableFirstColumn && column.isFirst) {
+      options = [
+        ...options,
+        {
+          key: 'pin-first-column',
+          label: hasStickyFirstColumn ? 'Unpin column' : 'Pin column',
```
shleewhite: Question: Should this be a translated string? Like use the internationalization work.

showcase/app/templates/page-components/advanced-table.hbs
```diff
@@ -738,9 +738,52 @@
...
+                @isSelectable={{selectableBool}}
+                {{! @glint-expect-error - will be fixed by https://hashicorp.atlassian.net/browse/HDS-5090}}
+                @model={{this.model.userDataShort}}
+                @columns={{array
+                  (hash key="id" label="ID" isSortable=sortableBool)
+                  (hash key="name" label="Name")
+                  (hash key="email" label="Email")
+                  (hash key="role" label="Role")
+                }}
+                @hasPinnableFirstColumn={{true}}
```
shleewhite: Thought: I'm not super sure about this API... since we already have `@hasStickyFirstColumn`. Say if the first column is pinned by default, you'd need to set `@hasPinnableFirstColumn` and `@hasStickyFirstColumn`?

It is odd for us to mix the terminology of "pinning" and "sticky". Could we simplify to just using `@hasStickyFirstColumn` and if they set it to false it is "pinnable" and if it is undefined it isn't? Just throwing an idea out there.

shleewhite (commented):
Overall, the interaction looks great. I think it would be good to add an example where you can unpin the column and it starts as pinned, currently all the examples have it default unpinned.

Also, could you add this feature to [the full page demo](https://hds-showcase-git-dchyun-adv-table-pin-col-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-advanced-table)?

showcase/app/templates/page-components/advanced-table.hbs
dchyun: Yeah I didn't really like the mix of "pin" and "sticky" either. I like the idea of only using `hasStickyColumn` to control it, and I've updated the functionality to match that. 

The only caveat is now the user will always be able to toggle the column pin if `hasStickyColumn` is true. So it is not possible for a consumer to make a column sticky and prevent the user from unpinning it.  

Not sure if that's an important interaction but that was why I originally created a separate argument, `hasPinnableFirstColumn`, for showing or hiding the pin context menu items.

packages/components/src/components/hds/advanced-table/th-context-menu.ts
dchyun: All strings in the context menu should be translated now

packages/components/src/components/hds/advanced-table/index.hbs
dchyun: I think `isStickyColumn` is already in the bound args 
Screenshot 2025-08-11 at 2 57 29 PM

packages/components/src/components/hds/advanced-table/index.hbs
dchyun: I think `hasStickyColumn` is already in the bound args.
Screenshot 2025-08-11 at 2 57 19 PM

packages/components/src/components/hds/advanced-table/index.hbs
```diff
@@ -48,11 +48,13 @@
               @onClickSort={{if column.key (fn this._tableModel.setSortBy column.key)}}
               @align={{column.align}}
               @tooltip={{column.tooltip}}
+              @hasStickyFirstColumn={{this.hasStickyFirstColumn}}
               @hasResizableColumns={{@hasResizableColumns}}
-              @isStickyColumn={{if (and (eq index 0) @hasStickyFirstColumn) true}}
+              @isStickyColumn={{if (and (eq index 0) this.hasStickyFirstColumn) true}}
```
zamoore: Nit: This is fine, but for the sake of consistency/brevity/doing less, you could use `column.isFirst`.

packages/components/src/components/hds/advanced-table/index.hbs
```diff
@@ -62,16 +64,18 @@
               @align={{column.align}}
               @column={{column}}
               @hasExpandAllButton={{this._tableModel.hasRowsWithChildren}}
+              @hasStickyFirstColumn={{this.hasStickyFirstColumn}}
               @hasResizableColumns={{@hasResizableColumns}}
               @isExpanded={{this._tableModel.expandState}}
               @isExpandable={{column.isExpandable}}
-              @isStickyColumn={{if (and (eq index 0) @hasStickyFirstColumn) true}}
+              @isStickyColumn={{if (and (eq index 0) this.hasStickyFirstColumn) true}}
```
zamoore: Same here with `column.isFirst`

packages/components/src/components/hds/advanced-table/th-context-menu.ts
```diff
@@ -26,45 +28,79 @@ interface HdsAdvancedTableThContextMenuOption {
...
-          label: 'Reset column width',
+          label: translatedResetWidthLabel,
           icon: 'rotate-ccw',
           action: this.resetColumnWidth.bind(this),
         },
       ];
     }
 
+    if (hasStickyFirstColumn !== undefined && column.isFirst) {
+      const translatedPinLabel = this.hdsIntl.t(
```
zamoore, line 83: Woo! Translations. Nice.

packages/components/src/components/hds/advanced-table/th-context-menu.ts
```diff
@@ -92,4 +128,18 @@ export default class HdsAdvancedTableThContextMenu extends Component<HdsAdvanced
...
+    column: HdsAdvancedTableColumn,
+    dropdownCloseCallback?: () => void
+  ): void {
+    const { onPinFirstColumn } = this.args;
+
+    if (typeof onPinFirstColumn === 'function') {
+      onPinFirstColumn();
+    }
+
+    dropdownCloseCallback?.();
```
zamoore, line 141: Nit: I think I like the style of `typeof === 'function'` for nil checking functions, but I think we should stick to one style of checking inside of one function

zamoore: Also, I know this is my own code, but I think dropdownCloseCallback should always be present, right?

zamoore (commented):
Seems to work pretty well so far

packages/components/src/components/hds/advanced-table/index.hbs
dchyun: Updated!

packages/components/src/components/hds/advanced-table/th-context-menu.ts
dchyun: Looking at other components `typeof functionName === 'function'` seems to be the style used for nil checking functions. Was there another way of checking you wanted to use?

`dropdownCloseCallback` has to be marked as optional, because in the Dropdown blocks `close` is marked as optional. When setting it to required in the context menu the following errors get thrown.
Screenshot 2025-08-12 at 9 20 55 AM

packages/components/src/components/hds/advanced-table/index.hbs
shleewhite: Ah - never mind.

packages/components/src/components/hds/advanced-table/th-sort.ts
```diff
@@ -100,9 +101,13 @@ export default class HdsAdvancedTableThSort extends Component<HdsAdvancedTableTh
   }
 
   get showContextMenu(): boolean {
-    const { hasResizableColumns } = this.args;
+    const { column, isStickyColumn, hasResizableColumns } = this.args;
 
-    return hasResizableColumns ?? false;
+    return (
+      (hasResizableColumns ||
+        (isStickyColumn !== undefined && column && column.isFirst)) ??
```
shleewhite: Suggestion: in the handlebars file, there is a check for if column is defined before this is called. It could be good to move all the logic to this getter. Like this: 

```
get showContextMenu(): boolean {
  const { column, isStickyColumn, hasResizableColumns } = this.args;
  
   if (column) {
      return hasResizableColumns || (isStickyColumn !== undefined && column.isFirst)
   }
   return false
}
```

shleewhite: Screenshot 2025-08-12 at 1 15 22 PM

Heres the relevant bit of the hbs file^

shleewhite: Actually, also now that im looking at it - why do you need to check this isStickyColumn and column.isFirstColumn? Doesnt the logic from the index file and pass the correct bool to the th component already?

dchyun: The check for it the column is defined has to happen in the template, because `@column` is passed into the `ThContextMenu`.

I did remove the check for the column being defined, and the check for `column.isFirstColumn` from the getter since they were not needed.

jorytindall:
Looks like there's a small gap between the pinned column and the overflow shadow. Not sure if that was introduced here or elsewhere though. Looks like this might be an issue only on the examples where there _isn't_ multi-selection.

Screenshot 2025-08-14 at 1 25 45 PM

dchyun:
> Looks like there's a small gap between the pinned column and the overflow shadow. Not sure if that was introduced here or elsewhere though. Looks like this might be an issue only on the examples where there _isn't_ multi-selection.

@jorytindall thanks for catching this. I looked into it and it is an existing issue, but I've added a fix for it in here. This issue is seen when the first column width is a px value. In the existing sticky column example the width is `auto` which does not cause the issue. That's why it wasn't seen until now, as the new pinning examples have px widths.

@shleewhite Since you worked on the scroll indicators originally, could you look at my fix to make sure everything is working as expected? 

The PR description has been updated with more details on the fix.

.changeset/tangy-bushes-matter.md
```diff
@@ -0,0 +1,11 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+<!-- START components/table/advanced-table -->
+`AdvancedTable` - Added support for pinning first column in context menu
```
jorytindall: If there are multiple changes for the same component, I would list them as bullet points which is the recommendation in the formatting guidelines.
 
```suggestion
`AdvancedTable` - Multiple changes around pinning a column including:
- Added support for pinning first column in the context menu
- Translated template strings in context menu
- Fixed style for scroll indicator when first column is sticky and pas a px width
```

dchyun:
> [nit] As an aside: I know it's something tricky that wasn't feasible to solve during the original sticky column work, but I wonder if now it's worth revisiting the very small jump that occurs in the sticky column when scrolling on the X axis. 

@jorytindall I tested it out a little trying to fix this, and I think it still needs a more dedicated look to solve. The main issue I can see is there is a difference between when the `sticky` positioning gets triggered and when the internal logic to apply the pinned column classes gets triggered. That slight time gap is causing situations where one is applied when the other isn't.

packages/components/src/components/hds/advanced-table/index.ts
```diff
@@ -81,7 +82,7 @@ const getScrollIndicatorDimensions = (
     });
 
     // offsets the left: -1px position if there are multiple sticky columns
```
shleewhite:
```suggestion
    // offsets the left: -1px position if there are multiple sticky columns or the first column has a fixed pixel width
```

packages/components/src/components/hds/advanced-table/th-context-menu.ts
```diff
@@ -26,45 +28,77 @@ interface HdsAdvancedTableThContextMenuOption {
...
+      const translatedUnpinLabel = this.hdsIntl.t(
+        'hds.advanced-table.th-context-menu.unpin',
+        { default: 'Unpin column' }
+      );
+      options = [
+        ...options,
+        {
+          key: 'pin-first-column',
+          label: isStickyColumn ? translatedUnpinLabel : translatedPinLabel,
+          icon: 'pin',
```
jorytindall, line 96: I'm not sure how I didn't catch this before and we can bump it to a follow-on task if you want. Is it feasible to change the icon inversely if the column is pinned to "unpin" or "pin-off". We don't have this icon in the library currently, but I mocked up what it could look like.

Image

jorytindall (dismissed):
This all looks good to me, tested it in all of the examples. One questions about the icon usage, but not a blocker, and something that could be added in the future.

dchyun:
> The code changes seem good, but there is a regression in the existing sticky + selectable columns where I can see a gap between the sticky columns

@shleewhite I'm not able to see this on my end. Would you mind checking again to see if you still see the issue?

packages/components/src/components/hds/advanced-table/th-context-menu.ts
dchyun: @jorytindall I'm fine waiting to get the new icon into this PR. The only thing is I think @zamoore is waiting on this PR to get in before the reordering goes up for review. 

Zack is the reordering going to be ready to go soon? If so we could make this a follow up task. Don't want to be a blocker.

shleewhite (approved):

packages/components/src/components/hds/advanced-table/th-context-menu.ts
dchyun: It's been decided this new icon will be added in a follow-up task.

packages/components/src/components/hds/advanced-table/th-context-menu.ts
shleewhite: This would be fixed by https://github.com/hashicorp/design-system/pull/3118

dchyun: @shleewhite As part of your PR should we update the `dropdownCloseCallback` to be required?

shleewhite: Yeah, the PR already makes it so `D.close` which is used here is always defined

https://github.com/hashicorp/design-system/blob/cf13cb9195d5d84df498cd894c691ef9ba07deae/packages/components/src/components/hds/advanced-table/th-context-menu.hbs#L17

shleewhite: Yeah, the PR already makes it so `D.close` is always defined

https://github.com/hashicorp/design-system/blob/cf13cb9195d5d84df498cd894c691ef9ba07deae/packages/components/src/components/hds/advanced-table/th-context-menu.hbs#L17

PR #3117, opened 2025-08-20
fix(components): remove deprecated `@task` decorator

aklkv:
### :pushpin: Summary

remove deprecated @task decorator

`@task` decorator has advised against since v3 and in v5 it's removed
task arrow function are much much better for typescript as you can see I was able to remove multiple comments and add correct types. https://ember-concurrency.com/docs/task-concurrency

### :link: External links

Figma file: [if it applies]

***

<details>

</details>

### DIFF
--- a/packages/components/src/components/hds/time/index.ts
+++ b/packages/components/src/components/hds/time/index.ts
@@ -129,11 +129,11 @@ export default class HdsTime extends Component<HdsTimeSignature> {
   }
 
   @action
-  didInsertNode(): void {
+  async didInsertNode(): Promise<void> {
     const date = this.date;
 
     if (dateIsValid(date)) {
-      this.hdsTime.register(date);
+      await this.hdsTime.register(date);
     }
   }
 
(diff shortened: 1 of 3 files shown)


zamoore (approved):

didoo:
@aklkv @zamoore question: should this PR have had a changeset entry?

aklkv:
I would say no as both old and "newer" (has been around since v3) formats are backward compatible, this change makes sure we are not triggering deprecation warnings on consumer side if they upgrade to v5

alex-ju:
> I would say no as both old and "newer" (has been around since v3) formats are backward compatible, this change makes sure we are not triggering deprecation warnings on consumer side if they upgrade to v5

Ideally any component code change should have an associated changelog entry so if consumers encounter an issue during the HDS upgrade they have an ideea of where to look at. I understand that may not be the case here we prefer over-communicating.

I suggest the following, but open to improvements

```
`Time` - removed deprecated `@task` decorator to avoid warnings
```

@aklkv would you be able to raise a follow-up PR for this?

PR #3129, opened 2025-08-26
Showcase: convert `Accordion` page to `gts`

shleewhite:
### :pushpin: Summary

If merged, this PR would convert the Accordion showcase page to the `gts` format.

#### 2. Wrap the template in a component

<details>
  <summary>Example</summary>

**From**
```hbs
{{page-title "Accordion Component"}}

<Shw::Text::H1>Accordion</Shw::Text::H1>

<section data-test-percy>
  // rest of page
</section>
```

**To**
```ts

export interface PageComponentsAccordionSignature {
  Element: HTMLDivElement;
}

export default class PageComponentsAccordion extends Component<PageComponentsAccordionSignature> {
  <template>
    {{page-title "Accordion Component"}}

    <Shw::Text::H1>Accordion</Shw::Text::H1>

    <section data-test-percy>
      // rest of page
    </section>
  </template>
}
```
</details>

#### 3. Add imports + update naming convention

Anything that is not explicitly defined in the file needs to be imported. This includes:
* HDS components
* Showcase components
* Helpers or modifiers

<details>
<summary>Example</summary>

**From**
```ts

export interface PageComponentsAccordionSignature {
  Element: HTMLDivElement;
}

export default class PageComponentsAccordion extends Component<PageComponentsAccordionSignature> {
  <template>
    {{page-title "Accordion Component"}}

    <Shw::Text::H1>Accordion</Shw::Text::H1>

    <section data-test-percy>
      <Hds::Accordion @type={{type}} as |A|>
         <A.Item>
           <:toggle>Item one</:toggle>
           <:content>
             <Shw::Placeholder @text="generic content" @height="40" />
           </:content>
         </A.Item>
       </Hds::Accordion>
      // rest of page
    </section>
  </template>
}
```

**To**
```ts
import { pageTitle } from 'ember-page-title';

import ShwTextH1 from 'showcase/components/shw/text/h1';
import ShwPlaceholder from 'showcase/components/shw/placeholder';

import {  HdsAccordion } from '@hashicorp/design-system-components/components';

export interface PageComponentsAccordionSignature {
  Element: HTMLDivElement;
}

export default class PageComponentsAccordion extends Component<PageComponentsAccordionSignature> {
  <template>
    {{pageTitle "Accordion Component"}}

    <ShwTextH1>Accordion</ShwTextH1>

    <section data-test-percy>
       <HdsAccordion @type={{type}} as |A|>
         <A.Item>
           <:toggle>Item one</:toggle>
           <:content>
             <ShwPlaceholder @text="generic content" @height="40" />
           </:content>
         </A.Item>
       </HdsAccordion>
      // rest of page
    </section>
  </template>
}
```
</details>

#### 4. Switch out local helpers for global util helpers

- If there is a `noop` function defined in the component controller, switch it out for the global helper in the `gts` file
```ts
import NOOP from 'showcase/utils/noop';
```

#### 5. (if applicable) Move all content from the controller file to the `gts` file

This is only necessary if there is a controller file associated with the route. This should just be a simple copy paste, move any functions/utilities from the controller file to the `gts` file. Include any relevant imports.

The controller file can be deleted at this point.

#### 6. (if applicable) Move all model data from the route file to the `gts` file

Anything that is exported from the model can just be directly imported in the `gts` file.

<details>
<summary>Example</summary>

**From**
```ts
export default class PageComponentsAccordion extends Component<PageComponentsAccordionSignature> {
  <template>
    {{pageTitle "Accordion Component"}}

    <ShwTextH1>Accordion</ShwTextH1>

    <section data-test-percy>
       {{#each @model.SIZES as |size index|}}
       {{/each}}
      // rest of page
    </section>
  </template>
}
```

**To**
```ts
import { SIZES } from '@hashicorp/design-system-components/components/hds/accordion/item/index';

export default class PageComponentsAccordion extends Component<PageComponentsAccordionSignature> {
  <template>
    {{pageTitle "Accordion Component"}}

    <ShwTextH1>Accordion</ShwTextH1>

    <section data-test-percy>
       {{#each SIZES as |size index|}}
       {{/each}}
      // rest of page
    </section>
  </template>
}
```
</details>

Afterwards, the route file should look like this
```ts
/**
 */

import type { ModelFrom } from 'showcase/utils/ModelFromRoute';

export type PageComponentsAccordionModel =
  ModelFrom<PageComponentsAccordionRoute>;

export default class PageComponentsAccordionRoute extends Route {}

```

#### 7. (if applicable) Switch out `@action` for arrow functions

This is only necessary if there are instances of `@action` in the file.

<details>
<summary>Example</summary>

**From**
```js 
@action
onClickToggleSingle {
  this.stateSingle = this.stateSingle === 'open' ? 'close' : 'open';
};
```

**To**
```js
onClickToggleSingle = () => {
  this.stateSingle = this.stateSingle === 'open' ? 'close' : 'open';
};
```
</details>

#### 8. Break up showcase into code fragments and sub-sections

The naming convention for sub-components is as follows
```
showcase/
  app/
    components/
      page-components/
        component-name/
          index.gts
          code-fragments/
            with-{unique name}.gts
          sub-sections/
            {section title or something unique}.gts
```

Avoid using the component name in example file names to shorten the file path (the component is implied by the folder structure).

As you go through a file to determine the code fragments and sub-sections, consider the following:
- What is the goal of this example? Does the example as it exists accomplish this? 
- Is this example a unique instance of this situation? 
- Is there a case the showcase page currently doesn't show?

##### Code Fragments

It is generally easiest to create code fragments first as you can see the full page of examples and notice patterns across sub-sections.

Any demos that have custom logic/functions should be turned into a code fragment. They can also be any repeated example where the same basic instance of a component is used with maybe one or 2 args different.

If there are arguments for these components, they should be pretty limited and ideally match arguments from the HDS component. This way, as you scan the main file you can quickly identify what is different about a specific example.

##### Sub-sections

After you make the code fragments, you can break the sub-sections into their own components.

Each major section of the showcase should be broken into its own section component. What constitutes a section can be flexible, but generally its each h2 + the content below it.

#### 9. Create `index.gts` component

Now, the majority of the content from the route file is going to be moved into an index file in the components directory.

<details>
<summary>Example</summary>

**From**
```ts 
const PageComponentsAccordion: TemplateOnlyComponent = <template>
  {{pageTitle "Accordion Component"}}

  <ShwTextH1>Accordion</ShwTextH1>

  <section data-test-percy>
    <SectionVariants />
    <SectionContent />
    <SectionContext />
    <SectionBaseElements />
  </section>
</template>;
```

**To**
```ts
const PageComponentsAccordion: TemplateOnlyComponent = <template>
  <AccordionIndex />
</template>;

export default PageComponentsAccordion;

```
</details>

### :link: External links

Jira ticket: [HDS-5323](https://hashicorp.atlassian.net/browse/HDS-5323)

***

<details>

</details>

[HDS-5323]: https://hashicorp.atlassian.net/browse/HDS-5323?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


showcase/app/utils/component-states.ts
```diff
@@ -0,0 +1,2 @@
+// array of interaction states for presentation purposes in the showcase
+export const INTERACTION_STATES = ['default', 'active', 'hover', 'focus'];
```
didoo: Not all the components use the same interaction states: https://github.com/search?q=repo%3Ahashicorp%2Fdesign-system+content%3A%2F%28%3F-i%29STATES+%3D+%2F&type=code

In that case, what do we do?

didoo:
> There are 2 categories of sub-components: Mocks Sections
> Mocks are any demos that have custom logic/functions. They can also be any repeated example where the same basic instance of a component is used with maybe one or 2 args different.

@shleewhite the "mock" thing has got a bit out of hand (my fault). Initially "mock" was meant to indicate the "mocking" of an application, that is something used in a standalone page (with a mock "shell"/frame/navigation) to demo how a component (or combination of components) would look like in the context of a product-application (not a showcase demo/snippet).

If we decide to proceed with this proposal, we have to come up with a better distinction between the different entities (mock components for the mock application, repeated components/code in showcase, sections, etc)

showcase/app/components/page-components/accordion/mocks/mock-with-placeholder-content.gts
didoo, line 1: > `mock-with-placeholder-content.gts`

Also the example above contained placeholder content. 


(not to be picky: only to show how this now opens the door to conversations about what a file should be called to be descriptive/meaningful/semantically correct)

showcase/app/templates/page-components/accordion.gts
```diff
@@ -0,0 +1,583 @@
...
+            </:content>
+          </HdsAccordionItem>
+        </SF.Item>
+      </ShwFlex>
+
+      <ShwDivider @level={{2}} />
+
+      <ShwTextH4>States</ShwTextH4>
+
+      <SectionAccordionItemStates />
```
didoo: Not sure if this double format (code blocks with "mock" components / "section" components) is meant to show two possible directions, or if this is what is being proposed.

What I would suggest we to do is to stick to one format, in particular "everything is a section". Think of the format of how we include sub-sections in the `index` markdown files in the website. How this single format makes it simple to find things (_don't make me think™_) and has served us well, even when the file that was split contained just a few lines of text in some cases

didoo (commented):
Left a few high-level comments (to be read in the context of the related document https://docs.google.com/document/d/1rttAPsxO8Vl5_4zQKiLdMNpuX45JtuKHdOGzW43N3Z8/)

showcase/app/templates/page-components/accordion.gts
shleewhite: This doesn't work though as we get to complex repeated examples, especially table/advanced table/key value inputs/etc. There is a genuine use case for having a sub component of just one example. Would that then lead to having section components that use sub-components?

showcase/app/utils/component-states.ts
shleewhite: I think we could export another const that doesnt have active and decide as a team how we handle disabled as that is the other "state" that is inconsistent across showcase pages.

Sometimes we show disabled states and other times we dont. In accordion, there are no examples of disabled accordions. Is a disabled accordion allowed? What about an accordion that has `containsInteractive=true`?

showcase/app/components/page-components/accordion/with-external-control.gts
```diff
@@ -0,0 +1,73 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+import Component from '@glimmer/component';
```
KristinLBradley, line 5:
```suggestion

import Component from '@glimmer/component';
```

showcase/app/templates/page-components/accordion.gts
```diff
@@ -0,0 +1,688 @@
...
+import ShwDivider from 'showcase/components/shw/divider';
+import ShwFlex from 'showcase/components/shw/flex';
+import ShwGrid from 'showcase/components/shw/grid';
+import ShwPlaceholder from 'showcase/components/shw/placeholder';
+import ShwTextBody from 'showcase/components/shw/text/body';
+import ShwTextH1 from 'showcase/components/shw/text/h1';
+import ShwTextH2 from 'showcase/components/shw/text/h2';
+import ShwTextH3 from 'showcase/components/shw/text/h3';
+import ShwTextH4 from 'showcase/components/shw/text/h4';
+import { INTERACTION_STATES } from 'showcase/utils/component-states';
```
KristinLBradley:
```suggestion

import { INTERACTION_STATES } from 'showcase/utils/component-states';
```

showcase/app/templates/page-components/accordion.gts
```diff
@@ -0,0 +1,688 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
```
KristinLBradley, line 4:
```suggestion
 */
 
```

showcase/app/templates/page-components/accordion.gts
```diff
@@ -0,0 +1,688 @@
...
+            <HdsAccordion @type={{type}} as |A|>
+              <A.Item>
+                <:toggle>
+                  <ShwPlaceholder @text="generic content" @height="24" />
+                </:toggle>
+                <:content>
+                  <ShwPlaceholder @text="generic content" @height="40" />
+                </:content>
+              </A.Item>
+            </HdsAccordion>
```
KristinLBradley: So if a component example is more one-off we keep it within the main GTS file vs. creating a sub-component? (just confirming)

showcase/app/utils/component-states.ts
KristinLBradley: Adding a few different variants for different included states here makes sense to me, as Lee suggested. I'm not sure how we would name the other variants though.

KristinLBradley: Maybe you could do something like this:
```
export const INTERACTION_STATES_ALL = ['default', 'active', 'hover', 'focus', 'disabled'];
export const INTERACTION_STATES_NO_ACTIVE = ['default', 'hover', 'focus', 'disabled'];
export const INTERACTION_STATES_NO_DISABLED = ['default', 'active', 'hover', 'focus'];
```

As long as the list doesn't grow crazy-long I don't think it looks _too_ bad...

showcase/app/components/page-components/accordion/mocks/mock-with-placeholder-content.gts
KristinLBradley: I think the new naming with just "with-blah.gts" and then using them within the main GTS file as "ComponentNameWithBlah" works pretty well (seems pretty clear to understand to me).

showcase/app/templates/page-components/accordion.gts
shleewhite: Yup! Seems like the easiest way (as opposed to sections)

showcase/app/utils/component-states.ts
didoo: these are the current _distinct_ combinations we have:
```
['default', 'hover', 'active', 'focus']; → 19 instances
['default', 'hover', 'active', 'focus', 'disabled']; → 3 instances
['default', 'hover', 'focus']; → 10 instances
['default', 'hover', 'focus', 'disabled']; → 1 instance
['default', 'hover', 'active']; → 2 instances
```
So, what do we do?

This for me is a "smell" that an approach that looked simple and obvious, a no-brainer, it's actually more complex. And this is just a simple example, so... what else are we not considering, in this first PR? Can we really infer the most common use cases and code pattern from a single component migration? Of course this is going to be an iterative process, I am conscious, but for me this means there should be more scrutiny on the patterns we're defining, the changes we're making, to avoid underestimating the effect of these changes and potentially introduce more problems than the ones we're solving.

showcase/app/templates/page-components/accordion.gts
didoo: > Would that then lead to having section components that use sub-components?

yeah, why not? it's easy to find things anyway:

- index → index of all the sections, just a simple list of "includes" (like for the website `index.md`)
- sections folder → all the "section-like" files in which the content is broken down (not necessary a single `<section>` or a block of every `<h2>` content) 
- snippets folder (name tbd) → all the repeated code blocks that we want to reuse in a section, in another code snippet, etc

Essentially, it's like using Atomic Design, only for folder/file structure.

showcase/app/templates/page-components/accordion.gts
didoo: So, this results in a half-hearted solution for me: we abstract away some code, but the "main" file remains extremely long, which was one of the pain-point that was highlighted. 

The previous file (`accordion.hbs)`) was 789 lines, the proposed one (`accordion.gts`) is still 688 lines, so "just" a ~10% reduction.

I think that, if we are decide to breaking down the code, we should go all-in and use "sections" too.

showcase/app/components/page-components/accordion/with-external-control.gts
didoo, line 1: I have strong opinions on this: I think the "fragments" (name tbd, repeated code blocks in a template/page) should be colocated with the "index" main page, and the sections (of course, organized under folders), under the "component-name" parent folder.

didoo (commented):
I have left a few more comments.

I have also created a branch (show is easier than tell) about what I had in mind:
https://github.com/hashicorp/design-system/compare/hds-5323/showcase-accordion-gts...hds-5323/cristiano-test?expand=1

(the "main" file `accordion.gts` should live under the `accordion` folder as `index.gts` but if I do it, for reasons that I don't understand, the page does not render its content, but it doesn't trigger an error)

showcase/app/templates/page-components/accordion.gts
KristinLBradley: Perhaps we could start with just converting pages to the GTS format and then figure out when & how to break things into subcomponents as a 2nd phase?

I think it may be easier to review PRs that way at least and we could make some progress on part of the intended changes. During the process of converting to GTS, we may also gain insights into logical ways to break down page content that would help with the 2nd phase.

KristinLBradley:
> I have left a few more comments.
> 
> I have also created a branch (show is easier than tell) about what I had in mind: https://github.com/hashicorp/design-system/compare/hds-5323/showcase-accordion-gts...hds-5323/cristiano-test?expand=1
> 
> (the "main" file `accordion.gts` should live under the `accordion` folder as `index.gts` but if I do it, for reasons that I don't understand, the page does not render its content, but it doesn't trigger an error)

This may be a better link to see Cristiano's test code showing his alternate structure proposal:
https://github.com/hashicorp/design-system/blob/5faa25e55682010e3fa624f3bbfb988422b1dd1b/showcase/app/templates/page-components/accordion.gts

showcase/app/templates/page-components/accordion.gts
KristinLBradley: @alex-ju mentioned to me that it may even be possible to use a codemod to do just the GTS conversion part, although that would mean less hands-on experience with the format for the team.

shleewhite: There is definitely an option to use the code mod - I just thought the changes might be overwhelming bc it would all happen at once and also thought people on the team would like hands on experience with the format. Like maybe if we go through this process with the showcase, we'd feel more comfortable using the code mod (and verifying its output) for the actual components.

shleewhite: @didoo that does not take into account the fact that the gts file also contains all the content from the model and controller files plus the imports. If you check out [this commit](https://github.com/hashicorp/design-system/pull/3129/commits/cb1db38cf64243b582b32750e68265810e31dd72) - you can see that after the initial steps to convert to gts, the file balloons to 935 lines. so getting it down to 688 is a more significant reduction.

I am trying to make a suggestion in this PR that balances the concern of keeping the showcase a playground where ideas can be quickly tested/added while reducing repeated logic and keep the files more focused. I am not sure that making every "section" its own component is inline with these goals. It seems like a lot more overhead for less payoff.

didoo: > @didoo that does not take into account the fact that the gts file also contains all the content from the model and controller files plus the imports.

No no I knew it. But it's the argument that "the conversion helps with reducing the length of our files" that's it's a bit moot (if the criteria is counting the overall lines of code in a single file).

didoo: >  I am not sure that making every "section" its own component is inline with these goals

Sure, but in doing so it's trying to preserve a (very important, at least for me) "feature" of the showcase, that I fear is getting lost: the consistency of how files are organized, internally structured, where to find things, etc

With the current proposal, there is a free for all on how to structure things: one component page could be a single page, others could be broken down in sub-parts, other in sub-parts with sub-parts, potentially some also in "sections" because they're too long anyway.

As I told you in different conversations, I am not just looking at what we're gaining from this conversion, but also at what we're losing, that is invisible because we're now used to it, it's a given and we don't even think of it, but has taken a lot of time (and sweat) to get to.

showcase/app/utils/component-states.ts
shleewhite: This list shows me that we sometimes treat it as its own state but other times we test how disabled + another interaction state works. We do not show the disabled state consistently across showcase pages, could we improve this? And then, if we decide to make it more consistent we could then consolidate some of these arrays.

You can see this with the Text Input page, it shows disabled + hover and disabled + default state. But for button, we only show disabled with the default state. 

Another thing is that some showcase pages don't show disabled states at all. the CodeBlock page does not show an example of a disabled CodeBlock copy button - but it is possible. Should we have an example of that?

showcase/app/templates/page-components/accordion.gts
shleewhite: I'm fine to say the file length thing is moot because it varies a lot from component to component. This page, it was small but for something like Advanced Table it could be much lager (the spike branch doesn't follow the same conventions as this PR, [but it was down 50%](https://github.com/hashicorp/design-system/pull/3038/commits/9d4a81320bbf91a8cf2e650b2011fcfe5b54ca99#diff-60222cfb24f95c4e6176326c3a16ffbc04ee86379d0cac682096e8bca4831d9f))

shleewhite:
> I have also created a branch (show is easier than tell) about what I had in mind: https://github.com/hashicorp/design-system/compare/hds-5323/showcase-accordion-gts...hds-5323/cristiano-test?expand=1

@didoo Yeah, I have tried collocating the sub-components in the initial spike for converting to gts and it basically doesn't work with ember. Ember everything in the template folder to be a route. The sub-components have to live in the components folder.

showcase/app/components/page-components/accordion/code-fragments/with-external-control.gts
dchyun, line 1: [Suggestion] Just to simplify things could we remove the `with` from the code fragement names?

dchyun: We could also move `CodeFragment` to the end of the class name. There's currently a misalignment with this compared to the section component names. The code fragment names are `CodeFragmentWithName` and the section names are `NameSection`. 

I don't have a strong preference if "CodeFragment" or "Section" goes at the beginning or end, but could align them.

shleewhite: Good point! I fixed this by switching how the sections are named.

showcase/app/components/page-components/accordion/sections/base-elements.gts
```diff
@@ -0,0 +1,267 @@
...
+import ShwPlaceholder from 'showcase/components/shw/placeholder';
+import ShwTextBody from 'showcase/components/shw/text/body';
+import ShwTextH2 from 'showcase/components/shw/text/h2';
+import ShwTextH3 from 'showcase/components/shw/text/h3';
+import ShwTextH4 from 'showcase/components/shw/text/h4';
+import NOOP from 'showcase/utils/noop';
+
+import { HdsAccordionItem } from '@hashicorp/design-system-components/components';
+import HdsAccordionItemButton from '@hashicorp/design-system-components/components/hds/accordion/item/button';
+import { SIZES } from '@hashicorp/design-system-components/components/hds/accordion/item/index';
```
dchyun:
```suggestion
import { SIZES, TYPES } from '@hashicorp/design-system-components/components/hds/accordion/item/index';
```
[Nit] Can consolidate imports

showcase/app/components/page-components/accordion/sections/variants.gts
```diff
@@ -0,0 +1,64 @@
...
+import type { TemplateOnlyComponent } from '@ember/component/template-only';
+import { capitalize } from '@ember/string';
+import { lt } from 'ember-truth-helpers';
+
+import ShwDivider from 'showcase/components/shw/divider';
+import ShwGrid from 'showcase/components/shw/grid';
+import ShwTextH2 from 'showcase/components/shw/text/h2';
+import ShwTextH3 from 'showcase/components/shw/text/h3';
+
+import { SIZES } from '@hashicorp/design-system-components/components/hds/accordion/item/index';
```
dchyun:
```suggestion
import { SIZES, TYPES } from '@hashicorp/design-system-components/components/hds/accordion/item/index';
```
[Nit] Can consolidate imports

showcase/app/components/page-components/accordion/code-fragments/with-external-control.gts
```diff
@@ -0,0 +1,73 @@
...
+import type { HdsAccordionForceStates } from '@hashicorp/design-system-components/components/hds/accordion/types';
+
+export interface WithExternalControlSignature {
+  Args: {
+    variant: 'all' | 'single';
+  };
+  Element: HTMLDivElement;
+}
+
+export default class WithExternalControl extends Component<WithExternalControlSignature> {
```
dchyun:
```suggestion
export default class CodeFragmentWithExternalControl extends Component<WithExternalControlSignature> {
```
[Suggestion][Nit] When these are imported into the section files they are renamed to `CodeFragmentWithName` but in these component files they're `WithName`. Could align them just for consistency's sake. The section class names are defined as `SectionName`.

dchyun (commented):
Everything is looking good to me structure wise. I'm on board with what is here. Just a few other comments

- In the PR description we could add notes in the breaking up section about creating the index componnet file and the main template file. 
- A task not related to this, but we will need to update our [confluence notes](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3197796483/How+to+use+the+Showcase+for+components+development) to match these changes.

showcase/app/templates/page-components/accordion.gts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+import type { TemplateOnlyComponent } from '@ember/component/template-only';
+
+import IndexAccordion from 'showcase/components/page-components/accordion';
```
didoo: How about `AccordionIndex`?
```suggestion
import AccordionIndex from 'showcase/components/page-components/accordion';
```

showcase/app/components/page-components/accordion/index.gts
```diff
@@ -0,0 +1,28 @@
...
+import { pageTitle } from 'ember-page-title';
+
+import ShwTextH1 from 'showcase/components/shw/text/h1';
+
+import SectionBaseElements from 'showcase/components/page-components/accordion/sections/base-elements';
+import SectionVariants from 'showcase/components/page-components/accordion/sections/variants';
+import SectionContent from 'showcase/components/page-components/accordion/sections/content';
+import SectionContext from 'showcase/components/page-components/accordion/sections/context';
+
+const IndexAccordion: TemplateOnlyComponent = <template>
```
didoo: Should this be `AccordionIndex`?
```suggestion
const AccordionIndex: TemplateOnlyComponent = <template>
```

showcase/app/components/page-components/accordion/index.gts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+import type { TemplateOnlyComponent } from '@ember/component/template-only';
+import { pageTitle } from 'ember-page-title';
+
+import ShwTextH1 from 'showcase/components/shw/text/h1';
+
+import SectionBaseElements from 'showcase/components/page-components/accordion/sections/base-elements';
```
didoo: since, as we've seen, the word `section` leads to confusion/misunderstanding (also, the `<section>` element is in the `index` component) I would suggest to call this folder `sub-sections` and the components `SubSectionXyx` so we have s little more latitude in what constitutes a sub-section.

showcase/app/components/page-components/accordion/sections/base-elements.gts
```diff
@@ -0,0 +1,269 @@
...
+import { HdsAccordionItem } from '@hashicorp/design-system-components/components';
+import HdsAccordionItemButton from '@hashicorp/design-system-components/components/hds/accordion/item/button';
+import {
+  SIZES,
+  TYPES,
+} from '@hashicorp/design-system-components/components/hds/accordion/item/index';
+
+const INTERACTION_STATES = ['default', 'active', 'hover', 'focus'];
+
+const SectionBaseElements: TemplateOnlyComponent = <template>
```
didoo: see comment above about calling this `SubSectionXyz`

didoo (commented):
I have left a few comments, mainly around naming things, but overall the structure and organization as is now works for me.

showcase/app/routes/page-components/accordion.ts
```diff
@@ -5,18 +5,9 @@
 
 import Route from '@ember/routing/route';
 
-import { SIZES } from '@hashicorp/design-system-components/components/hds/accordion/item/index';
-import { TYPES } from '@hashicorp/design-system-components/components/hds/accordion/item/index';
-
 import type { ModelFrom } from 'showcase/utils/ModelFromRoute';
 
 export type PageComponentsAccordionModel =
```
dchyun, line 10: [Question] Now that we've removed the accordion controller that was using this type, can we remove this?

showcase/app/components/page-components/accordion/sub-sections/base-elements.gts
```diff
@@ -0,0 +1,269 @@
...
+            <ShwPlaceholder @text="generic content" @height="40" />
+          </:content>
+        </HdsAccordionItem>
+      </SG.Item>
+    {{/each}}
+  </ShwGrid>
+
+  <ShwDivider @level={{2}} />
+
+  <ShwTextH4>isOpen</ShwTextH4>
```
dchyun, line 63: [Issue] Looks like the "containsInteractive" section was missed in the migration, which previously followed "isOpen".

showcase/app/routes/page-components/accordion.ts
shleewhite: I think its best practice to leave it, but @aklkv can confirm.

showcase/app/components/page-components/accordion/sub-sections/base-elements.gts
shleewhite: I actually removed it on purpose, it was a duplicate of the type section since it showed both types + containsInteractive=true/false.

I can put it back tho if we think it is necessary.

dchyun: Thanks for clarifying, think it should be all good to remove it then!

showcase/app/routes/page-components/accordion.ts
dchyun: If it's best practice then all good to leave it

didoo (approved):

PR #3142, opened 2025-09-02
Web docs - Adjusting guidance for forms in Flyouts

majedelass:
### :pushpin: Summary

If merged, this PR will address the "no forms in flyouts" where we provide a better more appropriate example and guidelines that align with how consumers are using it AND what we recommend.

### :link: External links

Jira ticket: [HDS-4825](https://hashicorp.atlassian.net/browse/HDS-4825)

***


website/docs/components/flyout/partials/guidelines/guidelines.md
```diff
@@ -68,13 +68,20 @@ While code snippets and terminal scripts are usually detailed, they are well sui
 
 #### Forms
 
-Don't use a form in a Flyout, whether the function is creating an object, or updating an existing one. This type of content is often complex and more appropriate on its own page, or in the case of a simple form, within a [Modal](/components/modal).
+Only use forms in a flyout when they directly support the page's primary function. For example, if the page's main purpose is to view a list of labels, editing them becomes a secondary action. In this case, it makes sense to put the edit function in a flyout.
```
jorytindall:
```suggestion
Only use forms in a Flyout when they directly support the page's primary function. For example, if the page's main purpose is to view a list of labels, editing them becomes a secondary action. In this case, it makes sense to put the edit function in a Flyout.
```

website/docs/components/flyout/partials/guidelines/guidelines.md
```diff
@@ -68,13 +68,20 @@ While code snippets and terminal scripts are usually detailed, they are well sui
...
 
-!!! Dont
+!!! Do
+
+In this context, editing is a simple task, and the form placed will require minimal effort from the user while also supporting their review of the contents of the table.
```
jorytindall: I might rework this a bit, what do you think of this:

```suggestion
It's acceptable to place a form in a Flyout for simple editing experiences that require minimal user effort when the object being edited is on the page that triggers the Flyout.
```

website/docs/components/flyout/partials/guidelines/guidelines.md
```diff
@@ -68,13 +68,20 @@ While code snippets and terminal scripts are usually detailed, they are well sui
...
+
+![A table with a list of labels, their associated values and applied workspaces and projects. There is a search bar above the table. The user is clicking on one of the edit buttons in the table.](/assets/components/flyout/flyout-forms-assisting-tasks-01.png)
+
+![A flyout is opened which has a form within that edits a single label, its values and other related inputs.](/assets/components/flyout/flyout-forms-assisting-tasks-02.png)
 
-![Form within a Flyout](/assets/components/flyout/form-in-flyout.png)
 !!!
 
+If you feel the contents of the form are too long, or complex (require multiple layers of editing features), it should be placed on its own page.
```
jorytindall: I think this should be turned into an inverse Don't block. Maybe alter the content to something like this:


```suggestion
Don't use a Flyout for forms that are long, complex, or require multiple layers of editing. Instead, place the form on a dedicated page.
```

jorytindall (commented):
I think this adds clarity to the form usage, but I think a broader initiative is probably necessary to understand the following (based on our sync today:
1. How consumers are using Flyouts outside of this recommendation, _why_ (this is the most important part and may warrant revisiting the recommendations holistically), and which of these examples is approved or has been built.
2. Whether we feel strongly enough about this guidance that we need to work with consumers to better adhere to it. Based on the spread of examples you shared there are some that are definitely reasonable, and some that are way outside of the guidelines.

majedelass:
> I think this adds clarity to the form usage, but I think a broader initiative is probably necessary to understand the following (based on our sync today:
> 
> 1. How consumers are using Flyouts outside of this recommendation, _why_ (this is the most important part and may warrant revisiting the recommendations holistically), and which of these examples is approved or has been built.
> 2. Whether we feel strongly enough about this guidance that we need to work with consumers to better adhere to it. Based on the spread of examples you shared there are some that are definitely reasonable, and some that are way outside of the guidelines.

@jorytindall Agreed to a broader initiative - also an initiative with HDS academy could be very useful!

didoo:
@majedelass can you add a quick note/comment on why the PR was closed, for future reference? thanks

PR #3146, opened 2025-09-04
Showcase: convert `CopyButton` page to `gts`

shleewhite:
### :pushpin: Summary

If merged, this PR would convert the CopyButton showcase page to `gts` following the steps defined here: https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/4220158068/How+to+convert+showcase+pages+to+gts+format

Most of the work is adapted from the previous spike (https://github.com/hashicorp/design-system/pull/3038/commits/95663f111a8d89045497587596274c453bab8e5d#diff-a1d613da429cbc1faaa78aa303e1d717892ad8424d6a5e0e3c93580ba920608a)

### :link: External links

Jira ticket: [HDS-5338](https://hashicorp.atlassian.net/browse/HDS-5338)

***

<details>

</details>


showcase/app/components/page-components/copy/button/code-fragments/with-generic-dialog-content.gts
```diff
@@ -0,0 +1,70 @@
...
+} from '@hashicorp/design-system-components/components';
+
+import ShwDivider from 'showcase/components/shw/divider';
+
+export interface CodeFragmentWithGenericDialogContentSignature {
+  Element: HTMLDivElement;
+}
+
+const CodeFragmentWithGenericDialogContent: TemplateOnlyComponent = <template>
+  <div class="shw-component-copy-button-demo-container">
```
aklkv: [nit]: to avoid potantial confusion, I suggest always including splattributes. There are couple of very useful rules that can be customized and catch this cases
* https://github.com/ember-template-lint/ember-template-lint/blob/main/docs/rule/require-splattributes.md
* https://github.com/ember-template-lint/ember-template-lint/blob/main/docs/rule/no-nested-splattributes.md

KristinLBradley:
Are these diffs with the copy buttons intended? (from Percy)

image

I mean the top row of buttons in the images in which the button text "Copy the input value" no longer displays and the copy button alignment has moved.

shleewhite:
> Are these diffs with the copy buttons intended? (from Percy)
> 
> image
> I mean the top row of buttons in the images in which the button text "Copy the input value" no longer displays and the copy button alignment has moved.

@KristinLBradley ah thanks for calling this out. Some of the diffs were intentional, but others weren't. 
- I forgot to copy over changes I made in the spike branch to the css for the page so the alignment was messed up, so that should be resolved now.
- I simplified it so all the examples with hds inputs next to a copy button have an icon only button so I could make it a resuable component more easily.

showcase/app/components/page-components/copy/button/sub-sections/demos.gts
```diff
@@ -0,0 +1,276 @@
...
+import ShwTextH2 from 'showcase/components/shw/text/h2';
+import ShwTextH3 from 'showcase/components/shw/text/h3';
+import ShwTextH4 from 'showcase/components/shw/text/h4';
+
+import {
+  HdsCopyButton,
+  HdsDropdown,
+} from '@hashicorp/design-system-components/components';
+
+import CopyButtonWithMaskedInput, {
```
dchyun:
```suggestion
import CodeFragmentWithMaskedInput, {
```
[Suggestion] The class name exported from the code fragment file starts with `CodeFragment`, so could align to that. I believe the accordion imports used `CodeFragment` as well. 

Note: This comment would apply across both sub section files

showcase/app/components/page-components/copy/button/sub-sections/demos.gts
```diff
@@ -0,0 +1,276 @@
...
+              <code>copy</code>
+              the
+              <a href="#">text</a>
+              in this
+              <cite>target</cite>
+              element.</p>
+          </li>
+        </ul>
+      </SG.Item>
+      <SG.Item @label="With hidden content">
```
dchyun, line 413: [Question] Is the intention that the hidden text here is copied as well? Currently I see only the visible text getting copied. I see this in `main` as well, so not something introduced from the conversion.

showcase/app/components/page-components/copy/button/sub-sections/demos.gts
```diff
@@ -0,0 +1,276 @@
...
+
+    <ShwGrid @columns={{2}} @gap="2rem" as |SG|>
+      <SG.Item @label="Code block">
+        <HdsCopyButton
+          @text="Copy the code block"
+          @targetToCopy="#test-code-block"
+        />
+        <pre class="shw-component-copy-button-code-block"><code
+            id="test-code-block"
+          >&lt;h1&gt;A page header example&lt;/h1&gt; &lt;p&gt;Some paragraph
```
dchyun:
```suggestion
          >&lt;h1&gt;A page header example&lt;/h1&gt;
&lt;p&gt;Some paragraph
```
[Nit] Make paragraph tag go to new line

showcase/app/components/page-components/copy/button/sub-sections/demos.gts
```diff
@@ -0,0 +1,276 @@
...
+      </SG.Item>
+      <SG.Item @label="Code block with 'contenteditable'">
+        <HdsCopyButton
+          @text="Edit and copy the code block"
+          @targetToCopy="#test-code-block-editable"
+        />
+        <pre class="shw-component-copy-button-code-block"><code
+            id="test-code-block-editable"
+            contenteditable="true"
+          >&lt;h1&gt;A page header example&lt;/h1&gt; &lt;p&gt;Some paragraph
```
dchyun:
```suggestion
          >&lt;h1&gt;A page header example&lt;/h1&gt;
&lt;p&gt;Some paragraph
```
[Nit] Make paragraph tag go to new line

showcase/app/components/page-components/copy/button/code-fragments/with-html-input.gts
```diff
@@ -0,0 +1,100 @@
...
+          <select id="test-input-{{@inputComponent}}">
+            <option>Lorem ipsum dolor</option>
+            <option>Sit amet</option>
+            <option>Consectetur adipiscing elit</option>
+          </select>
+        {{else}}
+          <input
+            type={{@inputComponent}}
+            value={{get inputComponentToValueMap @inputComponent}}
+            id="test-input-{{@inputComponent}}"
```
dchyun: [Suggestion] Use guidFor or make the id here more unique in some way. Currently there are multiple of the same ids on the page, which is causing mismatches when copying.

Ex: There are multiple `test-input-textarea` ids. If I copy the text of the "HTML input elements" textarea example it should say "Copy the generic textarea value" but what it actually copies is "This is a normal multiline text that should be copied", which comes from a text area in a previous section.

showcase/app/components/page-components/copy/button/sub-sections/demos.gts
shleewhite: Another reason to have integration tests for these things -____-

I'm assuming the intent is for all of it to be copied, but I'm going to look into it.

showcase/app/components/page-components/copy/button/code-fragments/with-html-input.gts
shleewhite: Totally! Don't know why I didnt do that from the start 🤔 It is updated now

showcase/app/components/page-components/copy/button/sub-sections/demos.gts
shleewhite: Ok I think it was never meant to copy the hidden content.

I found the PR where this example was added and it didnt work that way back then in the deployment: https://hds-showcase-git-release-3-0-0-hashicorp.vercel.app/components/copy/button

showcase/app/components/page-components/copy/button/code-fragments/with-html-input.gts
```diff
@@ -0,0 +1,102 @@
...
+    <div class="shw-component-copy-button-flex-container" ...attributes>
+      <div>
+        <label for="test-input-{{this.id}}" class="sr-only">
+          {{capitalize @inputComponent}}
+          input
+        </label>
+
+        {{#if (eq @inputComponent "textarea")}}
+          <textarea
+            id="test-input-{{@inputComponent}}"
```
dchyun:
```suggestion
            id="test-input-{{this.id}}"
```
[todo] This fix looks all good, just missed one spot.

didoo:
> * I simplified it so all the examples with hds inputs next to a copy button have an icon only button so I could make it a resuable component more easily.

The fact that we were testing both cases was intentional (I am pretty sure it was to emulate a use case that we found in a product, where they needed an explicit text for the copy button, near an readonly input)

showcase/app/components/page-components/copy/button/sub-sections/demos.gts
```diff
@@ -0,0 +1,276 @@
...
+  HdsCopyButton,
+  HdsDropdown,
+} from '@hashicorp/design-system-components/components';
+
+import CodeFragmentWithMaskedInput, {
+  MASKED_INPUT_VARIANTS,
+} from 'showcase/components/page-components/copy/button/code-fragments/with-masked-input';
+import type { CodeFragmentWithMaskedInputSignature } from 'showcase/components/page-components/copy/button/code-fragments/with-masked-input';
+import CodeFragmentWithHdsInput, {
+  INPUT_COMPONENTS,
```
didoo: [suggestion] maybe?

```suggestion
  INPUT_COMPONENTS as HDS_INPUT_COMPONENTS,
```

showcase/app/components/page-components/copy/button/code-fragments/with-html-input.gts
```diff
@@ -0,0 +1,102 @@
...
+  'number',
+  'url',
+  'email',
+  'date',
+  'time',
+  'range',
+  'color',
+  'textarea',
+  'select',
+];
```
didoo: [suggestion] according to Copilot, this could be simplified?

```suggestion
export const INPUT_COMPONENTS = [
  'text',
  'password',
  'number',
  'url',
  'email',
  'date',
  'time',
  'range',
  'color',
  'textarea',
  'select',
];

type inputComponent = typeof INPUT_COMPONENTS[number];
```

showcase/app/components/page-components/copy/button/code-fragments/with-html-input.gts
didoo: in this case, can we remove `test-input-` at all? not needed anymore

showcase/app/components/page-components/copy/button/code-fragments/with-masked-input.gts
```diff
@@ -0,0 +1,92 @@
...
+  HdsFormField,
+  HdsFormMaskedInputField,
+} from '@hashicorp/design-system-components/components';
+
+type variant =
+  | 'masked-input-base'
+  | 'masked-input-base-form-field'
+  | 'masked-input-field';
+
+export const MASKED_INPUT_VARIANTS: variant[] = [
```
didoo: [suggestion] see comment above about a possible simplification to avoid duplicating the list

showcase/app/components/page-components/copy/button/code-fragments/with-masked-input.gts
```diff
@@ -0,0 +1,92 @@
...
+  Args: {
+    isMultiline?: boolean;
+    variant: variant;
+  };
+  Element: HTMLDivElement;
+}
+
+const CodeFragmentWithMaskedInput: TemplateOnlyComponent<CodeFragmentWithMaskedInputSignature> =
+  <template>
+    <div ...attributes>
```
didoo: [personal consideration] we're adding `...attributes` even if not needed in the instantiation. what do you think of taking a more minimalistic approach, and add them only when needed?

showcase/app/components/page-components/copy/button/sub-sections/demos.gts
```diff
@@ -0,0 +1,276 @@
...
+
+  <template>
+    <ShwTextH2>Compositions</ShwTextH2>
+
+    {{#each MASKED_INPUT_VARIANTS as |variant|}}
+      <ShwFlex as |SF|>
+        {{#let (array false true) as |isMultilineOptions|}}
+          {{#each isMultilineOptions as |isMultiline|}}
+            <SF.Item @label={{get maskedInputVariantToLabelMap variant}}>
+              <CodeFragmentWithMaskedInput
```
didoo: [question] from what I see, `<CodeFragmentWithMaskedInput>` has only this instance. Why then have a code fragment for it (which comes with some extra code, like the `maskedInputVariantToLabelMap` above)? maybe the "Compositions" could be its own sub-section, and contain all the logic related to these examples?

showcase/app/components/page-components/copy/button/sub-sections/demos.gts
```diff
@@ -0,0 +1,276 @@
...
+  //   text,
+  //   target,
+  // }: {
+  //   trigger: HTMLElement;
+  //   text: string;
+  //   target: string;
+  // }) {
+  //   console.log('onError invoked in the controller', trigger, text, target);
+  // }
+
```
didoo: [suggestion] looking at the code, is nothing special, easy to re-create if needed, so probably we can clean up this old comments

```suggestion
```

showcase/app/components/page-components/copy/button/sub-sections/demos.gts
didoo: > Ok I think it was never meant to copy the hidden content.

Yes, that was the intent (I implemented the code at that time)

showcase/app/components/page-components/copy/button/code-fragments/with-masked-input.gts
shleewhite: I think its best practice to include just in case it is needed at some point. It definitely wont be needed for the sub section components but im not sure we can say that so definitively for the code fragments.

showcase/app/components/page-components/copy/button/sub-sections/demos.gts
shleewhite: Totally fair, I removed this code fragment.

showcase/app/components/page-components/copy/button/code-fragments/with-html-input.gts
```diff
@@ -0,0 +1,90 @@
...
+            value={{get inputComponentToValueMap @inputComponent}}
+            id={{this.id}}
+            min={{if (eq @inputComponent "range") "0" undefined}}
+            max={{if (eq @inputComponent "range") "10" undefined}}
+          />
+        {{/if}}
+      </div>
+      <HdsCopyButton
+        @text="Copy the {{@inputComponent}} input value"
+        @targetToCopy={{this.id}}
```
dchyun:
```suggestion
        @targetToCopy="#{{this.id}}"
```
[Issue] The copy button isn't working because the selector isn't quite right for the id.

showcase/app/components/page-components/copy/button/index.gts
```diff
@@ -0,0 +1,24 @@
...
+
+import ShwTextH1 from 'showcase/components/shw/text/h1';
+
+import SubSectionBaseElements from 'showcase/components/page-components/copy/button/sub-sections/base-elements';
+import SubSectionDemos from 'showcase/components/page-components/copy/button/sub-sections/demos';
+
+const CopyButtonIndex: TemplateOnlyComponent = <template>
+  {{pageTitle "CopyButton Component"}}
+
+  <ShwTextH1>CopyButton</ShwTextH1>
```
dchyun, line 16:
```suggestion
  <ShwTextH1>Copy::Button</ShwTextH1>
```
[Nit] Previously the `::` were included in the title

showcase/app/components/page-components/copy/button/sub-sections/demos.gts
```diff
@@ -0,0 +1,460 @@
...
+            @text="Copy the textarea value"
+            @targetToCopy="#test-textarea"
+          />
+        </div>
+      </SG.Item>
+      <SG.Item @label="Textarea component (readonly)">
+        <div class="shw-component-copy-button-flex-container">
+          <HdsFormTextareaBase
+            readonly
+            @value="This is a redonly
```
dchyun:
```suggestion
            @value="This is a readonly
```
[Nit] typo

showcase/app/components/page-components/copy/button/sub-sections/base-elements.gts
```diff
@@ -0,0 +1,120 @@
...
+import ShwTextH2 from 'showcase/components/shw/text/h2';
+
+import { HdsCopyButton } from '@hashicorp/design-system-components/components';
+import {
+  SIZES,
+  SUCCESS_ICON,
+  ERROR_ICON,
+} from '@hashicorp/design-system-components/components/hds/copy/button/index';
+
+const STATES = ['default', 'hover', 'active', 'focus'];
```
dchyun, line 21:
```suggestion
const STATES = ['default', 'active', 'hover', 'focus'];
```
[Question] I forget where we landed on aligning the order of states, but for the recent accordion conversion the order is slightly different from what it is here. Should we align to the accordion order?

showcase/app/components/page-components/copy/button/index.gts
shleewhite: I've been thinking about this. I think we should move to remove the `::` in strings in the showcase because it is not inline with the new syntax.

showcase/app/components/page-components/copy/button/sub-sections/base-elements.gts
shleewhite: I think we should leave it as is, we have another ticket to address these inconsistencies in the future.

showcase/app/components/page-components/copy/button/code-fragments/with-masked-input.gts
didoo: > its best practice

If you refer to our HDS components, yes but that's because they're an explicit API to provide escape hatches to our consumers (we don't know upfront what they may need). In our case we're both "producer" and "consumers" of these components, so there's no need to pre-emptively add `...attributes` everywhere.

See also answer here: https://github.com/hashicorp/design-system/pull/3155#discussion_r2339417072

PR #3150, opened 2025-09-07
`Table`, `KeyValueInputs`: adjust types so yielded data is typed

aklkv:
### :pushpin: Summary

We have been experiencing issues where `@model` shape is not passed to `B.data`, which resulted in a lot of glint exceptions or we need to create custom type casting function. 

If merged, this PR would use generic types so the type of the model or data is preserved when yielded back to the consumer.

***

<details>

</details>

### DIFF
--- a/.changeset/loud-readers-sneeze.md
+++ b/.changeset/loud-readers-sneeze.md
@@ -1,0 +1,11 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+<!-- START components/table/table -->
+`Table` - Update the model and returned `B.data` to use generic types, so the type of the data is retained.
+<!-- END -->
+
+<!-- START components/form/key-value-inputs -->
+`KeyValueInputs` - Update the data and returned `R.rowData` to use generic types, so the type of the data is retained.
+<!-- END -->


showcase/app/templates/page-components/form/key-value-inputs/index.hbs
```diff
@@ -212,10 +207,9 @@
 
     <:row as |R|>
       {{! @glint-expect-error - https://hashicorp.atlassian.net/browse/HDS-5090 }}
```
shleewhite: Why does this one still have a glint expect error?

showcase/app/templates/page-components/rich-tooltip.hbs
```diff
@@ -898,7 +896,6 @@
               <Hds::RichTooltip as |RT|>
                 <RT.Toggle>
                   <Hds::Badge
-                    {{! @glint-expect-error - https://hashicorp.atlassian.net/browse/HDS-5090 }}
                     @text={{capitalize B.data.status.text}}
                     {{! @glint-expect-error - https://hashicorp.atlassian.net/browse/HDS-5090 }}
```
shleewhite: Same here?

packages/components/src/components/hds/table/index.ts
```diff
@@ -90,7 +90,9 @@ export interface HdsTableSignature {
   Element: HTMLTableElement;
 }
 
-export default class HdsTable extends Component<HdsTableSignature> {
+export default class HdsTable<T = HdsTableModel> extends Component<
```
shleewhite, line 93: What is HdsTableModel here?

shleewhite (commented):
Could this fix also be applied to the SuperSelect and AdvancedTable components?

packages/components/src/components/hds/table/index.ts
aklkv: it was there previously I just re used it https://github.com/hashicorp/design-system/blob/main/packages/components/src/components/hds/table/types.ts#L95 

```
export type HdsTableModel = Array<Record<string, unknown>>;
```

showcase/app/templates/page-components/form/key-value-inputs/index.hbs
aklkv: I think it was the issues with data that was provided:

```
@isInvalid={{R.rowData.error}}
```

`error` - doesn't exist

showcase/app/templates/page-components/rich-tooltip.hbs
aklkv: `Hds::Badge` has very stricted enum expectations that need to be loosened we have many many issues on consumer side with that and also `Hds::Icon`

aklkv:
> Could this fix also be applied to the SuperSelect and AdvancedTable components?

yeah it could, but it is more involved and `AdvancedTable` has some bad patterns that are not ideal to how TS works, so I excluded it from this basic implementation, I think is this a good first step and can be propogated to other components that need it but since I do not know exactly some components should work I decided to avoid going to deep on them as implementation would have been a lot more complex

packages/components/src/components/hds/table/index.ts
shleewhite: Ahhh as the default. I see.

showcase/app/templates/page-components/form/key-value-inputs/index.hbs
```diff
@@ -206,8 +206,7 @@
     </:header>
 
     <:row as |R|>
-      {{! @glint-expect-error - https://hashicorp.atlassian.net/browse/HDS-5090 }}
-      <R.Field @isRequired={{true}} @isInvalid={{R.rowData.error}} as |F|>
```
shleewhite: It doesn't exist and we never show an error here so removed `@isInvalid`

showcase/app/templates/page-components/form/key-value-inputs/index.hbs
```diff
@@ -247,8 +246,7 @@
...
```
shleewhite: Same with this instance of `@isInvalid`

showcase/app/templates/page-components/table.hbs
```diff
@@ -805,8 +805,7 @@
         <B.Tr
           @selectionKey="{{B.data.id}}"
           @isSelected={{B.data.isSelected}}
-          {{! @glint-expect-error - will be fixed by https://hashicorp.atlassian.net/browse/HDS-5090}}
```
shleewhite, line 878: `B.data.lorem` doesn't exist, so just point it to the id instead.

showcase/app/templates/page-components/table.hbs
```diff
@@ -1039,14 +1037,14 @@
...
+            {{#if B.data.icon}}
+              <Hds::Icon @name={{B.data.icon}} @isInline={{true}} />
+            {{/if}}
             {{B.data.album}}
           </div>
         </B.Td>
         <B.Td>
-          {{! @glint-expect-error - will be fixed by https://hashicorp.atlassian.net/browse/HDS-5090}}
-          <Hds::Badge @text={{B.data.year}} @type="outlined" @color={{B.data.badge-color}} />
+          <Hds::Badge @text="{{B.data.year}}" @type="outlined" @color={{B.data.badge-color}} />
```
shleewhite: The issue with this one was that the year is typed as a number, just made it a string to resolve the issue.

showcase/app/templates/page-components/rich-tooltip.hbs
```diff
@@ -914,8 +913,7 @@
             </B.Td>
             <B.Td>
               <Hds::RichTooltip as |RT|>
-                {{! @glint-expect-error - https://hashicorp.atlassian.net/browse/HDS-5090 }}
-                <RT.Toggle @text={{B.data.tagsCount}} @icon="info" @iconPosition="leading" />
+                <RT.Toggle @text="{{B.data.tagsCount}}" @icon="info" @iconPosition="leading" />
```
shleewhite, line 916: `B.data.tagsCount` is a number in the default data, so just made it a string to resolve the issue.

showcase/app/templates/page-components/table.hbs
aklkv: I have raised this issue before we should, we often do things like `array.length` directly to this argument which results in this error, I think Badge should allow string or number

showcase/app/templates/page-components/rich-tooltip.hbs
aklkv: same as above, probably

shleewhite: Ok closing out this thread, this one is different because the rich tooltip requires a string (badge allows string or number). I do actually need to make it a string here

showcase/app/templates/page-components/table.hbs
shleewhite: Closing out this thread too, I removed forcing it to be a string. The issue was that it was being returned as a number or undefined. 

I just added a check to make sure it is defined before it is used in the Badge. Badge already accepts either a string or number for the text argument.

PR #3174, opened 2025-09-12
Clean up document changes to prevent test interference

alex-ju:
### :pushpin: Summary

`ember try:one ember-beta` seems to [struggle](https://github.com/hashicorp/design-system/actions/runs/17657733841/job/50185194464) to keep up with global, document-level changes (such as operations on `<body>`) so I suggest we reset them after each test to prevent interference.

This is currently polluting our automated checks with 'CI - Components / ember-beta (push)' failing.

### :link: External links

[Slack thread](https://hashicorp.slack.com/archives/C025N5V4PFZ/p1756848130224869)

***

<details>

</details>

### DIFF
--- a/showcase/tests/integration/components/hds/app-header/index-test.js
+++ b/showcase/tests/integration/components/hds/app-header/index-test.js
@@ -211,6 +211,8 @@ module('Integration | Component | hds/app-header/index', function (hooks) {
         .getAttribute('aria-controls'),
       this.element.querySelector('.hds-app-header__actions').getAttribute('id'),
     );
+    // Toggle the menu back to close to avoid interfering with other tests
+    await click('.hds-app-header__menu-button');
   });
 
   // A11Y Refocus


showcase/tests/integration/components/hds/app-header/index-test.js
```diff
@@ -211,6 +211,8 @@ module('Integration | Component | hds/app-header/index', function (hooks) {
         .getAttribute('aria-controls'),
       this.element.querySelector('.hds-app-header__actions').getAttribute('id'),
     );
+    // Toggle the menu back to close to avoid interfering with other tests
+    await click('.hds-app-header__menu-button');
```
alex-ju, line 215: Noting here that this action should not be required. We could make better use of destructors to make sure the document remains intact. I'll create a task for us to look into this.

shleewhite: Are the dropdown tests doing something so that this cleanup is not required there too?

PR #3189, opened 2025-09-15
`PopoverPrimitive` - Allow dynamic toggle element

zamoore:
### :pushpin: Summary

If merged, this PR will fix a bug where the `Hds::PopoverPrimitive` would stop working if its toggle element was dynamically swapped, ensuring the popover remains correctly associated and positioned.

The previous implementation only ran the logic to link the toggle and popover (`popovertarget`, `aria-controls`, and positioning) once when the component was first rendered. This caused a bug when a consumer dynamically changed the toggle element, as the new element was never linked to the popover.

**Example of Broken Behavior**

In the code below, clicking the "Original" button worked, but after a state change, the "New" button would not trigger the popover.

```hbs
<Hds::PopoverPrimitive @enableClickEvents={{true}} as |PP|>
  {{#if this.isSwapped}}
    <button {{PP.setupPrimitiveToggle}}>New Toggle</button>
  {{else}}
    <button {{PP.setupPrimitiveToggle}}>Original Toggle</button>
  {{/if}}
  <div {{PP.setupPrimitivePopover}}>Content</div>
</Hds::PopoverPrimitive>
```

### :link: External links

Jira ticket: [HDS-4701](https://hashicorp.atlassian.net/browse/HDS-4701)

***

<details>

</details>


packages/components/src/components/hds/popover-primitive/index.ts
```diff
@@ -106,13 +107,23 @@ export default class HdsPopoverPrimitive extends Component<HdsPopoverPrimitiveSi
...
       assert(
         `The toggle element of "Hds::PopoverPrimitive" must be a <button>; element received: <${element.tagName.toLowerCase()}>`,
         element instanceof HTMLButtonElement
       );
+
+      this._linkToggleAndPopover();
+
+      // Return a teardown function to clean up the modifier's side effects.
+      // This is a safeguard against bugs where this element might be
+      // cached and re-parented in the DOM, rather than being fully destroyed.
```
didoo, line 122: [praise] 👏 👏 👏 👏 👏 👏

showcase/tests/integration/components/hds/popover-primitive/index-test.js
```diff
@@ -103,6 +103,72 @@ module(
       // should go back to hidden
       assert.dom('#test-popover-primitive-content').isNotVisible();
     });
+    test('it should continue to work when the toggle element is dynamically swapped', async function (assert) {
```
didoo, line 106: [praise] 🧑‍🍳👌

.changeset/brave-owls-jump.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+<!-- START components/popover-primitive -->
+`PopoverPrimitive` - Added support for dynamic toggle elements.
```
didoo: Maybe something like this?
```suggestion
`PopoverPrimitive` - Added support for dynamic swap/injection of the toggle element.
```

packages/components/src/components/hds/popover-primitive/index.ts
```diff
@@ -167,27 +163,58 @@ export default class HdsPopoverPrimitive extends Component<HdsPopoverPrimitiveSi
...
+    }
+  );
+
+  // Apply the `hds-anchored-position` modifier to the "popover" element
+  // (notice: this function runs the first time when the element the modifier was applied to is inserted into the DOM, and it autotracks while running.
+  // Any tracked values that it accesses will be tracked, including the arguments it receives, and if any of them changes, the function will run again)
+  // This modifiers uses the Floating UI library to provide:
+  // - positioning of the "popover" in relation to the "toggle"
+  // - collision detection (optional)
+  private _updatePositioning(): void {
```
didoo: [suggestion] `_updatePositioning` reminds me of the action that the Floating UI library is performing, while here we're actually applying the modifier to the popover element
```suggestion
  private _applyAnchoredPositionModifier(): void {
```

didoo (approved):
Great PR! 🙌

I have left only a few suggestions, but nothing is a blocker.

alex-ju:
I followed the logic and I can see why this implementation is flagged as minor, which is great. Should we plan to test it with a release candidate though?

.changeset/brave-owls-jump.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+<!-- START components/popover-primitive -->
```
dchyun:
```suggestion
<!-- START utilities/popover-primitive -->
```
[Nit] Use correct docs path

shleewhite (commented):
Could you add an example of the toggle switching to the showcase?

didoo:
> Could you add an example of the toggle switching to the showcase?

[premise: no strong opinions] I thought the same at the beginning, but then I thought that maybe in this case it's not necessary, there is nothing UI related that a visual regression tests or a human test could catch, that the integration tests does not already catch

zamoore:
> I followed the logic and I can see why this implementation is flagged as minor, which is great. Should we plan to test it with a release candidate though?

Independently of the next release? We can, but I don't think it's necessary.

zamoore:
> > Could you add an example of the toggle switching to the showcase?
> 
> [premise: no strong opinions] I thought the same at the beginning, but then I thought that maybe in this case it's not necessary, there is nothing UI related that a visual regression tests or a human test could catch, that the integration tests does not already catch

@shleewhite @didoo 

This was my thinking as well. I can add the examples in the showcase if anyone has a strong opinion, but this more covers an edge case than provides additional intentional behavior.

shleewhite:
@zamoore @didoo ehh im convinced- we can leave it out for now.

didoo (approved):

PR #3205, opened 2025-09-18
`ApplicationState` - Remove center alignment rule

zamoore:
### :pushpin: Summary

If merged, this PR removed a style rule that center-aligns the `ApplicationState` component.

The current implementation of the `ApplicationState` component has a `margin: 0 auto;` rule on the root element (`.hds-application-state`). This forces a center alignment of the component, which consumers need to override.

### :link: External links

Jira ticket: [HDS-3584](https://hashicorp.atlassian.net/browse/HDS-3584)

***

<details>

</details>

### DIFF
--- a/packages/components/src/styles/components/application-state.scss
+++ b/packages/components/src/styles/components/application-state.scss
@@ -14,7 +14,6 @@ $hds-application-state-content-max-width: 480px;
   flex-direction: column;
   align-items: start;
   width: fit-content;
-  margin: 0 auto; // this will center the component in the parent container
 
   &.hds-application-state--align-center {
     align-items: center;
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+<!-- TEMPORARY ADDITION SO BRANCH DIVERGES FROM MAIN -->
+
 <picture>
   <source media="(prefers-color-scheme: dark)" srcset="https://github.com/hashicorp/design-system/assets/788096/5d6969b7-f8b4-4ad3-9ece-b16b4527522e" width="300">
   <source media="(prefers-color-scheme: light)" srcset="https://github.com/hashicorp/design-system/assets/788096/8e278094-427f-40cc-912f-9ccd3a0ff879" width="300">


zamoore:
> Now that we have removed the self-alignment, probably we should update the showcase to have two variants (default + center-aligned) for some of the cases (eg. when it's in a container) because I _think_ in most of the cases, the empty state will be visually centered (in a page, or in a table, or in a card). As far as I remember, the request came from Atlas/TFC and was an exception.
> 
> Which leads me to this consideration: what if introducing this change as is now (which is definitely breaking, will impact a lot of places, and can't be codemoded but will require a lot of manual checks) we define a new argument - `@hasMarginAuto ` or `@autoCenter` or or something similar - that by default is `true` (so non-breaking) but that consumers can turn off if/when needed?

I really like the idea of introducing a new property. `autoCenter` sounds good to me. This is similar to what we did for `isInline` in the icon component, so it has precedent.

.changeset/hot-owls-rescue.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": major
+---
+
+<!-- START components/table/advanced-table -->
```
didoo, line 5: @zamoore should this PR be rebased?

packages/components/src/styles/components/application-state.scss
```diff
@@ -26,6 +25,10 @@ $hds-application-state-content-max-width: 480px;
       width: auto;
     }
   }
+
+  &.hds-application-state--is-auto-centered {
```
didoo, line 30:
```suggestion
  // we want that by default the component is centered in the parent container
  &.hds-application-state--is-auto-centered {
```

.changeset/easy-owls-create.md
```diff
@@ -0,0 +1,10 @@
+---
+"@hashicorp/design-system-components": major
+---
+
+<!-- START components/application-state -->
+
+`ApplicationState` - Removed the opinionated `margin: 0 auto;` rule from the component's root element to delegate horizontal alignment control to the consumer.
+
+- Added the new `@isAutoCentered` argument (which defaults to `true`) to preserve the existing centering behavior while allowing consumers to disable it when needed.
```
didoo:
```suggestion
`ApplicationState` - Replaced the default opinionated `margin: 0 auto;` rule from the component's root element and replaced it with a new `@isAutoCentered` argument (which defaults to `true, to preserve the existing centering behavior) to delegate the horizontal alignment control to the consumers, allowing them to disable it when needed.
```

showcase/tests/integration/components/hds/application-state/index-test.js
```diff
@@ -63,6 +63,26 @@ module(
         .hasClass('hds-application-state--align-center');
     });
 
+    test('it should have the correct class when isAutoCentered is true', async function (assert) {
```
didoo: [thought] I am not convinced of this test: usually the way we check things like this is via two different tests:
- one test that checks the default behavior (without setting any argument value, relying on its default)
- one test that checks the alternative behaviour, setting the argument that controls the behaviour

didoo (commented):
left a couple of comments/suggestions

showcase/tests/integration/components/hds/application-state/index-test.js
zamoore: I think this test takes it a step further by showing that the class is reactive to the argument. Do you think it falls short in some way?

didoo: mmm, do you see a use case when consumer would dynamically toggle this argument? not blocking things on this, but I would keep things simple and just test default + argument with non-default value

zamoore: Fair enough. I split it up into 2 tests.

didoo (approved):
Approving, but there is a test on the website that is failing.

PR #3207, opened 2025-09-18
StepperList: Type fix for STATUSES export

dchyun:
### :pushpin: Summary

If merged, this PR would update the type of the `STATUSES` export in the StepperList::Step to fix linting issues.

Originally this PR was to update an incorrect import path in the showcase pages of the StepperList. Upon fixing this path, however a new linting error was thrown in the showcase. 

Screenshot 2025-09-18 at 3 22 45 PM

Upon further investigation, this linting error is due to the type of the `STATUES` export of the StepperList. The type of `STATUSES` is set to `string[]`, but the `@status` argument of the component is typed as the component specific type `HdsStepperStatuses`. When using `STATUSES` in the showcase examples this then causes a type mismatch. 

```
 {{#each STATUSES as |status|}}
      <SG.Item @label="{{capitalize status}}">
        <HdsStepperList @ariaLabel="Label" as |S|>
          <S.Step @status={{status}}>
          ...
```

To fix this issue, the type of the export `STATUSES` was updated from `string[]` to `HdsStepperStatuses[]`

```
export enum HdsStepperStatusesValues {
  Incomplete = 'incomplete',
  Progress = 'progress',
  Processing = 'processing',
  Complete = 'complete',
}

export type HdsStepperStatuses = `${HdsStepperStatusesValues}`;
```

### Potential follow-ups

In the component library, the majority of exported arrays of possible values for an argument are typed as the custom type for that component. However, there are various exports that follow this same pattern of typing the export as `string[]`. These exports will most likely run into the same type issues as those found here.

An effort could be taken to update these types to follow the changes made here.

Example from the Separator component
```
export const SPACING: string[] = Object.values(HdsSeparatorSpacingValues);
```

<details>

</details>

### DIFF
--- a/packages/components/src/components/hds/stepper/list/step.ts
+++ b/packages/components/src/components/hds/stepper/list/step.ts
@@ -21,7 +21,9 @@ import type {
 } from '../types.ts';
 
 export const DEFAULT_STATUS = HdsStepperStatusesValues.Incomplete;
-export const STATUSES: string[] = Object.values(HdsStepperStatusesValues);
+export const STATUSES: HdsStepperStatuses[] = Object.values(
+  HdsStepperStatusesValues
+);
 export const MAPPING_STATUS_TO_SR_ONLY_TEXT = HdsStepperStatusToSrOnlyText;
 
 export interface HdsStepperListStepSignature {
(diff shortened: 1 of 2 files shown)


dchyun:
> Should this have a changelog?

@didoo I think we should be okay without one. There were similar type fixes made recently to exports in https://github.com/hashicorp/design-system/pull/3054 and we didn't add a changeset there.

PR #3215, opened 2025-09-19
feat: improve logic for modal/flyout lifecycle handling

shleewhite:
### :pushpin: Summary

If merged, this PR would use `registerDestructor` to trigger the close callback when a Modal or Flyout is closed to ensure the proper cleanup happens when a Modal/Flyout is removed from the DOM.

**TODO:**
- [ ] test in cloud ui

This is a follow up from the Modal incident that happened earlier this summer when the modal is simply destroyed and removed from the DOM (eg. on submit) the onDismiss is not run, just the willDestroyNode method, which leaves the overflow applied to the body, making the page unscrollable.

This solution was previously suggested in https://github.com/hashicorp/design-system/pull/2962, but was not used because it was safer to just revert the change that caused the bug.

### :link: External links

Jira ticket: [HDS-5038](https://hashicorp.atlassian.net/browse/HDS-5038)

***

<details>

</details>

### DIFF
--- a/packages/components/src/components/hds/flyout/index.ts
+++ b/packages/components/src/components/hds/flyout/index.ts
@@ -10,6 +10,9 @@ import { assert } from '@ember/debug';
 import { getElementId } from '../../../utils/hds-get-element-id.ts';
 import { buildWaiter } from '@ember/test-waiters';
 import type { WithBoundArgs } from '@glint/template';
+import { modifier } from 'ember-modifier';
+import { registerDestructor } from '@ember/destroyable';
+import type Owner from '@ember/owner';
 
 import type { HdsFlyoutSizes } from './types.ts';
 
@@ -64,6 +67,27 @@ export default class HdsFlyout extends Component<HdsFlyoutSignature> {
   _element!: HTMLDialogElement;
   private _body!: HTMLElement;
   private _bodyInitialOverflowValue = '';
+  private _clickHandler!: (event: MouseEvent) => void;
+
+  constructor(owner: Owner, args: HdsFlyoutSignature['Args']) {
+    super(owner, args);
+
+    registerDestructor(this, (): void => {
+      // if the <dialog> is removed from the dom while open we emulate the close event
+      if (this._element && this._isOpen) {
+        this._element.dispatchEvent(new Event('close'));
+
+        this._element.removeEventListener(
+          'close',
+          // eslint-disable-next-line @typescript-eslint/unbound-method
+          this.registerOnCloseCallback,
+          true
+        );
+      }
+
+      document.removeEventListener('click', this._clickHandler, true);
+    });
+  }
 
   /**
    * Sets the size of the flyout
@@ -115,8 +139,7 @@ export default class HdsFlyout extends Component<HdsFlyoutSignature> {
     this._isOpen = false;
   }
 
-  @action
-  didInsert(element: HTMLDialogElement): void {
+  private _registerDialog = modifier((element: HTMLDialogElement) => {
     // Store references of `<dialog>` and `<body>` elements
     this._element = element;
     this._body = document.body;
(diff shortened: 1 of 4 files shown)


packages/components/src/components/hds/flyout/index.ts
```diff
@@ -135,19 +136,35 @@ export default class HdsFlyout extends Component<HdsFlyoutSignature> {
...
-        // eslint-disable-next-line @typescript-eslint/unbound-method
-        this.registerOnCloseCallback,
-        true
-      );
-    }
-  }
+    this._clickHandler = (event: MouseEvent) => {
+      // check if the click is outside the flyout and the flyout is open
+      if (!this._element.contains(event.target as Node) && this._isOpen) {
+        void this.onDismiss();
```
didoo:
```suggestion
        //  here we use `void` because `onDismiss` is an async function, but in reality we don't need to handle the result or wait for its completion
        void this.onDismiss();
```

packages/components/src/components/hds/modal/index.ts
```diff
@@ -151,19 +175,37 @@ export default class HdsModal extends Component<HdsModalSignature> {
...
+      if (this._element && this._isOpen) {
+        this._element.dispatchEvent(new Event('close'));
+
+        this._element.removeEventListener(
+          'close',
+          // eslint-disable-next-line @typescript-eslint/unbound-method
+          this.registerOnCloseCallback,
+          true
+        );
+      }
```
didoo: [thought] I have a feeling that this should be something like this (see comment in the main review message)

```suggestion
      if (this._element) {
        if (this._isOpen) {
          this._element.dispatchEvent(new Event('close'));
        }

        this._element.removeEventListener(
          'close',
          // eslint-disable-next-line @typescript-eslint/unbound-method
          this.registerOnCloseCallback,
          true
        );
      }
```

didoo (commented):
A few high-level comments:

---

All these explanations need to be updated to reflect the new code logic (`willDestroyNode` doesn't exist anymore)

Image
Image
Image

The same on the Modal showcase demos.

---

I suspect there is still something not 100% working at logical level, see video below where the listener for the `close` event seems not to be removed (that's my understanding, but I may be wrong).

https://github.com/user-attachments/assets/05219440-cbb5-4471-9019-b8d91a6944ae

I suggest to review all the use cases using a reduced set of examples, and making sure (via `debugger` that the order of execution is what you would expect, for different ways to close the flyout). I've sent you a code patch in DM if you want to use it.

shleewhite:
@didoo I looked into it more and I believe the fix you suggested was enough to make sure all the event listeners are removed correctly.

packages/components/src/components/hds/modal/index.ts
```diff
@@ -128,11 +130,33 @@ export default class HdsModal extends Component<HdsModalSignature> {
       }
     } else {
       this._isOpen = false;
+
+      // Reset page `overflow` property
+      if (this._body) {
```
didoo, line 135: you have made this change for the `modal`, but not the flyout (it's still in the `onDismiss` method). was it intentional? are there reasons for this different logic? maybe because the `dismiss` can be disabled in the modal? in which case maybe better to add a comment in both files to explain the differences

showcase/app/templates/page-components/modal.hbs
dchyun, line 1: [Note] One difference I was seeing between the existing showcase and this branch is the number of document click event listeners present on a demo modal. This is just due to all of the "open" modal examples above the demo section. I commented those out locally and only one click event listener was present per demo modal as expected.

**Prod**
Image

**This branch**
Image

dchyun (commented):
Minus the one open question on the differences between the modal and the flyout this looks all good to me!

packages/components/src/components/hds/modal/index.ts
```diff
@@ -151,19 +175,43 @@ export default class HdsModal extends Component<HdsModalSignature> {
...
       this.open();
     }
-  }
 
-  @action
-  willDestroyNode(): void {
-    if (this._element) {
-      this._element.removeEventListener(
+    // Note: because the Modal has the `@isDismissedDisabled` argument, we need to add our own click outside to dismiss logic. This is because `ember-focus-trap` treats the `focusTrapOptions` as static, so we can't update it dynamically if `@isDismissDisabled` changes.
+    this._clickOutsideToDismissHandler = (event: MouseEvent) => {
```
dchyun, line 180: [todo] Can we start rerunning this test for clicking outside the modal in the test suite now?
```
// TODO! while we decide what to do about the original bug
  skip('it should close the modal when clicking outside', async function (assert) {
```

didoo (approved):
LGTM 👌

packages/components/src/components/hds/modal/index.ts
shleewhite: Good point! Re-enabled the test.

dchyun (approved):

didoo (approved):

PR #2950, opened 2025-06-13
Upgrades ember-a11y-refocus to v5

MelSumner:
### :pushpin: Summary

If merged, this PR updates `ember-a11y-refocus` to v5.0.0.

***

### :eyes: Component checklist

- [ ] Percy was checked for any visual regression
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

<details>

<summary>:clipboard: PCI review checklist</summary>

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.
- [ ] If applicable, I've worked with GRC to document the impact of any changes to security controls.
  Examples of changes to controls include access controls, encryption, logging, etc.
- [ ] If applicable, I've worked with GRC to ensure compliance due to a significant change to the in-scope PCI environment.
  Examples include changes to operating systems, ports, protocols, services, cryptography-related components, PII processing code, etc.

</details>

PR #3234, opened 2025-09-24
Upgrade `ember-a11y-refocus` to `v5.1.0`

shleewhite:
### :pushpin: Summary

This PR upgrades `ember-a11y-refocus` to `v5.1.0`, which required a few related dependency, type, and style updates to ensure everything compiles and functions correctly.

#### Changes
* **Upgrade `@glimmer/component` to `v2.0`**
  * Required for the `EmberA11yRefocusRegistry` to correctly import type signatures for `NavigationNarrator`.
  * Without this, `NavigationNarrator` would remain untyped.
  * Added a global override in `package.json` to force `v2.0.0`. This prevents build failures in Showcase and Website caused by mismatched versions.
* **Fix types for `@a11yRefocusRouteChangeValidator` arguments**
  * In `SideNav` and `AppHeader`, this arg was previously typed as `string`.
  * It now correctly uses `(transition: Transition) => boolean` since it is forwarded to the `NavigationNarrator` component.
* **Correct `Owner` typing in `th-resize-handle`**
  * `@glimmer/component@2.0.0` throws if `Owner` is not typed correctly.
  * Updated the last remaining usage of `Owner` being typed as `unknown`.
* **Add `NavigationNarrator` styles to Website CSS**
  * `ember-a11y-refocus@5.1.0` no longer auto-imports its styles for the website nav.
  * Explicitly added them via `website/ember-cli-build.js` and `website/app/styles/app.scss` [following the recommended usage](https://github.com/ember-a11y/ember-a11y-refocus?tab=readme-ov-file#usage).

### Open discussion

There is still a question around what is the best way to handle the `ember-a11y-refocus` styles. Before, the way the build worked made it so the consumer would install `ember-a11y-refocus` by installing the design system and that would implicitly import the styles. This is no longer the case, so either we need to export the styles for them or they need to manually import them.

In this instance, we can treat the showcase app as a demo consumer of the design system. The easiest way to verify that the styles are imported are to check [the SideNav skip link example in the showcase](https://hds-showcase-git-chore-upgrade-ember-a11y-focus-hashicorp.vercel.app/components/side-nav#skip-link-a11y-refocus). If the link has an underline, the styles are not imported.

_**The solution in this PR is Lee's preference**_
Ultimately, the main style we need is `text-decoration: none` and ember-a11y-refocus is not likely to change the styles frequently so I think it is reasonably safe to copy their styles and include them in our exported stylesheet.

**Option 2: Require all consumers add `ember-a11y-refocus` as a dependency**

SPIKE PR: https://github.com/hashicorp/design-system/pull/3251

In this option, we don't export the ember-a11y-refocus styles so the consumer has to include the path in their ember-cli-build and add it as a dependency. This approach is similar to how the tokens styles work.

This would require all consumers to update their ember-cli-build file and we'd have to change the guidance on our website. If you don't do this, the design system package builds fine, but the showcase fails to build because it does not have access to the stylesheet.
***

<details>

</details>

### DIFF
--- a/.changeset/khaki-bananas-know.md
+++ b/.changeset/khaki-bananas-know.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Dependency ember-a11y-refocus upgraded
--- a/.changeset/khaki-bananas-know.md
+++ b/.changeset/khaki-bananas-know.md
@@ -3,9 +3,9 @@
 ---
 
 <!-- START components/side-nav -->
-`SideNav` - Fixed the type of `@a11yRefocusNavigationText` to match the expected type from `ember-a11y-refocus`. The new type is `(transition: Transition) => boolean` instead of `string`.
+`SideNav` - Fixed the type of `@a11yRefocusRouteChangeValidator` to match the expected type from `ember-a11y-refocus`. The new type is `(transition: Transition) => boolean` instead of `string`.
 <!-- END -->
 
 <!-- START components/app-header -->
-`AppHeader` - Fixed the type of `@a11yRefocusNavigationText` to match the expected type from `ember-a11y-refocus`. The new type is `(transition: Transition) => boolean` instead of `string`.
+`AppHeader` - Fixed the type of `@a11yRefocusRouteChangeValidator` to match the expected type from `ember-a11y-refocus`. The new type is `(transition: Transition) => boolean` instead of `string`.
 <!-- END -->
--- a/.changeset/khaki-bananas-know.md
+++ b/.changeset/khaki-bananas-know.md
@@ -2,8 +2,6 @@
 "@hashicorp/design-system-components": patch
 ---
 
-Dependencies `ember-a11y-refocus` and `@glimmer/component` upgraded.
-
 <!-- START components/side-nav -->
 `SideNav` - Fixed the type of `@a11yRefocusNavigationText` to match the expected type from `ember-a11y-refocus`. The new type is `(transition: Transition) => boolean` instead of `string`.
 <!-- END -->
--- a/.changeset/khaki-bananas-know.md
+++ b/.changeset/khaki-bananas-know.md
@@ -2,4 +2,12 @@
 "@hashicorp/design-system-components": patch
 ---
 
-Dependency ember-a11y-refocus upgraded
+Dependency `ember-a11y-refocus` and `@glimmer/component` upgraded.
+
+<!-- START components/side-nav -->
+`SideNav` - Fixed the type of `@a11yRefocusNavigationText` to match the expected type from `ember-a11y-refocus`. The new type is `(transition: Transition) => boolean` instead of `string`.
+<!-- END -->
+
+<!-- START components/app-header -->
+`AppHeader` - Fixed the type of `@a11yRefocusNavigationText` to match the expected type from `ember-a11y-refocus`. The new type is `(transition: Transition) => boolean` instead of `string`.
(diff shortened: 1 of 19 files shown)


packages/components/package.json
dchyun, line 1: [Suggestion] Looks like in the Percy diff, the skip to main link in the SideNav has an underline now. Previously the `ember-a11y-refocus-skip-link` class brought styles that set `text-decoration: none;`, but in the latest version those styles are no longer present.

We may need to adjust the `.hds-side-nav .ember-a11y-refocus-skip-link` styles, or any other skip link styles we have, to add `text-decoration: none`.

Previous styles:
Screenshot 2025-09-24 at 3 27 21 PM

aklkv:
we need to bring it styles manually now as addon it self now v2 which means styles will not be auto included https://github.com/ember-a11y/ember-a11y-refocus?tab=readme-ov-file#usage

also my comments on original PR is still valid https://github.com/hashicorp/design-system/pull/2950#issuecomment-2971874474

.changeset/khaki-bananas-know.md
```diff
@@ -0,0 +1,13 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+Dependencies `ember-a11y-refocus` and `@glimmer/component` upgraded.
```
didoo: [suggestion] I would probably split this in two entries, and add which version (at least the `major.minor`)

shleewhite: makes sense! I made the upgrades major and the type fixes as a patch (I checked and it doesnt seem like anyone is using those args)

didoo: This is what I meant :)
```suggestion
Upgraded `ember-a11y-refocus` dependency to version `5.0`
Upgraded `@glimmer/component` dependency to version `2.0`
```

website/app/app.js
```diff
@@ -8,6 +8,7 @@ import Resolver from 'ember-resolver';
 import loadInitializers from 'ember-load-initializers';
 import config from 'website/config/environment';
 import { importSync, isDevelopingApp, macroCondition } from '@embroider/macros';
+import 'ember-a11y-refocus/styles.css';
```
aklkv: this will not work in classic build, you need to move it into scss file, this method only work under Vite build

packages/components/rollup.config.mjs
```diff
@@ -61,6 +61,11 @@ const plugins = [
     fileName: 'styles/@hashicorp/design-system-power-select-overrides.css',
   }),
 
+  scss({
```
aklkv: this will not work nor will the line above

aklkv (approved):

MelSumner:
The addon has classes & Ember should let you add your own class when you invoke the component, so the styles shouldn't be an issue, even if usage has slightly changed out of necessity. 

I think your preferred solution is fine, honestly. I don't plan to change the styles, it's one of those things that is meant to be as barebones as possible so folks can use the classnames in their apps if they want to change things to match their own visual presentation styles.


> Fix types for `@a11yRefocusNavigationText` arguments

the navigation text should still be a string. Do you mean `a11yRefocusRouteChangeValidator` maybe?

shleewhite:
> > Fix types for `@a11yRefocusNavigationText` arguments
> 
> the navigation text should still be a string. Do you mean `a11yRefocusRouteChangeValidator` maybe?

Oop yes, mixed up the argument name + type into the wrong argument name haha

zamoore (approved):
Nice!

PR #3239, opened 2025-09-25
[WIP] [Project solaris / Phase 1 / Tokens Pipeline] Add `$modes` support to tokens pipeline

didoo:
### :pushpin: Summary

This is the implementation PR for adding support to co-located `$modes` values in the design tokens JSON files, the processing of such values, and the generation of distinct "themed" CSS files in output, with different formats.

This is a WIP built on top of 

### :link: External links

Figma file: [if it applies]

***

<details>

</details>


packages/tokens/dist/docs/products/tokens.json
KristinLBradley, line 1: These updates to the 2 JSON files seem to make sense. They are just adding the "mode" or "theme" values for each type of color token, correct? That seems fairly straight forward in and of itself to me.

packages/tokens/scripts/build-parts/generateThemingCssFiles.ts
KristinLBradley, line 1: My first question would be if there's perhaps a better/clearer name for this file than "generateExtraThemingFiles.ts"?

Any other options you considered? The word "extra" seems a bit ambiguous.

packages/tokens/scripts/build-parts/generateExtraThemingCssFiles.ts
```diff
@@ -0,0 +1,104 @@
...
+  for (const method of methods) {
+
+    let outputContent = `${header}\n\n`;
+
+    // CSS file for `prefers-color-scheme` (note: we use `cds-g0` for `light` and `cds-g100` for `dark`
+    if (method === 'prefers-color-scheme') {
+      outputContent = `${header}\n\n`;
+      //
+      // these are the themed `carbonized` tokens
+      outputContent += `@media (prefers-color-scheme: light) { ${cds0ThemedSource} }\n\n`;
```
KristinLBradley: https://drafts.csswg.org/mediaqueries-5/#prefers-color-scheme

In the W3C specs for the `prefers-color-scheme` media query, it mentions that a value of "light" indicates EITHER that the user expressly set a preference for a light color scheme OR that they did not set a preference and thus the default should be used.

Following this, they suggest using `(not (prefers-color-scheme: dark))` vs. `(prefers-color-scheme: 'light')` to allow for future updates to express a more active preference for a light color scheme or possible other color schemes.

Anyway, it doesn't seem too vital to worry about, but just something I thought we may want to consider.

packages/tokens/scripts/build-parts/generateExtraThemingCssFiles.ts
```diff
@@ -0,0 +1,104 @@
...
+
+import fs from 'fs-extra';
+import prettier from 'prettier';
+
+import type { Dictionary, PlatformConfig }  from 'style-dictionary';
+import { fileHeader } from 'style-dictionary/utils';
+
+export async function generateExtraThemingFiles(_dictionary: Dictionary, config: PlatformConfig ): Promise<void> {
+
+  const commonSource = await getSourceFromFileWithRootSelector(config, 'hds', 'common-tokens.css');
```
dchyun: [Issue] Preface: This may not be relevant if all core and semantic / component level tokens are themed. 

When a core level token has modes, and a semantic level token that references that token does not have modes, there can be some mismatches in the values output into the css files.

Example: `--token-color-blue-200` has modes, but `--token-color-foreground-action`, which is set to `--token-color-blue-200`, does not. 

```
"blue-200": {
  "$type": "color",
  "$value": "#1060ff",
  "group": "palette",
  "$modes": {
    "hds": "#1060ff",
    "cds-g0": "{carbon.color.blue.20}",
    "cds-g10": "{carbon.color.blue.20}",
    "cds-g90": "{carbon.color.blue.20}",
    "cds-g100": "{carbon.color.blue.20}"
  }
},
```

#### with-root-selector

In this situation, for the `with-root-selector` files, `--token-color-blue-200` is correctly added to `themed-tokens.css` with it's correct value.`--token-color-foreground-action` is still added to `common-tokens.css`, but it's value is equal to the themed value for `--token-color-blue-200` because of the token aliasing. 

This leads to a situation where a variable could have different values across the `common-tokens.css` files depending on the value of the themed core token it references.

**`hds/common-tokens.css`**
```
--token-color-foreground-action: #1060ff;
```

**`cds-g10/common-tokens.css`**
```
--token-color-foreground-action: #d0e2ff;
```

#### Extra theming files

For the extra theming files created here using `commonSource`, only the `common-tokens.css` from the `hds` theme is used. This leads to a situation where the semantic tokens is always set to the original blue 200 core token value, not a themed one. 

This all raises a question of if a core level token is themed is the expectation that the semantic level one should always reflect the themed value of the core tokens, or only the HDS value of the core token? 

However, if all of the semantic tokens have their own theming set then this is no longer an issue. So if that is the plan then this is a moot point.

packages/tokens/src/products/shared/typography.json
```diff
@@ -9,6 +9,16 @@
               "-apple-system",
               "BlinkMacSystemFont"
             ],
+            "$modes": {
```
dchyun, line 12: [Note] Overall think this approach is clean and I agree with it. 

My one thought was there could be many cases where all of the carbon tokens are equal for every carbon theme, essentially Carbon's "common tokens". With this approach there could be a lot of duplication of values for tokens across the carbon themes. One alternative could be to have a nesting of the themes (Carbon, HDS), and then modes within those themes (g0, g10, etc.).

Ultimately I think the complexity of that would outweigh the benefits, as HDS only has one theme anyway and there could be plenty of instances of carbon variables differing across themes depending on how the variable mapping goes. This approach is the most flexible to accommodate those kinds of things.

packages/tokens/src/products/shared/typography.json
dchyun, line 1: [Note] When testing adding `$modes` to `font-size` tokens it was successful, but there are some small things to note when theming them.

Our font size tokens are set using px, while Carbon's are set using rems. Both tokens have a `unit` field which specifies the unit of the value provided. 

When the tokens are generated both the px values and rem values are successfully converted to rems. However there are two small places where that conversion does not happen as accurately. 

- In the comment appended to the end of the CSS token, the values listed are the original HDS token values not the themed ones.
- In the token docs json the hds mode will be missing its unit
 ```
"$modes": {
    "hds": "30",
    "cds-g0": "2.625rem",
    "cds-g10": "2.625rem",
    "cds-g90": "2.625rem",
    "cds-g100": "2.625rem"
```

Comment (dchyun) [inline packages/tokens/src/products/shared/typography.json:1]: [Note] I did a test of adding `$modes` to most of the data types in the token library. Below are my preliminary results.

- [x] Color
  - [X] Hex codes
  - [X] Alpha values 
- [x] Typography
  - [x] font family
  - [x] font weight
  - [x] font size 
    - Mostly correct - see comments
  - [x] line height
    - Token comments incorrect
  - [x] letter spacing
- [x] Border radius
  - Carbon has no border radius tokens, but setting them to a hard coded value worked
- [ ] Animation easings
  - Differences in the json format across HDS and Carbon
- [x] Background image / icons
  - Fill / stroke colors will not be able to use theming tokens

File: packages/tokens/dist/docs/products/tokens.json [hunk 4]
```diff
@@ -886,10 +886,24 @@
     "$type": "color",
     "$value": "#ffffff",
     "group": "palette",
+    "$modes": {
```
KristinLBradley, line 889: For the use of the term "modes" in these definitions, you mentioned that it could be confusing to use in this context since it refers to values used for both themes (HDS, Carbon) and modes (light, dark).

Would using a completely different term in this context help?

It's hard to think of a good word but perhaps "variants" or "alternates" or something?

So you could think of it like:

Variants:
* HDS theme, light mode
* HDS theme, dark mode
* CDS-g0 theme, light mode
* CDS-g0 theme, dark mode
* CDS-g10 theme, light mode
* CDS-g10 theme, dark mode

Which to me seems to make sense but could still be confusing since we've used the term "variant" in different contexts before...

Btw, in the context of the JSON, I also think just using the word "$themes" would be clearer to understand the purpose of at a glance than "$modes".

packages/tokens/scripts/build-parts/generateThemingCssFiles.ts
KristinLBradley: Maybe just remove the word "extra" in the file name?

packages/tokens/scripts/build-parts/generateExtraThemingCssFiles.ts
KristinLBradley: This logic/structure is difficult to follow, but my thinking is if we are going to "theme" all the core tokens and the semantic tokens all reference the core tokens (which it seems they should) then it seems that the semantic tokens would all be themed as they should.

packages/tokens/dist/docs/products/tokens.json
KristinLBradley: Maybe it would also be clearer to use the word "value" in the name, such as "$mode-values" or "$theme-values". So then the relation to the existing "$value" is clearer.

KristinLBradley: Another idea would be to follow the naming of the `prefers-color-scheme` CSS media query and use the word "scheme" instead of "mode" to refer to "light" and "dark" variants.

(Sorry, I keep adding ideas related to naming here as they come to me)

didoo: @KristinLBradley please do! I didn't think of "scheme" which is also a good alternative (the only doubt I have is that then it gets confused with `prefers-color-scheme`). Another one Copilot suggest is "skin" but it sound strange to me (even though, what technically we're doing is "reskinning" our components)

packages/tokens/scripts/build-parts/generateThemingCssFiles.ts
didoo: The "extra" is to indicate that these file are in addition to the "themed-tokens" files already generated by StyleDictionary (the standard output). 
The naming convention is coming from the fact that we're using `generate` for consistency with a similar action (`generate-css-helpers`)

I'll think if I can come up with alternatives

packages/tokens/scripts/build-parts/generateExtraThemingCssFiles.ts
didoo: Oh I didn't know! I'll add this to the list of things to discuss

packages/tokens/src/products/shared/typography.json
didoo: This is something I already discussed with @alex-ju and we ended up agreeing that in order to keep things simple, repetition may be an acceptable solution, even if not perfect.

Let me try to explain what I mean with this:
- in theory, we could omit the `hds` value, and use directly the `$value`; _but_ there may be cases where the new value we want to use is different from the "old" standard value [for reasons we don't know yet] → first variant
- in theory, there could be cases where all the `cds-gx` values are the same, so we could have a convention that if a key `cds` is found, this means that its value applies to all the `cds` modes → second variant
- in theory, there could be cases where the light values (`cds-g0`/`cds-g10`) are the same and/or the dark values (`cds-g0`/`cds-g10`) are the same, so we could have a convention that if a key `cds-g10` is missing, we fallback to the `cds-g0` or `cds`, and if a key `cds-g90` is missing, we fallback to the `cds-g100` or `cds` → third variant

Now, this means that in the codebase there will be a lot of combinations of these "variants" and we end up in a situation where we don't know anymore if a value is missing intentionally (to use the conventional fallback) or it's an error/oversight). 

So we came to the conclusion that was better to make things simple and explicit, even at the cost of repetition.

Makes sense?

packages/tokens/src/products/shared/typography.json
didoo: OK, clearly a bug, I have to look into it

didoo: > Will probably need a custom solution to get carbon icons into the background-image tokens.

I think the current approach using `$modes` should work, no? (the icons are embedded as `data:image` anyway, so it would be a different string; the only problem I see is theming, which can't be achieved using tokens, so we will have to come up with a color that works for all the themes

packages/tokens/dist/docs/products/tokens.json
didoo: > CDS-g10 theme, dark mode

`cds-g10` can be only light, it's a light [thing] :)

didoo: > Btw, in the context of the JSON, I also think just using the word "$themes" would be clearer to understand the purpose of at a glance than "$modes".

@KristinLBradley not _strongly_ attached to it, the decision to go for `$modes` comes from looking at different proposals/discussion (eg. [here](https://github.com/design-tokens/community-group/issues/2)) in the Design Tokens Working Group, and finding this one the more compelling: https://github.com/design-tokens/community-group/issues/210#issue-1636612239

Looking at how the Terrazzo app declares these tokens (see [here](https://github.com/terrazzoapp/dtcg-examples/blob/main/github-primer.json#L15-L26) for example) they use the `$extensions.mode` container (where `$extension` is [this thing](https://www.designtokens.org/tr/drafts/format/#extensions)), but I think it's an unnecessary extra layer, so I went directly for `$modes`. 

I think we can discuss/decide this in the context of how we're going to call things :)

packages/tokens/src/products/shared/typography.json
dchyun: > better to make things simple and explicit, even at the cost of repetition.

Makes sense to me. This is my thinking as well.

packages/tokens/src/products/shared/typography.json
dchyun: Oh true, good point. It would just be a different string so icons should be fine. Yeah since the fills are hard-coded into the string that would be the one piece that would not be able to use the tokens directly. Updated my comment to reflect this.

didoo: > Animation easings / Differences in the json format across HDS and Carbon

This potentially could be accounted for in the `extract-carbon-tokens` script (where we read the design tokens from the `@carbon/***` packages and we process them and save under the `packages/tokens/src/carbon-extracted` folder. We could align the format in which we save their tokens to ours (this is also true for the differences in units for the typographic tokens). What do you think?

didoo: Potential solution here: https://github.com/hashicorp/design-system/pull/3239#discussion_r2452903617

dchyun: Yeah I think that would be a good solution. `carbon-extracted` can then be aligned to our own formatting and we don't need to worry about handling anything later on down the pipeline.

packages/tokens/scripts/build-parts/generateExtraThemingCssFiles.ts
```diff
@@ -0,0 +1,104 @@
...
+      // this is the common part
+      outputContent += `${commonSource}\n\n`;
+    }
+
+    // CSS file for `.class/[data]` selectors
+    if (method === 'css-selectors') {
+      outputContent = `${header}\n\n`;
+      //
+      // this is the fallback to the default `hds` mode
+      outputContent += `${hdsThemedSource.replace(/^:root/, ':root, .hds-theme-hds, [data-hds-theme="hds"]')}\n\n`;
```
KristinLBradley: As you mentioned, I think it would be clearer to rename ".hds-theme-hds" to something like ".hds-theme-default" or ".hds-theme-legacy" or something along those lines.

packages/tokens/src/products/shared/typography.json
didoo: Re. the comment issue I've created a follow-up ticket for this: https://hashicorp.atlassian.net/browse/HDS-5670

didoo: Created a ticket to handle the differences in token format for the animation values: https://hashicorp.atlassian.net/browse/HDS-5671 (it's not a blocker for now, in the dry run we are not tackling animations)

didoo: @dchyun re. the issue with the units, can you check that this is still a problem?

dchyun: @didoo for the json files under `docs/products` I'm still seeing the units missing from the `default` mode.

```
{
    "key": "{typography.display-500.font-size}",
    "$type": "font-size",
    "$value": "1.875rem",
    "unit": "px",
    "comment": "30px/1.875rem",
    "$modes": {
      "default": "30",
      "cds-g0": "2.625rem",
      "cds-g10": "2.625rem",
      "cds-g90": "2.625rem",
      "cds-g100": "2.625rem"
    },
```

packages/tokens/scripts/build-parts/generateExtraThemingCssFiles.ts
dchyun: @didoo [Note] This is still occurring, but likely won't be figured out till all of the mapping is done. Can be something to look at later.

 When a core token is themed, but the semantic token that reference it is not, the semantic token remains in the `common-tokens` file for all themes. However, it has different values across themes, depending on the value of the core token it is aliased to.

packages/tokens/scripts/build-parts/generateThemingCssFiles.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export async function generateThemingCssFiles(_dictionary: Dictionary, config: PlatformConfig): Promise<void> {
+
+  const commonSource = await getSourceFromFileWithRootSelector(config, 'default', 'common-tokens.css');
+  const defaultThemedSource = await getSourceFromFileWithRootSelector(config, 'default', 'themed-tokens.css');
+  const cds0ThemedSource = await getSourceFromFileWithRootSelector(config, 'cds-g0', 'themed-tokens.css');
+  const cds10ThemedSource = await getSourceFromFileWithRootSelector(config, 'cds-g10', 'themed-tokens.css');
+  const cds90ThemedSource = await getSourceFromFileWithRootSelector(config, 'cds-g90', 'themed-tokens.css');
+  const cds100ThemedSource = await getSourceFromFileWithRootSelector(config, 'cds-g100', 'themed-tokens.css');
+
```
dchyun, line 20:
```suggestion
  const lightThemeSource = cds0ThemedSource;
  const darkThemeSource = cds100ThemedSource;
```
[Suggestion][Nit] We could create variables for what themes are used for `hds-theme-light`, `hds-theme-dark`, and `prefers-color-scheme` here instead of referencing cds0 and cds100 directly. If they change in the future we only have to update it here.

packages/tokens/scripts/build.ts
```diff
@@ -14,18 +15,81 @@ import { fileURLToPath } from 'url';
...
+    preprocessor: (dictionary, _options) => {
+      // recursively traverse token objects and replace the `$value` with the corresponding colocated `$modes` theme value
+      // note: the `slice` is always an object (a token or a parent group)
+      function replaceModes(slice: DesignToken) {
+        if (slice.$modes) {
+          if (mode in slice.$modes) {
+            slice.$value = slice.$modes[mode];
+          } else {
+            // TODO! decide if we want to throw here (and test if it works, by removing a value from one of the test files) - see: https://hashicorp.atlassian.net/browse/HDS-5668
+            console.error(`❌ ERROR - Found themed token without '${mode}' value:`, JSON.stringify(slice, null, 2));
```
dchyun, line 48: [Praise] When I was testing adding in modes I had misspelled `default` and this caught it right away. Think it's pretty helpful to keep.

packages/tokens/dist/products/css/themed-tokens/with-css-selectors--migration/tokens.css
dchyun, line 1: [Question] Just to make sure I'm understanding correctly are these the use cases for each `with-css-selectors` file?

- `--migration`
  - Used during migration to implement theming via theming classes, but still have access to the default (HDS) theme with `hds-theme-default`.
  - Will be removed post-migration 
- `--advanced`
  - Gives access to all carbon themes so someone can use specific ones like `hds-mode-cds-g10` 
- Base file
  - Doesn't have `hds-theme-default`
  - Would be main file teams use post-migration once default mode (HDS) is removed and one of the carbon themes becomes the root tokens. 
  - Could also be used now if a team doesn't need `hds-theme-default`

dchyun: The one use case I see missing here is if I want to use specific themes, and not import any others in. This would be covered by the `with-root-selector` files I believe.

The `--advanced` file would not cover this, because it includes all available themes in one file.

packages/tokens/src/global/color/palette-neutrals.json
```diff
@@ -44,7 +44,14 @@
       "neutral-0": {
         "$type": "color",
         "$value": "#ffffff",
-        "group": "palette"
+        "group": "palette",
+        "$modes": {
+          "default": "#ffffff",
```
dchyun, line 49: [Note][Future discussion] Looking way forward to post-migration. If we are in a post-hds theme world, would we then remove `default` as a mode? And`$value` would be updated to reflect whatever our new "default" theme is (ex: cds-g0).

dchyun (commented):
Overall looking good to me. All my local testing worked as expected.

packages/tokens/dist/products/css/themed-tokens/with-css-selectors--advanced/tokens.css
```diff
@@ -0,0 +1,538 @@
+/**
+ * Do not edit directly, this file was auto-generated.
+ */
+
+:root,
+.hds-theme-default {
+  --token-color-palette-neutral-0: #ffffff;
```
KristinLBradley, line 7: Just curious, is there a reason to not use shorthand for the hex color values like this? (i.e., use "#fff" vs. "#ffffff")

packages/tokens/dist/products/css/themed-tokens/with-css-selectors--advanced/tokens.css
```diff
@@ -0,0 +1,538 @@
...
+}
+
+@media (prefers-color-scheme: light) {
+  .hds-theme-system {
+    --token-color-palette-neutral-0: #ffffff;
+    --token-color-foreground-primary: #161616;
+    --token-typography-font-weight-regular: 400;
+    --token-typography-font-weight-medium: 400;
+    --token-typography-font-weight-semibold: 600;
+    --token-typography-font-weight-bold: 600;
```
KristinLBradley, line 22: So is the purpose of repeating the weight values because some themes don't use all weights, and thus we want to map the unused weights to existing similar ones in a particular theme?

If this is the case, would it make sense to define core variables for standardized weights and then map them to the variables in the theme? For example, assuming we used a combo of Sass variables and CSS custom properties (which we wouldn't need to do, but I'm doing here for ease in reading the example code), we could do something like this:

```
// Font-weight definition variable:
$font-weight-light: 300;
$font-weight-regular: 400;
$font-weight-medium: 500;
$font-weight-semibold: 600;
$font-weight-bold: 700;

// Font-weights as used in theming:
.hds-theme-default {
  --token-typography-font-weight-medium: $font-weight-medium;
  --token-typography-font-weight-semibold: $font-weight-semibold;
  --token-typography-font-weight-bold: $font-weight-bold;
}

.hds-theme-system {
    --token-typography-font-weight-regular: $font-weight-regular;
    --token-typography-font-weight-medium: $font-weight-regular;
    --token-typography-font-weight-semibold: $font-weight-semibold;
    --token-typography-font-weight-bold: $font-weight-semibold;
}
```

To me, this makes it easier to tell what weight something is meant to be without needing to understand/remember what number = what weight. (I guess there would be no real value other than that though so it's just an idea...)

packages/tokens/dist/products/css/themed-tokens/with-css-selectors--advanced/tokens.css
```diff
@@ -0,0 +1,538 @@
...
+at the scope where it is declared—using whatever is visible at that level in the DOM (eg. `:root`).
+
+This block of "common/shared" CSS variables contains a lot of these aliases, so if we only declare these
+"computed" variables at `:root` level, their values do not update in themed blocks that override
+the "source" variables (e.g., `--source-value`) at a lower scope.
+This results in _stale_ variables, where the computed value is always stuck on the `:root` value,
+even inside a themed block where the source variable is assigned.
+
+To make sure every themed block/scoped context (eg. `.hds-theme-light`) correctly computes the aliased values,
+we redeclare each "computed" variable in every theme scope where its "source" variable is overridden.
```
KristinLBradley, line 90: Oh, I guess this explains why my idea for defining variables for the various standardized font weights wouldn't work...

packages/tokens/scripts/build-parts/generateThemingCssFiles.ts
KristinLBradley: I like Dylan's idea

packages/tokens/scripts/build-parts/getStyleDictionaryConfig.ts
```diff
@@ -0,0 +1,271 @@
...
+
+  if (target === 'devdot') {
+    return {
+      ...baseConfig,
+      'source': [
+        `src/carbon-extracted/**/*.json`,
+        `src/global/**/*.json`,
+        `src/products/shared/**/*.json`,
+        // just uncomment the line below to include overrides for 'devdot' tokens
+        `src/devdot/**/*.json`
```
KristinLBradley, line 174: Was this line meant to be commented out?

packages/tokens/src/global/color/palette-neutrals.json
KristinLBradley: That's what I would assume. Eventually, the HDS default theme will become obsolete and will be entirely replaced with Carbon-based themes.

didoo: Yes. In a post-migration world, will be much easier to base our design token strategy on defaults and conventions. For example:
- a `$value` without `$modes` implies that all the modes are the same (as `$value`) - this is already like this
- a `$modes` entry with just a subset of modes implies that these are overrides, the modes that are not included implicitly have the same value as `$value`
- potentially, we could even get rid of `$value` at all, and assume that
    - if there is a single entry in `$modes`, it applies to all the modes
    - if there are only two entries they apply for light/dark modes
    - etc

packages/tokens/dist/products/css/themed-tokens/with-css-selectors--migration/tokens.css
didoo: Yes to everything you said above :)

Plus, per [comment in the code](https://github.com/hashicorp/design-system/blob/f8caa0f11d6b63deaa8c88fec16d043e1e68cfd0/packages/tokens/scripts/build-parts/generateThemingCssFiles.ts#L23): _for the moment we're covering only the use cases we know of, later we may add more formats if needed_

packages/tokens/dist/products/css/themed-tokens/with-css-selectors--advanced/tokens.css
didoo: no specific reason, that's how they're added in the JSON files

packages/tokens/src/products/shared/typography.json
didoo: OK, I'll investigate this

packages/tokens/dist/products/css/themed-tokens/with-css-selectors--advanced/tokens.css
didoo: these are the `dist` files, the "compiled" ones, they're not meant for humans, but for the code compilers :)
the relevant files and declarations are the ones in the `src` folder: these are the declarations for the font weights
https://github.com/hashicorp/design-system/blob/f8caa0f11d6b63deaa8c88fec16d043e1e68cfd0/packages/tokens/src/products/shared/typography.json#L143-L188

packages/tokens/scripts/build-parts/generateExtraThemingCssFiles.ts
didoo: OK, added this to the follow-up investigations

packages/tokens/scripts/build-parts/getStyleDictionaryConfig.ts
didoo: yeah that comment is outdated, I've updated it

packages/tokens/dist/products/css/themed-tokens/with-css-selectors--advanced/tokens.css
didoo: See comment above

packages/tokens/scripts/build-parts/generateThemingCssFiles.ts
didoo: I am going to resolve this, for now. Why I see the thinking behind it, I still think it's too much of an abstraction: I don't see us having more than the existing Carbon themes in the future, and if we have to make changes, it will be relatively easy to introduce them

packages/tokens/src/products/shared/typography.json
KristinLBradley: @didoo Is there a follow-up to this? If so, perhaps you could add a link here for reference? (or explain briefly & resolve the thread?)

didoo: Sorry, forgot to link the follow-up ticket: https://hashicorp.atlassian.net/browse/HDS-5679

packages/tokens/scripts/build-parts/generateExtraThemingCssFiles.ts
didoo: Related follow-up tickets:
- https://hashicorp.atlassian.net/browse/HDS-5669
- https://hashicorp.atlassian.net/browse/HDS-5667

packages/tokens/scripts/build-parts/generateExtraThemingCssFiles.ts
didoo: In the end I went with the "traditional" selector.

packages/tokens/scripts/build.ts
```diff
@@ -4,7 +4,8 @@
  */
 
 import StyleDictionary from 'style-dictionary';
-import type { Config, DesignToken, Platform } from 'style-dictionary/types';
+// import { getReferences, usesReferences } from "style-dictionary/utils";
```
dchyun, line 7: [Question] Can this be removed?

didoo: It's a memo for myself, will be used when working on this: https://hashicorp.atlassian.net/browse/HDS-5667

PR #3240, opened 2025-09-25
[04] [Project Solar / Phase 1 / Showcase] Add support for theming and theme-switching to the showcase

didoo:
> [!IMPORTANT]
> This PR was accidentally closed by wrongly rebasing the parent branch (GitHub saw 0 differences, so it closed this PR). The history of the PR with all its comments is still relevant, but the work will continue in https://github.com/hashicorp/design-system/pull/3390

### :pushpin: Summary

See https://github.com/hashicorp/design-system/pull/3390 for details 

<details>

</details>

### DIFF
--- a/packages/components/src/components/hds/theme-context/index.ts
+++ b/packages/components/src/components/hds/theme-context/index.ts
@@ -17,7 +17,7 @@ import type { HdsThemes, HdsModes } from '../../../services/hds-theming.ts';
 
 export interface HdsThemeContextSignature {
   Args: {
-    // it can be an `HdsTheme` or an `HdsMode`
+    // can be either an `HdsTheme` or an `HdsMode`
     context: HdsThemeContexts;
   };
   Blocks: {
--- a/showcase/app/components/page-foundations/theming/sub-sections/contexts.gts
+++ b/showcase/app/components/page-foundations/theming/sub-sections/contexts.gts
@@ -24,7 +24,7 @@ import {
 import type { ShwStylesheets } from 'showcase/services/shw-theming';
 
 const styleIsAvailable = helper(
-  ([currentStylesheet, context]: [ShwStylesheets, string]) => {
+  ([currentStylesheet, context]: [ShwStylesheets, HdsThemeContexts]) => {
     const isAvailable =
       (context === 'default' &&
         (currentStylesheet === 'css-selectors--migration' ||
--- a/showcase/app/components/mock/app/sidebar/app-side-nav.gts
+++ b/showcase/app/components/mock/app/sidebar/app-side-nav.gts
@@ -156,7 +156,6 @@ export default class MockAppSidebarAppSideNav extends Component<MockAppSidebarAp
             {{yield to="extraAfter"}}
           </SNL.ExtraAfter>
         {{/if}}
-
       </HdsAppSideNavList>
     </HdsAppSideNav>
   </template>


showcase/app/components/shw/theme-switcher/popover.gts
```diff
@@ -0,0 +1,182 @@
...
+      case 'fixed-controls':
+        this._hasFixedControls = input.checked;
+        break;
+      case 'debugging-panel':
+        this._hasDebuggingPanel = input.checked;
+        break;
+    }
+  };
+
+  onApplyThemingPreferences = () => {
```
dchyun, line 99: [Issue] Somewhere in this flow the `_selectedLightTheme` and `_selectedDarkTheme` updates are not getting captured / propagated to other areas. 

If my theme, via the shw themeing select, is set to "carbon / dark (g100)", and I use the settings menu to set my light and dark themes to g10 and g90.

- My theme is visually updated to g90 as expected
- I use the select to change to any other themes, such as "carbon / light (g10)" 
- My theme is incorrectly updated to the default light theme of g0
- The select options all show the default light and dark options of g0 / g100
- The settings menu still states my applied light and dark themes as g10 / g90 even though those aren't being applied.

showcase/public/assets/styles/@hashicorp/design-system-components-theming-with-css-selectors.css
```diff
@@ -0,0 +1,11455 @@
...
+  --token-typography-font-weight-bold: 700;
+  /* added these to test theming in the showcase */
+  --token-color-foreground-strong: #0c0c0e;
+  --token-color-foreground-primary: #3b3d45;
+  --token-color-surface-strong: #f1f2f3;
+  --token-color-surface-primary: #ffffff;
+  --token-color-surface-faint: #ffffff;
+}
+
+.hds-theme-light,
```
dchyun: [Question/Suggestion] Was this testing file from a previous implementation of the CSS files from the tokens? [In the existing tokens PR](https://github.com/hashicorp/design-system/pull/3239/files#diff-7f466bd0eacefd19426da52040d0547c1593649da5081db9ac8077a7b562d370R14) there are no `hds-{light/dark}-theme` and `[data-hds-theme={light/dark}]` selectors.

I don't think these selectors would be feasible since a consumer could change what themes they want to use for the light and dark color schemes from the default choice of g0 and g100, but if these were included in our outputted css file like this they would always be tied to the default light and dark themes.

showcase/app/components/shw/theme-switcher/selector.gts
```diff
@@ -0,0 +1,98 @@
...
+    return {
+      'No theming': {
+        'no-theming|': 'HDS / Standard',
+      },
+      'Theming via prefers-color-scheme': {
+        'prefers-color-scheme|system': 'Carbon / System',
+      },
+      'Theming via CSS selectors': {
+        'css-selectors|': 'HDS / Default',
+        'css-selectors|light': `Carbon / Light (${this.gLight})`,
```
dchyun:
```suggestion
        'css-selectors|light|g0': `Carbon g0`,
        'css-selectors|light|g10': `Carbon g10`,
        'css-selectors|dark|g90': `Carbon g90`,
        'css-selectors|dark|g10': `Carbon g100`,
```
[Suggestion] For both the css selector and combined strategy options, it could be helpful for testing purposes to switch between all of the themes with the select, instead of what is currently selected as the light and dark themes. If we had a higher number of light and dark themes it would be quicker to look through each one just with the select instead of having to go update the advanced options each time.

showcase/app/styles/showcase-theming/index.scss
```diff
@@ -0,0 +1,49 @@
...
+
+// this works because the specificity of `:root` is 010, the same as a class-based and an attribute-based selector; when specificity is equal, the later rules override earlier ones due to cascade order
+
+@media (prefers-color-scheme: light) {
+  :root {
+    @include shw-theme-color-variables-light();
+  }
+}
+
+@media (prefers-color-scheme: dark) {
```
dchyun, line 20: [Note] I don't have a good solution to propose for this problem, but setting the color scheme dark showcase variables for the whole showcase can make reviewing default HDS theme components more challenging, when the dev has their system set to dark mode.

I have my system set to a dark color scheme, which means my whole showcase uses the dark showcase variables even when the default HDS theme is applied which does not have a dark version. 

Because all of the HDS theme options in the theme selector don't trigger the light showcase variables, the only way for me to see the HDS components on their intended lighter background is by updating my system to be light. 

Image

Maybe one solution could be using actual HDS token values here instead of hardcoded ones. Then they would still be light when no theme is applied

packages/components/src/services/hds-theming.ts
```diff
@@ -0,0 +1,232 @@
...
+        rootElement.setAttribute(HDS_THEMING_DATA_SELECTOR, this._currentMode);
+      } else if (this._currentCssSelector === 'class') {
+        rootElement.classList.add(
+          `${HDS_THEMING_CLASS_SELECTOR_PREFIX}-${this._currentMode}`
+        );
+      }
+    }
+
+    // store the current theme and theming options in local storage (unless undefined)
+    localStorage.setItem(
```
dchyun, line 162: [Suggestion] There could be use cases where a consumer wants to handle on their own how they keep track of a user's theme. They could want to handle it through their own cookies or storage or some other means. It would be useful to have an argument like `hasLocalStorage`, where if false then nothing would get set from our service in local storage.

A consumer could then use the `globalOnSetTheme` to listen for theme updates and handle them in their own way.

showcase/public/assets/styles/@hashicorp/design-system-components-common.css
dchyun, line 1: > do we want to add hds-theme-system or data-hds-theme="system" when the user-selected option is system?

Personally I don't think this is needed. We could say that if you are using the system theme you are then relying on the prefer-color-scheme selectors and shouldn't be trying to control that with some other selector.

packages/components/src/services/hds-theming.ts
dchyun: > do we want to keep the globalOnSetTheme callback? which use cases do we foresee?

Because of my point above I think it would be useful to keep the callback.

showcase/public/assets/styles/@hashicorp/design-system-components-common.css
KristinLBradley: I tend to agree with this.

showcase/app/components/shw/theme-switcher/popover.gts
didoo: yeah a bug introduced in one of the latest refactorings, I'm looking into it

didoo: Ok, main part of the bug has been fixed in https://github.com/hashicorp/design-system/pull/3240/commits/c3a9268502e53b7aac269314883cc21ab4741279. 

The fact that the selected values in the popover are out of sync with the actual values in the service is a second bug that I'll look into now

didoo: Second part fixed in https://github.com/hashicorp/design-system/pull/3240/commits/78002785926a476a19f5e31ad6965462083784eb (this was a bit trickier, I had to rethink some of the logic, due to the fact that the popover element is never removed from the DOM)

showcase/public/assets/styles/@hashicorp/design-system-components-theming-with-css-selectors.css
didoo: yes, this is still part of the experimentation we're doing, around "how we want to expose theming to our consumers". One option I'm considering is to have simpler CSS selectors  (`hds-{light/dark}-theme` and `[data-hds-theme={light/dark}]`) for the "standard" use cases (in the assumption that designers will want consumers to use only two themes, as general guidance) and then find a more complex alternative (but still pre-defined by us, eg. via the `combined-strategy`) to let them offer to their consumers all the available themes (eg. the end-user may choose which light theme to use (between `g0` and `g10`) and which dark one to use (between `g90` and `g100`) for example in their settings, like you can do in GitHub for example.

In any case, something to discuss further

showcase/app/components/shw/theme-switcher/selector.gts
didoo: This is my thinking: as a showcase user (developer/designer) I would use
- the "carbonized" pages (see https://github.com/hashicorp/design-system/pull/3241) to see all at once all the possible variants, states **and themes**, for a component
- the normal pages for more standard/simple testing (where maybe I don't need to test all the themes, expecially when all the foundations have been set, so we can "trust" the colors/themes to just work)

showcase/app/styles/showcase-theming/index.scss
didoo: yeah, that's also why I was considering having an explicit `hds-theme-system` class or a `data-hds-theme="system"` attribute applied to the root (https://github.com/hashicorp/design-system/pull/3240#discussion_r2449678027)

packages/components/src/services/hds-theming.ts
didoo: With the current implementation, what they could do would be to use the `setTheme()` call in the application controller, instead of the `initializeTheme()` ([see here](https://github.com/hashicorp/design-system/blob/project-solar/phase-1/HDS-5242_showcase/add-themes-support/showcase/app/controllers/application.ts#L24)) and set their own theme and options; this would write on localstorage, but they could simply ignore our values and use theirs way of storing this information (as you suggested, cookie for example).

The `setTheme()` method has a callback, so _in theory_ they may use that, instead of the global one, so we still have to find a good reason to keep it alive :)

showcase/public/assets/styles/@hashicorp/design-system-components-common.css
didoo: See [comment above](https://github.com/hashicorp/design-system/pull/3240#discussion_r2452796508) (I am not sure it would work, but that's why I am still keeping it open as a possibility)

showcase/app/styles/showcase-theming/index.scss
didoo: Another consideration for using an explicit selector, is that we need to provide our consumers that want to use `prefers-color-scheme` a way to turn on and off the "theming" functionality (opt-in, while they have already bumped the version and imported the "themed" CSS) e.g. to do the migration incrementally, while fixing their side of the theming. 

This (which I've just realized) makes me strongly think that we don't have an alternative.

packages/components/src/services/hds-theming.ts
```diff
@@ -0,0 +1,233 @@
...
+export enum HdsThemeValues {
+  // system settings (prefers-color-scheme)
+  System = 'system',
+  // user settings for dark/light
+  Light = 'light',
+  Dark = 'dark',
+}
+
+enum HdsModesBaseValues {
+  Hds = 'hds', // TODO understand if it should be `default`
```
KristinLBradley: If consumers don't do anything, the "Hds" theme is the one they should get so that would be a "default". Is that what you mean?

packages/components/src/services/hds-theming.ts
```diff
@@ -0,0 +1,233 @@
...
+}
+
+enum HdsModesLightValues {
+  CdsG0 = 'cds-g0',
+  CdsG10 = 'cds-g10',
+}
+
+enum HdsModesDarkValues {
+  CdsG90 = 'cds-g90',
+  CdsG100 = 'cds-g100',
```
KristinLBradley, line 26: I still don't really understand the purpose of having multiple Light & Dark value options.

packages/components/src/services/hds-theming.ts
```diff
@@ -0,0 +1,233 @@
...
+export const HDS_THEMING_LOCALSTORAGE_DATA = 'hds-theming-data';
+
+export const DEFAULT_THEMING_OPTION_LIGHT_THEME = HdsModesLightValues.CdsG0;
+export const DEFAULT_THEMING_OPTION_DARK_THEME = HdsModesDarkValues.CdsG100;
+export const DEFAULT_THEMING_OPTION_CSS_SELECTOR = 'data';
+
+export default class HdsThemingService extends Service {
+  @tracked _isInitialized: boolean = false;
+  @tracked _currentTheme: HdsThemes | undefined = undefined;
+  @tracked _currentMode: HdsModes = undefined;
```
KristinLBradley: Would it be clearer to include "undefined" for the type here (vs. including it in the type definition above) the same way it's defined for `_currentTheme` or is that a dumb question? (Or change the type definition for `HdsThemes`?)

(i.e. use `HdsModes | undefined = undefined;`)

packages/components/src/services/hds-theming.ts
KristinLBradley: Would using `globalOnSetTheme` be easier or more straightforward for them to use vs. `setTheme()` in any way?

showcase/app/services/shw-theming.ts
```diff
@@ -0,0 +1,102 @@
...
+      newStylesheet =
+        'assets/styles/@hashicorp/design-system-components-theming-with-css-selectors.css';
+      break;
+    case 'combined-strategies':
+      // this is used for local testing purposes
+      newStylesheet =
+        'assets/styles/@hashicorp/design-system-components-theming-with-combined-strategies.css';
+      break;
+    default:
+      // this is the standard CSS for HDS components, without any theming
```
KristinLBradley:
```suggestion
      // this is the "standard" CSS for HDS components, without any theming
```

(just to make it stand out a bit more clearly)

packages/components/src/services/hds-theming.ts
didoo: no, `globalOnSetTheme` is an (optional) callback function that gets called (if provided) when `setTheme()` is executed:
```
setTheme(
    ↳ onSetTheme()
    ↳ globalOnSetTheme()
)

```

packages/components/src/services/hds-theming.ts
didoo: I had a few difficulties including `undefined` in the `HdsThemes` type definition, like I've done for the `HdsModes`. TypeScript was complaining in some places, so I ended up using a slightly different definition.

packages/components/src/services/hds-theming.ts
didoo: I'll explain in today's sync

packages/components/src/services/hds-theming.ts
didoo: I think I may understand what you meant. I've made the update here: https://github.com/hashicorp/design-system/pull/3240/commits/fd95fc1107d9db650ad1cc8ccf90f658a308ad4a

showcase/app/components/shw/theme-switcher/selector.gts
dchyun: > the assumption that designers will want consumers to use only two themes, as general guidance

This could also depend on the outcome of this assumption. If design wants to have a suggested light and dark then only showing those in the theme switcher makes sense to me.

packages/components/src/services/hds-theming.ts
dchyun: > The setTheme() method has a callback, so in theory they may use that, instead of the global one, so we still have to find a good reason to keep it alive 

This was the main use case I was thinking to keep it for. It's good to have some way for consumers to listen for theme changes and `onSetTheme` covers that. So maybe it's not needed.

packages/components/src/services/hds-theming.ts
KristinLBradley: Thanks, I was mainly concerned about the inconsistency so the updates look good to me.

showcase/app/index.html
```diff
@@ -14,6 +14,30 @@
 
     {{content-for "head"}}
 
+    <!-- these styles are pre-compiled in the `tokens` and `components` packages -->
+    <!-- IMPORTANT: if you add a new entry here, make sure to check if it's needed also in the `showcase/tests/index.html` file, which is used to run the tests -->
+
+    <!-- this is the "standard" (non-themed) CSS file for the HDS components, which contains both the default HDS tokens as well as the HDS components' styles  -->
+    <link id="hds-components-stylesheet-default" rel="stylesheet" integrity=""
+      href="{{rootURL}}assets/styles/@hashicorp/design-system-components.css">
```
dchyun, line 22:
```suggestion
      href="{{rootURL}}assets/styles/@hashicorp/tokens.css">
```
[Suggestion] Since the `design-system-components.css` file with the tokens + component styles is kind of considered "legacy" with our new output, would it make more sense to add the base tokens file to `public/assets/styles/@hashicorp` and use that here instead? 

The component styles are already imported below from `design-system-components-common.css` so those wouldn't be needed here.

showcase/app/components/shw/theme-switcher/selector.gts
```diff
@@ -0,0 +1,98 @@
...
+      HdsThemes | '',
+    ];
+
+    // we set the `currentStylesheet` in the `shwTheming` service
+    this.shwTheming.setStylesheet(selectedStylesheet);
+    // we set the `currentTheme` in the `hdsTheming` service
+    this.hdsTheming.setTheme({
+      theme: selectedTheme === '' ? undefined : selectedTheme,
+      // example of how a consumer could use the `onSetTheme` callback by passing it to the `setTheme` function as extra option
+      // onSetTheme: ({ currentTheme, currentMode }) => {
```
dchyun, line 86: [Issue][Nit] When using the HDS theme selector on the foundations/theming page:

- if the showcase theming is set to "HDS / Standard" or "Carbon / System"
- Using the HDS theme switcher on the foundations/theming page I change the theme to "light" or "dark" from "system"
- The `hds-theming-data` storage variable gets updated, but the `shw-theming-current-stylesheet` remains on `standard` or `system`
- The showcase theme selector does not update to a light or dark theme

This is because this kind of change ideally requires the showcase theming to go between different `shw-theming-current-stylesheet` values. Since the hds default and system themes dont have their own light or dark, if a theme change is triggered when one of these is active we could have the theme go to the css selector stylesheet by default. We could listen for these kinds of changes with the `onSetTheme` and load in the needed stylesheet. 

The showcase theme selector is getting correctly if an HDS theme switch happens while the theme is set to a css selector or combined strategy theme.

showcase/tests/index.html
```diff
@@ -14,8 +15,20 @@
     {{content-for "head"}}
     {{content-for "test-head"}}
 
+    <!-- these styles are pre-compiled in the `components` package -->
+    <!-- IMPORTANT: the relevant/required entries here always need to be kept in sync with the `showcase/app/index.html` file -->
+
+    <!-- this is the "standard" (non-themed) CSS file for the HDS components, which contains both the default HDS tokens as well as the HDS components' styles  -->
+    <link rel="stylesheet" href="{{rootURL}}assets/styles/@hashicorp/design-system-components.css">
```
dchyun, line 22:
```suggestion
    <link rel="stylesheet" href="{{rootURL}}assets/styles/@hashicorp/design-system-components-common.css">
    <link rel="stylesheet" href="{{rootURL}}assets/styles/@hashicorp/tokens.css">
```
[Question] Should we use the `design-system-components-common` file here as well since that will be the preferred way of using the styles going forward?

showcase/app/index.html
```diff
@@ -14,6 +14,30 @@
...
+    <!-- these are the design tokens (CSS variables) for theming [NOTE: they are disabled by default and toggled on/off via JS by the `ShwThemeSwitcher` selector] -->
+    <link id="hds-tokens-with-prefers-color-scheme" rel="stylesheet" integrity="" disabled="true"
+      href="{{rootURL}}assets/styles/@hashicorp/themed-tokens/with-prefers-color-scheme/tokens.css">
+    <link id="hds-tokens-with-css-selectors" rel="stylesheet" integrity="" disabled="true"
+      href="{{rootURL}}assets/styles/@hashicorp/themed-tokens/with-css-selectors/tokens.css">
+    <link id="hds-tokens-with-combined-strategies" rel="stylesheet" integrity="" disabled="true"
+      href="{{rootURL}}assets/styles/@hashicorp/themed-tokens/with-combined-strategies/tokens.css">
+
+    <!-- this is the "components-only" CSS file for the HDS components, which needs to be associated with a themed CSS tokens file to work  -->
+    <link id="hds-components-stylesheet-common" rel="stylesheet" integrity="" disabled="true"
```
dchyun, line 33: [Note] A future task not related to this PR, but we should update the website to follow this new approach of using the separated token and component styles as well.

packages/components/src/services/hds-theming.ts
didoo: @KristinLBradley I just noticed that I didn't explain this in the sync. Essentially, since Carbon supports 4 themes/modes, we should do the same (in case some consumer, for reasons we didn't anticipate, may need to use a different combination that the standard `g0/g100`).

showcase/app/index.html
didoo: > is kind of considered "legacy" with our new output

Well, not yet 😀. This will remain the standard for quite some time, and the carbonization is going to be _**opt-in**_, not opt-out, so we should keep the showcase behaviour as similar as possible to what our consumers would be doing (choosing a different stylesheet/theme in the theme switcher is the equivalent of opting in)

showcase/tests/index.html
didoo: As mentioned above, the standard/current way of using the tokens and components styles will remain the same as is now for some time, so I would prefer to keep things as they are now.

showcase/app/index.html
didoo: Yes. How to update the website is going to be a big, complex task, done in the second part of the milestone 1 (we need to document not just how to opt-in in the carbonization/theming but also think how to update the components' pages to include theme switching, guidance on the usage of the carbonized HDS components, etc)

showcase/app/components/shw/theme-switcher/selector.gts
didoo: @dchyun what do you think of this solution? https://github.com/hashicorp/design-system/blob/project-solar/phase-1/HDS-5242_showcase/add-themes-support/showcase/app/components/page-foundations/theming/sub-sections/theme-switcher.gts#L21-L37

showcase/app/index.html
dchyun: > we should keep the showcase behaviour as similar as possible to what our consumers would be doing

That makes sense to me. It's really a question of whether we want to match what we want consumers to eventually do or what they're doing right now. Makes sense to match what they're doing right now for the time being.

showcase/app/components/shw/theme-switcher/selector.gts
dchyun: That looks great to me! Works for both the default theme and prefers color scheme. Small nit, personally I don't think the alert is necessary, but fine with having it too.

didoo: > Small nit, personally I don't think the alert is necessary

Yeah, I know it's annoying, but I wanted to make sure the user is aware we're changing the selection (that they potentially have just made) in the background, without their knowledge. I would have used a notification, if we had one, or some other mean, but this was the easiest/quickest. We can always improve it later, when things are more stable.

showcase/public/assets/styles/@hashicorp/design-system-components-theming-with-css-selectors.css
KristinLBradley: I'm in favor of having simpler light/dark selectors vs. g0, etc. selectors. I find the latter confusing and consumers may also.

showcase/app/styles/showcase-theming/index.scss
KristinLBradley: Testing this branch more locally (http://localhost:4201/components/button) ...

I set my system theme to "dark" then, in the Button Showcase, selected: "HDS/Standard" under "No theming". In the Advanced controls, the "CSS selector" was set to "data attribute"

image

I was still seeing a dark background on the page:
image

When I tested the Button page in this PR's generated preview the Button page with the same settings had a white background. I then tried selecting "HDS/Default" under "Theming via CSS selectors and the background turned dark". I switched back to "HDS/Standard" under "No theming" and then background remained dark instead of turning back to white.

This sounds somewhat similar to the behavior @dchyun described above.

packages/components/src/services/hds-theming.ts
KristinLBradley: I guess you just meant the naming. This could be called "default", "legacy", "standard" or something else along those lines to make it clearer vs. calling it "hds".

packages/components/src/services/hds-theming.ts
KristinLBradley: Ok, I understand we need to or should follow how it's done in Carbon, but I find it confusing.

packages/components/src/services/hds-theming.ts
didoo: Ticket to evaluate if make the local storage opt-out (or opt-in?) https://hashicorp.atlassian.net/browse/HDS-5566

showcase/app/components/shw/theme-switcher/selector.gts
didoo: I have now changed the logic, and it's much simpler (and the alert appears only in one specific case, when switching to `default` where the CSS is the one without the `default` theme). So I am going to close this for now, I think it's an OK solution

showcase/app/styles/showcase-theming/index.scss
didoo: I think this has been fixed now, with the new implementation, right? (unless I am misunderstanding the issue)

showcase/public/assets/styles/@hashicorp/design-system-components-common.css
didoo: resolving this now

showcase/public/assets/styles/@hashicorp/themed-tokens/with-css-selectors--advanced/tokens.css
```diff
@@ -0,0 +1,621 @@
...
+  --token-color-surface-strong: #f1f2f3;
+  --token-color-surface-primary: #ffffff;
+  --token-color-surface-faint: #ffffff;
+  --token-typography-font-stack-display:
+    -apple-system, BlinkMacSystemFont, Segoe UI, Helvetica, Arial, sans-serif,
+    Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol;
+  --token-typography-font-stack-text:
+    -apple-system, BlinkMacSystemFont, Segoe UI, Helvetica, Arial, sans-serif,
+    Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol;
+}
```
didoo, line 25: All these overrides are added to test the different themes application in the `foundations/theming` page. See screenshot below to understand what I mean:

Image

Note: they are overwritten with proper tokens in the follow-up PR https://github.com/hashicorp/design-system/pull/3241 so they're temporary, just for testing in this PR

showcase/app/styles/showcase-theming/index.scss
dchyun: @didoo the issue is still present, but I have a possible solution for it.

The main issue is that if a dev has their system theme as dark, which is what I have, then the dark showcase theme is applied even when the hds / default theme is active. 

This is an issue when looking at hds components because everything is on a black background, which makes it hard to see some of them because they are all intended to be a "light" theme.

If we change the selector for these styles from `:root` to `.hds-theme-system`, then the light theme showcase variables will be present anytime the hds theme is active regardless of the dev's local theme setting.

showcase/app/styles/showcase-theming/index.scss
```diff
@@ -0,0 +1,43 @@
...
+:root {
+  @include shw-theme-color-variables-light();
+}
+
+// PREFERS COLOR SCHEME
+
+// this works because the specificity of `:root` is 010, the same as a class-based and an attribute-based selector; when specificity is equal, the later rules override earlier ones due to cascade order
+
+@media (prefers-color-scheme: light) {
+  :root {
```
dchyun:
```suggestion
  .hds-theme-system {
```
[Suggestion] See previous thread about showcase theming variables. Using this selector would ensure the "light" theme showcase variables are always used when the hds theme is active.

showcase/app/styles/showcase-theming/index.scss
```diff
@@ -0,0 +1,43 @@
...
+    @include shw-theme-color-variables-light();
+  }
+}
+
+@media (prefers-color-scheme: dark) {
+  :root {
```
dchyun:
```suggestion
  .hds-theme-system {
```
[Suggestion] See previous thread about showcase theming variables. Using this selector would ensure the "light" theme showcase variables are always used when the hds theme is active.

**For dev with `(prefers-color-scheme: dark)` set locally:**

**Before**
Image

**After**
Image

showcase/app/styles/showcase-pages/theming.scss
```diff
@@ -0,0 +1,57 @@
...
+
+body.page-foundations-theming {
+  .shw-foundation-theming-light-background {
+    padding: 8px;
+    background: var(--shw-color-white);
+  }
+
+  .shw-foundation-theming-dark-background {
+    padding: 8px;
+    background: var(--shw-color-black);
```
dchyun, line 18: [Issue] The backgrounds for the `AppFooter` themes can change based on the active theme when ideally they should be theme-independent. 

Because these backgrounds are currently set to showcase theme variables they can change based on the active theme. For example, if carbon/dark is active then the "theme=dark" box actually becomes white because `--shw-color-black` is set to a light color for dark themes. 

This could probably be more easily handled once 05 gets brought in. Maybe this section could even use the `ShwCarbonizationComparisonGrid` from that PR instead. 

Image

showcase/app/styles/showcase-theming/index.scss
```diff
@@ -0,0 +1,43 @@
...
+  }
+}
+
+// CSS SELECTORS
+
+// after testing different options, we have decided to rely on `hds-` selectors instead of custom `shw-` selectors
+// to avoid extra un-necessary complexity (see different attempts in https://github.com/hashicorp/design-system/pull/3240)
+
+.hds-theme-light,
+.hds-theme-cds-g0,
```
dchyun:
```suggestion
.hds-mode-cds-g0,
```
[todo] Update class names for modes, as classes from styles are now `hds-mode-` not `hds-theme-`

showcase/app/styles/showcase-theming/index.scss
```diff
@@ -0,0 +1,43 @@
...
+.hds-theme-cds-g10 {
+  @include shw-theme-color-variables-light();
+}
+
+.hds-theme-dark,
+.hds-theme-cds-g90,
```
dchyun:
```suggestion
.hds-mode-cds-g90,
```
[todo] Update class names for modes, as classes from styles are now `hds-mode-` not `hds-theme-`

packages/components/src/services/hds-theming.ts
```diff
@@ -0,0 +1,207 @@
...
+      this._currentMode = undefined;
+    } else if (
+      theme === HdsThemeValues.System // system (prefers-color-scheme)
+    ) {
+      this._currentTheme = HdsThemeValues.System;
+      this._currentMode = undefined;
+    } else {
+      this._currentTheme = theme;
+      if (this._currentTheme === HdsThemeValues.Light) {
+        this._currentMode = this._currentLightTheme;
```
dchyun, line 136: [Question][Suggestion] Due to the `_currentMode` getting set when `_currentTheme` is `light` or `dark`, there are multiple theme classes getting added when the theme is light or dark.

For example, when the theme is "carbon/light" both the `hds-theme-light` and `hds-mode-cds-g0` classes are added.

Image

I think both of these classes don't need to be added. Really only the `hds-theme-light` should be added if the light and dark themes are the defaults we define for those utility classes.

So, maybe `_currentMode` should be `undefined` if the `_currentLightTheme` and `_currentDarkTheme` are the defaults?

In the use case where `_currentLightTheme` and `_currentDarkTheme` are not the defaults then I think only the `-mode-cds` utility class should be added. We wouldn't need `hds-theme-light`. So then `_currentTheme` would be `undefined` maybe?

packages/components/src/services/hds-theming.ts
```diff
@@ -0,0 +1,207 @@
...
+  darkTheme: HdsModesDark;
+};
+
+type SetThemeArgs = {
+  theme: HdsThemes | undefined;
+  options?: HdsThemingOptions;
+  onSetTheme?: OnSetThemeCallback;
+};
+
+export type OnSetThemeCallbackArgs = {
```
dchyun:
```suggestion
export type HdsThemingSetThemeCallbackArgs = {
```
[Nit] Include the hds service's name in the exported type name

packages/components/src/services/hds-theming.ts
```diff
@@ -0,0 +1,207 @@
...
+  currentTheme: HdsThemes | undefined;
+  currentMode: HdsModes | undefined;
+};
+
+export type OnSetThemeCallback = (args: OnSetThemeCallbackArgs) => void;
```
dchyun:
```suggestion
export type HdsThemingSetThemeCallback = (args: OnSetThemeCallbackArgs) => void;
```
[Nit] Include the hds service's name in the exported type name

packages/components/src/services/hds-theming.ts
```diff
@@ -0,0 +1,207 @@
...
+        };
+        this.setTheme({
+          theme,
+          options,
+        });
+      }
+    }
+  }
+
+  setTheme({ theme, options, onSetTheme }: SetThemeArgs) {
```
dchyun, line 96: [Question] Currently the service only supports setting a theme to either "light" or "dark", and customizing what the given "light" and "dark" modes are. Would we want to support using the service to set a specific theme, not just in the context of light/dark.

Ex: I want to set the theme to "cds-g10" or "cds-future-product-theme"

My inclination would be no, and that this is good for our current use cases, but wanted to poise the question.

showcase/app/styles/showcase-theming/index.scss
didoo: OK, made the change as suggested. Thanks 🙏

showcase/app/styles/showcase-theming/index.scss
didoo: Fixed. Thank you

showcase/app/styles/showcase-theming/index.scss
didoo: Fixed. Thank you

packages/components/src/services/hds-theming.ts
didoo: Good call. Done.

packages/components/src/services/hds-theming.ts
didoo: Good call. Done.

showcase/app/styles/showcase-theming/index.scss
didoo: OK, now it has been fixed. Thank you for checking again.

showcase/app/styles/showcase-theming/index.scss
didoo: done

showcase/public/assets/styles/@hashicorp/design-system-components-theming-with-css-selectors.css
didoo: This is now outdated. Closing the conversation.

packages/components/src/services/hds-theming.ts
didoo: Yes, in another time/context I would have certainly tried to be more future-proof, but for the Project Solar I am adopting a different approach, more realistic: it's _extremely_ unlikely that, in the timespan this work is supposed to cover (6-12-24 months?) we will need to have more future themes, or that Carbon will introduce new non-dark/light themes. So for the time being, I think it's OK to base our work on the assumption we'll need to support for major themes: `default` (migration-only), `system`, `light`, and `dark`.

showcase/app/styles/showcase-pages/theming.scss
didoo: Mmm, I am not convinced. The background is applied to the (fake/mock) wrapper of the AppFooter, not the AppFooter itself. And the fact that the background colors get inverted when the theme is switched, makes sense (since the colors of the elements inside the footer also get inverted, apart from the HashiCorp logo, but this is a bug and there is already [a ticket for that](https://hashicorp.atlassian.net/browse/HDS-4308))

Here there's no clean-cut solution: if we keep the background unchanged, then the text in the footer would become unreadable when switching themes:

**with `Carbon/light` theme:**
screenshot_5567

**with `Carbon/dark` theme:**
screenshot_5568

Probably is something related to the AppFooter theming implementation, which is not 100% ready for theming. In any case we will have to review all the "themable" HDS components, in a more holistic way, to understand what happens when they're in different theme context.

packages/components/src/services/hds-theming.ts
didoo: This in reality is intentional. Having double class allows to immediately spot not just which theme, but also which mode for that theme is applied. And the order of the declarations in the generated CSS files is made so that the alternative modes (`cds-g10`/`cds-g90`) can overwrite both the themes (`light`/`dark`) and the corresponding default modes (`cds-g0`/`cds-g100`). 

It's probably unintuitive, and I struggled a bit at the beginning to accept that it was OK to have both the selectors (theme and mode) for the light/dark themes, but at the end it made sense to me: it's the natural outcome of how themes/modes have been defined and implemented (similar to what happened with the types, where everything at a certain point clicked)

packages/components/src/components/hds/theme-context/index.ts
```diff
@@ -0,0 +1,69 @@
...
+
+import Component from '@glimmer/component';
+import { assert } from '@ember/debug';
+import type Owner from '@ember/owner';
+
+import {
+  HdsThemeContextThemesValues,
+  HdsThemeContextModesValues,
+} from './types.ts';
+import type { HdsThemeContexts } from './types.ts';
```
KristinLBradley, line 14:
```suggestion

import type { HdsThemeContexts } from './types.ts';
```

packages/components/src/components/hds/theme-context/index.ts
```diff
@@ -0,0 +1,69 @@
...
+
```
KristinLBradley, line 15:
```suggestion
```

packages/components/src/components/hds/theme-context/index.ts
```diff
@@ -0,0 +1,69 @@
...
+import type { HdsThemes, HdsModes } from '../../../services/hds-theming.ts';
+
+export interface HdsThemeContextSignature {
+  Args: {
+    // it can be an `HdsTheme` or an `HdsMode`
```
KristinLBradley:
```suggestion
    // can be either an `HdsTheme` or an `HdsMode`
```

showcase/app/services/shw-theming.ts
```diff
@@ -0,0 +1,114 @@
...
+    'hds-tokens-with-css-selectors--migration',
+    'hds-components-stylesheet-common',
+  ],
+  'css-selectors--advanced': [
+    'hds-tokens-with-css-selectors--advanced',
+    'hds-components-stylesheet-common',
+  ],
+};
+
+const updatePageStylesheets = (currentStylesheet: ShwStylesheets) => {
```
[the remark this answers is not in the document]
didoo: done

showcase/app/components/shw/theme-switcher/popover.gts
```diff
@@ -0,0 +1,197 @@
...
+        showAdvancedOptions: this._showAdvancedOptions,
+        hasFixedControls: this._hasFixedControls,
+        hasDebuggingPanel: this._hasDebuggingPanel,
+      });
+    }
+
+    // programmatically close the popover
+    const popoverElement = document.getElementById(this.args.popoverId);
+    if (popoverElement && 'hidePopover' in popoverElement) {
+      popoverElement.hidePopover();
```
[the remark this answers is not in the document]
didoo: it's for the showcase, it's OK to do it in such a simple way

showcase/app/components/page-foundations/theming/sub-sections/contexts.gts
```diff
@@ -0,0 +1,335 @@
...
+import ShwThemingService from 'showcase/services/shw-theming';
+import {
+  CONTEXTUAL_THEMES,
+  CONTEXTUAL_MODES,
+} from '@hashicorp/design-system-components/components/hds/theme-context/index';
+
+import type { ShwStylesheets } from 'showcase/services/shw-theming';
+
+const styleIsAvailable = helper(
+  ([currentStylesheet, context]: [ShwStylesheets, string]) => {
```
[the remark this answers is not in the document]
didoo: OK

packages/components/src/services/hds-theming.ts
```diff
@@ -0,0 +1,207 @@
...
+    DEFAULT_THEMING_OPTION_LIGHT_THEME;
+  @tracked _currentDarkTheme: HdsModesDark = DEFAULT_THEMING_OPTION_DARK_THEME;
+  @tracked globalOnSetTheme: HdsOnSetThemeCallback | undefined;
+
+  initializeTheme() {
+    const rawStoredThemingData = localStorage.getItem(
+      HDS_THEMING_LOCALSTORAGE_DATA
+    );
+    if (rawStoredThemingData !== null) {
+      const storedThemingData: unknown = JSON.parse(rawStoredThemingData);
```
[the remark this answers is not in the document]
didoo: Will be done in https://hashicorp.atlassian.net/browse/HDS-5689

packages/components/src/components/hds/theme-context/index.ts
didoo: I tend to visually separate what is imported from the `showcase` (relative or absolute), from what is imported from `packages/components`

showcase/app/styles/showcase-pages/theming.scss
dchyun: My main idea was just that the theming boxes should encapsulate the themes they are trying to showcase, like how the carbonization pr encapsulates the themes inside each example.

Ex: If we did something like add `hds-theme-light` around the light example and `hds-theme-dark` around the dark example then the components would always be visible even when the page theme is changed. 

Screenshot 2025-11-26 at 10 53 45 AM

That said, I'm fine leaving this as-is and revisiting it once the AppFooter is more ready for theming.

showcase/app/styles/showcase-theming/index.scss
dchyun: Looks great now thanks for updating it!

packages/components/src/components/hds/theme-switcher/index.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+const OPTIONS: Record<HdsThemes, ThemeOption> = {
+  default: { theme: 'default', icon: 'hashicorp', label: 'Default' },
+  system: { theme: 'system', icon: 'monitor', label: 'System' },
+  light: { theme: 'light', icon: 'sun', label: 'Light' },
+  dark: { theme: 'dark', icon: 'moon', label: 'Dark' },
+};
+
+interface HdsThemeSwitcherSignature {
```
dchyun, line 36:
```suggestion
export interface HdsThemeSwitcherSignature {
```
[Nit] Export the component signature

showcase/app/components/page-components/advanced-table/sub-sections/basic.gts
```diff
@@ -45,17 +46,14 @@ const BASIC_MUSIC_COLUMNS = [
 ];
 
 const SubSectionBasic: TemplateOnlyComponent = <template>
-  <ShwTextBody>
-    <a
-      class="shw-component-advanced-table__open-link"
```
dchyun: [todo] Can you remove the css for this class from the table showcase styles now that it's no longer needed?

showcase/app/components/page-foundations/theming/sub-sections/demo.gts
```diff
@@ -0,0 +1,29 @@
...
+  <ShwBodyLinkToRoute
+    @route="page-foundations.theming.frameless.demo-application-with-theme-switcher"
+    @isRouteExternal={{true}}
+  >
+    Open the demo application
+    <span class="sr-only">frame in a new window</span>
+  </ShwBodyLinkToRoute>
+
+  {{! NOTE: we don't use the frameless page because changing theme in the iframe interferes with the stylesheed in the parent page, and it may create confusion }}
+  {{! <ShwFrame
```
dchyun, line 21: [Question] Can this be removed?

showcase/app/components/mock/app/main/table-complex.gts
```diff
@@ -0,0 +1,164 @@
+import Component from '@glimmer/component';
```
dchyun, line 1:
```suggestion
/**
 * Copyright (c) HashiCorp, Inc.
 * SPDX-License-Identifier: MPL-2.0
 */
 
import Component from '@glimmer/component';
```
[Nit] Missing copyright header

showcase/app/components/page-foundations/theming/sub-sections/components.gts
```diff
@@ -0,0 +1,78 @@
+import type { TemplateOnlyComponent } from '@ember/component/template-only';
```
dchyun, line 1:
```suggestion
/**
 * Copyright (c) HashiCorp, Inc.
 * SPDX-License-Identifier: MPL-2.0
 */
 
import type { TemplateOnlyComponent } from '@ember/component/template-only';
```
[Nit] This file, and most of the files under `page-foundations/theming` are missing the copyright headers.

showcase/app/components/shw/theme-switcher/control/select.gts
```diff
@@ -0,0 +1,63 @@
+import Component from '@glimmer/component';
```
dchyun, line 1:
```suggestion
/**
 * Copyright (c) HashiCorp, Inc.
 * SPDX-License-Identifier: MPL-2.0
 */
 
import Component from '@glimmer/component';
```
[Nit] Missing copyright header

showcase/app/components/shw/theme-switcher/control/toggle.gts
```diff
@@ -0,0 +1,37 @@
+import Component from '@glimmer/component';
```
dchyun, line 1:
```suggestion
/**
 * Copyright (c) HashiCorp, Inc.
 * SPDX-License-Identifier: MPL-2.0
 */
 
import Component from '@glimmer/component';
```
[Nit] Missing copyright header

showcase/app/routes/page-foundations/theming/frameless/demo-application-with-theme-switcher.ts
```diff
@@ -0,0 +1,24 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Route from '@ember/routing/route';
+import type { ModelFrom } from 'showcase/utils/model-from-route';
+import { service } from '@ember/service';
+
+import ShwThemingService from 'showcase/services/shw-theming';
+
+export type PageComponentsFormFramelessDemoFormComplexModel =
```
dchyun, line 12:
```suggestion
export type PageFoundationsThemingFramelessDemoApplicationWithThemeSwitcherModel =
```
[todo] Update model name

showcase/app/routes/page-foundations/theming/frameless/demo-application-with-theme-switcher.ts
```diff
@@ -0,0 +1,24 @@
...
+  ModelFrom<PageComponentsFormFramelessDemoFormComplexRoute>;
```
dchyun, line 13:
```suggestion
  ModelFrom<PageFoundationsThemingFramelessDemoApplicationWithThemeSwitcherRoute>;
```
[todo] Update route name

showcase/app/routes/page-foundations/theming/frameless/demo-application-with-theme-switcher.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export default class PageComponentsFormFramelessDemoFormComplexRoute extends Route {
```
dchyun, line 15:
```suggestion
export default class PageFoundationsThemingFramelessDemoApplicationWithThemeSwitcherRoute extends Route {
```
[todo] Update route name

showcase/app/routes/page-foundations/theming/frameless/demo-application-with-theme-switcher.ts
dchyun, line 1: [todo] Every template page should have a route file, so one should be created for `page-foundations/theming/index.ts` as well.

showcase/app/styles/showcase-theming/dark-color-variables.scss
```diff
@@ -0,0 +1,22 @@
+// SHOWCASE COLORS > DARK THEME
```
dchyun, line 1:
```suggestion
/**
 * Copyright (c) HashiCorp, Inc.
 * SPDX-License-Identifier: MPL-2.0
 */
 
// SHOWCASE COLORS > DARK THEME
```
[todo] All of the `showcase-theming` scss files are missing the copyright header

showcase/app/templates/page-foundations/theming/frameless/demo-application-with-theme-switcher.gts
```diff
@@ -0,0 +1,441 @@
+import Component from '@glimmer/component';
```
dchyun, line 1:
```suggestion
/**
 * Copyright (c) HashiCorp, Inc.
 * SPDX-License-Identifier: MPL-2.0
 */
 
import Component from '@glimmer/component';
```
[todo] Add copyright header

showcase/app/templates/page-foundations/theming/index.gts
```diff
@@ -0,0 +1,8 @@
+import type { TemplateOnlyComponent } from '@ember/component/template-only';
```
dchyun, line 1:
```suggestion
/**
 * Copyright (c) HashiCorp, Inc.
 * SPDX-License-Identifier: MPL-2.0
 */
 
import type { TemplateOnlyComponent } from '@ember/component/template-only';
```
[todo] Add copyright header

dchyun (commented):
Functionality looks all good to me. Mainly just small things to clean up.

packages/components/src/components/hds/theme-switcher/index.ts
```diff
@@ -0,0 +1,108 @@
...
+  theme: HdsThemes | undefined;
+  icon: HdsIconSignature['Args']['name'];
+  label: string;
+}
+
+const OPTIONS: Record<HdsThemes, ThemeOption> = {
+  default: { theme: 'default', icon: 'hashicorp', label: 'Default' },
+  system: { theme: 'system', icon: 'monitor', label: 'System' },
+  light: { theme: 'light', icon: 'sun', label: 'Light' },
+  dark: { theme: 'dark', icon: 'moon', label: 'Dark' },
```
KristinLBradley, line 33: Just thinking...

In the labels, when "Default" is included as one of the options, would it be clearer to have the labels be something like:

--------
* Default
* Carbon Light
* Carbon Dark
--------

or

--------
* Default

**Carbon**:
* System
* Light
* Dark
--------

Something to more clearly distinguish that the Light, Dark, and System options are part of the Carbon theme/modes vs. the HDS theme? At least while the HDS theme is still an option. (After it's removed, then it wouldn't be necessary to include "Carbon")

Something more similar to the Switcher the Showcase uses which has groups of options together with labels seems clearer to me.

KristinLBradley: Since the design isn't finalized, I guess it's not important to worry about now though...

didoo:
@dchyun @KristinLBradley unfortunately a wrong origin definition in my local Git client made so that I pushed to the wrong branch (the parent) instead of this one, which made GitHub see 0 commits in this branch and close the PR, and now it can't be re-opened. I had to create a new PR https://github.com/hashicorp/design-system/pull/3390 and we'll continue the code review there, while at the same time I'll reply to your review comments in this one. Hopefully it should not be major changes, so _I think_ it's doable. Sorry for the extra work in having to review in one and comment in another.

showcase/app/styles/showcase-pages/theming.scss
didoo: I have created a ticket to review this later: https://hashicorp.atlassian.net/browse/HDS-5698

packages/components/src/components/hds/theme-switcher/index.ts
didoo: Yep. Done.

showcase/app/components/page-foundations/theming/sub-sections/demo.gts
didoo: If we decide to use query params for the Showcase instead of local storage (https://hashicorp.atlassian.net/browse/HDS-5673) then this becomes relevant again.

showcase/app/components/mock/app/main/table-complex.gts
didoo: I am letting the GH Action do this work for me :)

showcase/app/routes/page-foundations/theming/frameless/demo-application-with-theme-switcher.ts
didoo: Right, missed this one. Done in https://github.com/hashicorp/design-system/pull/3390

showcase/app/routes/page-foundations/theming/frameless/demo-application-with-theme-switcher.ts
didoo: Done in https://github.com/hashicorp/design-system/pull/3390

showcase/app/routes/page-foundations/theming/frameless/demo-application-with-theme-switcher.ts
didoo: Done in https://github.com/hashicorp/design-system/pull/3390

showcase/app/routes/page-foundations/theming/frameless/demo-application-with-theme-switcher.ts
didoo: That's an interesting one! @shleewhite @aklkv currently there is not route file, but the application works and the page is rendered correctly https://hds-showcase-git-project-solar-phase-1hds-5242-97b3b3-hashicorp.vercel.app/foundations/theming and I don't see errors in the browser's console. So why _technically_ is needed?

packages/components/src/components/hds/theme-switcher/index.ts
didoo: Yes, plus this is intended for the end-user, which may not be aware of what is Carbon (and even if it will be used by the product designers/developers while migrating, in production we don't expect to see the `default` option, so the end-users will always see only `System/Light/Dark`

PR #3259, opened 2025-09-30
[03] [Project Solar / Phase 1 / Themed CSS for HDS components] Rollup configuration and Sass processing for multiple files 

didoo:
### :pushpin: Summary

This is built on top of https://github.com/hashicorp/design-system/pull/3239 and is based on the initial spike done by @alex-ju in https://github.com/hashicorp/design-system/pull/3235

In this PR I have: 
- re-organized the components' Sass files architecture to split the common part related to HDS component in a separate file (`packages/components/src/styles/components/index.scss`)
- updated the Rollup configuration to process multiple Sass files and generate distinct CSS files in output:
    - `design-system-components.css` - current file for HDS components
    - `design-system-components-common.css` - new file for theming
        - contains just the styles for the HDS component, no CSS variables coming from the design tokens pipeline
        - requires that the CSS file with the "themed" design tokens is imported separately

Then, after discussion and agreement with @alex-ju, I have moved also the existing compilation of the `design-system-components.scss` and `design-system-power-select-overrides.scss` files to our custom plugin, so we can remove `rollup-plugin-scss` as dev-dependency
  - in the new configuration, the `design-system-power-select-overrides.scss` is not only copied in the `dist` folder, but also compiled to `.css` (and a `.map` file is generated)

### 🔎 How to review

You can look at the single commits, to see the sequence of steps and understand the explorations and thinking behind the PR, or you can look directly at the files, checking how the Scss files have been organized, how the Rollup plugin function has been implemented, and how the path for the `include` declarations have been updated to make it more obvious where a file was coming from.

### 🛠️ How to test

Check out the branch, delete the `dist` folder under `packages/components`, run the command `pnpm build` or `pnpm start`, and see how different files are copied/generated under `packages/components/dist/styles/@hashicorp` (source Scss files, compiled CSS files, map files)

You can also go one step further, and update (or add) some design tokens, run the `pnpm build` command in the `packages/tokens` folder, and then run the command `pnpm build` in the `packages/components` folder and see how the tokens changes are reflected in the generated CSS files for the components:
- if only `$modes` are changed, the file `packages/components/dist/styles/@hashicorp/design-system-components.css` should remain untouched
- if the `$value` is changed, the file `packages/components/dist/styles/@hashicorp/design-system-components.css` should be updated too (the CSS variable corresponding to the token that was changed)

### 🔗 External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-5505

***

- [ ] ~~A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))~~ 
    - changelogs will be added in the main feature branch

<details>

</details>

### DIFF
--- a/packages/components/src/styles/@hashicorp/design-system-components.scss
+++ b/packages/components/src/styles/@hashicorp/design-system-components.scss
@@ -6,5 +6,5 @@
 // these are the "standard" HDS tokens (the file comes from `packages/tokens/dist/`)
 @use "products/css/tokens";
 
-// these are the styles specific (and only) for the the HDS components
+// these are the styles specific (and only) for the HDS components
 @use "./design-system-components-common";


packages/components/src/styles/@hashicorp/design-system-components-theming-with-prefers-color-scheme.scss
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// these files come from `packages/tokens/dist/`
+@use "products/css/themed-tokens/with-prefers-color-scheme/tokens.css";
```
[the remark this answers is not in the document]
didoo: added to the TODOs to discuss with Dylan/Kristin

packages/components/src/styles/@hashicorp/design-system-components-theming-with-css-selectors.scss
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// these files come from `packages/tokens/dist/`
+@use "products/css/themed-tokens/with-css-selectors/tokens.css";
```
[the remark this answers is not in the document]
didoo: added to the TODOs to discuss with Dylan/Kristin

packages/components/src/styles/@hashicorp/design-system-components-theming-with-combined-strategies.scss
```diff
@@ -0,0 +1,22 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// these files come from `packages/tokens/dist/`
+@use "products/css/themed-tokens/with-combined-strategies/tokens.css";
```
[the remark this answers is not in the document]
didoo: added to the TODOs to discuss with Dylan/Kristin

packages/components/rollup.config.mjs
```diff
@@ -50,16 +83,34 @@ const plugins = [
...
-      'node_modules/@hashicorp/design-system-tokens/dist/products/css',
-    ],
-  }),
-
-  scss({
-    fileName: 'styles/@hashicorp/design-system-power-select-overrides.css',
-  }),
+  // We use a custom plugin for the Sass/SCSS compilation
+  // so we can have multiple input and multiple outputs
+  ...addScssCompilationPlugins([
```
dchyun, line 88: [Question] Will we be creating files here for the `with-root-selector` css tokens themes as well?

packages/components/src/styles/@hashicorp/design-system-components-theming-with-prefers-color-scheme.scss
```diff
@@ -0,0 +1,22 @@
...
+// TODO understand if these are common/shared or we should have different ones for the themed tokens
```
dchyun: [Note] I think these can stay as "common" and don't need to be changed based on the theme. All of the helper classes reference a CSS variable of the same name as the class (mostly). If the variables they reference are themed there shouldn't be any need to make any update to the classes themselves.

The one exception is the font weight classes which are hard coded, but this could be updated
```
.hds-font-weight-regular { font-weight: 400; }
.hds-font-weight-medium { font-weight: 500; }
.hds-font-weight-semibold { font-weight: 600; }
.hds-font-weight-bold { font-weight: 700; }
```

packages/components/rollup.config.mjs
```diff
@@ -50,16 +83,34 @@ const plugins = [
...
+    {
+      inputFile: 'design-system-components.scss',
+      outputFile: 'design-system-components.css',
```
dchyun, line 91: > would it make sense to have a "normal" CSS compiled for the "old" components, as is now, made of the "standard" non-themed files, and instead of compiling three different compiled "themed" CSS file, have only one for the components Scss code

[Discussion topic] I agree with your idea. I prefer there to be an "old" hds theme files that matches what exists today, and then separate theme and component style files that could be the future of how things work.

#### Theming use cases

To come to this opinion first I tried to lay out all the possible use cases a consumer would have for using themes both now and in the future

As a consumer I would like to...
- Use the original HDS theme only
    - I would prefer to import one file it is today, and not have to change my code to use one for themes and one for components
- Use all themes and switch between them
    - Combined strategy
    - CSS selectors
- Use prefers colors mode
    - Use the default themes selected by the HDS team for light and dark 
    - Customize which themes are set for each(?)
- Use only one theme (either from Carbon or HDS)
- Use some themes but not all

#### Proposed CSS file structure

If we keep the component styles and theme style separate it would improve flexibility of how consumers use the themes. They would know they always have to import the component styles, but could manage what themes they're using with more control.

This would be helpful with new projects if they are going to be built with Carbon. A team would just have to grab the component css and the root selector carbon theme css and be good to go. Maybe it could be helpful for teams wanting to use some of the themes, but not all as well. 

I would still like to keep around the existing `design-system-components.css` that contains both the tokens and components combined so teams aren't forced into any changes, but this could be removed eventually.

This would be my proposed CSS files structure:
- Themes
    - Root selector
        - Hds
        - Cds-g0
        - Cds-g10
        - Cds-g90
        - Cds-g100
        - Unknown future theme
    - Prefers color sceme
        - Fixed at having cds-g0 for light and cds-g100 for dark
    - Combined strategies
        - Fixed at having cds-g0 for light and cds-g100 for dark
- `components.css`
    - Only contains components styles
- `design-system-components.css`
    - Legacy file that contains both HDS root tokens and component styles
    - Could be removed in the future

packages/components/rollup.config.mjs
```diff
@@ -50,16 +83,34 @@ const plugins = [
...
+    },
+    {
+      inputFile: 'design-system-components-theming-with-css-selectors.scss',
+      outputFile: 'design-system-components-theming-with-css-selectors.css',
```
dchyun: (this comment has some overlap with the pipeline pr) 

[Suggestion][Discussion] For the `combined-strategies` and `css-selector` theming approaches, it would be beneficial to have one shared tokens file and separate files for each theme's css selector, similar to how `root-selector` is done. Each file for a theme would contain  only it's css selector tokens.

My basis for this idea is that it's most likely that the majority of tokens are themed, and the number of shared tokens is fairly small. It's also possible we add more component tokens in the future which would also need to be themed.

Based on the assumption that the majority of our tokens will be themed, that means for every theme there will be a high number of variables within these theme selector classes. That could bloat the size of this one file by a fair amount. It's also a possibility that new themes could be added in the future (maybe it's unlikely idk how Carbon handles those things). For every theme added we would be increasing the size of this file by a good amount.

There's also the use case that a consumer does not want all the themes. Maybe they only want g10 and g100. But if they want to use the css selector method they would need to bring in the tokens for those other themes if they use this file.

If we split up these selectors so there was one file for each, a consumer could choose to import in only the ones they need, and we could limit the length of these files.

dchyun (commented):
Have some general discussion comments around the structure of the css files, but overall the rollup is looking good to me. Did a test of adding new modes and all files were updated as expected.

packages/components/rollup.config.mjs
KristinLBradley: I agree with the idea of having separate "normal"/"old" CSS and themed files. I like Dylan's proposal for separating and structuring these files.

I think it's best if we plan things from an "opt-in" approach (at least to start) for consumers, in which, if they do nothing, they get what they have today. Otherwise, they have to take an explicit action to include and use themes as makes sense for their product without having to include too much unnecessary code.

packages/components/rollup.config.mjs
KristinLBradley: I like the idea of splitting of the files to reduce duplication and unnecessary code for consumers.

packages/components/rollup.config.mjs
didoo: mmm, no: consumer can import them directly from the `@hashicorp/design-system-tokens` package if they need them (this would be the case when they're implementing theming in such a custom way that none of the pre-implemented solutions we provide works for them, which is _really_ an edge case)

that said, we still have to discuss/consider the possibility that we don't generate different "themed" CSS files (embedding/including the tokens in the file) for the components, but instead generate a "common" CSS file for the components, and consumers import the pre-implemente themed CSS files with the CSS variables/tokens directly from the `@hashicorp/design-system-tokens` package, which would simplify a little the things

packages/components/src/styles/@hashicorp/design-system-components-theming-with-prefers-color-scheme.scss
didoo: > The one exception is the font weight classes which are hard coded, but this could be updated

Good point, I'll update the code for this as well

packages/components/rollup.config.mjs
didoo: > As a consumer I would like to...
> Use the original HDS theme only
> I would prefer to import one file it is today, and not have to change my code to use one for themes and one for components

Agree, in the case where consumers are not migrating to the carbonized HDS, everything should remain exactly the same as is now, with no changes to do on their side

> Use prefers colors mode
> Use the default themes selected by the HDS team for light and dark
> Customize which themes are set for each(?)

My expectation is that the HDS design team will limit the option to the two "main" themes `light/g0` and `dark/g100`. If a consumer _absolutely_ wants to use prefers colors mode with a custom set of modes, they can do it using the theming-specific `sass` mixins and build the theming [S]CSS files by themselves, or we can update the tokens pipeline to generate a custom file only for them

> Use all themes and switch between them
> Combined strategy
> CSS selectors

While the CSS selectors file currently supports multiple themes the idea is that once the designers have settled on the two themes/modes to use (likely `light/g0` and `dark/g100`) I'll update the code so that we generate only the two "recommended" selectors; potentially we could even consider using the more generic `hds-theme-light/hds-theme-dark` (in which case we should also update the `hdsTheming`/`shwTheming` code).

If a consumer really need to do custom things (eg, letting the end-user decide which light/dark mode to use, then the `combined-strategy` file would be the way to go (hence all the effort in the showcase to support this as realistic scenario, that required **a lot** of work and attempts to get it right)

> Use only one theme (either from Carbon or HDS)

Mmm, in that case they can just import the "with-root-selector" themed CSS files corresponding to the desired theme (and the common part, if we decide to keep it)

> Use some themes but not all

Mmm, I don't see this as a realistic scenario, and in that case they can use the theming-specific `sass` mixins and build the theming [S]CSS files by themselves, or we can update the tokens pipeline to generate a custom file only for them

didoo: > If we keep the component styles and theme style separate it would improve flexibility of how consumers use the themes. They would know they always have to import the component styles, but could manage what themes they're using with more control.
> This would be helpful with new projects if they are going to be built with Carbon. A team would just have to grab the component css and the root selector carbon theme css and be good to go. Maybe it could be helpful for teams wanting to use some of the themes, but not all as well.
> I would still like to keep around the existing design-system-components.css that contains both the tokens and components combined so teams aren't forced into any changes, but this could be removed eventually.

100% agree

didoo: > This would be my proposed CSS files structure:

I think we're saying exactly the same thing, but I _think_ you've missed one case, so let me rewrite it here:

- **Tokens CSS**
    - Root selector
        - Hds
        - Cds-g0
        - Cds-g10
        - Cds-g90
        - Cds-g100
        - Unknown future theme
    - Prefers color sceme
        - Fixed at having `light/cds-g0` and `dark/cds-g100
    - CSS selectors
        - Fixed at having `light/cds-g0` and `dark/cds-g100
    - Combined strategies
        - allows multiple "custom" (but pre-defined) strategies (see [modified example here](https://github.com/hashicorp/design-system/blob/project-solar/phase-1/HDS-5242_showcase/add-themes-support/showcase/public/assets/styles/%40hashicorp/design-system-components-theming-with-combined-strategies.css))
    - Scss mixins
         - for really custom implementations/needs
- **Components CSS**
- `design-system-components.css`
    - Legacy file that contains both HDS root tokens and component styles
    - Could be removed in the future
- `design-system-components-common.css`
    - Only contains components styles

Makes sense? 

In any case, once we have agreed (as I think we are) that we are not going to generate multiple "themed" CSS components files, but use directly the tokens package ones, this conversation will move into https://github.com/hashicorp/design-system/pull/3239 and https://github.com/hashicorp/design-system/pull/3240/

packages/components/rollup.config.mjs
didoo: We can postpone/delay the decision on this, once 1) we have more clarity on how the CSS files will be included by the consumers in the different scenarios, and 2) we have done some dry-run migrations (and we start to see how many tokens are common and how many are "themed")

packages/components/rollup.config.mjs
dchyun: > consumer can import them directly from the @hashicorp/design-system-tokens package if they need them

Not having theme root selector CSS files as part of the component package and letting teams just import the ones generated for those from the tokens package directly makes sense to me.

> this would be the case when they're implementing theming in such a custom way that none of the pre-implemented solutions we provide works for them, which is really an edge case

I wouldn't classify using only one theme via root selectors as an edge case. When teams are migrating over to use carbon themes, or a new project is getting set up, they most likely will not take on trying to implement a dark theme as well, since none (minus boundary) have a dark theme today.

They may say "to carbonize what we have today we only need the g0 theme, so that's all we want to use". 

I'm on board with the eventual goal to be having all products support light and dark themes, and then our pre-implemented solutions (prefers color scheme and css selectors) are what we should aim to have all teams use, but that day could be fairly far in the future. 

I guess a question is do we want to push teams to bring in our pre-implemented solutions now and carry the extra weight of those dark theme variables?

packages/components/rollup.config.mjs
dchyun: > let me rewrite it here:

This makes sense to me. The one piece I was missing was that if we split the components and theme styles the theming styles would only live within the tokens package, which makes sense to me.

packages/components/rollup.config.mjs
didoo: In theory they can always import the themed CSS, and assign a `hds-theme-[light/g0/g10]` class to their HTML, and it would work. They would have extra tokens in the shipped bundle/CSS, but it would be mostly compressed

packages/components/src/styles/@hashicorp/design-system-components-theming-with-prefers-color-scheme.scss
didoo: Created ticket for this (it can be done as a small separate PR): https://hashicorp.atlassian.net/browse/HDS-5563

packages/components/src/styles/components/index.scss
```diff
@@ -0,0 +1,52 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// Notice: this list can be automatically edited by the Ember blueprint, please don't remove the start/end comments
```
dchyun, line 6: [todo] In the ember blueprint `index.js` file we will need to update the path for the code that generates this file.

```
const cssFilePath = `${options.project.root}/../packages/components/src/styles/@hashicorp/design-system-components.scss`;
```

packages/components/src/styles/@hashicorp/design-system-components-common.scss
dchyun, line 1: [note for future] Not something that needs to be decided now. In the future, once the existing `design-system-components.css` file is deprecated, we could rename this file to `design-system-components` for simplicities sake. 

Would require consumers to make an update, but would allow us to say this `-common` prefix is just for this in-between period where both files exist. The `-common` really isn't needed if we get back to the case where only one file is being generated.

dchyun, line 1: [Note] The structure looks all good to me. I like the simplicity of not having the theming files be included in the components package. 

Did a test build and it generated as expected, and is much simpler than with the theming files. We do have some extra comments getting added in there when files get combined, but that is happening in our current build as well. 

We should update the showcase to follow our new recommendation of using the `design-system-components-common.css` file + a theme file (which I think has already been done in the 04 PR 🙌 )

Example of the extra comments:

```
/**
 */
/**
 */
/**
 */
/**
 */
.hds-accordion {
  display: flex;
  flex-direction: column;
}
```

packages/components/src/styles/components/index.scss
didoo: you're right! done.

packages/components/src/styles/@hashicorp/design-system-components-common.scss
didoo: yes, I had the same thought, but it's very down the line (in a year or so?)

didoo: > We should update the showcase to follow our new recommendation of using the design-system-components-common.css file + a theme file (which I think has already been done in the 04 PR 🙌 )

Yes, already done there. 

> We do have some extra comments getting added in there when files get combined, but that is happening in our current build as well.

Yeah, I noticed that too, but I didn't investigate why, and if there is something that we can fix/mitigate.

KristinLBradley: Perhaps we could add a step to strip all comments and then add one back at the top as a last step?

packages/components/rollup.config.mjs
dchyun, line 1: > You can also go one step further, and update (or add) some design tokens with $modes values, run the pnpm build command in the packages/tokens folder, and then run the command pnpm build in the packages/components folder and see how the tokens changes are reflected in the generated CSS files for the components

[Question] Is this comment in the description outdated now with the new structure? If I'm understanding correctly the two files created from the components package are theme agnostic with the new approach. The existing `design-system-components.css` still only includes the default theme values. So changing the `$modes` in the tokens shouldn't change anything in these files.

didoo: Yes, absolutely. I've missed that part, I'll update the PR description

packages/components/src/styles/@hashicorp/design-system-components.scss
```diff
@@ -3,63 +3,8 @@
...
-@use "../components/toast";
-@use "../components/tooltip";
-// END COMPONENT CSS FILES IMPORTS
-
-@use "../mixins/screen-reader-only" as *;
-// stylelint-disable-next-line selector-class-pattern
-.sr-only {
-  @include screen-reader-only();
-}
+// these are the styles specific (and only) for the the HDS components
```
KristinLBradley:
```suggestion
// these are the styles specific (and only) for the HDS components
```

KristinLBradley:
In packages/components/dist/styles/@hashicorp, the copyright in "design-system-components-common.css" and in "design-system-components.css" ends up below the first line of code. Probably not a big deal but it looks a little strange and causes the linter to complain about spacing.

image

In "design-system-power-select-overrides.css" and "design-system-power-select-overrides.scss" There's this linter complaint:
image

the linter also complains about spacing between style rules in "design-system-power-select-overrides.css" being absent.

didoo:
> In packages/components/dist/styles/@hashicorp, the copyright in "design-system-components-common.css" and in "design-system-components.css" ends up below the first line of code. Probably not a big deal but it looks a little strange and causes the linter to complain about spacing.
> 
> image
> In "design-system-power-select-overrides.css" and "design-system-power-select-overrides.scss" There's this linter complaint: image
> 
> the linter also complains about spacing between style rules in "design-system-power-select-overrides.css" being absent.

@KristinLBradley these are compiled files, there's not much we can do (as you mentioned, they're generated in the `dist` folder of `packages/components`)

KristinLBradley (approved):

PR #3266, opened 2025-10-01
`AdvancedTable` - Dynamic columns fix

zamoore:
### :pushpin: Summary

If merged, this PR fixes an issue when columns in the `AdvancedTable` do not match the internal `columnOrder` state.

A common pattern for our users is to add/remove columns from the `AdvancedTable`. Before this change, this would cause a mismatch between the ordered column keys and the available columns to order.

Now, instead of mapping over the column order keys and associating them with their column model, we reduce the column order keys and only match them to columns that exist.

Additionally, the `columnOrder` property only applies to tables where `hasReorderableColumns` is true. Tables that cannot be reordered will respect the order of the `columns` argument passed to them.

In the case that a user needs to have dynamic columns *as well as* reorderable columns, they will need to ensure that they are passing the `columnOrder` property with the *full* list of sorted column keys. *Even those that are not currently visible*. This is necessary for retaining order as columns are added or removed.

For example, if there are 3 *possible* columns: name, age, gender, but initially only 2 *visible* columns: name, and age, the `columnOrder` argument should still contain all possible column keys in their intended order.

toggle

***

<details>

</details>


showcase/app/components/page-components/advanced-table/sub-sections/functional-examples.gts
```diff
@@ -143,6 +168,22 @@ export default class SubSectionFunctionalExamples extends Component {
       </:topbarAction>
     </CodeFragmentWithDebugSelect>
 
+    <ShwTextH3>Toggle columns</ShwTextH3>
```
dchyun, line 224: [issue] I updated this test locally to see how removing columns would work if reorderable coluns were true. There seems to be an issue where if columns are reordered after a column is removed the order isn't reflecting right. 

https://github.com/user-attachments/assets/f31c6d81-b41f-48d5-8454-8261a602d168

showcase/tests/integration/components/hds/advanced-table/index-test.js
```diff
@@ -60,6 +69,7 @@ function getColumnByLabel(columns, label) {
 }
 
 async function getColumnOrder(columns) {
+  console.log({ columns })
```
dchyun:
```suggestion
```
[Nit] remove console log

showcase/app/components/page-components/advanced-table/sub-sections/functional-examples.gts
zamoore: I think I've addressed this now and added another demo for making reordering work with dynamic columns.

showcase/app/components/page-components/advanced-table/sub-sections/functional-examples.gts
```diff
@@ -70,6 +76,78 @@ export default class SubSectionFunctionalExamples extends Component {
     }, 5000);
   };
 
+  // TOGGLE COLUMN EXAMPLE
```
shleewhite, line 79: If you wanted, this could be its own code fragment for this example + the reorderable one - not a blocker at all tho.

showcase/app/components/page-components/advanced-table/sub-sections/functional-examples.gts
```diff
@@ -70,6 +76,78 @@ export default class SubSectionFunctionalExamples extends Component {
...
+      this.toggleColumnsVisibleColumns = [
+        ...this.toggleColumnsVisibleColumns,
+        columnKey,
+      ];
+    }
+  };
+
+  // TOGGLE REORDERABLE COLUMN EXAMPLE
+
+  toggleReorderableColumnsModel = USERS.slice(0, 4);
```
dchyun, line 110: [Nit][Non-blocker] You could move all the variables and tracked variables to the top of the file to align with the typical organization for those.

showcase/app/components/page-components/advanced-table/sub-sections/functional-examples.gts
```diff
@@ -143,6 +221,66 @@ export default class SubSectionFunctionalExamples extends Component {
...
+          }}
+        />
+      {{/each}}
+    </div>
+
+    <HdsAdvancedTable
+      @hasReorderableColumns={{true}}
+      @columnOrder={{this.toggleReorderableColumnsColumnOrder}}
+      {{! @glint-expect-error - will be fixed by https://hashicorp.atlassian.net/browse/HDS-5090}}
+      @model={{this.toggleReorderableColumnsModel}}
```
dchyun, line 270:
```suggestion
      @model={{USER_DATA_SHORT}}
```
[Suggestion][Non-blocker] You could define a const at the top of the file as we've done in some other data examples and not have to create a variable for the model.

```
const USER_DATA_SHORT = USERS.slice(0,4);
```

showcase/app/components/page-components/advanced-table/sub-sections/functional-examples.gts
zamoore: Right now they're organized by example, which I think might make it easier to read through

PR #3277, opened 2025-10-04
feat(components): remove `ember-get-config`

aklkv:
### :pushpin: Summary

`ember-get-config` is problematic in vite world and doesn't work without a [hack](https://github.com/hashicorp/cloud-admin/pull/1325/files/ef3a7012f2665823c7dfb12b4e7ff64a60bea1e8#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R22) this hack will not be release and it is advised to move to `appInstance` lookup

This just does that for components and website (which was pretty strange usecase)

+ bonus I removed no longer needed `ember-unique-id-helper-polyfill` support for this feature in included `ember-source` > `v4.4` - let me know if you want me to split it out

### :link: External links

Figma file: [if it applies]

***

<details>

</details>

### DIFF
--- a/packages/components/src/modifiers/hds-code-editor.ts
+++ b/packages/components/src/modifiers/hds-code-editor.ts
@@ -7,7 +7,7 @@ import Modifier from 'ember-modifier';
 import { assert, warn } from '@ember/debug';
 import { registerDestructor } from '@ember/destroyable';
 import { task } from 'ember-concurrency';
-import config from 'ember-get-config';
+import { macroCondition, isTesting } from '@embroider/macros';
 import { Compartment } from '@codemirror/state';
 import { EditorView } from '@codemirror/view';
 import { guidFor } from '@ember/object/internals';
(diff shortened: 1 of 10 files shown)


packages/components/src/modifiers/hds-code-editor.ts
```diff
@@ -208,8 +208,7 @@ export default class HdsCodeEditorModifier extends Modifier<HdsCodeEditorSignatu
     // if the editor does not exist, setup the editor
     else {
       // the intersection observer makes loading unreliable in tests
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-      if (config.environment === 'test') {
+      if (macroCondition(isTesting())) {
```
zamoore, line 211: Could you explain a bit how this works?

aklkv: that part essentially would remove this code at build time if `environment` is anything but `test` this pattern should be used when there is no way to get config from the host app or you want to be removing part of code at the build time

zamoore (approved):

PR #3281, opened 2025-10-06
`Website` update accordion-nesting-do image

andgen404:
### :pushpin: Summary

If merged, this PR will fix the nested `flush` accordion example. 

During a conversation in Design Review on August 27, we identified an issue with an example in the Accordion guidelines documentation. There is an inaccurate representation of the flush accordion within a container in the Nested Accordion section. 

accordion-nesting-do

### :link: External links

Jira ticket: [HDS-5425](https://hashicorp.atlassian.net/browse/HDS-5425)

[HDS-5425]: https://hashicorp.atlassian.net/browse/HDS-5425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jorytindall:
I'm trying to remember back to this conversation, but I don't think this image is correct. I think these are the updates that need to be made based on the image that is live:
- In the card with nested flush variants: The bottom border for the last nested flush accordion needs to be added
- The the flush example: the bottom border for the top-level accordion needs to be added and the bottom border for the last nested flush accordion needs to be added.

Somebody else double check me on this, but here's an example from the showcase:
Screenshot 2025-10-06 at 1 31 13 PM

andgen404:
> I'm trying to remember back to this conversation, but I don't think this image is correct. I think these are the updates that need to be made based on the image that is live:
> 
> * In the card with nested flush variants: The bottom border for the last nested flush accordion needs to be added
> * The the flush example: the bottom border for the top-level accordion needs to be added and the bottom border for the last nested flush accordion needs to be added.
> 
> Somebody else double check me on this, but here's an example from the showcase: Screenshot 2025-10-06 at 1 31 13 PM

Ahh, that jogs my memory on this, I believe you're correct. I went ahead and fixed the example! cc: @hashicorp/hds-design

PR #3525, opened 2026-01-29
[Project Solar / Phase 1 / Migration] `Accordion` "carbonization"

didoo:
### :pushpin: Summary

This PR intends to finalize the "carbonization" of the `Accordion` component.

In this PR I have:

- added component-level design tokens for the `Accordion` component
- updated the CSS files/styles for the `Accordion` component
- implemented the “carbonized” page for the `Accordion` component
- re-generated files in output

 👉 **Preview**: https://hds-showcase-git-project-solar-phase-1hds-5977-59b9bd-hashicorp.vercel.app/carbonization/components/accordion

#### Variants

Despite the apparent simplicity of the component, its "carbonization" involved **around 720 "variants" to consider/check** 😱! This back-of-the-envelope number comes out of this combinatorial calculation: (1 + 4 `modes` ) × (2 variants `card/flush`) × (3 sizes) × (2 `open/closed` states) × (3 variants `base` / `contains interactive` /  `is static`) × (4 interactive states `default/hover/active/focus`). Result: 🤯

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-5977

***

<details>

</details>


LilithJames-HDS:
Thanks @didoo! A couple nitpicks but this looks great

- The left and right padding should match in all the variants, currently it seems that the right padding on all the variants matches the top and bottom values instead (except for in the Large variants)
- The .ItemButton in Figma is a unique component that shares a close appearance with the DismissButton whereas the one used here appears more like the Button secondary variant

packages/tokens/src/products/shared/accordion.json
```diff
@@ -0,0 +1,420 @@
...
+                "cds-g10": "16",
+                "cds-g90": "16",
+                "cds-g100": "16"
+              }
+            }
+          }
+        }
+      },
+      "content": {
+        "padding": {
```
dchyun, line 318: [Question][Design] For the `item.content.padding` we are using a large top padding of 4px and a large bottom padding of 16px and the other sizes go down from these. The Carbon accordion uses 8px top padding and 24px bottom. Was it intentional not to align to this at least for the large size? 

Image

Image

showcase/app/components/page-carbonization/components/accordion/index.gts
```diff
@@ -0,0 +1,509 @@
...
+                <cds-accordion-item title="Nested item two/2">
+                  <ShwPlaceholder @text="generic content" @height="40" />
+                </cds-accordion-item>
+              </cds-accordion>
+            </cds-accordion-item>
+          </cds-accordion>
+        </:reference>
+      </ShwCarbonizationComparisonGrid>
+
+      {{!-- <ShwCarbonizationComparisonGrid>
```
dchyun: [todo] Can this commented section be removed?

showcase/app/components/page-carbonization/components/accordion/index.gts
```diff
@@ -0,0 +1,509 @@
...
+
+    <ShwTextH3>Rich content</ShwTextH3>
+
+    {{#each TYPES as |type|}}
+      <ShwCarbonizationComparisonGrid
+        @label="{{capitalize type}} with rich content in toggle"
+      >
+        <:theming>
+          <HdsAccordion @type={{type}} as |A|>
+            <A.Item @isOpen={{true}}>
```
dchyun:
```suggestion
            <A.Item @isOpen={{true}} @containsInteractive={{true}}>
```
[Issue] Adding links inside the accordion titles is only allowed when `containsInteractive` is true. Can update this example to reflect this.

showcase/app/components/page-carbonization/components/accordion/index.gts
```diff
@@ -0,0 +1,509 @@
...
+            <ShwCarbonizationComparisonGrid
+              @label={{state}}
+              @hideThemeLabels={{true}}
+              @hideCarbonLabels={{true}}
+            >
+              <:theming>
+                <ShwFlex @direction="column" as |SF|>
+                  {{#each booleans as |isOpen|}}
+                    <SF.Item>
+                      <HdsAccordionItem
```
dchyun, line 408: [Issue] For the active and hover mock states on the Card, those aren't being shown for the Carbon theme. They are present for the HDS theme.

Image

packages/components/src/styles/components/accordion.scss
dchyun:
```suggestion
    transform: rotate(180deg);
```
[Suggestion][Nit] The chevron icon on the Carbon component rotates the opposite of HDS. It rotates clockwise going from closed to open. We could add a carbon only mixin here to align to that.

packages/components/src/styles/components/accordion.scss
```diff
@@ -111,19 +152,15 @@
...
 .hds-accordion-item__button {
   padding: 0;
 
-  &:hover {
-    cursor: pointer;
-  }
-
   .hds-icon {
     @media (prefers-reduced-motion: no-preference) {
       transition: transform 0.3s;
```
dchyun:
```suggestion
      transition: transform 0.11s cubic-bezier(0.2, 0, 0.38, 0.9);
```
[Suggestion][Nit] The Carbon component has a slightly different animation timing. We could align to that in a carbon only mixin. 

Image

packages/components/src/styles/components/accordion.scss
```diff
@@ -198,15 +249,31 @@
   }
 }
 
-// Adjust content padding and border for nested flush accordions
+// NESTED ACCORDIONS
+// We need to adjust content padding and border for nested  accordions
 
 .hds-accordion-item--type-flush .hds-accordion-item__content:has(> .hds-accordion--type-flush) {
```
dchyun, line 268: [Suggestion] This existing style removes the top and bottom padding of the outer most accordion item content when a flush is used inside a flush. Since in Carbon themes both the card and flush styles look like flush, we should make a similar adjustment for any time accordions are nested.

Currently for both the card and mixed nested examples there is a padding gap between the outermost accordion and the inner ones. The nested flush does not have this problem. For carbon themes this results in visually identical nested accordions having different hover treatments

Nested flush
Image

Nested mixed / card

Image

packages/components/src/styles/components/accordion.scss
```diff
@@ -17,90 +20,128 @@
...
   &.hds-accordion--size-large {
-    gap: 12px;
+    gap: var(--token-accordion-card-gap-large);
   }
 }
 
-// ACCORDION ITEM COMPONENT (nested child)
+// ACCORDION ITEM > VARIANTS MAPPINGS
 
 .hds-accordion-item--size-small {
```
dchyun: [Suggestion] Could we simplify these size tokens to use a loop?

```
$hds-accordion-sizes: ("small", "medium", "large");

@each $size in $hds-accordion-sizes {
  .hds-accordion-item--size-#{$size} {
    --hds-accordion-item-toggle-padding-top: var(--token-accordion-item-toggle-padding-top-#{$size});
    --hds-accordion-item-toggle-padding-right: var(--token-accordion-item-toggle-padding-right-#{$size});
    --hds-accordion-item-toggle-padding-bottom: var(--token-accordion-item-toggle-padding-bottom-#{$size});
    --hds-accordion-item-toggle-padding-left: var(--token-accordion-item-toggle-padding-left-#{$size});
    --hds-accordion-item-toggle-gap: var(--token-accordion-item-toggle-gap-#{$size});
    --hds-accordion-item-toggle-icon-size: var(--token-accordion-item-toggle-icon-size-#{$size});
    --hds-accordion-item-content-padding-top: var(--token-accordion-item-content-padding-top-#{$size});
    --hds-accordion-item-content-padding-right: var(--token-accordion-item-content-padding-right-#{$size});
    --hds-accordion-item-content-padding-bottom: var(--token-accordion-item-content-padding-bottom-#{$size});
    --hds-accordion-item-content-padding-left: var(--token-accordion-item-content-padding-left-#{$size});
  }
}
```

packages/tokens/src/products/shared/accordion.json
```diff
@@ -0,0 +1,420 @@
...
+              "$value": "{color.surface.primary}",
+              "$modes": {
+                "default": "{color.surface.primary}",
+                "cds-g0": "transparent",
+                "cds-g10": "transparent",
+                "cds-g90": "transparent",
+                "cds-g100": "transparent"
+              }
+            },
+            "hover": {
```
dchyun: [Suggestion] For the `item.{card/flush}.surface.color.{default/hover}` would it make sense to put them under `item.{card/flush}.toggle.color.{default/hover}` since the hover state is only related to the toggle not the whole item?

packages/tokens/src/products/shared/accordion.json
```diff
@@ -0,0 +1,420 @@
...
+        "surface": {
+          "color": {
+            "default": {
+              "$type": "color",
+              "$value": "transparent"
+            },
+            "hover": {
+              "$type": "color",
+              "$value": "{color.surface.interactive-hover}",
+              "$modes": {
```
dchyun:
```suggestion
```
[todo] The alias for `interactive-hover` in the Carbon themes is `layerHover01` so you don't need modes here. Could even just use that semantic token directly in the sass.

packages/tokens/src/products/shared/accordion.json
```diff
@@ -0,0 +1,420 @@
...
+            }
+          }
+        },
+        "icon": {
+          "size": {
+            "small": {
+              "$type": "dimension",
+              "$value": "16",
+              "unit": "px",
+              "$modes": {
```
dchyun:
```suggestion
```
[todo] Can we remove the modes here since the value is the same for HDS and Carbon

dchyun (commented):
Think I got through all 720 😅. Looking really good, mostly some small comments.

showcase/app/components/page-carbonization/components/accordion/index.gts
didoo: Yes. Removed

showcase/app/components/page-carbonization/components/accordion/index.gts
didoo: Yeah, makes sense. Added.

packages/components/src/styles/components/accordion.scss
didoo: Good idea, I've added themed design tokens for this.

packages/components/src/styles/components/accordion.scss
didoo: Good idea, I've added themed design tokens for this.

showcase/app/components/page-carbonization/components/accordion/index.gts
didoo: What do you mean? there is no equivalent `carbon/***` examples? in that case it's intentional, to simplify the showcasing, so it's easier to compare the different variants (in any case, there are no mock states for the Carbon components, so there would be nothing to compare with)

packages/components/src/styles/components/accordion.scss
didoo: excellent idea, done

packages/tokens/src/products/shared/accordion.json
didoo: removed `$modes` but kept the token for consistency with the other variant (card)

packages/tokens/src/products/shared/accordion.json
didoo: yeah, probably some leftover. nice catch

packages/tokens/src/products/shared/accordion.json
didoo: @LilithJames-HDS 👆

packages/tokens/src/products/shared/accordion.json
didoo: OK, I actually decided to simplify things and remove the tokens for the surface colors and use a combination of semantic surface colors (where needed, not many places) and `hds-apply-only-if-carbon`. Have a look and let me know what you think.

packages/components/src/styles/components/accordion.scss
didoo: I _think_ I've fixed this, have a look

didoo:
> * The left and right padding should match in all the variants, currently it seems that the right padding on all the variants matches the top and bottom values instead (except for in the Large variants)

@LilithJames-HDS ~~let me double check~~ OK, I've checked and is more complex than I realized. Let's discuss during our sync today

> * The .ItemButton in Figma is a unique component that shares a close appearance with the DismissButton whereas the one used here appears more like the Button secondary variant

@LilithJames-HDS can you be more specific? I checked and also in code is a custom element (not an Hds Button instance, I mean) and the different colors for the different states look the same as in Figma 🤔

packages/components/src/styles/components/accordion.scss
```diff
@@ -111,22 +129,27 @@
...
+  width: var(--hds-accordion-item-toggle-icon-size);
+  height: var(--hds-accordion-item-toggle-icon-size);
   padding: 0;
 
-  &:hover {
-    cursor: pointer;
+  // animate `chevron-down` icon
+  &.hds-accordion-item__button--is-open {
+    .hds-icon {
+      transform: rotate(var(--token-accordion-item-toggle-icon-transform-rotate));
```
dchyun, line 151: [todo] I think the tokens build needs to be re-run. Animation tokens coming up as undefined

Image

showcase/app/components/page-carbonization/components/accordion/index.gts
dchyun: This wasn't related to there being no carbon equivalent. I don't think the mock state is being applied on the carbon themed HDS component correctly.

The hover state for the carbon theme has a grey background
Screenshot 2026-06-11 at 9 52 11 AM

What I would expect to see for the `AccordionItem` mock hover state would be this
Screenshot 2026-06-11 at 9 52 20 AM

But right now it looks like this
Screenshot 2026-06-11 at 9 52 16 AM

packages/components/src/styles/components/accordion.scss
dchyun: Still seeing a gap in the non-flush nested examples. There may be some further nuances with this. Can discuss at the sync today.

Screenshot 2026-06-11 at 9 56 43 AM

packages/components/src/styles/components/accordion.scss
didoo: Damn! I cleanup up all the commits, and forgot to re-run the build steps. Check now

packages/components/src/styles/components/accordion.scss
didoo: (the CDS version also has a gap, that's why it's there)

dchyun: Yeah I just noticed that, so should the flush + flush nested accordion have that gap as well? Since it's visually the same as the card + card or mixed accordions for Carbon.

Screenshot 2026-06-11 at 10 32 14 AM

showcase/app/components/page-carbonization/components/accordion/index.gts
didoo: I think now is fixed (🤞)

PS: I've refactored a non trivial amount of code, so now something else may have blown up, please double check

didoo:
@dchyun this should be ready now for re-review

packages/components/src/styles/components/accordion.scss
```diff
@@ -198,15 +262,34 @@
   }
 }
 
-// Adjust content padding and border for nested flush accordions
+// NESTED ACCORDIONS
```
dchyun, line 265: [Issue?] For the nested flush + flush there is a design difference between it any every other nested combo. (The issue with the gap below it is fixed now.) For the flush + flush the outermost accordion has a full-width bottom border. In the other instances there is no bottom border for the outermost accordion.

Not sure which should be considered the design we should align them all too.

**Flush + flush**

Image

**Card + card & Mixed**

Image

dchyun (approved):
See one design issue, but approving if we want to address that in a follow up PR

packages/components/src/styles/components/accordion.scss
dchyun: If we want to align to Carbon we should follow the version with no border under the top accordion item, but a border top on the first nested item.

didoo: My understanding is that they have different borders depending on the type (flush or not):

screenshot_6552

I doubt it's intentional, so we could consider this a minor issue, that doesn't need to be fixed since our way of treating borders is at least consistent:

accordion

So, I am going to merge the PR, and if we find that we want to make changes, we can always have follow-up PRs.

PR #3649, opened 2026-02-25
Add Composite component

zamoore:
### :pushpin: Summary

If merged, this PR adds a new `HdsComposite` component that provides accessible keyboard navigation (roving tabindex) for composite widget patterns such as toolbars, listboxes, and 2D grids. This will be used as a foundational building block for keyboard navigation in the `AdvancedTable` component.

The Composite component implements the [WAI-ARIA Composite](https://www.w3.org/WAI/ARIA/apg/patterns/) keyboard navigation pattern using a modifier-based API. It yields three modifiers via a contextual hash:

- `{{c.composite}}` - Applied to the container element. Listens for keydown events and manages the composite's `tabindex` when no item is active.
- `{{c.item}}` - Applied to each navigable item. Manages roving `tabindex` (`0` for active, `1` for inactive), `data-active-item` attribute, and `aria-disabled`. Accepts an optional `disabled` named argument.
- `{{c.group}}` - Applied to row/group containers to enable 2D grid navigation.

#### Supported features:

- `@orientation` - `"horizontal"`, `"vertical"`, or omitted for both axes. Controls which arrow keys are active.
- `@loop` - `true` (all axes), `"horizontal"`, or `"vertical"`. Wraps focus from last to first (or vice versa) within the axis.
- `@wrap` - `true`, `"horizontal"`, or `"vertical"`. In 2D grids, overflows navigation into adjacent rows/columns.
- @defaultCurrentId` - Pre-selects a specific item or explicitly sets no active item (`null`).

##### 2D grid navigation
When groups are present and no `@orientation` is set, switches to 2D mode with `ArrowLeft`/`ArrowRight` within rows, `ArrowUp`/`ArrowDown` across rows, `Home`/`End` for row edges, `Ctrl+Home`/`Ctrl+End` for grid edges, and `PageUp`/`PageDown` for column edges.

### :link: External links

Jira ticket: [HDS-6131](https://hashicorp.atlassian.net/browse/HDS-6131)

***

<details>

</details>

[HDS-6131]: https://hashicorp.atlassian.net/browse/HDS-6131?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/wide-aliens-count.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+<!-- START components/composite -->
+`Composite` - Added new component that provides accessible keyboard navigation (roving tabindex) for composite widget patterns such as toolbars, listboxes, and 2D grids.
```
shleewhite:
```suggestion
`Composite` - Added new component that provides accessible arrow key keyboard navigation for complex components such as Tabs and AdvancedTable.
```

Just to make the changeset clearer on why were doing this for to consumers.

packages/components/src/components/hds/composite/index.gts
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    this._syncCompositeElement();
+  }
+
+  private _syncItemElement(item: HdsCompositeItem): void {
+    const element = item.element;
+    const isCurrent = item.id === this._currentId;
+
+    if (item.disabled === true) {
```
shleewhite, line 275: Suggestion: I think something could be improved about how this handles disabled items... 


aria-disabled doesn't make the item not tabbable or look greyed out.

I also noticed that in the showcase for interactive elements, you're kind of setting disabled twice:

```
<HdsButton {{c.item disabled=true}} disabled @text="Disabled" />
```

might be better to have composite be the source of truth here and make this function set both `aria-disabled` and the actual `disabled` attribute and then you don't need to also set it in the showcase example.

the reason we need both is that since we're not adding elements into the dom, we cant guarantee that `disabled` is valid on the element.... probably too complicated to check if it is valid on the element and only add the one we need...

showcase/app/router.ts
```diff
@@ -53,6 +53,7 @@ Router.map(function () {
     this.route('card');
     this.route('code-block');
     this.route('code-editor');
+    this.route('composite');
```
shleewhite: NIt: composite should probably fall under utilities

showcase/app/templates/index.gts
```diff
@@ -122,6 +122,11 @@ const Index: TemplateOnlyComponent = <template>
             CodeEditor
           </LinkTo>
         </li>
+        <li>
+          <LinkTo @route="page-components.composite">
```
shleewhite: Nit: composite should probably go under utilities

showcase/app/templates/page-components/composite.gts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import type { TemplateOnlyComponent } from '@ember/component/template-only';
+
+import CompositeIndex from 'showcase/components/page-components/composite';
```
shleewhite:
```suggestion
import CompositeIndex from 'showcase/components/page-utilities/composite';
```

shleewhite: Part of moving composite to utilities

showcase/app/components/page-utilities/composite/sub-sections/default-current-id.gts
```diff
@@ -0,0 +1,52 @@
...
+        @ariaLabel="Pre-selected"
+        @defaultCurrentId="preset-third"
+        @role="toolbar"
+      />
+    </SF.Item>
+  </ShwFlex>
+
+  <ShwDivider @level={{2}} />
+
+  <ShwTextH3>Nonexistent ID</ShwTextH3>
```
shleewhite, line 40: Suggestion: add another demo of a toolbar where the defaultCurrentId is set to a disabled item

packages/components/src/components/hds/composite/index.gts
```diff
@@ -0,0 +1,369 @@
...
+      element.removeAttribute('aria-disabled');
+    }
+
+    if (isCurrent === true) {
+      element.setAttribute('data-active-item', '');
+    } else {
+      element.removeAttribute('data-active-item');
+    }
+
+    element.setAttribute('tabindex', isCurrent === true ? '0' : '-1');
```
shleewhite, line 291: Issue: there is currently no safeguard to check that the defaultCurrentId item is not disabled. 

You should not be able to focus a disabled item & tabindex should never be set to 0.

showcase/app/styles/showcase-pages/composite.scss
```diff
@@ -0,0 +1,27 @@
+/**
+ * Copyright IBM Corp. 2021, 2025
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// COMPOSITE
+
+.composite-with-grid {
```
shleewhite: Nit: we generally wrap the page styles with `body.page-utilities-composite`

shleewhite: Nit: we also generally prefix any custom classes with `shw-utility-composite-`, not just `composite-`

packages/components/src/components/hds/composite/navigation.ts
```diff
@@ -0,0 +1,352 @@
+import type {
+  HdsCompositeItem,
+  HdsCompositeNavigationConfig,
+  HdsCompositeNavigationSnapshot,
+  HdsCompositeOrientations,
+} from './types';
```
shleewhite:
```suggestion
} from './types.ts';
```

packages/components/src/components/hds/composite/navigation.ts
```diff
@@ -0,0 +1,352 @@
...
+  }
+
+  const targetItems = getItemsInGroup(items, groups[targetGroupIndex]!.id);
+
+  return resolveColumnTarget(targetItems, columnIndex);
+}
+
+function navigateColumnEnd(
+  snapshot: HdsCompositeNavigationSnapshot,
+  config: HdsCompositeNavigationConfig,
```
shleewhite:
```suggestion
```

config isn't used in this function so you probably don't need it.

packages/components/src/components/hds/composite/index.gts
```diff
@@ -0,0 +1,369 @@
...
+    }
+
+    if (this._currentId === null || this._currentId === undefined) {
+      element.setAttribute('tabindex', '0');
+    } else {
+      element.removeAttribute('tabindex');
+    }
+  }
+
+  compositeModifier = modifier<HdsCompositeCompositeModifierSignature>(
```
shleewhite, line 310: Question: why would you add a keyboard listener to the composite container? You shouldn't need to arrow navigate from the container to the items inside.

Thinking that this could also cause issues in the rare scenario where you have arrow key navigation in one axis and then overflow in another direction. You'd want to set tabindex="0" on the container so the user could use their arrow keys to move the overflowed content. 

Currently, if I set tabindex="0" on the element with {{c.composite}} it gets removed silently.

showcase/tests/integration/components/hds/composite/index-test.gts
```diff
@@ -0,0 +1,592 @@
...
+        </HdsComposite>
+      </template>,
+    );
+
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-1');
+    assert.dom('#item-1').hasAttribute('tabindex', '0');
+    assert.dom('#item-2').hasAttribute('tabindex', '-1');
+  });
+
+  test('it skips disabled items', async function (assert) {
```
shleewhite:
```suggestion
  test('it sets aria-disabled on disabled items', async function (assert) {
```

this test isnt checking that it doesnt navigate to disabled items.

shleewhite (commented):
This is looking great!! Only issues I found were with edge cases. Also think this should live under the utilities section of the showcase instead of components.

packages/components/src/components/hds/composite/index.gts
zamoore: I updated the modifier to set both disabled and aria-disabled automatically. Browsers safely ignore the native disabled attribute on non-form elements anyway, so it's an easy fix. I've removed the redundant args from the showcase too.

packages/components/src/components/hds/composite/index.gts
zamoore: The container listener uses event delegation to catch events bubbling up from the active item, which avoids attaching listeners to every child. The `tabindex` removal is the standard roving focus pattern. If the overflow scroll edge case is a blocker, I can update the logic to preserve a manually set `tabindex`. Let me know.

packages/components/src/components/hds/composite/index.gts
zamoore: Good catch. I'll update the initialization logic to fall back to the first enabled item if `defaultCurrentId` points to a disabled element. I'll also add a safeguard in `_syncItemElement` to ensure `tabindex` always remains `-1` for disabled items.

showcase/app/components/page-utilities/composite/sub-sections/default-current-id.gts
zamoore: Done

showcase/tests/integration/components/hds/composite/index-test.gts
```diff
@@ -0,0 +1,592 @@
...
+    assert.dom('[data-active-item]').hasAttribute('id', 'B3');
+
+    await triggerKeyEvent('#composite-grid', 'keydown', 'ArrowLeft');
+    assert.dom('[data-active-item]').hasAttribute('id', 'B2');
+
+    await triggerKeyEvent('#composite-grid', 'keydown', 'ArrowUp');
+    assert.dom('[data-active-item]').hasAttribute('id', 'A3');
+  });
+
+  test('it does not activate any item if @defaultCurrentId does not match', async function (assert) {
```
dchyun: [Suggestion] There could be a test added for `@defaultCurrentId` to test it works correctly when not null and when matching one of the item ids

showcase/tests/integration/components/hds/composite/index-test.gts
```diff
@@ -0,0 +1,592 @@
...
+    await focus('#item-2');
+    await triggerKeyEvent('#composite-root', 'keydown', 'ArrowRight');
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-1');
+
+    await focus('#item-1');
+    await triggerKeyEvent('#composite-root', 'keydown', 'ArrowLeft');
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-2');
+  });
+
+  test('it loops only horizontally in a 2D grid when @loop="horizontal"', async function (assert) {
```
dchyun, line 155: [Suggesiton] Could add some tests for the `@wrap` argument

packages/components/src/components/hds/composite/index.gts
```diff
@@ -0,0 +1,375 @@
...
+    (element: HTMLElement) => {
+      this._compositeElement = element;
+      this._syncCompositeElement();
+
+      const onKeyDown = (event: KeyboardEvent): void => {
```
dchyun, line 317: [Suggestion] When I was recently working on the tab keyboard nav in the filter bar, I ran into [an issue](https://github.com/hashicorp/design-system/pull/3386#discussion_r2615595776) where the page was scrolling on `keydown` when I was navigating the tabs. 

To solve this I added a `preventDefault` in the keyboard handler. I notice the same page scroll is occuring in these examples. We could add a `preventDefault` somewhere here as well to prevent this.

dchyun (commented):
Looking really great! Love the abstraction.

packages/components/src/components/hds/composite/index.gts
zamoore: That makes sense to me. I made some adjustments and added a test to cover it.

didoo:
Sorry @zamoore I deleted this branch in `origin` instead of local 🤦‍♂️.
Probably you will need to re-open the PR as a new one.

zamoore:
@didoo No worries - I pushed and reopened

showcase/app/components/page-utilities/composite/code-fragments/with-grid.gts
```diff
@@ -0,0 +1,67 @@
...
+    loop?: HdsCompositeSignature['Args']['loop'];
+    orientation?: HdsCompositeSignature['Args']['orientation'];
+  };
+  Blocks: {
+    default: [];
+  };
+  Element: HTMLDivElement;
+}
+
+const CompositeWithHorizontalButtonList: TemplateOnlyComponent<CompositeWithGridSignature> =
```
shleewhite:
```suggestion
const CompositeWithGrid: TemplateOnlyComponent<CompositeWithGridSignature> =
```

showcase/app/components/page-utilities/composite/sub-sections/wrap.gts
```diff
@@ -0,0 +1,43 @@
...
+import CompositeWithGrid from '../code-fragments/with-grid';
+
+const SubSectionWrap: TemplateOnlyComponent = <template>
+  <ShwTextH2>Wrap (2D)</ShwTextH2>
+
+  <ShwTextH3>Horizontal wrap</ShwTextH3>
+
+  <ShwFlex @direction="column" as |SF|>
+    <SF.Item
+      @label="@wrap='horizontal' | true — ArrowRight at end of row wraps to first item of next row, ArrowLeft at start wraps to last item of previous row"
```
shleewhite, line 21: Suggestion: if im thinking about it, when I hear horizontal/vertical wrap in the grid I expect that wrapping keeps me in the same row/column instead of moving to the next one. it seems odd that it would move me to another row/column.

showcase/app/components/page-utilities/composite/sub-sections/default-current-id.gts
```diff
@@ -0,0 +1,86 @@
...
```
shleewhite, line 40: Issue: in this example, now you cant navigate to the elements at all. it should default to the first element instead.

showcase/tests/integration/components/hds/composite/index-test.gts
```diff
@@ -0,0 +1,840 @@
...
+        </HdsComposite>
+      </template>,
+    );
+
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-1');
+    assert.dom('#item-1').hasAttribute('tabindex', '0');
+    assert.dom('#item-2').hasAttribute('tabindex', '-1');
+  });
+
+  test('it sets aria-disabled on disabled items', async function (assert) {
```
shleewhite: Suggestion: this test should also check that the HTML disabled attr is added

packages/components/src/components/hds/composite/index.gts
shleewhite: Hmm... I think it would be best to preserve the tabindex if thats ok.

shleewhite (commented):
Looking great! Just a few small nits.

packages/components/src/components/hds/composite/index.gts
zamoore: Let me know what you think of the new approach

showcase/app/components/page-utilities/composite/sub-sections/wrap.gts
zamoore: What you're describing is actually handled by the `@loop` argument. I reserved `@wrap` for overflow navigation, where hitting the edge drops your focus to the next line.

showcase/tests/integration/components/hds/composite/index-test.gts
```diff
@@ -0,0 +1,866 @@
...
+
+    await focus('#item-1');
+    await triggerKeyEvent('#composite-root', 'keydown', 'ArrowDown');
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-2');
+
+    await triggerKeyEvent('#composite-root', 'keydown', 'ArrowUp');
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-1');
+  });
+
+  test('it wraps focus if @loop is true', async function (assert) {
```
dchyun: [Question] Would it be useful to add a test that the up/down arrow keys wrap for `@orientation="vertical"` and `@loop={{true}}`? Or is that logic covered by just testing the horizontal orientation?

showcase/tests/integration/components/hds/composite/index-test.gts
```diff
@@ -0,0 +1,866 @@
...
+        <HdsComposite @defaultCurrentId="item-2" as |c|>
+          <div {{c.composite}} id="composite-root">
+            <HdsButton {{c.item}} id="item-1" @text="Item 1" />
+            <HdsButton {{c.item}} id="item-2" @text="Item 2" />
+          </div>
+        </HdsComposite>
+      </template>,
+    );
+
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-2');
```
dchyun, line 476:
```suggestion
    assert.dom('[data-active-item]').hasAttribute('id', 'item-2');
    assert.dom('#item-2').hasAttribute('tabindex', '0');
```
[Nit] Could also test that tabindex is set

showcase/tests/integration/components/hds/composite/index-test.gts
```diff
@@ -0,0 +1,866 @@
...
+        <HdsComposite @defaultCurrentId="does-not-exist" as |c|>
+          <div {{c.composite}} id="composite-root">
+            <HdsButton {{c.item}} id="item-1" @text="Item 1" />
+            <HdsButton {{c.item}} id="item-2" @text="Item 2" />
+          </div>
+        </HdsComposite>
+      </template>,
+    );
+
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-1');
```
dchyun, line 494:
```suggestion
    assert.dom('[data-active-item]').hasAttribute('id', 'item-1');
    assert.dom('#item-1').hasAttribute('tabindex', '0');
```
[Nit] Could also test that tabindex is set

showcase/tests/integration/components/hds/composite/index-test.gts
```diff
@@ -0,0 +1,866 @@
...
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-3');
+
+    await triggerKeyEvent('#composite-root', 'keydown', 'ArrowLeft');
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-2');
+
+    await triggerKeyEvent('#composite-root', 'keydown', 'ArrowUp');
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-1');
+  });
+
+  test('horizontal orientation ignores vertical arrow keys', async function (assert) {
```
dchyun, line 527: [Suggestion] This orientation's arrow keys are already tested above in "it moves focus with arrow keys (horizontal)". You could remove this test and add a check in the other one that the up/down arrow keys are ignored.

showcase/tests/integration/components/hds/composite/index-test.gts
```diff
@@ -0,0 +1,866 @@
...
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-1');
+
+    await triggerKeyEvent('#composite-root', 'keydown', 'ArrowUp');
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-1');
+
+    await triggerKeyEvent('#composite-root', 'keydown', 'ArrowRight');
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-2');
+  });
+
+  test('vertical orientation ignores horizontal arrow keys', async function (assert) {
```
dchyun, line 552: [Suggestion] This orientation's arrow keys are already tested above in "it moves focus with arrow keys (vertical)". You could remove this test and add a check in the other one that the left/right arrow keys are ignored.

showcase/tests/integration/components/hds/composite/index-test.gts
```diff
@@ -0,0 +1,866 @@
...
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-1');
+
+    await triggerKeyEvent('#composite-root', 'keydown', 'ArrowRight');
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-1');
+
+    await triggerKeyEvent('#composite-root', 'keydown', 'ArrowDown');
+    assert.dom('[data-active-item]').hasAttribute('id', 'item-2');
+  });
+
+  test('it sets tabindex=0 on composite when no item is active', async function (assert) {
```
dchyun, line 577: [Question] So if `@defaultCurrentId={{null}}` then the fallback is no item is active and the composite is focused, but from an earlier test if `@defaultCurrentId="id-not-found"` then the first item is active and focused. 

I think I would expect the behavior to be the same for these two situations (first item is active). Was there a reason no item was set as active when this is `null` vs when it's `id-not-found`?

showcase/tests/integration/components/hds/composite/index-test.gts
```diff
@@ -0,0 +1,866 @@
...
+    );
+
+    await triggerKeyEvent('#composite-root', 'keydown', 'ArrowLeft');
+    assert.strictEqual(
+      document.activeElement && document.activeElement.id,
+      'item-1',
+    );
+  });
+
+  test('focus moves to the correct item after dynamic changes', async function (assert) {
```
dchyun: [Question] Is this the same test as "it updates active item when an item is removed"

showcase/tests/integration/components/hds/composite/index-test.gts
```diff
@@ -0,0 +1,866 @@
...
+    await focus('#item-2');
+    context.showSecond = false;
+    await settled();
+    assert.strictEqual(
+      document.activeElement && document.activeElement.id,
+      'item-1',
+    );
+  });
+
+  test('custom item IDs are handled correctly for navigation and active state', async function (assert) {
```
dchyun: [Suggestion] Since all the examples to this point have been using custom ids I don't think this test is needed.

dchyun (commented):
Functionality all looks great to me! Just some small things on the test cases.

showcase/app/components/page-utilities/composite/sub-sections/wrap.gts
shleewhite: ahhh ok I understand now!

showcase/tests/integration/components/hds/composite/index-test.gts
zamoore: No harm in adding a test

showcase/tests/integration/components/hds/composite/index-test.gts
zamoore: I think separating them out makes it more explicit what we're testing. Going to keep it this way unless you feel strongly about this.

showcase/tests/integration/components/hds/composite/index-test.gts
zamoore: I think separating them out makes it more explicit what we're testing. Going to keep it this way unless you feel strongly about this.

showcase/tests/integration/components/hds/composite/index-test.gts
zamoore: `null` is how you explicitly say that nothing should be active (overriding the fallback)

showcase/tests/integration/components/hds/composite/index-test.gts
zamoore: Yeah that makes sense. Removed.

shleewhite (dismissed):

.changeset/wide-aliens-count.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+<!-- START components/composite -->
```
dchyun:
```suggestion
<!-- START utilities/composite -->
```
[Question] Are the component docs for this going to live under "Utilities"? If so this path needs to be updated.

dchyun (dismissed):

.changeset/wide-aliens-count.md
zamoore: Ah yes, good call

dchyun (dismissed):

dchyun (approved):

PR #3685, opened 2026-03-09
[Project Solar / Phase 1 / Components] `Tag` carbonization

dchyun:
### :pushpin: Summary

This PR carbonizes the `Tag` component and adds component tokens for it to the tokens library.

[Tag carbonization page](https://hds-showcase-git-project-solar-phase-1dchyunta-f74d45-hashicorp.vercel.app/carbonization/components/tag)

Summary of changes:

- Refactored borders to support link specific border colors

### :link: External links

Jira ticket: [HDS-6107](https://hashicorp.atlassian.net/browse/HDS-6107)

***

<details>

</details>

[HDS-6107]: https://hashicorp.atlassian.net/browse/HDS-6107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/styles/components/tag.scss
```diff
@@ -39,24 +43,33 @@ $hds-tag-border-radius: 50px;
   height: 12px;
 }
 
-.hds-tag__dismiss-icon,
+.hds-tag__dismiss,
 .hds-tag__text {
-  color: var(--token-color-foreground-primary);
+  color: var(--token-tag-foreground-color-default);
```
jorytindall: I think this is causing the icon in the dismiss button to match the text, which is not how the `primary` variant appears now. So the `x` in the dismiss button is matching the color of the text.

**Before**
Image

**After**
Image

jorytindall (commented):
I've matched up a couple of things in the Figma component (mainly the focus ring), I think this is looking solid, just left one note about the color of the icon in the dismiss button that is being inherited

showcase/app/components/page-carbonization/components/tag/index.gts
```diff
@@ -0,0 +1,149 @@
...
+          {{else}}
+            <pre>TODO: add static image here</pre>
+          {{/if}}
+        </:reference>
+      </ShwCarbonizationComparisonGrid>
+    {{/each}}
+
+    <ShwDivider @level={{2}} />
+
+    <ShwTextH2>Link Colors</ShwTextH2>
```
KristinLBradley:
```suggestion
    <ShwTextH2>Link color states</ShwTextH2>
```

KristinLBradley: I know this was copied from the original Tag Showcase but perhaps this suggested edit is a bit clearer/more accurate?

KristinLBradley: Or else change the `ShwTextH2` to `ShwTextH3` and the following headers to `ShwTextH4` so the hierchy puts them under the above "States" header. (Then add a ShwTextH3 for the examples under "States" making it clearer how those states relate to the tag vs. the "Link colors" ones.)

packages/components/src/styles/components/tag.scss
KristinLBradley: Looks like this style was there before (now gone in this PR) that was controlling the svg icon color:

```
.hds-tag__dismiss-icon, .hds-tag__text {
}
```

packages/components/src/styles/components/tag.scss
```diff
@@ -39,24 +43,33 @@ $hds-tag-border-radius: 50px;
   height: 12px;
 }
 
-.hds-tag__dismiss-icon,
```
KristinLBradley, line 42: Deleting this selector is what changed the SVG x icon color from before.

packages/components/src/styles/components/tag.scss
dchyun: This has been fixed now. I added in some icon color tokens like there are in Figma. I had to separate the link primary and secondary colors for the icon from the plain text color. So I have three new variables

- `tag.icon.color`
- `tag.link.icon.color.primary`
- `tag.link.icon.color.secondary`

showcase/app/components/page-carbonization/components/tag/index.gts
dchyun: I reorganized the whole tag page to make the separation between the content and the state examples more clear for both links and plain text.

packages/tokens/src/products/shared/tag.json
dchyun: [Note] I added a letter spacing token, because without it the cut off for long text would differ between Carbonized HDS and the Carbon component.

**Carbonized HDS w/out letter spacing token**

Image

**Carbon**

Image

didoo: Personally, I would leave the cutoff different (what is the issue)? I think we're overthinking this letter-spacing thing. For the end-user is not perceivable, it should be considered in the context of a wider, more complex UI.

heatherlarsen: +1 to what @didoo noted. I don't see the letter-spacing difference being unique enough to warrant the new token(s).

dchyun: Removed the letter spacing token

jorytindall (approved):
Looks good! Noting here for history: it isn't possible to assign the underline property in Figma to a variable, so for the link variants, the underline will persist in the Figma component.

showcase/app/components/page-carbonization/components/tag/index.gts
```diff
@@ -0,0 +1,246 @@
...
+                    viewBox="0 0 16 16"
+                    aria-hidden="true"
+                  ><path
+                      d="M7.2,2.3c-1,4.4,1.7,8.7,6.1,9.8c0.1,0,0.1,0,0.2,0c-1.1,1.2-2.7,1.8-4.3,1.8c-0.1,0-0.2,0-0.2,0C5.6,13.8,3,11,3.2,7.7	C3.2,5.3,4.8,3.1,7.2,2.3 M8,1L8,1C4.1,1.6,1.5,5.3,2.1,9.1c0.6,3.3,3.4,5.8,6.8,5.9c0.1,0,0.2,0,0.3,0c2.3,0,4.4-1.1,5.8-3	c0.2-0.2,0.1-0.6-0.1-0.7c-0.1-0.1-0.2-0.1-0.3-0.1c-3.9-0.3-6.7-3.8-6.4-7.6C8.3,3,8.4,2.4,8.6,1.8c0.1-0.3,0-0.6-0.3-0.7	C8.1,1,8.1,1,8,1z"
+                    ></path></svg>
+                </cds-operational-tag>
+              </SF.Item>
+            </ShwFlex>
+          {{else}}
+            <pre>n/a</pre>
```
didoo: [question] should we use the `NoEquivalent` placeholder?

showcase/app/components/page-carbonization/components/tag/index.gts
```diff
@@ -0,0 +1,246 @@
...
+            @onDismiss={{NOOP}}
+            mock-state-value={{if (eq state "default") undefined state}}
+            mock-state-selector="button"
+          />
+        </:theming>
+        <:reference>
+          {{#if (eq state "default")}}
+            <cds-dismissible-tag text="Lorem ipsum" />
+          {{else}}
+            <pre>TODO: add static image here</pre>
```
didoo:
```suggestion
            <pre>TODO: static image here</pre>
```

didoo: I've just realized the text is too long in some conditions, let's make it shorter everywhere

showcase/app/components/page-carbonization/components/tag/index.gts
```diff
@@ -0,0 +1,246 @@
...
+                  viewBox="0 0 16 16"
+                  aria-hidden="true"
+                ><path
+                    d="M7.2,2.3c-1,4.4,1.7,8.7,6.1,9.8c0.1,0,0.1,0,0.2,0c-1.1,1.2-2.7,1.8-4.3,1.8c-0.1,0-0.2,0-0.2,0C5.6,13.8,3,11,3.2,7.7	C3.2,5.3,4.8,3.1,7.2,2.3 M8,1L8,1C4.1,1.6,1.5,5.3,2.1,9.1c0.6,3.3,3.4,5.8,6.8,5.9c0.1,0,0.2,0,0.3,0c2.3,0,4.4-1.1,5.8-3	c0.2-0.2,0.1-0.6-0.1-0.7c-0.1-0.1-0.2-0.1-0.3-0.1c-3.9-0.3-6.7-3.8-6.4-7.6C8.3,3,8.4,2.4,8.6,1.8c0.1-0.3,0-0.6-0.3-0.7	C8.1,1,8.1,1,8,1z"
+                  ></path></svg>
+              </cds-operational-tag>
+            {{else if (eq color "secondary")}}
+              <pre>n/a</pre>
+            {{else}}
+              <pre>TODO: add static image here</pre>
```
didoo:
```suggestion
              <pre>TODO: static image here</pre>
```

showcase/app/components/page-carbonization/components/tag/index.gts
```diff
@@ -0,0 +1,246 @@
...
+            <pre>TODO: add static image here</pre>
+          {{/if}}
+        </:reference>
+      </ShwCarbonizationComparisonGrid>
+    {{/each}}
+
+    <ShwTextH3>Links</ShwTextH3>
+
+    {{#each COLORS as |color|}}
+      <ShwTextH3>{{capitalize color}}</ShwTextH3>
```
didoo:
```suggestion
      <ShwTextH4>{{capitalize color}}</ShwTextH4>
```

showcase/app/components/page-carbonization/components/tag/index.gts
```diff
@@ -0,0 +1,246 @@
...
+ */
+
+import type { TemplateOnlyComponent } from '@ember/component/template-only';
+import { pageTitle } from 'ember-page-title';
+import { eq, notEq, and } from 'ember-truth-helpers';
+import { capitalize } from '@ember/string';
+
+import ShwTextH1 from 'showcase/components/shw/text/h1';
+import ShwTextH2 from 'showcase/components/shw/text/h2';
+import ShwTextH3 from 'showcase/components/shw/text/h3';
```
didoo, line 13:
```suggestion
import ShwTextH3 from 'showcase/components/shw/text/h3';
import ShwTextH4 from 'showcase/components/shw/text/h4';
```

showcase/app/components/page-carbonization/components/tag/index.gts
```diff
@@ -0,0 +1,246 @@
...
+              />
+            </SF.Item>
+          </ShwFlex>
+        </:theming>
+        <:reference>
+          {{#if (notEq color "secondary")}}
+            <ShwFlex @direction="column" as |SF|>
+              <SF.Item>
+                <cds-operational-tag type="blue" text="Lorem ipsum" size="md">
+                  <svg
```
didoo: [todo] we don't need the icon here

showcase/app/components/page-carbonization/components/tag/index.gts
```diff
@@ -0,0 +1,246 @@
...
+                    ></path></svg>
+                </cds-operational-tag>
+              </SF.Item>
+              <SF.Item>
+                <cds-operational-tag
+                  type="blue"
+                  text="This is a very long text that should go on multiple lines"
+                  size="md"
+                >
+                  <svg
```
didoo: [todo] we don't need the icon here

showcase/app/index.html
```diff
@@ -62,6 +62,7 @@
     <script type="module" src="https://1.www.s81c.com/common/carbon/web-components/version/v2.47.0/link.min.js"></script>
     <script type="module"
       src="https://1.www.s81c.com/common/carbon/web-components/version/v2.48.0/modal.min.js"></script>
+    <script type="module" src="https://1.www.s81c.com/common/carbon/web-components/version/v2.48.0/tag.min.js"></script>
```
didoo, line 68: [question] have you tried to play with different version, to see if the visual bugs in the Carbon Web Components are fixed?

packages/components/src/styles/components/tag.scss
```diff
@@ -8,26 +8,30 @@
...
   max-width: 100%;
   vertical-align: middle;
-  background-color: var(--token-color-surface-interactive);
+  background-color: var(--token-tag-surface-color-default);
   border: 1px solid var(--token-color-border-strong);
-  border-radius: $hds-tag-border-radius;
+  border-radius: var(--token-tag-border-radius);
+
+  @include hds-apply-only-if-carbon() {
+    &:not(.hds-tag--color-primary, .hds-tag--color-secondary) {
```
didoo: [thought] instead of using this negation (always tricky) what if we add a new class to the `Tag` to determine when is a "simple/basic/text" tag, and when instead contains a link? in this way here we can less indirection for when this rule applies.

packages/components/src/styles/components/tag.scss
```diff
@@ -112,23 +126,50 @@ $hds-tag-border-radius: 50px;
 // COLORS (FOR LINK)
 
 .hds-tag--color-primary {
+  border: 1px solid var(--token-tag-link-border-color-primary);
```
didoo: [question] should we override only the color, not everything?

packages/components/src/styles/components/tag.scss
```diff
@@ -112,23 +126,50 @@ $hds-tag-border-radius: 50px;
...
     }
   }
+
+  .hds-tag__dismiss-icon {
+    color: var(--token-tag-link-icon-color-primary);
+  }
 }
 
 .hds-tag--color-secondary {
+  border: 1px solid var(--token-tag-link-border-color-secondary);
```
didoo: [question] should we override only the color, not everything?

didoo (commented):
@dchyun I have left a bunch of comments, but nothing major. I understand there are no 1-to-1 mappings for all the variants/combinations of this component, so everything is approximation anyway. So when you feel confident it's OK let me know and I can approve

showcase/app/components/page-carbonization/components/tag/index.gts
dchyun: Went ahead and edited all of the other carbonization files to use this shorter language as well

showcase/app/components/page-carbonization/components/tag/index.gts
dchyun: the `cds-operational-tag` isn't designed to be used without an svg. The padding is off without it, so have left it in for that reason

Screenshot 2026-03-31 at 11 21 30 AM

showcase/app/components/page-carbonization/components/tag/index.gts
dchyun: See above note about the icon

showcase/app/index.html
dchyun: I've tried playing around with the version and nothing worked.

Also a general note, no CDN for a version above `2.50` appears to be available even though the carbon docs all mentioned `2.51.1` as the latest. Always get a 404 when using those versions.

packages/components/src/styles/components/tag.scss
dchyun: I've added two classes `hds-tag--is-static` and `hds-tag--is-interactive` and leveraged those in the styles and added test cases for them.

packages/components/src/styles/components/tag.scss
dchyun: Updated to use `border-color`

showcase/app/components/page-carbonization/components/tag/index.gts
didoo: > the cds-operational-tag isn't designed to be used without an svg

Where did you get this information? (I looked for this in the Carbon website docs, but I couldn't find anything related to this). What I found is that in their own documentation ([here](https://carbondesignsystem.com/components/tag/style/#color) and [here](https://carbondesignsystem.com/components/tag/usage/#operational-tag)) the visual examples they show is _without_ icon:
screenshot_6082
tag-usage-operational

Plus, in [the React version](https://react.carbondesignsystem.com/?path=/story/components-tag--operational) if you remove the icon from the DOM, the padding is still correct:
screenshot_6084

Same in the Web Component version:
image

I suspect the issue is with the CSS in the version we're using (I checked before and some generated CSS selectors are all mangled). 

Maybe something to raise/discuss in our internal channel with the Carbon team?

showcase/app/index.html
didoo: Same as above, maybe something to ask/clarify in the internal Slack channel with the Carbon team?

packages/components/src/components/hds/tag/index.gts
```diff
@@ -138,13 +138,21 @@ export default class HdsTag extends Component<HdsTagSignature> {
...
+    const { href, route } = this.args;
+
     const classes = ['hds-tag'];
 
     // add a class based on the @color argument
     if (this.color) {
       classes.push(`hds-tag--color-${this.color}`);
     }
 
+    if (href || route) {
```
didoo, line 150: ❤️

showcase/app/index.html
dchyun: Have asked the Carbon team about this. Did a lot of digging on the cdn yesterday. There are two paths they seem to use. Either `carbon/web-components/version/vX.X.X` or `carbon/web-components/tag/v2/latest`. I've tried both ways and neither fixed the styling issues. Searched through the `tag.min.js` files and the `border-color` style we need doesn't seem to be present so not sure where its coming from on the storybook side.

showcase/app/components/page-carbonization/components/tag/index.gts
dchyun: After investigating this I believe its actually an issue with the Carbon component. Both their web component and react storybooks only show examples with the icon, but their docs say otherwise as you showed.

In the web component, the `cds-operational-tag` template always adds the `icon` slot to the internal `cds-tag` even if no icon is passed in by the consumer. 
```
<cds-tag
  part="tag"
  ?aria-pressed="${selected}"
  size="${size}"
  ?disabled="${disabled}"
  type="${type}">
  <slot name="icon" slot="icon"></slot>
  ${text}
  <slot name="decorator" slot="decorator"></slot>
  <slot name="ai-label" slot="ai-label"></slot>
  <slot name="slug" slot="slug"></slot>
</cds-tag>
```
This means the `has-custom-icon` attribute is always set on the internal tag, even if there is no icon in the slot. This attribute's presence is preventing the right padding from being applied.
Screenshot 2026-04-01 at 9 40 31 AM

I've asked them about these things and we shall see what they say.

didoo: So probably for our showcase we can remove the icon and "ignore" the small visual misalignment, what do you think?

dchyun: I can remove it for now. I suspect they will say that the icon can be removed and this is a bug. But if they say it can't then I'll add it back in.

dchyun: Icons have been removed 🪓

showcase/app/components/page-carbonization/utilities/dismiss-button/index.gts
```diff
@@ -53,7 +53,7 @@ const DismissButtonCarbonizationIndex: TemplateOnlyComponent = <template>
         <HdsDismissButton mock-state-value="hover" />
       </:theming>
       <:reference>
-        <pre>TODO: add static image here</pre>
+        <pre>TODO: static image here</pre>
```
KristinLBradley, line 56: Do we plan to add static images at a later stage?

dchyun: Yeah all the carbonization prs use this placeholder right now. I think we'll add images at a later time.

packages/tokens/src/products/shared/tag.json
```diff
@@ -0,0 +1,304 @@
...
+        "$modes": {
+          "default": 1.2308,
+          "cds-g0": "{carbon.typography.label01.line-height}",
+          "cds-g10": "{carbon.typography.label01.line-height}",
+          "cds-g90": "{carbon.typography.label01.line-height}",
+          "cds-g100": "{carbon.typography.label01.line-height}"
+        }
+      }
+    },
+    "icon": {
```
didoo: [thought] now that we've introduced the concept of "static" tag vs "link" tag, would make sense to have this as `tag.static.icon.color` for "simmetry" with the `tag.link.icon.color` one?

didoo: and at that point, also `tag.foreground.color` and `tag.surface.color` should live under `static`? (I haven't checked the semantic relation and usage in relation to the "link" tags, only an observation/question that came to mind)

packages/components/src/styles/components/tag.scss
```diff
@@ -37,26 +41,38 @@ $hds-tag-border-radius: 50px;
...
+    padding-bottom: calc(var(--token-tag-padding-bottom) - 1px); // account for 1px border on link variant
+    text-decoration: none;
+  }
+}
+
 .hds-tag__text-container {
   display: -webkit-box;
   overflow: hidden;
-  line-height: 1rem; // 16px - override `body-100`
+  font-weight: var(--token-typography-font-weight-regular);
```
didoo: [issue?] in production this text is "medium"; is this change intentional (eg. agreed with design)?

Image

packages/components/src/styles/components/tag.scss
```diff
@@ -8,26 +8,30 @@
...
   vertical-align: middle;
-  background-color: var(--token-color-surface-interactive);
+  background-color: var(--token-tag-surface-color-default);
   border: 1px solid var(--token-color-border-strong);
-  border-radius: $hds-tag-border-radius;
+  border-radius: var(--token-tag-border-radius);
+
+  @include hds-apply-only-if-carbon() {
+    &.hds-tag--is-static {
+      border: none;
```
didoo: [suggestion] as discussed, consider making the border transparent instead of removing completely, and decrease the padding tokens by 1px, so you can remove the custom padding for the `.hds-tag__link` (lines 64-65); for me is a neater solution, with less overrides and less layout differences (the visual space remains the same)

packages/tokens/src/products/shared/tag.json
dchyun: Reorganized all the base-level color tokens from `tag.{foreground/surface/etc}` to `tag.static.{foreground/surface/etc}`. Also re-organized the style selectors a bit so these static tokens weren't applied to the link unnecessarily.

packages/components/src/styles/components/tag.scss
dchyun: Thanks for catching this. I've added a font weight token to account for this.

packages/components/src/styles/components/tag.scss
dchyun: Updated to follow this approach. One note, had to adjust the padding of the dismiss button as well for carbon themes to keep the heights at `24px`.

```
  padding-top: 5px;
  padding-bottom: 5px;
}
```

packages/tokens/src/products/shared/tag.json
dchyun: Note: There's a bit of a nuance here as the static and secondary link variants share surface and foreground colors. Previously both inherited from the base level tokens. I've added `tag.link.secondary.surface/foreground` tokens to keep the `static/link` organization even though they are identical to the `tag.link.static.surface/foreground` tokens.

KristinLBradley:
The Percy diffs appear to be just from the issue with Icons not rendering so if it looks ok to you, you could mark it as approved to clear the failing check

packages/components/src/styles/components/tag.scss
```diff
@@ -8,55 +8,72 @@
...
   margin: 0; // reset default button margin
   padding: 6px 4px 6px 8px;
-  background-color: var(--token-color-surface-interactive);
+  background-color: var(--token-tag-static-surface-color-default);
   border: none; // reset default button border
   border-radius: inherit;
   border-top-right-radius: 0;
   border-bottom-right-radius: 0;
+
+  @include hds-apply-only-if-carbon() {
```
didoo: [note] I don’t think you need this? the parent is a flexbox, and even if you change the vertical padding to 2px, the height of the dismiss button remains the same; which for me it means the value of the vertical padding is not theme dependant

(please check, I tested in the browser, maybe I missed something?)

packages/components/src/styles/components/tag.scss
```diff
@@ -8,55 +8,72 @@
...
+  border-width: 1px;
+  border-radius: var(--token-tag-border-radius);
 }
 
 .hds-tag__dismiss {
   flex: 0 0 auto;
   margin: 0; // reset default button margin
   padding: 6px 4px 6px 8px;
-  background-color: var(--token-color-surface-interactive);
+  background-color: var(--token-tag-static-surface-color-default);
```
didoo: now that colors are differentiated by "static" and "link" should this declaration be moved under the `.hds-tag--is-static` block?

packages/components/src/styles/components/tag.scss
```diff
@@ -8,55 +8,72 @@
...
+
+.hds-tag--is-static {
+  background-color: var(--token-tag-static-surface-color-default);
+  border-color: var(--token-tag-static-border-color);
 }
 
 .hds-tag__dismiss-icon {
   width: 12px;
   height: 12px;
+  color: var(--token-tag-static-icon-color);
```
didoo: try to change this to `red` and you will see it gets applied only to the "static" version; this means for me that this color declaration should be moved under the `.hds-tag--is-static` block

Image

packages/components/src/styles/components/tag.scss
```diff
@@ -8,55 +8,72 @@
...
 }
 
-.hds-tag__dismiss-icon,
+.hds-tag__dismiss,
```
didoo: [todo] since the icon color is applied via `.hds-tag__dismiss-icon` we don't need this
```suggestion
```

packages/components/src/styles/components/tag.scss
```diff
@@ -8,55 +8,72 @@
...
 .hds-tag__text {
-  color: var(--token-color-foreground-primary);
+  color: var(--token-tag-static-foreground-color-default);
```
didoo: try to change this to `green` and you will see it gets applied only to the "static" version; this means for me that this color declaration should be moved under the `.hds-tag--is-static` block

Image

packages/components/src/styles/components/tag.scss
```diff
@@ -77,20 +94,22 @@ $hds-tag-border-radius: 50px;
...
   cursor: pointer;
 
+  &:focus,
+  &.mock-focus {
+    @include hds-focus-ring-basic();
+    z-index: 1; // ensures focus is not obscured by adjacent elements
+  }
+}
+
+.hds-tag__dismiss {
```
didoo: probably these interactive states color declarations should be moved under the `.hds-tag--is-static` block too

packages/components/src/styles/components/tag.scss
```diff
@@ -8,55 +8,72 @@
...
   border-top-right-radius: 0;
   border-bottom-right-radius: 0;
+
+  @include hds-apply-only-if-carbon() {
+    padding-top: 5px;
+    padding-bottom: 5px;
+  }
+}
+
+.hds-tag--is-static {
```
didoo: [suggestion] consider moving this block under the `// COLORS` comment (and remove the `(for link)` part)

didoo (commented):
A few comments about re-organizing the CSS code

packages/components/src/styles/components/tag.scss
```diff
@@ -109,26 +99,97 @@ $hds-tag-border-radius: 50px;
   }
 }
 
-// COLORS (FOR LINK)
+// STATIC COLORS
+
+.hds-tag--is-static {
```
didoo, line 105: [praise] so neat now! 🙌

packages/components/src/styles/components/tag.scss
```diff
@@ -65,8 +65,8 @@ $hds-tag-border-radius: 50px;
 
 .hds-tag__dismiss ~ .hds-tag__text,
 .hds-tag__dismiss ~ .hds-tag__link {
-  max-width: 160px;
-  padding: 3px 8px 5px 6px;
+  max-width: calc(var(--token-tag-max-width) - 6px);
```
didoo: [question] I just noticed now, where is this `6px` coming from? it's not the dismiss button width, so I am a bit confused 🤔

packages/components/src/styles/components/tag.scss
```diff
@@ -109,26 +99,97 @@ $hds-tag-border-radius: 50px;
...
+  .hds-tag__dismiss-icon {
+    color: var(--token-tag-link-icon-color-primary);
+  }
 }
 
 .hds-tag--color-secondary {
+  border-color: var(--token-tag-link-border-color-secondary);
+
+  .hds-tag__link,
+  &:has(.hds-tag__link) .hds-tag__dismiss {
```
didoo: [suggestion] I _think_ now we know for sure that the primary/secondary tags have `hds-tag__link`, no?
```suggestion
  .hds-tag__dismiss {
```

packages/components/src/styles/components/tag.scss
```diff
@@ -109,26 +99,97 @@ $hds-tag-border-radius: 50px;
...
+  }
+}
+
+// LINK COLORS
 
 .hds-tag--color-primary {
+  border-color: var(--token-tag-link-border-color-primary);
+
+  .hds-tag__link,
+  &:has(.hds-tag__link) .hds-tag__dismiss {
```
didoo: [suggestion] I _think_ now we know for sure that the primary/secondary tags have `hds-tag__link`, no?
```suggestion
  .hds-tag__dismiss {
```

didoo (commented):
left a couple of small suggestions and a question

packages/components/src/styles/components/tag.scss
dchyun: A bit going on with this. So the `max-width` is for cutting off the text. The cutoff needs to be the same length of characters for w/ and w/out the dismiss. The padding of the text is different depending on the presence of the dismiss so that needs to be accounted for in the `max-width`. 

Screenshot 2026-04-10 at 12 39 08 PM

Originally that padding diff was captured in the hard coded values, one was `160px` and one was `166px`. Now that the max-width is tokenized I accounted for that diff with the calc.

Problem this made me realize now is the padding diff between w/ and w/out dismiss also now changes for carbon, so this can't be left at 6px anymore actually. 

I could do a complicated calc with the values for the max width, left/right padding with dismiss, left/right padding w/out dismiss to determine this diff value, but it seemed easier to just make a token `dismissible.max-width` and account for the difference there.

So all that to say this value is now tokenized. `max-width: var(--token-tag-dismissible-max-width);`

didoo (approved):
It was a lot of work, but the final result is 👩‍🍳 💋 👌

PR #3767, opened 2026-03-31
Carbon icon mapping improvements

zamoore:
### :pushpin: Summary

If merged, this PR will change the source of the flight to carbon icon mapping from the `hds-carbon-icon-map.json` file (now deleted) to a single source of truth, the `catalog.json` file.

Also updates incorrect icons to correct ones (icons added and deleted).

### :link: External links

Jira ticket: [HDS-6112](https://hashicorp.atlassian.net/browse/HDS-6112)

[HDS-6112]: https://hashicorp.atlassian.net/browse/HDS-6112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


didoo:
@zamoore I see these errors when I run `pnpm build` in `packages/flight-icons`, is it expected?
screenshot_6092

zamoore:
@didoo - Yes, those warnings are intentional. `sparkle` is intentionally skipped do the `Do not use` in the mapping string. If that's too noisy, we can hide those.

For the other 3, I need to discuss with Jory. Seems like the mapping might be slightly off. There are some icons that aren't present in the `carbon` package at the 32px size (or any specific size), and some that just aren't present in the package at all. I'm not certain why that is.

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
```diff
@@ -93,9 +85,19 @@ export async function generateBundleSymbolJS({ config, catalog }: { config: Conf
             registry[baseName].flight[size] = `() => import('./flight/${fileName}.js')`;
 
             // --- CARBON ---
-            const carbonName = mapping[baseName];
 
-            if (carbonName && !registry[baseName].carbon) {
+            const carbonMatch = mapping?.match(/Maps to: (.+)/);
```
didoo: [thought] @zamoore @jorytindall I am bit worried that this format may lead to people manually editing the description field

Image

What if we instead make it more machine-readable and make people more wary about editing it? In the end, I don't think consumers will ever use this in any way.

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
```diff
@@ -93,9 +85,19 @@ export async function generateBundleSymbolJS({ config, catalog }: { config: Conf
...
+
+            if (carbonMatch != null && !registry[baseName].carbon) {
```
didoo: [suggestion] add a comment why the `!registry[baseName].carbon` check is needed (I had to ask Copilot, it was not obvious to me)

didoo: probably I would move it inside the `if` statement, as an early `continue` condition, so it's more clear the intent (and the comment is more specific)

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
```diff
@@ -93,9 +85,19 @@ export async function generateBundleSymbolJS({ config, catalog }: { config: Conf
...
+                const carbonName = carbonMatch[1].toLowerCase();
+
+                const doNotUse = carbonName.match(/\(Do not use\)/i);
+
+                if (doNotUse !== null) {
+                    console.warn(`⚠️ Skipping Carbon mapping for Flight icon ${fileName} due to "(Do not use)" note in mapping: ${mapping}`);
```
didoo: @jorytindall what does "Do not use" mean? (as far as I can see, there's only one icon with this, the `sparkle` one.

didoo (commented):
left a few comments

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
jorytindall: > manually editing the description field

Can you clarify what you mean by this? As in, people from our team manually editing this field? Personally I think that's pretty low risk since we all more or less know what is going on with this library. However, part of the purpose of including _something_ in the description was the make sure that the coverage was 100%, so if we deem this not valuable for consumers then I'm fine stripping it out or including something more machine-oriented.

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
jorytindall: Yeah there's only one icon in the Carbon library that has this (understanding this is a super edge case) and as far as I know it's because the Carbon components have specific AI icons and visual elements built into them (they have arguments is `isAI` or something like that). For that reason I don't think they want people using this icon directly, but for our case it was OK to include this mapping.

didoo: @jorytindall yes, but why not leave it as is, saying there is no mapping? Why the single exception (that then needs to be handled, in code as well as in Figma I suppose)

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
didoo: Let's say we have a new icon or something, and someone adds manually the description field using `Map to: ***` or `Mapping to: ***` instead of `Maps to: ***` because a typo or because they forgot what is the right™ format to use, the script would miss that entry. if it was something more machine oriented like `[carbon:***]` maybe it would be more obvious that it's meant for a machine, is not just human-readable text.

jorytindall: Yeah I think updating to something like that makes sense at this point in the process 👍

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
jorytindall: I'm trying to remember back exactly to what my thought process was here; there _is_ a mapping for the HDS `sparkle` icon to the Carbon/IBM `AI-label` icon, and it's mapped properly in the Figma library. That being said, I don't think this is necessary

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
zamoore: I discussed with @jorytindall and we opted for `[carbon:icon-name]` when available, and ``[carbon:]` when not

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
zamoore: Makes sense

packages/flight-icons/catalog.json
```diff
@@ -10,7 +10,7 @@
       "fileName": "loading-24",
       "iconName": "loading",
       "description": "loading, loader, spinner, animated",
-      "mapping": "No Carbon equivalent.\nTBD",
+      "mapping": "",
```
jorytindall: I'm not the one to make the decision on this, but just confirming that it's better/easier to have an empty string compared to being explicit with a fallback for icons that aren't mapped. I don't have a strong opinion, but if this field is empty in the catalog should it be excluded?

jorytindall (dismissed):
I'm a bit out of my depth here, but cursory examination looks good to me. Left a small point to discuss on the generated catalog, but it's not a blocker, I'm fine either way.

packages/flight-icons/catalog.json
zamoore: I think keeping an empty mapping field is a good explicit signal that it hasn't been mapped.

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
```diff
@@ -17,6 +17,14 @@ const prettierConfig = {
     trailingComma: 'none'
 } as const;
 
+// Some icons don't follow the same mapping convention, so we need to hardcode their paths here
+// NOTE: if this list grows we should consider moving it to a separate config file, but for now it's small enough to keep it here
+// { [iconName]: carbonIconPathWithoutExtension }
```
didoo: [question] is this commented line still relevant? 🤔
```suggestion
```

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
```diff
@@ -93,25 +94,37 @@ export async function generateBundleSymbolJS({ config, catalog }: { config: Conf
...
 
             // --- CARBON ---
-            const carbonName = mapping[baseName];
 
-            if (carbonName && !registry[baseName].carbon) {
-                const carbonPath = path.join(carbonIconsPath, '32', `${carbonName}.svg`);
+            // Proceed only if the tag exists
+            if (mapping) {
+                if (registry[baseName].carbon) {
+                    continue; // Carbon icon already processed for this baseName
```
didoo:
```suggestion
                    continue; // Carbon icon already processed for this baseName (e.g. for another size)
```

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
```diff
@@ -93,25 +94,37 @@ export async function generateBundleSymbolJS({ config, catalog }: { config: Conf
...
+                }
+                const carbonName = mapping.toLowerCase();
+                let carbonPath = path.join(carbonIconsPath, '32', `${carbonName}.svg`);
+
+                const overrideCarbonPath = CARBON_ICON_OVERRIDES[carbonName];
+                if (overrideCarbonPath !== undefined) {
+                    carbonPath = path.join(carbonIconsPath, `${overrideCarbonPath}.svg`);
+                }
```
didoo: [suggestion] maybe? (not feeling strong about it)
```suggestion
                let carbonPath;
                if (Object.keys(CARBON_ICON_OVERRIDES).includes(carbonName)) {
                    carbonPath = path.join(carbonIconsPath, `${CARBON_ICON_OVERRIDES[carbonName]}.svg`);                    
                } else {
                    carbonPath = path.join(carbonIconsPath, '32', `${carbonName}.svg`);
                }
```

packages/flight-icons/catalog.json
didoo: with the current implementation (`const mapping = mappingTagMatch?.[1]?.trim();`) if there is no `[carbon:***]` entry in the description field, the value of `mapping` is `undefined` so if the convention is that there should be an empty string, a fallback should be provided.

that said, I lean more on @jorytindall's view that if there is no mapping, the `mapping` field should be excluded (more explicit, for me, than an empty string). 

unless we want to differentiate even further:
- `[carbon:foo-bar-baz]` in the Figma description → `"mapping": "foo-bar-baz"`
- `[carbon:]` in the Figma description → `"mapping": ""`
- no `[carbon:]` entry in the Figma description → no `mapping` entry

didoo (commented):
I have left a couple of small suggestions, nothing major.

What concerns me the most is seeing diffs in this PR for the `symbol-js/carbon` files, and their mappings: is it because of these code changes (99.99% no)? is it because of something that changed in Figma (maybe)? is it because before we missed something in the mapping and now it's been fixed (also maybe)? something else?

probably it would be better to separate changes to the pipeline (this PR) from changes to the generated files (in a follow-up PR, that may include also changes to due to the paths differences in the exported files from Figma)

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
zamoore: It is - It's just explaining what `CARBON_ICON_OVERRIDES` maps to/from

packages/flight-icons/catalog.json
zamoore: I can get on board with excluding the `mapping` altogether. If there is no Carbon equiv, it will simply be `undefined`.

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
zamoore: I'm good with the suggestion

jorytindall (dismissed):
I'm not sure how much feedback I can provide other than what I already have, but looks good! 🫡

packages/flight-icons/scripts/build-parts/generateBundleSymbolJS.ts
didoo: Ah ok, I get it now. Maybe I would do something like this to make the comment clearer:
```suggestion
// Expected format: `{ [iconName]: carbonIconPathWithoutExtension }`
```

didoo (dismissed):
Left only one tiny suggestion, non blocker

didoo (approved):
👍

PR #3779, opened 2026-04-03
[Project Solar / Phase 1 / Components] `CodeBlock` carbonization

dchyun:
### :pushpin: Summary

This PR carbonizes the `CodeBlock` component and adds component tokens for it to the tokens library.

Additionally it takes the existing `--hds-code-block` CSS variables and moves them into the tokens library.

[CodeBlock carbonization page](https://hds-showcase-git-project-solar-phase-1dchyunco-28d676-hashicorp.vercel.app/carbonization/components/code-block)

Summary of changes:

- Reorganized theming CSS variables for CodeBlock to be moved from components package to tokens package

#### CodeBlock theme styles

Currently the `CodeBlock` gets its theming for both overall styles and a `themes.scss` file which contains a variety of CSS variables all of the format `--hds-code-block`. These variables are then used in the main file to style `prism.js` elements.

These variables range from "core" colors to semantic level tokens which reference the core colors.

```
--hds-code-block-color-blue: #2d8eff;
--hds-code-block-color-cyan: #32fff7;
--hds-code-block-color-green: #86ff13;
--hds-code-block-color-orange: #ffa800;
--hds-code-block-color-purple: #c76cff;
--hds-code-block-color-red: #ff3b20;
```

As part of the Carbonization effort these colors now can not be hard-coded and need to changed based on the theme. I have added the core color tokens to the tokens library and now they are generated with the format `--token-code-block-color-palette-blue`.

The semantic level code block tokens have remained unchanged, but now reference the tokens coming from the HDS tokens library.

```
--hds-code-block-color-atrule: var(--token-code-block-color-palette-blue);
--hds-code-block-color-code-block-attr-name: var(--token-code-block-color-palette-blue);
--hds-code-block-color-attr-value: var(--token-code-block-color-palette-blue);
```

### :link: External links

Jira ticket: [HDS-6079](https://hashicorp.atlassian.net/browse/HDS-6079)
Figma file: [if it applies]

***

<details>

</details>

[HDS-6079]: https://hashicorp.atlassian.net/browse/HDS-6079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/components/src/styles/components/code-block/index.scss
```diff
@@ -327,9 +341,18 @@ $hds-code-block-code-footer-height: 48px;
     display: block;
     height: $hds-code-block-code-footer-height;
     margin: 0 0 -#{$hds-code-block-code-padding} -1000px; // cover line highlights when line numbers are enabled
-    background: linear-gradient(360deg, #0d0e12 37.07%, rgba(13, 14, 18, 25%) 100%);
+    background: linear-gradient(
```
didoo, line 342: [praise] 👏

packages/components/src/styles/components/code-block/index.scss
```diff
@@ -327,9 +341,18 @@ $hds-code-block-code-footer-height: 48px;
...
+      360deg,
+      var(--token-code-block-color-surface-primary) 37.07%,
+      var(--token-code-block-color-surface-primary-with-alpha) 100%
+    ); // uses --code-block-color-surface-primary with 25% opacity
```
didoo:
```suggestion
    );
```

packages/components/src/styles/components/code-block/index.scss
```diff
@@ -269,11 +275,19 @@ $hds-code-block-code-footer-height: 48px;
...
     left: -$hds-code-block-code-padding;
-    background-color: var(--hds-code-block-color-line-highlight);
-    border: solid var(--hds-code-block-color-line-highlight-border);
+    background-color: var(--token-code-block-line-highlight-surface-color);
+    border: solid var(--token-code-block-line-highlight-border-color);
     border-width: 1px 0 1px 4px;
     mix-blend-mode: screen;
     pointer-events: none;
+
+    @include hds-apply-only-if-carbon($mode: "cds-g0") {
```
didoo: [todo] can you double check? I think you can reset the `mix-blend-mode` to `normal` for all the "cds" modes not just `g0/g10`

showcase/app/components/page-carbonization/components/code-block/index.gts
```diff
@@ -0,0 +1,454 @@
...
+            </SF.Item>
+          </ShwFlex>
+        </:reference>
+      </ShwCarbonizationComparisonGrid>
+
+      <ShwDivider @level={{2}} />
+
+      <ShwTextH2>Height limit</ShwTextH2>
+
+      <ShwCarbonizationComparisonGrid @layout="side-by-side">
```
didoo, line 212: [suggestion] do you think we can use code examples that have less wide text, so that we avoid the horizontal page scrolling?

(also in other examples)

didoo (commented):
Just a couple of minor comments, everything else is OK.

packages/components/src/styles/components/code-block/index.scss
dchyun: This has no effect on the g90/g100 themes but updated the mixin to be for all cds modes for simplicity.

showcase/app/components/page-carbonization/components/code-block/index.gts
dchyun: Removed any horizontal scrolling on the page

didoo (approved):
I am pre-approving. But probably better to do a final pass once the `CopyButton` is done too, just in case.

I have also left a question for the @hashicorp/hds-design in the shared document: https://docs.google.com/document/d/1b3lGaN1sJBD5G_1vPkpBubznLR1WacHKK9kzlULTO4c/edit?tab=t.gvcs3oirh9o7

LilithJames-HDS (approved):
Same note here as on CodeEditor. This can be approved but I opened a new question in Figma about the button states

jorytindall:
I think there's a visual bug in the alignment of the line highlight. Might because the change in font size also needs to be reflected in the sizing/height of the line highlight, but out of my realm of expertise here.

Screenshot 2026-04-16 at 5 07 55 PM

dchyun:
> I think there's a visual bug in the alignment of the line highlight. Might because the change in font size also needs to be reflected in the sizing/height of the line highlight, but out of my realm of expertise here.

Went down quite a rabbit hole here, but solved the issue. See below for long winded explanation. 

This issue was present with line highlighting only when line numbers weren't present. `prism.js` calculates the line highlighting based on the `pre` element (parent to the `code` element) when there are no line numbers and based on the `code` when there are. We only set our `font-size` on the `code` so this was creating a mismatch between the font size the calculations were made on and the actual font size used.

I've moved the font styles to apply to both the `pre` and `code` elements and this has fixed the issue.

jorytindall (approved):
Thanks for digging into that bug! Everything is looking good now

LilithJames-HDS (approved):
Thanks for making the changed to the color state tokens for the buttons

packages/tokens/src/products/shared/code-block.json
```diff
@@ -0,0 +1,376 @@
+{
+  "code-block": {
+    "color": {
+      "palette": {
```
didoo: [thought] In the [upcoming naming convention changes](https://docs.google.com/presentation/d/12MNvNr937iJpjRhKkBz-eMXJiilU7mE9yWpdSn1Iwxo/edit?slide=id.g3d6281424a5_0_0#slide=id.g3d6281424a5_0_0) we want to get rid of "palette" as term and use "core". Since core, in this context, does not mean much, what if we call these `code-block.color.syntax-highlight.***`?

packages/tokens/src/products/shared/code-block.json
```diff
@@ -0,0 +1,376 @@
...
+          "$modes": {
+            "default": "#ffeb3b",
+            "cds-g0": "{carbon.color.yellow.70}",
+            "cds-g10": "{carbon.color.yellow.70}",
+            "cds-g90": "{carbon.color.yellow.30}",
+            "cds-g100": "{carbon.color.yellow.30}"
+          }
+        }
+      },
+      "foreground": {
```
didoo: [todo] In the [upcoming naming convention changes](https://docs.google.com/presentation/d/12MNvNr937iJpjRhKkBz-eMXJiilU7mE9yWpdSn1Iwxo/edit?slide=id.g3d6281424a5_0_0#slide=id.g3d6281424a5_0_0) we want to consistently use `[foreground|surface|border]-color`, what do you think of making this change here too?

packages/tokens/src/products/shared/code-block.json
didoo, line 1: [issue?] General consideration: I was checking some tokens, that seemed not used in the codebase, and I could not find any instance. So I checked in the [branch](https://github.com/hashicorp/design-system/pull/3785/) for the `CodeEditor` (maybe they would be used there) but I couldn't find any usage there as well. So I asked Copilot to check the list of all the tokens added in this PR and this is the response:

> These tokens from your list are not used in the Sass files under the requested styles path:
> - `--token-code-block-color-foreground-interactive`
> - `--token-code-block-color-foreground-interactive-hover`
> - `--token-code-block-color-foreground-interactive-active`
> - `--token-code-block-color-foreground-action-hover`
> - `--token-code-block-color-foreground-action-active`
> - `--token-code-block-color-surface-interactive-active`
> - `--token-code-block-line-numbers-border-color`

Maybe I am missing something? 🤔 If not, do you mind checking?

didoo (commented):
Left a couple of comments, mainly related to tokens naming and usage. Everything else is OK

packages/tokens/src/products/shared/code-block.json
dchyun: Updated to use `code-block.color.syntax-highlight`

packages/tokens/src/products/shared/code-block.json
dchyun: Updated to align to the new format and use `[foreground|surface|border|focus]-color`

packages/tokens/src/products/shared/code-block.json
dchyun: Went through all the variables and these could be removed. Not used in the code block or editor. `surface-interactive-active` was referenced in another code block token `button.surface.active` for the HDS theme so I used the HDS value from `surface-interactive-active` directly there instead.

packages/tokens/src/products/shared/code-block.json
didoo: Given that now we have `foreground.color` consider using `code-block.syntax-highlight.color` instead (non blocker)

didoo (approved):
Left a non-blocking comment. Were you to make the change, don't wait for my approval, merge away :)

packages/tokens/src/products/shared/code-block.json
dchyun: Updated to `code-block.syntax-highlight.color`

PR #3801, opened 2026-04-10
[Project Solar / Phase 1 / Components] `AppSideNav` carbonization

dchyun:
### :pushpin: Summary

This PR carbonizes the `AppSideNav` component and adds component tokens for it to the tokens library.

It also makes a few updates to the existing list of `AppSideNav` tokens to align to our naming convention and remove unneeded tokens.

[AppSideNav carbonization](https://hds-showcase-git-project-solar-phase-1dchyunap-785e19-hashicorp.vercel.app/carbonization/components/app-side-nav)

Summary of changes:
- Aligned existing tokens to our naming convention and removed unneeded tokens

#### AppSideNav token changes

There have been several changes made to the existing `AppSideNav` tokens as part of the carbonization effort.

First, there are some variables that do not align to our token naming structure. The following token names have been updated.

- `--token-app-side-nav-body-list-item-spacing-vertical` -> `--token-app-side-nav-body-list-item-margin-top`
- `--token-app-side-nav-body-list-item-content-spacing-horizontal` -> `--token-app-side-nav-body-list-item-content-gap`

Second, several component tokens have been removed. The majority of these were `--token-app-side-nav-color` tokens which created a kind of core color palette for the app side nav. All of these tokens mapped to existing HDS tokens so those HDS tokens could be referenced directly instead.

- `--token-app-side-nav-color-foreground-primary`
- `--token-app-side-nav-color-foreground-strong`
- `--token-app-side-nav-color-foreground-faint`
- `--token-app-side-nav-color-surface-primary`
- `--token-app-side-nav-color-surface-interactive-hover`
- `--token-app-side-nav-color-surface-interactive-active`

There were also a few tokens for `--token-app-side-nav-header` which were unused.

- `--token-app-side-nav-header-home-link-padding`
- `--token-app-side-nav-header-home-link-logo-size`
- `--token-app-side-nav-header-home-link-logo-size-minimized`
- `--token-app-side-nav-header-actions-spacing`

### :link: External links

Jira ticket: [HDS-6074](https://hashicorp.atlassian.net/browse/HDS-6074)
Figma file: [if it applies]

***

<details>

</details>

[HDS-6074]: https://hashicorp.atlassian.net/browse/HDS-6074?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jorytindall:
Most everything is looking good here, but a couple of things I caught:
- The padding in the list items matches (7px 16px) between Figma, but the resulting height is different. In the showcase the list items are 36px, but in Figma they're 32px (which is what the Carbon component nets out at).
- Would it be possible to see a frameless example to see how the collapse button looks in context?
- The `BadgeCount` in the list item (and now that I'm looking) in general looks pretty off from Figma in terms of horizontal spacing. Something we can discuss during our sync I think.

dchyun:
@jorytindall 
> The padding in the list items matches (7px 16px) between Figma, but the resulting height is different. In the showcase the list items are 36px, but in Figma they're 32px (which is what the Carbon component nets out at).

We have an existing token that sets a min-height of 36px on the side nav list items. I've added theming for this token to set it to 32px for the carbon themes.

> Would it be possible to see a frameless example to see how the collapse button looks in context?

You can see some frameless examples if you go to the [main side nav showcase page](https://hds-showcase.vercel.app/components/app-side-nav#responsiveness) and set the theming toggle to the Carbon theme.

> The `BadgeCount` in the list item (and now that I'm looking) in general looks pretty off from Figma in terms of horizontal spacing. Something we can discuss during our sync I think.

As discussed in our sync it was intentional to not set a `min-width` on the badge as is done in the Carbon component. This decision may need to be reflected in Figma as well to match.

jorytindall (approved):
This is looking solid to me, the only thing that caught my attention was the line height of the list items. I technically created a new token for this which uses 20px for HDS and 18px for CDS themes (not possible to use a multiplier in Figma). I'm not sure how much we want to index on this since you're already using the `body-200` line-height, but it is technically what Carbon uses in `heading-compact-01`, but to me it isn't a total blocker since line-heights in general are somewhat different between Figma and our code tokens.

dchyun:
> This is looking solid to me, the only thing that caught my attention was the line height of the list items. I technically created a new token for this which uses 20px for HDS and 18px for CDS themes (not possible to use a multiplier in Figma). I'm not sure how much we want to index on this since you're already using the `body-200` line-height, but it is technically what Carbon uses in `heading-compact-01`, but to me it isn't a total blocker since line-heights in general are somewhat different between Figma and our code tokens.

@jorytindall generally we're avoiding adding line height tokens unless it creates a significant difference between the carbon component and the HDS one. I'd prefer to not add the token if you are okay with that.

jorytindall:
> > This is looking solid to me, the only thing that caught my attention was the line height of the list items. I technically created a new token for this which uses 20px for HDS and 18px for CDS themes (not possible to use a multiplier in Figma). I'm not sure how much we want to index on this since you're already using the `body-200` line-height, but it is technically what Carbon uses in `heading-compact-01`, but to me it isn't a total blocker since line-heights in general are somewhat different between Figma and our code tokens.
> 
> @jorytindall generally we're avoiding adding line height tokens unless it creates a significant difference between the carbon component and the HDS one. I'd prefer to not add the token if you are okay with that.

Yep, works for me, just noting it so we have a historical record of the decision.

packages/components/src/styles/components/app-side-nav/toggle-button.scss
```diff
@@ -56,32 +61,33 @@
...
+    @include hds-apply-only-if-carbon() {
+      content: none;
+    }
   }
 
   &:hover,
   &.mock-hover {
     width: 30px;
-    background-color: var(--token-app-side-nav-color-surface-interactive-hover);
+    background-color: var(--token-color-surface-interactive-hover);
```
didoo, line 78: there are specific tokens for the states of the toggle button I think:

Image

didoo: btw, it would be worth to use the same term in Figma ("collapse button") and in code ("toggle button")

packages/components/src/styles/components/app-side-nav/toggle-button.scss
didoo, line 1: I see this token in Figma for the toggle/collapse button, but I am not sure if it's related to the height in code and how:

Image

packages/components/src/styles/components/app-side-nav/content.scss
```diff
@@ -48,8 +48,9 @@
 .hds-app-side-nav__list-title {
   min-height: var(--token-app-side-nav-body-list-item-height);
   margin-top: var(--token-app-side-nav-body-list-margin-vertical);
-  padding: 9px var(--token-app-side-nav-body-list-item-padding-horizontal); // 8px = (min-height - body-100-line-height) / 2
-  color: var(--token-app-side-nav-color-foreground-faint);
+  padding: var(--token-app-side-nav-body-title-padding-vertical)
```
didoo: [idea] we could consider having a `app-side-nav.body.title.padding.horizontal` token that is an alias of `app-side-nav.body.list-item.padding.horizontal` (in JSON, and use it here)

didoo: Also, the component is `hds-app-side-nav__list-title` so maybe the token should be `--token-app-side-nav-body-list-title-padding-***`?

packages/tokens/src/products/shared/app-side-nav.json
```diff
@@ -41,112 +61,221 @@
...
+            "cds-g10": "0",
+            "cds-g90": "0",
+            "cds-g100": "0"
+          }
         },
-        "faint": {
-          "$type": "color",
-          "$value": "var(--token-color-foreground-faint)",
-          "group": "components"
+        "selected": {
```
didoo: we could
- move the "indicator" out of the "selected" parent
- more the "selected" variant at the end of "[foreground|surface]-color" (more consistent with naming convention)

packages/tokens/src/products/shared/app-side-nav.json
```diff
@@ -1,5 +1,18 @@
 {
   "app-side-nav": {
+    "surface": {
```
didoo: consider moving this under "wrapper" (more specific/correct)

packages/tokens/src/products/shared/app-side-nav.json
```diff
@@ -41,112 +61,221 @@
...
-        "interactive-hover": {
-          "$type": "color",
-          "$value": "var(--token-color-surface-interactive-hover)",
-          "group": "semantic"
-        },
-        "interactive-active": {
-          "$type": "color",
-          "$value": "var(--token-color-palette-neutral-300)",
-          "group": "semantic"
+      "title": {
```
didoo: consider renaming this "list-title" and move it between "list" and "list-item"

didoo (commented):
I have done a first pass and left a few comments, suggestions, thoughts.

I have also a few high-level things to discuss:

---

@jorytindall I am worried that there is not enough difference between a title (not interactive) and an item (interactive); it seemed the title was a broken link, at first, when I was interacting with it (before, at least, there was a visual differentiation)

Image

---

The focus ring is cut out by the overflow: hidden, any chance we can get it inside of it?

Image

Image

---

@dchyun could it be that the `hds-interactive-dark-theme` mixin never uses the `$is-within-app-side-nav` toggle, and the code could be simplified (in which case, I think we can open a separate PR for this cleanup)

Image

---

I see some artefacts in the toggle/collapse button:

Image

---

packages/components/src/styles/components/app-side-nav/toggle-button.scss
dchyun: On the eng side we don't need any toggle button specific surface color tokens. 

The hover color uses the semantic `surface-interactive-hover` which is equal to `backgroundHover` for Carbon themes and is what is existing for the HDS theme

dchyun: Also the Hds component is called `HdsAppSideNavToggleButton` so if we align we should align to "toggle"

packages/components/src/styles/components/app-side-nav/toggle-button.scss
dchyun: @jorytindall correct me if I'm wrong but I believe this was a figma-only token to remove the curves above and below the toggle that are present for the HDS theme.
Screenshot 2026-05-07 at 2 16 17 PM

packages/components/src/styles/components/app-side-nav/content.scss
dchyun: Updated to create an alias and use `list-title`

packages/tokens/src/products/shared/app-side-nav.json
dchyun: Updated to use this suggestion

packages/tokens/src/products/shared/app-side-nav.json
dchyun: Renamed to `list-title`

dchyun:
> @dchyun could it be that the hds-interactive-dark-theme mixin never uses the $is-within-app-side-nav toggle, and the code could be simplified (in which case, I think we can open a separate PR for this cleanup)

@didoo I've actually done this clean up already in the [AppHeader carbonization](https://github.com/hashicorp/design-system/pull/3819/changes#diff-4f4f49b1a56f79fb06318b1e1ba36d32ee390b7d3831a07b85a2ba2930c3ee86). The `hds-interactive-dark-theme` is only used by the `AppHeader`. In that PR I've deleted the file and moved all functionality inside the `app-header.scss`. In addition to that I've removed all of the logic for `$is-within-app-side-nav` since it's not used ever inside the side nav. 

> I see some artefacts in the toggle/collapse button:

Don't think I'm seeing these issue on my side.
Screenshot 2026-05-07 at 2 57 17 PM

didoo:
> Don't think I'm seeing these issue on my side. 

If you zoom in, you can see it in your screenshot (I _think_). My screenshot is from here: https://hds-showcase-git-project-solar-phase-1dchyunap-785e19-hashicorp.vercel.app/foundations/theming/frameless/demo-application-with-theme-switcher

packages/components/src/styles/components/app-side-nav/toggle-button.scss
jorytindall: Yes that's correct, there's some weird workarounds we have to do to make the shape for the collapse button work with height and auto layout.

jorytindall:
> @jorytindall I am worried that there is not enough difference between a title (not interactive) and an item (interactive); it seemed the title was a broken link, at first, when I was interacting with it (before, at least, there was a visual differentiation)

This is a good point, I've done some playing around with the tokens, it's a little challenging to get something that seems different enough solely via color. I'm not sure that changing the font-weight is the right approach, but I have some further exploration [here](https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/branch/GWMDKHuGr98DFxSptH6vgj/HDS-Components-v2.0?m=auto&node-id=102201-370014&t=SIKjfQvy7RloyPpM-1)

Screenshot 2026-05-07 at 12 58 36 PM

packages/components/src/styles/components/app-side-nav/toggle-button.scss
didoo: @jorytindall does it mean that the tokens in Figma are not needed as well? 🤔

didoo:
> > @jorytindall I am worried that there is not enough difference between a title (not interactive) and an item (interactive); it seemed the title was a broken link, at first, when I was interacting with it (before, at least, there was a visual differentiation)
> 
> This is a good point, I've done some playing around with the tokens, it's a little challenging to get something that seems different enough solely via color. I'm not sure that changing the font-weight is the right approach, but I have some further exploration [here](https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/branch/GWMDKHuGr98DFxSptH6vgj/HDS-Components-v2.0?m=auto&node-id=102201-370014&t=SIKjfQvy7RloyPpM-1)

Personally, I think the one that uses `text-primary` is distinct enough (considering that also the text is slightly smaller)

btw @dchyun I think there is an issue in code, because I see the font-size as `14px` but in Figma it's `13px` in HDS and `12px` in CDS (@jorytindall can you confirm?)
screenshot_6279

packages/components/src/styles/components/app-side-nav/toggle-button.scss
jorytindall: Technically these are different so we might have a mismatch; `surface-interactive-hover` maps to `neutral/100`, where I'm using `neutral/200` in the component. I did have to do some funky things in the Figma component to get this element to work (because it's a shape element, I had to create a different layer for the Carbon theme)

jorytindall:
> Personally, I think the one that uses text-primary is distinct enough (considering that also the text is slightly smaller)

@didoo agreed, @dchyun I've made this update in Figma.

dchyun:
> Personally, I think the one that uses `text-primary` is distinct enough (considering that also the text is slightly smaller)
> 
> btw @dchyun I think there is an issue in code, because I see the font-size as `14px` but in Figma it's `13px` in HDS and `12px` in CDS (@jorytindall can you confirm?)

@didoo @jorytindall First, I've updated to use `textPrimary` for the title with carbon themes.

On the font size, I think there is actually an issue with our core mapping. The list title uses `body-100` and the list link uses `body-200`. For HDS these have different font sizes (13px and 14px). However, for Carbon they map to `body01.fontSize` and `bodyCompact01.fontSize` which both equal 14px.

Should I use custom app header tokens here to create some font size difference between the two or should we revisit the core mapping?

jorytindall:
> btw @dchyun I think there is an issue in code, because I see the font-size as 14px but in Figma it's 13px in HDS and 12px in CDS (@jorytindall can you confirm?)

@didoo What element in Carbon are you referring to has a font-size of 12px? I'm generally seeing everything sidebar related (there are several examples) using 14px. That being said, I would say we should revisit the root mapping if that's really the issue rather than creating a unique token (given that it doesn't seem like the relevant Carbon components have the same type of element/structure)

didoo:
> > btw @dchyun I think there is an issue in code, because I see the font-size as 14px but in Figma it's 13px in HDS and 12px in CDS (@jorytindall can you confirm?)
> 
> @didoo What element in Carbon are you referring to has a font-size of 12px? I'm generally seeing everything sidebar related (there are several examples) using 14px. 

The List Title (that I've seen called also Section Title, which is confusing). When I use the Dev inspector in Figma (see [this frame](https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/branch/GWMDKHuGr98DFxSptH6vgj/HDS-Components-v2.0?m=auto&node-id=102440-23100&t=S00cMk0XLE9TvJmr-1)) this is what I see:

screenshot_6292

Not sure if it's Figma that's being confused by something, but by the look of it it's using 12px for `cds`

screenshot_6293

/cc @dchyun

packages/components/src/styles/components/app-side-nav/toggle-button.scss
didoo: @dchyun @jorytindall this is resolved now, right?

didoo (approved):
Nice!

PR #3814, opened 2026-04-14
[Project Solar / Phase 1 / Components] `Dropdown` carbonization (HDS-6083)

KristinLBradley:
**Note**:
* I am not touching anything related to focus outlines in component PRs other than positioning if needed (focus outline styling will be addressed separately as a whole)
* There is no Carbon web component version of the multiselect/checkbox Dropdown although there is a React one.

----

### :pushpin: Summary

If merged, this PR adds the carbonized `Dropdown` component to HDS.

👉 **Preview**: https://hds-showcase-git-project-solar-phase-1kristinh-ec9883-hashicorp.vercel.app/carbonization/components/dropdown

**Questions**: https://docs.google.com/document/d/1b3lGaN1sJBD5G_1vPkpBubznLR1WacHKK9kzlULTO4c/edit?tab=t.90bih7gkkj59

### :camera_flash: Screenshots

**_Partial Showcase screenshot:_**
image

### :link: External links

* Jira ticket: [HDS-6083](https://hashicorp.atlassian.net/browse/HDS-6083)
* Figma file: https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/branch/GWMDKHuGr98DFxSptH6vgj/HDS-Components-v2.0?node-id=2-10&p=f&t=gZI7paU6vdfow8CD-0
  * [Breakdown of interactive list item states](https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/branch/GWMDKHuGr98DFxSptH6vgj/HDS-Components-v2.0?m=auto&node-id=100917-122073&t=Patjv2VwSFS1FOtk-1)
  * [“Critical” interactive states](https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/branch/GWMDKHuGr98DFxSptH6vgj/HDS-Components-v2.0?m=auto&node-id=102417-327726&t=FzXE9gAZICar9lkO-1)

***

### :eyes: Component checklist

- [ ] Percy was checked for any visual regression
- ~~[ ] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))~~

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

<details>

<summary>:clipboard: PCI review checklist</summary>

[...]

PR #3819, opened 2026-04-20
[Project Solar / Phase 1 / Components] `AppHeader` carbonization

dchyun:
> [!IMPORTANT]
> DO NOT MERGE until the focus ring work merged in #3884  

### :pushpin: Summary

This PR carbonizes the `AppHeader` component and adds component tokens for it to the tokens library.

Additionally it removes the `hds-interactive-dark-theme` mixin, which was only used in the `AppHeader`.

[AppHeader carbonization page](https://hds-showcase-8qncv1ogn-hashicorp.vercel.app/carbonization/components/app-header)

**Note:** This PR contains a squash of the Dropdown carbonization in #3814 since the AppHeader uses the Dropdown. Do not merge until that PR is merged, or the squash is removed.

Summary of changes:
- Removed the `hds-interactive-dark-theme` mixin

#### `hds-interactive-dark-theme` removal

Currently the `AppHeader` utilizes the `hds-interactive-dark-theme` mixin for its interactive elements. It is the only component that utilizes this mixing currently. 

I have decided to remove this mixin, and implement all styles within the `AppHeader` for the following purposes:

- It is only utilized within the `AppHeader` currently
- Theming the AppHeader would be difficult as the mixin is only for "dark" components on light (HDS, cds-0, cds-10) themes, but the Carbonized AppHeader needs to be a "light" component on a light theme. There would be two main options
  - Remove reliance on the mixin and have all styles live within the AppHeader, then create tokens for those styles
  - Create an alternative `hds-interactive-light-theme` mixin and switch between using that mixin and the `dark-interavctive` based on `hds-apply-only-if-carbon` 

### :link: External links

Jira ticket: [HDS-6073](https://hashicorp.atlassian.net/browse/HDS-6073)
Figma file: [if it applies]

***

<details>

</details>

[HDS-6073]: https://hashicorp.atlassian.net/browse/HDS-6073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/tokens/dist/products/css/themed-tokens/with-css-selectors--advanced/tokens.css
```diff
@@ -1624,6 +1723,28 @@
 .hds-theme-light,
 .hds-mode-cds-g0 {
   --token-app-footer-link-focus-outline-color: #0f62fe;
+  --token-app-header-actions-gap: 0px;
+  --token-app-header-border-color: rgba(0, 0, 0, 0);
```
jorytindall: All of the Carbon themes have a border with 0% opacity, I've mapped them like this:
- CDS-0: border-subtle-00
- CDS-10: default-gray-30
- CDS-90: default-gray-70
- CDS-100: default-gray-80

packages/tokens/dist/products/css/themed-tokens/with-css-selectors--advanced/tokens.css
```diff
@@ -1624,6 +1723,28 @@
...
+  --token-app-header-interactive-foreground-color-disabled: rgba(
+    22,
+    22,
+    22,
+    0.25
+  );
+  --token-app-header-interactive-surface-color-active: rgba(141, 141, 141, 0.5);
+  --token-app-header-interactive-surface-color-default: rgba(0, 0, 0, 0);
+  --token-app-header-interactive-surface-color-disabled: #c6c6c6;
+  --token-app-header-interactive-surface-color-hover: rgba(141, 141, 141, 0.12);
```
jorytindall: For some reason I'm not seeing the interactive states for the home-link (not sure where else to comment this)

jorytindall (commented):
I did a quick run through of this and caught a couple of things right off the bat, but I think  we're going to need to make some decisions about the spacing and visual language of some of the interactive elements. I've left some comments for us to discuss in the planning doc.

packages/tokens/dist/products/css/themed-tokens/with-css-selectors--advanced/tokens.css
dchyun: My bad missed adding the border color. Added in now

packages/tokens/dist/products/css/themed-tokens/with-css-selectors--advanced/tokens.css
dchyun: The Carbon component doesn't have an interactive state for the home link, so I removed it from the HDS home link as well.

packages/components/src/styles/components/app-header.scss
```diff
@@ -18,10 +104,10 @@
   align-items: center;
   justify-content: space-between;
   min-height: var(--token-app-header-height);
-  padding: 12px 16px 11px 16px; // bottom padding of 11px is to account for the bottom border
-  color: var(--token-color-foreground-high-contrast);
-  background: var(--token-color-palette-neutral-700);
-  border-bottom: 1px solid var(--token-color-palette-neutral-500);
+  padding: var(--token-app-header-padding-top) 16px var(--token-app-header-padding-bottom) 16px;
+  color: var(--token-app-header-foreground-color);
```
jorytindall: Looking in a bit more detail at the Carbon, they use `text-secondary` for the color in all of the interactive elements (text, icons, in dropdowns, etc). I missed this originally, but I'll make the update.

jorytindall (commented):
Sorry @dchyun I thought I had submitted this review, but there's one cleanup item regarding text color, then I think this will be good to go

packages/components/src/styles/components/app-header.scss
dchyun: I've updated all the interactive colors to use `textSecondary` for the default state and `textPrimary` for hover to align to the Carbon component.

On the HomeLink I've kept the text color as `textPrimary` to align to the Carbon title. I've also update the font weight to be bold which aligns.

Note: These changes were done to the `app-header-interactive-foreground-color` token. I looked into the `app-header-foreground-color` token which you point to here more and it will only be used if someone places non-interactive text into the app header through one of the yields. 

I considered just removing this, but there is a color set in the existing HDS for these non-interactive text so I've left it even if it may never come up in practice.

jorytindall: Perfect, makes sense. I updated the interactive states in Figma as well

jorytindall (approved):

showcase/app/components/page-carbonization/components/app-header/index.gts
```diff
@@ -0,0 +1,563 @@
...
+
+const AppHeaderCarbonizationIndex: TemplateOnlyComponent = <template>
+  {{pageTitle "AppHeader - Carbonization"}}
+
+  <ShwTextH1>AppHeader - Carbonization</ShwTextH1>
+
+  <section>
+    <ShwTextH2>Content</ShwTextH2>
+
+    <ShwTextH3>Basic header with logo</ShwTextH3>
```
KristinLBradley: It looks like the only difference of note between the "Basic header with logo" examples and the following "Header with navigation and actions" is that the actual Carbon component switches the position of the Menu and the "header-name" text?

Perhaps you could consolidate the examples together to save space and since they aren't really functionally different except in the Carbon components? It would make it more obvious what's different if you could see the 2 Carbon component examples together also.

packages/components/src/styles/components/app-header.scss
```diff
@@ -7,7 +7,91 @@
...
+  // Active:
+  &:active,
+  &.mock-active {
+    color: var(
+      --token-app-header-interactive-foreground-color-default
+    ); // to avoid overrides by specificity (eg. `a:hover`)
+    background-color: var(--token-app-header-interactive-surface-color-active);
+
+    @if ($add-visible-border) {
+      border-color: var(--token-app-header-interactive-border-color-disabled);
```
KristinLBradley: The border-color shifted a bit from the original color. (#8c909c vs. #656a76 in prod) Is that as intended? (Was it changed to improve the contrast a bit?)

Image

packages/components/src/styles/components/app-header.scss
```diff
@@ -7,7 +7,91 @@
...
 
-@use "../mixins/interactive-dark-theme" as *;
+@use "../mixins/carbonization" as *;
+@use "../mixins/focus-ring" as *;
+@use "../mixins/button" as *;
+
+$hds-interactive-app-header-theme-focus-ring-action-box-shadow:
+  inset 0 0 0 1px var(--token-app-header-interactive-focus-action-internal),
+  0 0 0 3px var(--token-app-header-interactive-focus-action-external);
+// Used for interactive elements such as Button & Dropdown
```
KristinLBradley, line 14:
```suggestion

// Used for interactive elements such as Button & Dropdown
```

packages/components/src/styles/components/app-header.scss
```diff
@@ -18,10 +102,10 @@
   align-items: center;
   justify-content: space-between;
   min-height: var(--token-app-header-height);
-  padding: 12px 16px 11px 16px; // bottom padding of 11px is to account for the bottom border
-  color: var(--token-color-foreground-high-contrast);
-  background: var(--token-color-palette-neutral-700);
-  border-bottom: 1px solid var(--token-color-palette-neutral-500);
+  padding: var(--token-app-header-padding-top) 16px var(--token-app-header-padding-bottom) 16px;
+  // color: var(--token-app-header-foreground-color);
```
KristinLBradley: Would you want to delete this commented out line?

KristinLBradley (commented):
I left a few comments/questions and minor suggestions.

packages/components/src/styles/components/app-header.scss
dchyun: Thanks for catching that. Not intentional. Have aligned to previous color now.

showcase/app/components/page-carbonization/components/app-header/index.gts
dchyun: Removed this section

KristinLBradley:
I found another minor style difference.

See the border styling.

**in this PR**:  
image

**in prod**:  
image

KristinLBradley:
There's a border color difference in button active states from prod. Is this as intended?

image

KristinLBradley (commented):
I found a couple more minor style differences with borders that I added comments on.

dchyun:
> I found another minor style difference. See the border styling.

@KristinLBradley I think this is coming from the work done to carbonize the `DisclosurePrimitive` by @didoo. I would say this difference is acceptable. 

> There's a border color difference in button active states from prod. Is this as intended?

Good catch, that is a difference I missed. However, I think we are considering removing all active state styling from both HDS and carbon themes in #3830. @didoo is that that case? If so, I'll leave this as is with the intention the active state styling is removed completely as part of that PR.

didoo:
> > I found another minor style difference. See the border styling.
> 
> @KristinLBradley I think this is coming from the work done to carbonize the `DisclosurePrimitive` by @didoo. I would say this difference is acceptable.
> 
> > There's a border color difference in button active states from prod. Is this as intended?

@dchyun @KristinLBradley yes I can confirm (these "tests" are to check that there are not regressions in other use cases, with the overrides for the AppHeader)

> Good catch, that is a difference I missed. However, I think we are considering removing all active state styling from both HDS and carbon themes in #3830. @didoo is that that case? If so, I'll leave this as is with the intention the active state styling is removed completely as part of that PR.

@dchyun at the end we decided **_not_** to remove the active states (see [comment here](https://hashicorp.atlassian.net/browse/HDS-6227?focusedCommentId=1017393)). I've now closed #3830 (I forgot).

dchyun:
> > Good catch, that is a difference I missed. However, I think we are considering removing all active state styling from both HDS and carbon themes in https://github.com/hashicorp/design-system/pull/3830. @didoo is that that case? If so, I'll leave this as is with the intention the active state styling is removed completely as part of that PR.

> @dchyun at the end we decided not to remove the active states (see [comment here](https://hashicorp.atlassian.net/browse/HDS-6227?focusedCommentId=1017393)). I've now closed https://github.com/hashicorp/design-system/pull/3830 (I forgot).

@KristinLBradley I've updated the active state border now to set the correct color

KristinLBradley:
@jorytindall Dylan merged in the carbonized `Dropdown` into this branch. The spacing between the `Dropdown` text and the arrow looks too big in the carbon themes. What should the spacing be and how should it work? (Does it need to be responsive?)

image

### _This is the `Dropdown` design in Figma_:
image

KristinLBradley (commented):
Looks good but I added a comment with a question for @jorytindall about the spacing within the `Dropdown` trigger button between the text and arrow icon.

jorytindall:
> Looks good but I added a comment with a question for @jorytindall about the spacing within the Dropdown trigger button between the text and arrow icon.

@KristinLBradley @dchyun I think we decided that the spacing between the text and the chevron specifically in the `AppHeader` should be reduced to 8px.

KristinLBradley:
> > Looks good but I added a comment with a question for @jorytindall about the spacing within the Dropdown trigger button between the text and arrow icon.
> 
> @KristinLBradley @dchyun I think we decided that the spacing between the text and the chevron specifically in the `AppHeader` should be reduced to 8px.

@dchyun I think you may want to add an `AppHeader` specific style override of the `Dropdown` button gap value then? (or whatever you think is the best way to handle it.)

packages/tokens/src/products/shared/app-header.json
```diff
@@ -1,22 +1,274 @@
...
+            "default": "{color.foreground.high-contrast}",
+            "cds-g0": "{carbon.themes.white.textPrimary}",
+            "cds-g10": "{carbon.themes.g10.textPrimary}",
+            "cds-g90": "{carbon.themes.g90.textPrimary}",
+            "cds-g100": "{carbon.themes.g100.textPrimary}"
+          }
+        }
+      },
+      "logo": {
+        "width": {
```
didoo: [question] since width and height have the same value, why not use `size`? it's an acceptable term, from the nomenclature perspective

dchyun:
> > > Looks good but I added a comment with a question for @jorytindall about the spacing within the Dropdown trigger button between the text and arrow icon.
> > 
> > 
> > @KristinLBradley @dchyun I think we decided that the spacing between the text and the chevron specifically in the `AppHeader` should be reduced to 8px.
> 
> @dchyun I think you may want to add an `AppHeader` specific style override of the `Dropdown` button gap value then? (or whatever you think is the best way to handle it.)

Thanks for catching this @jorytindall @KristinLBradley. A dropdown toggle gap override was already present, but needed to add a specific override for the toggle icon padding. Should be as expected now.

KristinLBradley (approved):
Looking good! Thanks for fixing the Dropdown spacing issue.

packages/components/src/styles/components/app-header.scss
```diff
@@ -18,10 +95,9 @@
   align-items: center;
   justify-content: space-between;
   min-height: var(--token-app-header-height);
-  padding: 12px 16px 11px 16px; // bottom padding of 11px is to account for the bottom border
-  color: var(--token-color-foreground-high-contrast);
-  background: var(--token-color-palette-neutral-700);
-  border-bottom: 1px solid var(--token-color-palette-neutral-500);
+  padding: var(--token-app-header-padding-top) 16px var(--token-app-header-padding-bottom) 16px;
```
didoo: would make sense to have a `--token-app-header-padding-horizontal` too?

didoo: Also, looking at Figma (and the web component) it seems CDS uses `0` for the horizontal padding: 

Image

packages/components/src/styles/components/app-header.scss
```diff
@@ -7,7 +7,84 @@
...
 
-@use "../mixins/interactive-dark-theme" as *;
+@use "../mixins/carbonization" as *;
+@use "../mixins/focus-ring" as *;
+@use "../mixins/button" as *;
+
+$hds-interactive-app-header-theme-focus-ring-action-box-shadow:
+  inset 0 0 0 1px var(--token-app-header-interactive-focus-ring-color-action-internal),
+  0 0 0 3px var(--token-app-header-interactive-focus-ring-color-action-external);
+
```
didoo: I think this can be removed (see comment below)

```suggestion
```

packages/components/src/styles/components/app-header.scss
```diff
@@ -7,7 +7,84 @@
...
+// Used for interactive elements such as Button & Dropdown
+@mixin hds-interactive-app-header-theme($add-visible-border: true) {
+  // Override token used in focus ring mixin
+  --token-focus-ring-action-box-shadow: #{$hds-interactive-app-header-theme-focus-ring-action-box-shadow};
+
```
didoo: I don't think any of the focus rings uses anymore the box shadow, so this can be removed (but please double check)
```suggestion
```

packages/components/src/styles/components/app-header.scss
```diff
@@ -7,7 +7,84 @@
...
+  color: var(--token-app-header-interactive-foreground-color-default);
+  background-color: var(--token-app-header-interactive-surface-color-default);
+
+  // Default:
```
didoo: should these properties be grouped with the other above?

packages/components/src/styles/components/app-header.scss
```diff
@@ -7,7 +7,84 @@
...
+  border: 1px solid transparent;
+  border-radius: var(--token-button-border-radius);
```
didoo: not sure why this would be needed in this mixin. see if it can be removed, or if it needs to be specified only in a few specific places where the border radius doesn't come from the component itself (eg. Button or Dropdown)

packages/components/src/styles/components/app-header.scss
```diff
@@ -7,7 +7,84 @@
...
+  cursor: pointer;
+
+  @if ($add-visible-border) {
+    border-color: var(--token-app-header-interactive-border-color-default);
+  }
+
+  @include hds-apply-only-if-carbon() {
+    height: var(--token-app-header-height);
+  }
```
didoo, line 21: I have tried commenting this, and things looked OK, can you check if it's really needed?

packages/components/src/styles/components/app-header.scss
didoo: I have tried commenting the border-radius and the cursor, and things looked OK, can you check if they're really needed?

packages/components/src/styles/components/app-header.scss
```diff
@@ -7,7 +7,84 @@
...
+      border: none;
+    }
+  }
+
+  // Active:
+  &:active,
+  &.mock-active {
+    color: var(
+      --token-app-header-interactive-foreground-color-default
+    ); // to avoid overrides by specificity (eg. `a:hover`)
```
didoo:
```suggestion
    // to avoid overrides by specificity (eg. `a:hover`)
    color: var(
      --token-app-header-interactive-foreground-color-default
    );
```

packages/tokens/src/products/shared/app-header.json
```diff
@@ -1,22 +1,274 @@
...
+              "default": "{color.palette.neutral-600}",
+              "cds-g0": "{carbon.color.gray.30}",
+              "cds-g10": "{carbon.color.gray.30}",
+              "cds-g90": "{carbon.color.gray.60}",
+              "cds-g100": "{carbon.color.gray.70}"
+            }
+          }
+        }
+      },
+      "focus-ring": {
```
didoo, line 204: see comment above, _I think_ these can be removed

didoo (commented):
Did a first pass and left a few comments, but nothing major

packages/tokens/src/products/shared/app-header.json
dchyun: Updated to only have one size token

packages/components/src/styles/components/app-header.scss
dchyun: Added a token for both left and right, since we have separate top / bottom tokens

packages/components/src/styles/components/app-header.scss
dchyun: Upon looking into this, this border radius is needed for the home link to set the radius for the HDS theme. Since that is the only place it is needed I have moved it to that selector.

Screenshot 2026-06-03 at 9 58 06 AM

packages/components/src/styles/components/app-header.scss
dchyun: This is needed to align the height of the dropdowns with other items like the home link and icon buttons. Without it the dropdown does not stretch to the entire height of the app header for carbon themes as it is needed to do. Another alternative would be overriding the padding set on the dropdown to make it the right height but this was simpler to ensure all interactive elements are the same height.

Without height
Screenshot 2026-06-03 at 10 03 51 AM

With height
Screenshot 2026-06-03 at 10 04 03 AM

packages/tokens/src/products/shared/app-header.json
dchyun: I was able to remove `$hds-interactive-app-header-theme-focus-ring-action-box-shadow` and `--token-focus-ring-action-box-shadow`, but these tokens are still needed to set the colors for a "dark" focus ring on the HDS theme.

Elements on a dark background in the HDS theme have separate colors for the focus ring from those used for light bgs. These color overrides are captured in component tokens. 

Expected design
Screenshot 2026-06-03 at 10 24 40 AM

Without the external token the blue shade does not match Figma, and without the internal token there is no border
Screenshot 2026-06-03 at 10 38 24 AM

Another example of this is the CodeBlock which has component tokens for the internal and external colors as well.

Screenshot 2026-06-03 at 10 39 22 AM

didoo: Mmm, I fear this may become a real issue later. We're still playing and overriding colors in the HSD faux-dark theme, where we have actual dark themes in Carbon. Let me think about it

dchyun: Maybe we could have some foundational level solution instead of adding them within each component. Maybe something like in Figma where there are faux "dark" focus variables

Screenshot 2026-06-03 at 10 54 52 AM

packages/tokens/src/products/shared/app-header.json
```diff
@@ -1,22 +1,293 @@
...
+        "$modes": {
+          "default": "{color.palette.neutral-700}",
+          "cds-g0": "{color.page.primary}",
+          "cds-g10": "{color.page.primary}",
+          "cds-g90": "{color.page.primary}",
+          "cds-g100": "{color.page.primary}"
+        }
+      }
+    },
+    "interactive": {
```
didoo: looking at the proposed nomenclature, I think the `interactive` here should be a modifier (similar to what we have in the semantic `--token-surface-color-interactive` token). which means the colors above should be something like `base` or `default`?

packages/tokens/src/products/shared/app-header.json
```diff
@@ -1,22 +1,293 @@
...
+    "border": {
+      "color": {
+        "$type": "color",
+        "$value": "{color.palette.neutral-500}",
+        "$modes": {
+          "default": "{color.palette.neutral-500}",
+          "cds-g0": "{carbon.themes.white.borderSubtle00}",
+          "cds-g10": "{carbon.color.gray.30}",
+          "cds-g90": "{carbon.color.gray.70}",
+          "cds-g100": "{carbon.color.gray.80}"
```
didoo: [question] should the all be `{carbon.themes.***.borderSubtle00}`?

packages/tokens/src/products/shared/app-header.json
didoo: If we consider that we have (proposed/new nomenclature):

```
--token-surface-color-interactive
--token-surface-color-interactive-hover
--token-surface-color-interactive-active
--token-surface-color-interactive-disabled
```

maybe these color tokens for the AppHeader should be something like:

```
--token-app-header-surface-color
--token-app-header-border-color-interactive-default
--token-app-header-border-color-interactive-active
--token-app-header-foreground-color-interactive-default
--token-app-header-foreground-color-interactive-hover
--token-app-header-foreground-color-interactive-disabled
--token-app-header-surface-color-interactive-default
--token-app-header-surface-color-interactive-hover
--token-app-header-surface-color-interactive-active
--token-app-header-surface-color-interactive-disabled
```

and for the focus ring it's implicit that it's interactive (plus the qualifier is the `focus-ring` part) so maybe:

```
--token-app-header-focus-ring-color-action-internal
--token-app-header-focus-ring-color-action-external
```

what do you think?

packages/tokens/src/products/shared/app-header.json
didoo: OK, I have pushed a commit (as agreed) that should make the mixins and the focus rings work in a simpler way

didoo (commented):
OK, I have added a couple more comments, but I think at this point (if my commit is OK) we are down to some tokens nomenclature and nothing else

packages/tokens/src/products/shared/app-header.json
dchyun: I'm thinking maybe the problem here is `interactive` is not the right word. Because really this wasn't meant to mean `interactive` in the sense its a variant/modifier. It was more a catch all for all buttons (home link, dropdown, action button).

Using it as a modifier gets confusing. You could have `--token-app-header-border-color-base` and `--token-app-header-border-color-interactive-(active/default)`. But these are not related to the same element.

`border-color-base` sets the border color for the whole app header, and `border-color-interactive` sets it on the buttons. 

Originally I was using "interactive" for the group of interactive elements because that is what is used in the sass. What do you think if we renamed it to something else so it doesn't overlap with other tokens where we do use "interactive" as a modifier. We could name it "button" (`token-app-header-button-border-color`)? Or "interactive-item"?

didoo: I think `button` is OK

dchyun: Updated to use `button`

didoo (approved):
Well done!

PR #3829, opened 2026-04-24
Add tests for `hds-icon-registry` service

zamoore:
### :pushpin: Summary

If merged, this PR adds unit tests for the `hds-icon-registry` service.

### :link: External links

Jira ticket: [HDS-6118](https://hashicorp.atlassian.net/browse/HDS-6118)

***

<details>

</details>

[HDS-6118]: https://hashicorp.atlassian.net/browse/HDS-6118?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/showcase/tests/unit/services/hds-icon-registry-test.ts
+++ b/showcase/tests/unit/services/hds-icon-registry-test.ts
@@ -1,0 +1,404 @@
+/**
+ * Copyright IBM Corp. 2021, 2026
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { module, test } from 'qunit';
+import { setupTest } from 'ember-qunit';
+import { settled } from '@ember/test-helpers';
+import sinon from 'sinon';
+
+import HdsIconRegistryService, {
+  type HdsIconDefinition,
+} from '@hashicorp/design-system-components/services/hds-icon-registry';
+import {
+  HdsIconLibraryValues,
+  HdsIconSizeValues,
+  type HdsIconLoader,
+} from '@hashicorp/design-system-components/components/hds/icon/types';
+
+import type { IconName } from '@hashicorp/flight-icons/svg/index';
+
+const ROOT_ID = 'flight-sprite-empty-container';
+
+const makeSymbolIdFromDefinition = ({
+  library,
+  name,
+  size,
+}: HdsIconDefinition): string => {
+  const key =
+    library === HdsIconLibraryValues.Flight
+      ? `${library}-${name}-${size}`
+      : `${library}-${name}`;
+
+  return `hds-icon-${key.replace(/[^a-zA-Z0-9_-]/g, '-')}`;
+};
+
+const makeDefinition = (
+  overrides: Partial<HdsIconDefinition> = {},
+): HdsIconDefinition => ({
+  library: HdsIconLibraryValues.Flight,
+  name: 'alert' as IconName,
+  size: HdsIconSizeValues.Sixteen,
+  ...overrides,
+});
+
+const ensureRoot = (): SVGSVGElement => {
+  let root = document.getElementById(ROOT_ID) as SVGSVGElement | null;
+
+  if (root === null) {
+    root = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
+    root.id = ROOT_ID;
+
+    document.body.appendChild(root);
+  }
+
+  return root;
+};
+
+const removeRoot = (): void => {
+  document.getElementById(ROOT_ID)?.remove();
+};
+
+// awaits the full icon load: promises, DOM flush, and state updates
+const waitForRegistryCycle = async (): Promise<void> => {
+  // wait for icon loaders to resolve
+  await settled();
+
+  // wait for batched DOM injection via requestAnimationFrame
+  await new Promise<void>((resolve) => {
+    window.requestAnimationFrame(() => resolve());
+  });
+
+  // wait for final @tracked state updates
+  await settled();
+};
+
+interface HdsIconRegistryServicePrivate {
+  _resolveLoader(definition: HdsIconDefinition): HdsIconLoader;
+}
+
+const stubResolveLoader = (
+  service: HdsIconRegistryService,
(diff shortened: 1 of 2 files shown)


showcase/tests/unit/services/hds-icon-registry-test.ts
```diff
@@ -0,0 +1,404 @@
...
+} from '@hashicorp/design-system-components/services/hds-icon-registry';
+import {
+  HdsIconLibraryValues,
+  HdsIconSizeValues,
+  type HdsIconLoader,
+} from '@hashicorp/design-system-components/components/hds/icon/types';
+
+import type { IconName } from '@hashicorp/flight-icons/svg/index';
+
+const ROOT_ID = 'flight-sprite-empty-container';
```
didoo: [suggestion] you may consider re-exporting `ROOT_ID` in `hds-icon-registry` and avoid re-defining it here

showcase/tests/unit/services/hds-icon-registry-test.ts
didoo, line 1: Consider exporting `makeDomSafeId` and `makeSymbolIdFromKey` in the `hds-icon-registry` and add tests for these functions as well

showcase/tests/unit/services/hds-icon-registry-test.ts
```diff
@@ -0,0 +1,404 @@
...
+  library,
+  name,
+  size,
+}: HdsIconDefinition): string => {
+  const key =
+    library === HdsIconLibraryValues.Flight
+      ? `${library}-${name}-${size}`
+      : `${library}-${name}`;
+
+  return `hds-icon-${key.replace(/[^a-zA-Z0-9_-]/g, '-')}`;
```
didoo: if you export `makeDomSafeId` and `makeSymbolIdFromKey` from `hds-icon-registry.ts` then you can use hete here instead of the string concatenation and the regex (which is just a copy, so if one day we change the regex in that function, this test may pass anyway); plus you can test the functions themselves too (see comment above)

showcase/tests/unit/services/hds-icon-registry-test.ts
```diff
@@ -0,0 +1,404 @@
...
+    const symbolId = makeSymbolIdFromDefinition(definition);
+    const root = ensureRoot();
+    const existing = document.createElementNS(
+      'http://www.w3.org/2000/svg',
+      'symbol',
+    );
+
+    existing.id = symbolId;
+
+    root.appendChild(existing);
```
didoo: [thought] This case as is in this test is not realistic. I am not sure why we're using this approach (direct DOM injection of a symbol in the root container) instead of using the methods provided by the service itself

showcase/tests/unit/services/hds-icon-registry-test.ts
```diff
@@ -0,0 +1,404 @@
...
+      'warns once when sprite root is missing',
+    );
+    assert.strictEqual(
+      service.getSymbolId(definition),
+      null,
+      'icon remains unresolved when no sprite root exists',
+    );
+  });
+
+  test('it generates a symbol id key without size for carbon library icons', async function (assert) {
```
didoo: [nit]
```suggestion
  test('it generates a symbol id key without size for Carbon library icons', async function (assert) {
```

showcase/tests/unit/services/hds-icon-registry-test.ts
```diff
@@ -0,0 +1,404 @@
...
+    const definition = makeDefinition({
+      library: HdsIconLibraryValues.Carbon,
```
didoo, line 373: [issue?] I have changed this line to `library: HdsIconLibraryValues.Flight,` and the test was passing anyway, so I am not sure what this test is doing

showcase/tests/unit/services/hds-icon-registry-test.ts
didoo: I asked Copilot in my VSCode and that's what it told me:

> Given your confirmation that only the service populates the sprite root, these DOM checks are genuinely redundant — the in-memory _entries map is always the source of truth. If a symbol is in the DOM, the service put it there, meaning the entry already exists in _entries and requestLoad would have returned early on the this._entries.has(key) check anyway.
>
> The _symbolExists method and both of its call sites can be removed from the service, and this test can be deleted. The behavior it was testing — "don't re-load an already-present icon" — is already covered by the de-duplication test via the registry check.

So we should consider removing this test ("it uses an existing symbol in the DOM without invoking the loader") entirely.

didoo (commented):
Left a few comments

showcase/tests/unit/services/hds-icon-registry-test.ts
zamoore: That's fair - It was an edge case for simulating pre-existing sprite content outside of the service instance, but that's outside of an expected/typical use-case.

showcase/tests/unit/services/hds-icon-registry-test.ts
zamoore: I rewrote it to make it valuable instead of self-referential.

showcase/tests/unit/services/hds-icon-registry-test.ts
```diff
@@ -114,6 +101,34 @@ module('Unit | Service | hds-icon-registry', function (hooks) {
     assert.ok(service.getSymbolId.bind(service), 'getSymbolId exists');
   });
 
+  test('makeDomSafeId replaces non-DOM-safe characters', function (assert) {
+    assert.strictEqual(
+      makeDomSafeId('flight-alert-16'),
+      'flight-alert-16',
+      'keeps letters, numbers, underscores, and hyphens',
```
didoo, line 108: [nit] there is no underscore in the icon name, how can we be sure it does keep them?

didoo (approved):
Left only a small comment, non blocker

PR #3835, opened 2026-04-28
Use modern service imports in `showcase`

alex-ju:
### :pushpin: Summary

Replace the older import pattern `import { inject as service } from '@ember/service'` with the named `service` export across several components and the application template to align with the modern Ember API, improve consistency and avoid deprecation warnings.

***


didoo (approved):
👍

PR #3837, opened 2026-04-28
Suppress deprecation warnings for fastboot in `website`

alex-ju:
There is currently a lot of noise caused by the service inject deprecations, coming from various packages. This makes it very difficult to discover actual errors thrown by fastboot. I applied the `LOG_STACKTRACE_ON_DEPRECATION` config so fastboot output is much less noisy with deprecation warnings.

The alternative is patch those ember packages – I did try it locally, but as we update some of these ember dependencies we'll have to regenerate the patches.

***

### DIFF
--- a/website/config/environment.js
+++ b/website/config/environment.js
@@ -23,6 +23,7 @@ module.exports = function (environment) {
 
     EmberENV: {
       EXTEND_PROTOTYPES: false,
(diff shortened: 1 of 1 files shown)


didoo (approved):
thanks 🙏

PR #3838, opened 2026-04-28
[Project Solar / Phase 1 / Follow-up] Fix `hds-apply-only-if-carbon` Sass mixin

didoo:
### :pushpin: Summary

The current implementation of the `hds-apply-only-if-carbon()` Sass mixin, used to conditionally apply certain styles or overrides when a Carbon theme is applied, relies on the `.cds-g**` modes class names being applied to a parent element. This approach, though, doesn’t account for the case where the theme is `hds-theme-system` (which is applied to the root of the page without a corresponding `hds-mode-***` class like for the `hds-theme-light/dark` classes).

The reason for this very specific mixin implementation was that we were considering the possibility of nested modes; in reality up to now all the instances of `@include hds-apply-only-if-carbon()` are generic, not tied to a specific "mode".

This PR fixes this issue, updating the `hds-apply-only-if-carbon()` Sass mixin.

In this PR I have:
- updated the `hds-apply-only-if-carbon()` mixin to use `hds-theme-*` instead of `hds-mode-cds-*` as selectors
- updated the `ShwCarbonizationComparisonGridItem` to add `hds-theme-light/dark` to the containers’ class names
- re-generated the files in output

**Before:**
before

**After:**
after

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-6229

***

<details>

</details>


packages/components/src/styles/mixins/_carbonization.scss
```diff
@@ -5,29 +5,39 @@
...
-    }
+@mixin hds-apply-only-if-carbon() {
+  @at-root #{selector.nest(".hds-theme-light", &)},
+    #{selector.nest(".hds-theme-dark", &)},
+    #{selector.nest(".hds-theme-system", &)} {
+    @content;
   }
 }
+
+// NOTE: we don't expect this mixin to be used, but we're leaving it in place for now in case we need it later
```
KristinLBradley, line 16: Would you still want to keep this commented out one around as is or would you want to modify it similarly as above?

didoo: This is the old implementation, renamed to specify that is specific for "modes". But it's not used (and unlikely will ever be used, but I prefer to leave it for now, just in case)

packages/components/src/styles/mixins/_carbonization.scss
```diff
@@ -5,29 +5,39 @@
...
-    @at-root #{selector.nest(".hds-mode-cds-g0", &)},
-      #{selector.nest(".hds-mode-cds-g10", &)},
-      #{selector.nest(".hds-mode-cds-g90", &)},
-      #{selector.nest(".hds-mode-cds-g100", &)} {
-      @content;
-    }
+@mixin hds-apply-only-if-carbon() {
+  @at-root #{selector.nest(".hds-theme-light", &)},
+    #{selector.nest(".hds-theme-dark", &)},
+    #{selector.nest(".hds-theme-system", &)} {
```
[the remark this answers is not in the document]
didoo: **For future reference**: I have thought about it, and even if the comment when I've first read it made sense to me, in reality after some thinking and some testing I reached the conclusion that the current implementation works for the way in which we're going to use theming, and potentially nested theming.

The reason for this is that, in reality, the "modes" CSS selectors (`hds-mode-cds-g0/g10/g90/g100`) will never exist in a vacuum, but will only exist if the parent (root) element has a class `.hds-theme-light` or `.hds-theme-dark` applied to it, which means that the updated mixin implementation will work as expected. Nesting is not meant to add a Carbon-themed content inside a Classic-themed content, or viceversa. It's meant to nest a light (or dark) content inside a dark (or light) content, but in both cases they would be Carbon themes. And the fact that the mixin was called `hds-apply-only-if-carbon` and not `hds-apply-only-if-carbon-mode` is a "smell" (and the fact that up to now all the instances were not using a `$mode` argument, further confirms this point).

To make sure that my solution (and understanding of the problem) was correct, I have done a quick test (committed and then reverted) and it confirmed what I thought.

This is what the test looks like before the fix:
screenshot_6248

and this is what it looks like after the fix:
screenshot_6251

which is what should happen according to how the CSS styles and tokens are defined in the code.

So I am going to resolve the conversation/suggestion and move on with the PR review process.

packages/components/src/styles/mixins/_carbonization.scss
```diff
@@ -5,29 +5,39 @@
 
 @use "sass:selector";
 
-@mixin hds-apply-only-if-carbon($mode: false) {
```
dchyun, line 8: > in reality up to now all the instances of `@include hds-apply-only-if-carbon()` are generic, not tied to a specific "mode"

[Note] I have been close to adding some mode specific mixins in the past. I would classify these all as edge cases. Ex: Needing to set a hex code within a svg, hex codes within linear gradients. 

In each case so far I've been able to find a workaround to avoid the mode specific mixins but it potentially could come up again. I think it's fine for now to make this choice and reevaluate if there does seem to be a need for it. 

Another idea I had was what if we could have a `$theme` argument that only added styles for a light or dark theme, but idk how that would work with `system` since it could be light or dark.

dchyun (approved):

packages/components/src/styles/mixins/_carbonization.scss
didoo: Mmm but do you expect to have different values for different _modes_ (g0/g10/g90/g100) or different _themes_ (light/dark)? I suspect the latter. In any case, let's keep an eye on it.

PR #3839, opened 2026-04-29
`Tag` - Fix a runtime error when `ResizeObserver` is undefined

alex-ju:
### :pushpin: Summary

Fixed a runtime error in our Tag component when `ResizeObserver` is undefined (e.g. SSR in `website`)

We use observers in other parts of our component library, but the Tag component is the only place where it's not encapsulated within the backing class (for performance reasons). Because the observer sits outside the component instance, it gets picked up during SSR and throws a hard error. This change will result in a much cleaner console output when running `website`, helping us debug any further issues, without impacting our consumers.

***

### DIFF
--- a/.changeset/polite-beds-behave.md
+++ b/.changeset/polite-beds-behave.md
@@ -1,0 +1,9 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+<!-- START components/hds/tag -->
+
+`Tag` - Fixed a runtime error when `ResizeObserver` is undefined (e.g. SSR)
+
+<!-- END -->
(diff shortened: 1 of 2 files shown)


didoo (approved):
👍

.changeset/polite-beds-behave.md
```diff
@@ -0,0 +1,9 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+<!-- START components/hds/tag -->
```
dchyun, line 5:
```suggestion
<!-- START components/tag -->
```
@alex-ju just a note this path isn't quite right. We can update it during the release procedure before we generate the component entries

alex-ju: good spot! I relied on copilot to generate the changelog – certainly needs some guardrails

PR #3884, opened 2026-05-21
[Project Solar / Phase 1 / Follow-up] Focus ring refactoring for all components

didoo:
> [!NOTE]
> This PR is built on top of https://github.com/hashicorp/design-system/pull/3896 where the showcase pages have been updated to demo all the relevant focus ring variants for the components

### :pushpin: Summary

This PR is a large systematic refactoring of all the existing focus-ring / outline implementations for the HDS components, to support the carbonization of the components, and in particular the styling of the focus states

 👉 **Previews** 👀 : 
 - focus rings (single mode): https://hds-showcase-git-project-solar-phase-1hds-5695-307b74-hashicorp.vercel.app/foundations/focus-ring
 - carbonized focus rings (modes comparison): https://hds-showcase-git-project-solar-phase-1hds-5695-307b74-hashicorp.vercel.app/carbonization/foundations/focus-ring

In this PR I have:
- added handling of `mock-state-delay` in `addMockStateClasses` method (to support some cases where the DOM was manipulated after page rendering)
- simplified the tokens for `focus-ring`, removing some of the private tokens and using their declarations directly in the `box-shadow` definitions
- added two new mixins for `focus-ring`: `hds-focus-ring-basic` and `hds-focus-ring-advanced` (an older `hds-focus-ring-basic` was temporary renamed to avoid conflicts, and later removed)
- updated focus for different HDS components (see list of commits for details)
- removed the old mixins (not used/necessary anymore, once they have been replaced with the new ones)

> [!IMPORTANT]
> In the refactoring process I have introduced three simplifications (necessary to achieve the results in this PR):
> 
> - I have removed the complexity of the `&:focus / &:focus:not(:focus-visible) / &:focus-visible` definitions
> - I have generalized the focus states to use `&:focus-visible` only
> - I have removed the `&:focus:active` rules that removed the focus on `:active` states
>
> The result of these changes is that now the focus rings are visible only for keyboard-only interactions, similar to what is the behaviour in Carbon

#### Designers

You can look at the two preview pages:
 - focus rings (single mode): https://hds-showcase-git-project-solar-phase-1hds-5695-307b74-hashicorp.vercel.app/foundations/focus-ring
 - carbonized focus rings (modes comparison): https://hds-showcase-git-project-solar-phase-1hds-5695-307b74-hashicorp.vercel.app/carbonization/foundations/focus-ring
and see how the focus ring is implemented.

Since the focus ring is "mocked" (fixed) if you want to interact with the elements/components and see their actual behaviour, you will have to use the standard component pages and change themes/modes to see the differences (or their carbonization pages if you want to see the focus rings for `hds` and `cds` side by side.

> [!IMPORTANT]
> For the components that have not yet been “carbonized” (eg. the `StepperNav` or the `Form` elements or `Table/AdvancedTable`) the focus ring may not yet match what’s in the Figma designs; I have done my best to make the focus ring look nice, but the actual final implementation will be done when the components are completely “carbonized”

#### Engineers

There are two possible ways to review this PR:
- look at the changes commit by commit – I have kept the commits history as linear and clean as possible
- look at the overall changes – in this case you can have the overall picture of the changes introduced

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-5695

***

- ⚠️ Percy is not currently working, so I used local testing via before/after image diffs to capture unexpected regressions

<details>

</details>


packages/components/src/styles/@hashicorp/design-system-power-select-overrides.scss
```diff
@@ -38,11 +39,11 @@
...
     }
 
-    &:focus,
+    &:focus-visible,
     &.mock-focus {
       border-color: var(--token-color-focus-action-internal);
-      outline: 3px solid var(--token-color-focus-action-external);
-      outline-offset: 0;
+
+      @include hds-focus-ring-basic();
```
KristinLBradley, line 44: Generally, I believe we try to add the `@include` at top of regular style declrations. If there is a reason this needs to come after the border-color style in this case, could you add an explanatory comment?

packages/components/src/styles/@hashicorp/design-system-power-select-overrides.scss
KristinLBradley, line 1: Could you add an example of an Ember PowerSelect component using these style overrides to the focus ring examples in the Showcase?

packages/components/src/styles/components/form/radio-card.scss
```diff
@@ -55,10 +57,12 @@
     transition: var(--token-form-radiocard-transition-duration);
   }
 
+  // this is an exception, we use `focus-within` instead of `focus-visible`
   &:focus-within,
   &.mock-focus {
     border-color: var(--token-color-focus-action-internal);
```
KristinLBradley: The `RadioCard` is getting the same border-color style applied twice.

Image

Image

KristinLBradley:
Some of the components in the examples make use of nested `HdsLinkInline`, or `HdsButton`, etc. It could be useful to group these and compare to see which have styles overriding the default focus styles of those components vs. those that just rely upon the default focus styles of those nested components.

packages/components/src/styles/@hashicorp/design-system-power-select-overrides.scss
KristinLBradley: (There are other examples like this below btw. I won't comment on all.)

packages/components/src/styles/@hashicorp/design-system-power-select-overrides.scss
didoo: Good point, I've done it in https://github.com/hashicorp/design-system/pull/3896

packages/components/src/styles/components/form/radio-card.scss
didoo: Nice catch! Fixed

didoo:
> Some of the components in the examples make use of nested `HdsLinkInline`, or `HdsButton`, etc. It could be useful to group these and compare to see which have styles overriding the default focus styles of those components vs. those that just rely upon the default focus styles of those nested components.

@KristinLBradley  I don't think it's possible, in the sense that there are so many combinations that the only meaningful categorization is to keep the order similar to what we have on the index page (more or less, alphabetically, with a few exceptions here and there)

packages/components/src/styles/@hashicorp/design-system-power-select-overrides.scss
didoo: My "fear" is that if we add the properties after the focus mixin, we risk not noticing they're there.
Compare this:
screenshot_6392

With this:
screenshot_6393

 But maybe it's a "me" thing. @dchyun thoughts?

KristinLBradley: So for this one (and the similar examples below) I would just move the include above the border style like this:

```
    &:focus-visible,
    &.mock-focus {
      @include hds-focus-ring-basic();

      border-color: var(--token-color-focus-action-internal);
    }
```

Are you worried about inconsistency though?

jorytindall:
Thanks for putting all of these examples together, very helpful from a review perspective! Everything basically looks good from my view, but there are a couple of very subtle things that are worth mentioning.

Certain elements (I think those that are on a different Carbon surface color/layer color), there is a slight white (black on dark themes) border in addition to the focus outline. I think I'm largely noticing this in the components that we override in the HDS theme to create the faux "dark" theme (`CodeBlock`, `CodeEditor`) but also barely perceivable in the `RadioCard` and a few of the other form components as well.

**Showcase preview** 
Screenshot 2026-06-01 at 11 10 43 PM

**Figma**
Screenshot 2026-06-01 at 11 13 21 PM

In the `AppHeader`, there are a few issues.

The HDS theme home link does not have the inner lighter outline:

**Showcase preview**
Screenshot 2026-06-01 at 11 14 37 PM

**Figma**
Screenshot 2026-06-01 at 11 16 01 PM

There also seems to be an issue with the foreground color in these examples of the interactive elements (not directly related to the focus styles).
Screenshot 2026-06-01 at 11 14 24 PM

There's also something unexpected happening with the `AdvancedTable` context menu button as well. It appears that the outline is much thicker than in the other Carbon themes. We might also need to address the secondary button issue here as well, though it's strangely not a problem with the reorder handle.

Screenshot 2026-06-01 at 11 19 30 PM

didoo:
> Certain elements (I think those that are on a different Carbon surface color/layer color), there is a slight white (black on dark themes) border in addition to the focus outline. I think I'm largely noticing this in the components that we override in the HDS theme to create the faux "dark" theme (`CodeBlock`, `CodeEditor`) but also barely perceivable in the `RadioCard` and a few of the other form components as well.

@jorytindall this is something I want to discuss with you (designers). Many of these components with this double inner ring are derived from the HdsButton, and inherit this visual treatment. We _may/could potentially if absolutely necessary_ override this, but then we would end up again in the mess we were before (at least in code) of so many customizations and local exceptions that become a nightmare to maintain/refactor and get one's head around it.

As said, happy to discuss in a sync meeting once all the @hashicorp/hds-design designers have reviewed.

> In the `AppHeader`, there are a few issues.

Yes, I know. The idea is that it will be:
- fixed in https://github.com/hashicorp/design-system/pull/3819 if this PR gets merged first
- fixed in this PR if https://github.com/hashicorp/design-system/pull/3819 is merged first

Same for the Form components.

> There's also something unexpected happening with the `AdvancedTable` context menu button as well. It appears that the outline is much thicker than in the other Carbon themes. We might also need to address the secondary button issue here as well, though it's strangely not a problem with the reorder handle.

Let me have a look. OK, this was caused by the changes done in https://github.com/hashicorp/design-system/pull/3814 that were carried over in this branch after rebasing. Now it's fixed.

packages/components/src/styles/@hashicorp/design-system-power-select-overrides.scss
dchyun: I don't feel too strongly either way, but I would lean towards placing the `@include` at the top since that has been our pattern to-date. I don't foresee it taking up more than 4 or 5 lines and adding custom focus styles should not be a common occurrence given the focus ring mixins cover most things.

packages/components/src/styles/components/dropdown.scss
dchyun, line 1: [Issue] For the critical dropdown items there is not enough contrast between the item and the focus ring

Image
Image

packages/components/src/styles/mixins/_focus-ring.scss
```diff
@@ -7,122 +7,79 @@
...
-  $bottom: 0,
-  $left: 0,
-  $radius: var(--token-border-radius-medium),
-  $color: action,
-  $position: relative
+@mixin hds-focus-ring-advanced(
+  // pass false if you want to avoid (re)setting the element position
+  $element-position: relative,
+  $pseudo-element-z-index: -1,
+  // we assume a 1px border
```
dchyun, line 38: [Suggestion] Is it worth capturing these assumptions in variables? Or would that be overcomplicating things?

```
$element-border-width: 1px,
$pseudo-element-inset-hds: calc(-1 * #{$element-border-width}),
$pseudo-element-inset-cds: calc(#{$element-border-width} + 2px),
```

packages/components/src/styles/components/code-block/index.scss
```diff
@@ -144,6 +143,16 @@ $hds-code-block-code-footer-height: 48px;
   scrollbar-color: var(--token-color-palette-neutral-400) var(--token-color-palette-neutral-500);
   border-radius: inherit;
 
+  // special
```
dchyun:
```suggestion
```
[Nit] Can we remove this comment or elaborate on it?

packages/components/src/styles/components/code-editor/index.scss
```diff
@@ -108,32 +104,28 @@
     background-color: var(--token-code-block-button-surface-color-default);
     border: 1px solid var(--token-code-block-border-color-strong);
 
-    &:focus,
-    &.mock-focus,
     &:hover,
     &.mock-hover {
+      color: var(--token-code-block-foreground-color-primary);
```
dchyun, line 113:
```suggestion
```
[Issue] The hover state for the HDS actions buttons in the code editor has changed. There are a few styles not aligned to what they were before. One is the `color` which doesn't need to be set here. Another is the border color of the button needs to be a light blue.

Setting the hover border color to the focus color was not related to focus I believe but more a custom hover interaction for this component.

**Before**

Image

**After**

Image

KristinLBradley:
@didoo Will the icon colors be fixed once @dchyun's `AppHeader` work is merged in?

image

The focus ring colors also differ for the HDS theme in this branch from prod. Is it intentional?
image

@KristinLBradley yes, you can check in https://github.com/hashicorp/design-system/pull/3819 and [deployed preview](https://hds-showcase-8qncv1ogn-hashicorp.vercel.app/carbonization/foundations/focus-ring#appheader)

packages/components/src/styles/components/form/radio-card.scss
```diff
@@ -55,10 +57,10 @@
     transition: var(--token-form-radiocard-transition-duration);
   }
 
+  // this is an exception, we use `focus-within` instead of `focus-visible`
```
KristinLBradley, line 60: Maybe explain briefly why this is an exception?

packages/components/src/styles/components/breadcrumb.scss
```diff
@@ -183,7 +201,7 @@ $hds-breadcrumb-item-visual-horizontal-padding: 4px;
   @include hds-apply-only-if-carbon() {
     align-items: end;
     padding-bottom: 3px;
-    border: none;
+    border-color: transparent;
```
dchyun, line 204: [Issue] The breadcrumb truncation button now has a border on hover/active when it didn't previously. When this was `border: none` it would hide the border so it wasn't effected by the `border-color` set on the truncation button in the `hover` selector.

```
border-color: var(--token-color-border-strong);
```

Could either set this back to `border: none` or add a carbon only mixin inside the hover selector to make the border color transparent.

**Before**

Image

**After**

Image

packages/tokens/src/products/shared/tabs.json
```diff
@@ -29,7 +29,14 @@
         "radius": {
           "$type": "dimension",
           "$value": "5",
```
dchyun:
```suggestion
          "$value": "{border.radius.small}",
```
[Suggestion] Can use token value and remove modes

KristinLBradley:
@didoo **Nit**: Is it necessary to include the `ButtonSet` in the Showcases when it's just a layout container for standard HdsButtons, etc?

image

@KristinLBradley it's to see how the focus rings for different types of button align (in previous cases, when things were not correctly implemented, they were misaligned) and if the focus rings risk overlapping

dchyun:
> > In the AppHeader, there are a few issues.

> Yes, I know. The idea is that it will be:
> fixed in https://github.com/hashicorp/design-system/pull/3819 if this PR gets merged first
> fixed in this PR if https://github.com/hashicorp/design-system/pull/3819 is merged first

@jorytindall @didoo these issues have been fixed in #3819

packages/components/src/styles/components/dropdown.scss
jorytindall: This is coming directly from Carbon, so is that an issue for us to solve?

didoo:
> Will the icon colors be fixed once @dchyun's `AppHeader` work is merged in?

@KristinLBradley yes, you can check in [his branch](https://github.com/hashicorp/design-system/pull/3819) and [deployed preview](https://hds-showcase-8qncv1ogn-hashicorp.vercel.app/carbonization/foundations/focus-ring#appheader)

packages/components/src/styles/components/form/radio-card.scss
didoo: OK

packages/components/src/styles/mixins/_focus-ring.scss
didoo: probably OK to leave as is: the variables would be used only once, and would lead to more indirection

packages/components/src/styles/components/code-block/index.scss
didoo: honestly I don't remember why it was there. I've removed it

packages/tokens/src/products/shared/tabs.json
didoo: Good point, done.

didoo:
> **Nit**: Is it necessary to include the `ButtonSet` in the Showcases when it's just a layout container for standard HdsButtons, etc?

@KristinLBradley it's to see how the focus rings for different types of button align (in previous cases, when things were not correctly implemented, they were misaligned) and if the focus rings risk overlapping

packages/components/src/styles/components/breadcrumb.scss
didoo: Good catch. I've added a token to handle this, was the more consistent solution

packages/components/src/styles/components/code-editor/index.scss
didoo: Unfortunately there is a specificity battle between the styles for the `.hds-button`, the styles for the `.hds-copy-button`, and the styles for the `hds-code-editor__button` and `hds-code-editor__copy-button` 😱

I think I have it fixed now, can you check?

packages/components/src/styles/@hashicorp/design-system-power-select-overrides.scss
didoo: @KristinLBradley @dchyun see commit e23cbfaecd140bf559ce8d7ad3f5542ba6cd3a97 and let me know what you think.

packages/components/src/styles/components/code-editor/index.scss
dchyun: Issues with the hover state are fixed for the HDS theme. Now one difference is for the Carbon theme there is a white border on hover that wasn't there before.

**Before**

Image

**After**

Image

packages/components/src/styles/components/app-footer.scss
```diff
@@ -81,24 +83,30 @@ $hds-app-footer-icon-text-gap: 6px;
...
   }
 
-  &:active,
-  &.mock-active {
-    color: var(--app-footer-link-active-color);
+  &:focus-visible,
+  &.mock-focus {
+    color: var(--app-footer-link-foreground-color-focus);
+
+    @include hds-focus-ring-basic(
```
dchyun, line 96: [Nit] Can we move this `@include` above the selector style as well?

packages/components/src/styles/components/dropdown.scss
dchyun: Personally I would prefer to go away from Carbon to solve it since it's an a11y issue, but if we want to stick with Carbon anyway that's fine.

packages/components/src/styles/components/app-footer.scss
didoo: damn, I missed one! done.

packages/components/src/styles/components/code-editor/index.scss
didoo: Nice catch. Fixed now.

heatherlarsen:
The following is a list of discrepancies I found between the preview here and what's currently in Figma. 

1. `AppSideNav / List Item`: In the dark modes in Figma, the indicator color of an active list item is #0F62FE. In code, it's #78a9ff. I imagine the Figma one should be updated to match what's in code. 
2. `Button / Primary` & `Dropdown / Primary`: In the dark modes in Figma, the component doesn't match the preview or what is in Carbon. I imagine the Figma components should be updated to match what's in code. 
3. `CodeBlock` & `CodeEditor`: In code, all of the buttons use the double focus-ring, but in Figma, there is only a single focus ring. I'm not sure which is correct.
4. `Dropdown / Critical List Item`: In the dark modes in Figma, the critical list item remains using the blue focus ring, but it should match what's in code, which is the white focus ring.
5. `Standalone Link`: In code, the outline/border is 2px, but in Figma, it's 1px. Carbon also appears to use 1px. 
6. `Tabs`: In code in the dark modes, the bottom border of the active tab is using a blue color. The whole focus ring should be white. 
7. `Tag`: In Figma, the components use the double focus-ring, while in code, there's only a single focus-ring. Carbon uses an offset focus. I'm not sure which direction we should go in with this one though. 

cc @jorytindall @LilithJames-HDS Some of these likely need a Figma update instead of a code update. I tried to indicate which I think those were, but I would love to know what y'all think.

didoo:
> The following is a list of discrepancies I found between the preview here and what's currently in Figma.

@heatherlarsen @LilithJames-HDS @jorytindall + @dchyun see below for details/answers

> 1. `AppSideNav / List Item`: In the dark modes in Figma, the indicator color of an active list item is #0F62FE. In code, it's #78a9ff. I imagine the Figma one should be updated to match what's in code.

OK

> 2. `Button / Primary` & `Dropdown / Primary`: In the dark modes in Figma, the component doesn't match the preview or what is in Carbon. I imagine the Figma components should be updated to match what's in code.

OK.

> 3. `CodeBlock` & `CodeEditor`: In code, all of the buttons use the double focus-ring, but in Figma, there is only a single focus ring. I'm not sure which is correct.

The buttons in the  `CodeBlock`/`CodeEditor` are specialized `HdsButton[secondary]` or `CopyButton` (which is also a specialised `HdsButton[secondary]`) so they inherit their behaviour. We _could_ update their focus rings, but then we end up with a nightmare of inheritance and overrides that I would avoid as much as possible. TLDR: please align to the code implementation.

> 4. `Dropdown / Critical List Item`: In the dark modes in Figma, the critical list item remains using the blue focus ring, but it should match what's in code, which is the white focus ring.

OK

> 5. `Standalone Link`: In code, the outline/border is 2px, but in Figma, it's 1px. Carbon also appears to use 1px.

OK, fixed on the code side.

> 6. `Tabs`: In code in the dark modes, the bottom border of the active tab is using a blue color. The whole focus ring should be white.

I was not sure what to do (should I hide the indicator when focused?). I have aligned it now to having only a visible white focus ring even when is the active tab.

> 7. `Tag`: In Figma, the components use the double focus-ring, while in code, there's only a single focus-ring. Carbon uses an offset focus. I'm not sure which direction we should go in with this one though.

I was not sure which one to use, because they have two ways of applying the focus to the equivalent of our tag/dismiss and tag/text:
screenshot_6434
screenshot_6433
screenshot_6431

I have now updated the `cds` ring to have the offset, as you suggested.

dchyun:
> 3. CodeBlock & CodeEditor: In code, all of the buttons use the double focus-ring, but in Figma, there is only a single focus ring. I'm not sure which is correct.
> > The buttons in the CodeBlock/CodeEditor are specialized HdsButton[secondary] or CopyButton (which is also a specialised HdsButton[secondary]) so they inherit their behaviour. We could update their focus rings, but then we end up with a nightmare of inheritance and overrides that I would avoid as much as possible. TLDR: please align to the code implementation.

So a couple things on this one. First, in code @didoo the focus state still has a white border that I don't think should be there. It should just be a blue outline I believe. Second, @heatherlarsen in Figma the focus state has a white background which differs from the default state. Shouldn't the bg color match the default? Was there a reason to change it to white? 
**Code**
Screenshot 2026-06-04 at 9 07 30 AM
**Figma**
Screenshot 2026-06-04 at 9 04 37 AM

packages/components/src/styles/components/link/standalone.scss
```diff
@@ -31,7 +31,9 @@ $hds-link-standalone-sizes: ("small", "medium", "large");
     @include hds-focus-ring-advanced(
       // we want the focus ring to extend a bit around the content
       $pseudo-element-inset-hds: 0 -7px 0 -5px,
-      $pseudo-element-inset-cds: 0 -7px 0 -5px
+      $pseudo-element-inset-cds: 0 -7px 0 -5px,
+      // per design specs, we're using the same outline width as the normal link in Carbon (which is 1px)
```
dchyun, line 35: [Question] This isn't aligning as much to the equivalent carbon component as this was previously. I can't see a focus state in Figma, but was this the intended look and feel even if it's further from the Carbon component.

**Carbon component**

Image

**Previous HDS component - Carbon theme**

Image

**Current HDS component - Carbon theme**

Image

didoo:
> in code @didoo the focus state still has a white border that I don't think should be there. It should just be a blue outline I believe. 

@dchyun as mentioned above:

the `Button[secondary]` has an inner white border
screenshot_6440

the `CopyButton` is a `Button[secondary]` under the hood and as such has an inner white border because of this inheritance (you don't see it because the background is white, I've changed it here to make it more visible)
screenshot_6441

the `Copy` button in the `CodeEditor/Block` is a `CopyButton` under the hood who in turn is a `Button[secondary]` under the hood and as such has an inner white border because of this double inheritance; similarly all the other buttons in the `CodeEditor/Block` are also `Button[secondary]` buttons, which means... white border
screenshot_6442

Now _in theory_ we could override some styling of the focus ring in these cases, but I strongly advise against it because it creates a nightmare of special cases and overrides that is really hard to reason about and maintain (hence the three weeks it took me to get to this cleanup/rationalization).

Happy to discuss in our sync if you want

packages/components/src/styles/components/link/standalone.scss
didoo: Mmm, where did you see this above? 🤔

screenshot_6443

dchyun: My bad I guess I was looking at an outdated showcase preview link. Looks all good now.

didoo: OK, you mean in Figma. OK, never mind.

didoo: @LilithJames-HDS can you look into this? thanks

packages/components/src/styles/components/tag.scss
```diff
@@ -78,25 +78,46 @@
...
-
-  &:focus-visible::before {
-    box-shadow: none; // override default tooltip button focus styles
+    // we override the tooltip button focus ring (slighly different)
+    &::before {
+      border-radius: inherit;
+      inset: 0;
+
+      @include hds-apply-only-if-carbon() {
+        inset: -2px -2px -2px 1px;
```
dchyun:
```suggestion
        inset: -2px -2px -2px -2px;
```
[Suggestion] This inset needs to be -2px only when there is no dismiss.

Currently the focus ring looks right for when the dismiss is present, but when it isn't the outline isn't centered correctly.

Image

If it was updated to `-2px` it is centered.

Image

But this can't be done wholistically because then it doesn't look right when the dismiss is present. That case still needs to be `1px`.

Image

packages/tokens/src/products/shared/tabs.json
```diff
@@ -28,8 +28,15 @@
         },
         "radius": {
           "$type": "dimension",
-          "$value": "5",
-          "unit": "px"
+          "$value": "{border.radius.small}",
+          "unit": "px",
+          "$modes": {
```
dchyun: [Nit] We don't need `$modes` for this token since `border.radius.small` maps to 0x for carbon anyway

dchyun (commented):
A few small things but all good from me otherwise

packages/components/src/styles/@hashicorp/design-system-power-select-overrides.scss
KristinLBradley: Personally, I like it better and think the changes look good in the commit.

KristinLBradley:
Btw, I noticed that the browser tab/title for the Focus ring carbonization page doesn't include the word "carbonization" like the other Showcase page tabs do.

**Examples**:
_Typography carbonization page_:
image
image

_Focus ring carboniation page_:
image
image

packages/components/src/styles/components/dropdown.scss
KristinLBradley: If we had the white inner "border" on this one it would solve the contrast issue btw...

didoo:
> Btw, I noticed that the browser tab/title for the Focus ring carbonization page doesn't include the word "carbonization" like the other Showcase page tabs do.

Nice catch! Fixed :)

packages/tokens/src/products/shared/tabs.json
didoo: yep. fixed

packages/components/src/styles/components/tag.scss
didoo: OK should be fixed now

dchyun (approved):

PR #3918, opened 2026-06-10
`Toggle`: add aria-checked attribute to input

shleewhite:
### :pushpin: Summary

If merged, this PR would add a new modifier to the Toggle Base that keeps the aria-checked attribute in sync with the checked attribute. this is because ARIA 1.2 requires elements with role switch have aria-checked ([source](https://www.w3.org/TR/wai-aria-1.2/#switch)).

I did it this way to avoid creating a breaking change for consumers and so that all instances of toggle will be fixed after this is released. 

### :link: External links

Jira ticket: [HDS-6134](https://hashicorp.atlassian.net/browse/HDS-6134)

***

<details>

</details>

[HDS-6134]: https://hashicorp.atlassian.net/browse/HDS-6134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/components/hds/form/toggle/base.gts
+++ b/packages/components/src/components/hds/form/toggle/base.gts
@@ -16,7 +16,16 @@ export interface HdsFormToggleBaseSignature {
 export default class HdsFormToggleBase extends Component<HdsFormToggleBaseSignature> {
   private _syncAriaChecked = modifier((element: HTMLInputElement) => {
     const syncAriaChecked = () => {
-      element.setAttribute('aria-checked', element.checked ? 'true' : 'false');
+      const checkedValue = element.getAttribute('checked');
+      let ariaCheckedValue: 'true' | 'false';
+
+      if (checkedValue === 'mixed') {
+        ariaCheckedValue = 'false';
+      } else {
+        ariaCheckedValue = element.checked ? 'true' : 'false';
+      }
+
+      element.setAttribute('aria-checked', ariaCheckedValue);
     };
 
     syncAriaChecked();
--- a/packages/components/src/components/hds/form/toggle/base.gts
+++ b/packages/components/src/components/hds/form/toggle/base.gts
@@ -3,7 +3,8 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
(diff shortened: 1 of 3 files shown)


packages/components/src/components/hds/form/toggle/base.gts
```diff
@@ -12,7 +13,28 @@ export interface HdsFormToggleBaseSignature {
   Element: HTMLInputElement;
 }
 
-const HdsFormToggleBase: TemplateOnlyComponent<HdsFormToggleBaseSignature> =
+export default class HdsFormToggleBase extends Component<HdsFormToggleBaseSignature> {
+  private _syncAriaChecked = modifier((element: HTMLInputElement) => {
+    const syncAriaChecked = () => {
+      element.setAttribute('aria-checked', element.checked ? 'true' : 'false');
```
MelSumner: Question: would this handle `'mixed'` properly? Per the spec for aria-checked, it looks like it should be set to `'false'` in those scenarios.

packages/components/src/components/hds/form/toggle/base.gts
```diff
@@ -12,7 +13,28 @@ export interface HdsFormToggleBaseSignature {
...
+    };
+
+    syncAriaChecked();
+
+    const observer = new MutationObserver(syncAriaChecked);
```
MelSumner, line 33: I don't remember off the top of my head, so we should check to see if current Ember is okay with using observer, or if we should use the stopPropagation/preventDefault approach instead.

packages/components/src/components/hds/form/toggle/base.gts
shleewhite: Ill add that in!

packages/components/src/components/hds/form/toggle/base.gts
shleewhite: Code editor and advanced table use mutation observers, so I think it is ok?

zamoore (approved):

MelSumner (approved):

PR #3932, opened 2026-06-16
`Alert` - Add new generic content yielded blocks to match Figma (HDS-6287)

KristinLBradley:
### :pushpin: Summary

If merged, this PR:

* Adds new `GenericAbove` & `GenericBelow` yielded blocks
* Deprecated the existing `Generic` block

👉 **Preview**: https://hds-showcase-git-kristin-hds-6287-alert-generi-df250e-hashicorp.vercel.app/components/alert#actions

image

image

### :link: External links

* Jira ticket: [HDS-6287](https://hashicorp.atlassian.net/browse/HDS-6287)
* Figma file: https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/HDS-Components-v2.0?node-id=67382-70921&t=hjlIvesqawRNDpoe-0

***

<details>

</details>

[HDS-6287]: https://hashicorp.atlassian.net/browse/HDS-6287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/good-apples-chew.md
+++ b/.changeset/good-apples-chew.md
@@ -3,5 +3,5 @@
 ---
 
 <!-- START components/alert -->
-`Alert` - Deprecated the `Generic` yielded block which renders below actions and added `GenericAbove` and `GenericBelow` yielded blocks for including custom content either above or below `Alert` actions.
+`Alert` - Deprecated the `Generic` yielded block which renders below actions and added `GenericContent` and `GenericFooter` yielded blocks for including custom content either above or below `Alert` actions.
 <!-- END -->
--- a/.changeset/good-apples-chew.md
+++ b/.changeset/good-apples-chew.md
@@ -1,7 +1,7 @@
 ---
-"@hashicorp/design-system-components": patch
+"@hashicorp/design-system-components": minor
 ---
 
 <!-- START components/alert -->
-`Alert` - Changed position of generic yielded content within `Alert` to display above the actions instead of after
+`Alert` - Added `genericLayoutBottom` option with a default value of `true`. When set to `false`, it changes the position of generic yielded content within the `Alert` to display above the actions instead of after them at the bottom of the `Alert`
 <!-- END -->
--- a/.changeset/good-apples-chew.md
+++ b/.changeset/good-apples-chew.md
@@ -1,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`Alert` - Changed position of generic yielded content within `Alert` to display above the actions instead of after
--- a/.changeset/good-apples-chew.md
+++ b/.changeset/good-apples-chew.md
@@ -3,5 +3,5 @@
 ---
 
 <!-- START components/alert -->
-`Alert` - Added `genericLayoutBottom` option with a default value of `true`. When set to `false`, it changes the position of generic yielded content within the `Alert` to display above the actions instead of after them at the bottom of the `Alert`
+`Alert` - Deprecated the `Generic` yielded block which renders below actions and added `GenericAbove` and `GenericBelow` yielded blocks for including custom content either above or below `Alert` actions.
 <!-- END -->
--- a/.changeset/good-apples-chew.md
+++ b/.changeset/good-apples-chew.md
@@ -2,4 +2,6 @@
 "@hashicorp/design-system-components": patch
 ---
 
+<!-- START components/alert -->
 `Alert` - Changed position of generic yielded content within `Alert` to display above the actions instead of after
+<!-- END -->
--- a/packages/components/src/components/hds/alert/index.gts
+++ b/packages/components/src/components/hds/alert/index.gts
@@ -237,7 +237,6 @@ export default class HdsAlert extends Component<HdsAlertSignature> {
           {{yield (hash Description=HdsAlertDescription)}}
         </div>
 
-        {{yield (hash Generic=HdsAlertGenericDeprecated)}}
         {{yield (hash GenericAbove=HdsYield)}}
 
         <div class="hds-alert__actions">
@@ -249,6 +248,7 @@ export default class HdsAlert extends Component<HdsAlertSignature> {
           }}
         </div>
 
+        {{yield (hash Generic=HdsAlertGenericDeprecated)}}
         {{yield (hash GenericBelow=HdsYield)}}
       </div>
 
--- a/packages/components/src/components/hds/alert/index.gts
+++ b/packages/components/src/components/hds/alert/index.gts
@@ -4,7 +4,7 @@
  */
 
 import Component from '@glimmer/component';
-import { assert } from '@ember/debug';
+import { assert, deprecate } from '@ember/debug';
 import { guidFor } from '@ember/object/internals';
 import { tracked } from '@glimmer/tracking';
 import { eq } from 'ember-truth-helpers';
@@ -28,6 +28,26 @@ import HdsYield from '../yield/index.gts';
 import type { HdsAlertColors, HdsAlertTypes } from './types.ts';
 import type { HdsIconSignature } from '../icon/index.gts';
 
+class HdsAlertGenericDeprecated extends Component<{ Blocks: { default: [] } }> {
+  constructor(owner: Owner, args: Record<string, never>) {
+    super(owner, args);
+    deprecate(
+      'The "Generic" contextual component yielded by "Hds::Alert" is deprecated. Use "GenericAbove" or "GenericBelow" instead.',
+      false,
+      {
+        id: 'hds.alert.use-generic-above-or-generic-below',
+        until: '7.0.0',
+        for: '@hashicorp/design-system-components',
+        since: { available: '6.2.1' },
+      }
+    );
+  }
+  <template>
+    {{! template-lint-disable no-yield-only }}
+    {{yield}}
+  </template>
+}
+
 export const TYPES: HdsAlertTypes[] = Object.values(HdsAlertTypeValues);
 export const DEFAULT_COLOR: HdsAlertColors = HdsAlertColorValues.Neutral;
 export const COLORS: HdsAlertColors[] = Object.values(HdsAlertColorValues);
@@ -49,7 +69,6 @@ export interface HdsAlertSignature {
     type: HdsAlertTypes;
     color?: HdsAlertColors;
     icon?: HdsIconSignature['Args']['name'] | false;
-    genericLayoutBottom?: boolean;
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     onDismiss?: (event: MouseEvent, ...args: any[]) => void;
   };
@@ -58,7 +77,12 @@ export interface HdsAlertSignature {
       {
         Title?: typeof HdsAlertTitle;
         Description?: typeof HdsAlertDescription;
-        Generic?: typeof HdsYield;
+        /**
+         * @deprecated Use "GenericAbove" or "GenericBelow" instead.
+         */
+        Generic?: typeof HdsAlertGenericDeprecated;
+        GenericAbove?: typeof HdsYield;
+        GenericBelow?: typeof HdsYield;
         LinkStandalone?: WithBoundArgs<typeof HdsLinkStandalone, 'size'>;
         Button?: WithBoundArgs<typeof HdsButton, 'size'>;
       },
@@ -140,10 +164,6 @@ export default class HdsAlert extends Component<HdsAlertSignature> {
     }
   }
 
-  get genericLayoutBottom(): boolean {
-    return this.args.genericLayoutBottom ?? true;
-  }
-
   get classNames(): string {
     const classes = ['hds-alert'];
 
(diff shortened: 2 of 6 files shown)


MelSumner (dismissed):
:shipit:

heatherlarsen:
@KristinLBradley Would this cause an issue for anyone currently using this feature since it changes the location? Would it need to be released as a breaking change?

didoo (changes requested):
[issue/blocker] By changing the order in which the content inside the Alert (and Toast) is presented, this may change the "semantic" of the content presented to the user. We can't be sure of how consumers are using this block of content, so for me is a breaking change, we should not release this, unless we expose it via a (new) dedicated argument, or even better a new slot/named block, so it's not a breaking change and consumers can opt-in. (this will require changes also at Figma side, to have two slots)

KristinLBradley:
> [issue/blocker] By changing the order in which the content inside the Alert (and Toast) is presented, this may change the "semantic" of the content presented to the user. We can't be sure of how consumers are using this block of content, so for me is a breaking change, we should not release this, unless we expose it via a (new) dedicated argument, or even better a new slot/named block, so it's not a breaking change and consumers can opt-in. (this will require changes also at Figma side, to have two slots)

@didoo @shleewhite @heatherlarsen What would be the best way to handle this? I could add a new layout option, maybe `@layout-generic-above-actions=true` to control this. That's a pretty awkward name though so let me know if you have better ideas.

Or I could add a new yielded content slot. Maybe call it `GenericAbove` or something? That's a bit awkward maybe too though...

**UPDATE**: I added a new `@genericLayoutBottom` option which is `true` by default to match the current layout.

showcase/app/components/page-components/alert/sub-sections/actions.gts
```diff
@@ -64,8 +66,39 @@ const SubSectionActions: TemplateOnlyComponent = <template>
             <A.Generic>
               <div
                 class="shw-component-alert-sample-custom-content-after-actions"
-              >This for example could be extra text, specific for a special use
-                case.</div>
+              >This, for example, could be extra text
+                <strong>below the actions</strong>
+                for a special use case.</div>
```
didoo: [nit] once we expose the new API, this is not anymore a special case: both are valid

.changeset/good-apples-chew.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+<!-- START components/alert -->
+`Alert` - Added `genericLayoutBottom` option with a default value of `true`. When set to `false`, it changes the position of generic yielded content within the `Alert` to display above the actions instead of after them at the bottom of the `Alert`
```
shleewhite: Should we include a note that eventually the default will change?

showcase/app/components/page-components/alert/sub-sections/actions.gts
```diff
@@ -64,8 +66,39 @@ const SubSectionActions: TemplateOnlyComponent = <template>
...
+              aliqua.</A.Description>
+            <A.Button @text="Action" @color="secondary" />
+            <A.LinkStandalone
+              @icon="plus"
+              @text="Action"
+              @href="#"
+              @color="secondary"
+            />
+            <A.Generic>
+              <div
```
didoo: we tend to use the `<ShwPlaceholder>` when we want to indicate a "slot", it's easier to spot

packages/components/src/components/hds/alert/index.gts
```diff
@@ -139,6 +140,10 @@ export default class HdsAlert extends Component<HdsAlertSignature> {
     }
   }
 
+  get genericLayoutBottom(): boolean {
```
didoo: [nit] being a boolean, we tend to name these arguments with `is[something]`

.changeset/good-apples-chew.md
KristinLBradley: Will it? Anyway, changing the default seems like something we could add in a new changeset message when/if the default is later changed.

didoo (commented):
I have added a few comment, but I have to say this API doesn't fully convince me, something doesn't "click". 

What if instead we do something more radical, but also simpler: what if we yield two new named blocks, one called `genericBefore` and one called `genericAfter`, we deprecate the generic `generic`, and we remove it in the next major?

showcase/app/components/page-components/alert/sub-sections/actions.gts
KristinLBradley: So should I delete: "for a special use case"?

packages/components/src/components/hds/alert/index.gts
KristinLBradley: How about `@hasGenericLayoutBottom`?

Or can you think of another alternative?

showcase/app/components/page-components/alert/sub-sections/actions.gts
KristinLBradley: I can refactor this part once we decide on the naming/etc for the new option

showcase/app/components/page-components/alert/sub-sections/actions.gts
didoo: Yes, or even better (for me) use a `ShwPlaceholder` instead of the text (more visible/obvious that that is a slot)

packages/components/src/components/hds/alert/index.gts
didoo: Asked Copilot. `renderGenericSlot` = `"beforeActions" | "afterActions"` is not bad (we could even add more possible positions, in the future, if needed)
screenshot_6551

KristinLBradley: **Update**: I like the idea you posted above of adding `genericBefore` and `genericAfter` blocks plus deprecating the existing `generic` block so I could do something like that dependent on what @hashicorp/hds-design decides. (I missed that comment originally.)

KristinLBradley: `@genericSlotPlacement` for the naming if we go with adding a new argument sounds good to me. I will hold off making any new changes until there's a decision from @hashicorp/design-systems on what they want to do though.

KristinLBradley: **Update**: Following the [design decision](https://ibm-hashicorp.slack.com/archives/C09KTVB7GSX/p1781800542601829?thread_ts=1781640646.915309&cid=C09KTVB7GSX) I will go ahead and add 2 new yielded slots while deprecating the existing one.

KristinLBradley: New `genericBefore` and `genericAfter` blocks have been added along with deprecating the `generic` block.

showcase/app/components/page-components/alert/sub-sections/actions.gts
KristinLBradley: I've updated the Showcase to add examples of the new named blocks

showcase/app/components/page-components/alert/sub-sections/actions.gts
KristinLBradley: This has been refactored

.changeset/good-apples-chew.md
KristinLBradley: I updated the changelog mentioning the 2 new yielded blocks along with deprecation of the existing `generic` one.

packages/components/src/components/hds/alert/index.gts
```diff
@@ -212,6 +237,9 @@ export default class HdsAlert extends Component<HdsAlertSignature> {
           {{yield (hash Description=HdsAlertDescription)}}
         </div>
 
+        {{yield (hash Generic=HdsAlertGenericDeprecated)}}
```
didoo: [issue/blocker] If you move it here, it will be a breaking change
```suggestion
```

packages/components/src/components/hds/alert/index.gts
```diff
@@ -220,7 +248,8 @@ export default class HdsAlert extends Component<HdsAlertSignature> {
             )
           }}
         </div>
-        {{yield (hash Generic=HdsYield)}}
+
+        {{yield (hash GenericBelow=HdsYield)}}
```
didoo: [issue/blocker] it should be left in the same position to avoid a breaking change
```suggestion
        {{yield (hash Generic=HdsAlertGenericDeprecated)}}
        {{yield (hash GenericBelow=HdsYield)}}
```

didoo (changes requested):
There is a blocking issue (see comment).

Also, I am not convinced of the names `genericBefore`/`genericAfter`? before/after _what_? in theory they should be called `generic[Before|After]Actions` but that's too much, maybe.

So I was thinking: what if we use the names `genericContent`/`genericFooter` and a markup like this:

```
      <div class="hds-alert__content">
        <div
          class="hds-alert__text
            {{if
              (eq @type 'compact')
              'hds-typography-body-100'
              'hds-typography-body-200'
            }}"
        >
          {{yield (hash Title=HdsAlertTitle)}}
          {{yield (hash Description=HdsAlertDescription)}}
        </div>

        {{yield (hash GenericContent=HdsYield)}}

        <div class="hds-alert__actions">
          {{yield
            (hash
              Button=(component HdsButton size="small")
              LinkStandalone=(component HdsLinkStandalone size="small")
            )
          }}
        </div>

        {{yield (hash Generic=HdsAlertGenericDeprecated)}}
        {{yield (hash GenericFooter=HdsYield)}}

      </div>
```

We could even go a step further and have something like this:
```
      <div class="hds-alert__content">
        <div
          class="hds-alert__text
            {{if
              (eq @type 'compact')
              'hds-typography-body-100'
              'hds-typography-body-200'
            }}"
        >
          {{yield (hash Title=HdsAlertTitle)}}
          {{yield (hash Description=HdsAlertDescription)}}
        </div>

        <div class="hds-alert__generic-content">
          {{yield (hash GenericContent=HdsYield)}}
        </div>

        <div class="hds-alert__actions">
          {{yield
            (hash
              Button=(component HdsButton size="small")
              LinkStandalone=(component HdsLinkStandalone size="small")
            )
          }}
        </div>

        {{yield (hash Generic=HdsAlertGenericDeprecated)}}

        <div class="hds-alert__generic-footer">
          {{yield (hash GenericFooter=HdsYield)}}
        </div>

      </div>
```

where the elements `hds-alert__generic-content`/`hds-alert__generic-footer` have `display: contents` and we may even try to use something like:
```
.hds-alert__generic-content:not(:empty),
.hds-alert__generic-footer:not(:empty) {
   margin-top: xxx;
}
```
(in which case we should use the `~` to control the Ember whitespace)

KristinLBradley:
Cristiano:
> There is a blocking issue (see comment).
> 
> Also, I am not convinced of the names `genericBefore`/`genericAfter`? before/after _what_? in theory they should be called `generic[Before|After]Actions` but that's too much, maybe.
> 
> So I was thinking: what if we use the names `genericContent`/`genericFooter` and a markup like this:
> 
> ```
>       <div class="hds-alert__content">
>         <div
>           class="hds-alert__text
>             {{if
>               (eq @type 'compact')
>               'hds-typography-body-100'
>               'hds-typography-body-200'
>             }}"
>         >
>           {{yield (hash Title=HdsAlertTitle)}}
>           {{yield (hash Description=HdsAlertDescription)}}
>         </div>
> 
>         {{yield (hash GenericContent=HdsYield)}}
> 
>         <div class="hds-alert__actions">
>           {{yield
>             (hash
>               Button=(component HdsButton size="small")
>               LinkStandalone=(component HdsLinkStandalone size="small")
>             )
>           }}
>         </div>
> 
>         {{yield (hash Generic=HdsAlertGenericDeprecated)}}
>         {{yield (hash GenericFooter=HdsYield)}}
> 
>       </div>
> ```
> 
> We could even go a step further and have something like this:
> 
> ```
>       <div class="hds-alert__content">
>         <div
>           class="hds-alert__text
>             {{if
>               (eq @type 'compact')
>               'hds-typography-body-100'
>               'hds-typography-body-200'
>             }}"
>         >
>           {{yield (hash Title=HdsAlertTitle)}}
>           {{yield (hash Description=HdsAlertDescription)}}
>         </div>
> 
>         <div class="hds-alert__generic-content">
>           {{yield (hash GenericContent=HdsYield)}}
>         </div>
> 
>         <div class="hds-alert__actions">
>           {{yield
>             (hash
>               Button=(component HdsButton size="small")
>               LinkStandalone=(component HdsLinkStandalone size="small")
>             )
>           }}
>         </div>
> 
>         {{yield (hash Generic=HdsAlertGenericDeprecated)}}
> 
>         <div class="hds-alert__generic-footer">
>           {{yield (hash GenericFooter=HdsYield)}}
>         </div>
> 
>       </div>
> ```
> 
> where the elements `hds-alert__generic-content`/`hds-alert__generic-footer` have `display: contents` and we may even try to use something like:
> 
> ```
> .hds-alert__generic-content:not(:empty),
> .hds-alert__generic-footer:not(:empty) {
>    margin-top: xxx;
> }
> ```
> 
> (in which case we should use the `~` to control the Ember whitespace)

@didoo I updated the naming of the generic yielded blocks. I'm not sure about introducing new wrapper divs as that seems a bit beyond the scope of the ticket and could potentially introduce unforeseen issues such as with spacing.

didoo (approved):
👍

jorytindall (approved):

PR #3944, opened 2026-06-23
Website: actually render GTS code snippets in code previews

shleewhite:
### :pushpin: Summary

Currently, in the code group component - the preview is only ever the hbs version of the snippet. When you are on the gts snippet, it still renders the hbs version. This PR updates the component to actually render the gts component when the gts tab is active.

Currently, there is a difference in the Accordion size demo so it is easier to verify the new behavior, this will be reverted before the PR merges.

**[PREVIEW](https://hds-website-git-hds-5833-dynamic-template-refactor-hashicorp.vercel.app/components/accordion?tab=code#size)**

**Before**
Screenshot 2026-06-23 at 3 58 00 PM

**After**
Screenshot 2026-06-23 at 3 58 09 PM

### :link: External links

Jira ticket: [HDS-5833](https://hashicorp.atlassian.net/browse/HDS-5833)

***

<details>

</details>

[HDS-5833]: https://hashicorp.atlassian.net/browse/HDS-5833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/website/app/components/doc/code-group/index.gts
+++ b/website/app/components/doc/code-group/index.gts
@@ -120,15 +120,50 @@ export default class DocCodeGroup extends Component<DocCodeGroupSignature> {
     }
   }
 
-  // NOTE: the dynamic template requires a component to render a preview. If there is not a component (ex. only a sass/yaml/bash snippet), we hide the preview by default.
+  // NOTE: the dynamic template requires an Ember component to render a preview. If there is not a component (ex. only a sass/yaml/bash snippet), we hide the preview by default.
   get hidePreview() {
-    const hasRenderablePreview =
-      (this.args.hbsSnippet && this.args.hbsSnippet !== '') ||
-      (this.args.gtsSnippet &&
-        this.args.gtsSnippet !== '' &&
-        this.args.filename);
+    if (this.args.hidePreview === 'true') {
+      return true;
+    }
 
-    return !hasRenderablePreview || this.args.hidePreview === 'true';
+    const hasHbsPreview = this.hasSnippet(this.args.hbsSnippet);
+    const hasGtsPreview =
+      this.hasSnippet(this.args.gtsSnippet) && !!this.gtsPreviewComponentId;
+
+    return !(hasHbsPreview || hasGtsPreview);
+  }
+
+  private hasSnippet(snippet?: string) {
+    return !!snippet && snippet !== '';
+  }
+
+  get classicPreviewComponentId() {
+    return this.hasSnippet(this.args.filename) ? this.args.filename : undefined;
+  }
+
+  get gtsPreviewComponentId() {
+    return this.hasSnippet(this.args.gtsFilename)
+      ? this.args.gtsFilename
+      : undefined;
+  }
+
+  get previewMode() {
+    if (
+      this.currentView === 'gts' &&
+      this.hasSnippet(this.args.gtsSnippet) &&
+      this.gtsPreviewComponentId
+    ) {
+      return 'component-module';
+    }
+
+    if (
+      (this.currentView === 'hbs' || this.currentView === 'js') &&
+      this.hasSnippet(this.args.hbsSnippet)
+    ) {
+      return 'runtime-template';
+    }
+
+    return null;
   }
 
   get hbsSnippet() {
@@ -157,35 +192,21 @@ export default class DocCodeGroup extends Component<DocCodeGroupSignature> {
   }
 
   get previewTemplateString() {
-    // For .gts view, the preview comes from a real component module (via
-    // `@componentId`), so we intentionally skip passing a runtime template.
-    if (this.currentView === 'gts' && this.args.gtsSnippet !== '') {
-      return undefined;
-    }
-
-    return this.hbsSnippet === '' ? undefined : this.hbsSnippet;
+    return this.previewMode === 'runtime-template'
+      ? this.hbsSnippet
+      : undefined;
   }
 
   get previewComponentId() {
-    const { filename, gtsFilename } = this.args;
-
-    if (!filename) {
-      return undefined;
+    if (this.previewMode === 'component-module') {
+      return this.gtsPreviewComponentId;
     }
 
-    if (this.currentView === 'gts' && this.args.gtsSnippet !== '') {
-      if (gtsFilename && gtsFilename !== '') {
-        // .gts demos can live under a different component file than the .hbs
-        // classic demo, so prefer an explicit gts filename when provided.
-        return gtsFilename;
-      }
-
-      // Fallback: map `foo.classic` docs entries to the colocated modern
-      // component id (`foo`) used by the .gts demo implementation.
-      return filename.replace(/\.classic$/, '');
+    if (this.previewMode === 'runtime-template') {
+      return this.classicPreviewComponentId;
     }
 
-    return filename;
+    return undefined;
   }
 
   get currentSnippet() {
@@ -230,25 +251,28 @@ export default class DocCodeGroup extends Component<DocCodeGroupSignature> {
   };
 
   get shouldSyncLanguageSelection() {
-    return this.args.hbsSnippet !== '' && this.args.gtsSnippet !== '';
+    return (
+      this.hasSnippet(this.args.hbsSnippet) &&
+      this.hasSnippet(this.args.gtsSnippet)
+    );
   }
 
   get languageOptions() {
     const options: Array<LanguageOption> = [];
 
-    if (this.args.hbsSnippet !== '') {
+    if (this.hasSnippet(this.args.hbsSnippet)) {
       options.push({ label: '.hbs', value: 'hbs' });
     }
 
-    if (this.args.jsSnippet !== '') {
+    if (this.hasSnippet(this.args.jsSnippet)) {
       options.push({ label: '.js', value: 'js' });
     }
 
-    if (this.args.gtsSnippet !== '') {
+    if (this.hasSnippet(this.args.gtsSnippet)) {
       options.push({ label: '.gts', value: 'gts' });
     }
 
-    if (this.args.customLang && this.args.customSnippet !== '') {
+    if (this.args.customLang && this.hasSnippet(this.args.customSnippet)) {
       options.push({ label: `.${this.args.customLang}`, value: 'custom' });
     }
 
--- a/website/app/components/doc/code-group/index.gts
+++ b/website/app/components/doc/code-group/index.gts
@@ -23,6 +23,7 @@ import DynamicTemplate from 'website/components/dynamic-template';
 interface DocCodeGroupSignature {
   Args: {
     filename?: string;
+    gtsFilename?: string;
     hbsSnippet?: string;
     jsSnippet?: string;
     gtsSnippet?: string;
@@ -121,8 +122,13 @@ export default class DocCodeGroup extends Component<DocCodeGroupSignature> {
 
   // NOTE: the dynamic template requires a component to render a preview. If there is not a component (ex. only a sass/yaml/bash snippet), we hide the preview by default.
   get hidePreview() {
-    // TODO: refactor dynamic template to support gts components: https://hashicorp.atlassian.net/browse/HDS-5833
-    return this.args.hbsSnippet === '' || this.args.hidePreview === 'true';
+    const hasRenderablePreview =
+      (this.args.hbsSnippet && this.args.hbsSnippet !== '') ||
+      (this.args.gtsSnippet &&
+        this.args.gtsSnippet !== '' &&
+        this.args.filename);
+
+    return !hasRenderablePreview || this.args.hidePreview === 'true';
   }
 
   get hbsSnippet() {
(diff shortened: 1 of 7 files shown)


website/app/components/doc/code-group/index.gts
```diff
@@ -121,8 +122,13 @@ export default class DocCodeGroup extends Component<DocCodeGroupSignature> {
 
   // NOTE: the dynamic template requires a component to render a preview. If there is not a component (ex. only a sass/yaml/bash snippet), we hide the preview by default.
   get hidePreview() {
-    // TODO: refactor dynamic template to support gts components: https://hashicorp.atlassian.net/browse/HDS-5833
-    return this.args.hbsSnippet === '' || this.args.hidePreview === 'true';
+    const hasRenderablePreview =
+      (this.args.hbsSnippet && this.args.hbsSnippet !== '') ||
+      (this.args.gtsSnippet &&
+        this.args.gtsSnippet !== '' &&
+        this.args.filename);
```
didoo: [question] out of curiosity, why we need to check also `this.args.filename`?

website/app/components/doc/code-group/index.gts
```diff
@@ -150,6 +156,38 @@ export default class DocCodeGroup extends Component<DocCodeGroupSignature> {
...
+    // For .gts view, the preview comes from a real component module (via
+    // `@componentId`), so we intentionally skip passing a runtime template.
+    if (this.currentView === 'gts' && this.args.gtsSnippet !== '') {
+      return undefined;
+    }
+
+    return this.hbsSnippet === '' ? undefined : this.hbsSnippet;
+  }
+
+  get previewComponentId() {
```
didoo: I have asked Copilot to explain this logic, and visualize it in a matrix, because it was hard to me to understand it. This is the response (part of):

Image

Can the condition highlighted be true, in any way? My understanding is that you can have a gts snippet only if there is a gts file, and Copilot, when pressed, said this:

Image

So I am not sure if all this logic could be simplified in some way, by reducing the number of variables moved around, or maybe moved somewhere else and implement some other form of logic that decouples what to show from testing what is available.

I don't have a clear solution to this, because I am struggling to put all this pieces together in a single picture/flow.

website/app/components/dynamic-template.gjs
```diff
@@ -19,58 +19,78 @@ export default class DynamicTemplate extends Component {
...
+    if (factory?.class && typeof factory.class === 'function') {
+      return class extends factory.class {};
     }
 
-    let component = this.templateMap.get(templateString);
-    if (component === undefined) {
-      let compiledTemplate;
+    // if component couldn't be found the old way try importing it directly
+    let module;
+    try {
```
didoo: why is the `<DynamicTemplate>` trying to figure out which component to import, when we should know (_I think_) when we invoke it? or should this information be passed down to it, where we know it, instead of (literally) re-trying to figure out which of the two hbs/gts to render?

website/lib/markdown/markdown-process-demos.js
```diff
@@ -156,6 +156,7 @@ class MarkdownReplaceDemoBlocks extends Multifilter {
           };
 
           let fileNameToForward = '';
+          let gtsFileNameToForward = '';
```
didoo: I have tried to understand, with no luck, if it's really necessary to have a distinct `gtsFileNameToForward` or if instead we could simply reuse the `fileNameToForward` instead (we know the convention is to have the files to use the same name, only different extensions) and with that simplify the logic. 

If so, can you convince/explain me why? And in that case, should `fileNameToForward` be renamed to something like `classicFileNameToForward`?

also, nit: in some case it's spelled `filename` (lowecase `n`) in other `FileName` (uppercase `N`), maybe better to use the same convention?

didoo: Poke around with Copilot, and that's what it told me:

Image

Image

website/lib/markdown/markdown-process-demos.js
didoo, line 1: re. the naming of variables, should the names here be reviewed now, to make them more explicit/semantic?

```
const codeSnippets = {
  hbs: '',
  js: '',
  gts: '',
  compactGts: '',
  custom: '',
  customLang: '',
};
```

Something like:
```
const codeSnippets = {
  hbsCode: '',
  jsCode: '',
  gtsCode: '',
  gtsCodeCompact: '',
  customCode: '',
  customLang: '',
};
```

or 

```
const codeSnippets = {
  classicHbs: '',
  classicJs: '',
  modernGts: '',
  compactGts: '',
  customCode: '',
  customLang: '',
};
```

didoo: BTW, I just realized that maybe the answer is to just reuse these existing names?

```
interface DocCodeGroupSignature {
  Args: {
    filename?: string;
    gtsFilename?: string;
    hbsSnippet?: string;
    jsSnippet?: string;
    gtsSnippet?: string;
    compactGtsSnippet?: string;
    customSnippet?: string;
    customLang?: string;
    hidePreview?: 'true' | 'false';
    isExpanded?: 'true' | 'false';
  };
  Blocks: {
    default: [];
  };
}
```

(or update them to the new nomenclature you decide to use)

website/app/components/doc/code-group/index.gts
didoo: See comment here: https://github.com/hashicorp/design-system/pull/3944/changes#r3474054039

didoo (commented):
I have done a high-level review. My instinct tells me there is space for simplification of the code, expecially around the way in which we handle the different use cases/conditions, and for some clean-up too (naming things more consistently/semantically, adding comments to blocks of code that are not "obvious", etc)

If you want I can jump on a call, and we can discuss more in detail what I mean.

website/lib/markdown/markdown-process-demos.js
shleewhite: totally, I standardized based on the existing names

website/app/components/dynamic-template.gjs
shleewhite: I think this was a bad copilot rabbit hole I went down, I majorly simplified the code. The code group now passes the template string and the component id. if it is the template string, it does the "classic path" otherwise it does the "gts path".

component id is needed for both classic and gts paths because it is needed if there is any js for the component.

website/lib/markdown/markdown-process-demos.js
shleewhite: I fixed it so it name always does "filename" instead of "fileName"

as for passing both, I prefer to keep passing both to keep all the file name parsing/making assumptions about the file extensions in this file. previously weve been really explicit about passing the demo information to the code group and Id like to keep that pattern (ex. we could parse the compact gts code snippet in the code group, but opt to do it here instead).

As for renaming filename, sometimes it is the gts file name (like in cases where there is only a gts example - this happens on the getting started for engineers page). in the future when we dont support both anymore, id imagine wed just make it so filename is always the gts one and remove gtsFilename.

shleewhite: ok after thinking about it more, I made some more changes on the last point. now I just explicitly pass 'gtsFilename' and 'classicFilename' to make it as clear as possible.

website/app/components/doc/code-group/index.gts
```diff
@@ -119,10 +120,52 @@ export default class DocCodeGroup extends Component<DocCodeGroupSignature> {
...
+    }
+
+    const hasHbsPreview = this.hasSnippet(this.args.hbsSnippet);
+    const hasGtsPreview =
+      this.hasSnippet(this.args.gtsSnippet) && !!this.gtsPreviewComponentId;
+
+    return !(hasHbsPreview || hasGtsPreview);
+  }
+
+  private hasSnippet(snippet?: string) {
```
didoo: does this function need to be declared in the class? could it be a more general arrow function outside of the class? it's just returning if a string is defined and not empty

website/app/components/doc/code-group/index.gts
```diff
@@ -119,10 +120,52 @@ export default class DocCodeGroup extends Component<DocCodeGroupSignature> {
...
+      (this.currentView === 'hbs' || this.currentView === 'js') &&
+      this.hasSnippet(this.args.hbsSnippet)
+    ) {
+      return 'runtime-template';
+    }
+
+    return null;
   }
 
   get hbsSnippet() {
```
didoo: [question] I see that these functions (`hbsSnippet`, `gtsSnippet`, `jsSnippet`, `compactGtsSnippet`, `customSnippet`) are fundamentally doing the same thing; why not merge them in a single function and pass the snippet as argument, and the function just unescape it?

website/app/components/doc/code-group/index.gts
didoo: also, `!!snippet` already ensures `snippet !== ''` which means this function could really be checking only `!!snippet`, which means (**for me**) that it's not needed, it's an unnecessary extra abstraction, you can just check for `!![snippet]` wherever the function is  invoked

didoo (commented):
I left a few comments, but sorry to insist I still think there is work to do in terms of reducing the overhead in logic branching and getter. So, sorry if I copy verbatim what Copilot confirmed when asked, but 1) it express the issue more clearly of what I could do and 2) it confirms what my senses are telling me, which is there are too many abstractions/indirections/getters for something that probably could be much simpler/linear:

Image

Image

It also added this comments, which _I think_ make sense:

Image

shleewhite:
@didoo I have simplified the logic the way you described, I def agree that it is much clearer now.

website/app/components/doc/code-group/index.gts
```diff
@@ -119,59 +120,78 @@ export default class DocCodeGroup extends Component<DocCodeGroupSignature> {
...
 
   get currentSnippet() {
+    const {
+      jsSnippet,
+      gtsSnippet,
+      hbsSnippet,
+      customSnippet,
+      customLang,
+      compactGtsSnippet,
+    } = this.args;
```
didoo, line 167: [nit] let's keep the order the same as in other declarations
```suggestion
    const {
      hbsSnippet,
      jsSnippet,
      gtsSnippet,
      compactGtsSnippet,
      customSnippet,
      customLang,
    } = this.args;
```

website/app/components/doc/code-group/index.gts
```diff
@@ -119,59 +120,78 @@ export default class DocCodeGroup extends Component<DocCodeGroupSignature> {
...
   get hidePreview() {
-    // TODO: refactor dynamic template to support gts components: https://hashicorp.atlassian.net/browse/HDS-5833
-    return this.args.hbsSnippet === '' || this.args.hidePreview === 'true';
-  }
-
-  get hbsSnippet() {
-    const { hbsSnippet } = this.args;
-    return hbsSnippet ? unescapeCode(hbsSnippet) : '';
-  }
+    if (this.args.hidePreview === 'true') {
```
didoo, line 124: [suggestion/simplification] 
- the `isExpanded` argument is converted to a boolean in the constructor, if you convert also `hidePreview` there you can remove this string comparison.
- you could change the getter from a negative ("hide something") to a positive ("show something", like `renderPreview` or `previewShouldBeRendered`) and it would resolve the double negation below (`{{#if (notEq this.hidePreview true)}}`)

website/lib/markdown/markdown-process-demos.js
didoo: [todo] a couple of changes to this regex:
- the `.js` extension is not needed (`getComponentIdFromPath` is never used for `.js` files`)
- using `.classic.hbs` instead of `.hbs` leads to the two filenames be the same for `gts` and `hbs`
```suggestion
  /((?:getting-started|foundations|components|layouts|utilities|overrides|patterns|testing).*?)\.(?:classic\.hbs|gts)/;
```

didoo (commented):
I have left a few more comments. Plus, I am going to send you a diff to show how a single `filename` (well, `componentId`) for both `hbs` and `gts` would work (not a hill I am going to die for, but I think simplifies/rationalises a bit the code).

One other thing I noticed is that in the local environment I see this messages in the console (not in the deployed preview). Something we should  "fix"?

Image

website/app/components/doc/code-group/index.gts
```diff
@@ -22,7 +21,8 @@ import DynamicTemplate from 'website/components/dynamic-template';
 
 interface DocCodeGroupSignature {
   Args: {
-    filename?: string;
+    gtsComponentId?: string;
+    classicComponentId?: string;
```
didoo: [nit]
```suggestion
    classicComponentId?: string;
    gtsComponentId?: string;
```

website/lib/markdown/markdown-process-demos.js
```diff
@@ -147,15 +151,16 @@ class MarkdownReplaceDemoBlocks extends Multifilter {
...
+            hbsSnippet: '',
+            jsSnippet: '',
+            gtsSnippet: '',
+            compactGtsSnippet: '',
+            customSnippet: '',
             customLang: '',
           };
 
-          let fileNameToForward = '';
+          let classicComponentIdToForward = '';
```
didoo: [nit] why we need the `ToForward` too, in the variable name?

website/app/components/doc/code-group/index.gts
didoo: consider updating also the naming and logic upstream, so that instead of `hidePreview` you have `showPreview` (and the getter could be `renderPreview/shouldRenderPreview/isPreviewVisible/isPreviewRendered` to differentiate) and in the process maybe rename the arguments so they have the same name across the different forwardings (easier to track them), eg. `hidePreview="${shouldHidePreview}"` or `expanded="${isExpanded}"`

also, as mentioned above, maybe the `hidePreview/showPreview` argument could be converted to boolean in the constructor, similar to what it's done for the `isExpanded`

didoo (dismissed):
Left a few non-blocking comments. Up to you if you want to address them or not. In any case I am approving, the code is much better now, thanks. 👍

website/app/components/doc/code-group/index.gts
shleewhite: I think im going to leave it as is, isExpanded is a dynamic value so to it makes sense to convert it to a boolean in the constructor. hidePreview is only checked in this getter and its value doesnt change so im comfortable leaving it as a string check.

zamoore (approved):

didoo (approved):

PR #3950, opened 2026-06-25
MCP - Add flight-icons resource

zamoore:
### :pushpin: Summary

If merged, this PR adds Flight Icons MCP resources (`hds://icons` and `hds://icons/{iconName}`) backed by a new icon catalog store + schema/lookup layer, and registers them in MCP resource bootstrapping.

- Adds `packages/mcp/src/resources/flight-icons.ts` with:  
  - catalog index resource (`get_hds_icons`)  
  - per-icon resource (`get_hds_icon`)  
- Adds `packages/mcp/src/stores/flight-icons/`:  
  - `schema.ts` for catalog validation (Zod)  
  - `lookup.ts` for icon record shaping + normalized lookup keys  
  - `store.ts` for loading `@hashicorp/flight-icons/catalog.json`, metadata, lookup, and filtering/search support  
- Updates `packages/mcp/src/resources/index.ts` to register Flight Icons resources.  

Focus review on resource contract shape, lookup normalization behavior (icon name + filename), and not-found/error responses.

Jira ticket: [HDS-6484](https://hashicorp.atlassian.net/browse/HDS-6484)

***

<details>

</details>

[HDS-6484]: https://hashicorp.atlassian.net/browse/HDS-6484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/mcp/src/resources/flight-icons/get-icon-by-name.ts
```diff
@@ -0,0 +1,115 @@
...
+  limit = 100,
+): string[] => {
+  const query = value.trim().toLowerCase();
+  const matches: string[] = [];
+
+  for (const summary of store.listIcons()) {
+    const icon = store.getIconByName(summary.iconName);
+    const aliases = [
+      summary.iconName,
+      ...(icon?.variants.map((variant) => variant.fileName) ?? []),
```
dchyun, line 39:
```suggestion
```
[Suggestion] Since the `fileName` is the same as the `iconName` just with `-24` or `-16` maybe it's easier to only search by the `iconName`.

packages/mcp/src/resources/flight-icons/get-icon-by-name.ts
dchyun, line 1: [Suggestion] It could be helpful to have a `get-icon-by-description` call as well. `get-icon-by-name` only returns icons which match a name exactly. We could leverage the existing `description` field in the catalog to also have a mcp call which searches by those keywords. This would be similar to the [icon library search](https://helios.hashicorp.design/icons/library) we have on the website which can search by exact icon name or through those keywords.

packages/mcp/src/resources/index.ts
```diff
@@ -5,10 +5,12 @@
 
 import { withSafeResourceHandler } from "./utils.js";
 
+import flightIconsResources from "./flight-icons/index.js";
```
dchyun, line 8:
```suggestion
import flightIconsResources from "./flight-icons/index";
```

[Question] A lot of the imports across the mcp packages suffix their imports with `.js` when all the files are `.ts`. I think we could remove this `.js` since they are actually `ts` files.

packages/mcp/tests/unit/resources/flight-icons/utils.test.ts
```diff
@@ -0,0 +1,45 @@
...
+describe("Flight icon serialization", () => {
+  const icon = toIconRecord([buildIconAsset()]);
+
+  it("serializes an explicit summary field set", () => {
+    expect(toSerializableIconSummary(icon)).toStrictEqual({
+      iconName: "alert-triangle",
+      description: "alert, warning, caution",
+      category: "Alerts",
+      sizes: ["16"],
+      hasMapping: false,
```
dchyun, line 26: [Suggestion] Could you test for an icon with mapping as well?

packages/mcp/tests/unit/resources/flight-icons/store/index.test.ts
```diff
@@ -0,0 +1,96 @@
...
+        limit: 10,
+        category: " ALERTS ",
+        size: "24",
+        hasMapping: true,
+      }),
+    ).toHaveLength(1);
+    expect(
+      store.searchIcons({ query: "arrow", limit: 10, hasMapping: true }),
+    ).toHaveLength(0);
+    expect(store.searchIcons({ query: "", limit: 1 })).toHaveLength(1);
```
dchyun, line 94:
```suggestion
    expect(store.searchIcons({ query: "", limit: 1 })).toHaveLength(1);
    expect(store.searchIcons({ query: "warning", size: "99" })).toHaveLength(0);
```
[Suggestion] Can add a test for if an icon search doesn't have a given size

dchyun (commented):
Tested with the inspector and all the queries seem to be working as expected.

packages/mcp/src/resources/flight-icons/get-icon-by-name.ts
zamoore: Keeping fileName lookup is basically free (a few extra Map keys, built once) and it covers the case where an LLM asks for `alert-triangle-16`.

packages/mcp/src/resources/flight-icons/get-icon-by-name.ts
zamoore: Good call, but this is better modeled as a tool than a resource since it takes multiple input parameters (query, category, size, etc) and computes a filtered result rather than reading a known URI. We can add a tool as a followup for sure

packages/mcp/src/resources/index.ts
zamoore: It's required in this case because this is an ESM project. The path has to point to the built file.

packages/mcp/tests/unit/resources/flight-icons/utils.test.ts
zamoore: Good call - Added

dchyun (approved):

PR #3954, opened 2026-06-30
[Project Solar / Phase 2 / Tokens renaming] Rename `--hds-***` CSS variables to `--hds-var-***`

didoo:
> [!NOTE]
> The markdown files used with and/or generated by Copilot are left in the PR only for reference; they will be removed before merging the branch.

### :pushpin: Summary

In the RFC for the new design tokens nomenclature we have decided to change the prefix of all our design tokens from `--token-***` to `--hds-***`

The reason is not just for consistency with other entities (which usually are prefixed with `hds/HDS`) but also because in this way we can immediately spot tokens used in the consumers' codebase that need to be updated (new prefix + potentially different token name to follow the convention).

Now, to avoid confusion with tokens coming from the pipeline (which will have the `--hds` prefix) we want to rename all the existing CSS variables currently used in code (in SCSS/JS/TS/GTS files) that currently have a `--hds` prefix and use instead a `--hds-var` prefix, to distinguish them from the normal tokens (the `var` part becomes a new second-level namespace) . 

In this PR I have:
- discussed and drafted an initial prompt (with Copilot) to use as starting point for the planning session
- started to work on the actual planning session, but after closer analysis realized that the `--hds` variables used on the consumers codebase are just a dozen or so ([reference](https://github.com/search?q=%2F--hds-%2F+user%3Ahashicorp+repo%3Ahashicorp%2Fcloud-ui++OR+repo%3Ahashicorp%2Fatlas&type=code&ref=advsearch)), which means a simple find&replace will be the most pragmatic and simple thing to do, instead of implementing complex codemods and AI skills/prompts to do the replacement; so I pivoted the prompting and asked to focus only on the migration within the HDS monorepo
- finalized the plan (with Copilot) and executed it (via Copilot); a report was generated about the outcome of the execution
- as a result, all the `--hds` variables have been renamed to `--hds-var` (under `packages/components`, `showcase`, `website`)

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-6493

***

<details>

</details>


MelSumner (commented):
seems okay to merge

website/docs/layouts/flex/partials/code/code-snippets/flex-custom-gap-complex.classic.hbs
dchyun, line 1: [Issue] Caught one unchanged `--hds` reference in `website/docs/layouts/flex/partials/code/how-to-use.md`

Image

website/docs/layouts/grid/partials/code/code-snippets/grid-spacing-custom-complex.classic.hbs
dchyun, line 1: [Issue] Caught one unchanged `--hds` reference in `website/docs/layouts/grid/partials/code/how-to-use.md`

Image

.changeset/twenty-ads-travel.md
```diff
@@ -0,0 +1,155 @@
...
+- `hds-app-side-nav-animation-easing` → `hds-var-app-side-nav-animation-easing`
+- `hds-app-side-nav-toggle-button-width` → `hds-var-app-side-nav-toggle-button-width`
+- `hds-app-side-nav-width-expanded` → `hds-var-app-side-nav-width-expanded`
+- `hds-app-side-nav-width-fixed` → `hds-var-app-side-nav-width-fixed`
+- `hds-app-side-nav-width-minimized` → `hds-var-app-side-nav-width-minimized`
+<!-- END -->
+
+<!-- START components/code-block -->
+`CodeBlock` - Changed the component CSS custom property prefix from `--hds-*` to `--hds-var-*`.
+- `hds-code-block-color-atrule` → `hds-var-code-block-color-atrule`
```
dchyun:
```suggestion
- `hds-code-block-color-{property}` → `hds-var-code-block-color-{property}`
```
[Nit][Suggestion] Since there are so many listed here maybe worth it to condense them to one line instead of listing them all.

packages/components/src/styles/components/app-side-nav/vars.scss
```diff
@@ -13,21 +13,21 @@
 
 :root {
   // widths
-  --hds-app-side-nav-width-minimized: 16px;
-  --hds-app-side-nav-width-expanded: 280px;
-  --hds-app-side-nav-width-fixed: var(--hds-app-side-nav-width-expanded);
+  --hds-var-app-side-nav-width-minimized: 16px;
```
dchyun, line 16: [Suggestion][Non-blocker] Not related to this PR, but I reviewed what `-hds` tokens are [used by consumers](https://github.com/search?q=org%3Ahashicorp%20--hds-&type=code) and there are only three sets of variables used

- `--hds-code-block-color`
  - These references could be replaced by the new `--token-code-block` variables
- `--hds-elevation-inset-box-shadow`
  - Not sure why this is being referenced as it's not an existing HDS variable
- `--hds-app-side-nav-width-minimized` and `--hds-app-side-nav-width-expanded`

Maybe it would be worth it to add a component tokens for the app side nav minimized and expanded widths. Then we could update consumers to point to those tokens package variables and have a general rule that consumers should not be using our internal `--hds-var` variables.

.changeset/twenty-ads-travel.md
didoo: Yeah, makes sense! 👍

website/docs/layouts/flex/partials/code/code-snippets/flex-custom-gap-complex.classic.hbs
didoo: Fixed. Thanks

website/docs/layouts/grid/partials/code/code-snippets/grid-spacing-custom-complex.classic.hbs
didoo: Fixed. Thanks.

packages/components/src/styles/components/app-side-nav/vars.scss
didoo: > `--hds-code-block-color` These references could be replaced by the new --token-code-block variables

OK

> `--hds-elevation-inset-box-shadow` Not sure why this is being referenced as it's not an existing HDS variable

Yeah, I saw it too, probably a bug/typo

> there are only three sets of variables used

Yes, but even if not used by the consumers the `--hds-(var)-layout-[flex|grid]-[row| column]-gap` are officially public APIs, so we can't really declare it a rule.

Also, considered that these variables are used only in a bunch of places, and they're easy to spot (and if needed, refactor) I would not worry too much about them: we can consider them acceptable exceptions (in the grand scheme of things)

dchyun (approved):

PR #3968, opened 2026-07-02
[Spike] reproduce modal close issue

shleewhite:
### :pushpin: Summary

This came up when @pkj-web was using Helios for his project. We noticed that when the Modal wasn't within one of the yielded AppFrame components, it would work but scroll isn't properly restored when it closes. The fix is to either put it within a yielded component (like AF.Main) or to move it outside the AppFrame entirely.

Not sure this is a bug or not and if this is something we should address (I can't think of a time when the Modal would be used in this way in a consumer's app), but thought I would make a reproduction so we can discuss.

To reproduce the issue:
* Go to: https://hds-showcase-git-spike-modal-app-frame-hashicorp.vercel.app/layouts/app-frame/frameless/demo-full-app-frame-with-functional-modal
* Close the modal
* Notice you cant scroll the page

***

<details>

</details>

### DIFF
--- a/packages/components/src/components/hds/app-frame/index.gts
+++ b/packages/components/src/components/hds/app-frame/index.gts
@@ -68,24 +68,23 @@ export default class HdsAppFrame extends Component<HdsAppFrameSignature> {
 
   <template>
     <div class={{this.classNames}} ...attributes>
-      {{#if this.hasHeader}}
-        {{yield (hash Header=HdsAppFrameHeader)}}
-      {{/if}}
-      {{#if this.hasSidebar}}
-        {{yield (hash Sidebar=HdsAppFrameSidebar)}}
-      {{/if}}
-      {{!
+      {{!--
         IMPORTANT: since the modals may be injected via portal or in-element with code that lives in the "main" container,
         the "modal" container needs to be present in the DOM _before_ the "main" block, otherwise it may cause errors
         where the target DOM element is not found (for example in tests where the modal may be immediately opened on first render).
+
+        All sub-components are yielded in a single hash to avoid multiple {{yield}} calls, which would cause any
+        content placed directly in the block (e.g. a modal with {{#if}}) to be instantiated once per yield.
+      --}}
+      {{yield
+        (hash
+          Header=HdsAppFrameHeader
+          Sidebar=HdsAppFrameSidebar
+          Modals=HdsAppFrameModals
+          Main=HdsAppFrameMain
+          Footer=HdsAppFrameFooter
+        )
       }}
-      {{#if this.hasModals}}
-        {{yield (hash Modals=HdsAppFrameModals)}}
-      {{/if}}
-      {{yield (hash Main=HdsAppFrameMain)}}
-      {{#if this.hasFooter}}
-        {{yield (hash Footer=HdsAppFrameFooter)}}
-      {{/if}}
     </div>
   </template>
 }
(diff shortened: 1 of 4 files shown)


didoo:
@shleewhite probably a bug, I see a spurious inline `overflow: hidden` style still applied to the `body` element even after the modal has been closed.

screenshot_6682

shleewhite:
Through debugging with Bob, we found that because the AppFrame has 5 yield statements, it meant that the Modal was actually being rendered 5 times.

From Bob:
> The template had 5 separate {{yield}} calls — one per sub-component, each gated by a {{#if hasX}}. In Glimmer, every {{yield}} re-executes the entire caller's block. So placing <HdsModal> as a direct (non-<Frame.Modals>) child of <HdsAppFrame> caused 5 independent HdsModal instances to be created, each calling open() and locking document.body with overflow: hidden, and each recording _bodyInitialOverflowValue = "hidden" (after the first one already set it). When the modal closed, only one instance's registerOnCloseCallback ran and correctly removed overflow: hidden — but the 4 remaining teardowns each ran with _isOpen=true, called _performCloseCleanup(), and re-set overflow back to "hidden" (because _bodyInitialOverflowValue was "hidden" when they were set up).

> The fix: collapse all 5 {{yield}} calls into a single {{yield}} with all sub-components in one hash. The block now executes exactly once, so any content placed directly inside it (like a conditionally-rendered <HdsModal>) is instantiated exactly once.

Ultimately, it is not a bug with the modal lifecycle, but the AppFrame.

If you collapse it to just one yield, the bug is fixed. Not sure if this is an approach we want to take though as it changes the AppFrame component quite a bit.

I pushed the fix + extra console logs in the modal to show the issue. If you pull the branch locally and revert the change to the appframe, you can see that the events are fired a bunch of times.

didoo:
> The template had 5 separate {{yield}} calls — one per sub-component, each gated by a {{#if hasX}}. In Glimmer, every {{yield}} re-executes the entire caller's block. So placing  as a direct (non-<Frame.Modals>) child of  caused 5 independent HdsModal instances to be created

@shleewhite @pkj-web yes, that's what I thought as well, when I looked (time ago) at the PR: there is a specific `<Frame.Modals>` container for this kind of element, on purpose (see also the HDS docs [here](https://helios.hashicorp.design/layouts/app-frame#modals-container)). I don't think is a bug, and I don't think it should be fixed (if someone has to use a modal that is _not_ contained in the appframe, they just have to inject it as sibling of the appframe.

shleewhite:
> > The template had 5 separate {{yield}} calls — one per sub-component, each gated by a {{#if hasX}}. In Glimmer, every {{yield}} re-executes the entire caller's block. So placing  as a direct (non-<Frame.Modals>) child of  caused 5 independent HdsModal instances to be created
> 
> @shleewhite @pkj-web yes, that's what I thought as well, when I looked (time ago) at the PR: there is a specific `<Frame.Modals>` container for this kind of element, on purpose (see also the HDS docs [here](https://helios.hashicorp.design/layouts/app-frame#modals-container)). I don't think is a bug, and I don't think it should be fixed (if someone has to use a modal that is _not_ contained in the appframe, they just have to inject it as sibling of the appframe.

@didoo I totally agree, it is using the app frame incorrectly.

MelSumner:
A little food for thought; for a11y purposes, the modal should be next the element that triggers it (in the DOM).

didoo:
> A little food for thought; for a11y purposes, the modal should be next the element that triggers it (in the DOM).

@MelSumner I can't find any reference to proximity of the DOM elements: https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/ can you point me to where this requirement is specified? thanks

shleewhite:
Closing as we have determined this is not actually a bug.

PR #3988, opened 2026-07-08
[Project Solar / Phase 1 / Migration] `Table` & `AdvancedTable` carbonization

shleewhite:
### :pushpin: Summary

If merged, this PR would carbonize the `Table` component and adds the relevant component tokens. It also adds a new state for the row that toggles the isSelected class.

* [Table carbonization page](https://hds-showcase-git-project-solar-phase-1shleewhi-7e3d13-hashicorp.vercel.app/carbonization/components/table)
* [AdvancedTable carbonization page](https://hds-showcase-git-project-solar-phase-1shleewhi-7e3d13-hashicorp.vercel.app/carbonization/components/advanced-table)

Note: for reorderable columns, the HTML element to show the preview column as you're dragging is appended to the DOM so it doesn't show as themed on the carbonization page. To validate that it works correctly, check the normal [AdvancedTable showcase page](https://hds-showcase-git-project-solar-phase-1shleewhi-7e3d13-hashicorp.vercel.app/components/advanced-table) and change the theme.

Read through similar carbonization pages under showcase/app/components/page-carbonization and compare those to the changes made here to ensure this PR aligns with previous carbonization efforts

### :link: External links

Jira ticket: [HDS-6104](https://hashicorp.atlassian.net/browse/HDS-6104)

***

<details>

</details>

[HDS-6104]: https://hashicorp.atlassian.net/browse/HDS-6104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/tokens/src/products/shared/table.json
```diff
@@ -0,0 +1,360 @@
...
+        }
+      },
+      "surface": {
+        "color": {
+          "hover": {
+            "$type": "color",
+            "$value": "{surface.color.primary}",
+            "$modes": {
+              "default": "{surface.color.primary}",
+              "cds-g0": {
```
dchyun: [Nit] For the carbon modes if they only contain a `$value` they can reference the value directly.

```
"cds-g0": "{carbon.themes.white.backgroundHover}",
```

packages/components/src/components/hds/table/tr.gts
```diff
@@ -71,19 +74,45 @@ export default class HdsTableTr extends Component<HdsTableTrSignature> {
     return undefined;
   }
 
+  get isSelected(): boolean {
+    return this.isSelectedOverride ?? this.args.isSelected ?? false;
```
dchyun, line 79: [Issue] The selected state styles are still getting applied when the selected state of a row is changed based on the table header select/unselect all. If a row is selected and then the row header checkbox is used to unselect all of the rows the row still has the selected styling.

Image

packages/components/src/styles/components/table.scss
```diff
@@ -38,21 +35,68 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
+  }
+}
+
+// row height
+
+.hds-table__th,
+.hds-table__td {
+  .hds-table--density-short & {
+    @include hds-apply-only-if-carbon() {
+      // CSS doesn't allow minimum height set on table cells, need to set height instead
```
dchyun: [Question] Can this selector block be removed since it's empty?

packages/components/src/styles/components/table.scss
```diff
@@ -38,21 +35,68 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
 
     .hds-table__th {
-      padding: $hds-table-cell-padding-medium;
+      // we need to account for the `1px` border when calculating the actual padding
+      padding: var(--hds-table-cell-padding-vertical-medium) var(--hds-table-cell-padding-horizontal-medium)
+        calc(var(--hds-table-cell-padding-vertical-medium) - 1px);
       text-align: left;
+
+      .hds-table--density-short & {
+        @include hds-apply-only-if-carbon() {
```
dchyun: [Suggestion] Maybe instead of using the carbon only mixins to account for the nuance that the HDS theme needs to maintain the medium size padding for all densities we could create table header padding tokens and use aliases to the other row padding tokens

```
"header": {
  "padding": {
    "vertical": {
      "short": {
        "$type": "dimension",
        "$value": "{table.cell.padding.vertical.medium}",
        "$modes": {
          "default": "{table.cell.padding.vertical.medium}",
          "cds-g0": "{table.cell.padding.vertical.short}",
          "cds-g10": "{table.cell.padding.vertical.short}",
          "cds-g90": "{table.cell.padding.vertical.short}",
          "cds-g100": "{table.cell.padding.vertical.short}"
        }
      }
    }
  }
}
```
```
.hds-table--density-short & {
  padding: var(--hds-table-header-padding-vertical-short) var(--hds-table-header-padding-horizontal-short)
    calc(var(--hds-table-header-padding-vertical-short) - 1px);
}
```

showcase/app/components/page-carbonization/components/table/index.gts
```diff
@@ -0,0 +1,537 @@
...
+        <HdsTableThButtonTooltip
+          @tooltip="Here is more information"
+          mock-state-value={{state}}
+        />
+        <br />
+      {{/each}}
+    </:theming>
+  </ShwCarbonizationComparisonGrid>
+
+  <ShwTextH3>ThButtonSort</ShwTextH3>
```
dchyun, line 513: [Issue?] Looking at the Carbon table the sort doesn't have a background color or border on hover / active. Also the icon stays black/white while active instead of the action color. We could align to this.

Image

Active state for carbon
Image

showcase/app/index.html
dchyun, line 111: [todo] You can add all of the `cds-table-` element tags to the below list to hide them until their styles are loaded.

packages/tokens/src/products/shared/table.json
dchyun, line 1: All the token names look good to me!

dchyun (commented):
Looking good on a first pass!

packages/components/src/styles/components/table.scss
```diff
@@ -38,21 +35,69 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
+          // we need to account for the `1px` border when calculating the actual padding
+          padding: var(--hds-table-cell-padding-vertical-short) var(--hds-table-cell-padding-horizontal-short)
+            calc(var(--hds-table-cell-padding-vertical-short) - 1px);
+        }
+      }
```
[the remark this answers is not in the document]
shleewhite: This is intentional, in helios the table header cells always have medium padding.

packages/components/src/styles/components/table.scss
```diff
@@ -38,21 +35,69 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
+
+      .hds-table--density-tall & {
+        @include hds-apply-only-if-carbon() {
+          // we need to account for the `1px` border when calculating the actual padding
+          padding: var(--hds-table-cell-padding-vertical-tall) var(--hds-table-cell-padding-horizontal-tall)
+            calc(var(--hds-table-cell-padding-vertical-tall) - 1px);
+        }
+      }
```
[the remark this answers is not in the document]
shleewhite: This is intentional, in helios the table header cells always have medium padding.

packages/components/src/styles/components/table.scss
shleewhite: oh im not sure why its empty, i must have removed the height on accident. adding it back.

didoo: It was removed by Copilot (see commits)

didoo:
@shleewhite I had a quick overview of the PR, and I think is better if you do also the AdvancedTable in the same PR (similar to what Kristin is doing with the form controls), not only to avoid back and forth between decisions but more importantly to have a more holistic view of the carbonization of these two very similar components. What do you think?
I know it's a big ask, given this is your first carbonization, but I think it's still something worth to do it in this way, even if it takes longer.

/cc @alex-ju let me/us know if you disagree with my assessment above

alex-ju:
> /cc @alex-ju let me/us know if you disagree with my assessment above

I don’t have strong opinions – if it’s easier for reviewers to compare and contrast both options simultaneously then that would be preferable indeed.

That being said, I see value in seeking early feedback on the Table component first, before replicating some of this work across the AdvancedTable.

packages/tokens/src/products/shared/table.json
```diff
@@ -0,0 +1,447 @@
+{
+  "table": {
+    "cell": {
+      "padding": {
+        "horizontal": {
+          "short": {
+            "$type": "dimension",
+            "$value": "16",
+            "unit": "px",
+            "$modes": {
```
dchyun:
```suggestion
```
[todo] If the carbon values and hds are the same we can remove the `$modes`. This could apply to a few tokens in here.

packages/components/src/styles/components/table.scss
```diff
@@ -118,8 +157,8 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
 
   &:hover,
   &.mock-hover {
-    color: var(--hds-foreground-color-primary);
-    background-color: var(--hds-surface-color-interactive);
+    color: var(--hds-table-header-button-foreground-color-hover);
+    background-color: var(--hds-table-header-button-surface-color-hover);
     border-color: var(--hds-border-color-strong);
```
dchyun: [Question] There's a border for the carbon theme header buttons, but the carbon table doesn't have the border on hover. Was this difference intentional? 

Image

packages/components/src/styles/components/table.scss
```diff
@@ -162,14 +201,78 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
+
   .hds-table__tr {
-    color: var(--hds-foreground-color-primary);
+    color: var(--hds-table-cell-foreground-color-default);
     background-color: var(--hds-surface-color-primary);
 
+    @include hds-apply-only-if-carbon() {
+      .hds-table__th,
+      .hds-table__td {
+        border-bottom: var(--hds-table-row-border-width) solid var(--hds-table-row-border-color-default);
```
dchyun: [Issue?] I don't think the striped carbon table has this same bottom border color. It's not visible right now because it blends into the striped colors, but you can see it on the last row in the table. 

Image

packages/tokens/src/products/shared/table.json
```diff
@@ -0,0 +1,447 @@
...
+            "cds-g10": "{carbon.themes.g10.layerAccent01}",
+            "cds-g90": "{carbon.themes.g90.layerAccent01}",
+            "cds-g100": "{carbon.themes.g100.layerAccent01}"
+          }
+        }
+      },
+      "cell": {
+        "padding": {
+          "horizontal": {
+            "short": {
```
dchyun: [Suggestion] Maybe we don't need to have short/medium/tall tokens since they are all 16px. Could we just have one `cell-padding-horizontal`?

packages/tokens/src/products/shared/advanced-table.json
dchyun, line 1: [Suggestion] For the advanced table tokens that are identical to the table tokens for the same thing could we use the table tokens directly in the sass instead of making new tokens? Kind of matches to something similar we did using code block tokens in the code editor.

showcase/app/components/page-carbonization/components/advanced-table/index.gts
dchyun, line 1: [Suggestion] Could you add an example for the adv table + filter bar?

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -282,13 +324,22 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
...
       }
     }
 
     // Only apply the thick border to the last sticky column
     .hds-advanced-table__th--is-sticky-column:not(:has(+ .hds-advanced-table__th--is-sticky-column)) {
-      border-right: 3px solid $hds-advanced-table-border-color;
+      border-right: 3px solid var(--hds-advanced-table-row-border-color-default);
+
+      @include hds-apply-only-if-carbon() {
+        border-right: none;
```
dchyun: [Question][Design] In the HDS theme the pinned column is indicated by the wider border width, but with the borders removed in the carbon theme it's not indicated to the user as significantly what the pinned column is. Are we okay with this difference?

Image

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -334,21 +385,37 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
...
+      border-bottom: none;
+    }
 
     .hds-advanced-table__scroll-indicator-top {
       top: unset;
       bottom: -11px;
+
+      @include hds-apply-only-if-carbon() {
+        // less offset because the carbonized table doesn't have borders
+        bottom: -8px;
```
dchyun: [Suggestion] Could we make a token for this instead of using the mixin?

packages/tokens/src/products/shared/advanced-table.json
```diff
@@ -0,0 +1,200 @@
...
+    "row": {
+      "border": {
+        "width": {
+          "$type": "dimension",
+          "$value": "{table.row.border.width}"
+        },
+        "color": {
+          "default": {
+            "$type": "color",
+            "$value": "{table.row.border.color.default}"
```
dchyun, line 132: [Suggestion] We could add `$modes` for the header and cell border colors which are `transparent` instead of using the mixin to set `border: none`

showcase/app/components/page-carbonization/components/advanced-table/index.gts
dchyun, line 1: [Question][Design] With the borders removed is there a risk when the columns are really wide like this a user could think the sort icon is associated with the column after it not the column its in? The carbon table doesn't run into this issue because the entire header cell is a button and has a hover effect.

Image

packages/tokens/src/products/shared/advanced-table.json
shleewhite: oh I was following how i think the form elements are doing it... I think we talked about it at one of the syncs. @didoo which would be your preference? should I use the table tokens in the advanced table sass or make new tokens that reference the table values?

showcase/app/components/page-carbonization/components/advanced-table/index.gts
shleewhite: @jorytindall @heatherlarsen do you guys have thoughts? it is def awkward that the sort button is so far away.

packages/components/src/styles/components/advanced-table.scss
shleewhite: I think we are? I showed this briefly at the sync on monday, but @heatherlarsen and @jorytindall can you confirm?

packages/tokens/src/products/shared/advanced-table.json
didoo: Yes, I suggested Lee to create tokens for the `AdvancedTable` and alias them to the `Table` tokens when meaningful. For the `CodeEditor` we don't have at all tokens, so in the `CodeEditor` Sass code all the rules refer to the `CodeBlock` tokens. For the `AdvancedTable` instead in the Sass code we would have a mix of component tokens for `AdvancedTable` (generated) and tokens from `Table`. I think it would be a bit messy/confusing. Makes sense?

packages/tokens/src/products/shared/advanced-table.json
shleewhite: ok, I have separated them into 2 tokens, a cell border (which is the main one the helios theme uses) and a row border (the carbon tokens)


that way I can avoid having the mixin as much.

packages/components/src/styles/components/table.scss
shleewhite: should match carbon now, although I feel its odd that theres no visual indicator for mouse users. we can talk with design tho about the experience.

packages/components/src/styles/components/table.scss
shleewhite: should be fixed now!

showcase/app/components/page-carbonization/components/advanced-table/index.gts
shleewhite: Added. I ended up removing the border for the filter bar to match the table styles.

packages/tokens/src/products/shared/advanced-table.json
```diff
@@ -0,0 +1,269 @@
...
+            "$type": "color",
+            "$value": "{surface.color.primary}",
+            "$modes": {
+              "default": "{surface.color.primary}",
+              "cds-g0": "#E0E0E0",
+              "cds-g10": "#F1F1F1",
+              "cds-g90": "#474747",
+              "cds-g100": "#373737"
+            },
+            "comment": "the carbon hover colors have alpha values so when a column is pinned + hovered, you can see the content behind the pinned column. These are the closest match to the carbon hover colors without alpha values."
```
dchyun:
```suggestion
            "comments": {
              "cds": "the carbon hover colors have alpha values so when a column is pinned + hovered, you can see the content behind the pinned column. These are the closest match to the carbon hover colors without alpha values."
            }
```
[Nit] Since these comments are only related to the carbon theme we can use this format to only show them there. This applies to the other comments here as well.

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -768,6 +944,10 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
 .hds-advanced-table__actions .hds-filter-bar {
   border-bottom: none;
   border-radius: $hds-advanced-table-border-radius $hds-advanced-table-border-radius 0 0;
+
+  @include hds-apply-only-if-carbon() {
+    border: none;
```
dchyun: Looks like the carbonizer filter bar won't have the border so this can be removed and it will be handled within the filter bar carbonization

dchyun (commented):
Thanks for addressing my earlier comments, all of those changes are looking good to me.

packages/components/src/styles/components/table.scss
heatherlarsen: This is because the header cell is actionable and receives the hover treatment in Carbon, whereas in HDS, only the individual buttons are actionable. I agree that it's odd to have no visual indicator (other than the cursor change), so I think it would be good to mimic the visual hover Carbon uses for the header cell on these buttons. Is that possible?

shleewhite: I mean, we could - but in carbon the only actionable thing in the header is the sort button. but we can have expand button, tooltip, sort, drag handle, and context menu. would all of those make the whole cell background color change?

heatherlarsen: Sorry for the confusion; I mean adding the visual changes to the buttons, not the whole header. These buttons should mimic the `secondary muted` buttons. 

For example: 
secondary muted button in table header

Here's a link to this in Figma: https://www.figma.com/design/GWMDKHuGr98DFxSptH6vgj/-LJ--Token-Migration?node-id=107798-27050&t=AIiQ5W5bZDpTT13H-4

packages/components/src/styles/components/advanced-table.scss
heatherlarsen: We're losing a bit of affordance by not having the border present. I didn't see this feature available in Carbon's Data Table, so we can make a call on this. I recommend displaying a border; let me play around with this in Figma, and I'll tag you for review.

showcase/app/components/page-carbonization/components/advanced-table/index.gts
heatherlarsen: @jorytindall Do you recall why we moved the sort to the right of the header cells? I remember it being an intentional decision, but I don't remember what the reason was.

packages/components/src/styles/components/advanced-table.scss
heatherlarsen: Tagged you both, but here's a link if you didn't get a notification: https://www.figma.com/design/GWMDKHuGr98DFxSptH6vgj/-LJ--Token-Migration?node-id=107798-61804&t=AIiQ5W5bZDpTT13H-4

showcase/app/components/page-carbonization/components/advanced-table/index.gts
dchyun: This was discussed with the team and we will leave it as is for now.

packages/tokens/src/products/shared/table.json
```diff
@@ -0,0 +1,386 @@
...
+        },
+        "surface": {
+          "color": {
+            "active": {
+              "$type": "color",
+              "$value": "{surface.color.interactive-active}"
+            },
+            "hover": {
+              "$type": "color",
+              "$value": "{surface.color.interactive}"
```
dchyun: [Issue?] I think this token will need `$modes`. The hover color in figma looks darker than what's here right now.

**Code**
Image

**Figma**

Image

Image

packages/tokens/src/products/shared/advanced-table.json
```diff
@@ -0,0 +1,298 @@
...
+      "border": {
+        "color": {
+          "$type": "color",
+          "$value": "{table.cell.border.color}"
+        },
+        "width": {
+          "$type": "dimension",
+          "$value": "{table.cell.border.width}"
+        },
+        "pinned": {
```
dchyun: [Suggestion] To align with the token format the modifier `pinned` can be added to the end. `border-color-pinned` and `border-width-pinned`. Could update the other tokens to `border-color-default` and `border-width-default`.

packages/components/src/components/hds/advanced-table/index.gts
didoo, line 1: [question] could these changes be cherry-picked in a separate PR targeting `main`? in this way we reduce the footprint of this PR, and we have proper changeset/history for them (and we release them separately). Then we you rebase your branch on `main` (once approved/merged) you get them directly as already done.

packages/components/src/components/hds/advanced-table/tr.gts
didoo, line 1: [question] could these changes be cherry-picked in a separate PR targeting `main`?

packages/components/src/components/hds/table/index.gts
didoo, line 1: [question] could these changes be cherry-picked in a separate PR targeting `main`?

packages/components/src/components/hds/table/tr.gts
didoo, line 1: [question] could these changes be cherry-picked in a separate PR targeting `main`?

showcase/tests/integration/components/hds/table/tr-test.gts
didoo, line 1: [question] could these changes be cherry-picked in a separate PR targeting main?

showcase/tests/integration/components/hds/table/index-test.gts
didoo, line 1: [question] could these changes be cherry-picked in a separate PR targeting main?

showcase/tests/integration/components/hds/advanced-table/tr-test.gts
didoo, line 1: [question] could these changes be cherry-picked in a separate PR targeting main?

showcase/tests/integration/components/hds/advanced-table/features/multi-selection-test.gts
didoo, line 1: [question] could these changes be cherry-picked in a separate PR targeting main?

showcase/app/index.html
```diff
@@ -142,6 +144,13 @@
       cds-pagination:not(:defined),
       cds-pagination-nav:not(:defined),
       cds-selectable-tag:not(:defined),
+      cds-table:not(:defined),
```
didoo, line 147: I did a quick check, and to me the list of `<cds-table[-***]` tags seems to be:
```
cds-table
cds-table-head
cds-table-header-row
cds-table-header-cell
cds-table-body
cds-table-cell
cds-table-row
cds-table-expanded-row
cds-table-expand-header-cell
```

can you double check this list with `:not(:defined)` is correct?

showcase/app/styles/showcase-pages/carbonization/advanced-table.scss
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright IBM Corp. 2021, 2025
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// CARBONIZATION > ACCORDION
```
didoo:
```suggestion
// CARBONIZATION > ADVANCED TABLE
```

showcase/app/styles/showcase-pages/carbonization/advanced-table.scss
```diff
@@ -0,0 +1,14 @@
...
+
+body.page-carbonization-components-advanced-table {
+  .shw-component-advanced-table-reorder-handle-container
+    .hds-advanced-table__th-reorder-handle {
```
didoo:
```suggestion
  .shw-component-advanced-table-reorder-handle-container .hds-advanced-table__th-reorder-handle {
```

packages/components/src/styles/components/table.scss
```diff
@@ -38,20 +33,48 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
+      height: 2.5rem;
+    }
+  }
+
+  .hds-table--density-tall & {
+    @include hds-apply-only-if-carbon() {
+      height: 3rem;
+    }
+  }
 }
```
didoo: [suggesiton] Would something like this work? because in this case it's easier to read, and reduces significantly the complexity of the CSS selector generated 
```
.hds-table__th,
.hds-table__td {
    // CSS doesn't allow minimum height set on table cells, need to set height instead
    height: var(--hds-var-table-cell-height);
  }
}

.hds-table--density-short {
  --hds-var-table-cell-height: 2rem;
}

.hds-table--density-medium {
  --hds-var-table-cell-height: 2.5rem;
}

.hds-table--density-tall {
  --hds-var-table-cell-height: 3rem;
}
```

(in your case the generated selector is: `.hds-theme-light .hds-table--density-short .hds-table__th, .hds-theme-light .hds-table--density-short .hds-table__td, .hds-theme-dark .hds-table--density-short .hds-table__th, .hds-theme-dark .hds-table--density-short .hds-table__td, .hds-theme-system .hds-table--density-short .hds-table__th, .hds-theme-system .hds-table--density-short .hds-table__td` and this is multiplied by 3 sizes)

packages/components/src/styles/components/table.scss
```diff
@@ -38,20 +33,48 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
+
   .hds-table__tr {
     color: var(--hds-foreground-color-strong);
     background-color: var(--hds-surface-color-strong);
 
     .hds-table__th {
-      padding: $hds-table-cell-padding-medium;
+      // we need to account for the `1px` border when calculating the actual padding
+      padding: var(--hds-table-header-cell-padding-vertical) var(--hds-table-cell-padding-horizontal)
+        calc(var(--hds-table-header-cell-padding-vertical) - 1px);
```
didoo: [issue?] when it's `cds` there's no border bottom, so the padding bottom is incorrect in that mode?

packages/components/src/styles/components/table.scss
```diff
@@ -113,14 +136,14 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
   padding: 0;
   color: var(--hds-foreground-color-faint);
   background-color: transparent;
```
didoo: [suggestion] for completeness, maybe it's worth to add also these two to the component-level tokens (even if they have a common value between `hds` and `cds`)
- `--hds-table-header-button-foreground-color-default`
- `--hds-table-header-button-surface-color-default`

packages/components/src/styles/components/table.scss
```diff
@@ -162,14 +185,79 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
+
+      .hds-table__th,
+      .hds-table__td {
+        border-bottom: var(--hds-table-row-border-width) solid var(--hds-table-row-border-color-hover);
+      }
+    }
+
     // striped rows
 
+    .hds-table--striped &.hds-table__tr {
```
didoo: [issue] this generates this selector: `.hds-table--striped .hds-table__tbody.hds-table__tr .hds-table__th`; probably there should be a space after the `&`?

packages/components/src/styles/components/table.scss
```diff
@@ -162,14 +185,79 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
+        background-color: var(--hds-table-cell-surface-color-striped-hover);
+      }
+    }
+
+    // selected rows
+
+    &.hds-table__tr--is-selected {
+      color: var(--hds-table-cell-foreground-color-selected);
+
+      // setting it only with carbon to avoid the striped styles being affected by the selected styles in the helios theme
```
didoo, line 232: [suggestion] I think something like this would work as well, can you check?

```
      // setting it only with carbon to avoid the striped styles being affected by the selected styles in the helios theme
        background-color: var(--hds-table-cell-surface-color-selected-default);

        .hds-table__th,
        .hds-table__td {
          border-bottom: var(--hds-table-row-border-width) solid var(--hds-table-row-border-color-selected-default);
        }

        &:hover,
        &.mock-hover {
          background-color: var(--hds-table-cell-surface-color-selected-hover);

          .hds-table__th,
          .hds-table__td {
            border-bottom: var(--hds-table-row-border-width) solid var(--hds-table-row-border-color-selected-hover);
          }
        }
      }
```

packages/components/src/styles/components/table.scss
didoo: plus, you could define also a local variable for the padding and simplify these lines (below in the code):

```
  .hds-table__th,
  .hds-table__td {
    text-align: left;

    // density

    .hds-table--density-short & {
      padding: var(--hds-table-cell-padding-vertical-short) var(--hds-table-cell-padding-horizontal)
        calc(var(--hds-table-cell-padding-vertical-short) - 1px);
    }

    .hds-table--density-medium & {
      padding: var(--hds-table-cell-padding-vertical-medium) var(--hds-table-cell-padding-horizontal)
        calc(var(--hds-table-cell-padding-vertical-medium) - 1px);
    }

    .hds-table--density-tall & {
      padding: var(--hds-table-cell-padding-vertical-tall) var(--hds-table-cell-padding-horizontal)
        calc(var(--hds-table-cell-padding-vertical-tall) - 1px);
    }
  }

```

packages/tokens/src/products/shared/table.json
```diff
@@ -0,0 +1,386 @@
...
+          "color": {
+            "active": {
+              "$type": "color",
+              "$value": "{foreground.color.primary}"
+            },
+            "hover": {
+              "$type": "color",
+              "$value": "{foreground.color.primary}"
+            },
+            "sorted": {
```
didoo, line 72: [todo] let's try to keep the order of the tokens definitions consistent (in this case the children should be `default`/`hover`/`focus`/`active`); similar for the semantic tokens, where usually the order is `foreground`/`surface`/`border`)

check also other definitions, if possible

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -10,18 +10,13 @@
...
+@use "../mixins/carbonization" as *;
 
 $hds-advanced-table-border-radius: var(--hds-border-radius-medium);
-$hds-advanced-table-border-width: 1px;
 $hds-advanced-table-inner-border-radius: calc(
-  #{$hds-advanced-table-border-radius} - #{$hds-advanced-table-border-width}
+  #{$hds-advanced-table-border-radius} - var(--hds-advanced-table-row-border-width)
 );
-$hds-advanced-table-border-color: var(--hds-border-color-primary);
 $hds-advanced-table-border-resize-handle-hover-color: #b1b1b5;
```
didoo: [maybe?] should this be a token? In both cases, the name should be `hds-advanced-table-resize-handle-border-color-hover`

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -10,18 +10,13 @@
...
-$hds-advanced-table-header-height: 48px;
-$hds-advanced-table-cell-padding-medium: 14px 16px 13px 16px; // the 1px difference is to account for the bottom border
-$hds-advanced-table-cell-padding-short: 6px 16px 5px 16px; // the 1px difference is to account for the bottom border
-$hds-advanced-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to account for the bottom border
 $hds-advanced-table-button-size: 24px; // the size of the buttons and dropdown triggers in the header cell
```
didoo: [maybe?] should this be a token too? In both cases, for consistency with the `Table` the name should be `hds-advanced-table-header-button-size`

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -10,18 +10,13 @@
...
 $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
```
didoo: [maybe?] should this be a token too? in both cases, the name should be `hds-advanced-table-drag-preview-surface-color`

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -77,6 +72,29 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
   border-bottom-left-radius: inherit;
 }
 
+// row height
```
didoo: [suggestion] see comment for similar code in `Table`

packages/tokens/src/products/shared/advanced-table.json
```diff
@@ -0,0 +1,298 @@
...
+      },
+      "surface": {
+        "color": {
+          "default": {
+            "$type": "color",
+            "$value": "{surface.color.primary}"
+          },
+          "hover": {
+            "$type": "color",
+            "$value": "{surface.color.primary}",
```
didoo: should this (and the `default`) point to `{table.cell.surface.color.hover}`?

packages/tokens/src/products/shared/advanced-table.json
```diff
@@ -0,0 +1,298 @@
...
+            }
+          },
+          "striped": {
+            "default": {
+              "$type": "color",
+              "$value": "{table.cell.surface.color.striped.default}"
+            },
+            "hover": {
+              "$type": "color",
+              "$value": "{surface.color.faint}",
```
didoo: should this (and the `default`) point to `{table.cell.surface.color.striped.hover}`?

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -527,17 +588,88 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
...
+    }
+
+    &:hover,
+    &.mock-hover {
+      color: var(--hds-advanced-table-cell-foreground-color-hover);
+
+      .hds-advanced-table__th,
+      .hds-advanced-table__td {
+        background-color: var(--hds-advanced-table-cell-surface-color-hover);
+        border-color: var(--hds-advanced-table-row-border-color-hover);
```
didoo:
```suggestion
        border-bottom-color: var(--hds-advanced-table-row-border-color-hover);
```

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -527,17 +588,88 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
...
+      }
+
+      .hds-advanced-table__th,
+      .hds-advanced-table__td {
+        border-bottom: var(--hds-advanced-table-row-border-width) solid var(--hds-advanced-table-row-border-color-hover);
```
didoo:
```suggestion
        border-bottom-color: var(--hds-advanced-table-row-border-color-hover);
```

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -527,17 +588,88 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
...
+      }
+    }
 
     // striped rows
 
+    .hds-advanced-table--striped &.hds-advanced-table__tr {
```
didoo: see comment before
```suggestion
    .hds-advanced-table--striped & .hds-advanced-table__tr {
```

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -527,17 +588,88 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
...
+    }
+
+    // selected rows
+
+    &.hds-advanced-table__tr--is-selected {
+      color: var(--hds-advanced-table-cell-foreground-color-selected);
+
+      .hds-advanced-table__th,
+      .hds-advanced-table__td {
+        // setting it only with carbon to avoid the striped styles being affected by the selected styles in the helios theme
```
didoo: see comment in the `table.css` file about a possible simplification

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -592,15 +724,18 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
     // density
 
     .hds-advanced-table--density-short & {
-      padding: $hds-advanced-table-cell-padding-short;
+      padding: var(--hds-advanced-table-cell-padding-vertical-short) var(--hds-advanced-table-cell-padding-horizontal)
```
didoo: see comment in `table.css` about possible simplification/abstraction using a dedicated local variable

didoo (commented):
Left a few comments and suggestions. I focused only on the code, didn't review the actual visual implementation (I'll do a second pass once these comments/suggestions have been addressed or closed)

packages/components/src/components/hds/advanced-table/index.gts
shleewhite: https://github.com/hashicorp/design-system/pull/4021

packages/components/src/styles/components/advanced-table.scss
shleewhite: Yes- it should be a token, however when its used its actually the background color. I made it `hds-advanced-table-resize-handle-surface-color-hover`

packages/components/src/styles/components/table.scss
shleewhite: makes sense! I refactored for both table & advanced table

packages/components/src/styles/components/table.scss
shleewhite: ah yes, I made a new token `--hds-table-header-cell-border-offset` to account for this (also made the change for advanced table)

packages/components/src/styles/components/table.scss
shleewhite: Updated for both table + advancedtable

packages/components/src/styles/components/table.scss
shleewhite: yeah I believe this works! will update.

showcase/app/index.html
shleewhite: ah yeah, I missed the expand ones.

packages/tokens/src/products/shared/table.json
shleewhite: for sure, I think I have fixed all of the defs in table + advanced table

packages/tokens/src/products/shared/table.json
shleewhite: added the modes!

packages/tokens/src/products/shared/advanced-table.json
```diff
@@ -0,0 +1,339 @@
...
+            "default": {
+              "$type": "color",
+              "$value": "{table.cell.surface.color.striped.default}"
+            },
+            "hover": {
+              "$type": "color",
+              "$value": "{table.cell.surface.color.striped.hover}",
+              "$modes": {
+                "default": "{table.cell.surface.color.striped.hover}",
+                "cds-g0": "#E8E8E8",
```
dchyun, line 242: [Issue?] Could we tweak this value a bit to more closely align to the carbon color? Right now it has almost the exact same color as the header for the g0 theme. The Carbon table color is lighter. 

Image

showcase/app/components/page-carbonization/components/advanced-table/index.gts
```diff
@@ -0,0 +1,921 @@
...
+              {{/each}}
+            </cds-table-body>
+          </cds-table>
+        </:reference>
+      </ShwCarbonizationComparisonGrid>
+    </section>
+
+    <ShwDivider />
+
+    <ShwTextH2>With sticky columns + header</ShwTextH2>
```
dchyun, line 390: [Question][Non-blocker] Not table related, but should we removed this extra padding that is above and below the dropdown list items? I don't see this in the Carbon dropdown.

Image

showcase/app/components/page-carbonization/components/advanced-table/index.gts
```diff
@@ -0,0 +1,921 @@
...
+                  mock-state-value={{state}}
+                />
+              </div>
+            </SF.Item>
+          </ShwFlex>
+        {{/each}}
+      </:theming>
+    </ShwCarbonizationComparisonGrid>
+
+    <ShwTextH3>ThResizeHandle</ShwTextH3>
```
dchyun, line 827: [Question] I saw a comment in Figma that said this would follow the `secondary-muted` button styles. This seems to be slightly different as it's using `surface-interactive` and the colors are slighly different from `secondary-muted`. Can't see any Figma designs to confirm but are the currently colors what's expected?

Image

Image

packages/tokens/src/products/shared/table.json
```diff
@@ -0,0 +1,420 @@
...
+              "cds-g0": "{carbon.themes.white.textSecondary}",
+              "cds-g10": "{carbon.themes.g10.textSecondary}",
+              "cds-g90": "{carbon.themes.g90.textSecondary}",
+              "cds-g100": "{carbon.themes.g100.textSecondary}"
+            }
+          },
+          "hover": {
+            "$type": "color",
+            "$value": "{foreground.color.primary}",
+            "$modes": {
```
dchyun: [Nit] Since `foreground-color-primary` already maps to `textPrimary` for the carbon themes you could remove this token and reference foreground primary directly. Or just remove the modes if you want to keep the token.

packages/tokens/src/products/shared/table.json
```diff
@@ -0,0 +1,420 @@
...
+              "cds-g0": "{carbon.themes.white.textPrimary}",
+              "cds-g10": "{carbon.themes.g10.textPrimary}",
+              "cds-g90": "{carbon.themes.g90.textPrimary}",
+              "cds-g100": "{carbon.themes.g100.textPrimary}"
+            }
+          },
+          "selected": {
+            "$type": "color",
+            "$value": "{foreground.color.primary}",
+            "$modes": {
```
dchyun: [Nit] Since `foreground-color-primary` already maps to `textPrimary` for the carbon themes you could remove this token and reference foreground primary directly. Or just remove the modes if you want to keep the token.

packages/tokens/src/products/shared/table.json
```diff
@@ -0,0 +1,420 @@
...
+                "cds-g90": "{carbon.themes.g90.layerSelectedHover01}",
+                "cds-g100": "{carbon.themes.g100.layerSelectedHover01}"
+              }
+            }
+          }
+        }
+      },
+      "foreground": {
+        "color": {
+          "default": {
```
dchyun: [Nit] If you do decide to remove the `-hover` and `-selected` tokens (see previous comments) this could be simplified to just `foreground-color`.

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -77,6 +69,39 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
...
+// density
+
+.hds-advanced-table__th,
+.hds-advanced-table__td {
+  @include hds-apply-only-if-carbon() {
+    min-height: var(--hds-var-advanced-table-cell-height);
+  }
+}
+
+.hds-advanced-table--density-short {
```
dchyun: [Suggestion] These styles could be handled in a loop instead.

```
$hds-advanced-table-densities: ("short", "medium", "tall");

@each $density in $hds-advanced-table-densities {
  .hds-advanced-table--density-#{$density} {
    ...
  }
}
```

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -444,40 +485,49 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
...
   &.mock-hover {
     box-shadow: var(--hds-elevation-low-box-shadow);
   }
+
+  &:focus-visible,
+  &.mock-focus {
+    &::before {
+      // these are needed to make the dropdown icon focus state match the size of the other header buttons focus state
+      inset: 0 !important;
+      border-width: 0 !important;
```
dchyun: [Issue] This style is making the HDS theme dropdown lose its inner border color.

Image

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -808,4 +936,8 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
   border: 1px solid var(--hds-border-color-primary);
   border-bottom-right-radius: $hds-advanced-table-border-radius;
   border-bottom-left-radius: $hds-advanced-table-border-radius;
+
+  @include hds-apply-only-if-carbon() {
+    border: none;
```
dchyun: [Suggestion] An `empty-state-border-width` token could be created and set it to `0px` instead of using the mixin.

packages/components/src/styles/components/table.scss
```diff
@@ -38,20 +33,51 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
+
+.hds-table__th,
+.hds-table__td {
+  @include hds-apply-only-if-carbon() {
+    // CSS doesn't allow minimum height set on table cells, need to set height instead
+    height: var(--hds-var-table-cell-height);
+  }
+}
+
+.hds-table--density-short {
```
dchyun: [Suggestion] These styles could be handled in a loop instead.

```
$hds-table-densities: ("short", "medium", "tall");

@each $density in $hds-table-densities {
  .hds-table--density-#{$density} {
    ...
  }
}
```

showcase/app/components/page-carbonization/components/advanced-table/index.gts
shleewhite: ah yes, I missed that they are styled differently than the other header buttons

packages/tokens/src/products/shared/table.json
shleewhite: im leaving the tokens and just removing the modes

showcase/app/components/page-carbonization/components/advanced-table/index.gts
dchyun: nvm I mistaking thought the carbon dropdown doesn't do this, but it does.

packages/components/src/styles/components/advanced-table.scss
shleewhite: removed that style! should be better now.

showcase/app/index.html
```diff
@@ -142,6 +144,15 @@
...
+      cds-table-header-row:not(:defined),
+      cds-table-header-cell:not(:defined),
+      cds-table-head:not(:defined),
+      cds-table-expanded-row:not(:defined),
+      cds-table-expand-header-cell:not(:defined),
+      cds-table-cell:not(:defined),
+      cds-table-body:not(:defined),
+      cds-table-body-row:not(:defined),
```
didoo:
```suggestion
      cds-table-row:not(:defined),
```

packages/tokens/src/products/shared/table.json
```diff
@@ -0,0 +1,444 @@
...
+          "$value": "0",
+          "unit": "rem",
+          "$modes": {
+            "default": "0",
+            "cds-g0": "2",
+            "cds-g10": "2",
+            "cds-g90": "2",
+            "cds-g100": "2"
+          } 
+        },
```
didoo: [suggestion] what do you think of doing something like this and remove the `@include hds-apply-only-if-carbon()` condition in the SCSS file?
```suggestion
        "short": {
          "$value": "auto",
          "$modes": {
            "default": "auto",
            "cds-g0": { "$value": "2", "$type": "dimension", "unit": "rem" },
            "cds-g10": { "$value": "2", "$type": "dimension", "unit": "rem" },
            "cds-g90": { "$value": "2", "$type": "dimension", "unit": "rem" },
            "cds-g100": { "$value": "2", "$type": "dimension", "unit": "rem" }
          },
          "comment": "this token is applied as `min-height` to the cell"
          }
        },
```

packages/tokens/src/products/shared/table.json
didoo, line 1: [suggestion] consider sorting the entities following a more "semantic" order:
```
table
header
header-cell
header-button
body
row
cell
```

same for the color categories:
```
foreground
surface
border
```
(this would also be consistent with the order we use in most of the JSON files too)

packages/tokens/src/products/shared/advanced-table.json
didoo, line 1: [suggestion] see comment for the order of entities in the `table.json` file

packages/tokens/src/products/shared/advanced-table.json
```diff
@@ -0,0 +1,369 @@
...
+        "color": {
+          "default": {
+            "$type": "color",
+            "$value": "{table.row.border.color.default}"
+          },
+          "hover": {
+            "$type": "color",
+            "$value": "{table.row.border.color.hover}"
+          },
+          "striped": {
```
didoo: [thought] should this be `striped.default` for consistency? 🤔

packages/tokens/src/products/shared/table.json
```diff
@@ -0,0 +1,444 @@
...
+            "$value": "{border.color.primary}",
+            "$modes": {
+              "default": "{border.color.primary}",
+              "cds-g0": "{carbon.themes.white.layerHover01}",
+              "cds-g10": "{carbon.themes.g10.layerHover01}",
+              "cds-g90": "{carbon.themes.g90.layerHover01}",
+              "cds-g100": "{carbon.themes.g100.layerHover01}"
+            }
+          },
+          "striped": {
```
didoo: [thought] should this be `striped.default` for consistency? 🤔

packages/tokens/src/products/shared/advanced-table.json
```diff
@@ -0,0 +1,369 @@
...
+          }
+        }
+      }
+    },
+    "resize-handle": {
+      "surface": {
+        "color": {
+          "focus": {
+            "$type": "color",
+            "$value": "{core.color.blue-300}"
```
didoo: [todo] we should add a comment here explaining why we're using a `core` color (not theme-aware) and not a semantic color

packages/tokens/src/products/shared/advanced-table.json
```diff
@@ -0,0 +1,369 @@
...
+          },
+          "hover": {
+            "$type": "color",
+            "$value": "#b1b1b5"
```
didoo: [todo] we should add a comment here explaining why we're using an HEX color (not theme-aware) and not a semantic color

packages/tokens/src/products/shared/advanced-table.json
```diff
@@ -0,0 +1,369 @@
...
+          }
+        }
+      }
+    },
+    "drag-preview": {
+      "surface": {
+        "color": {
+          "$type": "color",
+          "$value": "rgba(204, 227, 254, 30%)"
```
didoo: [todo] we should add a comment here explaining why we're using this specific color (not theme-aware or mode-aware) and not a semantic color

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -77,6 +69,27 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
...
 
+// density
+
+$hds-advanced-table-densities: ("short", "medium", "tall");
+
+@each $density in $hds-advanced-table-densities {
+  .hds-advanced-table--density-#{$density} {
+    --hds-var-advanced-table-cell-height: var(--hds-advanced-table-cell-height-#{$density});
+    --hds-var-advanced-table-cell-padding-top: var(--hds-advanced-table-cell-padding-vertical-#{$density});
+    --hds-var-advanced-table-cell-padding-bottom: calc(
```
didoo, line 80: [praise] 👏👏👏

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -77,6 +69,27 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
...
+      var(--hds-advanced-table-cell-padding-vertical-#{$density}) - 1px
+    );
+  }
+}
+
+.hds-advanced-table__th,
+.hds-advanced-table__td {
+  @include hds-apply-only-if-carbon() {
```
didoo: [note] see suggestion in the JSON file about the possibility of removing this conditional inclusion

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -396,9 +425,10 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
     justify-content: center;
     width: 24px;
     height: 16px;
-    color: var(--hds-foreground-color-faint);
+    color: var(--hds-advanced-table-header-button-foreground-color-default);
+    // We don't use the existing `hds-advanced-table-header-button-foreground-color-default` token here because it is transparent. since the reorder handle sits halfway between the header and body, transparent background makes the button look split
```
didoo, line 436: [praise] 👏👏👏

packages/components/src/styles/components/table.scss
```diff
@@ -38,20 +33,43 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
+  .hds-table--density-#{$density} {
+    --hds-var-table-cell-height: var(--hds-table-cell-height-#{$density});
+    --hds-var-table-cell-padding-top: var(--hds-table-cell-padding-vertical-#{$density});
+    --hds-var-table-cell-padding-bottom: calc(var(--hds-table-cell-padding-vertical-#{$density}) - 1px);
+  }
+}
+
+.hds-table__th,
+.hds-table__td {
+  @include hds-apply-only-if-carbon() {
```
didoo: [note] see comment about potentially removing this conditional include

didoo (commented):
I have added a few comments, nothing major.

One thing I noticed while testing is that there is an issue with the borders in the `cds` modes with sticky column: there is some transparent border that allows the content that is being scrolled to appear underneath the sticky columns:

https://github.com/user-attachments/assets/17b33441-14ea-4b6d-8f3d-881e81db5ba7

packages/tokens/src/products/shared/table.json
shleewhite: oh nice! i didnt realize I could set the tokens like that. updated.

packages/tokens/src/products/shared/advanced-table.json
shleewhite: I dont think theres a why, I added comment saying that we should tokenize this (+ the other 2 non-semantic colors)

packages/components/src/styles/components/advanced-table.scss
dchyun, line 538:
```suggestion
    @include hds-focus-ring-advanced($pseudo-element-inset-hds: 0, $pseudo-element-inset-cds: 0);
    
    // Override focus styles from dropdown toggle icon
    color: var(--hds-foreground-color-faint);
    background-color: transparent;
```
[Issue] The dropdown toggle icon has a different background/foreground color from the other header cell buttons on focus. After carbonization there are now colors set on the dropdown toggle for focus that aren't getting overridden. 

Image

**Current HDS**

Image

**Carbonization**

Image

dchyun (commented):
Minus the couple open comments everything looks good to me! Just caught one small visual bug.

packages/components/src/styles/components/advanced-table.scss
shleewhite: fixed!

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -435,49 +465,59 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
...
 }
 
 .hds-advanced-table__th-context-menu .hds-dropdown-toggle-icon {
-  width: $hds-advanced-table-button-size;
-  height: $hds-advanced-table-button-size;
+  width: var(--hds-advanced-table-header-button-size);
+  height: var(--hds-advanced-table-header-button-size);
   margin: -2px 0;
   padding: 0;
   color: var(--hds-foreground-color-faint);
```
dchyun:
```suggestion
  color: var(--hds-advanced-table-header-button-foreground-color-default);
```
[Suggesiton] Since the `background-color` is now aligned to the token, can do the same for the foreground color.

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -435,49 +465,59 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
...
   &.mock-hover {
     box-shadow: var(--hds-elevation-low-box-shadow);
   }
+
+  &:focus-visible,
+  &.mock-focus {
+    // matching the same focus ring as th-button so the context menu + th buttons are the same size when focused. if use the default dropdown toggle icon focus state it ends up being a pixel too big.
+    @include hds-focus-ring-advanced($pseudo-element-inset-hds: 0, $pseudo-element-inset-cds: 0);
+    // setting it to the default color to match the th-button styles. when focused, th-buttons do not have a background color set so it defaults to the default background color. however, dropdown toggle icon does have a background color on focus, so using the default token here.
+    background-color: var(--hds-advanced-table-header-button-surface-color-default);
```
dchyun, line 509:
```suggestion
    color: var(--hds-advanced-table-header-button-foreground-color-default);
    background-color: var(--hds-advanced-table-header-button-surface-color-default);
```
[Issue] The foreground color also needs to match the th-button in addition to the background color. The above comment will need to be updated as well.

packages/components/src/styles/components/advanced-table.scss
dchyun: Thanks! The fix needs to be done for both color and background color. Right now it's only for background color.

packages/tokens/src/products/shared/advanced-table.json
```diff
@@ -0,0 +1,405 @@
...
+                "cds-g0": "{carbon.themes.white.backgroundHover}",
+                "cds-g10": "{carbon.themes.g10.backgroundHover}",
+                "cds-g90": "{carbon.themes.g90.backgroundHover}",
+                "cds-g100": "{carbon.themes.g100.backgroundHover}"
+              }
+            }
+          }
+        }
+      },
+      "foreground": {
```
didoo: [nit] move before `surface` sibling

packages/tokens/src/products/shared/advanced-table.json
didoo: @jorytindall @heatherlarsen do you remember why these colors were used?

packages/tokens/src/products/shared/advanced-table.json
didoo: @jorytindall @heatherlarsen do you remember why these colors were used?

packages/tokens/src/products/shared/advanced-table.json
didoo: @jorytindall @heatherlarsen do you remember why these colors were used?

packages/components/src/styles/components/advanced-table.scss
```diff
@@ -435,49 +466,60 @@ $hds-advanced-table-drag-preview-background-color: rgba(204, 227, 254, 30%);
...
 
   &:hover,
   &.mock-hover {
     box-shadow: var(--hds-elevation-low-box-shadow);
   }
+
+  &:focus-visible,
+  &.mock-focus {
+    // matching the same focus ring as th-button so the context menu + th buttons are the same size when focused. if use the default dropdown toggle icon focus state it ends up being a pixel too big.
+    @include hds-focus-ring-advanced($pseudo-element-inset-hds: 0, $pseudo-element-inset-cds: 0);
```
didoo, line 506: I think there is a specificity issue (another declaration is winning) in `cds` mode: 

Image

didoo (approved):
Pre-approving (only a minor thing to fix, and a nitpick)

Note: I have reviewed mainly the code, since @dchyun has already done a fine-combed visual review

packages/components/src/styles/components/advanced-table.scss
shleewhite: Ah, I fixed it.

didoo (approved):

dchyun (approved):
🎉🎉🎉

packages/tokens/src/products/shared/advanced-table.json
jorytindall: This is for the drag "box" surface color and is `blue-100` with opacity so that the table and drag target is visible underneath it. We don't have a token for it in Figma either (though it uses `table-column-reorder-surface-color` as a component token.

FWIW the value that is set in Figma is `rgba(204, 227, 254, 50%)`, rather than 30% opacity. I'm not sure if that's a previously existing inconsistency.

shleewhite: I think it is a previous inconsistency. I can update that

I added a comment to the token that we should theme it at some point in the future.

packages/components/src/styles/components/table.scss
```diff
@@ -38,20 +33,41 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
 .hds-table__thead {
+  background-color: var(--hds-table-header-surface-color);
+
   .hds-table__tr {
     color: var(--hds-foreground-color-strong);
     background-color: var(--hds-surface-color-strong);
 
     .hds-table__th {
-      padding: $hds-table-cell-padding-medium;
+      padding: var(--hds-table-header-cell-padding-vertical) var(--hds-table-cell-padding-horizontal)
```
jorytindall, line 69: [question] I'm not quite sure where to add this, but the rendered height short HDS table is a bit taller (~2px) than the Carbon table and I think it's because of a 2px difference in the line height of the text. Is this something worth indexing on, or would we consider this a bit nit-picky?

packages/components/src/styles/components/table.scss
```diff
@@ -162,14 +178,75 @@ $hds-table-cell-padding-tall: 22px 16px 21px 16px; // the 1px difference is to a
...
+
+      .hds-table__th,
+      .hds-table__td {
+        border-bottom: var(--hds-table-row-border-width) solid var(--hds-table-row-border-color-hover);
+      }
+    }
+
     // striped rows
 
+    .hds-table--striped & {
```
jorytindall, line 206: [issue] When I apply sorting to a table that is striped, the alternating striping seems to be ignored and it's possible to have two adjacent rows that are the same color.

Image

jorytindall: Disregard this, it's because the selected row has a different surface color that I didn't realize 🤦

packages/tokens/dist/products/css/themed-tokens/with-root-selector/cds-g0/themed-tokens.css
```diff
@@ -540,6 +551,36 @@
...
+  --hds-table-cell-padding-vertical-short: 7px;
+  --hds-table-cell-padding-vertical-tall: 7px;
+  --hds-table-cell-surface-color-hover: rgba(141, 141, 141, 0.12);
+  --hds-table-cell-surface-color-selected-default: #e0e0e0;
+  --hds-table-cell-surface-color-selected-hover: #d1d1d1;
+  --hds-table-cell-surface-color-striped-default: #e0e0e0;
+  --hds-table-cell-surface-color-striped-hover: rgba(141, 141, 141, 0.12);
+  --hds-table-header-button-border-color-active: rgba(0, 0, 0, 0);
+  --hds-table-header-button-border-color-hover: rgba(0, 0, 0, 0);
+  --hds-table-header-button-foreground-color-sorted-active: #161616;
```
jorytindall, line 580: [issue] I'm seeing a slight inconsistency in the tooltip button and the sort button foreground colors. In the g0 theme:
- The resting/default foreground color for the tooltip button and the indeterminate sort is #525252
- The resting/default foreground color for the descending/ascending sort button is #161616

I think all of these should have the same resting/default foreground color, and this is what is reflected in Figma. Even though these interactive elements are different from how Carbon implements these features.

jorytindall: Actually looks like this might be an issue with the `secondary-muted` variant (sorry if this comment is in a weird spot, kind of hard to know where it's best to comment).

jorytindall (commented):
Sorry for this barrage of comments that I wasn't really sure what to pin to. I'll try to organize them categorically.

## Scroll indicator

We're losing the scroll indicator gradient in the `AdvancedTable` on the dark Carbon themes. There isn't really a Figma token for this, so LMK how it's best to define that.

## Reordering/resizing

There may have been a few bugs introduced in the `AdvancedTable`:
- the reordering button displays on the hover action for the column resizing feature.
Image
- When I select "Move column" from the context menu, the drag handle displays, but is not focused (in the production branch it does do this)

## Empty state

There's an issue with the empty state (either in the showcase examples, or otherwise), where hover the tooltip there's a z-index issue causing the empty state to overlap the tooltip. Additionally, the header is scrollable. However, the scroll only appears after the tooltip has been interacted with. I wasn't able to reproduce it in the main branch, but there also aren't tooltips in any of the examples.
Image

dchyun:
@jorytindall  

It looks like the reordering / resizing and empty state bug you caught are both ones present on `main` today. We could address those in a separate bug fix PR and merge it into `main` directly. 

CC: @shleewhite

dchyun:
> @jorytindall
> 
> It looks like the reordering / resizing and empty state bug you caught are both ones present on `main` today. We could address those in a separate bug fix PR and merge it into `main` directly.
> 
> CC: @shleewhite

Created a JIRA ticket for this https://hashicorp.atlassian.net/browse/HDS-6573

packages/components/src/styles/components/table.scss
shleewhite: yuppppp

packages/components/src/styles/components/table.scss
shleewhite: Good q, I think its up to you... carbon line height is 1.28572 where ours is 1.4286. I can always make a token for it

packages/tokens/dist/products/css/themed-tokens/with-root-selector/cds-g0/themed-tokens.css
shleewhite: ahh I see the issue. I can update

packages/components/src/styles/components/table.scss
shleewhite: Update: I made a cell line height token bc carbon is partially using the line height to determine the height of the rows

shleewhite:
Note: I removed the border bottom fix from this PR bc it is addressed in its own PR against main: https://github.com/hashicorp/design-system/pull/4042

packages/components/src/styles/components/table.scss
dchyun: I see the `table-cell-line-height` token now exists but I don't see it being used in the styles anywhere. Is it still needed? Does it need to be used in the styles still?

packages/tokens/src/products/shared/advanced-table.json
```diff
@@ -0,0 +1,462 @@
...
+              "cds-g0": "{focus.color.action.external}",
+              "cds-g10": "{focus.color.action.external}",
+              "cds-g90": "{focus.color.action.external}",
+              "cds-g100": "{focus.color.action.external}"
+            }
+          }
+        }
+      }
+    },
+    "drag-preview": {
```
dchyun, line 414: [Question][Design] @jorytindall For the dark themes do we want to change the color of the blue drop place border in between the columns to also be white to align to the drag preview? 

Image

didoo (approved):
For me, good to go! 🚀 🙌

dchyun (approved):

packages/components/src/styles/components/table.scss
jorytindall: As discussed, probably not worth it here to bother.

packages/tokens/dist/products/css/themed-tokens/with-root-selector/cds-g0/themed-tokens.css
jorytindall: Looks good now!

packages/components/src/styles/components/table.scss
shleewhite: oh I did end up making the token and use it in the css now

packages/tokens/src/products/shared/advanced-table.json
jorytindall: Good catch on this, I missed a couple of these as well. Let's fix it [like this](https://www.figma.com/design/iweq3r2Pi8xiJfD9e6lOhF/branch/GWMDKHuGr98DFxSptH6vgj/HDS-Components-v2.0?m=auto&node-id=109244-44174&t=Oh0JBTVAqRwXrtLC-1) so the drag and drop indicators match the resizing.

Screenshot 2026-08-05 at 3 15 04 PM

dchyun (approved):

packages/tokens/src/products/shared/advanced-table.json
dchyun: New token looks good to me!

packages/tokens/src/products/shared/advanced-table.json
```diff
@@ -439,6 +439,21 @@
         }
       }
     },
+    "drop-preview": {
+      "surface": {
+        "color": {
+          "$type": "color",
+          "$value": "{focus.color.action.internal}",
```
didoo, line 446: would make sense here to alias directly to `{advanced-table.resize-handle.surface.color.focus}` which is the design intention, from what I understand? and also add a comment explaining why this connection (more explicitly)

didoo (approved):
One minor comment, I pre-approve anyway

packages/tokens/src/products/shared/advanced-table.json
shleewhite: the values are the same... but they are different features (resize vs reorder), so im hesitant to directly tie the tokens together.

packages/tokens/src/products/shared/advanced-table.json
jorytindall: Resolved in #4051

PR #3998, opened 2026-07-15
MCP: Improve build, testing, and resource error handling

zamoore:
### :pushpin: Summary

Improves MCP server reliability and developer workflows by preventing duplicate tool registration, safely handling resource errors, and separating build and type-check configurations.

- Wraps resource handlers with consistent error handling and adds unit coverage.
- Type-checks both source and tests while limiting build output to source files.
- Adds shared test utilities for resource responses and request handlers
- Removes duplicate tool registration during server init.

Focus review on MCP resource callback typing and error-response behavior, plus the separation between type-checking and build output.


packages/mcp/src/index.ts
```diff
@@ -40,8 +40,6 @@ const buildServer = (): McpServer => {
     version: getServerVersion(),
   });
 
-  registerTools(server);
```
shleewhite, line 43: why was this removed?

shleewhite (approved):

packages/mcp/src/index.ts
zamoore: It's called elsewhere. This was a second call from an older setup that was causing errors.

PR #4006, opened 2026-07-21
[Project Solar / Phase 2 / Tokens renaming] Tooling for HDS tokens replacement

didoo:
### :pushpin: Summary

This PR introduces a tooling (developed with the help of Copilot) that, given a target codebase, replace all the existing `--token` with the new "carbonized" (renamed/refactored) tokens across all the code (`scss/css` + `hds/gts/ts/js`).

#### Phase A - Mapping

The mapping between the old and the new tokens has been done in the "Phase A" of the tooling implementation. This mapping groups every old `--token-*` name into one of these buckets by how it changed during the "carbonization" of the codebase (I asked copilot to compare the tokens in`main` with the ones in the feature branch in https://github.com/hashicorp/design-system/pull/3237):

- known renames — old→new pairs that follow a rule:
    - prefix-only: just the `--token-`→`--hds-` swap, structure unchanged
    - systematic families: palette (`color-palette-*` → `core-color-*`), product (`color-{product}-*` → `product-{product}-*-color`), semantic (`color-{semantic}-*` → `{semantic}-color-*`), focus-ring, transition-function
    - other: structurally renamed but not matching any rule; mapped individually
- removed — old token with no successor (`after: null` → flagged with a TODO instead of rewritten)
- added — brand-new token in the post set only (`before: null` → reference-only, nothing to migrate)

#### Phase B - Script

"Phase B" instead is the actual implementation of a config-driven, agnostic script that, when pointed to a target repo:

- rewrites every stale `--token-*` token to its new `--hds-*` counterpart
- flags removed tokens (that don't have a counterpart) with TODO markers
- verifies that zero stale tokens remain

It's been validated against the `main` branch of this repo and dry-run against Atlas as consumer app.

> [!NOTE]
> The automatic replacement of tokens doesn't need to cover 100% of the cases, it's intended to do the heavy lifting, and then we can handle what's left (or the exceptions) manually. In any case, we can use the tooling developed in https://github.com/hashicorp/design-system/pull/3985 to validate the tokens after the replacement, and this will highlight what tokens still need to be manually handled post-refactoring.

### 👀 How to review

This is not code that will be used by our consumers, it's code that **we** will use when working at the carbonization of an existing codebase to automate as much as possible the migration of the refactored/renamed design tokens from the "old" names to the new ones. 

For this reason, I don't think it's necessary to do an in depth code review (the output has already been tested and the results were what we expected). Just have a general look, and go more in detail if there is something you want to see how it's done (or you see something strange/unexpected).

If you want to have a broad idea of what this tooling does, and how it works (and it's been implemented) these files may help:
- `copilot-plans/full-tokens-replacement/generated-plan.md`
- `copilot-plans/full-tokens-replacement/tooling/README.md`
- `copilot-plans/full-tokens-replacement/HANDOVER-full-tokens-replacement.md`

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-6504

***

<details>

</details>


dchyun:
Just for my own understanding what is the workflow steps to run this in a consumer repo? I'm comparing the files present here to those in your [atlas PR](https://github.com/hashicorp/atlas/pull/29836). From what I can tell, were these the steps you followed? 

1. Copied over `tooling/migration-token.mjs` (from phase 2)
2. Ran a prompt with the aI to generate `tooling/README.md` and `generated-plan.md`
3. Use the plan to generate all the other files like `config` and `reports`
4. Ran the token conversion from those generated files

Is this correct? I'm trying to understand what the workflow looks like from our side. It could be helpful to list out this workflow somewhere for us to reference when doing these conversions.

Also if the prompt you used to feed these things to the ai was non-trivial we could create a template for that prompt and include it here as well. Then people could just copy paste it into Bob.

copilot-plans/full-tokens-replacement/generated-plan.md
```diff
@@ -0,0 +1,316 @@
+# Generated Plan — HDS Token "Carbonization" Migration Tooling
+
+> Two-phase, config-driven, dependency-free tooling to migrate CSS design-token usages from the **pre-carbonization** token set to the **post-carbonization** token set.
+>
+> **Phase A** generates the old→new token map (an editable JSON that is the shared source of truth). **Phase B** applies that map to consumer code and verifies the result.
+>
+> Status: **TOOLING BUILT & VALIDATED — map accepted as generated.** Phase A implemented and run; its `tooling/reports/hds/token-map.generated.json` is accepted as-is (the `__other` = 6 and `removed` = 16 buckets were inspected and are correct — no manual finalization needed). Phase B implemented and validated end-to-end against a `main` worktree (1176 replacements, 34 TODOs, 33 interpolated skipped, 0 stale remaining, idempotent); the resulting applied diff was reviewed in a GUI client and approved. Next step: run Phase B against the real downstream target(s).
```
KristinLBradley, line 7: Q/Nit: Just wondered if there's a clearer word or words to use vs. "idempotent"? (I had to look it up myself.)

KristinLBradley: I guess it doesn't really matter though since Copilot is the real consumer of this...

KristinLBradley (approved):

copilot-plans/full-tokens-replacement/generated-plan.md
didoo: Yes, that word is used in both maths and computer science (hence the use by Copilot)

didoo:
> Just for my own understanding what is the workflow steps to run this in a consumer repo?
> It could be helpful to list out this workflow somewhere for us to reference when doing these conversions.

@dchyun the handover markdown file should be sufficient, I think, for both a human (https://github.com/hashicorp/design-system/blob/project-solar/phase-2/10/HDS-6504/full-tokens-replacement/copilot-plans/full-tokens-replacement/HANDOVER-full-tokens-replacement.md?plain=1#L132) as well as an AI tool.

> I'm comparing the files present here to those in your [atlas PR](https://github.com/hashicorp/atlas/pull/29836). 

Nah, that was just a test, it's a subset of what needs to be done for the migration. I have closed that PR now, it's outdated.

> Also if the prompt you used to feed these things to the ai was non-trivial we could create a template for that prompt and include it here as well. Then people could just copy paste it into Bob.

It's the handover file, it should be enough.

dchyun (approved):

PR #4007, opened 2026-07-21
MCP - Website tools

zamoore:
### :pushpin: Summary

If merged, this PR adds two new MCP tools - `hds_search_docs` and `hds_read_doc1 - that give AI assistants the ability to search and read the full Helios documentation catalog. It also adds the build pipeline (generation script, validation script, asset copy) that produces the `catalog.json` from the website's compiled output.

#### Docs catalog generation (`packages/mcp/scripts/`)

Three new scripts were added to generate, validate, and copy the docs catalog as part of the build:

- `generate-docs-catalog.mjs`: walks `website/dist/docs/`, parses each page's JSON output, splits content by tab and heading into section-level records, and writes `src/catalogs/docs/catalog.json`.
- `validate-docs-catalog.mjs`: validates a catalog against a Zod schema.
- `copy-assets.mjs`, copies the catalog JSON into `dist/` as part of the build since it isn't handled by the TypeScript compiler.

#### Docs store (`packages/mcp/src/tools/docs/store/`)

The store loads `catalog.json` lazily via `readFileSync` on first use (no static import), then builds an in-memory page map and pre-normalized search index used by both tools.

`hds_search_docs tool`

Keyword search over the catalog. Normalizes query and section text, scores by term frequency, filters by optional scope, and returns ranked results with a snippet and direct URL per section.

`hds_read_doc tool`

Paginated reading of a doc page. Accepts a `docId` + optional `sectionId` to start from a known section, or a `cursor` to continue from a previous call. Returns a `nextCursor` when more content remains.

### :link: External links

Jira ticket: [HDS-6535](https://hashicorp.atlassian.net/browse/HDS-6535)

***

<details>

</details>

[HDS-6535]: https://hashicorp.atlassian.net/browse/HDS-6535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/mcp/package.json
dchyun, line 27: [todo] The catlog build needs to be ran during the normal build step. Also right now `copy-assets` isn't called anywhere so it doesn't get added to the `dist`. To get it to run locally I did something like this

```
"catalog:docs:copy": "node ./scripts/copy-assets.mjs",
"catalog:docs:build": "pnpm -C ../../website build:packages && pnpm -C ../../website exec ember build --environment=development && pnpm catalog:docs && pnpm catalog:docs:copy",
```

```
"build": "pnpm tsc --project tsconfig.build.json && pnpm catalog:docs:build",
```

packages/mcp/scripts/generate-docs-catalog.mjs
```diff
@@ -0,0 +1,215 @@
...
+    .toLowerCase()
+    .trim()
+    .replace(/<[^>]+>/g, "")
+    .replace(/[^a-z0-9\s-]/g, "")
+    .replace(/\s+/g, "-")
+    .replace(/-+/g, "-");
+
+const splitSections = (markdown, pageTitle, tab) => {
+  const content = markdown.replace(/<!-- file included: .*? -->/g, "").trim();
+  const headingPattern = /^(#{2,6})\s+(.+)$/gm;
```
dchyun, line 61: [Discussion] I know this is just a first stab at this, but I'm curious what the best way to structure the docs catalog is to prioritize the most relevant information.

For example splitting up each individual header can make some content lose the context of the higher-level headers it was under. Would it better for example to only have objects for each level 3 and above header? 

Also would it be more helpful to reduce the amount of content to search and exclude pages we don't think are relevant for an MCP (about us, principles, etc) or are there ways we can help make use cases clearer? ex: Read "guidelines" for design guidance and "code" for code guidance

Not a blocker but it would be interesting to test different catalog formats for their effectiveness.

dchyun, line 61: [Issue?] Splitting up each header item is breaking apart the component api in a way which I think will hurt its searchability.

The "Component API" header is in its own object, and then the actual component api is fractured across a few different objects which don't have any context as to being related to a component api. Maybe there's a way to handle this section specially. 

Image

packages/mcp/scripts/generate-docs-catalog.mjs
```diff
@@ -0,0 +1,215 @@
...
+      level: match[1].length,
+      ...(tab === undefined ? {} : { tab }),
+      markdown: sectionMarkdown,
+    });
+  });
+
+  return sections;
+};
+
+const addSectionIds = (sections) => {
```
dchyun, line 109: [Question] Would it be possible to add in the page or component name into the section id? There are a lot of duplicate ids because a lot of pages have similar headings. Ex: `guidelines--overview` or `code--how-to-use-this-component`.

Not sure how relevant the IDs are to searches so can ignore this comment too.

packages/mcp/scripts/generate-docs-catalog.mjs
dchyun, line 1: [Suggestion] Might be worth excluding the version histories and changelog entries from the catalog to reduce noise.

packages/mcp/scripts/generate-docs-catalog.mjs
dchyun: Though I guess the component api here may not be as relevant if that information will come from the components catalog anyway.

packages/mcp/src/tools/docs/store/lookup.ts
```diff
@@ -0,0 +1,103 @@
...
+  
+  let score = 0;
+
+  if (title === query) {
+    score += 100;
+  } else if (title.startsWith(query)) {
+    score += 60;
+  } else if (title.includes(query)) {
+    score += 40;
+  }
```
dchyun, line 53:
```suggestion
  } else if (query.includes(title)) {
    score += 30;
  }
```

[Suggestion][Non-blocker] I know the score will need to be tweaked a lot based on testing so this is all non-blocking.

The scoring needs to be more flexible to queries which don't directly match a title or heading. One example here is only looking for `title.includes(query)` means the entire query needs to be found in the title. It would be beneficial to look for the inverse, if a query includes a title, since the title is just the component name in most cases.

The use case I found for this was the query `button colors`. When searching with the existing scoring all of the scores were very low, and thus the results not relevant. (You can see the scores in the output)

Image

When making this update the results were able to pick up on the "button" part of the query much easier.

Image

A few similar ideas to this tweak  + some others I was thinking about
- Check the title and heading for each "term" not just the whole query string
- Weight the searchable text term results more
- Have some list of component names and if the query includes a component name score that component page results higher

packages/mcp/src/tools/docs/store/lookup.ts
```diff
@@ -0,0 +1,103 @@
...
+  page: DocsPage;
+  section: DocsSection;
+  searchableText: string;
+}
+
+export const normalizeDocsText = (value: string): string =>
+  value
+    .toLowerCase()
+    .replace(/<!--([\s\S]*?)-->/g, " ")
+    .replace(/<[^>]+>/g, " ")
```
dchyun, line 24: [Issue?] This is stripping all of the HTML elements from the content. This means the returned results do not include any code snippets. Do we want to handle searching for code snippets in it's own unique way? 

I think it's valuable to be able to search the docs for the snippets. It's something the Carbon MCP does specifically with its own tool I believe.

This also is effecting the component api results as all of the `<Doc` elements are removed.

packages/mcp/src/tools/docs/store/schema.ts
```diff
@@ -0,0 +1,106 @@
...
+  scope: docsSearchScopeSchema.exclude(["all"]),
+  sections: z.array(docsSectionSchema),
+});
+
+export const docsCatalogSchema = z.object({
+  version: z.literal(1),
+  pages: z.array(docsPageSchema),
+});
+
+export const docsSearchResultSchema = z.object({
```
dchyun, line 36: [Suggestion] Would it be useful to include the "heading" in the search results as well?

packages/mcp/src/tools/docs/read-doc.ts
```diff
@@ -0,0 +1,61 @@
...
+import { getOrLoadDocsStore } from "./store/index.js";
+import { docsReadToolOutputSchema } from "./store/schema.js";
+
+import type { McpTool } from "../types.js";
+import type { DocsStore } from "./store/index.js";
+import type { DocsReadOutput } from "./store/schema.js";
+
+const inputSchema = {
+  docId: z.string().trim().min(1).optional(),
+  sectionId: z.string().trim().min(1).optional(),
```
dchyun, line 23: [Suggestion] Since the sections are more granular, usually only a paragraph, is there a way to incorporate the top level divisions of guidelines, code, accessibility, etc.? 

Like it's hard to know that to start reading the code docs you need to start with the `code--how-to-use-this-component` section. Would there be a way a request could say only show me the "guidelines" or "code" and start with the first section in that?

packages/mcp/tests/unit/tools/docs/store/index.test.ts
```diff
@@ -0,0 +1,101 @@
...
+      scope: "components",
+      limit: 10,
+    });
+
+    expect(results[0]).toMatchObject({
+      docId: "components/button/index",
+      sectionId: "guidelines--usage",
+      url: "/components/button?tab=guidelines#usage",
+    });
+    expect(results.every((result) => result.title === "Button")).toBe(true);
```
dchyun, line 41: [Issue?] I don't know if I'm doing something wrong, but when I search "button action" the results include sections which do not have "button" in the title.

Image

dchyun (commented):
Overall great to see we can actually translate all the website content into json and read it in the MCP! Very useful

packages/mcp/scripts/generate-docs-catalog.mjs
jorytindall: Great point, if possible I think it would be ideal to somehow group content together under h2s as that's the highest level of hierarchy in the heading structure. Everything grouped under an h2 until the subsequent h2 should ideally be related, if taken out of context it might be confusing.

zamoore:
Closing in favor of these stacked PRs

https://github.com/hashicorp/design-system/pull/4075
https://github.com/hashicorp/design-system/pull/4084

PR #4008, opened 2026-07-21
Segmented Group: add SuperSelect and SuperSelectMultiple yielded components

pkj-web:
### :pushpin: Summary

Adds `SuperSelect` and `SuperSelectMultiple` as yielded components to `Hds::SegmentedGroup` and updates the styling so they integrate seamlessly with existing segmented controls.

**Preview**: https://hds-showcase-git-pk-hds-4836-super-select-hashicorp.vercel.app/components/segmented-group#superselect

This PR adds support for using `SuperSelect` and `SuperSelectMultiple` within `Hds::SegmentedGroup`, allowing consumers to build segmented filtering UIs without requiring custom styling.

Specifically, this PR:

* Adds `SuperSelect` and `SuperSelectMultiple` as yielded components from `Hds::SegmentedGroup`.
* Updates the segmented group styling so the SuperSelect components match the appearance and behavior of the existing supported controls.
* Updates the showcase/examples to demonstrate the new supported usage.

This feature was requested by the ILM team as part of migrating existing dropdowns to SuperSelect.

Please verify that:

* `SuperSelect` and `SuperSelectMultiple` are correctly yielded by `Hds::SegmentedGroup`.
* The segmented group styling (borders, border radius, spacing, and focus states) is applied correctly.
* Existing yielded components continue to behave as expected.

*Add screenshots of the updated Segmented Group showcase demonstrating `SuperSelect` and `SuperSelectMultiple`.*
Screenshot 2026-07-21 at 3 56 39 PM

### :link: External links 

Jira ticket: [[HDS-4836](https://hashicorp.atlassian.net/browse/HDS-4836)](https://hashicorp.atlassian.net/browse/HDS-4836)
Figma file: N/A

---

* [x] Percy was checked for any visual regression
* [x] A changelog entry was added via [[Changesets](https://github.com/changesets/changesets)](https://github.com/changesets/changesets) if needed (see [[templates here](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates)](https://hashicorp.atlassian.net/wiki/spaces/HDS/pages/3243114706/Changelog+authoring+best+practices#Templates))

:speech_balloon: Please consider using [[conventional comments](https://conventionalcomments.org/)](https://conventionalcomments.org/) when reviewing this PR.

<details>

* [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.
* [ ] If applicable, I've worked with GRC to document the impact of any changes to security controls.
* [ ] If applicable, I've worked with GRC to ensure compliance due to a significant change to the in-scope PCI environment.

</details>

[HDS-4836]: https://hashicorp.atlassian.net/browse/HDS-4836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


shleewhite (commented):
Can you add a check that the Super Selects are rendered to this test? https://github.com/hashicorp/design-system/blob/31b8464047a4f0062bf64abfc57492e4580b6f85/showcase/tests/integration/components/hds/segmented-group/index-test.gts#L25

shleewhite (commented):
Another small thing to do in this PR, you need to generate a changelog entry.

.changeset/fine-peas-share.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+<!-- START components/segmented-group -->
+`SegmentedGroup` - Added `SuperSelect` and `SuperSelectMultiple` as yielded contextual components. Fixed a visual bug where the `SuperSelect` dropdown content appeared flush against the trigger when rendered inside `SegmentedGroup`.
```
shleewhite:
```suggestion
`SegmentedGroup` - Added `SuperSelect` and `SuperSelectMultiple` as yielded contextual components.
```

.changeset/fine-peas-share.md
```diff
@@ -0,0 +1,7 @@
+---
+"@hashicorp/design-system-components": patch
```
shleewhite:
```suggestion
"@hashicorp/design-system-components": minor
```

since this is a new feature, it is a minor change.

.changeset/fine-peas-share.md
shleewhite: You don't need to mention the visual bug thing since it was discovered & fixed in the process of adding the new feature.

packages/components/src/components/hds/form/super-select/multiple/base.gts
```diff
@@ -40,7 +40,7 @@ export interface HdsFormSuperSelectMultipleBaseSignature {
     isInvalid?: boolean;
   };
   Blocks: PowerSelectSignature['Blocks'];
-  Element: PowerSelectSignature['Element'];
```
shleewhite, line 43: Why did you make this change?

packages/components/src/components/hds/form/super-select/multiple/base.gts
```diff
@@ -184,7 +184,7 @@ export default class HdsFormSuperSelectMultipleBase extends Component<HdsFormSup
 
   <template>
     {{! Important: if an argument is added in base.hbs, it must also be added/processed in the Base component used in field.hbs }}
-    <div class={{this.classNames}} {{style this.styles}}>
+    <div class={{this.classNames}} {{style this.styles}} ...attributes>
```
shleewhite: Changing where the attributes are passed is a breaking change. Why was this necessary?

packages/components/src/styles/components/segmented-group.scss
```diff
@@ -10,10 +10,15 @@
 .hds-segmented-group {
   display: inline-flex;
 
+  > .hds-form-super-select .ember-basic-dropdown-content--in-place {
```
shleewhite, line 23: Can you add a comment explaining why this was necessary? Just so we dont forget in the future.

pkj-web: Yes, I will add a comment explaining why this was necessary. 👍

packages/components/src/components/hds/form/super-select/multiple/base.gts
pkj-web: Yeah, it was not necessary. I changed this so that I am use a similar format as the other base-test files. i.e select/base-test.gts "assert.dom('#test-form-select').hasClass('hds-form-select');" However, I reverted this change in the latest push.

packages/components/src/components/hds/form/super-select/multiple/base.gts
pkj-web: I made this change due to pretty much the same reason as the ...attributes. Since, I moved the attributes to the div and not the <PowerSelect />, I changed the element to match/follow. In the latest push/commit you can see the new base-test changes or you can say the original.

showcase/tests/integration/components/hds/form/super-select/single/base-test.gts
```diff
@@ -21,15 +21,10 @@ module(
     test('it should render the component with a CSS class that matches the component name', async function (assert) {
       await render(
         <template>
-          <HdsFormSuperSelectSingleBase
```
shleewhite, line 24: Can you revert the changes in this file too?

showcase/tests/integration/components/hds/form/super-select/multiple/base-test.gts
```diff
@@ -23,15 +23,10 @@ module(
     test('it should render the component with a CSS class that matches the component name', async function (assert) {
       await render(
         <template>
-          <HdsFormSuperSelectMultipleBase
```
shleewhite, line 26: Can you revert the changes in this file?

showcase/app/components/page-components/segmented-group/sub-sections/content.gts
```diff
@@ -2,7 +2,9 @@
  * Copyright IBM Corp. 2021, 2026
  * SPDX-License-Identifier: MPL-2.0
  */
-import type { TemplateOnlyComponent } from '@ember/component/template-only';
+import Component from '@glimmer/component';
+import { tracked } from '@glimmer/tracking';
+import { action } from '@ember/object';
```
shleewhite: Action is not needed in gts files, you can just use arrow functions.

showcase/app/components/page-components/segmented-group/sub-sections/content.gts
```diff
@@ -39,322 +41,457 @@ const DROPDOWN_OPTIONS = [
...
+const SUPER_SELECT_OPTIONS = [
+  'Boundary',
+  'Consul',
+  'Packer',
+  'Terraform',
+  'Vault',
+  'Waypoint',
+];
+
+export default class SubSectionContent extends Component {
```
shleewhite: We try to keep these files pretty focused and avoid having multiple states controlling demos  like this. Can you make a code fragment: `with-super-select`

In the fragment file, it would just render a segmented group with the super select and have a tracked state for the currently selected options + the set values function.

The code fragment would take a couple arguments:
 * type (idk about the name, but something like that): which accepts single or multiple and controls what type of super select is rendered.
 * position: which accepts leading, middle, or trailing and controls where in the group it is rendered


Then you can import those fragments into this file and it can stay a template only component.

The reason we do this is to make sure each demo is self contained and states cant affect other demos if you copy/paste the wrong state name.

Here is an example of another code fragment we reuse: https://github.com/hashicorp/design-system/blob/main/showcase/app/components/page-components/accordion/code-fragments/with-placeholder-content.gts

jorytindall:
I'm not sure if this is something we can resolve, but it's worth noting a difference between how the SuperSelect behaves in isolation and how it behaves within the SegmentedGroup:
- In isolation, the SuperSelect (at least as a Field) fills it's container and does not change size based on the selected items
- Within the SegmentedGroup the width is initially constrained to the width of the label of the selected item (or the placeholder). When an item (or items) is selected, the width changes causing some layout shift

CleanShot 2026-07-28 at 12 15 53

I'm not sure if there's any easy solution to prevent the layout shift other than specifically targeting the SuperSelect or if that's worth it. The default Select component also does not have this problem.

shleewhite:
@jorytindall ahh that is a good point.. I feel like we wouldnt know what the right max width would be as a baseline, but a consumer could theoretically add their own css to constrain the width of the segmented group.

For the select, the input takes the width of the longest option and we dont have any examples of super long options.
Screenshot 2026-07-28 at 4 44 47 PM

shleewhite:
@hashicorp/hds-engineering I added a `@width` argument to the SuperSelect. It didnt fully end up working as intended, because if there is not enough room in the container for the specified width, it will truncate the SuperSelect.

See this example where both SuperSelects have `@width="300px"`:
Screenshot 2026-08-05 at 2 23 29 PM

The issue of the SuperSelect changing width has gone away because now it fills all the available space, but it feels odd as a consumer to set a width and not have it fully respected.

[Showcase examples](https://hds-showcase-git-pk-hds-4836-super-select-hashicorp.vercel.app/components/segmented-group#superselect)

didoo:
> The issue of the SuperSelect changing width has gone away because now it fills all the available space, but it feels odd as a consumer to set a width and not have it fully respected.

@shleewhite I _think_ you have to apply a `flex-shrink: 0` to the SuperSelect, not sure if via CSS (always) or via JS (only when there is a width applied to it). @KristinLBradley what do you think?

shleewhite:
> > The issue of the SuperSelect changing width has gone away because now it fills all the available space, but it feels odd as a consumer to set a width and not have it fully respected.
> 
> @shleewhite I _think_ you have to apply a `flex-shrink: 0` to the SuperSelect, not sure if via CSS (always) or via JS (only when there is a width applied to it). @KristinLBradley what do you think?

no... because then it just takes the width from the adjacent inputs like this:
Screenshot 2026-08-06 at 8 29 04 AM

didoo:
> no... because then it just takes the width from the adjacent inputs like this

yeah, but that is correct, no? If I set a width of 300px then the width is 300px. I don't know in a flexbox item if it's possible to do proportional widths, using relative units for `width` (or maybe `flex-basis`) can you try?

shleewhite:
@didoo to me, this is where the `isInline` argument on the segmented group would be helpful and just let the segmented group grow to fill the available space

def not thinking of trying to maintain proportions or anything super complicated.

didoo:
> @didoo to me, this is where the `isInline` argument on the segmented group would be helpful and just let the segmented group grow to fill the available space
> 
> def not thinking of trying to maintain proportions or anything super complicated.

Yeah but in this case, we're doing too much (from a consumer perspective):

- consumer: _I have a problem, the SuperSelect I've added in the SegmentedGroup changes size whenever I select an option and the content shifts_
- us: _apply an `@inline=false` to the SegmentedGroup, and the effect should be mitigated; ah btw now your SegmentedGroup will fill all its available space too_ 
- consumer: mmmm 🤔 not what I was expecting

Unless what we want is that the SuperSelect can grow (and shift content) but up to a point, in which case we can use `max-width` instead (but in that case probably the argument should be `@maxWidth` to avoid confusion) 

Attached is a patch of a few tests I've done on the showcase (and components) to test different ideas/solutions:
[patch.diff.zip](https://github.com/user-attachments/files/30795340/patch.diff.zip)

BTW, probably we should decouple this issue to a separate ticket, to allow @pkj-web to merge their code changes :)

packages/components/src/components/hds/segmented-group/index.gts
```diff
@@ -20,6 +22,8 @@ export interface HdsSegmentedGroupSignature {
         Dropdown?: typeof HdsDropdown;
         Select?: typeof HdsFormSelectBase;
         TextInput?: typeof HdsFormTextInputBase;
+        SuperSelect?: typeof HdsFormSuperSelectSingleBase;
```
KristinLBradley: Would it make sense to instead call this `SuperSelectSingle` to contrast with `SuperSelectMultiple`?

KristinLBradley:
> > The issue of the SuperSelect changing width has gone away because now it fills all the available space, but it feels odd as a consumer to set a width and not have it fully respected.
> 
> @shleewhite I _think_ you have to apply a `flex-shrink: 0` to the SuperSelect, not sure if via CSS (always) or via JS (only when there is a width applied to it). @KristinLBradley what do you think?

@didoo As @shleewhite posted, adding `flex-shrink: 0` to the SuperSelect can cause it to steal too much width from other "segments" or can cause the segments to exceed the bounds of the hds-segmented-group parent element if you try to force a wider width on the other segments (adding up to greater than 100% width of the parent).

I tried experimenting with different combos of widths vs. max-width, etc and so far there's trade-offs with everything.

Maybe it would help to first identify the ideal behavior that we would like for the `SuperSelect` width vs. the other sgment widths first? And then see how close we can get to that. (Would we like its width to behave more like a Form `select` in which it takes up the width of the widest option? That seems like it's probably not feasible though so what would our 2nd ideal behavior be for its width? How about the widths of all the other segments, either on their own or in relationship to each other?)

showcase/app/components/page-components/segmented-group/sub-sections/content.gts
```diff
@@ -340,6 +343,37 @@ const SubSectionContent: TemplateOnlyComponent = <template>
 
   <ShwDivider @level={{2}} />
 
+  <ShwTextH3>SuperSelect</ShwTextH3>
+
+  <ShwTextBody>
+    Note: all of the SuperSelects have
```
KristinLBradley: Can you please add examples of SuperSelects without a width set and with a maxWidth value of 300px set instead of width for comparison?

KristinLBradley (commented):
Would it be possible to change the behavior of the maxWidth property to work the same as the new width property when the `SuperSelect` is in the context of a `SegmentedGroup`?

Since the new width property actually behaves more like you would expect a `maxWidth` to behave that may make more sense to consumers.

showcase/tests/integration/components/hds/segmented-group/index-test.gts
```diff
@@ -20,6 +21,46 @@ module('Integration | Component | hds/segmented-group/index', function (hooks) {
...
+  test('it should not render the full-width CSS class by default', async function (assert) {
+    await render(
+      <template><HdsSegmentedGroup id="test-segmented-group" /></template>,
+    );
+    assert
+      .dom('#test-segmented-group')
+      .doesNotHaveClass('hds-segmented-group--is-full-width');
+  });
+
+  test('it should render the full-width CSS class when @isFullWidth is true', async function (assert) {
```
KristinLBradley:
```suggestion
  test('it renders the full-width CSS class when @isFullWidth is true', async function (assert) {
```

Nit: for clarity/simplicity

showcase/tests/integration/components/hds/segmented-group/index-test.gts
```diff
@@ -20,6 +21,46 @@ module('Integration | Component | hds/segmented-group/index', function (hooks) {
...
+      <template>
+        <HdsSegmentedGroup id="test-segmented-group" @isFullWidth={{true}} />
+      </template>,
+    );
+    assert
+      .dom('#test-segmented-group')
+      .hasClass('hds-segmented-group--is-full-width');
+  });
+
+  test('it should set the max-width CSS custom property when @maxWidth is provided', async function (assert) {
```
KristinLBradley:
```suggestion
  test('it sets the max-width CSS custom property when @maxWidth is provided', async function (assert) {
```

KristinLBradley (dismissed):
Looks good, nice work!

Added a few suggested wording edits in tests but feel free to ignore if you don't think they're helpful/necessary

KristinLBradley:
Try re-basing on main to clear some of the visual diffs in Percy.

jorytindall (dismissed):

showcase/tests/integration/components/hds/segmented-group/index-test.gts
pkj-web: Sounds good!

KristinLBradley (approved):
### ⛵𓂃

PR #4017, opened 2026-07-27
`AdvancedTable` focus bug fix

zamoore:
### :pushpin: Summary

Clicking a cell in `AdvancedTable` made the page jump. Removes the `center` scroll alignment on `Td`/`Th` so focus scrolls by the minimum distance instead.

Cells are `Composite` items with a `tabindex` so clicking one focuses it and fires `hdsScrollIntoViewOnFocus`. That modifier was passed `block="center" inline="center"`, and `scrollIntoView` walks up every scroll ancestor including the document, so `center` recentered the whole page even when the cell was already visible.

Dropping the override falls back to the default `nearest` behavior.

#### Background

The bug was introduced in #3818 when we switched keyboard navigation to use the Composite component.

### :link: External links

Jira ticket: [HDS-6545](https://hashicorp.atlassian.net/browse/HDS-6545)

***

<details>

</details>

[HDS-6545]: https://hashicorp.atlassian.net/browse/HDS-6545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


didoo:
@zamoore can you provide links/references to where this bug was introduced (in the PR description)? in this way we can compare the before/after and have more context about the fix itself

.changeset/whole-crabs-lay.md
```diff
@@ -0,0 +1,5 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+`AdvancedTable` - Fixed an issue where clicking a cell scrolled the page to center that cell
```
dchyun, line 7: [todo] Add start and end comments around entry

packages/components/src/components/hds/advanced-table/th.gts
```diff
@@ -468,7 +468,7 @@ export default class HdsAdvancedTableTh extends Component<HdsAdvancedTableThSign
         )
       }}
       {{@compositeItem disabled=@isCompositeItemDisabled}}
-      {{hdsScrollIntoViewOnFocus options=(hash block="center" inline="center")}}
+      {{hdsScrollIntoViewOnFocus}}
```
dchyun: [Issue?] In tables which have horizontal scrolling, clicking on a partially visible cell scrolls the table to that cell. 

https://github.com/user-attachments/assets/b0c53b89-13af-41aa-99d3-8feb28e7c8d6

I was looking into the page scrolling issue a bit with Bob and it has suggested wrapping the `scrollIntoView` into a check that the table cell has `:focus-visible`. Maybe this could help? Seemed to work when testing locally, but checking a css selector through js to trigger it is a bit quirky. 

```
const onFocus = (): void => {
  if (element.matches(':focus-visible')) {
    element.scrollIntoView(scrollOptions);
  }
};
```

zamoore: I'm personally ok with this behavior, but I'm curious what design has to say

didoo (commented):
There is something that I don't understand: I checked out the code at this PR merge https://github.com/hashicorp/design-system/pull/3818 and run the showcase; when I click on a cell inside the AdvancedTable, nothing happens (there is no movement of the scrolled table, no centering of the clicked cell). 

This tells me that between then and now, a core behaviour of the AdvancedTable has been changed (but I don't see mention of this change in [the version history of the component](https://helios.hashicorp.design/components/table/advanced-table?tab=version%20history)). Now the table "centers" the clicked cell, which is something... unexpected, as a user. Why do we infer that, if I click a cell, I also intend to center it in the table? This is an assumption that we're making, and I am not sure it's expected (I checked other data tables implementations, and I could not find similar behaviour) expecially because it shifts the visual position of the content and is really disorienting.

I would challenge this implementation (sorry I didn't notice before) and would consider removing entirely this "feature" (and with this, also avoid the bug).

cc @shleewhite @alex-ju

shleewhite:
> There is something that I don't understand: I checked out the code at this PR merge #3818 and run the showcase; when I click on a cell inside the AdvancedTable, nothing happens (there is no movement of the scrolled table, no centering of the clicked cell).
> 
> This tells me that between then and now, a core behaviour of the AdvancedTable has been changed (but I don't see mention of this change in [the version history of the component](https://helios.hashicorp.design/components/table/advanced-table?tab=version%20history)). Now the table "centers" the clicked cell, which is something... unexpected, as a user. Why do we infer that, if I click a cell, I also intend to center it in the table? This is an assumption that we're making, and I am not sure it's expected (I checked other data tables implementations, and I could not find similar behaviour) expecially because it shifts the visual position of the content and is really disorienting.
> 
> I would challenge this implementation (sorry I didn't notice before) and would consider removing entirely this "feature" (and with this, also avoid the bug).
> 
> cc @shleewhite @alex-ju

@didoo this feature is needed for keyboard users. without the scrolling behavior, your focus can be outside of the view which is a wcag failure.

I believe this was introduced when we switched to the composite component.

as for other workarounds, one option @zamoore and I talked about was to add a check to the scrolling function that checks if the focus event was initiated by the cursor or keyboard. If it is keyboard initiated, then do the scroll. otherwise, ignore. does that option seem better to you?

zamoore:
> other

@didoo @shleewhite - Yes, that is correct. It wasn't explicitly a feature added to the `AdvancedTable`, it was just a change in the implementation of an existing feature that introduced a regression.

Edit: Ah, you're saying that the bug is not present in #3818 . That's interesting. I'm not certain what else would have changed between then and now?

didoo:
> @didoo this feature is needed for keyboard users. without the scrolling behavior, your focus can be outside of the view which is a wcag failure.

Yes, that's what I meant (hence the use of the word `click`). For keyboard users obviously moving the focused cell into view is the expected behaviour/intent (as user).

zamoore:
@didoo @shleewhite - I updated the code to only react to keyboard events

packages/components/src/modifiers/hds-scroll-into-view-on-focus.ts
didoo, line 1: [super-mega-nit] changes to this file should be in the "fixed bug" commit, not the "updated test" commit

didoo (commented):
There may be an issue (?) with sticky header/columns, where the cell focused via keyboard is not visible (it's behind the sticky header/column):

https://github.com/user-attachments/assets/f8a29e6d-39ce-4ba2-a98c-fda4112bb940

No idea how (and even if) this can be solved.

packages/components/src/components/hds/advanced-table/td.gts
```diff
@@ -121,7 +122,7 @@ export default class HdsAdvancedTableTd extends Component<HdsAdvancedTableTdSign
         )
       }}
       {{@compositeItem disabled=@isCompositeItemDisabled}}
-      {{hdsScrollIntoViewOnFocus options=(hash block="center" inline="center")}}
+      {{hdsScrollIntoViewOnFocus}}
```
shleewhite: @didoo I think if we go back to having it scroll to the center it should always be visible

packages/components/src/modifiers/hds-scroll-into-view-on-focus.ts
```diff
@@ -25,6 +25,11 @@ const hdsScrollIntoViewOnFocus =
       const scrollOptions = named.options ?? DEFAULT_OPTIONS;
 
       const onFocus = (): void => {
+        // only scroll for keyboard-driven focus
+        if (element.matches(':focus-visible') === false) {
```
dchyun, line 29: [Nit] Would it be simpler to do the inverse of this?

```
if (element.matches(':focus-visisble') {
  element.scrollIntoView(scrollOptions);
}
```

packages/components/src/components/hds/advanced-table/td.gts
dchyun: Now that we have the check to only run this when focused via keyboard we could revert to what was there before.

packages/components/src/components/hds/advanced-table/th.gts
dchyun: With the change to run the scroll on keyboard focus only this issue is fixed.

jorytindall:
I found this while testing the resizing PR #4020 but it might be more related to this; traversing the grid with arrow keys works fine until I sort a column, which then seems to keep the same focus order but results in jumping around because the focus order is different from the visual order of the cells in the column. 

Is this a known issue already or expected?

CleanShot 2026-07-29 at 09 56 32

zamoore:
> I found this while testing the resizing PR #4020 but it might be more related to this; traversing the grid with arrow keys works fine until I sort a column, which then seems to keep the same focus order but results in jumping around because the focus order is different from the visual order of the cells in the column.
> 
> Is this a known issue already or expected?
> 
> CleanShot 2026-07-29 at 09 56 32

This isn't related, but it's also not a known issue. Can you create a ticket so we can address it on its own? Good find

packages/components/src/modifiers/hds-scroll-into-view-on-focus.ts
zamoore: I prefer early returns but that's just a stylistic thing

packages/components/src/components/hds/advanced-table/td.gts
zamoore: Agreed

.changeset/whole-crabs-lay.md
```diff
@@ -0,0 +1,8 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+<!-- START components/table/advanced-table -->
+
+`AdvancedTable` - Fixed an issue where clicking a cell scrolled the page to center that cell
+<<!-- END -->>
```
shleewhite:
```suggestion
<!-- END -->
```

shleewhite (approved):

PR #4039, opened 2026-08-04
[Project Solar / Phase 1 / Release] 03 - `HdsTheming` service and `HdsThemeContext` component

didoo:
### :pushpin: Summary

This is the third of a set of stacked PRs that cherry-pick the changes from https://github.com/hashicorp/design-system/pull/3237 into `main`.

This PR cherry picks the implementation of the `HdsTheming` service, broadly used across the codebase, and the `HdsThemeContext` component, a headless utility component to apply local/contextual themes.

> [!NOTE]
> This PR does **not** include the `HdsThemeSwitcher` component, which was implemented in https://github.com/hashicorp/design-system/pull/3237 as demo only; since the design has decided to provide a pattern to the consumers, not a plug&play component, we don't want to bring in that exploration code in `main`.

### 🚧 TODOs
- [ ] understand what to do with the `start/end` delimiter in the changeset file (https://github.com/hashicorp/design-system/pull/4039#discussion_r3865378376)

### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-6574

***

> [!WARNING]
> **Note**:  tests are expected to fail in this PR, since the code changes are cherry-picked and only partial; what we expect is that the tests pass in the last PRs in the stack, where all the code changes become full and consistent.

<details>

</details>

### DIFF
--- a/.changeset/carbonization-components.md
+++ b/.changeset/carbonization-components.md
@@ -1,0 +1,11 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+Added `HdsTheming` service for theming
+
+<!-- START theming/theme-context -->
+`ThemeContext` - Added headless component
+<!-- END -->
+
+
--- a/packages/components/src/components/hds/theme-context/index.gts
+++ b/packages/components/src/components/hds/theme-context/index.gts
@@ -1,0 +1,73 @@
+/**
+ * Copyright (c) HashiCorp, Inc.
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import Component from '@glimmer/component';
+import { assert } from '@ember/debug';
+import type Owner from '@ember/owner';
+
+import {
+  HdsThemeContextThemesValues,
+  HdsThemeContextModesValues,
+} from './types.ts';
+import type { HdsThemeContexts } from './types.ts';
+
+import type { HdsThemes, HdsModes } from '../../../services/hds-theming.ts';
+
+export interface HdsThemeContextSignature {
+  Args: {
+    // can be either an `HdsTheme` or an `HdsMode`
+    context: HdsThemeContexts;
+  };
+  Blocks: {
+    default: [];
+  };
+  Element: HTMLElement;
+}
+
+export const CONTEXTUAL_THEMES: HdsThemes[] = Object.values(
+  HdsThemeContextThemesValues
+);
+export const CONTEXTUAL_MODES: HdsModes[] = Object.values(
+  HdsThemeContextModesValues
+);
+export const CONTEXTUAL_VALUES: HdsThemeContexts[] = [
+  ...CONTEXTUAL_THEMES,
+  ...CONTEXTUAL_MODES,
+];
+
+export default class HdsThemeContext extends Component<HdsThemeContextSignature> {
+  constructor(owner: Owner, args: HdsThemeContextSignature['Args']) {
+    super(owner, args);
+
+    const { context } = args;
+
+    assert(
+      `@context for "Hds::ThemeContext" must be one of the following: ${CONTEXTUAL_VALUES.join(
+        ', '
+      )}; received: ${context}`,
+      CONTEXTUAL_VALUES.includes(context)
+    );
+  }
+
+  // Get the class names to apply to the component.
+  get classNames(): string {
+    const classes = ['hds-theme-context'];
+
+    const { context } = this.args;
+
+    // add "theme" or "mode" classes based on the @context arguments
+    if (CONTEXTUAL_THEMES.includes(context as HdsThemes)) {
+      classes.push(`hds-theme-${context}`);
+    } else if (CONTEXTUAL_MODES.includes(context as HdsModes)) {
+      classes.push(`hds-mode-${context}`);
+    }
+
+    return classes.join(' ');
+  }
+
+  <template>
+    <div class={{this.classNames}} ...attributes>{{yield}}</div>
+  </template>
+}
--- a/packages/components/src/components/hds/theme-context/index.gts
+++ b/packages/components/src/components/hds/theme-context/index.gts
@@ -23,7 +23,7 @@ export interface HdsThemeContextSignature {
   Blocks: {
     default: [];
   };
-  Element: HTMLElement;
+  Element: HTMLDivElement;
 }
 
 export const CONTEXTUAL_THEMES: HdsThemes[] = Object.values(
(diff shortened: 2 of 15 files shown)


.changeset/carbonization-components.md
```diff
@@ -0,0 +1,11 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+Added `HdsTheming` service for theming
+
+<!-- START theming/theme-context -->
```
didoo, line 7: [TODO] @didoo understand what we need to do in order for this to be correctly processed (since this page does not exist in the website)

shleewhite: We could add a changelog section to the foundations/tokens page? seems like a logical place to put information about the context/how to use the themes.

otherwise, we could make a utilities/theme-context page

packages/components/src/components/hds/theme-context/index.gts
```diff
@@ -0,0 +1,73 @@
...
+
+export interface HdsThemeContextSignature {
+  Args: {
+    // can be either an `HdsTheme` or an `HdsMode`
+    context: HdsThemeContexts;
+  };
+  Blocks: {
+    default: [];
+  };
+  Element: HTMLElement;
```
shleewhite:
```suggestion
  Element: HTMLDivElement;
```

shleewhite (commented):
Looks good, just had one small suggestion.

.changeset/carbonization-components.md
shleewhite: Looking at the docs branch, this path seems fine. you just need to add it to the list of options here: https://github.com/hashicorp/design-system/blob/9883338afec75ddf9ba73e3784c0ce098a5e6a95/website/lib/changelog/generate-component-changelog-entries.mjs#L23

shleewhite (dismissed):
Changes look good! Just need to decide if the website changelog script should be updated in this pr or in the docs branch

.changeset/carbonization-components.md
didoo: Ohhhhh, I forgot about this! Thanks 🙏

showcase/tests/unit/services/hds-theming-test.js
```diff
@@ -0,0 +1,244 @@
...
+    assert.ok(
+      onSetThemeSpy.calledWith({
+        currentTheme: HdsThemeValues.Light,
+        currentMode: DEFAULT_THEMING_OPTION_LIGHT_THEME,
+      }),
+      'onSetTheme callback received correct arguments',
+    );
+  });
+
+  test('getters return correct values', function (assert) {
```
dchyun, line 220: [Suggestion] Should we test the `isCarbonThemeEnabled` getter as well?

didoo: Done

dchyun (dismissed):

packages/components/src/components/hds/theme-context/index.gts
```diff
@@ -0,0 +1,73 @@
...
+  get classNames(): string {
+    const classes = ['hds-theme-context'];
+
+    const { context } = this.args;
+
+    // add "theme" or "mode" classes based on the @context arguments
+    if (CONTEXTUAL_THEMES.includes(context as HdsThemes)) {
+      classes.push(`hds-theme-${context}`);
+    } else if (CONTEXTUAL_MODES.includes(context as HdsModes)) {
+      classes.push(`hds-mode-${context}`);
```
dchyun, line 64: [Issue] When using a `@mode` an `hds-theme-{light/dark/system}` class is not present on the context. This means any `hds-apply-only-if-carbon` styles must get their selector from the outer level page.

The `hds-apply-only-if-carbon` styles only rely on the `hds-theme-` classes. Because the context does not add those classes when using a mode it leads to some slight issues

1. When the page theme is dark but the context is light the selector used inside the context is still `hds-theme-dark`. (The tokens are still correct values because they are defined at the context level)

Image

2. When using a Carbon mode inside an HDS theme page (not a use case in practice, but comes up on the docs) the `hds-apply-only-if-carbon` styles are not applied at all.

In this example the alert is missing the wider carbon-theme left border, which is applied with the mixin

Image

packages/components/src/components/hds/theme-context/index.gts
dchyun, line 1: [Issue] When a carbon context is used inside an HDS theme page any carbon-theme functionality of a component dependent on the `hdsTheming` service is not triggered because the service only depends on the global theme.

Ex: In the alert the icon rendered is different for the carbon themes, but in this use case the HDS icon is still being shown.

**Expected icon**

Image

**Actual icon**

Image

dchyun (commented):
In reviewing #4117 I caught a few issues with the `HdsThemeContext` when it is used to set a carbon theme inside an hds / default theme page. This is not a use case in practice, but it does cause some visual issues for situations like the docs example.

packages/components/src/components/hds/theme-context/index.gts
didoo: @dchyun you're right. I've added a commit _"explicitly added “theme” class to “mode” contexts"_ with the changes suggested, have a look.

packages/components/src/components/hds/theme-context/index.gts
didoo: @dchyun yes, this is a known limitation and I don't know how to work around this (unless you have an idea). 

In theory, the `ThemeContext` will be used to create different contexts within the light/dark modes (and not to mix "classic" themes with "carbon" themes. And in fact, the `hdsTheming` service in the HDS components is used (from a logic perspective) only to discriminate when carbon is active, so this should be safe to use as is now:
screenshot_1559

What I am going to do is to specify these limitations when I document it.

packages/components/src/components/hds/theme-context/index.gts
dchyun: Looks great thanks for adding that. I think this will clear up the largest point of confusion for the docs examples since the mixin is used widely.

packages/components/src/components/hds/theme-context/index.gts
dchyun: Yeah I think it's not a limitation worth fighting to fix. I don't have any ideas either. I think documenting the limitation is the best idea.

In these nested context examples I would also avoid using components which rely on the theming service (like the alert) just to avoid confusion.

didoo: I have updated the documentation for the `ThemeContext component, let me know what you think: https://hds-website-git-project-solar-phase-1-docs-upd-599b91-hashicorp.vercel.app/theming/theme-context

dchyun: Docs updates look good to me

PR #4062, opened 2026-08-11
MCP - Components resources

zamoore:
### 📌 Summary

If merged, this PR adds MCP resources for discovering HDS components and retrieving detailed component metadata from the component catalog.

### 🛠️ Detailed description

- Adds the hds://components catalog resource.
- Adds the hds://components/{componentName} detail resource.
- Adds catalog validation, lazy loading, and normalized component-name lookup.
- Supports case-insensitive names with optional Hds prefixes.
- Adds component-name completion and structured not-found responses.
- Deletes existing component catalog in favor of one generated by typescript.
- Adds unit and integration tests for stores, resources, completion, serialization, registration, and the generated build artifact.
- Adds script to the components project for generating a manifest json file from type definitions and derived website documentation paths

### :link: External links

Jira ticket: [HDS-6532](https://hashicorp.atlassian.net/browse/HDS-6532)

***

<details>

</details>

[HDS-6532]: https://hashicorp.atlassian.net/browse/HDS-6532?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/mcp/src/resources/components/store/index.ts
```diff
@@ -0,0 +1,76 @@
...
+    totalComponentCount: number;
+  };
+  listComponents: () => ComponentSummary[];
+  getComponentByName: (componentName: string) => ComponentRecord | null;
+}
+
+const require = createRequire(import.meta.url);
+
+const getComponentCatalogPath = (): string => {
+  return require.resolve("@hashicorp/design-system-components/component-catalog.json");
```
dchyun, line 30: [Question] Because this `component-catalog` is re-generated on every components build, is there a risk that the MCP could be referencing features yet to be released? If new features are pushed to the components package, but not released yet could the MCP be referencing those?

Maybe the component catalog could only be re-generated as part of the release procedure?

packages/mcp/src/resources/components/get-component-by-name.ts
```diff
@@ -0,0 +1,136 @@
...
+    return componentName;
+  }
+};
+
+export const completeComponentNames = (
+  components: ComponentSummary[],
+  value: string,
+  limit = 100,
+): string[] => {
+  const query = value.trim().toLowerCase();
```
dchyun, line 33: [Suggestion] It would be helpful to support searching for the gts component names as well without the `::`. Right now searches for the gts names do not return the component information.

Image

packages/mcp/tests/unit/resources/components/store/index.test.ts
```diff
@@ -0,0 +1,98 @@
...
+  const catalog = {
+    components: [
+      buildComponentCatalogEntry(),
+      buildComponentCatalogEntry({
+        name: "Hds::AdvancedTable::Th",
+        modulePath: "hds/advanced-table/th",
+        docsPath: "components/table/advanced-table",
+        element: "HTMLTableCellElement",
+        args: [],
+        blocks: [{ name: "default", yields: [] }],
```
dchyun, line 53: [Suggestion] Could also add a test for an entry which has yielded blocks that aren't just the default.

dchyun (commented):
Overall server seems to be running well and getting all the expected information

packages/mcp/src/resources/components/store/index.ts
zamoore: Good catch and the answer is sort of complicated.

Published: no risk. `prepublishOnly` runs `pnpm build`, which regenerates the catalog, so each tarball's catalog is generated from exactly the source being published. It always matches the version you installed.

Monorepo checkout: yes, real. The committed catalog tracks src, so between a merge and its release it describes unreleased API, and that's the only way to run the server today. Mostly it lands on contributors already working on unreleased code, but it would bite anyone checking out the repo just to run the MCP while writing app code against a release.

TLDR; It should be fine for now. In the version that is ultimately released it won't be an issue.

zamoore: Ultimately it will build the JSON file from whatever version the consumer is running (as long as they install the MCP server as a dev dependency)

packages/mcp/src/resources/components/get-component-by-name.ts
```diff
@@ -0,0 +1,136 @@
...
+    const aliases = [component.name, component.modulePath];
+    const isMatch =
+      query.length === 0 ||
+      aliases.some((alias) => alias.toLowerCase().includes(query));
+
+    if (isMatch) {
+      matches.push(component.name);
+    }
+
+    if (matches.length >= limit) {
```
shleewhite, line 50: Is there a technical reason to limit the number of matches? thinking if you search "table" or "input" you may want to see multiple components

packages/components/component-catalog.json
```diff
@@ -0,0 +1,12293 @@
...
+      ],
+      "blocks": [
+        {
+          "name": "default",
+          "yields": []
+        }
+      ]
+    },
+    {
+      "name": "Hds::AdvancedTable::ThReorderHandle",
```
shleewhite, line 2035: Should there be a concept of "private" components? Would we want people to know about the sub-components within AdvancedTable?

Im sure theres other examples of sub-components we dont want people to use directly

packages/mcp/src/resources/components/get-component-by-name.ts
zamoore: Good call. Added

packages/mcp/tests/unit/resources/components/store/index.test.ts
zamoore: Good call. Added

packages/mcp/src/resources/components/get-component-by-name.ts
zamoore: It's a limitation of the SDK. It will only return 100 items, but will also return hasMore: true, which should mitigate that.

packages/components/component-catalog.json
zamoore: It doesn't hurt for the MCP to have knowledge of some of the "private" components. It will pull implementation details from the docs anyways, so there shouldnt be a risk of the MCP recommending using the components directly.

packages/mcp/src/resources/components/get-component-by-name.ts
shleewhite: Sounds good! That makes sense.

dchyun (approved):

PR #4074, opened 2026-08-20
MCP: Fail when the committed component catalog is stale

zamoore:
### :pushpin: Summary

`component-catalog.json` records a `docsPath` for every component, the website page
that documents it. The MCP server hands those paths to AI agents as links.

The generator matches a component to its page using the page's `links.github`
frontmatter, not the page's URL. That's deliberate and normally good: rename a docs
folder and it still finds the right page. But it also means a rename **silently
changes the answer** instead of failing.

So: a docs-only PR renames `website/docs/components/badge` -> `components/status-badge`.
CI rebuilds the catalog, gets the new path, and throws it away, nothing compares it
to the committed file. The PR merges green, and `main` now ships a pointer to a page
that no longer exists. Every future build stays green too, because the generator keeps
self-correcting in a workspace nobody inspects.

Regenerate the catalog, then `git diff HEAD` it. If the committed file no longer
matches what the sources produce, fail and print the diff.

`ci-components` and `ci-website` already run the build that regenerates it, so this is
one added step in each rather than a new job. They cover different causes: a docs
rename (`ci-website`) and a source change whose catalog was never committed
(`ci-components`).

When it fails, the corrected file is already in your working tree — commit it, and the
rename and its fix land in the same PR.

### :link: External links

Jira ticket: [HDS-6598](https://hashicorp.atlassian.net/browse/HDS-6598)

***

<details>

</details>

[HDS-6598]: https://hashicorp.atlassian.net/browse/HDS-6598?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/scripts/check-component-catalog.mjs
+++ b/packages/components/scripts/check-component-catalog.mjs
@@ -1,0 +1,53 @@
+#!/usr/bin/env node
+/**
+ * Copyright IBM Corp. 2021, 2026
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+// fails when the committed `component-catalog.json` is not what the current sources produce
+
+import { execFileSync } from 'node:child_process';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const PACKAGE_ROOT = path.resolve(fileURLToPath(import.meta.url), '../..');
+const CATALOG_FILE = 'component-catalog.json';
+const GENERATOR = './scripts/build-component-catalog.mjs';
+const MAX_DIFF_LINES = 100;
+
+function fail(message) {
+  console.error(`\n\x1b[31m⚠️  Error: ${message}\x1b[0m\n`);
+
+  process.exit(1);
+}
+
+function git(args) {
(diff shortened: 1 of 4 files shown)


packages/components/scripts/check-component-catalog.mjs
```diff
@@ -0,0 +1,53 @@
...
+  return execFileSync('git', args, { cwd: PACKAGE_ROOT, encoding: 'utf8' });
+}
+
+try {
+  execFileSync(process.execPath, [GENERATOR], {
+    cwd: PACKAGE_ROOT,
+    stdio: 'inherit',
+  });
+} catch {
+  process.exit(1);
```
alex-ju, line 34: I would consider capturing and printing the caught error to make CI logs easier to scan.

```suggestion
  console.error('Generator failed:', err && err.message);
  process.exit(1);
```

alex-ju (approved):
Looks like a sensible CI check – left only one suggestion for consideration

PR #4075, opened 2026-08-20
MCP - Website catalog

zamoore:
### :pushpin: Summary

If merged, this PR adds a build script to generate a JSON catalog of website documentation and refactors shared normalization utilities.

- Created `build-docs-catalog` script in `packages/mcp/scripts/` that processes markdown files from the website docs directory and generates a structured JSON catalog (`docscatalog.json`)                                         
- The script extracts frontmatter metadata, parses markdown content into searchable chunks, validates cross-references, and joins the component catalog
- Refactored `normalizeLookupValue` function to a shared utility (`src/shared/normalize.ts`) to eliminate duplication across components, flight-icons, and tokens stores
- Added comprehensive test coverage for the new shared normalize utility                                                                
- Added yaml dependency to parse frontmatter blocks in documentation files

### :link: External links

Jira ticket: [HDS-6535](https://hashicorp.atlassian.net/browse/HDS-6535)

[HDS-6535]: https://hashicorp.atlassian.net/browse/HDS-6535?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/mcp/scripts/build-docs-catalog/index.mjs
dchyun, line 1: [Suggestion] Might be worth excluding the version histories and changelog entries from the catalog to reduce noise. (comment brought over from #4007)

packages/mcp/scripts/build-docs-catalog/chunks.mjs
```diff
@@ -0,0 +1,87 @@
...
+
+  return chunks.map(({ level, heading, lines }) => ({
+    level,
+    heading,
+    body: lines.join("\n").trim(),
+  }));
+}
+
+// walks back through shallower headings to build the breadcrumb
+export function headingTrailFor(chunks, index) {
```
dchyun, line 69: [Suggestion] The `headingTrail` looks really useful for keeping context. I wonder if it would help an AI more to have the plain text values in the array like you do now or the catalog item ids. Might be something worth testing to see which works better.

packages/mcp/scripts/build-docs-catalog/search-text.mjs
```diff
@@ -0,0 +1,100 @@
...
+];
+
+const HTML_ENTITIES = new Map([
+  ["&amp;", "&"],
+  ["&lt;", "<"],
+  ["&gt;", ">"],
+  ["&quot;", '"'],
+]);
+
+function harvestSymbols(code, into) {
```
dchyun, line 49: [Suggestion] It could be useful keep this set of symbols, or other similar import related ones, out since they appear in every gts snippet and probably don't provide any useful context.

```
"symbols": [
  "@ember",
  "@hashicorp",
```

packages/mcp/tests/unit/shared/normalize.test.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Copyright IBM Corp. 2021, 2026
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+import { describe, expect, it } from "vitest";
+import { normalizeLookupValue } from "../../../src/shared/normalize.js";
+
+describe("normalizeLookupValue", () => {
```
dchyun, line 9: [Question] Previously you'd tested for the empty value on this function. Is it still useful to test for that?

```
it("returns empty string for empty input", () => {
    expect(normalizeLookupValue("")).toBe("");
  });
```

dchyun (commented):
I think all of my [feedback from the previous version](https://github.com/hashicorp/design-system/pull/4007#pullrequestreview-4900728936) of this work has been addressed or isn't relevant but could you double check everything is addressed? Same applies for #4075

packages/mcp/scripts/build-docs-catalog/index.mjs
zamoore: Makes sense to me

packages/mcp/scripts/build-docs-catalog/chunks.mjs
zamoore: The plain text form is better for the LLM use case. The heading trail is there to give reading context, and natural language headings like `["Code", "How to use this component", "Full width"]` are immediately readable. The slugified/ID form is already embedded in the chunk's `id` field anyway, so there's no information lost by keeping the trail human-readable.

packages/mcp/scripts/build-docs-catalog/search-text.mjs
zamoore: Good call

packages/mcp/tests/unit/shared/normalize.test.ts
zamoore: No, I don't think it is anymore

packages/mcp/scripts/build-docs-catalog/index.mjs
zamoore: I'm actually going to keep the version history and changelog entries. They're downranked by default, so it shouldnt actually show up unless someone is looking for them.

shleewhite (approved):
LGTM - the catalog is missing the component page content, but it is addressed in https://github.com/hashicorp/design-system/pull/4084

PR #2274, opened 2024-07-24
`ApplicationState` Enhancements - Implementation

zamoore:
### :pushpin: Summary

If merged, this PR updates the design and functionality of the `ApplicationState` component.

### :hammer_and_wrench: Detailed description

- Adds a yielded `Media` child component that takes no arguments. Yields to any content.
- Moves the error code above the title/icon line
- Changes the vertical spacing between the footer and body components
- Adds a new `align` (`left` or `center`) argument that changes the orientation of content
- Sets a max width of content inside the component (other than the Media block) of `480px`.
- Automatically orients the direct children of the footer element as follows
  - The first two children will display on the left side of the footer automatically.
  - Any other children will be displayed on the right.
- Removes the visual separator above the footer
- The footer now yields `Dropdown` and `Button` as actions

### :camera_flash: Screenshots

New default layout for existing `ApplicationState` implementations without media
Screenshot 2024-07-24 at 1 14 04 PM

`ApplicationState` without media, new alignment options
Screenshot 2024-07-24 at 1 14 42 PM

`ApplicationState` with media, new alignment options
Screenshot 2024-07-24 at 1 15 01 PM

`ApplicationState` with extra-wide media, new alignment options
Screenshot 2024-07-24 at 1 15 17 PM

### :link: External links

Jira ticket: [HDS-3542](https://hashicorp.atlassian.net/browse/HDS-3542)
Figma file: [Figma](https://www.figma.com/design/noyY6dUMDYjmySpHcMjhkN/branch/PHekpm994tyTcf1fIY2Pgn/HDS-Components?node-id=31113-50004&t=ats1qMOpRLUN9RS9-0)

***

### 👀 Component checklist

- [x] Percy was checked for any visual regression
- [x] A changelog entry was added via [Changesets](https://github.com/changesets/changesets) if needed (see [templates here](https://github.com/hashicorp/design-system/blob/main/wiki/Website-Changelog.md#templates-for-npm-packages))

:speech_balloon: Please consider using [conventional comments](https://conventionalcomments.org/) when reviewing this PR.

[HDS-3542]: https://hashicorp.atlassian.net/browse/HDS-3542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4081, opened 2026-08-24
Exploration: Identifying inconsistencies in component API docs

annawanggg:
> [!IMPORTANT]
> This PR is just for reviewing the docs attached, do not merge!! 

### :pushpin: Summary

This PR aims to identify any consistencies that may exist between the type definitions of component arguments and their corresponding API documentation. Two documents are attached: `initial-prompt.md` and `final-analysis.md`. 
- `initial-prompt.md` includes AI instructions to audit each HDS component and its contextual components, specifically looking at: argument names, required flags, the `@values` array, default values, and types. 
     - Potential source code issues were also noted.
- `final-analysis.md` presents all of the raised issues by component, with a brief description of the problem and where to look in the source code and component API docs. 
     - Summary statistics on the most common issue types are also noted at the top. 
     - The contents of this file are not Bob's original output, as a second audit was performed afterward to remove any spuriously flagged issues and clean up the document.

Feedback would be appreciated to check if there is any criteria I may have missed or need to edit in `initial-prompt.md` and to make sure the raised issues in `final-analysis.md` are valid. Each issue was manually checked over once already, but some might be incorrectly flagged still based on context and implementation detail that I am missing. I've left in-line comments calling out some issues I wanted a second opinion on, especially for the potential source code bugs that were flagged 👍

### :link: External links

Jira ticket: [HDS-6584](https://hashicorp.atlassian.net/browse/HDS-6584)

***

<details>

</details>

[HDS-6584]: https://hashicorp.atlassian.net/browse/HDS-6584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/component-api-docs-audit/final-analysis.md
+++ b/component-api-docs-audit/final-analysis.md
@@ -1,10 +1,10 @@
 # Component API Audit Summary
 
-**Total issues: 34** (+ 8 potential source bugs noted separately, + 2 minor docs formatting inconsistencies noted but excluded from count)
+**Total issues: 37**
 
 | Issue type                          | Count |
 |-------------------------------------|-------|
-| Wrong default                       | 6     |
+| Wrong default                       | 9     |
 | Wrong type                          | 7     |
 | Missing arg                         | 8     |
 | Wrong required flag                 | 4     |
@@ -14,10 +14,12 @@
 | Wrong values                        | 2     |
 | Wrong arg name                      | 1     |
 | Wrong content                       | 1     |
-| Potential source bug                | 8     |
-| Minor docs formatting inconsistency | 2     |
 
-**Components with issues: 27 / 66 audited**
+Noted separately from total count:
+- 8 potential source bugs
+- 2 minor docs formatting inconsistencies
+
+**Components with issues: 28 / 66 audited**
 
 ---
 
@@ -32,13 +34,15 @@ Docs: `website/docs/components/accordion/partials/code/component-api.md`
 ### [A].Item
 
 - **`ariaLabel`: default specified but none exists**
-  - Description: The docs claim a hardcoded string default that does not exist in source; the actual fallback is a completely different accessibility mechanism.
+  - Description: The docs claim a hardcoded string default of `Toggle display` that does not exist in source. The actual fallback is a different accessibility mechanism.
   - Source (`item/index.gts:42`):
     - `ariaLabel?: string`, optional, no default value
   - Source (`item/index.gts:68-71`):
     - When `ariaLabel` is absent, `ariaLabelledBy` is set to `_titleId`, pointing to the element wrapping the toggle block content; the accessible name is derived from whatever the consumer puts in the toggle block
   - Docs (`component-api.md:50`):
-    - `@default='"Toggle display"'`, the string `"Toggle display"` does not exist anywhere in the source
+    - `@default='"Toggle display"'`
+  - Docs (`how-to-use.md:45`):
+    - `The default value is "Toggle display" but you can set a custom value useful for translated text for example.`
 
 ---
 
@@ -62,6 +66,20 @@ Docs: `website/docs/components/table/advanced-table/partials/code/component-api.
     - `model: T[]`, no `?`, required
   - Docs (`component-api.md:39`):
     - Not marked `@required`
+
+- **`reorderedMessageText`: wrong default**
+  - Description: The docs default omits the word "position" that appears in the actual default string.
+  - Source (`index.gts:753`):
+    - Default: `` `Moved ${column.label} column to position ${newPosition}` ``
+  - Docs (`component-api.md:101`):
+    - `@default="Moved (label) column to (position)"` — missing "position" before the placeholder
+
+- **`sortedMessageText`: wrong default**
+  - Description: The docs default includes a comma after `(label)` that is not present in the actual default string.
+  - Source (`index.gts:554`):
+    - Default: `` `Sorted by ${this.currentSortBy} ${this.currentSortOrder}ending` ``
+  - Docs (`component-api.md:174`):
+    - `@default="Sorted by (label), (asc/desc)ending"` — spurious comma after `(label)`
 
 ### AdvancedTable::Th
 
@@ -368,6 +386,22 @@ Docs: `website/docs/components/rich-tooltip/partials/code/component-api.md`
 
 ---
 
+## Table
+
+Source: `packages/components/src/components/hds/table/index.gts`, `th-sort.gts`, `types.ts`
+Docs: `website/docs/components/table/table/partials/code/component-api.md`
+
+### Table (root)
+
+- **`sortedMessageText`: wrong default**
+  - Description: The docs default includes a comma after `(label)` that is not present in the actual default string.
+  - Source (`index.gts:181`):
+    - Default: `` `Sorted by ${this.sortBy} ${lowerCaseTranslatedSortOrder}` ``
+  - Docs (`component-api.md:112`):
+    - `@default="Sorted by (label), (asc/desc)ending"` — spurious comma after `(label)`
+
+---
+
 ## Time
 
 Source: `packages/components/src/components/hds/time/index.gts`, `single.gts`, `range.gts`
@@ -419,7 +453,7 @@ Docs: `website/docs/components/form/layout/partials/code/component-api.md`
   - Docs (`component-api.md:42`):
     - `@valueNote="any valid CSS width (px, rem, etc)" @default="672px"`, `@default` is dead markup here; `672px` never appears on the rendered page
   - Note:
-    - Have it match the pattern used for documenting `Dropdown` `@width`.
+    - Should maybe match pattern used for documenting `Dropdown` `@width`
 
 ### Form::Header::Title
 
@@ -454,6 +488,8 @@ Docs: `website/docs/components/form/primitives/partials/code/component-api.md`
     - `id` absent from `Args`
   - Docs (`component-api.md:176`):
     - Correctly documented as `<C.Property @name="id" @type="string">`
+  - Note:
+    - Should match the pattern of `Form::Field` with `id` in `Args`
 
 ---
 
--- a/component-api-docs-audit/final-analysis.md
+++ b/component-api-docs-audit/final-analysis.md
@@ -1,0 +1,259 @@
+# Component API Audit — Summary
+
+> This summary covers components audited so far. It will be updated as more components are reviewed.
+**Total issues: 19** (+ 2 potential source bugs noted separately)
+
+| Issue type                          | Count |
+|-------------------------------------|-------|
+| Wrong default                       | 3     |
+| Wrong type                          | 6     |
+| Missing arg                         | 2     |
+| Wrong required flag                 | 3     |
+| Default specified but none exists   | 1     |
+| Undocumented default                | 1     |
+| Wrong values                        | 1     |
+| Wrong arg name                      | 1     |
+| Wrong content                       | 1     |
+| Potential source bug                | 2     |
+
+**Components with issues: 7 / 9 audited**
+
+**Flagged for review: 2** (`density`, `isSelected`)
+
+---
+
+
+# Component API Audit — Findings
+
+---
+
+## Accordion
+
+Source: `packages/components/src/components/hds/accordion/index.gts`, `item/index.gts`, `types.ts`
+Docs: `website/docs/components/accordion/partials/code/component-api.md`
+
+### [A].Item
+
+- **`ariaLabel`: default specified but none exists**
+  - Description: The docs claim a hardcoded string default that does not exist in source; the actual fallback is a completely different accessibility mechanism.
+  - Source (`item/index.gts:42`):
+    - `ariaLabel?: string`, optional, no default value
+  - Source (`item/index.gts:68-71`):
+    - When `ariaLabel` is absent, `ariaLabelledBy` is set to `_titleId`, pointing to the element wrapping the toggle block content; the accessible name is derived from whatever the consumer puts in the toggle block
+  - Docs (`component-api.md:50`):
+    - `@default='"Toggle display"'`, the string `"Toggle display"` does not exist anywhere in the source
+
+---
+
+## AdvancedTable
+
+Source: `packages/components/src/components/hds/advanced-table/index.gts`, `tr.gts`, `th.gts`, `td.gts`, `types.ts`
+Docs: `website/docs/components/table/advanced-table/partials/code/component-api.md`
+
+### AdvancedTable (root)
+
+- **`columns`: wrong required flag**
+  - Description: `columns` is required in source but not marked required in the docs.
+  - Source (`index.gts:168`):
+    - `columns: HdsAdvancedTableColumn[]`, no `?`, required
+  - Docs (`component-api.md:50`):
+    - Not marked `@required`
+
+- **`model`: wrong required flag**
+  - Description: `model` is required in source but not marked required in the docs.
+  - Source (`index.gts:174`):
+    - `model: T[]`, no `?`, required
+  - Docs (`component-api.md:39`):
+    - Not marked `@required`
+
+- **`selectionAriaLabelSuffix`: missing arg**
+  - Description: This arg exists in source but is not documented on the root component.
+  - Source (`index.gts:176`):
+    - `selectionAriaLabelSuffix?: string`
+  - Docs:
+    - Not documented
+
+- **`align`: missing arg**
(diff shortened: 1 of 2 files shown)


component-api-docs-audit/final-analysis.md
```diff
@@ -0,0 +1,597 @@
...
+    - `columns: HdsAdvancedTableColumn[]`, no `?`, required
+  - Docs (`component-api.md:50`):
+    - Not marked `@required`
+
+- **`model`: wrong required flag**
+  - Description: `model` is required in source but not marked required in the docs.
+  - Source (`index.gts:174`):
+    - `model: T[]`, no `?`, required
+  - Docs (`component-api.md:39`):
+    - Not marked `@required`
```
annawanggg: I was unsure if columns and model args were intentionally not marked as required bc they are so essential to even a basic advanced table working, but since they are still args + required in the type definition, it may be good to explicitly mark them as required on the website.

component-api-docs-audit/final-analysis.md
```diff
@@ -0,0 +1,597 @@
...
+    - `colspan?: number`
+  - Docs (`component-api.md:264`):
+    - `@type="string"`
+
+- **`rowspan`: wrong type**
+  - Description: Source defines `rowspan` as `number` but docs document it as `string`.
+  - Source (`td.gts:32`):
+    - `rowspan?: number`
+  - Docs (`component-api.md:261`):
+    - `@type="string"`
```
annawanggg, line 114: The getters for `colspan` and `rowspan` in AdvancedTable::Th and AdvancedTable::Td do have `string` as the return type. However, since the component API docs are meant to align with the type definition specifically, it might be better to document them as `number` on the website.

component-api-docs-audit/final-analysis.md
```diff
@@ -0,0 +1,597 @@
...
+    - `iconPosition?: HdsLinkIconPositions`, optional, no default; passed as `@iconPosition={{@iconPosition}}`
+  - Docs (`component-api.md:130`):
+    - `@default="trailing"`
+
+- **`color`: potential source bug**
+  - Description: `color` is declared as an optional arg in the signature but is hardcoded to `"secondary"` in the template and never forwarded, so the arg has no effect. Should it be removed from the `Args` interface?
+  - Source (`link.gts:19`, `link.gts:34`):
+    - `color?: HdsLinkColors` in `Args`; template hardcodes `@color="secondary"` and does not pass `@color={{@color}}`
+  - Docs:
+    - Correctly not documented
```
annawanggg, line 171: Was wondering if this was intentional, should it be cleaned up or left as is?

component-api-docs-audit/final-analysis.md
```diff
@@ -0,0 +1,597 @@
...
+Docs: `website/docs/components/application-state/partials/code/component-api.md`
+
+### [A].Footer
+
+- **`hasDivider`: potential source bug**
+  - Description: `hasDivider` is declared as an optional arg in the signature but is never referenced in the template, so it has no effect. Should it be removed from the `Args` interface?
+  - Source (`footer.gts:15`):
+    - `hasDivider?: boolean` in `Args`; template does not reference it
+  - Docs:
+    - Correctly not documented
```
annawanggg, line 228: Similarly wanted to call this one out and ask if issues like these should be cleaned up or left as is?

component-api-docs-audit/final-analysis.md
```diff
@@ -0,0 +1,597 @@
...
+
+Source: `packages/components/src/components/hds/filter-bar/index.gts`, `types.ts`
+Docs: `website/docs/components/filter-bar/partials/code/component-api.md`
+
+### Filter types
+
+- **`type`: minor docs formatting inconsistency**
+  - Description: The `@required` attribute uses a string value `"true"` instead of the boolean `{{true}}`, which is inconsistent with the rest of the docs. This is a docs template authoring quirk, not a source/docs mismatch; excluded from the issue count.
+  - Docs (`component-api.md:45`):
+    - `@required="true"` (string), should be `@required={{true}}`
```
annawanggg: sorry REALLY minor, but noted this in case

component-api-docs-audit/final-analysis.md
```diff
@@ -0,0 +1,597 @@
...
+
+- **`size`: wrong type**
+  - Description: `size` is documented as `@type="number"` but the source type `HdsIconSizes` resolves to the string union `'16' | '24'`; consumers pass a string, not a number.
+  - Source (`index.gts:24`, `types.ts:6-11`):
+    - `size?: HdsIconSizes`, which is `` `${HdsIconSizeValues}` `` resolving to `'16' | '24'`
+  - Docs (`component-api.md:7`):
+    - `@type="number"`
+  -Note:
+    - Passing a number works at runtime due to Handlebars coercion, but since the TypeScript type is '16' | '24', Glint would reject the number if enabled.
```
annawanggg, line 313: Calling out the note I left, but although passing a number does technically work, should the documented type be changed to `string` ideally?

component-api-docs-audit/final-analysis.md
```diff
@@ -0,0 +1,597 @@
...
+### Time (root)
+
+- **`isoUtcString`: potential source bug**
+  - Description: `isoUtcString` is declared in `Args` but the getter `get isoUtcString()` always derives the value from `this.date` via the time service and never reads `this.args.isoUtcString`. The argument is appropriately omitted from the docs currently since it isn't used currently. Should it be removed from the `Args` interface?
+  - Source (`index.gts:32`):
+    - `isoUtcString?: string` declared in `Args`
+  - Source (`index.gts:95-105`):
+    - `get isoUtcString()` ignores `this.args.isoUtcString` entirely
+  - Docs:
+    - Correctly not documented
```
annawanggg, line 419: This doesn't impact the component API docs at all, but again wondering if it should be cleaned up or left as is!

component-api-docs-audit/final-analysis.md
```diff
@@ -0,0 +1,597 @@
...
+    - `@name="href"` with no `@default` specified
+
+### [AF].Link
+
+- **`iconPosition`: wrong default**
+  - Description: The docs document a default of `"trailing"` but the source does not set any default; `iconPosition` is passed directly to `HdsLinkInline` without a fallback.
+  - Source (`link.gts:21`, `link.gts:44`):
+    - `iconPosition?: HdsLinkIconPositions`, optional, no default; passed as `@iconPosition={{@iconPosition}}`
+  - Docs (`component-api.md:130`):
+    - `@default="trailing"`
```
annawanggg, line 164: @dchyun found that the default of `trailing` is actually set within `HdsLinkInline` so this isn't actually an inconsistency!

component-api-docs-audit/final-analysis.md
```diff
@@ -0,0 +1,597 @@
...
+    - `if (this.args.visibilityToggleAriaLabel) { return this.args.visibilityToggleAriaLabel; }`, no mechanism to pass state-dependent overrides
+  - Docs:
+    - Not a docs issue
+
+- **`visibilityToggleAriaMessageText`: potential source bug**
+  - Description: Same limitation as `visibilityToggleAriaLabel`: only a single static string can be provided, with no way to supply different text for the masked and unmasked states.
+  - Source (`base.gts:103-121`):
+    - `if (this.args.visibilityToggleAriaMessageText) { return this.args.visibilityToggleAriaMessageText; }`, no mechanism to pass state-dependent overrides
+  - Docs:
+    - Not a docs issue
```
annawanggg, line 540: My other question for `Form::MaskedInput::Base` and also `Form::TextInput::Field` down below:

Should both "Show masked content" and "Hide masked content" be shown in the defaults on the docs? It only mentions "Show masked content" currently., but was unsure if we intentionally only want to show the initial default state. (Same goes for "Input content is hidden" and "Input content is visible")

component-api-docs-audit/final-analysis.md
shleewhite: yes, a number is correct. its a string for the getters bc we have to translate the number (which is how normal table cells set colspan/rowspan) to  a valid value for CSS grid which is used to style the advanced table.

component-api-docs-audit/final-analysis.md
shleewhite: The argument should probably be removed, this was likely an oversight when it was converted to typescript

component-api-docs-audit/final-analysis.md
shleewhite: Maybe @zamoore or @jorytindall remembers bc I think they were the ones who worked on this component - but this feels like something that accidentally got left in after the design changed

component-api-docs-audit/final-analysis.md
shleewhite: yeah, honestly the type should be documented as "16" | "24"

component-api-docs-audit/final-analysis.md
```diff
@@ -0,0 +1,633 @@
...
+- **`hiddenText`: missing arg**
+  - Description: `Form::Label` accepts a `hiddenText` arg that renders visually hidden accessible text, but it is not documented.
+  - Source (`label/index.gts:20`):
+    - `hiddenText?: string`, renders a `<span class="sr-only">` when present
+  - Docs (`component-api.md:3-21`):
+    - Not documented
+
+### Form::Fieldset
+
+- **`id`: potential source bug**
```
shleewhite, line 485: This is accurate, we should fix.

component-api-docs-audit/final-analysis.md
```diff
@@ -0,0 +1,633 @@
...
+
+## Form::MaskedInput
+
+Source: `packages/components/src/components/hds/form/masked-input/base.gts`, `field.gts`
+Docs: `website/docs/components/form/masked-input/partials/code/component-api.md`
+
+### Form::MaskedInput::Base
+
+- **`visibilityToggleAriaLabel`: potential source bug**
+  - Description: The arg accepts only a single static string, so a consumer cannot provide different labels for the masked and unmasked states. Passing any value means the same string is used regardless of state.
```
shleewhite, line 529: Hmm... @MelSumner can weigh in here, but im of the opinion that the label should never change based on the state, aria-pressed (or some other attribute) should be used to convey the state. This is a larger thing tho, probably would need its own ticket.

component-api-docs-audit/final-analysis.md
shleewhite: We can leave it as is, but add to the description that by default it toggles between the 2 depending on the visibility state?

component-api-docs-audit/initial-prompt.md
```diff
@@ -0,0 +1,95 @@
+# Component API Documentation Audit
+
+## Overview
+
+The goal is to audit every HDS component and check the type definitions of component arguments in `packages/components/src/components/hds/` with its corresponding API documentation in `website/docs/components/`. The goal is to identify any inconsistencies in the documentation so they can be corrected to more accurately reflect our components to better serve our consumers and ongoing initiatives like the MCP server.
```
KristinLBradley:
```suggestion
The objective is to audit all HDS components in `packages/components/src/components/hds/` and verify that their argument type definitions match the API documentation in `website/docs/components/`. The results of the audit will be used to identify and correct any inconsistencies in the documentation.
```

How about something more like the above? I attempted to simplify and clarify the text.

component-api-docs-audit/final-analysis.md
annawanggg: gotcha, so rather than the existing `Show/Hide masked content` label that changes, it should stick with the single customizable label like the `Show masked content` default in the docs, and then maybe set `aria-pressed` to true and false (with the doc description clarifying this)?

also as a general question for learning purposes, is the `toggle` component implemented using a switch that uses `aria-checked` because its a toggle controlling a more persistent state, while is the toggle for the masked input a button instead bc it's a temporary action?

component-api-docs-audit/initial-prompt.md
annawanggg: this is great thanks kristin!

component-api-docs-audit/final-analysis.md
shleewhite: its a subtle difference, but a switch is a type of checkbox so it is a form element, whereas a toggle button (aka a button with aria pressed) is not.

the masked input toggle is not a part of the form, it just changes the visibility of the form content so toggle button is more appropriate.

shleewhite: its a subtle difference, but a switch is a type of checkbox so it is a form element, whereas a toggle button (aka a button with aria pressed) is not.

the masked input toggle is not a part of the form, it just changes the visibility of the form content so toggle button is more appropriate.

annawanggg: ooo gotcha, tyty!

component-api-docs-audit/final-analysis.md
dchyun: I think it can be removed. It is used in a few showcase examples, but isn't have any effect and I don't see it used in the component template.

```
<HdsApplicationState as |A|>
  <A.Header @title="Empty state title" />
  <A.Body
    @text="Sorry, an unexpected error has occurred. Please try again later or contact support for assistance."
  />
  <A.Footer @hasDivider={{true}} as |F|>
    <F.LinkStandalone
      @icon="arrow-left"
      @text="Go to previous page"
      @href="/"
    />
    <F.LinkStandalone
      @icon="help"
      @text="Need Help"
      @href="/components/alert"
      @iconPosition="trailing"
    />
  </A.Footer>
</HdsApplicationState>
```

Image

component-api-docs-audit/final-analysis.md
dchyun: @KristinLBradley this catch looks valid as `@isoUtcString` isn't used in the `HdsTime`. There is an argument with the same name on the `HdsTimeSingle` maybe it's only needed there and can be removed from the `HdsTime` signature?

component-api-docs-audit/final-analysis.md
MelSumner: I agree with @shleewhite that the label shouldn't change. It could say "Toggle masked content" and a screen-reader will announce if it's pressed or not.

shleewhite: @annawanggg do you mind creating a ticket in the backlog for this?

annawanggg: yep creating tickets for everything rn!

component-api-docs-audit/final-analysis.md
KristinLBradley: Here's the PR in which the divider was removed following design updates made to the ApplicationState component: https://github.com/hashicorp/design-system/pull/2274

So it seems the arg was just left by accident and should be cleaned up to remove it.

component-api-docs-audit/final-analysis.md
KristinLBradley: Investigating...

KristinLBradley: It can be safely removed. It looks like this it was leftover during part of the [original development process](https://github.com/hashicorp/design-system/pull/2515) when I refactored the Time component implementation to create the `TimeSingle` component.

annawanggg: sounds good, thanks guys!

component-api-docs-audit/final-analysis.md
```diff
@@ -0,0 +1,633 @@
...
+
+## Form::KeyValueInputs
+
+Source: `packages/components/src/components/hds/form/key-value-inputs/index.gts`, `field.gts`, `add-row-button.gts`, `delete-row-button.gts`, `generic.gts`
+Docs: `website/docs/components/form/key-value-inputs/partials/code/component-api.md`
+
+### [F].AddRowButton
+
+- **`ariaLabel`: missing arg**
+  - Description: `HdsFormKeyValueInputsAddRowButton` accepts an `ariaLabel` arg but it is not documented.
```
annawanggg, line 504: @shleewhite realizing that the `ariaLabel` arg is never actually used anywhere in the `AddRowButton `component. consumers are able to pass in `aria-label` fine to the `HdsButton` via splattributes, but do we want `ariaLabel` to be added to the website docs and have the code explicitly pass `aria-label={{@ariaLabel}}` with a fallback value for `ariaLabel`, or to keep things as is and remove it from the component args?

shleewhite: ahh good catch! this arg is probably never used... but the safest option is to explicitly set `aria-label={{@ariaLabel}}`.

To remove the arg could potentially be a breaking change.

PR #4109, opened 2026-09-03
MCP: Add search tools for icons and tokens 

zamoore:
### :pushpin: Summary

Adds two MCP search tools, `search_hds_icons` and `search_hds_tokens`, and wires them into the server alongside the existing components and docs tools.

Also renames `flight-icons` → `hds-icons` throughout the package so all five tools read `*_hds_*`.

**New tools**

- `search_hds_tokens` — finds tokens by name *or* value (searching a hex code finds the token that defines it) and returns the `cssVar` to write. Filters by type and category.
- `search_hds_icons` — finds icons by name or keyword, filtered by category.

Both return bounded payloads with `totalMatches` / `truncated`, so a caller knows when to narrow rather than assuming it saw the best matches. An unknown filter value comes back in `unknownFilters` alongside the values that do exist, instead of an empty result with no explanation.

**Store support**

The icons and tokens stores now back search directly:

- `searchIcons` / `searchTokens` return `CatalogSearchOutcome` (`{ totalMatches, hits }`) rather than a bare array, so the total survives the limit.
- `getMeta()` exposes `categories`, used to answer unknown filters.
- The token search blob now includes `cssVar`, so a token is findable by the variable name a consumer would actually type.

**Shared constants**

`DEFAULT_SEARCH_LIMIT`, `MAX_SEARCH_LIMIT` and `MAX_QUERY_LENGTH` moved from `tools/components/constants.ts` to a new `tools/shared/constants.ts`, and `tools/components/utils.ts` became `tools/shared/output-schema.ts` — both are now used by all search tools.

**Rename**

`flight-icons` → `hds-icons` across tools, stores, resources, and tests, plus `search_flight_icons` → `search_hds_icons`. The `@hashicorp/flight-icons` package specifier and the `"Flight Icons"` docs tab label are unchanged — those are the real package and real catalog data, not naming. Resource URIs (`hds://icons`) are unchanged.

**Testing**

Integration tests for both new tools. `tests/unit/tools/index.test.ts` pins the registered tool-name set: a tool whose barrel is empty or missing from `src/tools/index.ts` now fails there, rather than passing its own tests while being unreachable by clients.

Also removes `tests/integration/tools/docs.test.ts`, which duplicated `read-hds-docs.test.ts` except for one comment, and renames `search-components.test.ts` → `search-hds-components.test.ts` to match the tool it covers.

### :link: External links

Jira ticket: [HDS-6643](https://hashicorp.atlassian.net/browse/HDS-6643)
***

<details>

</details>

[HDS-6643]: https://hashicorp.atlassian.net/browse/HDS-6643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/mcp/src/tools/hds-icons/search-icons.ts
```diff
@@ -0,0 +1,189 @@
...
+  limit: number;
+  category?: string;
+  size?: string;
+  hasMapping?: boolean;
+}
+
+export type SearchIconsPayload = z.infer<typeof searchIconsOutputSchema>;
+
+const DESCRIPTION = [
+  "Find Flight icons — the icon set Helios ships — by name or keyword, from the catalog inside the installed @hashicorp/flight-icons package.",
```
dchyun:
```suggestion
  "Find HDS icons — the icon set Helios ships — by name or keyword, from the catalog inside the installed @hashicorp/flight-icons package.",
```
[Nit] Rename to "HDS" to align with other content

packages/mcp/src/tools/hds-icons/search-icons.ts
```diff
@@ -0,0 +1,189 @@
...
+      toJsonToolResponse(searchIcons(getStore(), input)),
+    );
+
+  return defineTool<
+    typeof searchIconsInputShape,
+    typeof searchIconsOutputShape
+  >({
+    name: SEARCH_ICONS_TOOL_NAME,
+    config: {
+      title: "Search Flight icons",
```
dchyun:
```suggestion
      title: "Search HDS icons",
```
[Nit] Rename to "HDS" to align with other content

packages/mcp/src/stores/hds-icons/index.ts
dchyun, line 1: [Question] I think this is just an mcp inspector issue but wanted to raise it to be sure. When using the inspector if I type a value into the `category` and execute a search it works as expected. If I then delete the category value and execute another search it keeps returning an error.

Image

packages/mcp/src/tools/constants.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Copyright IBM Corp. 2021, 2026
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+export const DEFAULT_SEARCH_LIMIT = 20;
```
dchyun, line 6: [Suggestion] Would we want to align the docs tool to also use these `DEFAULT_SEARCH_LIMIT` and `MAX_SEARCH_LIMIT` values instead of its own?

```
export const DEFAULT_SEARCH_LIMIT = 8;
// cap keeps a search from ever approaching limits
export const MAX_SEARCH_LIMIT = 25;
```

packages/mcp/tests/integration/tools/search-hds-icons.test.ts
```diff
@@ -0,0 +1,239 @@
...
+    const payload = search({ query: "spaceship" });
+
+    expect(payload.results).toStrictEqual([]);
+    expect(payload.totalMatches).toBe(0);
+    expect(payload.truncated).toBe(false);
+    expect(payload.totalIconCount).toBe(3);
+  });
+
+  it("narrows by category and mapping, echoing what it applied", () => {
+    const payload = search({ query: "arrow", category: " NAVIGATION " });
```
dchyun, line 118: [Question] Not sure if this is a inspector issue but if I filter on a category with whitespace like `  Navigation  ` it returns no results.

packages/mcp/tests/integration/tools/search-hds-icons.test.ts
```diff
@@ -0,0 +1,239 @@
...
+
+    expect(payload.unknownFilters).toStrictEqual(["category: Spaceships"]);
+    expect(payload.availableCategories).toStrictEqual(["Alerts", "Navigation"]);
+    expect(payload.results).toStrictEqual([]);
+  });
+
+  it("leaves the valid-value list out when the filter landed", () => {
+    expect(
+      search({ query: "arrow", category: "Navigation" }),
+    ).not.toHaveProperty("availableCategories");
```
dchyun, line 155: [Suggestion] A similar test should be added for `availableSizes` not being included when the filter lands

packages/mcp/PR_DESCRIPTION.md
dchyun, line 1: [todo] I think this file can be removed.

packages/mcp/src/tools/tokens/search-tokens.ts
```diff
@@ -0,0 +1,202 @@
...
+    .string()
+    .min(1)
+    .max(MAX_FILTER_LENGTH)
+    .optional()
+    .describe(
+      "Restrict to one token category, e.g. 'color', 'border' or 'typography'. A value the catalog does not use comes back in unknownFilters alongside the categories that exist.",
+    ),
+};
+
+const searchTokensResultShape = z.object({
```
dchyun, line 60: [Suggestion] It'd be useful to provide some additional token information which is present in the docs json.

- Any comments contained in the `comment` value
- Original alias values for semantic tokens contained in the `original.value` value

It could be helpful to look at this [revamp of the website tokens catalog](https://github.com/hashicorp/design-system/pull/4113) done for the carbonization in which a lot of these values are now surfaced in the docs.

packages/mcp/src/tools/tokens/search-tokens.ts
```diff
@@ -0,0 +1,202 @@
...
+  limit: number;
+  type?: TokenType;
+  category?: string;
+}
+
+export type SearchTokensPayload = z.infer<typeof searchTokensOutputSchema>;
+
+const DESCRIPTION = [
+  "Find Helios design tokens by name or value, from the token catalog inside the installed @hashicorp/design-system-tokens package.",
+  "Every result carries the cssVar to write in a stylesheet — var(--token-color-foreground-action) — so use this to replace a hard-coded colour, spacing or radius with the token that holds it.",
```
dchyun:
```suggestion
  "Every result carries the cssVar to write in a stylesheet — var(--token-color-foreground-action) — so use this to replace a hard-coded color, spacing or radius with the token that holds it.",
```
[super nit] Don't use british english 🦅 🇺🇸

packages/mcp/src/tools/components/get-component.ts
dchyun, line 1: [Question] I don't see this new tool mentioned in the PR description. Could you elaborate on its purpose?

I see some conflicting use cases potentially between this and the `search_hds_components` tool. The search tool mentioned that is intended for search only to see if components exist, and not to provide api information. 

In the description it directs you to the `search_hds_docs` tool for getting api information
> A result's docsPath is what search_hds_docs takes as its docsPath filter, so that is the way through to what a component's arguments are and how to use it

But it looks like this `get_hds_components` tool provides that api information. So there's a bit of a conflict that the search is pointed to the docs tool for api information when this tool provides that. 

 Would it make sense to include this api information directly within the search tool results, like is done for `search_hds_components` or `search_hds_tokens`? Or is there a way to direct the search to use this tool.

packages/mcp/src/tools/hds-icons/search-icons.ts
zamoore: Updated this to "HDS icons" for consistency with the tool name and the rest of the MCP surface. The description still identifies @hashicorp/flight-icons as the catalog source.

packages/mcp/src/stores/hds-icons/index.ts
zamoore: I confirmed that the server accepts the category when it is omitted. The inspector appears to retain the cleared field as `category: ""`, which is correctly rejected by the schema because a present category must be non-empty. I’ve kept the schema strict rather than treating malformed empty-string input as an omitted optional property.

packages/mcp/src/tools/constants.ts
zamoore: I’ve kept the docs limits separate because docs results are substantially larger.

packages/mcp/tests/integration/tools/search-hds-icons.test.ts
zamoore: Surrounding whitespace is supported at the server level. Category values are trimmed and lowercased before comparison, and the integration test passes `" NAVIGATION "` and receives the expected results. I wasn’t able to reproduce this outside inspector, so it appears to be inspector state.

packages/mcp/src/tools/tokens/search-tokens.ts
zamoore: Added optional comment and alias fields to token search results. Rather than returning the entire raw original object, `alias`r is derived specifically from unresolved token references such as `{color.palette.neutral-700}`. Both fields are also searchable, and tests cover their extraction and output.

packages/mcp/src/tools/tokens/search-tokens.ts
zamoore: Whoops

packages/mcp/src/tools/components/get-component.ts
zamoore: Good catch. The intended split is discovery versus API versus guidance: `search_hds_components` finds the component, `get_hds_component` returns its generated API contract, and `search_hds_docs` returns prose guidance, examples, and accessibility recommendations.

I updated the search tool description to make that flow explicit. I kept the complete API out of search results because searches may return many components and some argument value sets are very large, so the two-step flow keeps responses focused and bounded.

dchyun (approved):

PR #4118, opened 2026-09-08
`Form::MaskedInput` + `Form::TextInput`: Visibility toggle a11y fixes

annawanggg:
> [!IMPORTANT]
> Do not merge until further notice

### :pushpin: Summary

If merged, this PR would address a11y issues raised during the component API docs audit:

#### `aria-label` should not change with state
Context: The current `visibilityToggleAriaLabel` implementation has the aria-label on the `Form::MaskedInput::Base` component change from “Show masked content” to “Hide masked content.” This also occurs on the `Form::TextInput::Field` component, where the aria-label changes from “Show password” to “Hide password.” The label of a toggle button should never change when its state changes.

Fix: There is now a single default `visibilityToggleAriaLabel` value that consumers can still override, along with a new `aria-pressed` value passed to the  `HdsFormVisibilityToggle` that conveys the pressed state of the toggle button.

#### allowing custom `sr-only` messages for both visible and hidden states
Context: `Form::MaskedInput::Base` and `Form::TextInput::Field` both have a `visibilityToggleAriaMessageText` arg that supplies the `sr-only` message the screenreader announces upon state change. The current implementation has different default messages for the visible and hidden states, but it only allows consumers to pass one custom message currently. 

Fix: To support this, the arg `visibilityToggleAriaMessageTextWhenVisible` was added so consumers can set a custom message for the visible state. ~~As a backwards compatible solution, `visibilityToggleAriaMessageText` still supplies the value for both states if no `visibilityToggleAriaMessageTextWhenVisible` is provided as a guardrail. However, after looking in consumer repos, it seems no one uses the arg currently 🥲, so let me know if I should keep this or not.~~ Based on feedback from review, I got rid of the backwards compatibility and cleaned up the code accordingly. `visibilityToggleAriaMessageText` is only responsible for the hidden state now. There's also a new `ariaMessageTextWhenVisible` arg in `Form::VisibilityToggle`.

I conducted manual accessibility testing and updated the integration tests accordingly. I have a number of tests that are in both `Form::MaskedInput::Base` and `Form::MaskedInput::Field` to be safe, but let me know if they are too redundant. Also, there wasn't any a11y-related testing for `Form::TextInput`, so I added a bunch of tests for the visibility toggle if that looks okay!

Form::MaskedInput
[before](https://helios.hashicorp.design/components/form/masked-input?tab=code#component-api) | [after](http://hds-website-git-annawanggg-form-a11y-fixes-hashicorp.vercel.app/components/form/masked-input?tab=code#component-api)

Form::TextInput
[before](https://helios.hashicorp.design/components/form/text-input?tab=code#component-api) | [after](https://hds-website-git-annawanggg-form-a11y-fixes-hashicorp.vercel.app/components/form/text-input?tab=code#component-api)

### :link: External links

Jira ticket: [HDS-6621](https://hashicorp.atlassian.net/browse/HDS-6621)

***

<details>

</details>

[HDS-6621]: https://hashicorp.atlassian.net/browse/HDS-6621?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/big-plums-sing.md
+++ b/.changeset/big-plums-sing.md
@@ -1,0 +1,16 @@
+---
+"@hashicorp/design-system-components": patch
+---
+
+<!-- START components/form/masked-input -->
+`Form::MaskedInput` - Fixed a few accessibility issues in `Form::MaskedInput::Base` and `Form::MaskedInput::Field`, including:
+- Changed `@visibilityToggleAriaLabel` to have a static value for visible and hidden states. Added the `aria-pressed` attribute to the visibility toggle button to communicate state.
+- Added `@visibilityToggleAriaMessageTextWhenVisible` argument to set a custom message for the visible state.
+- 
+<!-- END -->
+
+<!-- START components/form/text-input -->
+`Form::TextInput` - Fixed a few accessibility issues in `Form::TextInput::Field`, including:
+- Changed `@visibilityToggleAriaLabel` to have a static value for visible and hidden states. Added the `aria-pressed` attribute to the visibility toggle button to communicate state.
+- Added `@visibilityToggleAriaMessageTextWhenVisible` argument to set a custom message for the visible state.
+<!-- END -->
--- a/.changeset/big-plums-sing.md
+++ b/.changeset/big-plums-sing.md
@@ -1,16 +1,19 @@
 ---
-"@hashicorp/design-system-components": patch
+"@hashicorp/design-system-components": minor
 ---
 
 <!-- START components/form/masked-input -->
 `Form::MaskedInput` - Fixed a few accessibility issues in `Form::MaskedInput::Base` and `Form::MaskedInput::Field`, including:
 - Changed `@visibilityToggleAriaLabel` to have a static value for visible and hidden states. Added the `aria-pressed` attribute to the visibility toggle button to communicate state.
 - Added `@visibilityToggleAriaMessageTextWhenVisible` argument to set a custom message for the visible state.
-- 
 <!-- END -->
 
 <!-- START components/form/text-input -->
 `Form::TextInput` - Fixed a few accessibility issues in `Form::TextInput::Field`, including:
 - Changed `@visibilityToggleAriaLabel` to have a static value for visible and hidden states. Added the `aria-pressed` attribute to the visibility toggle button to communicate state.
 - Added `@visibilityToggleAriaMessageTextWhenVisible` argument to set a custom message for the visible state.
+<!-- END -->
+
+<!-- START components/form/visibility-toggle -->
+`Form::VisibilityToggle` - Added `@ariaMessageTextWhenVisible` argument to set a message for the visible state.
 <!-- END -->
--- a/.changeset/big-plums-sing.md
+++ b/.changeset/big-plums-sing.md
@@ -14,6 +14,6 @@
 - Added `@visibilityToggleAriaMessageTextWhenVisible` argument to set a custom message for the visible state.
 <!-- END -->
 
-<!-- START components/form/visibility-toggle -->
+<!-- START components/form/primitives/visibility-toggle -->
 `Form::VisibilityToggle` - Added `@ariaMessageTextWhenVisible` argument to set a message for the visible state.
 <!-- END -->
--- a/packages/components/src/components/hds/form/masked-input/base.gts
+++ b/packages/components/src/components/hds/form/masked-input/base.gts
@@ -88,7 +88,7 @@ export default class HdsFormMaskedInputBase extends Component<HdsFormMaskedInput
       return this.hdsIntl.t(
         'hds.components.form.masked-input.base.show-masked-content',
         {
-          default: 'Show masked content',
+          default: 'Toggle masked content',
         }
       );
     }
--- a/packages/components/src/components/hds/form/masked-input/base.gts
+++ b/packages/components/src/components/hds/form/masked-input/base.gts
@@ -33,7 +33,7 @@ export interface HdsFormMaskedInputBaseSignature {
     value?: string;
     visibilityToggleAriaLabel?: HdsFormVisibilityToggleSignature['Args']['ariaLabel'];
     visibilityToggleAriaMessageText?: HdsFormVisibilityToggleSignature['Args']['ariaMessageText'];
-    visibilityToggleAriaMessageTextWhenVisible?: HdsFormVisibilityToggleSignature['Args']['ariaMessageText'];
+    visibilityToggleAriaMessageTextWhenVisible?: HdsFormVisibilityToggleSignature['Args']['ariaMessageTextWhenVisible'];
     width?: string;
     height?: string;
     ariaDescribedBy?: string;
@@ -86,7 +86,7 @@ export default class HdsFormMaskedInputBase extends Component<HdsFormMaskedInput
       return this.args.visibilityToggleAriaLabel;
     } else {
       return this.hdsIntl.t(
-        'hds.components.form.masked-input.base.show-masked-content',
+        'hds.components.form.masked-input.base.toggle-masked-content',
         {
           default: 'Toggle masked content',
         }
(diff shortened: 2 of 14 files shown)


packages/components/src/components/hds/form/text-input/field.gts
```diff
@@ -57,23 +58,33 @@ export default class HdsFormTextInputField extends Component<HdsFormTextInputFie
...
 
-  get visibilityToggleAriaLabel(): string | undefined {
+  get visibilityToggleAriaLabel(): string {
     if (this.args.visibilityToggleAriaLabel) {
       return this.args.visibilityToggleAriaLabel;
-    } else if (this._isPasswordMasked) {
-      return 'Show password';
     } else {
-      return 'Hide password';
+      return 'Toggle password visibility';
```
annawanggg: `Form::TextInput` doesn't use hdsIntl.t() while Form::MaskedInput does, should I add internationalization here?

packages/components/src/components/hds/form/masked-input/base.gts
```diff
@@ -83,40 +84,45 @@ export default class HdsFormMaskedInputBase extends Component<HdsFormMaskedInput
...
     } else {
-      return this.hdsIntl.t(
-        'hds.components.form.masked-input.base.input-content-is-visible',
-        {
-          default: 'Input content is visible',
-        }
-      );
+      if (this.args.visibilityToggleAriaMessageTextWhenVisible) {
+        return this.args.visibilityToggleAriaMessageTextWhenVisible;
+      } else if (this.args.visibilityToggleAriaMessageText) {
```
annawanggg: I added the `else if` branch for backwards compatibility to account for the current implementation where `@visibilityToggleAriaMessageText` applies to both states... however, no consumers actually use this arg currently :( should I keep these guardrails, or get rid of it + clean the code up?

dchyun: Considering no teams are using the existing argument, and we're not removing the argument I think it's fine to not add this backwards compatibility. I would not consider it a breaking change, more so just the functionality of the argument changing.

packages/components/src/components/hds/form/text-input/field.gts
dchyun: Good catch! Yes let's add internationalization support for this.

packages/components/src/components/hds/form/masked-input/base.gts
```diff
@@ -83,40 +84,45 @@ export default class HdsFormMaskedInputBase extends Component<HdsFormMaskedInput
...
-      return this.hdsIntl.t(
-        'hds.components.form.masked-input.base.show-masked-content',
-        {
-          default: 'Show masked content',
-        }
-      );
     } else {
       return this.hdsIntl.t(
-        'hds.components.form.masked-input.base.hide-masked-content',
+        'hds.components.form.masked-input.base.show-masked-content',
```
dchyun:
```suggestion
        'hds.components.form.masked-input.base.toggle-masked-content',
```
[Suggestion] Since the text is now related to "toggle" instead of "show" I would update the translation yaml to reflect this. This change will need to be made in the translation file as well.

packages/components/translations/hds/components/form/masked-input/base/en-us.yaml
```diff
@@ -1,5 +1,4 @@
 copy-masked-content: Copy masked content
-show-masked-content: Show masked content
-hide-masked-content: Hide masked content
+show-masked-content: Toggle masked content
```
dchyun:
```suggestion
toggle-masked-content: Toggle masked content
```
See previous comment

packages/components/src/components/hds/form/masked-input/base.gts
```diff
@@ -83,40 +84,45 @@ export default class HdsFormMaskedInputBase extends Component<HdsFormMaskedInput
...
         }
       );
     }
   }
 
+  get visibilityToggleAriaPressed(): string {
+    return this.isContentMasked ? 'false' : 'true';
+  }
+
   get visibilityToggleAriaMessageText(): string {
```
dchyun: [Suggestion] What do you think of instead of adding the functionality to decide when to show which aria message text into both the masked input and text input adding that support into the visibility toggle directly?

We could add a `ariaMessageTextWhenVisible` arg to the visibility toggle and then just pass the appropriate arguments in from the masked input and text input.

In the visibility toggle template
```
<span class="sr-only" aria-live="polite">{{if @isVisible @ariaMessageTextWhenVisible @ariaMessageText}}</span>
```

In the masked input template
```
@ariaMessageText={{this.visibilityToggleAriaMessageText}}
@ariaMessageTextWhenVisible={{this.visibilityToggleAriaMessageTextWhenVisible}}
```

website/docs/components/form/masked-input/partials/code/component-api.md
dchyun, line 1: [todo] Please have the docs updates be in a separate PR. When a feature requires docs updates we have it in a separate PR so that we can merge the docs PR only when a release is about to go out.

Any merge to `main` updates the website so we want to avoid publishing new docs information before the features are available.

packages/components/src/components/hds/form/masked-input/base.gts
```diff
@@ -172,6 +178,7 @@ export default class HdsFormMaskedInputBase extends Component<HdsFormMaskedInput
         @ariaLabel={{this.visibilityToggleAriaLabel}}
         @ariaMessageText={{this.visibilityToggleAriaMessageText}}
         aria-controls={{this.id}}
+        aria-pressed={{this.visibilityToggleAriaPressed}}
```
annawanggg: @MelSumner @shleewhite When testing using VoiceOver, the screenreader correctly announces it as a "toggle button" now. However, it says “selected” when `aria-pressed=true`, so it uses checkbox terminology and VoiceOver says its a checkbox when telling the user how to interact with it. It doesn’t announce the state at all for `aria-pressed=false`. 

I've compared my implementation with VoiceOver's behavior on various toggle button with `aria-pressed` examples, such as from the [APG](https://www.w3.org/WAI/ARIA/apg/patterns/button/examples/button/), where I've gotten the exact same behavior. 

There's no VoiceOver specific documentation or discourse online, but I'm thinking that this is a VoiceOver quirk. I wanted to double check with you guys on whether we should leave it since (I believe) the implementation falls in line with best practices, or if we need to find a workaround. Thanks!

.changeset/big-plums-sing.md
```diff
@@ -0,0 +1,19 @@
+---
+"@hashicorp/design-system-components": minor
+---
+
+<!-- START components/form/masked-input -->
+`Form::MaskedInput` - Fixed a few accessibility issues in `Form::MaskedInput::Base` and `Form::MaskedInput::Field`, including:
+- Changed `@visibilityToggleAriaLabel` to have a static value for visible and hidden states. Added the `aria-pressed` attribute to the visibility toggle button to communicate state.
+- Added `@visibilityToggleAriaMessageTextWhenVisible` argument to set a custom message for the visible state.
```
MelSumner, line 8: Question: Why? The SR should hear something like "icon name, switch, pressed" or "icon name, switch, not pressed" for a standard experience.

annawanggg: The current experience when I test in showcase is "aria label value, selected, toggle button" and "aria label value, *nothing read for not pressed*, toggle button", where the value is something like "Toggle password" or "Toggle visibility". Should I have it read the icon name instead of the aria label? 

Also, I left a [comment above](https://github.com/hashicorp/design-system/pull/4118#discussion_r3982780505) with some questions/concerns since my VoiceOver reads "selected" for `aria-pressed=true` and nothing at all for false, rather than "pressed" and "not pressed", even when I test on other toggle button examples online. Thanks!!

PR #4147, opened 2026-09-18
pnpm 12

Dhaulagiri:
### :link: External links

Jira ticket: https://hashicorp.atlassian.net/browse/HDS-6689


package.json
```diff
@@ -18,44 +18,9 @@
     "dotenv": "^16.4.7",
     "prettier-plugin-ember-template-tag": "^2.0.5"
   },
-  "packageManager": "pnpm@10.11.0+sha512.6540583f41cc5f628eb3d9773ecee802f4f9ef9923cc45b69890fb47991d4b092964694ec3a4f738a420c918a333062c8b925d312f42e4f0c263eb603551f977",
+  "packageManager": "pnpm@12.4.2",
```
zamoore, line 21: [question] Is it ok that we're dropping the integrity hash?

Dhaulagiri: good catch, that got dropped for some reason but I've put it back

